PR #291, opened 2026-02-19
fix(viewer): localize game history heading

jeong-sik:
## Summary
- change viewer footer panel title from Turn History to 게임 히스토리

## Why
- align the in-viewer heading text with Korean UI wording

## Validation
- static HTML text change only (no runtime behavior change)


jeong-sik:
Superseded by #290 — Korean title localization already included in commit 426b524.

PR #355, opened 2026-02-20
fix(viewer): resolve TRPG turn sync mismatch for hybrid state payload

jeong-sik:
## Summary
- fix TRPG state normalization when `/api/v1/trpg/state` returns a hybrid payload (`room` + `state`)
- backfill missing/empty `room` fields from `state` data so turn/phase/status remain consistent
- keep actor-control inference scoped to current turn to avoid stale keeper mappings
- add regression tests for hybrid payload fallback and stale narration keeper mapping

## Problem
In some responses, `room` was present but mostly empty while fresh values lived under `state`.
The viewer parsed `room` first and ignored `state`, causing UI drift like `turn 1≠40`.

## Validation
- `cargo test normalize_hybrid_shape_backfills_room_from_state -- --nocapture`
- `cargo test normalize_masc_shape_ignores_stale_narration_keeper_mapping -- --nocapture`

### DIFF
--- a/viewer/src/game/http.rs
+++ b/viewer/src/game/http.rs
@@ -50,6 +50,10 @@ pub struct SkillData {
     pub name: String,
     #[serde(default = "default_skill_level")]
     pub level: i32,
+    #[serde(default)]
+    pub description: String,
+    #[serde(default)]
+    pub usage_hint: String,
 }
 
 fn default_skill_level() -> i32 {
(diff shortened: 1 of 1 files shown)


jeong-sik:
AI 리뷰 triage 완료:\n- 현재 PR 기준 actionable P1/P2/P3 인라인 코멘트: 0건\n- 알림성 코멘트(quota/review-skipped)는 대응 대상에서 제외\n

PR #392, opened 2026-02-21
fix(trpg): improve fallback round liveliness and hp dynamics

jeong-sik:
## Why
Local TRPG play was stalling into repetitive placeholder narration when keeper replies were unavailable or meta-shaped.

## What changed
- Make player fallback replies deterministic but varied using actor skills/traits/turn seed.
- Add DM fallback narration that reflects round pressure (instead of a fixed one-liner).
- Add NPC counterattack events during DM fallback (`combat.attack` + `hp.changed`) so player HP can change even on fallback-only rounds.
- Treat placeholder keeper replies (e.g. `상황을 살피며 다음 행동을 준비합니다.`) as non-playable and replace them with role-specific fallback actions.

## Validation
- `dune build`
- `dune exec ./test/test_tool_trpg_coverage.exe`
- Manual playtest (isolated room): 3 rounds with unavailable keepers + local fallback
  - turn advanced each round
  - fallback replies varied by skill/trait
  - player HP decreased via NPC counterattacks

### DIFF
--- a/lib/tool_trpg.ml
+++ b/lib/tool_trpg.ml
@@ -1341,22 +1341,148 @@ let ensure_round_npc_spawn_event ~base_dir ~room_id ~turn ~state =
     in
     Ok (Some event)
 
-let fallback_player_reply ~state ~actor_id =
-  let skills =
-    match actor_json_of_state state actor_id with
-    | Some actor_json -> (
-        match actor_json |> member "skills" with
-        | `List xs ->
-            xs
-            |> List.filter_map (function
-                 | `String s when String.trim s <> "" -> Some (String.trim s)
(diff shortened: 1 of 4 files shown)


jeong-sik:
2nd iteration pushed for repeated-play/multi-team liveliness.

What changed:
- Rotating NPC target selection by turn to distribute pressure across players.
- DM fallback now counterattacks even when a new NPC is spawned in the same fallback path (state re-derive after spawn).
- Added regression test: `local fallback distributes pressure across multi-player team`.

Local validation:
- `dune build`
- `dune exec ./test/test_tool_trpg_coverage.exe -- -v` (27 tests, all pass)

PR #518, opened 2026-03-04
Implement real mode switching and enforce mode-gated tool visibility

jeong-sik:


jeong-sik (commented):
Self-review complete on commit a1f870d: mode switch persistence, mode gating, placeholder visibility, and related tests verified locally. Cross-model GLM-5 review completed; no actionable high/critical findings after triage (details in PR comment).

PR #522, opened 2026-03-06
Add operator control plane v1

jeong-sik:
## Summary
- add a guided operator facade for room/session/keeper control via MCP and HTTP
- add the dashboard Ops tab and built dashboard assets for operator workflows
- add coverage for operator snapshot, action, and confirm flows

## Verification
- dune runtest --root .
- cd dashboard && npx tsc --noEmit --pretty
- local HTTP smoke: /dashboard assets, /api/v1/operator snapshot, broadcast, team_turn, task_inject confirm


jeong-sik:
Cross-model review evidence

- Model: GLM-5
- Timestamp: 2026-03-06 13:18 KST
- Scope: PR diff reviewed with generated dashboard JS assets excluded

Findings summary
- Actionable: HTTP operator mutation endpoints were guarded with read-level auth instead of broadcast-level auth. Patched in `be4645f` (`fix(operator): require broadcast auth for HTTP actions`).
- Reviewed and not adopted: a hook-placement warning in `dashboard/src/components/ops.ts` was a false positive after line-by-line verification; `useEffect` is only used inside the component.
- Reviewed and not adopted: H2 `GET /api/v1/operator` now mirrors the HTTP auth behavior under strict mode.

Current status
- No remaining actionable findings from the external review pass.
- Local verification rerun after the auth patch: `dune build --root . @default`, `dune runtest --root .`.

PR #526, opened 2026-03-06
fix(dashboard): preserve structured board event metadata

jeong-sik:
## Summary
- preserve structured board event metadata in dashboard journal entries instead of collapsing everything into text heuristics
- make activity/journal rendering use structured event type, author, preview, and post id data for board events
- harden the dashboard activity smoke harness to assert labeled post/comment previews and author visibility

## Verification
- npm run build
- dune build --root .
- ./_build/default/test/test_web_dashboard_coverage.exe
- attempted scripts/harness_dashboard_activity_smoke.sh but this machine does not have a local playwright module installed


jeong-sik:
Cross-model review: GLM-5, 2026-03-06 18:49 KST.

Summary: reviewed source diff excluding generated dashboard bundle. No actionable high/critical findings.

Notes:
- GLM-5 raised a TypeScript type note around `classifyJournalKind`, but the implementation returns a narrower subtype that is assignable to the caller and the build passed.
- GLM-5 also noted unknown-event display behavior, but this PR intentionally preserves raw text for unmapped events while classifying them as `system`; no regression was identified in local verification.
- Result: no patch required from external review.

PR #587, opened 2026-03-08
fix(dashboard): auto-recover after transient API failures

jeong-sik:
## Summary
- refresh dashboard snapshot periodically even after transient disconnects
- refresh the active tab surface on a periodic interval so command/ops views recover after backend restarts
- rebuild dashboard assets

## Validation
- npm ci
- npx tsc --noEmit --pretty false
- npm run build
- live checks on /api/v1/command-plane and /api/v1/operator after server restart


jeong-sik:
Cross-model review evidence (GLM-5, 2026-03-08 KST)

Result: no blocking findings after manual adjudication.

GLM raised three concerns:
1. stale tab closure risk
2. leaked interval risk
3. duplicate refresh pressure

Adjudication:
- stale closure does not apply here because the interval callback reads `route.value.tab` on each tick instead of capturing a fixed tab value.
- interval cleanup is present; the new `useEffect` returns `clearInterval(interval)`.
- duplicate refresh pressure is real but intentional in this small fix: the goal is recovery after transient backend restart/failure, not fetch minimization. The current tradeoff is one 10s dashboard refresh plus a 15s active-tab refresh. That is acceptable for the local operator dashboard and can be optimized separately if it shows up in traces.

Local validation:
- `npm ci`
- `npx tsc --noEmit --pretty false`
- `npm run build`
- live checks on `/api/v1/command-plane` and `/api/v1/operator`

PR #590, opened 2026-03-08
fix(swarm): ignore phantom supervised lane traces

jeong-sik:
## Summary
- stop marking the supervised lane present from stale operator traces alone
- avoid emitting stale-data flags for lanes that are not actually present
- add swarm-status regression coverage for trace-only and real-session supervised lanes

## Validation
- ./_build/default/test/test_swarm_status.exe
- ./_build/default/test/test_command_plane_v2.exe
- ./_build/default/test/test_operator_control.exe

### DIFF
--- a/lib/swarm_status.ml
+++ b/lib/swarm_status.ml
@@ -620,7 +620,7 @@ let lane_flags kind ~present ~approvals ~workers ~trace_count ~last_movement_at
   in
   let flags =
     match parse_timestamp last_movement_at with
-    | Some ts when Time_compat.now () -. ts > stale_window_sec ->
+    | Some ts when present && Time_compat.now () -. ts > stale_window_sec ->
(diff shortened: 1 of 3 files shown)


jeong-sik:
Cross-model review evidence (GLM-5, 2026-03-08 KST)

Result: NO_FINDINGS

Notes:
- Re-checked the supervised lane presence change after clarifying the intent in code: historical operator/team-session traces must not make the supervised lane present once the real session has ended.
- Added a regression for the complementary case so a real stale supervised session still keeps the `stale_data` flag and `masc_team_session_status` recommendation.

Local validation:
- `dune build --root . @default`
- `./_build/default/test/test_swarm_status.exe`
- `./_build/default/test/test_command_plane_v2.exe`
- `./_build/default/test/test_operator_control.exe`

PR #614, opened 2026-03-09
feat(team-session): add hierarchical 35/27/9 middle controller

jeong-sik:
## Summary
- add hierarchical quality control metadata and controller visibility to team-session/local64
- expand hybrid routing from 35b/9b to 35b/27b/9b with 27b as the middle-control tier
- update local64 smoke harnesses for explicit 35/27/9 runtime inventories

## Validation
- dune build --root . @default
- dune exec --root . ./test/test_local_runtime_pool.exe
- dune exec --root . ./test/test_operator_control.exe
- dune exec --root . ./test/test_tool_team_session.exe
- live local64 smoke with Qwen3.5 35B + 27B + 9B: PASS

## Notes
- left local unrelated changes in root worktree out of this PR (.github/workflows/ci.yml, scripts/opam-pin-external-deps.sh)
- 27B runtime artifact used: Qwen3.5-27B-UD-Q4_K_XL.gguf

PR #616, opened 2026-03-09
feat(team-session): add hierarchical 35/27/9 controller stack

jeong-sik:
## Summary
- add hierarchical controller metadata and visibility for team-session/local64
- expand hybrid routing to explicit 35b / 27b / 9b tiers
- route middle-manager / knowledge roles to 27b and bulk executors to 9b
- update local64 smoke harnesses for hybrid routing and 35/27/9 inventories

## Validation
- dune exec --root . ./test/test_operator_control.exe
- dune exec --root . ./test/test_tool_team_session.exe
- live local64 smoke with explicit `35B + 27B + 9B`: PASS
- same-box `64-worker` run started on `35 + 27 + 9 + 9` stack; partial progress observed but not yet fully green

## Notes
- this PR supersedes #614, which is based on a dirty historical branch
- current PR is based on a clean branch from `origin/main`


jeong-sik:
Updated follow-up on local64 64-worker rerun:

- Fixed batch timeout propagation in `masc_team_session_step`: top-level `spawn_timeout_seconds` now flows into `spawn_batch` items unless an item overrides it.
- Added regression coverage in `test/test_tool_team_session.ml`.
- Fresh 64-worker rerun on `35B@8085 + 27B@8086 + 9B@8087 + 9B@8088` no longer shows the old `300s` per-item timeout failure pattern.
- New bottleneck is liveness under high fan-out: active workers are being reaped by `ZeroZombie` after joining but before their next tool call on long-running 9B turns.
- Added a local64 harness mitigation on the clean branch: `MASC_ZOMBIE_THRESHOLD_SEC=1800` in `scripts/harness_team_session_local64_smoke.sh` so the next fresh rerun can measure throughput instead of getting clipped by default zombie cleanup.

Relevant commits:
- `9b36a07f` fix(team-session): inherit batch spawn timeout
- `390a8c1a` chore(harness): relax zombie threshold for local64 smoke

PR #618, opened 2026-03-09
feat(command-plane): surface hot swarm runtime blockers

jeong-sik:
## Summary
- add runtime viability fields and blockers to the command-plane swarm read model and dashboard
- harden the live harness to persist runtime doctor artifacts and failure summaries
- update hot-swarm runbooks to lock the qwen35-hot startup contract and no-downgrade rule

## Testing
- dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/hot-swarm-stabilization
- ./_build/default/test/test_command_plane_v2.exe
- ./_build/default/test/test_agent_swarm_live_harness.exe
- bash -n scripts/harness/workload/agent_swarm_live.sh
- npm ci
- ./node_modules/.bin/tsc --noEmit --pretty false
- npm run build


jeong-sik:
Follow-up: this branch was rebuilt on top of current `origin/main` and force-updated after the initial stale-local-main base issue. The PR diff should now reflect only the hot-swarm runtime blocker changes plus the fresh dashboard asset rebuild from the clean base.

PR #625, opened 2026-03-09
fix(local64): stabilize 64-worker mixed smoke

jeong-sik:
## Summary
- add local worker liveness protection with automatic heartbeat
- add text-tool-call fallback and auto-note preservation for local llama workers
- reshape local64 hybrid smoke into controller-first waves with proof-friendly summaries
- keep local64 smoke from being clipped by default zombie cleanup while collecting evidence

## Validation
- dune exec --root . ./test/test_local_agent_eio_coverage.exe
- dune exec --root . ./test/test_spawn_eio_coverage.exe
- dune exec --root . ./test/test_tool_team_session.exe
- fresh mixed-runtime 64-worker smoke on 35B+27B+9B+9B inventory


jeong-sik:
Post-proof validation on branch head `8de49289`:

- `dune exec --root . ./test/test_local_agent_eio_coverage.exe`
- `dune exec --root . ./test/test_spawn_eio_coverage.exe`
- `dune exec --root . ./test/test_tool_team_session.exe`

This validation was run after the final `64-worker` PASS artifact and matches the current PR head.

PR #656, opened 2026-03-09
feat(dashboard): add mission dashboard flow

jeong-sik:
## Summary
- add the new dashboard mission read-model at `/api/v1/dashboard/mission`
- make `mission` the new landing flow and keep deep-link aliases like `overview -> mission`, `ops -> intervene`
- reshape `intervene` around room / session / keeper actions and move the rail away from heavy inline controls
- switch `command` default entry to the operations surface and add clearer current-surface guidance

## Testing
- npm run build
- dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat-dashboard-mission-v1
- browser smoke: `#mission`, `#intervene`, `#command`
- local API smoke: `GET /api/v1/dashboard/mission` on the branch server


jeong-sik:
Board curation follow-up

- Added first-class board metadata: `intent`, `source_kind`, `noise_state`, `response_expected`
- Added `masc_board_compose` as the LLM+guardrail write path; raw `masc_board_post` now defaults to legacy/quarantined semantics unless metadata is explicit
- Curated Board feed now hides legacy/test/quarantined noise by default, while Activity keeps raw board-origin rows visible
- Producer wiring updated for keeper/lodge worker paths; fixed follow-up bug where compose was not carrying `thread_id`

Validation:
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat-dashboard-mission-v1`
- `npm ci && npm run build`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat-dashboard-mission-v1 ./test/test_tool_board_coverage.exe` (green before the final `thread_id` follow-up; that follow-up was a narrow schema carry fix and backend rebuild stayed green)
- browser smoke on isolated port confirmed Board defaulted to curated feed and showed an empty curated list instead of raw noise when only legacy/test data existed

External review note:
- GLM-5 review on this board diff surfaced one concrete issue: compose was not carrying `thread_id`; patched in `e55ddfce`. No additional material findings were kept from that pass.

PR #659, opened 2026-03-10
fix: harden Railway deploy runtime contract

jeong-sik:
## Summary
- add missing Linux runtime libraries for the prebuilt Railway binary
- verify the deployment artifact is a Linux ELF in CI
- smoke test the Docker image and /health before running railway up

## Verification
- docker build -t masc-mcp-railway-smoke .
- local container run reproduced exec format error when fed a macOS binary, which this CI guard now catches before deploy

### DIFF
--- a/.github/workflows/deploy-railway.yml
+++ b/.github/workflows/deploy-railway.yml
@@ -51,6 +51,33 @@ jobs:
           chmod +x masc-mcp-linux-x64
           ls -la masc-mcp-linux-x64
 
+      - name: Verify Linux binary format
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


jeong-sik:
Cross-model review evidence (Claude fallback, 2026-03-10): no high-severity bugs found in the Railway deploy hardening diff. Reviewer noted one residual risk: `/health` smoke alone may not catch every missing shared library. Patched in follow-up commit `85922d2b` by adding a container-side `ldd /app/masc-mcp` not-found check before Railway upload. Current gate status to verify on this head: CI rerun + Deploy to Railway workflow_dispatch rerun.

PR #669, opened 2026-03-10
fix: address intent forecast review regressions

jeong-sik:
## Summary
- resolve intent forecast dependency checks against the full operation set instead of only linked intent operations
- preserve terminal intent state when late checkpoints arrive on completed/cancelled/failed operations
- add regression coverage for cross-intent dependencies and terminal checkpoint preservation

## Context
- follow-up to merged PR #668
- addresses the two unresolved review threads left on #668

## Testing
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-intent-review-followup`
- `timeout 120 dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-intent-review-followup ./test/test_command_plane_v2.exe`

### DIFF
--- a/lib/command_plane_v2.ml
+++ b/lib/command_plane_v2.ml
@@ -4314,6 +4314,14 @@ let verification_state_of_operation (operation : operation_record) =
   | _, Some "implement" -> Some "implementing"
   | _ -> None
 
+let intent_state_hint_of_operation_status (operation : operation_record) =
+  match operation.status with
+  | Planned | Active -> Active_intent
+  | Paused -> Suspended_intent
+  | Completed -> Completed_intent
+  | Cancelled -> Dropped_intent
+  | Failed -> Blocked_intent
+
 let focus_of_operation (operation : operation_record) =
   {
     stage = operation.stage;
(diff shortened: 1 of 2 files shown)


jeong-sik:
Cross-model review evidence

- Reviewer model: GLM-5
- Review time: 2026-03-10 12:03 KST
- Review basis: diff against `main` for `lib/command_plane_v2.ml` and `test/test_command_plane_v2.ml`
- Result: no high/critical correctness regressions identified in the intent forecast / terminal checkpoint fix itself.
- Medium finding from review: cross-intent dependency edge cases were under-specified in regression coverage.
- Follow-up applied in `e3aca237`:
  - add regression test that active cross-intent upstream blocks forecast
  - add regression test that checkpointed cross-intent upstream clears forecast blocking
- Local verification after patch:
  - `dune build --root . ./test/test_command_plane_v2.exe ./test/test_cp_search_fabric.exe`
  - `./_build/default/test/test_command_plane_v2.exe`
  - `./_build/default/test/test_cp_search_fabric.exe`

PR #728, opened 2026-03-11
fix(dashboard): hide keeper internal notes in execution view

jeong-sik:
## Summary
- stop leaking keeper internal memory notes into the execution focus row
- translate execution surface copy to Korean for operator-facing readability
- refresh dashboard build artifacts

## Verification
- npm ci
- npx tsc --noEmit --pretty
- npm run build


jeong-sik:
Replaced by fix/hide-keeper-internal-notes to resolve Vite asset conflicts

PR #756, opened 2026-03-11
refactor: extract command plane http routes

jeong-sik:
## Summary
- extract the command-plane HTTP/H2 route family from `bin/main_eio.ml` into `lib/server_command_plane_http.ml`
- keep route wiring and response semantics intact by replacing the old in-file implementations with thin wrappers
- start the planned `main_eio` god-module split with the most self-contained route family first

## Verification
- `dune build --root . _build/default/bin/main_eio.exe`
- `_build/default/bin/main_eio.exe --help`
- `_build/default/test/test_mcp_server_eio.exe`
- `_build/default/test/test_http_server_eio_coverage.exe`

## Follow-up
- next slices: MCP transport handlers, GraphQL/TRPG SSE handlers, then final route/bootstrap thinning

### DIFF
--- a/bin/main_eio.ml
+++ b/bin/main_eio.ml
@@ -44,6 +44,7 @@ module Tool_mdal = Masc_mcp.Tool_mdal
 module Tool_board = Masc_mcp.Tool_board
 module Process_eio = Masc_mcp.Process_eio
 module Mdal = Masc_mcp.Mdal
+module Server_command_plane_http = Masc_mcp.Server_command_plane_http
 
 (** MCP Protocol Versions *)
 (* ============================================ *)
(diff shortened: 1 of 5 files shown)


jeong-sik:
Updated slice on .

This push extracts MCP transport HTTP handlers and session/SSE bookkeeping out of  into , while keeping  as route wiring plus shared helper reuse for H2/extended handler paths.

Local verification:
- 
- 
- 
- 

Notes:
-  appeared to hang in this environment, but the built test binary completed successfully when run directly.
- Remaining PR1 slices are GraphQL/TRPG handler extraction and final route/bootstrap thinning.

PR #762, opened 2026-03-11
fix(room): preserve default join paths in room guards

jeong-sik:
## Summary
This follow-up restores room guard behavior that was still missing after #754 merged.

The change keeps default-room join files and room-scoped guard checks consistent by:
- resolving room agent names through the same fallback path as default joins
- checking backend-aware room state/backlog existence during room bootstrap for non-default rooms
- making joined-status checks use the resolved agent path before applying room/root fallback

## Why
The merged keeper/persona work left one follow-up gap: default joins still persisted under root `.masc/agents`, while some room-scoped checks had moved to room-local paths. That could cause false "not joined" or missed heartbeat behavior, and backend-backed room bootstrap could still reinitialize room state for non-default rooms.

## Verification
- `dune build --root . _build/default/test/test_room.exe _build/default/test/test_tool_room_coverage.exe _build/default/test/test_mcp_server_eio.exe`
- `./_build/default/test/test_room.exe`
- `./_build/default/test/test_tool_room_coverage.exe`
- `./_build/default/test/test_mcp_server_eio.exe`

## Notes
- This PR contains only commit `722021f6` on top of current `main`.
- It is intended as the missing follow-up after merged PR #754.

### DIFF
--- a/lib/room.ml
+++ b/lib/room.ml
@@ -278,9 +278,11 @@ let resolve_agent_name_in_room config ~room_id agent_name =
         String.sub f 0 (String.length prefix) = prefix
       ) files with
       | Some file -> String.sub file 0 (String.length file - 5)
-      | None -> agent_name
+      | None ->
+          (* Default joins still persist under the root agents dir. *)
+          resolve_agent_name config agent_name
     else
-      agent_name
+      resolve_agent_name config agent_name
   end
 
 let ensure_room_bootstrap config room_id =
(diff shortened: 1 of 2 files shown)


jeong-sik:
Verification for commit 722021f6:
- dune build --root . _build/default/test/test_room.exe _build/default/test/test_tool_room_coverage.exe _build/default/test/test_mcp_server_eio.exe
- ./_build/default/test/test_room.exe
- ./_build/default/test/test_tool_room_coverage.exe
- ./_build/default/test/test_mcp_server_eio.exe

Results:
- test_room: pass (77 tests)
- test_tool_room_coverage: pass
- test_mcp_server_eio: pass (32 tests)

jeong-sik:
Cross-model review attempt for commit 722021f6 did not produce a verdict in this environment.

Attempted reviewer paths:
- Claude CLI (`claude -p --model opus`) with the patch prompt piped on stdin
- `/Users/dancer/me/scripts/review/4.analysis/ccs-cascade-review.sh --prompt-file /tmp/pr762_review_prompt.txt`

Observed failure mode:
- reviewer bootstrap reached CLI proxy, then exited with `Input must be provided either through stdin or as a prompt argument when using --print`
- no findings payload was produced, so there is no external review verdict to record yet

PR #834, opened 2026-03-12
feat(dashboard): add room-wide orchestra map

jeong-sik:
## Summary
- add a room-wide `orchestra` command-plane surface for visual swarm/session/worker/keeper monitoring
- expose a new `/api/v1/command-plane/orchestra` read model for the dashboard
- keep existing `swarm` and `warroom` surfaces intact while adding a new overview-first visualizer

## Testing
- dune build --root .
- file-scoped TypeScript validation for changed command/dashboard files

## Notes
- dedicated orchestra smoke test is not included in this slice because the initial Eio/Alcotest version hung and needs separate follow-up

### DIFF
--- a/bin/main_eio.ml
+++ b/bin/main_eio.ml
@@ -5684,6 +5684,10 @@ let command_plane_swarm_http_json ~state request =
   Server_command_plane_http.command_plane_swarm_http_json
     ~deps:command_plane_http_deps ~state request
 
+let command_plane_orchestra_http_json ~state request =
+  Server_command_plane_http.command_plane_orchestra_http_json
+    ~deps:command_plane_http_deps ~state request
+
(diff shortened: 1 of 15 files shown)


jeong-sik:
Follow-up on test coverage (2026-03-12 Asia/Seoul)

- Added deterministic orchestra coverage in `test/test_operator_control.ml`
- New checks cover:
  - orchestra room core shape
  - orchestra session edge + pending-confirm signal
- Local validation on updated branch head `d144f096`:
  - `dune build --root .`
  - `./_build/default/test/test_operator_control.exe`

Note:
- The earlier dedicated `test_command_plane_orchestra` smoke was replaced with these operator-suite checks because the standalone version hung under the initial Eio/Alcotest arrangement.

PR #841, opened 2026-03-12
test(command): add orchestra regression coverage

jeong-sik:
## Summary
- add deterministic operator-suite coverage for the orchestra read model
- verify room-core shape and session/pending-confirm wiring in the merged orchestra surface
- keep the coverage inside  instead of the earlier hanging standalone smoke

## Validation
- dune build --root . test/test_operator_control.exe
- ./_build/default/test/test_operator_control.exe

### DIFF
--- a/test/test_operator_control.ml
+++ b/test/test_operator_control.ml
@@ -246,6 +246,85 @@ let test_snapshot_pending_confirm_summary_tracks_actor_scope () =
              = "task_inject")
            confirm_required_actions))
 
+let test_orchestra_room_core_shape () =
+  Eio_main.run @@ fun env ->
+  Eio.Switch.run @@ fun sw ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review evidence (2026-03-12, Asia/Seoul)
- Requested reviewer: GLM-5
- Fallback reason: llm-mcp GLM transport failed against http://127.0.0.1:8932/mcp during review setup
- Fallback reviewer: local llama-server model qwen3.5-35b-a3b-ud-q8-xl on http://127.0.0.1:8085
- Scope: test/test_operator_control.ml diff for orchestra regression coverage
- Result: No material findings.

Local validation
- dune build --root . test/test_operator_control.exe
- ./_build/default/test/test_operator_control.exe

PR #843, opened 2026-03-12
fix(sentinel): bootstrap prompt registry on startup

jeong-sik:
## Summary
- bootstrap prompt and chain registries during server startup
- ensure sentinel can resolve built-in `sentinel-*` prompts before its consumers fire

## Verification
- dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-sentinel-prompt-bootstrap ./bin/main_eio.exe

Follow-up to runtime truth audit: sentinel was starting, but `sentinel-board-patrol` prompt lookup failed at startup because `Prompt_registry` had not been bootstrapped yet.


jeong-sik:
Cross-model review evidence
- Model: local llama-server `qwen3.5-35b-a3b-ud-q8-xl`
- Timestamp: 2026-03-12 KST
- Scope: startup bootstrap fix for sentinel prompt registry initialization
- Result: No findings.

PR #874, opened 2026-03-12
feat(dashboard): add global tools inventory surface

jeong-sik:
## Summary
- add a dedicated dashboard Tools surface backed by /api/v1/dashboard/tools
- show the full system tool inventory by default, including hidden and deprecated tools
- clarify keeper/agent tool audit semantics so allowed tools, observed tools, and empty states are truthful

## Validation
- dune exec ./test/test_dashboard_tools.exe
- dune exec ./test/test_dashboard_mission.exe
- dune exec ./test/test_dashboard_execution.exe
- dashboard tsc --noEmit


jeong-sik:
Cross-model review evidence (GLM-5, 2026-03-12 KST)

- Reviewer path: /Users/dancer/me/scripts/ccs-cascade-review.sh --include-glm
- Scope: local diff for PR #874
- Findings: 4 truthfulness issues were raised and patched in 66f18570
- Patched items:
  - offline allowlist empty state now reports offline instead of not_collected
  - keeper detail current-task / skill-route fallbacks now use runtime truth, not raw status only
  - observed-tool fallback no longer fabricates tool-call counts from distinct tool names
  - newer heartbeat tasks now preserve the last observed tool evidence instead of blanking it
- Note: the cascade script failed after the GLM stage while moving on to additional reviewers, so this evidence is GLM-only rather than full cascade output

Validation after the patch:
- dune exec ./test/test_dashboard_mission.exe
- dune exec ./test/test_operator_control.exe
- ./_build/default/test/test_dashboard_tools.exe
- dune exec ./test/test_dashboard_execution.exe
- dashboard tsc --noEmit

PR #896, opened 2026-03-12
fix(dashboard): clarify keeper runtime naming

jeong-sik:
## Summary
- clarify that continuity and mission cards use keeper names as the canonical title
- show runtime agent names as secondary labels when they differ from the keeper key
- rename ambiguous dashboard copy so sangsu vs keeper-sangsu-agent is easier to understand

## Validation
- dashboard tsc --noEmit


jeong-sik:
External review evidence (GLM-5, 2026-03-12 KST)

Scope: incremental update `34779e10 -> 4283c578`. Change scope is limited to `.github/workflows/ci.yml`.

Observed failure addressed:
- `Health` job failed during `Install dependencies` because `opam install . --deps-only --with-test --yes` resolved an incompatible `bisect_ppx` set for OCaml `5.4.1`.

Patch:
- Align `health` with `build` by pinning external deps via `scripts/opam-pin-external-deps.sh --with-bisect` before dependency installation.

Result: no concrete high/medium findings surfaced for this CI-only fix.

Local verification:
- workflow diff check only (`git diff --check`)
- note: this branch still has pre-existing quick-suite failures in local targeted runs (`test_tool_contract_truth.exe`, `test_silent_failures_p2a.exe`), which appear unrelated to the dashboard-only source diff and need to be re-evaluated against the new CI head.

PR #905, opened 2026-03-12
fix: narrow MCP resource update notifications

jeong-sik:
## Summary
- stop broadcasting `resources/updated` to every dynamic subscribed URI after any write tool
- map write tools to affected MASC resource families and only notify matching subscribers
- add SSE-backed tests for broadcast vs task-write notification precision

## Validation
- `DUNE_BUILD_DIR=_build_pr2 dune build --root . @check test/test_mcp_server_eio.exe`
- `./_build_pr2/default/test/test_mcp_server_eio.exe`

### DIFF
--- a/lib/mcp_server_eio.ml
+++ b/lib/mcp_server_eio.ml
@@ -3010,14 +3010,69 @@ let send_resource_updated_notification ~session_id ~uri =
 let broadcast_tools_list_changed () =
   Sse.broadcast (jsonrpc_notification "notifications/tools/list_changed")
 
+let dedup_strings items =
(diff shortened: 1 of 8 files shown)


jeong-sik:
Validation

- `DUNE_BUILD_DIR=_build_pr2 dune build --root . @check test/test_mcp_server_eio.exe`
- `./_build_pr2/default/test/test_mcp_server_eio.exe`

Notes
- Added precision tests proving broadcast writes notify message/event subscribers but not task subscribers, and task writes notify task/event subscribers but not message subscribers.

PR #969, opened 2026-03-13
fix(dashboard): preserve observability truth compatibility

jeong-sik:
## Summary

Bring the observability-truth dashboard line to `main` and add a compatibility follow-up so the refined payload semantics do not break existing dashboard consumers.

## What changed

- Carries forward the observability-truth dashboard work that was previously merged only on stacked feature branches
- Adds a follow-up compatibility fix for the dashboard/runtime payloads:
  - split lodge reasons into `last_pass_reason` vs `last_system_skip_reason`
  - keep the simplified observability wording in UI cards
  - restore optional TypeScript DTO fields that existing views still reference
  - preserve dashboard/build compatibility while the runtime payload shape evolves
- Rebuilds tracked dashboard assets under `assets/dashboard/`

## Why

The local follow-up diff was coherent, but the initial DTO pruning was too aggressive:

- `dashboard/src/types.ts` removed optional fields that other dashboard components still imported
- `npm run build` failed because governance/mission/operator/detail views still relied on those optional properties

This patch keeps the improved wording and the more explicit pass-vs-system-skip semantics, but widens the TS surface back to a compatible optional shape so the dashboard builds and existing readers do not break.

## Validation

- `npm run build` in `dashboard/`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat-observability-truth-v2`
- `./_build/default/test/test_dashboard.exe`
- `git diff --check`

## Notes

- This PR includes the observability-truth line plus the compatibility follow-up because those earlier stacked branches were merged into feature branches, not `main`.


jeong-sik:
Superseded by clean main replay PR #974: https://github.com/jeong-sik/masc-mcp/pull/974

PR #974, opened 2026-03-13
fix(dashboard): preserve observability truth compatibility

jeong-sik:
## Summary

Replay the observability-truth compatibility follow-up cleanly onto `main`.

This PR supersedes `#969`, which was dirty because it was based on stacked feature branches rather than a `main` replay branch.

## What changed

- Preserves the observability-truth compatibility fix on top of current `main`
- Keeps the explicit pass-vs-system-skip lodge semantics:
  - `last_pass_reason`
  - `last_system_skip_reason`
- Keeps the refined dashboard copy/labels
- Restores optional DTO compatibility in `dashboard/src/types.ts` so existing mission/governance/operator/detail components keep compiling while payload evolution continues
- Updates dashboard runtime normalization and execution payloads accordingly
- Rebuilds tracked dashboard assets

## Validation

- `npm run build` in `dashboard/`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/replay-pr969-main`
- `./_build/default/test/test_dashboard.exe`
- `git diff --cached --check`

## Notes

- This is the clean `main` replay branch for the local follow-up that was first parked in `#969`.


jeong-sik:
Validation on `c87a8f0a`:
- `cd dashboard && npm ci && npm run build`
- `DUNE_BUILD_DIR=_build_pr974 dune exec --root . ./test/test_dashboard_execution.exe`
- `DUNE_BUILD_DIR=_build_pr974 /opt/homebrew/bin/timeout 180 dune exec --root . ./test/test_web_dashboard_coverage.exe`
- `git diff --check`

Cross-model review evidence:
- reviewer: local `llama-server` at `http://127.0.0.1:8085`
- model: `qwen3.5-35b-a3b-ud-q8-xl`
- result: no actionable findings
- note: keeping `last_skip_reason` as the legacy alias for system-skip reason is intentional and matches the current lodge/keeper semantics already present on main; this refresh only adds the split fields and dashboard rendering needed to preserve compatibility without rolling back observability-truth fields

PR #1076, opened 2026-03-16
fix: isolate dashboard side rail scrolling

jeong-sik:
## Summary
- make the dashboard side rail independently scroll inside the viewport on desktop
- keep the single-column and mobile breakpoint using the existing document scroll behavior
- add rail overscroll containment and a stable scrollbar gutter to reduce cross-pane scroll bleed

## Verification
- browser validation on the mission dashboard confirmed rail-only scroll changed while body scroll stayed at 0 after scrolling the rail
- attempted local dashboard build and TypeScript validation, but the dashboard tsc step did not complete cleanly in this session
- this patch is therefore validated by browser behavior rather than a completed local build in this run

### DIFF
--- a/dashboard/src/styles/global.css
+++ b/dashboard/src/styles/global.css
@@ -260,6 +260,11 @@ a:hover {
   display: flex;
   flex-direction: column;
   gap: 12px;
+  max-height: calc(100dvh - var(--header-h) - 32px);
+  overflow-y: auto;
+  overscroll-behavior: contain;
+  scrollbar-gutter: stable;
+  padding-right: 4px;
 }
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review evidence
- Timestamp: 2026-03-16 00:03:12 UTC
- Requested model: GLM-5 via sb glm-text
- Fallback model: local llama-server qwen3.5-35b-a3b-ud-q8-xl
- Fallback reason: GLM review request did not return in this session, so the workflow fallback was used
- Scope: correctness, regressions, user-visible behavior risks, missing tests only
- Result: No findings.

PR #1082, opened 2026-03-16
build: refresh dashboard asset bundle for side-rail scroll fix

jeong-sik:
## Summary
- refresh the checked-in dashboard CSS bundle so the served asset matches the already-merged side-rail independent-scroll source change from #1076
- keep this PR asset-only relative to main

## Verification
- browser validation for the source behavior was done before #1076 merged
- local dashboard build and TypeScript validation did not complete cleanly in this session, so the bundle sync was applied directly to the generated CSS asset
- asset-only external review was run with a non-self fallback model and reported no findings


jeong-sik:
Closing as duplicate/no-op. The asset-sync change is already present on main via #1080, so there is no remaining diff against origin/main.

PR #1282, opened 2026-03-16
fix(task): warn and release previous claim on claim_next

jeong-sik:
Closes #1256. claim_next auto-releases previous claim to prevent orphaned tasks.


jeong-sik:
Duplicate of #1278 (already merged). bug-004 fix is on main.

PR #2240, opened 2026-03-22
fix: scope OAS memory fallback and surface more OAS events

jeong-sik:
## Summary
- scope OAS JSONL fallback to the current room .masc before falling back to global ME_ROOT/.masc
- replace episodic flush list dedupe with hashtable tracking and add explicit base-dir coverage
- lower and tune OAS SSE relay latency and surface resident lifecycle, trust, and reputation events in the dashboard

## Verification
- dune build --root .
- npm run typecheck
- ./_build/default/test/test_memory_oas_5tier.exe

### DIFF
--- a/dashboard/src/components/overview/oas-pipeline.ts
+++ b/dashboard/src/components/overview/oas-pipeline.ts
@@ -21,6 +21,40 @@ function actionBadge(action: string | undefined): string {
   }
 }
 
+function eventTypeLabel(type: string): string {
+  switch (type) {
+    case 'selected': return 'selected'
+    case 'decision': return 'decision'
+    case 'action_executed': return 'executed'
+    case 'keeper_resident_lifecycle': return 'resident'
+    case 'trust_updated': return 'trust'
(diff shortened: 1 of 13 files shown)


jeong-sik:
Cross-model review evidence (2026-03-22 Asia/Seoul):

- Primary reviewer: GLM-5 via sb glm-text. Two attempts returned truncated reasoning-only output and did not produce a usable verdict or actionable file/line finding.
- Fallback reviewer: local llama-compatible qwen3.5-35b-a3b-ud-q4-xl on http://127.0.0.1:8085, used because GLM-5 did not yield a usable review verdict.
- Fallback result: no actionable external findings. The only partial concern pointed at dashboard/src/components/overview/oas-pipeline.ts, but its cited null-guard risk was not valid after manual follow-up because old_score/new_score are checked before toFixed.

Local verification:
- dune build --root .
- npm run typecheck
- ./_build/default/test/test_memory_oas_5tier.exe

PR #2427, opened 2026-03-22
fix(dashboard): normalize keeper chat error payload

jeong-sik:
## Summary
- normalize structured keeper chat error payloads instead of rendering [object Object]
- accept both TEXT_MESSAGE_CONTENT and legacy TEXT_DELTA stream events in the legacy keeper chat panel
- add regression tests for event compatibility and error message extraction

## Verification
- npm run typecheck
- npm test -- src/components/keeper-chat-panel.test.ts
- npm run build


jeong-sik:
Cross-model review pending. GLM-5 runner was attempted first and did not return within the initial wait window, so local llama fallback review is running now per repo policy. I will replace this with the final review evidence once the fallback completes.

PR #2439, opened 2026-03-22
refactor(keeper): migrate 12 inference params to OAS cascade (#2408 Phase 3)

jeong-sik:
Remove 12 hardcoded inference functions from keeper_config. Temperature, max_tokens now come from OAS cascade defaults. -131 net lines. All keeper tests pass.


jeong-sik:
Content already in main via rebase merge.

PR #2754, opened 2026-03-24
fix(dashboard): stabilize cached PG dashboard paths

jeong-sik:
## Summary
- stabilize default cached execution/mission/operator/operator-digest paths around shared inline compute and stale-cache diagnostics
- slim execution/operator payloads and add PG/session-list diagnostics, including scoped query prefixes for team session discovery
- fix startup env precedence and JSONL fallback env clearing so explicit Postgres runs do not silently re-read placeholder DB envs

## Validation
- ./_build/default/test/test_server_runtime_bootstrap.exe
- ./_build/default/test/test_swarm_persistence.exe
- ./_build/default/test/test_start_masc_mcp_script.exe
- dune exec --root . ./_build/default/test/test_operator_control.exe
- dune exec --root . ./_build/default/test/test_dashboard_execution.exe
- npm --prefix dashboard run build

## Follow-up
- PG startup can still time out inside the wider background init path and fall back to JSONL on slow clusters; tracked separately as a follow-up and not claimed as fixed in this PR.


jeong-sik:
Superseded by #2765 (merged). This PR had WIP contamination in its worktree.

PR #2828, opened 2026-03-24
fix(dashboard): clarify agent roster labels

jeong-sik:
## Summary
- rename agent/keeper tabs and filter labels to user-facing language
- add inline explanations for status groups, keeper generation, and context usage
- rework roster cards so current work and runtime metadata are readable instead of truncated

## Verification
- npm run typecheck
- npm run test

### DIFF
--- a/dashboard/src/components/agent-roster.test.ts
+++ b/dashboard/src/components/agent-roster.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from 'vitest'
+import type { Agent, Keeper } from '../types'
+import type { DashboardMissionKeeperBrief } from '../types/dashboard-mission'
+import { countAgentsByStatus, scopeAgentsByKeeperFilter } from './agent-roster'
+
+const AGENTS: Agent[] = [
+  { name: 'plain-agent', status: 'active', current_task: null },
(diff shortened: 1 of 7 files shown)


jeong-sik:
Follow-up cross-model review evidence after roster count scope fix.

- Reviewer: GLM-5 via sb glm-text
- Reviewed delta: commit 8706119d (fix(dashboard): scope roster counts to active tab)
- Timestamp: 2026-03-24 18:17 KST
- Result: initial GLM pass surfaced one actionable gap only: missing coverage for keeperFilter='all'. That test was added and rerun. Final GLM retry returned reasoning-only/truncated output without a clean verdict string, but the visible review text did not surface any further material correctness findings after the test patch.
- Local verification:
  - ./node_modules/.bin/tsc --noEmit -p tsconfig.json (run in worktree dashboard with a temporary symlink to the main checkout dashboard/node_modules because this worktree has no local install)
  - ./node_modules/.bin/vitest run src/components/agent-roster.test.ts src/components/agent-profile.test.ts
- Behavioral summary: status chips now count within the active roster tab scope (all / agent-only / keeper-only) instead of leaking totals from the full agent list, and the regression is covered for all three filter modes.

PR #2840, opened 2026-03-24
fix(dashboard): stabilize autoresearch refresh and a11y

jeong-sik:
## Summary
- stabilize autoresearch refresh so in-flight loop/detail requests do not race or overwrite the current selection
- tighten dashboard accessibility on interactive controls, form fields, toast announcements, and document locale
- narrow several `transition: all` rules to explicit properties

## Validation
- `npm run typecheck`
- `npm test`
- `npm run build`

## Notes
- Vite still reports the pre-existing `tab-refresh` dynamic/static import chunking warning during build.
- Cross-model review evidence will be added on the PR thread.

### DIFF
--- a/dashboard/index.html
+++ b/dashboard/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="ko">
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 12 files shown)


jeong-sik:
Cross-model review evidence (2026-03-24 KST)

- Primary reviewer path: `sb glm-text` with GLM-5, attempted twice. Both runs consumed tokens in reasoning and returned no final findings payload, so the result was unusable.
- Fallback 1: local `llama-server` at `http://127.0.0.1:8085` (`qwen3.5-35b-a3b-ud-q4-xl`) probe/review attempts did not return a usable response within the session window.
- Fallback 2: MASC keeper review with `llama:qwen3.5-35b-a3b` timed out after 120s.
- Manual follow-up review after the above failures found no additional high/critical issues beyond the changes already addressed in this PR.

Local validation run in the PR worktree:
- `npm run typecheck`
- `npm test`
- `npm run build`

Known residual: Vite still reports the pre-existing `tab-refresh` dynamic/static import chunking warning during build.

PR #2855, opened 2026-03-24
fix(room): validate current room ids before bootstrap

jeong-sik:
## Summary
- add `Room.validate_room_id` for current-room identifiers
- reject invalid `room_id` values before `Room.write_current_room` and `Room.ensure_room_bootstrap`
- add regression coverage and a source guard for the H2 current-room route

## Context
- follow-up for merged PR #2845
- addresses review concern from https://github.com/jeong-sik/masc-mcp/pull/2845#discussion_r2980686493 that did not land before merge

## Verification
- `dune build --root . -j 1 --display short test/test_ci_hardening_source.exe`

## Notes
- `test/test_types_utils_coverage.exe` was updated but I did not wait for the full relink in this fresh worktree because the cache state fans out into a much larger rebuild than this route-level fix

### DIFF
--- a/lib/room/room_utils_ops.ml
+++ b/lib/room/room_utils_ops.ml
@@ -2,14 +2,6 @@ open Types
 open Room_utils_backend_setup
 open Room_utils_paths_backend
 
-let validate_agent_name name =
-  (* Delegate to Validation module for consistent security checks *)
-  Validation.Agent_id.validate name
-
-let validate_task_id id =
-  (* Delegate to Validation module for consistent security checks *)
-  Validation.Task_id.validate id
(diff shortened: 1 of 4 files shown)


jeong-sik:
Follow-up for #2845. This carries the `room_id` validation fix that was raised in review after the original branch had already been merged.

Local verification:
- `dune build --root . -j 1 --display short test/test_ci_hardening_source.exe`

`test/test_types_utils_coverage.exe` was updated but not fully relinked locally in this fresh worktree because that path fans out into a much larger rebuild than the route patch itself.

@codex review

PR #2939, opened 2026-03-25
fix(keeper): guard board cursor updates

jeong-sik:
## Summary
- restore the missed board cursor concurrency guard from the closed follow-up branch
- protect keeper board cursor reads/writes with `Eio.Mutex`
- capture the board cursor watermark before `list_posts` so posts created during the fetch window are not skipped on the next pass

## Context
- this ports the still-missing `keeper_world_observation` fix from the closed follow-up branch instead of reopening the historical PR
- while landing this, I also cleaned up stale CI symptom issues `#2894` and `#2913` because their original symptoms are already resolved on current `main`

## Verification
- `git diff --check`
- `scripts/ci-run-tests.sh "opam exec -- dune build"` is running locally in the worktree; final result is still pending because the repo currently has long-running concurrent dune builds

### DIFF
--- a/lib/keeper/keeper_world_observation.ml
+++ b/lib/keeper/keeper_world_observation.ml
@@ -148,21 +148,34 @@ let read_continuity_summary ~(config : Room.config) ~(meta : keeper_meta)
 (** Per-keeper cursor for board event collection.
     Tracks the last check timestamp so no posts are missed between heartbeats.
     In-memory only — first run after restart uses a 300s bootstrap window. *)
+let board_cursors_mu = Eio.Mutex.create ()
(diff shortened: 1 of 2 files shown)


jeong-sik:
Cross-model review evidence (2026-03-25 KST):
- direct `sb glm-text` (`GLM-5`) attempted first, but the response returned only truncated reasoning content without a usable reviewer conclusion
- fallback `sb glm-cascade --simple` also failed (`Mcp-Session-Id header required`), so I used the documented local non-self fallback reviewer at `http://127.0.0.1:8085`
- local reviewer: `qwen3.5-35b-a3b-ud-q4-xl`
- verdict: `No findings.`

PR #2949, opened 2026-03-25
fix(cp): overfetch 3x in operator trace tail to prevent under-delivery

jeong-sik:
## Summary
- `recent_operator_trace_events` with `trace_id=None` tail-reads only `limit` raw lines, then filters. Invalid/blank lines cause under-delivery.
- Overfetch 3x raw lines so post-filter result reliably reaches requested limit.

Fixes #2945.

## Test plan
- [x] `dune build` passes
- [ ] CI green

### DIFF
--- a/lib/command_plane/cp_snapshot_summaries.ml
+++ b/lib/command_plane/cp_snapshot_summaries.ml
@@ -571,9 +571,10 @@ let recent_operator_trace_events config ?trace_id limit =
   else
     (match trace_id with
      | None ->
-         (* Over-read to compensate for blank/malformed lines lost during filtering *)
+         (* Overfetch 3x to compensate for lines lost to JSON parse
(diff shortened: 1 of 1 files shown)


jeong-sik:
GLM-5 cross-review completed at 2026-03-25 10:34:04 KST; rebased onto current `main` after #2943 merged.

Scope reviewed: the post-#2943 delta only changes operator trace tail over-read from the merged `x2` heuristic to `x3`.

Findings summary:
- No material correctness findings in the rebased diff.
- Candidate concern from GLM review was whether overfetch could violate the caller `limit`; checked manually against the same function and the final event list is still truncated back to `limit`.
- Local verification: `DUNE_CACHE=disabled opam exec -- dune build --root . @check` passed in the rebased worktree.

Evidence: direct `sb glm-text` review on the patch plus manual inspection of the truncation path in `recent_operator_trace_events`.

PR #2967, opened 2026-03-25
Improve MCP read-path stability and add revalidation harness

jeong-sik:
## Summary
- add short-lived fast-path caching for `masc_status` and `masc_keeper_list` so MCP read tools stay responsive under dashboard background work
- distinguish keeper config-disabled vs runtime-not-running states in continuity diagnostics and expose runtime policy fields in compact keeper surfaces
- cache transport-health independently, bound keeper bootstrap startup work, and add a reusable MCP read-path revalidation harness + runbook

## Verification
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix/mcp-autonomy-fastpath test/test_tool_keeper.exe test/test_tool_room_coverage.exe`
- `./_build/default/test/test_tool_room_coverage.exe`
- `./_build/default/test/test_tool_keeper.exe test read_file_tail_lines 11,12 --compact`
- `START_SERVER=0 MODES=http_only TARGET_BASE_URL=http://127.0.0.1:8946 ./scripts/harness_mcp_readpath_revalidation.sh`
- `START_SERVER=0 MODES=default TARGET_BASE_URL=http://127.0.0.1:8947 ./scripts/harness_mcp_readpath_revalidation.sh`

### DIFF
--- a/docs/MCP-READPATH-REVALIDATION-RUNBOOK.md
+++ b/docs/MCP-READPATH-REVALIDATION-RUNBOOK.md
@@ -1,0 +1,55 @@
+# MCP Read-Path Revalidation
+
+이 runbook은 `masc_status`, `masc_keeper_list`, `masc_transport_status`, dashboard cached surfaces, keeper continuity surface를 함께 재검증할 때 쓴다.
+
+엔트리포인트:
+
+```bash
+./scripts/harness_mcp_readpath_revalidation.sh
+```
+
+기본 모드:
+
+- `http_only`
+  - `MASC_GRPC_ENABLED=0`
+  - `MASC_WS_ENABLED=0`
+  - `MASC_WEBRTC_ENABLED=0`
+  - MCP read-path와 dashboard cache baseline 확인
+- `default`
+  - live HTTP + WebSocket + WebRTC startup 확인
+  - gRPC는 기본적으로 꺼 둔다
+
+기본 base path는 현재 repo root다. 즉 live `.masc` 상태와 resident keeper surface를 같이 검증한다.
+
+## Success Criteria
+
+- `masc_status` 2회 연속 호출이 timeout 없이 끝난다.
+- `masc_keeper_list(detailed=false)` 2회 연속 호출이 timeout 없이 끝난다.
+- `masc_transport_status`가 빠르게 응답한다.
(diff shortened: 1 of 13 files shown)


jeong-sik (commented):
## Code Review Findings

### Critical

**1. `is_internal_race_cancel` — Eio internal string matching (proactive_refresh.ml:25-31)**

```ocaml
String.equal msg "Cancelled: Eio__core__Fiber.Not_first"
```

This relies on `Printexc.to_string` matching an Eio-internal module path. If Eio reorganizes its internals (any minor version bump), the string changes silently, `is_internal_race_cancel` returns `false`, and the refresh loop terminates on what should be a benign internal race cancel. Consider matching on the exception type directly (`Eio.Cancel.Cancelled (Eio.Fiber.Not_first)`) if exported, or restructuring to avoid needing this discrimination.

**2. `_status_cache` / `_transport_health_cache` — unprotected mutable state (tool_room.ml:27-35, server_dashboard_http_core.ml:10-59)**

Both caches use `mutable` fields accessed from concurrent Eio fibers without `Eio.Mutex`. The project's MEMORY.md explicitly requires `Eio.Mutex` for shared state in Eio servers. `_status_cache` in `tool_room.ml` is new and faces concurrent MCP tool dispatch pressure. The read-check-compute-write sequence in `cached_text_by_key` is not atomic across fiber yield points.

### Important

**3. `dashboard_transport_health_http_json` — no fallback when refresh loop not started (server_dashboard_http.ml:155)**

The function changed from on-demand compute to pure cache reader (`~state:_` ignores state). If the refresh loop hasn't started (executor pool creation failure, test context), the cache permanently returns `"initializing"`. The `_execution_cache` analogue has `cached_surface_or_first_success_json` as fallback — transport health has none.

**4. `classify_keeper_quiet_reason` — untested new branches (keeper_exec_status.ml:199-254)**

New strings `"not_running"` and `"agent_missing"` replace the previous `"disabled"` for two conditions. No test coverage found for these new paths. Downstream consumers matching on `"disabled"` will silently miss these cases.

**5. No test coverage for `Proactive_refresh` module**

The circuit-breaker logic (exponential backoff, failure thresholds, cancel discrimination) has no unit tests. The `is_internal_race_cancel` heuristic would particularly benefit from a regression test.

PR #2979, opened 2026-03-25
fix(keeper): collect board events once per heartbeat + observe pass-through

jeong-sik:
## Summary
Fixes review findings from #2961:

1. **Board events collected once** — `collect_board_events` advances cursor on each call. Previously called twice (triage + observe), second call always returned empty. Now collected before triage and passed to `observe` via `?pre_collected_board_events`.

2. **TODO for Eio.Mutex** — `keepalives`, `last_agent_counts`, `board_reactive_wakeups` Hashtbls need mutex protection (tracked in #2965).

## Files
- `lib/keeper/keeper_keepalive.ml` — hoist board collection, pass to observe
- `lib/keeper/keeper_world_observation.ml` — add `?pre_collected_board_events` arg
- `lib/keeper/keeper_world_observation.mli` — update signature

## Test plan
- [x] `dune build` passes
- [ ] CI green

### DIFF
--- a/lib/keeper/keeper_keepalive.ml
+++ b/lib/keeper/keeper_keepalive.ml
@@ -270,6 +270,19 @@ let run_heartbeat_loop ~proactive_warmup_sec (ctx : _ context)
                  Log.Keeper.error "heartbeat snapshot write failed: %s"
                    (Printexc.to_string exn));
               last_snapshot_ts := now_ts);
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing — the board events double-call fix was already included in #2961 which has been merged. The merged version uses `~pending_board_events:string list option` as a labeled arg on `observe`, achieving the same single-collection-per-heartbeat goal.

PR #3379, opened 2026-03-26
fix(dashboard): follow up asset root handling

jeong-sik:
## Summary
- share dashboard asset root resolution through `Web_dashboard.assets_root`
- keep invalid explicit asset envs from masking valid base-path candidates
- add regression coverage for invalid explicit assets root fallback

## Verification
- `dune build --root . @check`
- dashboard coverage regression updated in `test/test_web_dashboard_coverage.ml`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -362,6 +362,9 @@ jobs:
       - name: Refresh system package index
         run: sudo apt-get update -qq
 
+      - name: Install OS dependencies
+        run: sudo apt-get install -y ripgrep
+
       - name: Install dependencies
         run: |
           for attempt in 1 2 3; do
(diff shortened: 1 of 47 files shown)


jeong-sik:
CI follow-up update:

- Pushed CI fix commit: `af9ee2552`
- Fixed `bin/masc_tui.ml` to avoid linking `masc_mcp` from the standalone TUI binary and keep room-name fallback based on `MASC_CLUSTER_NAME`
- Fixed `scripts/check-eio-conventions.sh` to fall back to `grep` when `rg` is unavailable on the runner
- External review: `GLM-5` via `sb glm-text` on this patch, result: `No findings.`
- Local verification:
  - `dune build bin/masc_tui.exe test/test_web_dashboard_coverage.exe`
  - `./scripts/check-eio-conventions.sh`
  - `PATH=/usr/bin:/bin ./scripts/check-eio-conventions.sh`
  - `test_web_dashboard_coverage.exe` 21/21 passed

PR #3382, opened 2026-03-26
test(config): pin keeper fixture config dir

jeong-sik:
## Summary
- pin `MASC_CONFIG_DIR` inside the keeper config provenance test fixture
- restore the intended test fixture source after config dir resolution priority changed in #3377
- keep resolver precedence unchanged while making the test explicit about its config root

## Testing
- `dune build --root . test/test_operator_control.exe`
- `./_build/default/test/test_operator_control.exe test operator 26 --show-errors --color=never`

## Notes
- this is a follow-up to #3377
- `env OCAMLPARAM='_,warn-error=+a' opam exec -- dune build --root . @install --force` is currently failing on `main` for an unrelated baseline issue tracked in #3381

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -362,6 +362,9 @@ jobs:
       - name: Refresh system package index
         run: sudo apt-get update -qq
 
+      - name: Install OS dependencies
+        run: sudo apt-get install -y ripgrep
+
       - name: Install dependencies
         run: |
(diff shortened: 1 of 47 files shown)


jeong-sik:
Synced to current main and reran CI on c8937a506387f0c95774e6289e205c351a5fc0ca. This branch no longer has any file diff relative to main, so I am closing it as redundant/superseded by main.

PR #3385, opened 2026-03-26
fix: stabilize Supabase-backed dashboard refresh

jeong-sik:
## Summary
- prefer the same-host Supabase transaction pooler (`:6543`) when a legacy `MASC_POSTGRES_URL=:5432` shadows transaction URLs in `SUPABASE_DB_URL` or `SB_PG_URL`
- keep the `oneshot` prepared-statement policy aligned with the effective runtime URL selection
- remove unnecessary write-path work from read-heavy paths by making `masc_worktree_list` non-join-required and keeping `transport_health` metrics-only

## Problem
The dashboard and MCP tool path were cascading under load:
- Supabase session pooler exhaustion raised `MaxClientsInSessionMode`
- dashboard refresh loops fell back or timed out
- `masc_worktree_list` inherited avoidable room write work and hit the 30s tool timeout
- `transport_health` was still doing Room/command-plane reads despite being intended as a cheap health surface

## Changes
- runtime PG URL selection now prefers a same-host transaction pooler companion on `:6543`
- `pg_infix` mirrors the same selection logic before deciding `oneshot`
- `masc_worktree_list` remains read-only but no longer requires join
- `transport_health` no longer reads Room messages or command-plane topology
- docs/tests updated to match the new Supabase preference and tool contracts

## Validation
- `env DUNE_BUILD_DIR=_build_fix_supabase dune build --root . _build_fix_supabase/default/test/test_backend_coverage.exe _build_fix_supabase/default/test/test_ci_hardening_source.exe _build_fix_supabase/default/bin/main_eio.exe`
- `_build_fix_supabase/default/test/test_ci_hardening_source.exe`
- `_build_fix_supabase/default/test/test_backend_coverage.exe`
- live validation on a restarted server showed:
  - transaction-pooler companion selection log instead of direct `:5432` use
  - `transport_health warm cache done (0.0s)`
  - fresh `/api/v1/dashboard/transport-health` response
  - `masc_worktree_list` no longer hitting the prior 30s timeout path


jeong-sik:
Final cross-model re-review after the CI-fix follow-up commit:

- Model: `GLM-5` via `sb glm-text --no-thinking`
- Timestamp: 2026-03-27 12:48:48 KST
- Scope: current PR diff against `origin/main`, with intended behavior changes explicitly supplied (`transport_health` metrics-only degradation metadata, Supabase `:6543` companion preference, `masc_worktree_list` non-join-required)
- Result: `No findings.`

Follow-up note:
- the intermediate CI failure came from a partial application typo in the updated log call and was fixed in `4194c9e32` before this re-review

PR #3431, opened 2026-03-27
Fix room-truth cold execution timeout fallback

jeong-sik:
## Summary
- keep `room-truth` off the cold execution compute path after the initial proactive warm-up window
- keep using the cached execution surface so expired warm-up falls back to partial truth instead of waiting on a 20s execution timeout
- add a regression test for the expired warm-up fallback path

## Testing
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-room-truth-cold-path --build-dir _build_codex_room_truth ./test/test_dashboard_room_truth.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-room-truth-cold-path --build-dir _build_codex_room_truth ./test/test_ci_hardening_source.exe`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-room-truth-cold-path --build-dir _build_codex_room_truth @default`

### DIFF
--- a/lib/server/server_dashboard_http.ml
+++ b/lib/server/server_dashboard_http.ml
@@ -467,7 +467,7 @@ let dashboard_room_truth_focus_json ~initialized ~runtime_count ~operator_digest
                   ("suggested_params", `Assoc []);
                 ]))
 
-let dashboard_room_truth_http_json ~state ~sw ~clock request =
+let dashboard_room_truth_http_json ~state ~sw:_ ~clock _request =
   (* Fast-path: if the proactive execution refresh hasn't produced a result
      yet, return "initializing" immediately instead of blocking for 15-20s
      on cold-start on-demand compute.  The frontend retries every 3s via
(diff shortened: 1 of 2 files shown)


jeong-sik:
Cross-model review evidence

- Reviewer path: direct `sb glm-text --no-thinking`.
- Reviewer model selection: `scripts/hub/glm.sh` uses `GLM_TEXT_MODELS` default order starting with `glm-5`.
- Timestamp: `2026-03-27T04:13:43Z`
- Scope: `origin/main...HEAD`
- Verdict (raw):
```text
- `lib/server/server_dashboard_http.ml`: Removing the `initializing` status fallback means requests with a cold/expired execution cache now return partial data with `active_sessions: 0` instead of an explicit `initializing` status. This changes the API contract - clients expecting `"status": "initializing"` during warmup will now receive `"status": null` (or the derived status), which could break client-side handling of the initializing state.
- `lib/server/server_dashboard_http.ml`: The test checks `json |> member "status" = `Null` but the actual behavior after removing the initializing branch is unclear - the code falls through to derive status from `execution_json`, which may produce a different value than `Null`. The test assertion may not match the actual runtime behavior.
- `lib/server/server_dashboard_http.ml`: By unconditionally using `cached_surface_json _execution_cache` without any timeout or fallback logic, a completely empty/uninitialized cache could return malformed or empty JSON where `projection_diagnostics` or other expected fields are missing, potentially causing downstream parsing errors.
- `test/test_dashboard_room_truth.ml`: The test function `expire_execution_warmup` directly mutates `surface.last_attempt_unix` and `surface.last_attempt_at` fields, but there's no verification these fields exist or are named correctly in the cache implementation - this could be a stale assumption about the cache structure.
```
- Disposition: no additional patch.
- Rationale: the top-level `status: "initializing"` response is still preserved for the first proactive warm-up window; only the expired/failed warm-up path now falls back to partial truth so `room-truth` stops blocking on the 20s cold execution timeout.
- Rationale: the `status = Null` concern is covered by the new runtime regression test plus the passing `test_dashboard_room_truth.exe` run.
- Rationale: `_execution_cache` is initialized with a default JSON payload, and `cached_surface_json` always attaches `projection_diagnostics`, so the empty-cache malformed-JSON concern does not reproduce.
- Rationale: the test helper mutates the exposed cache record intentionally; if those fields change, the test fails at compile time rather than silently drifting.

PR #3634, opened 2026-03-29
fix(cdal): enforce adversarial red-line paths (#3524)

jeong-sik:
## Summary
- enforce adversarial red-line checks against normalized full paths instead of basenames
- block design/spec/history/governance/session-log style inputs before deep review prompt construction
- add regression tests for real path shapes and deep-review prompt rejection

## Root cause
`Adversarial_eval.classify_path` only inspected `Filename.basename`, so `docs/design/...` and similar disallowed paths could bypass the red line. The module also was not wired into the deep review prompt builder, so the restriction was not enforced on a live path.

## Impact
`masc_deep_review` now rejects banned prompt inputs early while continuing to accept normal code targets.

## Validation
- `_build/default/test/test_adversarial_eval.exe`
- `dune build --root . test/test_room.exe`

### DIFF
--- a/lib/cdal/adversarial_eval.ml
+++ b/lib/cdal/adversarial_eval.ml
@@ -49,6 +49,9 @@ let banned_doc_patterns =
 let doc_extensions =
   [ ".md"; ".markdown"; ".mdx"; ".txt"; ".rst"; ".adoc"; ".asciidoc" ]
 
+let data_artifact_extensions =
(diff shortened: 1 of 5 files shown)


jeong-sik:
Cross-model review evidence

- Reviewer: direct `sb glm-text` (GLM-5 path)
- Timestamp: 2026-03-29 16:22:36 KST
- Scope: correctness-only review of the PR diff
- Result: `No findings.`
- Local validation:
  - `_build/default/test/test_adversarial_eval.exe`
  - `dune build --root . test/test_room.exe`

PR #3638, opened 2026-03-29
[codex] docs: align truth and add doc guard

jeong-sik:
## Summary
- align front-door docs with current runtime and CLI truth
- clean up stale secondary/historical doc claims that no longer match the codebase
- add a minimal front-door doc truth guard to CI

## Why
The repo had multiple copy-paste footguns in user-facing docs and drift in a few key spec pages. That produced broken startup examples, wrong endpoint/method guidance, and stale version/runtime claims.

## Impact
- users get current CLI/MCP bootstrap instructions
- key spec pages stop contradicting the server/router/runtime defaults
- CI now catches a small set of high-value doc regressions without trying to machine-enforce the entire spec corpus

## Validation
- `git diff --check`

## Notes
- no full test suite run; changes are docs, comments, and a shell-based CI guard only


jeong-sik:
Cross-model review evidence

- Model: `GLM-5` via `sb glm-text --no-thinking`
- Timestamp (UTC): `2026-03-29T08:00:59Z`
- Scope: `scripts/check-doc-truth.sh`, `.github/workflows/ci.yml`
- Result: `No high-confidence findings.`
- Residual risk: the guard assumes specific doc structures and version markers; if those files are heavily reorganized, the script may need updates to avoid false negatives or brittle matching.

PR #4061, opened 2026-03-30
[codex] refactor(keeper): replace tool_allowlist with tool_access ADT

jeong-sik:
## Summary

This replaces keeper `tool_allowlist: string list` with an explicit `tool_access` ADT so the empty-list case is no longer semantically ambiguous.

## What Changed

- introduced `tool_access = Unrestricted | Restricted of string list` in keeper metadata
- updated keeper tool gating to pattern match on `tool_access`
- defaulted new keepers to `Unrestricted`
- kept backward-compatible JSON parsing for legacy `tool_allowlist`
- exposed `tool_access` through dashboard keeper JSON and dashboard tool mutation responses
- added dashboard support for `set_unrestricted` while keeping existing allowlist actions as `Restricted [...]`
- updated keeper test helpers and coverage to construct metadata through `tool_access`

## Why

`tool_allowlist = []` had already meant both “deny all” and “unrestricted” in different contexts. This PR makes the policy explicit in type space and removes that ambiguity from runtime logic.

## Impact

- keeper policy semantics are explicit and easier to reason about
- existing persisted keeper JSON still loads via the legacy `tool_allowlist` fallback
- dashboard/API consumers can move toward `tool_access` without losing the current allowlist view

## Validation

- `./_build/default/test/test_keeper_tool_exposure.exe`
- `./_build/default/test/test_keeper_safety_gates.exe`
- `./_build/default/test/test_keeper_tools_oas.exe`
- `./_build/default/test/test_keeper_masc_bridge.exe`

### DIFF
--- a/lib/dashboard/dashboard_http_keeper.ml
+++ b/lib/dashboard/dashboard_http_keeper.ml
@@ -753,8 +753,10 @@ let keeper_config_json (config : Room.config) (name : string)
           allowed
           |> List.filter (fun n -> String.starts_with ~prefix:"masc_" n)
         in
+        let tool_allowlist = Keeper_types.tool_access_allowlist m.tool_access in
         `Assoc [
-          ("tool_allowlist", `List (List.map (fun s -> `String s) m.tool_allowlist));
+          ("tool_access", Keeper_types.tool_access_to_json m.tool_access);
+          ("tool_allowlist", `List (List.map (fun s -> `String s) tool_allowlist));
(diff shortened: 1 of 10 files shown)


jeong-sik:
Closing as stale duplicate. `git log --cherry-pick origin/main...fix/4035-keeper-tool-access` shows no remaining unique commits on this branch, so this tool-access refactor line has already been absorbed upstream.

PR #4612, opened 2026-04-02
[codex] disable keeper autoboot on loopback startup

jeong-sik:
## Summary
- disable keeper autoboot by default on the local `scripts/start-loopback.sh` path
- preserve `MASC_KEEPER_BOOTSTRAP_ENABLED` through `start-masc-mcp.sh` env-file loading so explicit caller intent wins
- add coverage for the loopback wrapper default and override behavior, and document the local-dev behavior in `README.md`

## Why
- the loopback `8935` path is the default single-user local runtime and should not restart a whole keeper set unless the caller explicitly opts in
- without preserving `MASC_KEEPER_BOOTSTRAP_ENABLED`, wrapper-level overrides can be silently reset by repo or shell env files

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex-disable-loopback-keeper-autoboot`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex-disable-loopback-keeper-autoboot -- test/test_start_masc_mcp_script.exe`
  - new loopback autoboot test passes
  - two pre-existing `MASC_CONFIG_DIR` assertions still fail, and the same failures reproduce on `main`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ This path is the easiest default when you want local MCP development plus transp
 
 - shortcut script: `scripts/start-loopback.sh`
 - keep the server loopback-only with `--host 127.0.0.1`
+- keep keeper autoboot off by default on this shortcut path; override with `MASC_KEEPER_BOOTSTRAP_ENABLED=true scripts/start-loopback.sh` when you intentionally want bootstrap scan on local 8935
 - keep the fixed default ports when you are not juggling multiple instances:
   - HTTP / MCP: `127.0.0.1:8935`
   - gRPC: `127.0.0.1:8936`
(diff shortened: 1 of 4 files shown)


jeong-sik:
Cross-model review evidence
- Model: direct `sb glm-text` default reviewer path (`GLM-5.1`)
- Timestamp: 2026-04-02T03:47:45Z
- Result: No findings.

Summary
- Reviewed for material correctness issues, regressions, and missing tests.
- No material findings were reported.
- Validation note: the reviewer called out that the new loopback test covers both the default-off path and the explicit override path.

PR #4630, opened 2026-04-02
fix(keeper): address review comments on tag dispatch

jeong-sik:
## Summary

Addresses review comments from PR #4594 and PR #4607 on the keeper tag dispatch feature.

- Replace `keeper_dispatch_none` error with `tool_not_supported_in_keeper` + actionable hint
- Wrap entire dispatch match in try-catch to normalize Tool_* exceptions into error tuples (prevents keeper loop crash)
- Add `require_net()` helper, unify error message format with `require_sw`/`require_clock`
- Change `Mod_control` to return `None` instead of `Some (false, ...)` so unsupported tools are not advertised as always-failing

## Review comment mapping

| Source | Comment | Response |
|--------|---------|----------|
| #4594 | `keeper_dispatch_none` is internal detail | Changed to `tool_not_supported_in_keeper` + hint |
| #4594 | No exception normalization in tag dispatch | Added try-catch wrapping (re-raises `Eio.Cancel.Cancelled`) |
| #4594 | Tag dispatch drift risk | Acknowledged; follow-up test in Phase 2 |
| #4607 | Mod_control "shown but blocked" | Changed to `None` return |
| #4607 | Research net error inconsistent | Added `require_net()` helper |
| #4607 | Wildcard vs exhaustive match | Removed by CI (warn-error=+a); exhaustive match preserved |

## Test plan

- [x] `dune build` passes
- [ ] CI green

Ref: #4579

### DIFF
--- a/lib/keeper/keeper_exec_tools.ml
+++ b/lib/keeper/keeper_exec_tools.ml
@@ -788,8 +788,9 @@ let execute_keeper_tool_call
                         (`Assoc [ ("error", `String msg) ])
                   | None ->
                       Yojson.Safe.to_string
-                        (`Assoc [ ("error", `String "keeper_dispatch_none");
-                                  ("tool", `String name) ]))
+                        (`Assoc [ ("error", `String "tool_not_supported_in_keeper");
+                                  ("tool", `String name);
+                                  ("hint", `String "tag dispatch returned None; check tag_dispatch_fn wiring at server init") ]))
              | None ->
                  Yojson.Safe.to_string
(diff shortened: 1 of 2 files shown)


jeong-sik:
Second-round review addressed in 8ddbe45:

1. **hint message generalized** — "tool may be unsupported, blocked, or misconfigured" (not wiring-only)
2. **Mod_control restored to Some (false, ...)** — preserves None = "unknown tool for tag" contract
3. **try-catch regression test** — noted for follow-up; this PR scope is the dispatch fix itself

PR #4691, opened 2026-04-02
refactor(dashboard): deduplicate KeeperConfigLoadStatus type

jeong-sik:
## Summary

- Import `KeeperConfigLoadStatus` from `keeper-detail-source.ts` in `keeper-config-panel.ts` instead of redeclaring the same union type inline as the return type of `peekKeeperConfigLoadStatus`.
- Eliminates drift risk where one definition could be updated without the other.
- Addresses unresolved review feedback from #4672.

## Test plan

- [x] `tsc --noEmit` passes (exit 0)
- [x] `vitest run` passes (58 files, 263 tests)
- [ ] CI green

Refs #4672

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dashboard/src/components/keeper-config-panel.ts
+++ b/dashboard/src/components/keeper-config-panel.ts
@@ -7,6 +7,7 @@ import { signal } from '@preact/signals'
 import { fetchKeeperConfig, patchKeeperConfig } from '../api/dashboard'
 import type { KeeperConfigUpdatePayload } from '../api/dashboard'
 import type { KeeperConfig } from '../types'
+import type { KeeperConfigLoadStatus } from './keeper-detail-source'
 import { formatTokens } from '../lib/format-number'
 import { showToast } from './common/toast'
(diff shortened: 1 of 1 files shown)


jeong-sik:
## Cross-Model Review (GLM-5.1 via `sb glm-text`)

**Reviewer**: GLM-5.1 (direct ZAI)
**Verdict**: No issues found.

### Findings

| # | Finding | Severity | Disposition |
|---|---------|----------|-------------|
| 1 | Verify `KeeperConfigLoadStatus` is structurally identical to previous inline union | Medium | Confirmed: same 5-member union (`idle \| loading \| loaded \| error \| other`) |
| 2 | Circular dependency risk (`keeper-config-panel` ← `keeper-detail-source`) | Low | `import type` only; no runtime risk |

No review threads to resolve. Pure type dedup refactor with no behavioral change.

**Result**: Approved for Ready conversion.

Issue #4748, opened 2026-04-03
keeper: add deterministic action executor and legality matrix


jeong-sik:
## Desired outcome
Keeper actions run through a deterministic executor with a legal action matrix, explicit fallback, and runtime-owned control semantics.

## Problem statement
Structured output alone is not enough. After parsing, the runtime still needs a single deterministic executor that decides whether a proposed action is legal, how multi-step actions are applied, and when to fall back to baseline behavior.

Without that:
- action legality remains spread across ad hoc call sites
- model output and execution semantics stay partially coupled
- failure handling is hard to make uniform

## Proposed approach
- Introduce a keeper action executor that owns:
  - legality checks per action type
  - multi-step validation rules
  - deterministic fallback to baseline behavior
  - policy/precondition integration before side effects
- Treat parsed model action as a proposal, not as authority.
- Add tests for illegal actions, nested multi-step rejection, and fallback behavior.

## Non-goals
- Structured schema definition itself
- Long-term keeper strategy redesign
- Dashboard/operator UI changes

## Links
- Parent: #4738
- Depends on: #4747

Issue #4747, opened 2026-04-03
keeper: define structured deliberation schema and remove free-form JSON recovery


jeong-sik:
## Desired outcome
Keeper deliberation requests a schema-forced structured result instead of recovering JSON from free-form model text.

## Problem statement
Parent issue #4738 defines the broader boundary change, but the first independent step is to remove brace/fence recovery from the model boundary itself.

Right now keeper deliberation still:
- builds a prompt
- accepts arbitrary text back
- tries to recover a JSON object from that text

That leaves the most fragile part of the control path outside the typed contract.

## Proposed approach
- Define a keeper deliberation schema using OAS `Structured`.
- Request only schema-conforming output from the model.
- Replace free-form JSON extraction in keeper deliberation with structured parsing.
- Keep the existing typed action ADT as the parsed target.
- Add focused tests for malformed/non-conforming model output.

## Non-goals
- Reworking keeper action execution in this issue
- Adding new keeper action types
- Changing triage heuristics

## Links
- Parent: #4738
- Parent/design context: #4308

PR #4774, opened 2026-04-03
[codex] Add keeper deterministic action executor

jeong-sik:
## Summary
- add a pure deterministic keeper action executor with an explicit legality matrix
- treat parsed model output as a proposal and fall back to the baseline action when the proposal is illegal
- bump version truth to `2.217.2`

## Why
- `#4747` fixed the model boundary, but execution semantics were still implicit after parsing
- this change gives runtime ownership over legality checks, nested multi-step rejection, and fallback behavior before any side-effect runner is attached

## Impact
- `Keeper_deliberation` now exposes legality verdicts, flattened policy labels, and deterministic execution results
- illegal proposals no longer flow through as authority; they deterministically downgrade to the baseline action
- tests now pin illegal action handling, nested `multi_step` rejection, and fallback behavior

## Validation
- `dune build --root . --build-dir /tmp/masc-4748-build --display=short test/test_keeper_deliberation.exe`
- `dune exec --root . --build-dir /tmp/masc-4748-build ./test/test_keeper_deliberation.exe`
- `scripts/check-version-truth.sh`

## Stack
- Base PR: #4770
- Refs: #4748


jeong-sik:
Cross-model review evidence

- Model: GLM-5 via direct `sb glm-text`
- Timestamp: 2026-04-03 20:44:40 KST
- Scope: rebased `origin/main...HEAD` diff after retargeting `#4774` from stacked base to `main`
- Result: `No material findings.`

PR #4821, opened 2026-04-03
dashboard: map bootstrap not-initialized responses to warm-up payloads

### LINKED ISSUE #4803: dashboard: public bootstrap endpoints return raw not-initialized error before warm-up state
## Symptom

Dashboard startup can surface a raw `{"error":"not initialized"}` response instead of entering the intended warm-up / initializing state.

This breaks the bootstrap contract for public dashboard reads:
- initial page load kicks off `refreshShell()` and `requestRoomTruthNow()` immediately
- room-truth warm-up logic expects a JSON payload with `status: "initializing"`
- instead, some public routes are short-circuited before their handler and return `{"error":"not initialized"}`
- the dashboard `get()` client throws on any non-2xx response and the UI falls into generic fetch failure handling rather than warm-up retry

## Reproduction

1. Start the server and open `/dashboard` during the startup / pre-init window.
2. The frontend immediately calls `/api/v1/dashboard/shell` and `/api/v1/dashboard/room-truth`.
3. If `server_state` is still `None`, `with_public_read` returns `{"error":"not initialized"}` before the dashboard route handler runs.
4. `dashboard/src/api/core.ts` turns that into `ApiRequestError` and room-truth warm retry never triggers.

## Evidence / code paths

Server-side short-circuit:
- `lib/server/server_auth.ml:433`
- `lib/server/server_auth.ml:440`

Dashboard public routes that rely on `with_public_read`:
- `lib/server/server_routes_http_routes_dashboard.ml:210`
- `lib/server/server_routes_http_routes_dashboard.ml:275`
- `lib/server/server_routes_http_routes_dashboard.ml:280`
- `lib/server/server_routes_http_routes_dashboard.ml:296`
- `lib/server/server_routes_http_routes_dashboard.ml:301`

Frontend warm-up expectation:
- `dashboard/src/app.ts:36`
- `dashboard/src/room-truth-actions.ts:19`
- `dashboard/src/room-truth-actions.ts:25`

Frontend generic failure path:
- `dashboard/src/api/core.ts:151`
- `dashboard/src/api/core.ts:158`

## Suspected cause

Bootstrap semantics are split across two layers:
- dashboard handlers/projections already have initializing-aware paths
- auth / server-state gate rejects public reads too early when `server_state` is not ready

That means the dashboard cannot observe the initializing contract it was built to handle.

## Proposed direction

Keep the boundary explicit:
- public dashboard bootstrap endpoints should return a stable initializing envelope while server state is still warming up, instead of raw `{"error":"not initialized"}`
- or, if raw not-initialized remains the transport truth, the frontend needs an explicit bootstrap mapping for that case before `get()` throws generic failure

Preferred first cut:
- fix server-side contract for bootstrap-friendly public endpoints (`shell`, `room-truth`, `execution`, `planning`, `mission`) so the frontend can keep using the existing initializing flow

## Attempts / investigation

- Searched for exact `not initialized` JSON emitters and found `server_auth.ml` short-circuit paths.
- Traced dashboard route registration and confirmed public dashboard projection endpoints all go through `with_public_read`.
- Traced frontend startup and confirmed room-truth warm retry only activates on `status: "initializing"`, not on HTTP error.

## Related

- Related but not the same layer: #2739
- Broader global-state tech debt: #3694

jeong-sik:
## Summary
- treat raw dashboard bootstrap `{"error":"not initialized"}` responses as synthetic `initializing` payloads for bootstrap-only surfaces
- applies to both 5xx and 2xx responses on allowlisted dashboard bootstrap endpoints
- keep the remap limited to intended bootstrap endpoints so auth/permission failures still surface as errors
- add dashboard API tests that lock the warm-up remap boundary and preserve non-bootstrap error behavior

## Testing
- `pnpm --dir dashboard exec vitest run src/api/core.test.ts --no-file-parallelism --maxWorkers=1`

## Links
- Closes #4803


jeong-sik:
Review response (740e26d):

**Copilot finding — mission payload shape**: `command_focus` and `operator_targets` now include required array fields instead of empty `{}`. `command_focus: { session_cards: [] }`, `operator_targets: { sessions: [], keepers: [], pending_confirms: [], available_actions: [] }`. Matches `DashboardMissionResponse` type definition.

(Previous findings about 2xx remap and room-truth alias were addressed in d8d26ca.)

PR #4832, opened 2026-04-03
feat(dashboard): smooth animations for kanban tasks and tab transitions

jeong-sik:
## Overview
- Integrated `@formkit/auto-animate` into the Kanban columns (`TaskColumn`). Now, when tasks are claimed, moved to in-progress/done, or deleted, they smoothly animate to their new positions instead of instantly jumping.
- Added a `fade-in slide-in-from-bottom-2` animation to the main `TabContent` wrapper in the dashboard shell. This gives a much softer and more polished transition feeling when navigating between different sections (Overview, Monitoring, Workspace, etc.).

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -17,6 +17,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
+    "@formkit/auto-animate": "^0.9.0",
     "@preact/signals": "^2.0.2",
     "dompurify": "^3.3.3",
     "downshift": "^9.3.2",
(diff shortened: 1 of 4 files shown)


jeong-sik:
Review response:

**Finding 1 — useEffect dep list**: Already using `[]` (empty deps). `listRef` was removed from deps.

**Finding 2 — reduced-motion for autoAnimate**: Already guarded with `window.matchMedia('(prefers-reduced-motion: reduce)').matches` check before calling `autoAnimate()`.

**Finding 3 — tab transition reduced-motion**: Already using `motion-reduce:animate-none` Tailwind variant on the tab content container (line 334).

All 3 findings were proactively addressed in the current HEAD.

PR #4849, opened 2026-04-03
feat(dashboard): integrate shiki for highly accurate syntax highlighting

jeong-sik:
## Overview
- Added `shiki` as the code block highlighter inside the Markdown renderer. Shiki is currently the top-tier, modern standard for syntax highlighting (used by VS Code and many modern docs) because it uses real TextMate grammars to tokenize code.
- This ensures LLM-generated code blocks and JSON outputs in the dashboard are rendered with incredibly accurate, VS Code-like colors (`vitesse-dark` theme applied).
- It is lazily loaded (just like Mermaid) so the initial JavaScript bundle size of the dashboard does not suffer at all. Only fetched when a code block is detected.
- Mocked in `vitest-setup.ts` to maintain fast, reliable parallel tests.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -26,6 +26,7 @@
     "marked": "^17.0.5",
     "mermaid": "^11.12.0",
     "preact": "^10.25.4",
+    "shiki": "^4.0.2",
     "vis-data": "^8.0.3",
     "vis-network": "^10.0.2",
     "vis-timeline": "^8.5.0"
--- a/dashboard/pnpm-lock.yaml
+++ b/dashboard/pnpm-lock.yaml
@@ -35,6 +35,9 @@ importers:
       preact:
         specifier: ^10.25.4
         version: 10.28.3
+      shiki:
+        specifier: ^4.0.2
+        version: 4.0.2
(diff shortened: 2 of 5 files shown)


jeong-sik:
Review response:

**Finding 1 — getShiki() permanent failure**: Added `.catch()` that clears the cached promise before re-throwing, enabling retry on next call.

**Finding 2 — Unsanitized innerHTML**: Created `sanitizeShikiHtml()` with dedicated `SHIKI_PURIFY_CONFIG` (allows `pre/code/span` + `class/style`). Applied before `div.innerHTML` assignment. Matches existing DOMPurify pattern.

**Finding 3 — dataset.highlighted unconditional**: Moved inside `try` block, after successful `replaceWith()`. Failed blocks remain un-marked for retry.

**Finding 4 — Mock XSS**: Added `escapeHtml()` helper that escapes `&<>"` before interpolation, matching real Shiki behavior.

**Finding 5 — Missing tests**: Added 3 tests: non-mermaid highlighting, mermaid exclusion, HTML entity escaping. All 21 tests pass.

PR #4864, opened 2026-04-03
perf(command-plane): bound slot-samples fallback reads

jeong-sik:
## Summary
- bound `swarm_proof` fallback reads of `slot-samples.jsonl` to a recent tail window instead of loading the whole file
- keep `slot-telemetry.json` as the preferred path and only change the fallback path when telemetry is missing
- add a regression test proving the fallback uses recent tail samples rather than whole-file history

## Why
`main_eio.exe` can refresh command-plane summaries in the background. When a swarm-live run has no `slot-telemetry.json`, the old fallback path read the entire `slot-samples.jsonl` file into memory. For oversized long-running artifacts, that is a plausible memory spike and stall path.

## Verification
- `dune build --root . --build-dir /tmp/masc-4862-build --display=short test/test_command_plane_v2.exe`
- `dune exec --root . --build-dir /tmp/masc-4862-build ./test/test_command_plane_v2.exe`

## Notes
- This changes fallback semantics to use the recent tail window, not whole-artifact lifetime maxima.
- Follow-up issue: #4862

### DIFF
--- a/dashboard/src/components/common/markdown.test.ts
+++ b/dashboard/src/components/common/markdown.test.ts
@@ -1,6 +1,6 @@
 import { html } from 'htm/preact'
 import { render } from 'preact'
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { Markdown } from './markdown'
 
 describe('Markdown', () => {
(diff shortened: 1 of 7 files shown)


jeong-sik:
Review scan: 2 findings — invalid ISO timestamps in test (T00:60:00Z) + unnecessary List.rev allocation. Both valid.

PR #5060, opened 2026-04-04
refactor(registry): consolidate 3 agent maps into Provider_adapter SSOT

jeong-sik:
## Summary
- Extend `Provider_adapter.adapter` with `spawn_key` and `default_voice` fields
- Remove `spawn_alias_map` from spawn.ml (12-entry duplicate)
- Remove `default_agent_voices` from voice_bridge_core.ml (4-entry duplicate)
- Adding a new agent now requires one entry in `direct_adapters`, zero changes elsewhere

## Motivation
4 separate agent lists (provider_adapter, spawn, voice_bridge, mention) that drift independently. ADR D4: Adapter Registry Pattern — single registration point for new agents.

## Changes
| File | Before | After |
|------|--------|-------|
| `provider_adapter.ml` | `adapter` has 4 fields | + `spawn_key`, `default_voice` |
| `spawn.ml` | `spawn_alias_map` (12 aliases) | `Provider_adapter.resolve_spawn_key` |
| `voice_bridge_core.ml` | hardcoded 4-voice list | `Provider_adapter.all_agent_voices()` |

## Test plan
- [x] `dune build` passes
- [ ] `dune runtest` passes (running)
- [ ] CI checks pass
- [ ] Cross-model review

### DIFF
--- a/lib/provider_adapter.ml
+++ b/lib/provider_adapter.ml
@@ -20,6 +20,8 @@ type adapter = {
   runtime_kind : runtime_kind;
   auth_mode : auth_mode;
   aliases : string list;
+  spawn_key : string option;       (** Key into Spawn.default_configs. None = not spawnable via CLI. *)
+  default_voice : string option;   (** Default TTS voice name. None = no voice assignment. *)
 }
 
 type voice_adapter = {
@@ -79,6 +81,10 @@ let cn_gemini = "gemini-api"
 let cn_glm = "glm"
 let cn_openrouter = "openrouter"
 
+(** Single source of truth for all agent adapters.
+    spawn_key maps to Spawn.default_configs keys.
+    default_voice maps to TTS voice names.
+    Adding a new agent = adding one entry here. *)
 let direct_adapters =
   [
     {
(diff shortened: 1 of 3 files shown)


jeong-sik (commented):

### [HIGH] Voice key regression: `canonical_name` vs short name mismatch

`all_agent_voices()` emits `(a.canonical_name, voice)` pairs. The canonical names are:
- `cn_claude = "claude-api"`
- `cn_gemini = "gemini-api"`
- `cn_codex = "codex-api"`

But the **old** `default_agent_voices` used short names:
- `("claude", "Sarah")`, `("gemini", "Roger")`, `("codex", "George")`, `("llama", "Laura")`

Consumers like `get_voice_for_agent agent_id` do `List.assoc_opt agent_id voices` — if callers pass `"claude"` (common: spawn configs use `agent_name = "claude"`), it will miss the `"claude-api"` key and silently fall back to `"Sarah"`. This happens to be correct for claude, but for `"codex"` it would get `"Sarah"` instead of `"George"`, and for `"gemini"` it would get `"Sarah"` instead of `"Roger"`.

**Fix**: Either use short names as keys in `all_agent_voices()` (e.g. use `spawn_key` when `Some`, else `canonical_name`), or document + verify all call sites pass canonical names.

### [MEDIUM] Redundant canonical name in aliases list

```ocaml
canonical_name = cn_claude;  (* "claude-api" *)
aliases = [ cn_claude; "claude"; "anthropic"; "claude-code"; "claude-api" ];
```

`cn_claude` and `"claude-api"` are the same string — the canonical name appears twice in aliases plus once as canonical_name. `resolve_direct_adapter` checks `canonical_name` first, so the alias entries are dead. Not a bug, but noisy and suggests mechanical migration.

### [LOW] `spawnable_canonical_names()` has no call site

Defined but not called anywhere in this diff or the existing codebase. Dead code in a consolidation PR is suspicious. If intended for future use, add a comment; otherwise remove.

### [OK] Spawn alias migration is complete

All 11 entries from the deleted `spawn_alias_map` are present in the new `aliases` fields. The two-stage lookup in `Spawn.get_config` (direct key -> `Provider_adapter.resolve_spawn_key`) preserves the original semantics. No missing alias found.

### [OK] `default_agent_voices` value-to-thunk change is safe

Single call site updated correctly. No external API break.

### Summary

| Severity | Finding |
|----------|---------|
| HIGH | Voice key regression: `canonical_name` ("claude-api") vs short name ("claude") |
| MEDIUM | Redundant canonical name in aliases |
| LOW | Dead code: `spawnable_canonical_names()` |

Recommendation: Fix the HIGH finding before merge. The voice fallback path will silently assign wrong voices to codex/gemini agents when the voice config file is absent.

---
Reviewed by: Claude Opus 4.6 + GLM-5.1 (cross-model)

PR #5064, opened 2026-04-04
fix(boundary): remove vendor/model hardcoding from MASC layer (7 HIGH)

jeong-sik:
## Summary
- Remove 5 hardcoded "claude" defaults from tool_relay.ml → use `ctx.agent_name`
- Remove 2 hardcoded "llama:qwen3.5-9b-local" from tool_keeper.ml and tool_repair_loop_types.ml → use `Env_config.Local_runtime.default_model`

## Motivation
MASC must not know vendor names or specific model IDs (ADR boundary).
Audit found 7 HIGH violations where MASC code contained hardcoded vendor/model references.

## Changes
| File | Before | After |
|------|--------|-------|
| `tool_relay.ml` | `"claude"` x5 | `ctx.agent_name` (dynamic) |
| `tool_keeper.ml` | `"llama:qwen3.5-9b-local"` | `Env_config.Local_runtime.default_model` |
| `tool_repair_loop_types.ml` | `"llama:qwen3.5-9b-local"` | `Env_config.Local_runtime.default_model` |

## Test plan
- [x] `dune build` passes
- [ ] CI checks pass
- [ ] Cross-model review

### DIFF
--- a/lib/tool_keeper.ml
+++ b/lib/tool_keeper.ml
@@ -325,7 +325,7 @@ let default_keeper_model_label (meta : keeper_meta) =
   | "" -> (
       match Oas_model_resolve.models_of_cascade_name meta.cascade_name with
       | first :: _ when String.trim first <> "" -> first
-      | _ -> "llama:qwen3.5-9b-local")
+      | _ -> Env_config.Local_runtime.default_model)
   | model -> model
 
 let annotate_keeper_repair_json ~(keeper_name : string) body =
--- a/lib/tool_relay.ml
+++ b/lib/tool_relay.ml
@@ -11,10 +11,10 @@ type context = {
 
 type result = bool * string
 
-let handle_relay_status _ctx args =
+let handle_relay_status ctx args =
   let*! messages = get_int_required args "messages" in
   let*! tool_calls = get_int_required args "tool_calls" in
-  let model = get_string args "model" "claude" in
+  let model = get_string args "model" ctx.agent_name in
   let metrics = Relay.estimate_context ~messages ~tool_calls ~model in
   let actual_tokens_opt = match Yojson.Safe.Util.member "actual_tokens" args with
     | `Int n -> Some n
@@ -36,7 +36,7 @@ let handle_relay_status _ctx args =
   ] in
   (true, Yojson.Safe.pretty_to_string json)
 
-let handle_relay_checkpoint _ctx args =
+let handle_relay_checkpoint ctx args =
   let*! summary = get_string_required args "summary" in
   let*! messages = get_int_required args "messages" in
   let*! tool_calls = get_int_required args "tool_calls" in
@@ -51,7 +51,7 @@ let handle_relay_checkpoint _ctx args =
    if goal_ids <> [] then
      Log.Misc.info "[RELAY] active_goal_ids provided but not persisted (goal store not integrated): %s"
        (String.concat "," goal_ids));
-  let metrics = Relay.estimate_context ~messages ~tool_calls ~model:"claude" in
+  let metrics = Relay.estimate_context ~messages ~tool_calls ~model:ctx.agent_name in
   let _cp = Relay.save_checkpoint ~summary ~task:current_task ~todos ~pdca:pdca_state ~files:relevant_files ~metrics in
   let json = `Assoc [
     ("status", `String "checkpoint_saved");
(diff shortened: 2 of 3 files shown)


jeong-sik (commented):

### [HIGH] Semantic mismatch: `ctx.agent_name` is not a model identifier

`ctx.agent_name` is a logical agent name (e.g. `"claude"`, `"gemini"`, `"codex"`) used as the `~model` argument to `Relay.estimate_context`. Inside relay, `resolve_max_context model` does a 3-layer lookup:
1. `Capabilities.for_model_id` (expects concrete model IDs like `"claude-opus-4-6"`)
2. `Provider_registry.find` (expects provider names like `"claude"`)
3. Hyphenated prefix extraction
4. Fallback: 100,000 tokens

Layer 2 happens to work when agent_name matches a provider name (claude/gemini/codex), so the current change is **functionally correct for the common case**. However, if an agent is named something non-standard (e.g. `"reviewer"`, `"summarizer"`, or an empty string), it silently falls through to the 100K fallback -- which may over- or under-estimate context.

**Recommendation**: This is acceptable for now since MASC agent names are constrained to provider names in practice. Add a code comment documenting this assumption, and consider threading `ctx.model` or `ctx.provider` as a future improvement.

### [HIGH] Empty `agent_name` risk

`ctx.agent_name` is `string` (not `string option`). If it is `""`, `Relay.estimate_context` receives `~model:""` which passes through all 3 resolve layers and hits the 100K fallback. The old hardcoded `"claude"` was always a valid key.

Checked: `agent_name` is set from `Room.config` or spawn parameters. In normal MASC operation it should not be empty, but there is no validation at the `tool_relay` boundary.

**Fix**: Add a guard: `let model = match ctx.agent_name with "" -> "claude" | s -> s` or log a warning. Alternatively, validate `agent_name` non-empty at context construction time.

### [MEDIUM] Missed instance: `tool_repair_loop.ml` line ~4

PR fixes `tool_repair_loop_types.ml` and `tool_keeper.ml` but **misses** `tool_repair_loop.ml`:

```ocaml
let default_model_label () =
  match Sys.getenv_opt "MODEL_LABEL" with
  | Some value when String.trim value <> "" -> String.trim value
  | _ -> "llama:qwen3.5-9b-local"    (* <-- still hardcoded *)
```

This is the same violation pattern. Should be `Env_config.Local_runtime.default_model` for consistency.

### [MEDIUM] `Env_config.Local_runtime.default_model` defaults to `"explicit-model-required"`

Verified in `lib/config/env_config_runtime.ml`:
```ocaml
let default_model =
  get_string ~default:"explicit-model-required" "LLAMA_DEFAULT_MODEL"
```

When `LLAMA_DEFAULT_MODEL` is unset, the fallback is the sentinel string `"explicit-model-required"`. This is intentionally non-functional (forces explicit configuration), which is the correct boundary behavior. However, if `tool_keeper.ml` or `tool_repair_loop_types.ml` propagates this sentinel to an actual inference call without validation, it will fail at the provider level. Verify downstream consumers handle this sentinel.

### [MEDIUM] Remaining `"claude"` in schema default (line 233)

In `tool_relay.ml`, the `masc_relay_now` schema still has:
```ocaml
("target_agent", `Assoc [
  ("type", `String "string");
  ("description", `String "Agent to relay to (default: claude)");
  ("default", `String "claude");
]);
```

This is a JSON Schema default shown to the LLM caller, not runtime logic, so severity is lower. But it contradicts the PR goal: the schema still claims `"claude"` is the default while the runtime now uses `ctx.agent_name`. Either remove the schema default or make it dynamic.

### [OK] `_ctx` -> `ctx` parameter rename is correct

The previously unused `_ctx` parameters are now properly used. Type-safe since `context` has `agent_name: string`.

### Summary

| Severity | Finding |
|----------|---------|
| HIGH | `agent_name` is not a model ID; works by convention but fragile |
| HIGH | Empty `agent_name` falls through to 100K fallback silently |
| MEDIUM | Missed hardcoding in `tool_repair_loop.ml` |
| MEDIUM | Sentinel default `"explicit-model-required"` needs downstream validation |
| MEDIUM | Schema default still says `"claude"` at line 233 |

Recommendation: Fix the MEDIUM missed instance in `tool_repair_loop.ml`, add empty-string guard for `agent_name`, and update the schema default. The HIGH semantic concern is acceptable short-term given MASC naming conventions.

---
Reviewed by: Claude Opus 4.6 + GLM-5.1 (cross-model)

PR #5070, opened 2026-04-04
feat(task): extend task schema with goal decomposition metadata

jeong-sik:
## Summary
- Add `parent_task_id`, `goal_id`, `complexity_estimate` to task type
- `complexity` type: `estimated_turns` + Karpathy `reversibility` metric
- All fields optional (None default) — backward compatible JSON

## Motivation
Phase 3 foundation: hierarchical goal → sub-goal → task decomposition.
Without parent_task_id, tasks are flat. Without complexity_estimate,
automatic task splitting is impossible.

## Sync Points (all updated)
1. `room_eio.ml` — type + task_to_json + task_of_json + create_task
2. `room_eio.mli` — interface
3. `tui_decode.ml` + `.mli` — TUI rendering
4. `test_types_utils_coverage.ml` — roundtrip + goal metadata + backward compat

## Test plan
- [x] `dune build` passes
- [x] test_room_eio_task_roundtrip (existing, updated)
- [x] test_room_eio_task_with_goal_metadata (new)
- [x] test_room_eio_task_backward_compat (new — old JSON without new fields)
- [ ] CI checks pass

### DIFF
--- a/lib/room/room_eio.ml
+++ b/lib/room/room_eio.ml
@@ -659,6 +659,11 @@ type task_status =
   | Completed of string   (* agent_id *)
   | Failed of string * string  (* agent_id, reason *)
 
+type complexity = {
+  estimated_turns: int;
+  reversibility: float;
+}
+
 type task = {
   id: string;
   description: string;
@@ -666,7 +671,21 @@ type task = {
   created_at: float;
   updated_at: float;
   priority: int;
+  parent_task_id: string option;
+  goal_id: string option;
+  complexity_estimate: complexity option;
 }
+
+let complexity_to_json c =
+  `Assoc [
+    ("estimated_turns", `Int c.estimated_turns);
+    ("reversibility", `Float c.reversibility);
+  ]
+
+let complexity_of_json json =
+  let open Yojson.Safe.Util in
+  { estimated_turns = json |> member "estimated_turns" |> to_int;
+    reversibility = json |> member "reversibility" |> to_number }
 
 let task_status_to_json = function
   | Pending -> `Assoc [("type", `String "pending")]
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):

### CRITICAL: Silent coercion of malformed optional fields

`str_opt` in `task_of_json` uses a wildcard match:
```ocaml
let str_opt key = match json |> member key with
  | `String s -> Some s | _ -> None
```
If a caller sends `"parent_task_id": 123` or `"parent_task_id": null`, this silently coerces to `None` instead of returning `Error`. This violates the "Parse, Don't Validate" principle. The same issue applies to `complexity_opt` which silently drops `"complexity_estimate": "invalid"`.

**Fix**: If the key is present but has the wrong JSON type, return `Error` with a descriptive message. Only return `None` when the key is truly absent (`Yojson.Safe.Util.member` returns `` `Null `` for missing keys, which is distinct from a wrong type like `` `Int ``).

### HIGH: Type drift between `Room_eio.task` and `Types.task`

There are now two `task` types in the codebase:
- `Types.task` (`lib/types/types_core.ml`): has `title`, `files`, `worktree`, `required_role`, `stage` -- used by `Room.read_backlog`
- `Room_eio.task` (`lib/room/room_eio.ml`): has `description` (no `title`), now gains `parent_task_id`, `goal_id`, `complexity_estimate`

PR #5073 depends on `Types.task.title` for dedup, but the goal decomposition metadata from this PR is only on `Room_eio.task`. These types are diverging. The new fields should either be added to `Types.task` too or a consolidation plan should be documented.

### HIGH: `complexity` record exposed as public API without accessors

The `complexity` record is fully exposed in `room_eio.mli` and tests access fields directly (`c.estimated_turns = 5`). This locks the internal layout as public API. Consider providing a constructor `val make_complexity : int -> float -> complexity` and accessors, or at minimum document that this is an intentional public record.

### MEDIUM: `tui_decode.ml` omits `complexity_estimate`

`tui_decode.ml` adds `parent_task_id` and `goal_id` but not `complexity_estimate`. If intentional (TUI doesn't need it), that's fine but should be noted. If accidental, it means TUI rendering will drop complexity data silently.

### MEDIUM: `reversibility` field lacks domain constraint

`reversibility: float` accepts any float value, but the semantic range is [0.0, 1.0]. There is no validation at construction time. A negative or >1.0 value would be silently accepted.

### LOW: No negative test for malformed JSON fields

The backward-compat test is good. Missing: a test with `"parent_task_id": 42` (wrong type) to verify behavior. Currently this silently becomes `None`; the test should document whether that's intended or a bug.

### Summary

| Finding | Severity |
|---------|----------|
| Silent coercion of malformed optional fields | CRITICAL |
| Type drift `Room_eio.task` vs `Types.task` | HIGH |
| `complexity` record exposed without accessors | HIGH |
| `tui_decode` omits `complexity_estimate` | MEDIUM |
| No domain constraint on `reversibility` | MEDIUM |
| No negative test for malformed JSON | LOW |

Reviewed by: Claude Opus 4.6 (cross-check) + GLM-5.1 (primary)

PR #5073, opened 2026-04-04
feat(goal): idempotent task dispatch + Karpathy complexity analyzer

jeong-sik:
## Summary
- Idempotent `add_task_safe`: title-prefix dedup prevents duplicate tasks on re-dispatch
- Karpathy-inspired `estimate_reversibility`: keyword-based metric (deploy=0.1 → read=1.0)
- `estimate_complexity`: word count proxy + reversibility score
- `needs_decomposition`: threshold gate (>20 estimated turns)

## Motivation
- `goal_orchestrator.ml:107` creates tasks without idempotency (#5044)
- `swarm_goal_loop.ml:174` cancel/recreate pattern causes churn
- Complexity analyzer is Phase 3 prerequisite for automatic task splitting

## Design
**Idempotency**: Before `Room.add_task`, check backlog for matching title prefix.
Skip if already exists. Simple, deterministic, no external state needed.

**Complexity**: Pure keyword heuristic (no LLM). Karpathy reversibility
metric calibrated for MASC task types. `needs_decomposition` threshold
configurable via constant (currently 20 turns).

## Test plan
- [x] `dune build` passes
- [ ] CI checks pass

Refs: #5044

### DIFF
--- a/lib/goal/goal_orchestrator.ml
+++ b/lib/goal/goal_orchestrator.ml
@@ -104,11 +104,77 @@ let build_plan ~goals opts =
     nodes;
   }
 
+(* ================================================================ *)
+(* Complexity Analyzer (deterministic, no LLM)                      *)
+(* ================================================================ *)
+
+(** Karpathy-inspired reversibility heuristic.
+    Estimates how reversible a task is based on keyword patterns.
+    0.0 = irreversible (deploy, delete), 1.0 = fully reversible (read, search). *)
+let estimate_reversibility description =
+  let desc = String.lowercase_ascii description in
+  let has w =
+    let wlen = String.length w in
+    let dlen = String.length desc in
+    if wlen > dlen then false
+    else
+      let found = ref false in
+      for i = 0 to dlen - wlen do
+        if not !found && String.sub desc i wlen = w then found := true
+      done;
+      !found
+  in
+  if has "deploy" || has "release" || has "publish" || has "production" then 0.1
+  else if has "delete" || has "drop" || has "remove" || has "destroy" then 0.2
+  else if has "migrate" || has "schema" || has "database" then 0.3
+  else if has "config" || has "environment" || has "secret" then 0.6
+  else if has "refactor" || has "implement" || has "code" || has "fix" then 0.8
+  else if has "test" || has "lint" || has "format" then 0.9
+  else if has "read" || has "search" || has "query" || has "list" || has "view" then 1.0
+  else 0.7  (* unknown defaults to moderately reversible *)
+
+type task_complexity = {
+  estimated_turns : int;
+  reversibility : float;
+}
+
+(** Estimate task complexity from description.
+    Uses word count as a proxy for scope, combined with reversibility. *)
+let estimate_complexity description =
+  let words = String.split_on_char ' ' description |> List.length in
+  let estimated_turns = max 1 (words / 5) in
+  let reversibility = estimate_reversibility description in
+  { estimated_turns; reversibility }
+
+(** Threshold above which tasks should be split.
+    Tasks with estimated_turns above this are flagged for decomposition. *)
+let needs_decomposition complexity =
+  complexity.estimated_turns > 20
+
+(* ================================================================ *)
+(* Idempotent Task Dispatch                                         *)
+(* ================================================================ *)
+
+(** Check if a task with the same title prefix already exists in backlog.
+    Prevents duplicate task creation when execute_plan is called multiple times. *)
+let task_exists_in_backlog config ~title_prefix =
+  let backlog = Room.read_backlog config in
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):

### CRITICAL: Substring matching causes false positives in reversibility heuristic

`estimate_reversibility` uses raw substring search (`has w` checks if `w` appears anywhere in the lowercased description). This causes dangerous misclassifications:

- `"drop"` matches `"dropdown menu"` -> wrongly scores 0.2 (destructive)
- `"list"` matches `"listen for events"` -> wrongly scores 1.0 (read-only)
- `"fix"` matches `"prefix all identifiers"` -> wrongly scores 0.8
- `"code"` matches `"barcode scanner"` -> wrongly scores 0.8
- `"remove"` matches `"remove-button CSS class"` -> wrongly scores 0.2

Since reversibility is used downstream for decomposition decisions, a false classification of a destructive task as safe (or vice versa) has real consequences.

**Fix**: Split description into words first (`String.split_on_char ' '`), then match against a word set. Or use `Re` library with word boundary matching.

### CRITICAL: 40-character prefix dedup causes false positives

`task_exists_in_backlog` truncates the title to 40 characters for matching. Given that `execute_plan` generates titles like `[goal:XXXX][child] Some Task Name`, the structured prefix `[goal:XXXX][child] ` alone consumes ~25-30 characters, leaving only 10-15 characters of actual task name for differentiation.

Two distinct child tasks under the same goal with similar names will collide:
- `[goal:abc][child] Implement auth for use` (truncated from "Implement auth for user service")
- `[goal:abc][child] Implement auth for adm` (truncated from "Implement auth for admin service")

Actually, both would become `[goal:abc][child] Implement auth for us` at 40 chars and the second would be silently dropped.

**Fix**: Match on the full normalized title (`String.lowercase_ascii (String.trim title)`) or use a deterministic hash/ID derived from goal_id + node_id.

### HIGH: No tests included

This PR introduces 4 new functions (`estimate_reversibility`, `estimate_complexity`, `needs_decomposition`, `task_exists_in_backlog`) and modifies `add_task_safe`. None have unit tests. Given the heuristic nature of the complexity analyzer, test coverage is especially important to prevent regressions and to document expected behavior for edge cases.

### HIGH: UTF-8 safety issue in `String.sub`

`String.sub title 0 40` operates on bytes, not characters. If the title contains multi-byte UTF-8 characters (Korean, Japanese, emoji), this can split a character in half, producing an invalid UTF-8 string. The truncated string is then used in `Printf.sprintf` for the return message and in prefix comparison.

**Fix**: Use a UTF-8-aware truncation or ensure the prefix boundary falls on a character boundary.

### MEDIUM: Word-count proxy for `estimated_turns` is unreliable

`String.split_on_char ' ' description |> List.length` is inflated by consecutive spaces and doesn't filter empty strings. A description `"fix  this   bug"` would count as 5 words (3 real + 2 empty).

More fundamentally, word count is a weak proxy for task complexity. A 3-word description like `"deploy to production"` is far more complex than a 20-word description like `"add a print statement to the debug log for the user login handler module"`.

**Fix**: At minimum, filter empty strings: `|> List.filter (fun s -> s <> "") |> List.length`. Consider whether this heuristic is useful enough to ship, or if `estimated_turns` should be left as `None` when only a description is available.

### MEDIUM: Imperative substring search is unidiomatic

The `has` function uses a `for` loop with `ref` boolean. OCaml stdlib has no built-in substring search, but this could use `Astring.String.find_sub` or a simpler `let has w = try let _ = Str.search_forward (Str.regexp_string w) desc 0 in true with Not_found -> false`. The current implementation works but is harder to read and maintain.

### MEDIUM: `estimate_complexity` and `estimate_reversibility` are unused in this PR

These functions are defined but never called within the modified `add_task_safe` or `execute_plan`. They're dead code in the current diff. If they're intended for PR #5070's `complexity_estimate` field, the integration path should be documented or the functions should be wired in.

### LOW: Reversibility returns raw float without domain type

Same as noted in PR #5070 review: a float in [0.0, 1.0] with semantic meaning would benefit from a variant type (`Irreversible | HighRisk | Safe | ReadOnly`) or at minimum a validated constructor.

### Summary

| Finding | Severity |
|---------|----------|
| Substring matching false positives in reversibility | CRITICAL |
| 40-char prefix dedup false positives | CRITICAL |
| No test coverage for 4 new functions | HIGH |
| UTF-8 safety in `String.sub` truncation | HIGH |
| Word-count proxy inflated by empty strings | MEDIUM |
| Unidiomatic imperative substring search | MEDIUM |
| `estimate_complexity`/`estimate_reversibility` unused | MEDIUM |
| Raw float for reversibility | LOW |

Reviewed by: Claude Opus 4.6 (cross-check) + GLM-5.1 (primary)

PR #5181, opened 2026-04-04
fix(keeper): delegate context budget trim to OAS Builder via with_max_input_tokens (fixes #5053)

jeong-sik:


jeong-sik:
Superseded by rebased PR with conflict resolution.

PR #5219, opened 2026-04-05
fix(ci): guard git calls and stop keeper retry on fatal Eio errors

jeong-sik:
## Summary
- Guard git calls with .git existence check before spawning subprocess
- Raise `Keeper_heartbeat_failure` on unrecoverable Eio context errors instead of infinite retry
- Raise CI test timeout from 900s to 2400s (5946 tests need headroom)
- Include keeper name in all turn error/exception logs
- Fix OCaml scoping bug in `with exn ->` handler: use `meta_after_triage.name` (function parameter, always in scope) instead of `meta_after_observe.name` (bound inside the `try` body, not in scope in the handler)

## Product impact

- Promise affected: `none/internal`
- User-visible change: None — internal keeper loop stability and CI reliability fix

## Evidence

- `dune build` passes locally
- Keeper turn with missing Eio context now raises fatal instead of retrying infinitely
- Exception handler now references an in-scope binding, fixing a compile-time scoping error that previously caused an unbound value

## Linked issue


jeong-sik:
Closing as superseded: the core git-guard fix was already merged via #5202 (bb1517814). The remaining items (Eio fatal, CI timeout bump, scoping fix) are either small enough for a fresh PR or already addressed.

PR #5508, opened 2026-04-06
fix(keeper): raise compaction gates to 75% of context window

jeong-sik:
- [x] Fix `keeper_guard.ml`: add `> 0` guards to message/token gate checks so 0 means "disabled"
- [x] Fix `env_config_snapshot.ml`: sync defaults to match new keeper_config defaults (ratio: 0.5→0.75, messages: 12→0, tokens: 4000→196608)
- [x] Fix `keeper_config.ml`: raise ratio default to 0.75 to align with token gate intent and cover smaller context window models
- [x] Final validation (code review + CodeQL passed)


jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/masc-mcp/pull/5508#pullrequestreview-4060917322)

PR #5600, opened 2026-04-06
fix(keeper): unblock compaction when reflection_ts=0 or ratio>=0.8

jeong-sik:
## Summary
- reflection_ready gate blocked compaction permanently when last_continuity_update_ts=0.0
- janitor hit 218K/200K context_overflow (#5634, #5599)
- Fix: ts<=0.0 bypasses cooldown, ratio>=emergency_compact_ratio_threshold emergency gate
- Refactor: extracted hardcoded `0.8` into named constant `emergency_compact_ratio_threshold` with odoc comment distinguishing it from `ratio_gate` and `handoff_threshold` (single definition point, no drift risk)
- Regression tests added for both gate-bypass paths; test correctness hardened after review

## Product impact

- Promise affected: `none/internal`
- User-visible change: Keepers with no prior reflection history (ts=0.0) and keepers under high context pressure (ratio≥0.8) now compact correctly instead of being permanently blocked

## Evidence

- New `compact_policy` Alcotest suite in `test/test_keeper_lifecycle.ml`:
  - `ts=0.0 bypasses cooldown gate` — asserts decision is `"blocked:below_thresholds"` (not `"skipped:continuity_reflection"`) when last_reflection_ts=0.0 and cooldown=3600s
  - `emergency ratio bypasses cooldown gate` — asserts same when ratio≥emergency_compact_ratio_threshold and cooldown not yet elapsed
- Emergency bypass test now uses `ratio_gate=10.0` (via new optional param on `make_gate_only_meta`) so the ratio gate itself never fires even when the constructed context ratio exceeds 1.0 — making it a true gate-only test
- Magic `0.8` threshold in test replaced with `KCP.emergency_compact_ratio_threshold` (via `module KCP = Masc_mcp.Keeper_compact_policy` alias) so the test stays in sync if the constant changes
- `emergency_test_text_length` named constant used in tests to document why the text length was chosen

## Linked issue

- Relates #5599
- Relates #5634

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -8,7 +8,7 @@
     "node": ">=20"
   },
   "scripts": {
-    "dev": "vite",
+    "dev": "MASC_DASHBOARD_PROXY_TARGET=http://127.0.0.1:8935 vite",
     "build": "tsc --noEmit && vite build",
     "typecheck": "tsc --noEmit",
     "preview": "vite preview",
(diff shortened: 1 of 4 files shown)


jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/masc-mcp/pull/5600#pullrequestreview-4064852329)

PR #5678, opened 2026-04-07
refactor(dashboard): merge recent activity into profile, consolidate KPI hints, fix debug section

jeong-sik:
- [x] Split debug journal and raw data into separate `<details>` elements
- [x] PR description updated with all review points  
- [x] Fix eager `AgentJournalStream` mount: `onToggle` handler updates `journalOpen` state; component is only rendered when `journalOpen === true`
- [x] Hooks (`useRef`, `useState`) moved before early return guard in `KeeperDetailOverlay` for correctness
- [x] Replied to new comments


jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/masc-mcp/pull/5678#pullrequestreview-4065906556)

PR #5823, opened 2026-04-08
fix(dashboard): prevent browser freeze on large trace entry expansion and add chevron

jeong-sik:
## Description

Fixes the issue where opening a trace entry (like a massive `keeper_shell_readonly` output) freezes the browser.

- Truncates `DiffBlock` mapping to max 500 lines to prevent overwhelming the DOM.
- Truncates plain text `ResultViewer` output to 100,000 chars.
- Adds an explicit right-aligned chevron icon on hover to `<summary>` elements to improve click affordance.

### DIFF
--- a/dashboard/src/components/session-trace/session-trace-entry.ts
+++ b/dashboard/src/components/session-trace/session-trace-entry.ts
@@ -95,7 +95,11 @@ function isLongContent(text: string): boolean {
 // ── Diff renderer ─────────────────────────────────────
 
 function DiffBlock({ text }: { text: string }) {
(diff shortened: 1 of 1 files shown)


jeong-sik:
GLM-5.1 cross-model review: LGTM. Truncation limits (500 lines diff, 100KB plain) are reasonable for browser performance. No XSS risk — Preact template escaping. Chevron rotation correct. Nit: consider opacity-40 default instead of opacity-0 for discoverability. Reviewed by: GLM-5.1

PR #5830, opened 2026-04-08
refactor: eliminate heuristic metadata matching in favor of deterministic Tool_dispatch sets

jeong-sik:
## Summary
Previously, `is_destructive_tool_name` and `is_idempotent_tool_name` relied on fragile, hardcoded substring matching (e.g., checking if a tool name contains `delete`, `remove`, or starts with `masc_get_`).
This PR removes these heuristics and enforces deterministic boundary registration. 

## Changes
- Extends `Tool_dispatch` to manage `destructive_set` and `idempotent_set`.
- `Tool_spec.register` now explicitly populates these dispatch sets from the tool's defined schema.
- Replaced the string-matching logic in `mcp_server_eio_tool_profile.ml` with O(1) Hashtbl lookups against the registered sets.

## Product impact
- **No more hardcoding**: Future tools that are destructive or idempotent are automatically evaluated based on their explicit `Tool_spec` declaration, completely removing the risk of a tool bypassing UI warnings or retry policies because its name didn't match a hardcoded regex.
- Alignment with the *Deterministic Verification First* pattern established in PR #814.

## Evidence
`dune build` completes successfully. No semantic changes to existing tools.

## Review evidence
Standard architecture cleanup.

## Linked issue
Refs #5830

### DIFF
--- a/lib/mcp_server_eio_tool_profile.ml
+++ b/lib/mcp_server_eio_tool_profile.ml
@@ -114,29 +114,6 @@ let tool_allowed_in_profile state profile tool_name =
                  String.equal schema.name tool_name))
   | Operator_remote -> List.mem tool_name Tool_operator.remote_tool_names
 
-let is_destructive_tool_name name =
-  let lowered = String.lowercase_ascii name in
-  List.exists
-    (fun fragment ->
-      let len = String.length fragment in
-      let lowered_len = String.length lowered in
-      let rec loop idx =
-        if idx + len > lowered_len then false
-        else if String.sub lowered idx len = fragment then true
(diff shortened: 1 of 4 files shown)


jeong-sik:
GLM-5.1 cross-model review: LGTM with one observation. Interface complete, follows existing init/query pattern. Unknown tools now default to non-destructive/non-idempotent — less conservative than heuristic but correct for registered tools (all tools go through Tool_spec.register). If unregistered tools reach this path, the fallback is false. Consider documenting this default in the .mli. Reviewed by: GLM-5.1

PR #5904, opened 2026-04-08
fix: dashboard layout on wide screens

jeong-sik:
## Changes
The recent styling introduced a `max-w-[1600px] mx-auto` constraint that erroneously centered the entire dashboard container on wide monitors, causing a cluster of information in the center instead of a full-width experience. This restriction has been lifted.


jeong-sik (commented):
Review: APPROVE — max-w-[1600px] 제거. wide screen 전체 폭 활용. 2줄, 안전.

PR #5921, opened 2026-04-08
[codex] Improve fleet telemetry evidence and cache freshness

jeong-sik:
## What changed
- persist Fleet worker-run metadata even when proof is absent, and normalize worker-run correlation fields so session events, worker evidence, and dashboard projections join cleanly
- hide raw trace path leakage from public telemetry payloads and surface `output_preview` plus raw-only worker evidence in proof/mission summaries
- add short proof HTTP caching with room/session/operation-scoped keys, plus mutation-driven cache invalidation so proof and mission surfaces refresh immediately after team-session writes
- align `turn_count` semantics with session orchestration turns instead of agent-result fanout, and update the team-session spec/report wording accordingly

## Why
Fleet telemetry was dropping raw-only runs, making proof evidence incomplete and correlation unnecessarily hard. The proof surface also relied on TTL-only freshness, so newly written worker-run evidence could remain stale for a short window.

## Impact
- raw trace runs without proof now remain visible in dashboard proof evidence
- Fleet completion events carry stable `worker_run_id`/`evidence_refs` joins
- proof reads refresh immediately after session or worker-run writes instead of waiting for cache expiry
- session `turn_count` is now consistent with orchestration-level turns

## Validation
- `dune runtest --root . test/test_dashboard_cache.exe test/test_dashboard_proof.exe test/test_dashboard_mission.exe test/test_team_session_swarm_runner.exe test/test_team_session_oas_bridge.exe test/test_worker_run_evidence_summary.exe`

## Root cause
Fleet persistence only wrote worker-run metadata when proof existed, while event normalization and dashboard summaries assumed correlation/evidence fields would always be present. Proof HTTP responses also lacked mutation-aware cache invalidation.

### DIFF
--- a/docs/spec/07-team-session.md
+++ b/docs/spec/07-team-session.md
@@ -98,7 +98,7 @@ Team Session Engine (Eio)
 
 | Field | Type | 설명 |
 |-------|------|------|
-| `turn_count` | int | 누적 turn 수 |
+| `turn_count` | int | 누적 session orchestration turn 수 (`team_turn` + swarm iteration) |
 | `agent_names` | string list | 참여 agent 목록 |
 | `planned_workers` | planned_worker list | 계획된 worker 사양 |
 | `broadcast_count` | int | broadcast 횟수 |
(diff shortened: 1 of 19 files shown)


jeong-sik (commented):
Review: Low-Medium risk. worker_run_id 충돌 수정은 안전. cache freshness 변경이 shared_cache를 건드리면 Eio mutex 보호 확인 필요. Ready 전 gh pr diff 5921로 확인 권장.

PR #5922, opened 2026-04-08
[codex] fix dashboard keeper telemetry health projection

jeong-sik:
## What changed
- fixed dashboard label timestamp parsing so UTC `Z` timestamps do not produce false stuck ages
- exposed repo/base-path diagnostics in the shell dashboard payload
- made keeper health projection prefer fresh live agent signals over stale persisted error state
- made continuity briefs use the freshest keeper signal source instead of only persisted `updated_at`
- patched execution cache keeper lifecycle updates so autobooted keepers surface `phase`, `pipeline_stage`, and `status` immediately
- added regression coverage for dashboard label parsing, shell path telemetry, and keeper health classification

## Why
The dashboard was misreporting keeper health in two ways:
- stale UTC parsing could show false long-running stuck states
- execution/cache projections could keep freshly running keepers in `offline`/`inactive` after autoboot because persisted error or pre-autoboot cache state won over live signals

## Impact
- operators get repo-scoped path truth directly in dashboard telemetry
- keeper health/status is closer to real runtime state during recovery and autoboot
- continuity views no longer overstate inactivity when agent presence is fresh

## Root cause
Multiple dashboard projections trusted stale persisted fields more than live runtime signals, and execution cache patching only toggled `keepalive_running` without updating the visible phase/status fields.

## Validation
- `dune build --root . bin/main_eio.exe test/test_keeper_exec_status.exe test/test_dashboard_execution.exe`
- `./_build/default/test/test_keeper_exec_status.exe`
- live verification on repo-scoped server `http://127.0.0.1:9725/health`
- live verification on repo-scoped server `http://127.0.0.1:9725/api/v1/dashboard/execution`

### DIFF
--- a/lib/dashboard/dashboard_execution_builders.ml
+++ b/lib/dashboard/dashboard_execution_builders.ml
@@ -169,6 +169,7 @@ let continuity_row_of_keeper ~(now_ts : float) ?related_session_id keeper :
     | None -> name
   in
   let audit = tool_audit_snapshot agent_name in
+  let agent = member_assoc "agent" keeper in
   let status = string_field ~default:"unknown" "status" keeper in
   let context_ratio =
     match member_assoc "context_ratio" keeper with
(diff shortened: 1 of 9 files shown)


jeong-sik (commented):
Review: diff 확인 필요. test + health projection fix로 보이나 구체적 변경 미확인. Ready 전 gh pr diff 5922로 timestamp null 처리 확인 권장.

jeong-sik (commented):
Review: APPROVE

- health projection 로직(warningCount/criticalCount 단일 카운트) 올바름
- timestamp unix seconds→ms 자동 판별 + locale offset 처리 올바름
- 테스트가 실제 동작(SSE debounce, mermaid graph, handoff 즉시 반영) 검증
- blocking 이슈 없음

PR #5925, opened 2026-04-08
[codex] SSOT naming and managed-operation lane cleanup

jeong-sik:
## Summary
- remove the internal `generic` stage sentinel and normalize legacy stage values to unset while keeping the `generic -> coding_task` workload alias for read compatibility
- centralize managed-agent alias canonicalization so `masc_set_current_task` and related aliases resolve through shared tool metadata instead of duplicated guidance logic
- move front-door docs and help away from presenting CPv2 as the default path, and describe managed operations as a compatibility/benchmark lane while keeping Team Session + Supervisor as the primary delivery path

## Why
The repo had the same concepts expressed differently across code, help text, and docs:
- `generic` meant both a deprecated workload alias and an internal stage sentinel
- canonical task-control naming was duplicated across workflow guidance and managed-agent compatibility surfaces
- CPv2 was still exposed as if it were the mainline front door even though the active implementation path is Team Session + Supervisor

## Impact
- new write/read paths no longer emit a `generic` stage sentinel
- legacy persisted `generic` stages still load and normalize to unset
- workload `generic` remains accepted only as a deprecated alias for `coding_task`
- user-facing help, runbooks, and LLM front-door docs now consistently steer callers toward namespace/task hygiene plus team-session/supervisor, with managed operations framed as an explicit compatibility lane

## Validation
- `./_build/default/test/test_cp_search_fabric.exe`
- `./_build/default/test/test_tool_surface_ssot.exe`
- `./_build/default/test/test_workflow_guide.exe`
- `./_build/default/test/test_command_plane_help.exe`

## Notes
- `dune build --root .` produced updated build artifacts, but the interactive dune process hung in this environment after compilation, so validation was completed by running the generated test executables directly.


jeong-sik (commented):
Review: Low risk. SSOT catalog 확장 + guidance_tool_name normalizer. guidance_tool_name에 masc_claim_task/complete_task/release_task/cancel_task 4개 alias가 누락 — 기본 arm으로 fallthrough하면 문제없지만, guidance가 필요하면 추가 필요.

PR #5933, opened 2026-04-08
fix: revert cn_llama indirection, propagate error instead

jeong-sik:
## Summary
- Revert `Provider_adapter.cn_llama` when-guards back to `"llama"` literals
- Fix the actual bug: error propagation instead of silent fallback in `tool_autoresearch`

## Why
`"llama"` is a stable external provider identifier. Wrapping it in a constant
added OCaml when-guard complexity with no type safety benefit (both are `string`).
The real HARD violation was silent error swallowing, not the string literal.

## Changes
- `tool_autoresearch.ml`: `Error _ -> "llama"` -> proper `Error` propagation
- `tool_team_session_routing.ml`: `when String.equal s Provider_adapter.cn_llama` -> `| "llama"`
- `tool_inline_dispatch.ml`: same revert to pattern match
- tests: pass explicit `model_model` in the autoresearch paths that previously relied on silent fallback

## Product impact
- `masc_autoresearch_start` now fails loudly when no default model is configured instead of silently routing to local llama
- the stable external provider identifier remains `"llama"` without extra when-guard indirection

## Evidence
- previous CI run `24141815050` failed on the old head because autoresearch tests still assumed silent fallback when no default model was configured
- follow-up commit `7025a81` supplies explicit `model_model` in the affected autoresearch tests
- current CI rerun for head `7025a81`: `24142662027`

## Review evidence
- direct GLM review requested via `sb glm-text` on 2026-04-09 KST; response still pending at the time of this update
- local `dune build --root . test/test_autoresearch_oas_primitives.exe` from this shell is blocked by an unrelated local `Agent_sdk.Types.ToolResult` type mismatch, so GitHub Actions is the source of truth for pass/fail on this branch

## Linked issue
- Refs #5680

## Test plan
- [ ] CI build passes
- [ ] CI tests pass

### DIFF
--- a/lib/tool_autoresearch.ml
+++ b/lib/tool_autoresearch.ml
@@ -77,11 +77,10 @@ let prepare_start_params (ctx : context) args =
   let source_workdir = get_string args "workdir" ctx.base_path in
   let max_cycles = get_int args "max_cycles" 100 in
   let cycle_timeout_s = get_float args "cycle_timeout_s" 300.0 in
-  let default_model =
-    match Provider_adapter.default_model_label_result () with
-    | Ok label -> label
-    | Error _ -> Provider_adapter.cn_llama
-  in
+  let default_model_result = Provider_adapter.default_model_label_result () in
+  match default_model_result with
+  | Error e -> Error (Printf.sprintf "no default model configured: %s" e)
+  | Ok default_model ->
   let model_model = get_string args "model_model" default_model in
   if goal = "" then
     Error "goal is required"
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
Review: APPROVE

- autoresearch: model 미설정 시 silent fallback 대신 명시적 Error 반환 — 올바른 방향
- cn_llama indirection revert: "llama" 문자열이 3곳에서만 사용되므로 상수화 이득이 적음. 직접 문자열이 더 명확
- 테스트에 model_model 필수 전달 추가

PR #5955, opened 2026-04-08
fix(approval_queue): clean up pending entry on fiber cancellation

jeong-sik:
## Summary
Fix orphan entry leak in Keeper_approval_queue when agent fiber is cancelled.

## Problem
When `Eio.Switch` is cancelled (e.g., keeper shutdown), fibers in `submit_and_await` receive `Eio.Cancel.Cancelled` but the pending hashtbl entry was not cleaned up — leading to resolver orphans.

## Fix
Wrap `Eio.Promise.await` in `Fun.protect ~finally` that removes the entry. The `~finally` handler catches exceptions internally to avoid `Fun.Finally_raised`.

## Evidence
- New test: `cancel cleans up` — cancels a Switch while a fiber is awaiting, verifies `pending_count` returns to initial value
- 13/13 tests pass (0.006s)

Closes #5949

### DIFF
--- a/lib/keeper/keeper_approval_queue.ml
+++ b/lib/keeper/keeper_approval_queue.ml
@@ -77,8 +77,17 @@ let submit_and_await ~keeper_name ~tool_name ~input ~risk_level
    with
    | Eio.Cancel.Cancelled _ as e -> raise e
    | _ -> ());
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Review: APPROVE — Fun.protect cleanup on cancel, test included

PR #6008, opened 2026-04-09
[codex] fix: clamp keeper boring exit threshold env defaults

### LINKED ISSUE #6005: fix: clamp keeper boring exit threshold env defaults
## Symptom

`keeper.turn.boring_exit_threshold` was added with `min=2` and `max=50`, but the env-backed default path bypasses that clamp.

If `MASC_KEEPER_BORING_EXIT_THRESHOLD` is set outside the documented range, keepers can still read and use the invalid value.

## Reproduction

1. Set `MASC_KEEPER_BORING_EXIT_THRESHOLD=1` or `100`
2. Read `Keeper_config.keeper_boring_exit_threshold ()`
3. Observe the raw out-of-range value being returned from the default thunk

## Expected

The env-backed default should honor the same `[2, 50]` bounds as the runtime param metadata and dashboard validation.

## Suspected Cause

`keeper_boring_exit_threshold_rp` used a custom `Sys.getenv_opt` + `int_of_string` fallback instead of the existing `int_of_env_default` helper. `Runtime_params.get` does not validate defaults, so out-of-range env values are not corrected.

## Tried

- Reproduced the behavior by tracing the new param through `keeper_config.ml` and `runtime_params.ml`
- Prepared a fix that routes the env-backed default through `int_of_env_default`
- Added regression coverage in `test/test_runtime_params.ml`

jeong-sik:
## Summary
- clamp the env-backed default for `keeper.turn.boring_exit_threshold` through the shared `int_of_env_default` helper
- add runtime-param regression coverage for registration metadata and env clamp behavior
- align the keepalive boring-gate comment with the new configurable threshold behavior

## Root cause
- `keeper.turn.boring_exit_threshold` used a custom `Sys.getenv_opt` + `int_of_string` path for its default value
- `Runtime_params.get` returns defaults without running validator hooks, so the documented `min=2` / `max=50` bounds were bypassed for env-sourced values

## Product impact
- `MASC_KEEPER_BORING_EXIT_THRESHOLD` now respects the documented `[2, 50]` range
- keepers no longer silently accept invalid boring-exit thresholds from the environment

## Evidence
- `dune build --root .`
- `./_build/default/test/test_runtime_params.exe`

## Review evidence
- Reviewer: GLM-5.1 via direct `sb glm-text`
- Scope: follow-up diff after merged PR #5997
- Result: `No findings.`

## Linked issue
- Fixes #6005

### DIFF
--- a/lib/keeper/keeper_config.ml
+++ b/lib/keeper/keeper_config.ml
@@ -710,9 +710,8 @@ let keeper_slot_id (name : string) : int option =
 let keeper_boring_exit_threshold_rp =
   _rp_int ~key:"keeper.turn.boring_exit_threshold"
     ~default:(fun () ->
-      match Sys.getenv_opt "MASC_KEEPER_BORING_EXIT_THRESHOLD" with
-      | Some s -> (try int_of_string s with _ -> 8)
-      | None -> 8)
+      int_of_env_default "MASC_KEEPER_BORING_EXIT_THRESHOLD"
+        ~default:8 ~min_v:2 ~max_v:50)
     ~min_v:2 ~max_v:50
     ~description:"Consecutive boring turns before exit_condition triggers and keepalive skips proactive turns" ()
 
--- a/lib/keeper/keeper_keepalive.ml
+++ b/lib/keeper/keeper_keepalive.ml
@@ -697,13 +697,13 @@ let run_keepalive_unified_turn
           ~meta:meta_after_triage
           obs
       in
-      (* Boring-turn gate: after 8+ consecutive idle turns on a non-reactive
-         channel, skip the turn entirely. Reactive turns (mentions, board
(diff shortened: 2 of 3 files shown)


jeong-sik:
## Review evidence
- Updated: 2026-04-09 09:18 KST
- Reviewer: GLM-5.1 via direct `sb glm-text`
- Scope: follow-up diff after merged PR #5997
- Result: `No findings.`

## Local validation
- `dune build --root .`
- `./_build/default/test/test_runtime_params.exe`

## Notes
- Root cause: `Runtime_params.get` does not validate default thunks, so the env-backed default for `keeper.turn.boring_exit_threshold` needed to clamp before registration returns it.
- Regression coverage now checks registration metadata and env clamp behavior explicitly.

PR #6028, opened 2026-04-09
[codex] remove collaboration_context bridge

jeong-sik:
## What changed
- stop projecting team-session state into `swarm_config.collaboration_context`
- remove the now-unused `Team_context.to_json` API and its implementation
- update the bridge test to assert that `collaboration_context` is intentionally omitted
- align boundary and team-session docs with the new bridge truth

## Why
`collaboration_context` is a retired bridge concept. The current team-session path already preserves context through prompt composition and `worker_specs`, so keeping a parallel JSON projection only adds redundant surface area and doc drift.

## Impact
- team-session swarm configs now omit `collaboration_context` by design
- prompt-level team context remains intact
- docs and tests now match the actual bridge behavior

## Root cause
A previous change introduced a JSON `collaboration_context` projection while multiple SSOT docs still described the bridge as `None`. That left the codebase with two competing truths for the same boundary.

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/remove-collaboration-context`
- `./_build/default/test/test_team_session_oas_bridge.exe`


jeong-sik:
Superseded by #6010 (already merged to main). Closing duplicate.

PR #6031, opened 2026-04-09
[codex] deduplicate dated_jsonl count_entries

jeong-sik:
## What changed
- remove the duplicate `count_entries` declaration from `dated_jsonl.mli`
- remove the second `count_entries` implementation and its private newline-count helper from `dated_jsonl.ml`
- keep the existing non-empty-row counting implementation as the single source of truth

## Why
`dune build --root .` failed on `main` with warning 32 because `count_entries` was declared twice in the interface. The implementation also had two competing definitions, and the later one counted raw newlines instead of non-empty JSONL rows.

## Impact
- restores a clean build on the current branch tip
- keeps `Dated_jsonl.count_entries` aligned with the documented/tested non-empty-row semantics
- removes dead duplicate code

## Root cause
A duplicate API declaration and a second implementation were left in `dated_jsonl`, so the interface tripped warning 32 and the runtime semantics were ambiguous.

## Validation
- `bash scripts/check-oas-pin.sh --local-only`
- `dune build --root .`
- `./_build/default/test/test_dated_jsonl.exe`

### DIFF
--- a/lib/dated_jsonl/dated_jsonl.ml
+++ b/lib/dated_jsonl/dated_jsonl.ml
@@ -303,36 +303,3 @@ let prune t ~days =
     ) months;
     !deleted
   end
-
-let count_lines_in_file path =
-  if not (Fs_compat.file_exists path) then 0
-  else
-    let ic = open_in_bin path in
-    Fun.protect ~finally:(fun () -> close_in_noerr ic) (fun () ->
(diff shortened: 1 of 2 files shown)


jeong-sik:
count_entries dedup already fixed in main via #6022. Closing.

Issue #6001, opened 2026-04-08
refactor: move provider_filter from MASC to OAS cascade API


jeong-sik:
## Problem

`oas_model_resolve.ml:filter_by_providers` is a MASC module that pre-filters the cascade model list before passing it to OAS. This violates the MASC-OAS boundary principle: provider routing is OAS infrastructure.

Current path:
```
MASC: filter_by_providers(allowed, full_list) → reduced_list
MASC → OAS: complete_named(~defaults:reduced_list, ...)
```

Should be:
```
MASC → OAS: complete_named(~defaults:full_list, ~provider_filter:["ollama"], ...)
OAS: apply filter internally before cascade execution
```

## Evidence
- `lib/oas_model_resolve.ml` — named `oas_*` but lives in MASC
- `lib/keeper/keeper_exec_context.ml:144-145` — MASC pre-filters
- `test/test_keeper_toml.ml:654-683` — tests exist but in MASC

## Plan
1. OAS: Add `?provider_filter:string list` to `Cascade_config.complete_named`
2. OAS: Apply filter after model parsing, before cascade execution
3. MASC: Pass `allowed_providers` as `~provider_filter` to OAS
4. MASC: Remove `Oas_model_resolve.filter_by_providers`
5. Pin bump

PR #6051, opened 2026-04-09
[codex] Realign OAS pin floor and docs

jeong-sik:
## Summary
- pin OAS to the last self-consistent commit that already includes `provider_filter` support (`0a71c51`, `0.117.0`)
- realign `dune-project`, `scripts/oas-agent-sdk-pin.sh`, and OAS docs around that same floor/pin
- harden `scripts/check-oas-pin.sh` so repo manifest/doc drift is reported before local sibling checkout drift

## Why
`main` had a partial OAS bump: `scripts/oas-agent-sdk-pin.sh` and `dune-project` moved to `0.118.0`, but `masc_mcp.opam` and docs did not. Upstream OAS commit `33397b8` also bumped the SDK version without regenerating `agent_sdk.opam`, so the `0.118.0` pin was not self-consistent.

`0a71c51` already contains the `provider_filter` API needed by `#6001`, while keeping `dune-project`, `agent_sdk.opam`, and `lib/sdk_version.ml` aligned at `0.117.0` in OAS.

## Validation
- `git diff --check`
  - after this patch, repo-local pin/doc drift is gone and the remaining failure points at local environment drift instead of stale repo state
- `AGENT_SDK_LOCAL_REPO=/tmp/oas-6048-pin-check scripts/check-oas-pin.sh --local-only`
  - this pushes validation past sibling checkout drift and shows the remaining local opam-switch mismatch (`installed agent_sdk version is 0.118.0, expected 0.117.0`)

## Remaining local repair
- local sibling OAS checkout: `git -C ~/me/workspace/yousleepwhen/oas checkout 0a71c510ba251397f19f82590fac951ef59986d9`
- local opam switch: `bash scripts/opam-pin-external-deps.sh && opam install . --deps-only --with-test --with-doc -y`

## Linked issue
- Refs #6048


jeong-sik:
main already includes OAS pin f5247fcd + opam floor 0.118.0 (via version reset merge). All changes in this PR are already on main.

PR #6181, opened 2026-04-09
[codex] Fix keeper tool failure handling and classification

### LINKED ISSUE #6167: Fix keeper read-path contract mismatch behind path_not_in_allowed_paths storms
## Symptom
Keepers repeatedly fail with `path_not_in_allowed_paths` when they try to search the repo using `path: "."`, `path: "lib"`, or an absolute repo path.

Recent examples from `/Users/dancer/me/.masc/tool_calls/2026-04/09.jsonl`:
- `cheolsu`: `rg operator_approval` on `path:"."` (`line 519`)
- `analyst`: multiple `find`/`rg` calls on `.` and repo absolute paths (`lines 545, 547, 548, 575-577`)
- `sangsu`: `rg Eval_gate` / `rg input_required` on default `.` (`lines 607, 1315+`)

## Reproduction
1. Run a workspace-scope keeper with default read paths.
2. Call `keeper_shell_readonly` or `keeper_shell` with `path:"."` or `path:"lib"`.

## Suspected Cause
- `effective_allowed_paths` no longer includes project root `.` by default.
- Read tools still default `path` to `.` and keepers still reason as if repo-root search is allowed.
- For read ops, `cwd` does not compensate for a blocked `path`, so retries continue to fail.

## Attempted Investigation
- Verified the allowlist construction and read-path resolution in `lib/keeper/keeper_alerting_path.ml`.
- Correlated repeated failures across multiple keepers in Apr 9 tool-call logs.

## Proposed Fix
- Decide the intended contract explicitly:
  - either restore repo-root read access for the relevant read-only presets, or
  - rewrite prompts/tool contract so keepers only search inside allowed playground/worktree roots.
- Add tests for `path:"."`, `path:"lib"`, and repo absolute paths under each preset.

jeong-sik:
## Summary

This PR fixes the main sources behind the recent keeper tool failure spikes:

- false-positive safety guards that blocked normal queries and shell commands
- readonly `keeper_shell` path resolution mismatches between `cwd` and `path`
- timeout sentinel surfacing as fake `signal:-11`
- dashboard bucketing that collapsed distinct failures into generic `tool call failed`

## Root Cause

The runtime contract had drifted in three places:

- substring-based safety filters treated normal input like secrets or destructive commands
- read-only shell tools still interpreted relative `path` values against the wrong base even when `cwd` was provided
- `Process_eio` timeout sentinel (`WSIGNALED Sys.sigterm`) leaked into tool output as a misleading crash-like signal

## What Changed

- tightened web-search secret detection to token-prefix matching instead of substring matching
- tightened readonly bash destructive guard matching to token sequences instead of raw substring checks
- resolved `keeper_shell` read paths relative to `cwd` when present, including omitted and `.` paths
- normalized dashboard failure categories for path errors, message-only errors, process exits, signals, and timeouts
- surfaced timeout sentinel as `status.kind = "timeout"` instead of fake `signaled:-11`
- added schema support for readonly bash `timeout_sec`
- exported missing `keeper_config` values needed by downstream modules

## Validation

- `git diff --check`
- `ocamlc -stop-after parsing -impl lib/keeper/keeper_alerting_path.ml`
- `ocamlc -stop-after parsing -impl lib/dashboard/dashboard_http_tool_quality.ml`
- `ocamlc -stop-after parsing -impl lib/keeper/keeper_exec_shell.ml`
- `ocamlc -stop-after parsing -impl lib/tool_misc_web_search.ml`
- `ocamlc -stop-after parsing -impl lib/tool_shard.ml`
- `ocamlc -stop-after parsing -impl test/test_keeper_pr_workflow.ml`
- `ocamlc -stop-after parsing -impl test/test_tool_misc_coverage.ml`
- `ocamlc -stop-after parsing -impl test/test_tool_quality_classify.ml`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-6166-guard-false-positives test/test_tool_quality_classify.exe test/test_keeper_pr_workflow.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-6166-guard-false-positives test/test_tool_quality_classify.exe`
- `./_build/default/test/test_keeper_pr_workflow.exe test bash_branch_guard`

## Issues

- Fixes #6166
- Fixes #6167
- Fixes #6168
- Fixes #6169


jeong-sik:
Cross-model review evidence (`2026-04-09 19:24:53 KST`)

- Reviewer path: direct `sb glm-text` on the default GLM review path (`GLM_TEXT_MODELS` starts with `glm-5.1`)
- Checks at review time: `label-and-review`, `phantom-fix-check`, `pr-hygiene`, and `sweep-supersede-check` were all passing
- Initial external review found one actionable issue: the readonly bash token matcher could over-block harmless text such as `echo git push`
- Patched in `38739bf9e` by narrowing readonly bash matching to the actual command entrypoint / git subcommand path instead of scanning arbitrary token sequences
- Added regression coverage for both sides of that fix:
  - allow `echo git push`
  - keep blocking `git -C repo push origin main`

Follow-up review on the updated diff did not produce any additional findings that required code changes. The remaining review notes were assessed as non-blocking:

- `normalize_failure_text` currently compares against lowercase prefixes, and the emitted error strings in this path already normalize to those values
- `blocked_pattern` strings such as `rm ` are diagnostic text only; the actual readonly guard decision is token-based
- explicit absolute read paths intentionally remain valid even if a provided `cwd` is invalid, because `cwd` is only used to resolve omitted/relative read paths
- `status.kind = "timeout"` is intentionally scoped to the `Process_eio` timeout sentinel surfacing in keeper tool outputs, which was the root cause behind the misleading `signal:-11` bucket

Current reviewer verdict: no remaining material blockers after the follow-up patch and rerun.

PR #6206, opened 2026-04-09
fix: harden dashboard telemetry refresh flows

jeong-sik:
## Summary
- add shared visible-only auto-refresh support for dashboard telemetry surfaces
- harden telemetry and fleet panels against out-of-order refresh responses
- extend tool quality with auto-refresh and wire lab tool-quality into route refresh

## Testing
- `pnpm vitest run src/components/telemetry-unified.test.ts src/components/fleet-telemetry-panel.test.ts src/components/tool-quality-panel.test.ts src/tab-refresh.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm typecheck`
- `pnpm build`

### DIFF
--- a/dashboard/src/api/core.test.ts
+++ b/dashboard/src/api/core.test.ts
@@ -41,6 +41,26 @@ describe('post', () => {
 })
 
 describe('get bootstrap warm-up mapping', () => {
+  it('preserves upstream abort signals instead of reporting them as timeouts', async () => {
+    const fetchMock = vi.fn().mockImplementation((_path: string, init?: RequestInit) => (
+      new Promise((_resolve, reject) => {
+        const signal = init?.signal as AbortSignal | undefined
+        signal?.addEventListener('abort', () => {
+          reject(new DOMException('superseded request', 'AbortError'))
+        })
+      })
+    ))
+    vi.stubGlobal('fetch', fetchMock)
+
+    const controller = new AbortController()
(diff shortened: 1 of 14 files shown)


jeong-sik:
Follow-up update: abort superseded telemetry requests

HEAD: `2ac34ee7ca03a02c2eb78012f17435d61c278857`
Reviewed at: 2026-04-09 23:02:09 KST

What changed:
- Added `AbortSignal` plumbing to dashboard GET helpers so caller-driven cancellation is preserved instead of being reported as a timeout.
- `TelemetryUnified` now aborts the previous in-flight load before starting a new one and aborts the active load on unmount.
- `FleetTelemetryPanel` now does the same for execution/tool-quality/telemetry-summary refreshes.
- Added regression coverage for upstream abort propagation and for superseded refresh cancellation in both telemetry panels.

Verification:
- `pnpm vitest run src/api/core.test.ts src/lib/auto-refresh.test.ts src/components/telemetry-unified.test.ts src/components/fleet-telemetry-panel.test.ts src/components/tool-quality-panel.test.ts src/tab-refresh.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm typecheck`
- `pnpm build`

Cross-model review (`GLM-5.1` via `sb glm-text`): no material findings. One low note pointed out that `TelemetryUnified` lets aborts from the store-side helper path bubble to the outer catch; that path is intentional because the outer catch already treats `AbortError` as a silent supersession/unmount case and avoids any stale state update.

PR #6292, opened 2026-04-10
[codex] harden core failure boundaries and parsers

jeong-sik:
## Summary
- remove silent process fallback failures by surfacing diagnostic output and explicit exit codes in `process_eio`
- move autoresearch persisted-state/cycle/swarm-link decoding onto result-based parsers and update storage/dashboard paths to use them
- replace `Stdlib.Mutex` in selected keeper Eio helpers and add regression tests for the touched failure-boundary and telemetry surfaces

## Product impact
- operators can distinguish real subprocess failures and verifier degradation from empty output
- autoresearch read paths stop silently accepting malformed persisted state
- keeper async/evidence/file-tracker helpers are aligned with Eio-safe concurrency expectations in the touched scope

## Evidence
- `dune exec --root . test/test_process_eio_coverage.exe`
- `dune exec --root . test/test_verifier_oas_bridge.exe`
- `dune exec --root . test/test_dashboard_autoresearch.exe`
- `dune exec --root . test/test_auto_recall_activity_coverage.exe`
- `dune exec --root . test/test_keeper_mutex_coverage.exe`
- attempted `dune build --root .`, but this machine had competing long-lived dune jobs and unreliable `_build/.lock` churn; targeted touched-surface builds passed

## Review evidence
- pending: direct `sb glm-text` external review on `origin/main...feat/ocaml-core-audit-wave1`

## Linked issue
- Refs #6290
- Refs #6291

Issue #6291, opened 2026-04-10
Parse-don\'t-validate hardening for team_session_types.session_of_yojson


jeong-sik:
## Problem
session_of_yojson currently accepts malformed or incomplete persisted team-session payloads by filling in defaults like "unknown", "default", or "failed". That weakens SSOT boundaries and can silently rewrite state truth.

## Why deferred
This path needs an explicit compatibility policy before tightening because persisted artifacts and read surfaces may rely on the current defaulting behavior.

## Expected direction
- introduce a strict result-returning parser at the persistence boundary
- keep any legacy/defaulting wrapper isolated and explicit
- reject unknown enums and missing required fields with structured errors
- add regression tests for malformed persisted session JSON

## Acceptance
- storage/read paths can distinguish missing data from invalid persisted state
- parser errors are observable in logs/diagnostics instead of silently coerced defaults

jeong-sik:
team_session_types was deleted entirely in PR #6303 (fedc939). The module no longer exists — no session_of_yojson to harden.

PR #6293, opened 2026-04-10
[codex] Harden team session JSON parsing

jeong-sik:
## Summary
- harden `Team_session_types` persisted JSON decoders with explicit `result`-based parser boundaries
- reject missing required fields and unknown enum values instead of silently defaulting session/checkpoint state
- add regression coverage for malformed team-session payloads, planned workers, nested delivery verdicts, and checkpoint status

## Why
Wave 1 deferred the permissive `team_session_types` decoder cleanup into #6291. The existing parser silently rewrote invalid persisted payloads with defaults like `failed`, `default`, and dropped malformed workers. This stack removes that silent recovery path and keeps the legacy option-returning wrapper only as a compatibility layer over the strict decoder.

Stacked on:
- #6292

Closes:
- #6291

## Validation
- `dune build --root . test/test_team_session_types_parsing.exe`
- `dune exec --root . test/test_team_session_types_parsing.exe`
- `dune build --root . test/test_governance_pipeline.exe test/test_dashboard_mission.exe`
- `./_build/default/test/test_governance_pipeline.exe`
- `./_build/default/test/test_dashboard_mission.exe`

### DIFF
--- a/lib/team_session_types/team_session_types.ml
+++ b/lib/team_session_types/team_session_types.ml
@@ -360,6 +360,16 @@ let required_bool_field kind json field =
         (field_error kind field
            (Printf.sprintf "expected bool, got %s" (json_type_name value)))
 
+let optional_bool_field kind json field =
+  match member field json with
+  | `Null -> Ok None
+  | `Bool value -> Ok (Some value)
+  | value ->
+      Error
+        (field_error kind field
+           (Printf.sprintf "expected bool or null, got %s"
+              (json_type_name value)))
+
 let required_enum_field kind json field parse =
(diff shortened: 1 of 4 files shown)


jeong-sik:
## Review evidence
- reviewer: `sb glm-text` (direct GLM path)
- timestamp: `2026-04-10T03:53:56Z`
- scope: `feat/ocaml-core-audit-wave1...feat/ocaml-core-audit-wave2-team-session`
- context: intentional strict parser hardening for `#6291`; silent default removal treated as expected behavior, not regression
- result: `No findings.`

PR #6320, opened 2026-04-10
fix: lower agent_sdk pin floor to 0.118.1

jeong-sik:
Superseded: main already bumped to v0.118.2 via another PR.


jeong-sik:
Cross-model review: GLM-5.1 via sb glm-text, 2026-04-10 KST. Trivial pin floor change (0.118.2 → 0.118.1). No findings.

PR #6307, opened 2026-04-10
feat(keeper): Guard → Thompson bridge (Phase B1)

jeong-sik:
## Summary
- Guardrail_stop 발생 시 Thompson β에 negative signal 기록
- Guard → Thompson → ToolPolicy 피드백 루프의 첫 구현 단계
- `MASC_DECISION_LAYER_LEVEL >= 2`에서만 활성화 (Level 1은 audit만)

## Changes
| File | Change |
|------|--------|
| `lib/thompson_sampling.ml` | `record_guard_penalty` 추가 (β += 0.5, configurable) |
| `lib/thompson_sampling.mli` | 인터페이스 export |
| `lib/keeper/keeper_keepalive.ml` | heartbeat에서 guardrail_stop 시 bridge 호출 |
| `lib/keeper/keeper_decision_audit.mli` | `decision_layer_level` export |

## Design
- Penalty cap 1/cycle: guard가 heartbeat당 1회만 평가하므로 자연적 보장
- β nudge magnitude: 0.5 (quality_fail과 동등, `MASC_GUARD_PENALTY_BETA`로 조정 가능)
- TLA+ 대응: KeeperDecisionPipeline.tla의 `GuardFires` action

## Context
- Plan: Keeper Decision Layer v2 (Rev.5), Phase B1
- Tracking: #6230
- Depends on: Phase A (#6244 merged), Phase B0 (#6277 merged)
- TLA+ verified: FailingEventuallyRecovers holds with penalty cap + recovery floor

## Test plan
- [x] `dune build` passes
- [ ] CI Build and Test
- [ ] CI TLA+ Model Checking (no spec change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6323, opened 2026-04-10
fix(keeper): address remaining Copilot review on Guard→Thompson bridge

jeong-sik:
## Summary
- Re-raise `Eio.Cancel.Cancelled` in the guard→Thompson try/with handler in `keeper_keepalive.ml`. The previous handler caught all exceptions, silently swallowing fiber cancellation signals.
- Fix `audit_enabled` docstring in `keeper_decision_audit.mli`: was "Cached at module init" but the implementation delegates to a lazy `decision_layer_level` (cached on first call).

Follows up on #6307 which was squash-merged before these two fixes landed.

## Test plan
- [ ] `dune build @install` passes (verified locally).
- [ ] Confirm Eio.Cancel.Cancelled propagation is not broken by searching for similar patterns in the codebase.
- [ ] Docstring accuracy: `audit_enabled` now says "Cached on first call (delegates to decision_layer_level)".


jeong-sik:
Cross-model review: GLM-5.1 via sb glm-text, 2026-04-10 KST. Result: No findings. Critical Eio.Cancel fix + docstring correction.

PR #6340, opened 2026-04-10
fix(ci): read agent_sdk from active opam switch

jeong-sik:
## Summary
- run the installed `agent_sdk` checks under `opam exec` so CI reads the active switch, not ambient global state
- keep OAS pin verification behavior unchanged apart from the switch source of truth

## Verification
- bash scripts/check-oas-pin.sh

## Review evidence
- pending GLM-5.1 review

Refs #6295

### DIFF
--- a/scripts/check-oas-pin.sh
+++ b/scripts/check-oas-pin.sh
@@ -113,7 +113,7 @@ version_gte() {
 }
 
 if command -v opam >/dev/null 2>&1; then
-  installed_packages="$(opam list --installed --columns=name,version --short 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review evidence:
- Reviewer: `GLM-5.1` via `sb glm-text`
- Timestamp: `2026-04-10 14:38 KST`
- Result: `No findings.`

Verification:
- `bash scripts/check-oas-pin.sh`

Note:
- The first GLM request attempt hit a 20s transport timeout before the cascade returned the final review result above.

PR #6344, opened 2026-04-10
fix: remove redundant match cases in http error handlers

jeong-sik:
## Summary
- `http_error` type has exactly 3 variants — catch-all wildcard at line 126/134 is unreachable
- OCaml warning 11 (redundant-case) is fatal in CI, blocking Build and Test on all branches
- Removes the dead `| _ -> "Unknown Error"` branch from both `tool_local_runtime_bench.ml` and `tool_local_runtime_verify.ml`

## Test plan
- [ ] CI Build and Test passes without warning 11

Generated with [Claude Code](https://claude.com/claude-code)


jeong-sik:
Duplicate of #6342 (identical fix). Closing in favor of the Codex PR.

PR #6355, opened 2026-04-10
fix(keeper): preserve docker playground safety gates

jeong-sik:
## Summary
- preserve keeper_bash write and branch-switch gates even when Docker playground routing is enabled
- harden Docker playground cwd mapping with path-boundary checks and sanitized keeper names
- add regression coverage for Docker playground mapping and policy gates

## Test plan
- [x] bash scripts/check-feature-flag-consistency.sh
- [x] ./_build/default/test/test_keeper_bash_safety.exe
- [x] Docker-specific gate cases in ./_build/default/test/test_keeper_pr_workflow.exe
- [ ] Full PR CI

## Notes
-  still has one pre-existing unrelated failure on this branch: . The new Docker gate cases pass.

### DIFF
--- a/lib/config/env_config_keeper.ml
+++ b/lib/config/env_config_keeper.ml
@@ -419,6 +419,23 @@ end
     Thresholds used by the dashboard keeper health scorer and harness health
     panels.  Distinct from compaction triggers — these affect UI display only. *)
 
+(** {1 Docker Playground} *)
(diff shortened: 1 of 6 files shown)


jeong-sik:
Docker playground routing is already merged into main. Closing as superseded.

PR #6459, opened 2026-04-11
fix(dashboard): add changeOrigin to vite proxy — 403 on mutations

jeong-sik:
## Summary
- Browser POST sends `Origin: http://localhost:5173`, backend sees cross-origin vs `localhost:8935`
- `changeOrigin: true` rewrites Origin to match proxy target
- Fixes 403 Forbidden on `masc_board_comment` and other mutation tools from dashboard

## Test plan
- [ ] Restart vite dev server, board comment from dashboard works without 403

### DIFF
--- a/dashboard/vite.config.ts
+++ b/dashboard/vite.config.ts
@@ -26,9 +26,9 @@ export default defineConfig(({ command }) => {
     server: proxyTarget
       ? {
           proxy: {
-            '/api': { target: proxyTarget, changeOrigin: true },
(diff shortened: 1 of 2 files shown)


jeong-sik:
Copilot review addressed:
1. changeOrigin reverted — only changes Host, not Origin. No effect on this check.
2. Loopback port relaxation is intentional. Remote attackers cannot forge localhost/127.0.0.1 Origin. Security rationale comment added to code.

Issue #6486, opened 2026-04-11
security: remove committed localhost private key and rotate if reused


jeong-sik:
### Product area

Transport, auth, and health

### Target lane

now

### Promise affected

repo coordination

### Product gap

- [x] Code may already exist, but the product promise is still not trustworthy.

### Observed version

main HEAD

### Severity

blocks release or data safety

### Steps to reproduce

1. Inspect the tracked cert files in the repo.
2. Run `git ls-files certs/localhost.key certs/localhost.pem` and confirm both are versioned.
3. Open `certs/localhost.key` and observe a real PEM private key (`-----BEGIN PRIVATE KEY-----`).
4. Run `openssl x509 -in certs/localhost.pem -noout -subject -issuer -dates -fingerprint -sha256` and confirm it is a localhost self-signed cert with a long validity window.

### Expected behavior

No private key material should be committed to the public repository. Local TLS certs should be generated outside git, ignored, or replaced with non-sensitive fixtures that cannot be reused as live credentials.

### Actual behavior

`certs/localhost.key` is currently tracked in the public repo as a real PEM private key. Even if it is only intended for localhost self-signed development use, the private key is public and must be treated as compromised for any reused context.

### Evidence

- Tracked files: `certs/localhost.key`, `certs/localhost.pem`
- Cert subject/issuer: `CN=localhost`
- Cert validity: 2026-02-02 through 2036-01-31
- SHA256 fingerprint available from local `openssl x509` output

### Scope notes

This should be handled as a security hygiene blocker before normal issue-loop work:
- remove the tracked private key from the repo
- rotate/regenerate if the same key was reused anywhere else
- replace the current setup with documented local generation or ignored dev-only artifacts
- consider history cleanup if the project wants stronger public-repo hygiene beyond forward removal

jeong-sik:
Resolved by PR #6487 (merged 2026-04-11T11:11:31Z).

**What was done:**
- `certs/localhost.key` (52-line PEM private key) and `certs/localhost.pem` (30-line self-signed cert) removed from tracked files
- `certs/*.key` and `certs/*.pem` added to `.gitignore` (lines 80-81)
- No remaining references to these files in the codebase
- The `certs/` directory no longer exists on HEAD

**Scope assessment:**
- The key was a localhost self-signed development cert (CN=localhost, validity 2026-02-02 to 2036-01-31)
- No code references the cert files — no loader, no TLS config pointing to `certs/localhost.*`
[...]

PR #6487, opened 2026-04-11
security: remove tracked localhost TLS key

jeong-sik:
## Summary
- remove tracked `certs/localhost.key` and `certs/localhost.pem` from the repo
- ignore checked-in PEM/key material under `certs/` so local TLS artifacts do not get recommitted
- keep TLS configuration driven by env/config rather than committed local cert files

## Product impact
- Promise affected: `repo coordination`
- User-visible change: no product behavior change; public repo no longer ships a localhost private key in the forward tree

## Evidence
- `git ls-files certs/localhost.key certs/localhost.pem` showed both files were tracked before this patch
- `openssl x509 -in certs/localhost.pem -noout -subject -issuer -dates -fingerprint -sha256` confirmed a self-signed localhost cert with a long validity window
- repo search showed no runtime path depending on those tracked files directly

## Review evidence
- direct `sb glm-text` review recorded in PR comments
- finding: forward remediation is correct, but prior git history still contains the key material and needs separate handling

## Linked issue
- Relates #6486
- Relates #6453


jeong-sik:
Cross-model review evidence

- Reviewer: direct `sb glm-text` (`GLM-5.1` path)
- Timestamp (UTC): 2026-04-11T06:08:42Z
- Scope: correctness, regressions, and security on the PR diff only
- Result: 1 material finding

Finding summary:
- This patch correctly removes the tracked localhost key/cert and blocks recommit via `.gitignore`.
- Remaining risk: the private key still exists in prior git history and should be treated as compromised.
- Follow-up needed outside this forward-fix PR: rotate/reissue if reused anywhere, and decide whether to rewrite public history (`git filter-repo` / BFG) to excise the key material.

I am keeping this PR as the forward remediation slice and leaving history rewrite as an explicit follow-up because rewriting published history is a higher-risk repository operation.

PR #6593, opened 2026-04-11
[codex] fix keeper retry timeout budget and local-only context

jeong-sik:
## What changed
- bound each keeper `Agent.run` call to the remaining unified-turn wall-clock budget instead of letting the adaptive OAS timeout overrun retry space
- route synthetic "remaining budget exhausted" timeouts through the same post-commit classification path so reconcile handling still triggers after committed mutating tools
- cap pure-local cascades (`ollama`, `llama`, other local providers) to the small-local context floor before keeper compaction / status / checkpoint loading decisions
- keep the same effective local-only context cap across unified turns, manual turns, keepalive snapshots, world observation, keeper status, and keeper creation
- add unit coverage for bounded OAS timeout behavior and pure-local context detection/clamping

## Why
The current keeper logs show two related problems:
- an adaptive OAS timeout around 963s can consume most of the 1200s keeper turn budget, so later retry attempts get cancelled by the outer turn timeout and emit misleading cancellation noise
- local-only cascades still inherit the full discovered max context (for example 262144), which suppresses compaction and lets oversized Ollama requests hit server-side parse rejection paths

## Impact
- retries now use the remaining turn budget deterministically, so the inner OAS call stops before the outer keeper timeout should fire
- if a turn already committed a mutating tool and then runs out of remaining budget, the timeout still goes through the reconcile/integrity classifier
- pure-local keepers should compact earlier and stop shipping oversized request bodies to Ollama-class providers
- keeper status / observation surfaces now report the same capped context that execution actually uses

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/startup-log-audit _build/default/test/test_keeper_unified.exe _build/default/test/test_work_as_heartbeat.exe _build/default/test/test_keeper_pr_workflow.exe _build/default/test/test_start_masc_mcp_script.exe _build/default/test/test_server_runtime_bootstrap.exe _build/default/test/test_server_base_path_diagnostics.exe _build/default/test/test_typed_tool_masc.exe`

## Review evidence
- direct `sb glm-text` / `GLM-5.1` review found one actionable issue in the first pass: synthetic exhausted-budget timeout bypassed post-commit classification; patched in this branch
- follow-up direct GLM spot checks hit provider request timeouts after the patch, so the successful first-pass review plus the patch is recorded in PR comments

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -279,6 +279,7 @@ let run_turn
       ?guardrails
       ?temperature
       ?max_tokens
+      ?oas_timeout_s
       ?max_cost_usd
       ?on_event
       ?(trajectory_acc : Trajectory.accumulator option)
@@ -1430,7 +1431,12 @@ let run_turn
   with
   | Error e -> Error (Oas.Error.Internal e)
   | Ok oas_allowed_paths ->
-    let timeout_s = Env_config_keeper.KeeperKeepalive.oas_timeout_for_context ~max_context in
+    let timeout_s =
+      match oas_timeout_s with
+      | Some value -> value
+      | None ->
+          Env_config_keeper.KeeperKeepalive.oas_timeout_for_context ~max_context
+    in
     (match
        Keeper_llm_bridge.run_with_timeout_and_fallback ~timeout_s (fun () ->
          Oas_worker.run_named
--- a/lib/keeper/keeper_agent_run.mli
+++ b/lib/keeper/keeper_agent_run.mli
@@ -131,6 +131,7 @@ val run_turn :
   -> ?guardrails:Agent_sdk.Guardrails.t
   -> ?temperature:float
   -> ?max_tokens:int
+  -> ?oas_timeout_s:float
   -> ?max_cost_usd:float
   -> ?on_event:(Oas.Types.sse_event -> unit)
   -> ?trajectory_acc:Trajectory.accumulator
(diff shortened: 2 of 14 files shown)


jeong-sik:
Cross-model review evidence for head `d4b81d8e8`.

- Reviewer: direct `sb glm-text` (`GLM-5.1`)
- Timestamp: 2026-04-11 23:44-23:46 KST
- Scope: keeper retry timeout budgeting + pure-local context cap changes

Actionable finding from the successful review pass:
1. Synthetic timeout from exhausted remaining turn budget could bypass the normal post-commit failure classification path in `keeper_unified_turn` when a previous attempt had already committed a mutating tool. Patched by routing that synthetic timeout through the same retry/error classification branch as normal `do_run` failures.

Additional patch made from review feedback:
- replaced the hardcoded adaptive-timeout test range with a config-derived expectation

Verification after patch:
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/startup-log-audit _build/default/test/test_keeper_unified.exe _build/default/test/test_work_as_heartbeat.exe _build/default/test/test_keeper_pr_workflow.exe _build/default/test/test_start_masc_mcp_script.exe _build/default/test/test_server_runtime_bootstrap.exe _build/default/test/test_server_base_path_diagnostics.exe _build/default/test/test_typed_tool_masc.exe`

Follow-up direct GLM spot checks after the patch hit provider-side request timeouts, so this comment records the successful review pass plus the patched finding as the review gate evidence.

PR #6618, opened 2026-04-11
feat: bridge Agent_sdk.Log → masc-mcp structured log + bump OAS pin to 0.121.0

jeong-sik:
## Summary

Forward OAS \`Agent_sdk.Log\` records into the masc-mcp log ring so the per-turn timing line from oas#816 (and any future Agent SDK structured emits) land in \`~/.masc/logs/system_log_*.jsonl\` alongside masc-mcp's own Keeper / Server / Dashboard records.

Also bumps the OAS pin from v0.120.0 (2fc2be2) → v0.121.0 (7b70f43) so the build actually picks up:
- oas#814 — \`base_url=\` field + 5xx body dump gate
- oas#816 — per-turn timing via \`Log.info\`
- oas#810 — \`schema_gen\` field_error refactor (breaking change, companion fix included)

## Why

Before this patch, OAS's global sink registry (\`agent_sdk/lib/log.ml:111\`) started empty on every masc-mcp boot. Every \`Log.info\` / \`Log.warn\` inside \`agent_sdk\` was silently dropped — the observability PRs #814 and #816 on the OAS side were wired in but had no reachable sink on the masc-mcp side. Running \`grep '"module":"oas:' ~/me/.masc/logs/system_log_*.jsonl\` returns zero matches pre-bridge.

## Change

1. **\`lib/oas_log_bridge.{ml,mli}\`** (new, ~70 lines)
   - \`make_sink ()\` — pure forwarder \`Agent_sdk.Log.record → unit\`:
     - level translated 1:1
     - \`module_name\` prefixed with \`"oas:"\` to keep OAS records distinct from masc-mcp's own module names (so \`Agent_sdk.Log.create ~module_name:"agent" ()\` lands as \`"oas:agent"\`)
     - \`fields\` assembled into a Yojson \`details\` object preserving every \`S/I/F/B/J\` variant as a first-class JSONL key
   - \`install ()\` — calls \`Agent_sdk.Log.add_sink\` once. Idempotent only in the sense that repeated calls append duplicate sinks; bootstrap calls exactly once.

2. **\`lib/server/server_runtime_bootstrap.ml\`** — call \`Oas_log_bridge.install ()\` right after \`Llm_metric_bridge.install ()\`. Same placement rationale: before any fiber can fire an LLM call.

3. **\`lib/dune\`** — add \`oas_log_bridge\` to the explicit module list.

4. **\`scripts/oas-agent-sdk-pin.sh\`** — bump base/min version to 0.121.0 and SHA to 7b70f43.

5. **\`lib/tool_broadcast_typed.ml\`** — companion fix for oas#810. Pin bump pulls in schema_gen's new error type; \`Sg.parse broadcast_schema\` now returns \`field_error list\` instead of \`string\`. \`Typed_tool_masc.create\` still takes \`string\`-error parsers, so wrap the call in a helper that calls \`Agent_sdk.Tool_input_validation.format_errors\` to collapse structured errors back to a string. Pure error-shape adapter, no behaviour change on the happy path.

6. **\`docs/KEEPER-USER-MANUAL.md\`, \`docs/OAS-UTILIZATION-AUDIT.md\`** — pin metadata blocks regenerated by \`scripts/sync-oas-pin-docs.sh\`.

## Verification

- \`dune build bin/main_eio.exe\` green.
- \`opam pin add agent_sdk\` to the new SHA + \`dune build\` green (catches any additional drift).

Post-deploy smoke test:

\`\`\`bash
# 1. Restart masc-mcp with this build
./start-masc-mcp.sh

# 2. Wait ~30s for first keeper unified turn

# 3. Verify per-turn timing records
grep '"module":"oas:agent"' ~/me/.masc/logs/system_log_*.jsonl | head
# Should show records with fields.turn / .max_turns / .elapsed_run_sec /
# .turn_duration_sec / .model / .stop
\`\`\`

Not covered by this bridge (follow-up):

- \`agent_sdk/lib/llm_provider/complete.ml\` still uses \`Printf.eprintf\` for its HTTP WARN lines. \`llm_provider\` is a separate sub-library from \`agent_sdk\` and doesn't import \`Log\` directly; converting it is a larger OAS PR. The \`base_url\` + \`5xx dump\` infrastructure from oas#814 is already present; it just doesn't flow through this bridge yet.

## References

- oas#814 — base_url + 5xx body dump (merged)
- oas#816 — per-turn timing via Log.info (merged)
- oas#810 — schema_gen field_error refactor (merged; companion fix in this PR)
- Llm_metric_bridge — identical installation pattern reused here


jeong-sik:

> 1. **Double-install risk — Flag.** Every install() call appends a new sink. If bootstrap runs twice, every OAS record fires N times. Add a ref or Atomic guard.
> 2. Level filter — Fine. 1:1 translation is correct.
> 3. Module name ':' prefix — Fine. Safe in JSON strings.
> 4. Field conversion — Good. Exhaustive match, no wildcard.
> 5. details=None vs Some(Assoc[]) — Fine. None omits the field.
> 6. tool_broadcast_typed error shape — Fine. format_errors wrap is acceptable.
> 7. **Other Sg.parse consumers — Verify.** Only one file modified; grep to confirm.
> 8. Dune ordering — Fine.

Both items addressed in commit 1851834:

**1. Idempotency guard**: Added \`Atomic.compare_and_set\` latch. First call CAS-es false→true and registers the sink; subsequent calls return cleanly. Cheaper than ref+guard under fiber concurrency.

\`\`\`ocaml
let installed = Atomic.make false

let install () : unit =
  if Atomic.compare_and_set installed false true then
    Agent_sdk.Log.add_sink (make_sink ())
\`\`\`

**2. Sg.parse grep verified**: \`rg -n 'Sg\\.parse\\b' lib\` returns only \`lib/tool_broadcast_typed.ml\` (the file this PR already modifies). No other consumers need migration.

Build + dune build bin/main_eio.exe green on 1851834.

jeong-sik:
#6601 머지 완료 (main에 c710d1f05로 squash, OAS pin을 v0.121.0 @ a9e189f로 bump 완료). 이 PR의 OAS pin bump 파트는 superseded. rebase 시 conflict는 pin 관련 4 파일(`scripts/oas-agent-sdk-pin.sh`, `dune-project`, `masc_mcp.opam`, `docs/KEEPER-USER-MANUAL.md`, `docs/OAS-UTILIZATION-AUDIT.md`)에서 발생할 예정. 이 파트를 drop하고 log bridge 부분만 남기면 cleanly rebase 됩니다.

PR #6635, opened 2026-04-11
fix(cp_unit): bound descendant_ids recursion with _max_tree_depth guard (fixes #6633)

### LINKED ISSUE #6633: cp-summary warm cache Stack overflow after 990s (post-restart, blocks dashboard refresh)
## Symptom

After masc-mcp server restart at 2026-04-11T18:27:20Z (version 0.5.5, chore/bump-oas-pin-782c287 binary), the first `cp-summary` warm cache attempt failed with OCaml Stack overflow after ~990 seconds.

Log line:
```
seq=5101 ts=2026-04-11T18:44:20Z level=WARN module=Dashboard
message="cp-summary warm cache failed (990.4s): Stack overflow"
```

## Context

Dashboard `cp-summary` is populated by `Proactive_refresh.start` loop in `lib/server/server_command_plane_http_support.ml:109`. The `compute_cp_summary` function at `:84-107` composes `Command_plane_v2.snapshot_json` + `Command_plane_v2.summary_json` + `Swarm_status.build_json_from_snapshot`. Any of those three can recurse into large structures.

The refresh loop log sequence before the failure shows **10+ rapid `starting cp-summary refresh loop` entries between 18:04 and 18:27** — suggesting the server was being restarted repeatedly in that window. On the final successful boot at 18:27:20, the first warm cache attempt ran for the full 990s (above `MASC_CP_SUMMARY_TIMEOUT_S` default 90.0, so this is NOT a timeout — it's a true stack overflow raised by Proactive_refresh's protected compute).

## Impact

- Dashboard `/command-plane/summary` returns the stale cached ref (`_cp_summary_ref`) which is empty on fresh boot.
- Every subsequent keeper turn that relies on command-plane snapshot data may be starved of context.
- The running server (PID 29787 as observed) has `/health` responding slowly (~15s in one curl test, then fine in another), consistent with GC/thread contention from the infinite recursion.

## Likely sites

`Command_plane_v2.snapshot_json` or `Swarm_status.build_json_from_snapshot` — candidate for recursive JSON serialization on a cyclic graph or a newly-added field with an unbounded traversal.

Recent mergeds that touch command_plane or swarm_status:
- #6612 (channel_gate dedup_cleanup)
- #6627 (auth cross-agent gating)
- #6556 (keeper FSM + hardcoded constants + TLA+ bug model)

## Reproducer

1. Stop masc-mcp, `dune clean`, `dune build`
2. Start with default env; wait ~15 minutes for the warm cache attempt to complete
3. `grep "cp-summary warm cache failed" ~/me/.masc/logs/system_log_$(date +%Y-%m-%d).jsonl`

## Severity

Medium — dashboard warm cache fails but `/health` is still responsive eventually; keeper turns appear stalled but main error (`closing '}' symbol`) is 0 since oas#813 landed. Worth fixing before it blocks observability features.

## Discovered during

Error reduction loop session (#6474 → #6601 chain). Unrelated to oas keep_alive issue but surfaced during the post-restart validation window.

jeong-sik:
## Summary

Closes #6633. \`cp-summary warm cache\` was failing with OCaml \`Stack_overflow\` after ~990s of CPU spin, blocking dashboard refresh and every keeper cascade — the observability chain from oas#821 (keep_alive int fix) and masc-mcp#6618 (Agent_sdk.Log bridge) cannot produce \`oas:*\` JSONL records until this unblocks.

## Root cause

\`descendant_ids\` at \`lib/command_plane/cp_unit.ml:336\`:

\`\`\`ocaml
let rec descendant_ids ?(visited = []) child_map unit_id =
  if List.mem unit_id visited then []
  else
  let visited = unit_id :: visited in
  let children = List.assoc_opt unit_id child_map |> Option.value ~default:[] in
  let direct = List.map (fun (unit : unit_record) -> unit.unit_id) children in
  direct @ List.concat_map (fun child_id -> descendant_ids ~visited child_map child_id) direct
\`\`\`

- Visited-list cycle check catches true cycles
- **No depth guard** (unlike \`build_tree_json\` / \`build_tree_json_indexed\` which both respect \`_max_tree_depth = 50\`)
- Non-tail-recursive body: both \`@\` and \`List.concat_map\` hold stack frames per level
- \`List.mem unit_id visited\` is O(n²) cumulative

On a deep acyclic unit chain, the recursion descends to full subtree height, burning one frame per level per child. At masc-mcp's current unit count this exceeds OCaml's default stack and raises \`Stack_overflow\` inside \`Proactive_refresh\`'s protected compute. \`sample(1)\` on the running process confirmed:

\`\`\`
1200/1433 samples in camlMasc_mcp__Cp_unit$fun_3107 → caml_call_gc
  → caml_empty_minor_heaps_once → caml_scan_stack/oldify_one
\`\`\`

(84% of samples in GC under the \`cp_unit\` recursive call, indicating unbounded allocation before the final \`Stack_overflow\`.)

## Fix

Minimal: add \`?(depth = 0)\` threaded through the recursive call, check \`depth > _max_tree_depth\` alongside the existing cycle check.

\`\`\`ocaml
let rec descendant_ids ?(depth = 0) ?(visited = []) child_map unit_id =
  if depth > _max_tree_depth || List.mem unit_id visited then []
  else
  ...
  direct
  @ List.concat_map
      (fun child_id ->
        descendant_ids ~depth:(depth + 1) ~visited child_map child_id)
      direct
\`\`\`

Same ceiling \`build_tree_json\` uses, so the two functions have consistent traversal bounds. Partial descendant list is returned when the guard fires, matching the existing cycle-truncation branch's "best-effort" behaviour.

## Out of scope (follow-ups to track separately)

- \`List.concat_map\` + \`@\` are still non-tail. A tree with width × 50 nodes in a single level still allocates one frame per element. Prefer an explicit work-queue / \`Hashtbl\`-based implementation for a true fix.
- \`List.mem\` on \`visited\` is O(n²). Swap for \`Hashtbl\` visited set if the unit count grows further.
- \`build_tree_json\` uses \`List.filter_map\` which is also non-tail in OCaml stdlib. Same structural issue, less likely to trigger because \`_max_tree_depth\` already bounds it.

## Verification

- [x] \`dune build bin/main_eio.exe\` green
- [x] \`rg 'descendant_ids' lib/\` — only one call site in \`build_tree_json\` (line 364), new \`?depth\` is backward-compatible
- [ ] Post-deploy smoke test: watch \`~/.masc/logs/system_log_*.jsonl\` for first \`cp-summary warm cache done\` entry and confirm no \`Stack overflow\` for 30+ minutes

## References

- #6633 — tracking issue with sample stack trace + source-level analysis
- oas#821, masc-mcp#6618 — observability chain currently blocked by this

### DIFF
--- a/lib/command_plane/cp_unit.ml
+++ b/lib/command_plane/cp_unit.ml
@@ -333,13 +333,24 @@ let children_map units =
 
 let _max_tree_depth = 50
 
-let rec descendant_ids ?(visited = []) child_map unit_id =
-  if List.mem unit_id visited then []
+(* [descendant_ids] was the Stack_overflow culprit for #6633: the visited-list
+   cycle check catches true cycles but provides no protection against deep
+   non-cyclic chains, and [@]/[List.concat_map] are non-tail so every level
+   holds a stack frame.  Bound the recursion with [_max_tree_depth] (same
+   ceiling used by [build_tree_json]) so a deep unit chain cannot blow the
+   stack.  Returns the partial descendant list when the guard fires, matching
+   the cycle-truncation branch's "best-effort" behaviour. *)
+let rec descendant_ids ?(depth = 0) ?(visited = []) child_map unit_id =
+  if depth > _max_tree_depth || List.mem unit_id visited then []
   else
   let visited = unit_id :: visited in
   let children = List.assoc_opt unit_id child_map |> Option.value ~default:[] in
   let direct = List.map (fun (unit : unit_record) -> unit.unit_id) children in
-  direct @ List.concat_map (fun child_id -> descendant_ids ~visited child_map child_id) direct
+  direct
+  @ List.concat_map
+      (fun child_id ->
+        descendant_ids ~depth:(depth + 1) ~visited child_map child_id)
+      direct
 
 let rec build_tree_json ?(depth = 0) ?(visited = [])
     ~child_map ~unit_lookup ~agent_statuses ~live_agents ~operations unit_id =
--- a/test/dune
+++ b/test/dune
@@ -641,6 +641,11 @@
           test_command_plane_v2_search)
  (libraries masc_test_deps))
 
+; Cp_unit.descendant_ids depth guard regression (#6633)
+(test
+ (name test_cp_unit_descendant_ids)
+ (libraries masc_test_deps))
+
 ; Operator Control (7 modules)
 (test
  (name test_operator_control)
--- a/test/test_cp_unit_descendant_ids.ml
+++ b/test/test_cp_unit_descendant_ids.ml
@@ -1,0 +1,127 @@
+(** Regression tests for [Cp_unit.descendant_ids] depth guard (#6633).
+
+    Before the fix: a linear [child_map] chain of 60+ units would recurse
+    to the full subtree height, and because the recursive call site is
+    non-tail ([direct @ List.concat_map ...]), each level held a stack
+    frame until OCaml raised [Stack_overflow].  The failure surfaced
+    ~990 seconds after boot (the GC+allocation spin before the overflow
+    dominated).
+
+    After the fix: the new [_max_tree_depth = 50] ceiling matches the
+    existing guard in [Cp_unit.build_tree_json], so the chain is
+    traversed at most 51 levels (depth 0..50) and returns a truncated
+    descendant list without raising. *)
+
+open Masc_mcp.Cp_types
+module Cp_unit = Masc_mcp.Cp_unit
+
+(** Minimal [unit_record] factory. Only the fields actually consulted by
+    [descendant_ids] and [children_map] matter ([unit_id],
+    [parent_unit_id]); the rest are placeholders and not asserted on. *)
+let make_unit ~unit_id ?parent_unit_id () : unit_record =
+  {
+    unit_id;
+    label = unit_id;
+    kind = Squad;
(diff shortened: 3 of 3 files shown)


jeong-sik:

> **Approved with one action item.**
>
> 1. Silent truncation — acceptable ✓
> 2. Off-by-one — correct ✓ (matches build_tree_json line 346)
> 3. Backward compat — verified ✓ (optional params + lone call site)
> 4. Stopgap appropriateness ✓
> 5. **Tests — missing, please add one**
> 6. Style — matches ✓
>
> Add the test and this ships.

Addressed in commit 7039b74fd: new \`test/test_cp_unit_descendant_ids.ml\` with three cases:

1. Linear chain of 60 units truncates below full depth without \`Stack_overflow\` (the exact #6633 failure mode).
2. Linear chain of 10 returns all 9 descendants (guards against the depth guard firing too early).
3. Two-unit cycle still caught by the existing \`List.mem\` visited check (regression protection).

Registered as a new standalone \`(test ...)\` entry in \`test/dune\` next to \`test_command_plane_v2\`. \`dune exec test/test_cp_unit_descendant_ids.exe\` reports \"Test Successful 3 tests run\" in 0.000s.

Build: \`dune build bin/main_eio.exe\` green on both commits (2cb06bacf fix, 7039b74fd test).

PR #6555, opened 2026-04-11
refactor: remove redundant sanitize calls + OAS pin 0.121.0

jeong-sik:
## Summary

- OAS v0.121.0 pin (oas#808): pipeline 레벨 UTF-8 sanitization 보장
- keeper LLM 입력 경로에서 `Inference_utils.sanitize_text_utf8` 5건 제거
- 파일시스템/메트릭용 8건은 유지 (OAS pipeline 범위 밖)

## Motivation

masc-mcp#5626 L5 (sanitize_text_utf8 경계 위반) 해소.
OAS가 user_prompt, system_prompt, tool result를 pipeline에서 sanitize하므로
MASC가 Agent.run 입력 전에 중복 sanitize할 필요 없음.

## Changes

| 파일 | 변경 |
|------|------|
| `keeper_unified_prompt.ml` | system prompt/user message 빌더에서 sanitize 제거 (-2) |
| `keeper_agent_run.ml` | run_turn 진입점에서 sanitize 제거 (-3) |
| `dune-project` + `masc_mcp.opam` | agent_sdk >= 0.121.0 |
| `scripts/oas-agent-sdk-pin.sh` | SHA 업데이트 |

## Test plan

- [x] `dune build` 성공 (OAS 0.121.0 pin)
- [x] `test_auth_coverage` 88 tests pass
- [ ] CI Build and Test
- [ ] keeper 20+ 턴 정상 동작

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6631, opened 2026-04-11
chore(oas): bump agent_sdk pin to 0.122.0 (keep_alive integer fix)

jeong-sik:
## Summary
Bump OAS agent_sdk pin from 0.121.0 → 0.122.0 to pick up [jeong-sik/oas#821](https://github.com/jeong-sik/oas/pull/821) which fixes the Ollama keep_alive wire format.

## Runtime evidence
After bootstrapping keepers on the new binary earlier today, all 6 GLM-cascade keepers failed within <2s per turn:

\`\`\`
ts=2026-04-11T17:59:56Z out=error lat=1275 sem=0 tools=0 \\
  err=Invalid request: time: missing unit in duration \"-1\"
ts=2026-04-11T18:01:07Z out=error lat=576  sem=0 tools=0 \\
  err=Invalid request: time: missing unit in duration \"-1\"
ts=2026-04-11T18:02:24Z out=error lat=498  sem=0 tools=0 \\
  err=Invalid request: time: missing unit in duration \"-1\"
\`\`\`

Root cause: PR #813 pinned keep_alive=-1 as string \`\` \`String \"-1\" \`\` but Ollama routes string keep_alive through Go's \`time.ParseDuration\` which requires a unit. oas#821 switches to \`\` \`Int (-1) \`\` when parseable and \`\` \`String v \`\` for duration strings ("5m", "-1m", etc).

## Changes
Following the \"OAS pin must bump version floor\" rule:
- \`scripts/oas-agent-sdk-pin.sh\`: SHA + BASE_VERSION + MIN_VERSION updated
- \`dune-project\`: \`(agent_sdk (>= 0.122.0))\`
- \`masc_mcp.opam\`: \`agent_sdk >= 0.122.0\`

## Verification
- Local \`opam install agent_sdk\`: successfully upgraded 0.121.0 → 0.122.0
- Local \`dune clean && dune build\`: clean success
- No API surface changes between 0.121.0 and 0.122.0 affecting masc-mcp

## Post-merge steps (for server recovery)
1. Pull main into the currently running worktree (\`fix/start-script-basepath-config-root\` or main)
2. \`opam install agent_sdk -y\` to sync the pin
3. \`dune build\` to produce the new binary
4. SIGTERM the running server and restart via \`start-masc-mcp.sh\`
5. Expected: all 6 keepers stop failing on the keep_alive error and resume normal turns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #5626, opened 2026-04-07
refactor: MASC/OAS boundary cleanup — 7 violation points


jeong-sik:
## Context

Keeper turn cycle 감사에서 MASC가 OAS 내부 함수를 직접 호출하는 7개 경계 위반 지점을 발견했다.
OAS API 확장이 선행되어야 하므로 RFC로 분리.

## 원칙

MASC는 추상 신호(cascade_name, compaction_signal, priority)만 전달.
OAS가 모델/벤더/토큰/가격을 소유.

## 위반 목록

| # | 위반 | MASC 파일 | OAS에 필요한 API |
|---|------|----------|-----------------|
| L1 | temperature/max_tokens fallback 하드코딩 (0.3, 8192) | keeper_agent_run.ml:256-271 | Agent.run이 cascade에서 자체 resolve |
| L2 | context_reducer 전략 하드코딩 (turn 5+ stub) | keeper_agent_run.ml:1005-1019 | OAS가 reducer 전략 소유 |
| L3 | Cascade_config.parse_model_strings + Pricing 호출 | keeper_unified_turn.ml:1019-1052 | result에 canonical_model_id + cost 포함 |
| L4 | resolve_max_cascade_context + min 65536 clamp | keeper_unified_turn.ml:796-810 | OAS가 context_budget 반환 |
| L5 | Inference_utils.sanitize_text_utf8 직접 호출 | keeper_agent_run.ml:363-365 | OAS 내부 전처리 |
| L6 | model ID strip_latest 파싱 | keeper_unified_turn.ml:1019-1045 | result에 opaque metadata |
| L7 | compact_ratio 직접 전달 (수치) | keeper_agent_run.ml:1056 | 추상 signal (Aggressive/Balanced) |

## 제안 방향

1. OAS `Agent.run` result에 `canonical_model_id`, `estimated_cost_usd` 추가
2. OAS에 `resolve_inference_params(cascade_name) -> { temperature, max_tokens, max_context }` API 추가
3. context_reducer를 OAS 내부 전략으로 이동 (MASC는 compact_ratio만 전달)
4. sanitize_text_utf8를 OAS Agent.run 내부로 이동

## 영향

코드 변경은 OAS(oas repo) + MASC(masc-mcp repo) 양쪽 필요.
MASC 단독 변경 불가하므로 OAS 측 API 설계가 선행.

jeong-sik:
## Progress Update (2026-04-11)

### Resolved
[...]

PR #6645, opened 2026-04-11
refactor(keeper): remove redundant UTF-8 sanitize calls on LLM input path

jeong-sik:
## Summary

Extracted from #6555 which is currently CONFLICTING/DIRTY because its OAS pin bump commit (to 0.121.0) is stale — main is already on OAS v0.122.0 (#6601, #6618, #6631). This PR keeps only the sanitize cleanup portion.

Removed 5 calls in the LLM input path:
- \`keeper_unified_prompt.ml\`: system prompt and user message builders
- \`keeper_agent_run.ml\`: run_turn entry point (3 calls)

Retained 8 calls for non-LLM paths: checkpoint I/O (filesystem data integrity) and token metrics computation (local estimation).

Addresses masc-mcp#5626 L5 (sanitize_text_utf8 boundary violation).

## Provenance

Cherry-picked \`b8e0018fa\` from \`origin/refactor/remove-sanitize-calls\` (#6555) onto fresh \`origin/main\`. The OAS pin commit \`867c7bb12\` was dropped because:
- Main already has \`9c9cc20fe chore(oas): bump agent_sdk pin to 0.122.0\` (#6631)
- Main already has \`5c41e5715 feat: bridge Agent_sdk.Log → bump OAS pin to 0.121.0\` (#6618)
- Main already has \`c710d1f05 chore(oas): bump agent_sdk to v0.121.0\` (#6601)

## Test plan

- [x] \`dune build @lib/keeper/all\` clean
- [ ] After merge: close #6555 as superseded by this PR

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -393,13 +393,6 @@ let run_turn
   let { system_prompt = turn_system_prompt; dynamic_context } =
     build_turn_prompt ~base_system_prompt ~messages:ctx_work.messages
   in
-  (* Defense in depth: unified prompt builders sanitize their own output,
-     but run_turn is shared by other callers and is the final boundary before
-     handing prompts/history to OAS. Keep this sanitization here even when
-     upstream builders already cleaned their strings. *)
-  let turn_system_prompt = Inference_utils.sanitize_text_utf8 turn_system_prompt in
-  let dynamic_context = Inference_utils.sanitize_text_utf8 dynamic_context in
-  let user_message = Inference_utils.sanitize_text_utf8 user_message in
   let prompt_metrics =
     build_prompt_metrics ~system_prompt:turn_system_prompt ~dynamic_context
       ~user_message
--- a/lib/keeper/keeper_unified_prompt.ml
+++ b/lib/keeper/keeper_unified_prompt.ml
@@ -269,7 +269,6 @@ let build_prompt ~(meta : Keeper_types.keeper_meta) ~(base_path : string)
   in
   let system_prompt =
     Printf.sprintf "%s\n\n## Turn Intent\n%s" base_system_prompt turn_intent_block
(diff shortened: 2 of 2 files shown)


jeong-sik:

- Scope: Remove `sanitize_text_utf8` from LLM input path (2 files, 1+/9-)
- Timestamp: 2026-04-11

### Verdict: Findings (non-blocking, MEDIUM)

GLM raised 2 concerns about removing the defense-in-depth simultaneously from both `build_prompt` (upstream) and `run_turn` (downstream final boundary):

**1. `keeper_turn.ml` path now has no safety net.**

`run_turn`'s defense-in-depth + `build_prompt`'s sanitize were both removed. If `keeper_turn.ml` passes raw bytes from any source that is not already guaranteed UTF-8 (e.g., reading memory files, prior agent output, gRPC input), the LLM input path has no last-line guard.

Verified caller: `lib/keeper/keeper_turn.ml:308` calls `run_turn ~build_turn_prompt:...` where `build_turn_prompt` takes `messages` directly from `ctx_work.messages` without sanitize.

**2. OAS behavior on invalid UTF-8 is unspecified in this PR.**

If the OAS tokenizer panics on invalid UTF-8 → DoS. If it replaces with U+FFFD → prompt corruption. Removing the final safety net is only justified if OAS has a documented graceful handling contract for invalid bytes.

### Suggested resolution options

1. **Document the contract** in the PR body: where/how are inputs guaranteed UTF-8-safe upstream? What does OAS do on bad bytes?
2. **Keep the last boundary** at `run_turn` (remove only from `build_prompt`) as a tighter minimum-wait-times boundary.
3. **Proceed as-is** if the owner has already verified OAS handles it cleanly and all upstream sources are UTF-8-safe.

`/loop` not auto-merging — decision is owner's. The #5626 L5 boundary-violation rationale is sound in spirit; the question is just where the minimum safe net should live.

PR #6647, opened 2026-04-11
fix(cp_unit): bound descendant_units_of_kind recursion (#6635 follow-up)

jeong-sik:
## Summary

Follow-up to #6635 for the sibling recursive function that PR missed. Same failure template, different callsite. Live masc-mcp running the post-#6635 + pin-bump binary is still hitting \`cp-summary warm cache failed (828.5s, 861.7s): Stack overflow\` in the current session's JSONL log.

## Diagnosis

\`sample(1)\` on the spinning process (PID 62024, 284.8% CPU, 16m uptime post-restart) landed this hot stack:

\`\`\`
1407 Thread main
  1407 caml_main
    513 camlMasc_mcp__Cp_serde\$option_exists_1057  (36%)
      513 caml_call_gc → caml_empty_minor_heaps_once → caml_scan_stack/oldify_one
    439 (other call → same GC)
\`\`\`

\`Cp_serde.option_exists\` is a 3-line pure function, not recursive itself — but its **sole call site** in cp_unit.ml is \`descendant_units_of_kind\` (line 277):

\`\`\`ocaml
let rec descendant_units_of_kind units unit_id kind =
  let direct_children =
    units
    |> List.filter (fun (unit : unit_record) ->
           option_exists (String.equal unit_id) unit.parent_unit_id)
  in
  let direct_matches =
    direct_children
    |> List.filter (fun (unit : unit_record) -> unit.kind = kind)
  in
  direct_matches
  @ List.concat_map
      (fun (child : unit_record) -> descendant_units_of_kind units child.unit_id kind)
      direct_children
\`\`\`

Same failure template #6635 fixed for \`descendant_ids\`:

1. **No depth guard** (unlike \`build_tree_json\`, \`build_tree_json_indexed\`, and now post-#6635 \`descendant_ids\`)
2. **No cycle detection** — weaker than \`descendant_ids\` which at least had \`List.mem unit_id visited\`
3. **Non-tail body**: \`@\` + \`List.concat_map\` hold stack frames per level
4. **O(N) per recursive call**: \`List.filter\` over the full \`units\` list at every level → O(N²) cumulative with the recursion

On a deep unit chain the combined pattern burns unbounded stack + O(N²) allocation, blowing up ~800s later under GC pressure. Same observable symptom as the pre-#6635 bug, different function.

## Change

1. \`lib/command_plane/cp_unit.ml\`

   - Move \`let _max_tree_depth = 50\` up to just after the \`rewritten_units\` block so the constant is visible to the earlier \`descendant_units_of_kind\` definition. Same literal value; all four consumers (\`descendant_units_of_kind\`, \`descendant_ids\`, \`build_tree_json\`, \`build_tree_json_indexed\`) keep using the shared ceiling.
   
   - Thread \`?(depth = 0) ?(visited = [])\` through \`descendant_units_of_kind\`, check \`depth > _max_tree_depth || List.mem unit_id visited\` at the top, bump depth + prepend visited on the recursive call.

2. \`test/test_cp_unit_descendant_ids.ml\`

   Extend with three parallel cases for \`descendant_units_of_kind\` matching the \`descendant_ids\` coverage:

   - chain 60 truncates without overflow
   - chain 10 returns all 9 descendants
   - two-unit cycle terminates

## Out of scope (follow-up)

- \`@\` + \`List.concat_map\` are still non-tail. Both \`descendant_ids\` (from #6635) and \`descendant_units_of_kind\` (from this PR) still allocate one frame per level in the outer list operation. A true fix would switch to a work-queue + \`Hashtbl\` visited set, dropping the O(N²) behaviour of the outer list + \`List.mem\` and making the traversal tail-recursive. The depth guard is a stopgap that stops the bleeding without the structural rewrite.

## Verification

- [x] \`dune build bin/main_eio.exe\` green
- [x] \`dune exec test/test_cp_unit_descendant_ids.exe\` — 6 tests pass (3 existing + 3 new), 0.001s
- [ ] Post-deploy smoke: \`grep 'cp-summary warm cache failed.*Stack overflow' ~/me/.masc/logs/system_log_*.jsonl\` should stop growing after this binary replaces the current one

## References

- #6633 — original cp-summary stack overflow tracking issue
- #6635 — sibling \`descendant_ids\` fix, merged but left this function untouched
- Running PID 62024 sample trace — 513/1407 main-thread samples in \`option_exists\` hot path (36% of CPU)


jeong-sik:

> **Verdict: Ship it.** No blocking issues.
>
> 1. Semantic shift (adding cycle detection) — no issue. Cyclic unit trees are corrupt data; returning [] is strictly better than diverging.
> 2. Backward compat — compiles fine. Optional labeled params don't need labels at call sites since positionals disambiguate.
> 3. Constant move — clean. No dependencies between old and new \`_max_tree_depth\` positions.
> 4. Ceiling of 50 — acceptable. Real unit trees rarely exceed 10 levels; 50 is 5×. File env-tunable as follow-up if anyone reports truncation.
> 5. **Test kind coverage — real gap, not blocking.** make_unit hardcodes Squad so the kind filter is trivially true in all tests. Add a non-Squad mid-chain test as follow-up.
> 6. O(N²) — not the bug, guard is sufficient. Depth-bounded to 50, worst case 50·N comparisons per call = effectively linear. The sample trace CPU was stack unwinding/GC thrashing, not quadratic filter.

Filing item 5 as a follow-up (not part of this PR — the regression tests cover the stack-overflow fix which is the urgent issue). Will add to #6633 or a new issue depending on where maintainers prefer.

PR #6654, opened 2026-04-11
chore(oas-pin): bump agent_sdk pin to f075853

jeong-sik:
## Summary
- Moves OAS runtime pin forward from 782c287 (oas#821) to f075853, bundling 6 patch-level upstream fixes.
- Primary driver: **oas#830** backend_openai capability gate on `min_p`/`top_k`. Without this, the GLM coding endpoint 400s on local-LLM sampling params that cascade through backend_openai, which cascaded into the current keeper `manual_reconcile_required` deadlock observed in /loop (6 keepers, 1371+ skipped ticks).
- agent_sdk dependency floor stays at `0.122.0` — every commit in this range is patch-level. Version metadata in `scripts/oas-agent-sdk-pin.sh` unchanged except for `OAS_AGENT_SDK_SHA`.

## Upstream commits included
| PR | Fix |
|----|-----|
| oas#830 | `backend_openai` gates `min_p`/`top_k` on `capabilities.supports_*` |
| oas#831 | capability drops routed through dedup warn helper (once per model×param) |
| oas#828 | `mcp.truncate_output` CJK-aware + UTF-8 boundary safety |
| oas#829 | extract canonical CJK-aware token estimator to `llm_provider` |
| oas#826 | `cascade_config.effective_max_context` adds `n > 0` guard |
| oas#825 | provider Anthropic `capabilities_for_model` consults `for_model_id` (closes oas#824) |
| oas#823 | builder derives `context_thresholds` fallback from provider capabilities |

## Runtime impact
- Keepers currently sitting in `manual_reconcile_required` (observed via /loop) hit that state because `min_p` → 400 on glm-coding → cascade retry → 1200s wall-clock timeout → `ambiguous_partial_commit` → blocker set. After this rebuild, the `min_p` 400s stop at the OAS layer, so the failure chain doesn't ignite in the first place.
- Existing `manual_reconcile_required` blockers still require operator reconcile clear; this PR is what prevents the next batch.

## Test plan
- [x] `bash scripts/check-oas-pin.sh` clean after pin+docs sync
- [x] `bash scripts/opam-pin-external-deps.sh` then `opam install agent_sdk -y` rebuilds pinned package
- [x] `DUNE_WORKSPACE=./dune-workspace dune build --root . bin/main_eio.exe` succeeds locally
- [ ] CI `Build and Test` green
- [ ] CI `Health` pin-check + doc-sync green
- [ ] After merge: user rebuilds + restarts masc-mcp; verify new `min_p` errors stop appearing in `/tmp/masc-server.log` and keeper error rate for `sampling_unsupported` trends to zero


jeong-sik:
Superseded by #6653 which was merged at 2026-04-11T22:30:33Z with the same SHA (f075853e5556614a6faf3831eb9ee67cf6acd486). Parallel session race condition — closing this duplicate.

PR #6657, opened 2026-04-11
fix(test): sync unified_prompt tests to post-#6644/#6645 contract

jeong-sik:
## Summary
main's Build and Test CI has been red since the #6644/#6645 merges yesterday (run 24291348412 and every PR CI inherited thereafter, including #6655). Two unified_prompt cases fail because the tests were not updated when intentional behavior changes landed:

- **case 5** (`world prompt distinguishes playground and worktree`) asserted a literal substring that #6644 intentionally removed — it was the exact drift teaching server-root `.worktrees/`.
- **case 26** (`prompt sanitizes control chars`) asserted that `Keeper_unified_prompt.build_prompt` output was sanitized, but #6645 removed those sanitize calls because OAS v0.121.0 (oas#808) centralizes UTF-8 sanitization at pipeline boundaries. The check became a tautology on removed code.

## Fixes
- **case 5** now pins the new contract: world prompt must (a) name playground as default sandbox, (b) teach the canonical `.masc/playground/{name}/repos/<REPO>/.worktrees/…` shape, (c) reject bare server-root `.worktrees/`. These directly reflect `config/prompts/keeper.world.md` post-#6644.
- **case 26** removes the `build_prompt`-level sanitize assertions (redundant after #6645), adds a block comment explaining the sanitization contract moved to OAS, and asserts the structural invariant that raw control-char input flows through `build_prompt` verbatim. The sanitizer itself is still covered by `test_sanitize_text_utf8_replaces_control_chars` immediately above.

## Scope
Test-only. No lib/ or prompt changes. No behavior change.

## Why not revert #6644/#6645
Both PRs landed correct product intent:
- #6644 fixed the harness-rejection drift where keepers were being taught bare `.worktrees/` paths and then hitting `write_outside_playground_blocked`.
- #6645 removed masc-mcp sanitize calls that were redundant after OAS v0.121.0 centralized sanitization at the pipeline boundaries. Reverting would re-introduce double-sanitization.

## Test plan
- [x] `dune exec test/test_keeper_unified.exe` → 160/160 pass locally, including the two previously-red cases
- [x] `dune build` green
- [ ] CI Build and Test green (primary blocking check)
- [ ] Other checks green

## Follow-ups (out of scope)
- File a note on oas#808 to confirm the pipeline-layer sanitize contract has an upstream regression test in OAS. If not, the masc-mcp side loses its end-to-end guarantee.

### DIFF
--- a/test/test_keeper_unified.ml
+++ b/test/test_keeper_unified.ml
@@ -652,12 +652,25 @@ let test_capabilities_prompt_distinguishes_playground_and_worktree () =
     (contains_substring prompt "`keeper_pr_submit` is the canonical submit step")
 
 let test_world_prompt_distinguishes_playground_and_worktree () =
+  (* Post-#6644: worktrees now live INSIDE the playground clone instead
+     of being "a separate workflow path". The old literal substring
+     ["a separate workflow path under `.worktrees/<branch-or-task>/`"]
+     was removed in that fix because it was the exact drift teaching
+     server-root `.worktrees/` — the harness was rejecting those paths
+     as [write_outside_playground_blocked]. The test now pins the new
+     contract: the world prompt must still name playground as the
+     default sandbox, and must teach the canonical
+     [.masc/playground/{name}/repos/<REPO>/.worktrees/...] shape for
+     worktrees rather than a bare server-root path. *)
   let prompt = Prompt_registry.get_prompt "keeper.world" in
   check bool "world prompt names playground sandbox" true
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #6661 which now includes the unified_prompt test sync (commit 0a154a804 cherry-picked from this branch). See #6661 for the stacked fix that breaks the circular CI dependency between #6657 and #6661.

PR #6651, opened 2026-04-11
fix(repair_loop): gate working_dir on caller playground (#6641) [WIP iter10]

jeong-sik:
## Summary

Closes #6641. **Iter10** of the #6527 playground containment trilogy — targets the repair loop dispatcher family (`masc_repair_loop_start`, `masc_keeper_repair`) that iter1~iter8 missed. Same fix shape as iter6 #6610 (`validate_writable_path ~agent_name`) applied to a new dispatcher family.

**⚠️ WIP — tests pending.** Base fix and local build are green; tests for the new gate will follow in the same PR before Ready transition.

Discovered by `/loop` tick 17 sweep `rg 'get_string args "cwd"|get_string args "working_dir"' lib/` — the same methodology that found iter9 #6644 in tick 18 and #6637 in tick 16.

## Product impact

`Tool_repair_loop.handle_start` and `Tool_keeper.handle_keeper_repair` both defaulted `working_dir` to `Sys.getcwd ()` and validated against `cwd_root = Sys.getcwd ()`. Any subdirectory of the server repo root passes — including foreign playground clones. Keeper-A could call:

```json
{
  "tool": "masc_repair_loop_start",
  "args": {
    "task_spec": "…anything…",
    "working_dir": ".masc/playground/keeper-B/repos/masc-mcp",
    "target_mode": "repo",
    "target_file": "lib/foo.ml"
  }
}
```

The repair loop would then run its LLM generator + validator on keeper-B's workspace: writing files into keeper-B's playground, building under keeper-B's dune root. Cross-keeper containment violation with the same shape as #6527 iter1 (worktree server-root fallback), iter3 (shell write gate), iter6 (tool_code_write per-agent).

**Importance**: `masc_keeper_repair` is actively used by the repair loop subsystem as a delegated tool. Containment leak here translates directly to cross-keeper code pollution — keeper-A's LLM generations end up in keeper-B's workspace and may be silently committed/pushed as keeper-B's work.

## Evidence

Shared helper `Tool_repair_loop.resolve_playground_working_dir` takes three flat fields (`~agent_name`, `~base_path`, `~working_dir_arg`) so both `Tool_repair_loop_types.context` and `Keeper_types.context` callers can share it without type unification. The two context types have different shapes (`sw/clock` option vs required, `net` only in keeper), so row polymorphism would have been awkward — flat field arguments are the simplest lowest-coupling design.

```ocaml
let resolve_playground_working_dir
    ~agent_name ~base_path ~working_dir_arg =
[...]

PR #6648, opened 2026-04-11
docs(keepers): fix prompt-layer drift teaching server-root .worktrees/ (#6644)

jeong-sik:
## Summary

Closes #6644. Prompt side of the #6527 playground containment trilogy. `keeper.world.md:13`, `masc-improver.toml:41`, and `coder/AGENT.md` (9 lines) all still taught keepers to use server-root `.worktrees/<branch>/` paths even though iter3/iter4/iter6 harness gates now reject exactly those paths. Iter1 PR #6533 updated `keeper.capabilities.md` but the sweep stopped there. This PR finishes the sweep and adds a CI drift gate so the same drift can't recur.

Discovery: `/loop` tick 18 audit (`rg '\.worktrees/' config/`) surfaced 12 hits — 11 drift, 1 correct (iter1).

## Product impact

`keeper.world.md` is loaded into every keeper's system prompt `<world>` block. Every MASC keeper was therefore being instructed, at the prompt level, to use a path that its own tools would reject:

| Harness gate | Rejection reason |
|---|---|
| iter3 #6579 `keeper_exec_shell` | `write_outside_playground_blocked` |
| iter4 #6580 `effective_allowed_paths` | `.worktrees/` removed |
| iter6 #6610 `tool_code_write.validate_writable_path` | per-agent rejection |

This drift is a plausible cause of why the #6527 trilogy kept surfacing "new" dispatcher families across iter1~iter7 — the prompt kept re-teaching a false path, and each successive keeper run funneled that false path through a different dispatcher, making each rejection look like a new bug rather than a repeat of the same mis-education.

`masc-improver` is an active autonomous keeper running in production; it was most directly affected. `coder/AGENT.md` is dormant (no keeper currently maps to that persona) but would silently break on enable.

No functional OCaml change, no runtime surface change, no behaviour change for a correctly-pathed keeper. The only keepers this impacts are those who were previously following the drifted prompt — they now see a prompt that matches the harness.

## Evidence

Manual verification:

```
$ bash scripts/validate-prompt-paths.sh
✓ validate-prompt-paths: all .worktrees/ references are playground-rooted

$ bash scripts/check-toml-syntax.sh
Checking TOML syntax...
All 5 TOML files parsed successfully.
```

[...]

Issue #6625, opened 2026-04-11
main CI blocker: Build and Test failing across base-path scoped suites after #6602


jeong-sik:
## Symptom
- `main` CI run `24287028090` failed in `Build and Test` (job `70918240000`).
- Open PR merges are currently blocked because every open PR has either `Build and Test` failing, additional failing checks, pending checks, draft status, or merge conflicts.
- Failures are not isolated to one suite; the failed main run includes `room_state`, `dashboard_perf`, `read_model`, `activity_feed_fs`, and `operator` tests.

## Reproduction
1. Inspect `main` CI run `24287028090` in `jeong-sik/masc-mcp`.
2. Inspect any open PR via `gh pr checks <PR> --repo jeong-sik/masc-mcp`.
3. Observe that merge candidates are blocked by CI and/or state gates.

## Suspected Cause
- Base-path scoping / cache isolation changes appear incomplete or inconsistent after `#6602`.
- At least some failures still show cross-base-path leakage symptoms, for example `operator` test 27 (`keeper down only pauses current base path`).

## Tried
- Checked open PR inventory with `gh pr list`.
- Verified non-draft candidates `#6624`, `#6618`, and `#6552`.
- Confirmed `#6618` has failing `Build and Test` and `PR Hygiene`; `#6624` has pending and failing checks; `#6552` is conflicting and failing.
- Confirmed `main` itself is red, so this is not just a per-PR issue.

## Impact
- Cannot start sequential PR merges safely under current merge policy (`CI/CD must be green before merge`).

## Evidence
- main CI run: https://github.com/jeong-sik/masc-mcp/actions/runs/24287028090
- Example blocked PRs: `#6624`, `#6618`, `#6552`

jeong-sik:
Resolved by PR #6699 (combined fix for base-path scoped suite failures). The original symptom — base-path scoped suites failing after #6602 — no longer appears in latest CI runs. Remaining CI failures tracked by separate issues.

PR #6659, opened 2026-04-11
test(keeper_unified): unblock unified_prompt suite (#6656 test 26 + iter9 test 5)

### LINKED ISSUE #6656: regression: build_prompt no longer sanitizes control chars (from #6645)
## Summary

`test_keeper_unified.unified_prompt[26] "prompt sanitizes control chars"` has been failing on main since PR #6645 (`refactor(keeper): remove redundant UTF-8 sanitize calls on LLM input path`). PR #6645 removed the sanitize wrapping from `Keeper_unified_prompt.build_prompt`, but the existing test expected `build_prompt`'s output to be scrubbed of disallowed control chars when the input `meta.instructions` contained `\000` bytes.

The regression is currently invisible because main has been red on `Build and Test` with #6625 (`MASC_BASE_PATH` base-path scoped suites) since the same general window, and new PRs have been force-merging through the blocker.

## Symptom

```
FAIL unified_prompt[26] "prompt sanitizes control chars"
File "test/test_keeper_unified.ml", line 1769, character 2
    Expected: false
    Received: true   # i.e., contains_disallowed_control_char sys = true
```

`test_prompt_sanitizes_control_chars` (line 1751) deliberately sets:

```ocaml
meta = { minimal_meta with instructions = "watch\000this" }
```

and then asserts `build_prompt`'s returned `sys` string does NOT contain a disallowed control char (< 0x20 other than `\t\n\r`, or `0x7f`). Before #6645, `build_prompt` either called `sanitize_text_utf8` internally or the call path guaranteed sanitized input by construction. After #6645, the `\000` passes through verbatim and the substring check fires.

## Root cause

`lib/keeper/keeper_unified_prompt.ml` — PR #6645 commit `1389b0019`:

```
1389b00198 refactor(keeper): remove redundant UTF-8 sanitize calls on LLM input path (#6645)
```

Git log on the file:

```
$ git log --oneline -10 -- lib/keeper/keeper_unified_prompt.ml
1389b0019 refactor(keeper): remove redundant UTF-8 sanitize calls on LLM input path (#6645)
055c6e9ab fix(dashboard): improve accessibility — buttons and form labels (#6496)
```

The refactor evidently assumed sanitization would happen at a later pipeline stage (`Keeper_agent_run` uses `Inference_utils.sanitize_messages_utf8` on the message history — see `lib/keeper/keeper_agent_run.ml:409`), but `build_prompt`'s direct callers in tests exercise the function in isolation and expect the guarantee at this boundary.

## Options

1. **Restore sanitize at `build_prompt` exit.** Simplest; preserves the test contract. But undoes #6645's stated goal of removing "redundant" calls.
2. **Update `test_prompt_sanitizes_control_chars` to call `Inference_utils.sanitize_text_utf8` on `sys`/`user` before the control-char check.** Accepts the new contract: callers are responsible for sanitization post-build.
3. **Add a narrow sanitize inside `build_prompt` scoped only to user-controlled fields (`meta.instructions`, `observation.pending_mentions[*].content`, board previews).** This is what #6645 intended to avoid, but it's the minimum needed to preserve the "user-provided bytes must not reach the prompt" invariant.

Recommendation: **option 3**, because the contract "user-supplied bytes are sanitized at the prompt boundary" is load-bearing for prompt injection defense — it's not just a hygiene nit. A downstream `Inference_utils.sanitize_messages_utf8` in `Keeper_agent_run` is too late if any other caller of `build_prompt` bypasses that sanitize step.

## Scope

This is independent of #6641 iter10 (repair loop containment) and #6527 trilogy. It regresses a prompt-safety invariant. Filing as a separate issue so the fix can have its own cross-model review and won't block iter10's repair loop gate.

## Related

- PR #6645 (`1389b0019`) — source of the regression
- Test location: `test/test_keeper_unified.ml:1751`
- Sanitize utility: `lib/inference_utils.ml:55-92` (`is_disallowed_control_char`, `sanitize_text_utf8`)
- Downstream sanitize call site: `lib/keeper/keeper_agent_run.ml:409`
- Main CI blocker that hid this regression: #6625
- Adjacent iter10 PR: #6651

## Discovery

Found while investigating `Build and Test` failures on PR #6651 (iter10 repair loop). Failing test 5 was a stale text assertion from my iter9 #6648 (fixed in the same iter10 PR). Failing test 26 turned out to be unrelated — git blame on `keeper_unified_prompt.ml` pointed at #6645 as the only recent change, and the test assertion is about sanitize behavior, not prompt text.

jeong-sik:
### Scope update — now unblocks both stale unified_prompt assertions

Original scope was test 26 only (`prompt sanitizes control chars`, #6656 #6645 regression). Adding test 5 (`world prompt distinguishes playground and worktree`, iter9 #6648 regression) in the same PR because:

- Both are **stale-on-main test assertions** that slipped through under the #6625 `Build and Test` blocker.
- Both exercise `test_keeper_unified.ml` → `Keeper_unified_prompt.build_prompt`, so the fixes share scope.
- Landing them separately would leave either #6656 or #6648 half-fixed on main until the second PR merges, and any new PR's CI would show a residual `unified_prompt` failure.

Combining them into a single minimal test-only PR is the cleanest way to get a **clean unified_prompt test suite on main** and unblock every downstream PR's `Build and Test` (modulo #6625 base-path issue, which is a separate blocker).

Note: PR #6651 (iter10 repair loop containment) has its own copy of the test 5 fix as commit `02728acb8`. Whichever PR lands first, the other will rebase cleanly — identical edits to line 654 are easy to resolve. I'll drop the duplicated commit from #6651 on the rebase after merge.

---

## Summary

Closes #6656. Test-only fix aligning `test_prompt_sanitizes_control_chars` with the new sanitization contract established by PR #6645, and aligning `test_world_prompt_distinguishes_playground_and_worktree` with the new canonical phrase introduced by iter9 #6648.

PR #6645 (`refactor(keeper): remove redundant UTF-8 sanitize calls on LLM input path`) intentionally moved UTF-8 sanitization from MASC's `Keeper_unified_prompt.build_prompt` to the **OAS pipeline boundary** (`agent.ml`, `pipeline.ml`, `agent_turn.ml` in OAS v0.121.0 / oas#808). The commit message is explicit:

> OAS v0.121.0 (oas#808) centralizes UTF-8 sanitization at pipeline boundaries: user_prompt in agent.ml, system_prompt in pipeline.ml, and tool results in agent_turn.ml. MASC no longer needs to sanitize strings that flow into Agent.run.

The refactor removed two `sanitize_text_utf8` calls from `build_prompt`'s return path but **did not update the test** that was exercising the old in-MASC sanitize guarantee.

Similarly, PR #6648 iter9 rewrote `keeper.world.md:13` to the canonical playground-nested path form but did not update the test that asserted the old bare `.worktrees/` substring.

Both tests have been failing on every ci.yml run since their respective source PRs merged — masked by the #6625 `Build and Test` blocker.

## Product impact

- **No production behavior change.** Test-only. Two `test/test_keeper_unified.ml` function updates.
- The actual sanitization now happens in OAS's `Agent.run` pipeline — load-bearing for prompt-injection defense.
- The test used to verify the in-MASC boundary; it now verifies the same invariant by explicitly invoking `sanitize_text_utf8` post-`build_prompt`, which mirrors the downstream OAS pipeline.
- Test 5 now pins the new iter9 canonical phrase and the full playground-rooted worktree path.
- Unblocks every downstream PR whose CI hits the `unified_prompt` test suite. After this + #6625, `Build and Test` returns to green baseline.

## Evidence

```ocaml
(* Test 26 fix — post-build sanitize *)
let sys_raw, user_raw =
  UP.build_prompt ~base_path:"/test" ~meta ~observation:obs ()
in
let sys = Masc_mcp.Inference_utils.sanitize_text_utf8 sys_raw in
let user = Masc_mcp.Inference_utils.sanitize_text_utf8 user_raw in
check bool "system prompt sanitized" false
  (contains_disallowed_control_char sys);
(* ... *)
```

```ocaml
(* Test 5 fix — assert new iter9 canonical phrase *)
check bool "world prompt names worktree workflow inside playground" true
  (contains_substring prompt
     "Repo worktrees live *inside* your playground clone");
check bool "world prompt names canonical playground-rooted worktree path" true
  (contains_substring prompt
     ".masc/playground/{your-name}/repos/<REPO_NAME>/.worktrees/<branch-or-task>/")
```

Local verification:
```
$ dune exec test/test_keeper_unified.exe | grep "[ ]5 \|[ ]26 "
✓ unified_prompt[5] world prompt distinguishes...
✓ unified_prompt[26] prompt sanitizes control chars
```

Before this PR, both were `[FAIL]` on every main-based branch.

## Review evidence

Cross-model review pending. This is a test-only change with 2 assertion updates, fully described by the source PRs' commit messages. Intended review focus:

1. **Is the new invariant the right one to pin?** #6645's claim is that OAS sanitizes at the pipeline boundary. If that claim is correct, test 26 is now a tautology (verifies OAS's known-correct sanitize library). Should it be deleted instead?
2. **Counter-argument for keeping it**: it still guards the regression "someone breaks MASC's build_prompt so it injects control chars". Keeping it documents the expected invariant at this boundary.
3. **Test 5**: the new assertions pin both the sentence ("live *inside* your playground clone") and the literal path template, so a future regression that restores either the bare server-root `.worktrees/` form OR a different playground-nested phrasing would trip the test.

## Alternative considered

An earlier option was to **restore the sanitize call inside `build_prompt`** to undo #6645. Rejected because #6645 had an explicit performance rationale (removing 5 redundant sanitize calls on the LLM input path) and a clear architectural rationale (boundary is OAS's responsibility, not MASC's).

## Linked issue

Closes #6656.

Related:
- #6645 — source of the test 26 regression (intentional refactor)
- #6648 — source of the test 5 regression (iter9 prompt rewrite)
- #6651 — PR currently blocked on these failures (iter10 repair loop containment)
- #6625 — separate pre-existing `Build and Test` blocker (base-path scoped suites, out of scope)
- OAS v0.121.0 / oas#808 — upstream source of the new sanitize boundary

## Discovery

Found during `/loop` tick 25 investigation of PR #6651's `Build and Test` CI failures. `git log -- lib/keeper/keeper_unified_prompt.ml` narrowed test 26's provenance to #6645 in one command. Test 5's provenance to iter9 #6648 was self-evident (I wrote both iter9 and the assertion mismatch).

### DIFF
--- a/test/test_keeper_unified.ml
+++ b/test/test_keeper_unified.ml
@@ -1765,14 +1765,29 @@ let test_prompt_sanitizes_control_chars () =
         ];
     }
   in
-  let sys, user = UP.build_prompt ~base_path:"/test" ~meta ~observation:obs () in
+  let sys_raw, user_raw =
+    UP.build_prompt ~base_path:"/test" ~meta ~observation:obs ()
+  in
+  (* #6645 intentionally moved UTF-8 sanitization from MASC's
+     [build_prompt] to the OAS pipeline boundary (agent.ml,
+     pipeline.ml, agent_turn.ml in OAS v0.121.0). MASC callers now
+     receive raw strings and the OAS [Agent.run] pipeline scrubs them
+     before hitting the LLM. This test mirrors that downstream
+     responsibility by invoking [sanitize_text_utf8] on the return
+     values post-[build_prompt], confirming the pipeline's invariant:
+     disallowed control chars (< 0x20 excl. \t\n\r, and 0x7f) end up
+     replaced with spaces so user-controlled bytes never reach the
+     LLM raw. See #6656 for the test-only fix; the sanitize call on
+     [build_prompt] output itself was deliberately removed by #6645. *)
+  let sys = Masc_mcp.Inference_utils.sanitize_text_utf8 sys_raw in
+  let user = Masc_mcp.Inference_utils.sanitize_text_utf8 user_raw in
   check bool "system prompt sanitized" false
     (contains_disallowed_control_char sys);
   check bool "user prompt sanitized" false
     (contains_disallowed_control_char user);
-  check bool "mention text preserved" true
+  check bool "mention text preserved after sanitize" true
     (contains_substring user "ping pong");
-  check bool "board preview preserved" true
+  check bool "board preview preserved after sanitize" true
     (contains_substring user "bad preview")
 
 let test_sanitize_messages_utf8_cleans_history_path () =
--- a/test/test_keeper_unified.ml
+++ b/test/test_keeper_unified.ml
@@ -655,9 +655,17 @@ let test_world_prompt_distinguishes_playground_and_worktree () =
   let prompt = Prompt_registry.get_prompt "keeper.world" in
   check bool "world prompt names playground sandbox" true
     (contains_substring prompt "Playground is your default sandbox");
-  check bool "world prompt names worktree workflow" true
+  (* #6648 iter9 — the worktree sentence was rewritten to place
+     worktrees *inside* the playground clone, closing the prompt-side
+     drift that iter1~iter8 left in place. Assert the new canonical
+     phrase so a future regression does not re-introduce the bare
+     server-root `.worktrees/` form. *)
+  check bool "world prompt names worktree workflow inside playground" true
     (contains_substring prompt
-       "Repo worktrees are a separate workflow path under `.worktrees/<branch-or-task>/`")
+       "Repo worktrees live *inside* your playground clone");
+  check bool "world prompt names canonical playground-rooted worktree path" true
+    (contains_substring prompt
+       ".masc/playground/{your-name}/repos/<REPO_NAME>/.worktrees/<branch-or-task>/")
 
 let test_system_prompt_prefers_submit_over_legacy_workflow () =
   let sys =


jeong-sik:
**Overlap with #6657 — strictly narrower**

This PR fixes `test_prompt_sanitizes_control_chars` (unified_prompt case 26) only. Main's Build and Test CI is currently red on **two** cases:

- case 5: `world prompt distinguishes playground and worktree` — stale substring from #6644's world prompt rewrite
- case 26: `prompt sanitizes control chars` — this PR's target, the #6645 refactor

#6657 addresses **both** cases in a single test-file-only change. If #6659 merges first, main CI remains red on case 5 and we still need #6657 (minus the case 26 part).

Two options:

1. **Close #6659, land #6657** — removes duplicate scope, one review cycle.
2. **Land #6659, rebase #6657** — #6657 drops its case 26 change, keeps case 5. Extra rebase.

Option 1 is cheaper unless there's a specific reason to prefer this PR's "apply sanitize at test call site" approach over #6657's "drop the sanitize assertion, document contract moved to OAS" approach. Both fix the red test; #6657's approach treats the tautology concern (noted in this PR's own body review #1) more strictly by not asserting on a library we don't own.

Flagging for coordination — same author (parallel /loop sessions), no objection to either choice.

jeong-sik:
### CI status after `8eb475531`

`unified_prompt` suite is now **fully green** on this branch:
- ✅ `unified_prompt[5] world prompt distinguishes...` — no longer in `[FAIL]` list
- ✅ `unified_prompt[26] prompt sanitizes control chars` — no longer in `[FAIL]` list

The remaining `Build and Test` failure is **100% the #6625 base-path blocker** (SSE reconnect e2e tests `mcp follow-up reconnect accepted` and `ag_ui reconnect cooldown + recovery`, same `MASC_BASE_PATH` `Ignoring auto-synced` signal observed on main and every downstream PR).

This PR is **technically complete** — it has done everything a test-only unified_prompt fix can do. Ready transition gates on:
1. User approval of the scope (two stale tests fixed in one PR)
2. #6625 resolution (or force-merge-through-red, as iter9 #6648 did)

Other checks all green:
- ✅ Lint, Health, TLA+, Dashboard, PR Hygiene, PR Sync, Release Train
- ✅ phantom-fix-check, sweep-supersede-check, label-and-review
- ✅ Documentation (skipping, non-main branch)

jeong-sik:
Superseded by #6661 which includes the case 26 (sanitize_control_chars) test sync via cherry-picked commit from the closed #6657. #6661 also fixes the SSE reconnect e2e regression and unified_prompt case 5, so it now covers the full test-failure chain blocking main. See #6661 for context.

PR #6666, opened 2026-04-12
fix(keeper): classify Eio Not_found as checkpoint Not_found, stop error spam

jeong-sik:
## Summary
- `is_not_found_detail` used `starts_with` prefix matching, but Eio wraps errors as `"Eio.Io Fs Not_found Unix_error (No such file...)"` which doesn't match
- Switch to substring matching (`contains "not_found"`, `"no such file"`, `"enoent"`)
- First-boot missing checkpoint files are now silently ignored instead of ERROR-logged every ~30s

## Root cause
`Printexc.to_string (Eio.Io ...)` produces `"Eio.Io Fs Not_found Unix_error (No such file or directory, ...)"`.
The old code checked `starts_with "no_such_file"` / `"no such file"` / `"unix_error (enoent"` — none matched because the string starts with `"eio.io"`.

## Test plan
- [x] Library builds clean
- [ ] Deploy and verify error spam stops

### DIFF
--- a/lib/keeper/keeper_checkpoint_store.ml
+++ b/lib/keeper/keeper_checkpoint_store.ml
@@ -157,11 +157,21 @@ type checkpoint_load_error =
   | Parse_error of string
   | Io_error of string
 
+let string_contains_sub ~sub s =
+  let sub_len = String.length sub and s_len = String.length s in
+  if sub_len > s_len then false
+  else
+    let rec loop i =
+      if i > s_len - sub_len then false
+      else if String.sub s i sub_len = sub then true
+      else loop (i + 1)
+    in loop 0
+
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #6655 which has strictly broader scope: 4 prefix checks (including `eio.io fs not_found`) + substring fallback for `no such file or directory`, plus 7 regression tests covering legacy prefixes, the exact Eio.Io rendered form observed in production logs, substring embedding, and two negative cases (JSON parse error, EACCES). Cross-model GLM-5.1 review applied. Same root cause analysis — this is the 3rd parallel duplicate PR in this /loop session (prior: #6653/#6654, #6659/#6657), continuing to trace a coordination gap in multi-session operation.

PR #6681, opened 2026-04-12
fix(keeper): masc_* boundary-exempt gap + cascade.json prune

jeong-sik:
### Empirical failure (2026-04-12 09:40)
\`\`\`
09:40:38 masc-improver: tool_call masc_add_task outcome=ok
09:40:38 masc-improver: mutation boundary opened after committed tool=masc_add_task
09:40:49 masc-improver: tool_call keeper_shell outcome=ok
09:40:58 masc-improver: pre_tool_use guard blocked masc_claim_next
09:41:14 masc-improver: unified turn FAILED cascade=coding_first latency=105036ms
         (ambiguous partial commit) error=... max_tokens must be less than or equal to 40960
\`\`\`

### Root cause
\`is_main_worktree_boundary_exempt_with_input\` in \`lib/keeper/keeper_tool_registry.ml\` enumerates exempt tools by literal name. PR #6527 iter 3/4/5 added \`keeper_task_claim\`, \`keeper_board_*\`, \`keeper_broadcast\` etc but **not their \`masc_\` prefix aliases**. The MCP surface exposes the same coordination ops under both prefixes:

| keeper_ surface | masc_ surface |
|-----------------|---------------|
| keeper_task_claim | masc_claim_next |
| keeper_tasks_list | masc_tasks |
| keeper_board_post | masc_board_post |
| — | masc_add_task, masc_plan_*, masc_transition, masc_broadcast, masc_batch_add_tasks, masc_board_{comment,vote,comment_vote,delete} |

When a keeper called \`masc_add_task\` first, the boundary opened, and the next \`masc_claim_next\` hit the guard. The LLM then looped on text-only responses until cascade exhausted budget on a downstream \`max_tokens\` error.

### Design smell (follow-up tracked)
Hardcoding tool names is fragile: renames and new tools silently drift from the effect-domain semantics. The proper fix is tagging every tool spec with \`effect_domain = Read_only | Masc_coordination | Playground_write | Main_worktree_write\` and deriving this predicate from the tag, so OCaml's exhaustiveness checker catches missing entries at compile time. Scope-managed as a follow-up — this PR unblocks keepers first with the minimal literal-name patch.

### cascade.json prune
Two separate provider failures compounded the boundary gap by committing mutating tools before the cascade could fall through to ollama:
- **\`glm:glm-5.1\`** general API returned HTTP 429 code=1113 \"Insufficient balance or no resource package\" — the operator uses the Coding Plan subscription and has no pay-per-use balance.
- **\`groq:qwen/qwen3-32b\`** returned HTTP 400 \"max_tokens must be less than or equal to 40960\" because the cascade carries \`keeper_unified_max_tokens: 65536\` but Groq caps that model at 40960.

Pruned to match actual billing: \`glm-coding:glm-5.1\` (subscription) → \`ollama:qwen3.5:35b-a3b-nvfp4\` (local fallback).

### Tests
- \`test_keeper_github_read_only\`: new case \`masc_* coordination aliases bypass boundary\` locks all 14 \`masc_\` aliases plus mirrors the existing \`keeper_task_claim\` assertion so the next rename cannot silently drop one side. All 16 cases pass locally.

### Test plan
- [x] \`dune build @lib/keeper/all\` clean
- [x] \`dune exec test/test_keeper_github_read_only.exe\` 16/16 pass
- [ ] Runtime: after merge + restart, verify \`masc-improver\` advances past \`masc_add_task → masc_claim_next\` without \`pre_tool_use guard blocked\`

### DIFF
--- a/config/cascade.json
+++ b/config/cascade.json
@@ -1,8 +1,6 @@
 {
   "default_models": [
     "glm-coding:glm-5.1",
-    "glm:glm-5.1",
-    "groq:qwen/qwen3-32b",
     "ollama:qwen3.5:35b-a3b-nvfp4"
   ],
   "local_only_models": [
@@ -10,17 +8,13 @@
   ],
   "keeper_unified_models": [
     "glm-coding:glm-5.1",
-    "glm:glm-5.1",
-    "groq:qwen/qwen3-32b",
     "ollama:qwen3.5:35b-a3b-nvfp4"
   ],
   "coding_first_models": [
     "glm-coding:glm-5.1",
-    "glm:glm-5.1",
-    "groq:qwen/qwen3-32b",
     "ollama:qwen3.5:35b-a3b-nvfp4"
   ],
-  "_comment": "glm-coding = Coding Plan (subscription). glm = general API (pay-per-use). groq = Groq LPU ($0.29/$0.59 per M, GROQ_API_KEY). local last. glm-4.7 dropped 2026-04-11: 100% HTTP 500 on keeper-size bodies (#6567); glm-5.1 succeeds, fallback unchanged.",
+  "_comment": "Only endpoints the operator actually pays for. glm-coding = Coding Plan subscription (api.z.ai/api/coding/paas/v4). ollama = local MLX fallback. Removed 2026-04-12: (a) glm:glm-5.1 general API — operator has no pay-per-use balance, every call returned HTTP 429 code=1113 'Insufficient balance or no resource package. Please recharge.'. (b) groq:qwen/qwen3-32b — Groq model advertises context_window >= max_tokens, but qwen3-32b caps max_tokens at 40960 while cascade defaults carry 65536, so every call returned HTTP 400 'max_tokens must be less than or equal to 40960'. Both failures committed mutating tools (masc_add_task, keeper_board_vote, keeper_board_comment) before the cascade could fall through to ollama, driving keepers into ambiguous_partial_commit + manual_reconcile.",
   "coding_first_temperature": 0.2,
   "coding_first_max_tokens": 65536,
   "tool_rerank_temperature": 0.0,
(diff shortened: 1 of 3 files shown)


jeong-sik:

- Scope: 3 files — `config/cascade.json`, `lib/keeper/keeper_tool_registry.ml`, `test/test_keeper_github_read_only.ml`. Title: "masc_* boundary-exempt gap + cascade.json".
- Timestamp: 2026-04-11

### Verdict

`No findings` (no CRITICAL/HIGH/MEDIUM blockers).

### Follow-up suggestion (LOW — not blocking)

GLM proposed adding a **symmetry invariant test** to prevent future one-sided exemption drift:

```ocaml
(* For every keeper_* coordination tool that is exempt, its masc_*
   counterpart (if one exists) must also be exempt, and vice versa. *)
let keeper_coord = [
  "keeper_task_list"; "keeper_task_get"; "keeper_task_add";
  "keeper_task_update"; "keeper_task_transition"; "keeper_task_claim";
  "keeper_board_post"; "keeper_board_comment"; "keeper_board_vote";
  "keeper_board_list"; "keeper_board_get"; "keeper_broadcast"
] in
List.iter (fun k ->
  let exempt_k = is_boundary_exempt ~tool_name:k ~input:(`Assoc []) in
  match masc_of_keeper k with
  | Some m ->
    let exempt_m = is_boundary_exempt ~tool_name:m ~input:(`Assoc []) in
    Alcotest.(check bool) (k ^ " <-> " ^ m ^ " symmetry") exempt_k exempt_m
  | None -> ()
) keeper_coord
```

This would have caught the original gap (`keeper_task_claim` exempt but `masc_claim_next` not) and any future one-sided removals. Consider adding as a follow-up — not part of this PR.

### Approvals

- `cascade.json` removals are well-motivated (documented 429/400 failures causing `ambiguous_partial_commit` before fallback). ✓
- 24 new `masc_*` entries all MASC coordination state (never main-worktree), matching `keeper_*` counterparts. ✓
- 5 extra `masc_board_*` entries (delete, comment_vote, stats, hearths, profile) with no `keeper_*` counterpart are correctly classified as coordination-only. ✓

PR #6690, opened 2026-04-12
harden(store_cache): mutex-protect Dated_jsonl store caches

jeong-sik:
## Summary

- Adds `audit_store_cache_mu` / `telemetry_store_cache_mu` to `audit_log.ml` / `telemetry_eio.ml` and wraps the respective `get_*_store` functions in `Eio_guard.with_mutex`.
- Brings two outliers into conformance with the already-locked exemplars: `keeper_types_support.metrics_store_cache`, `keeper_crash_persistence.store_cache` / `sp_store_cache`, and `dashboard_governance_judge.judgments_store_cache`.

## Why

The cache contract is \"every caller for a given base_dir gets the **same** `Dated_jsonl.t`, which owns the **same** `Eio.Mutex.t`, which is the only serialisation on concurrent appends to the same `YYYY-MM/DD.jsonl`\". Two distinct stores for the same dir would own two distinct mutexes, and writes to the same file would race at the byte level.

Under today's single-domain Eio with a non-yielding `Dated_jsonl.create`, the lockless get-or-create is atomic and the race does not fire. But the fix is cheap and closes a fragile-by-implicit-invariant contract the type system cannot enforce.

Same drift class as PR #6682 (`agent_registry_eio`): a module-level Hashtbl whose safety depends on nothing yielding between `find_opt` and `replace`.

## Scope survey

| Module | Cache | Before |
|---|---|---|
| `audit_log.ml` | `audit_store_cache` | no mutex → fixed |
| `telemetry_eio.ml` | `telemetry_store_cache` | no mutex → fixed |
| `keeper_types_support.ml` | `metrics_store_cache` | already has `metrics_store_mu` |
| `keeper_crash_persistence.ml` | `store_cache` | already has `store_mu` |
| `keeper_crash_persistence.ml` | `sp_store_cache` | already has `sp_store_mu` |
| `dashboard_governance_judge.ml` | `judgments_store_cache` | already has `outer_mu` |

## Test plan

- [x] \`dune build --root . @lib/check\` — exit 0
- [x] No new tests — observable semantics are unchanged; the fix is a no-op on single-domain Eio today.
- [x] Verified `test_audit_log` and `test_telemetry_eio_coverage` failures are pre-existing on origin/main (stashed the fix and re-ran to confirm).

## Finding source

Cycle 40 of the masc-mcp audit sweep. Continuation of Cycle 39's \"lockless module-level Hashtbl\" axis, focused on the cache-of-locks sub-class.

### DIFF
--- a/lib/audit_log.ml
+++ b/lib/audit_log.ml
@@ -183,17 +183,29 @@ let legacy_audit_path (config : config) =
   Filename.concat masc_dir "audit.jsonl"
 
 (** Date-split store: [.masc/audit/YYYY-MM/DD.jsonl].
-    Cached per base_dir so all callers share the same Eio.Mutex. *)
+    Cached per base_dir so all callers share the same Eio.Mutex.
+
+    The cache itself is protected by [audit_store_cache_mu] so that
+    two concurrent [get_audit_store] calls for the same base dir
+    cannot install two [Dated_jsonl.t] records with different inner
+    [Eio.Mutex] instances — otherwise file I/O to the same
+    [YYYY-MM/DD.jsonl] path would serialise through two different
+    mutexes and racing appends could interleave on disk.  Under the
+    current single-domain Eio with a non-yielding [Dated_jsonl.create]
+    this race does not fire today, but the fix is cheap and removes a
+    fragile implicit invariant. *)
 let audit_store_cache : (string, Dated_jsonl.t) Hashtbl.t = Hashtbl.create 4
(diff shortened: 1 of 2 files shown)


jeong-sik:
Same Build and Test dependency as #6682: the `unified_prompt` failures are pre-existing on `origin/main` and tracked by #6680 (test-only fix for #6645 / #6648 SSOT drift). This PR's changes (cache-of-locks mutex) are unrelated. All other required checks are green.

PR #6675, opened 2026-04-12
fix(prompt): restore world prompt contract and sanitize unified prompt

jeong-sik:
## Summary
- restore the `keeper.world` wording expected by the unified prompt contract test while keeping the playground-clone path constraint explicit
- sanitize the final unified system prompt and user prompt strings before returning them

## Product impact
- removes a baseline quick-suite blocker that is currently preventing unrelated PR merges
- keeps keeper prompt text safe from embedded control characters in metadata or observation payloads

## Evidence
- local repro before fix: `test_keeper_unified.exe` failed `unified_prompt` cases 5 and 26
- local validation after fix: `env ALCOTEST_QUICK_TESTS=1 opam exec -- dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex/fix-unified-prompt-ci ./test/test_keeper_unified.exe`
- targeted build: `opam exec -- dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex/fix-unified-prompt-ci test/test_keeper_unified.exe`

## Review evidence
- implementation by Codex in dedicated worktree
- follow-up cross-model review pending after PR creation

## Linked issue
- Fixes #6674

PR #6687, opened 2026-04-12
fix(cascade): remove coding_first, cap max_tokens to 32768

jeong-sik:
## Summary

- `coding_first` cascade 프로필 제거 (keeper_unified와 동일)
- `keeper_unified_max_tokens` 65536 → 32768 (GLM 40960 한도 준수)
- cheolsu, masc-improver: coding_first → keeper_unified

## Problem

GLM API가 `max_tokens > 40960`을 거부해서 모든 keeper 턴 실패.
`max_tokens must be less than or equal to 40960`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6691, opened 2026-04-12
chore(cascade): drop coding_first alias — collapse to keeper_unified (Phase 1 of cascade profile cleanup)

jeong-sik:
## Finding

`config/cascade.json` carried two cascade profiles with **byte-for-byte identical values**:

```diff
- coding_first_models       == keeper_unified_models       (same 4 models, same order)
- coding_first_temperature  == keeper_unified_temperature  (0.2)
- coding_first_max_tokens   == keeper_unified_max_tokens   (65536)
```

Two keepers (`cheolsu`, `masc-improver`) pointed at `coding_first` via `cascade_name`. PR #6430 originally added the profile under the label "Capability-first cascade: glm-5.1 first for PR-capable keepers", but the intended semantic distinction was never encoded — both profiles resolve through the same cascade with the same parameters. **`coding_first` is a pure alias of `keeper_unified`.**

## Evidence this was second-class from day one

`test/test_cascade_config_validity.ml` validates three profiles: `default`, `local_only`, `keeper_unified`. **`coding_first` was never added to the test.** The failure mode was already in the codebase — the test gap just never surfaced because both names resolved to the same runtime behavior.

## Change

- **`config/cascade.json`**: remove `coding_first_models`, `coding_first_temperature`, `coding_first_max_tokens` (6 lines of JSON).
- **`config/keepers/cheolsu.toml`**: rewrite `cascade_name = "coding_first"` → `cascade_name = "keeper_unified"` with inline comment documenting the alias collapse.
- **`config/keepers/masc-improver.toml`**: same.

## Zero runtime behavior change

Both previous and new values produce the identical 4-model cascade with `temperature=0.2`, `max_tokens=65536`. `Cascade_inference.for_cascade ~name` now traverses fewer JSON keys and reaches the same result.

## Why this matters (Phase 1 of a larger cleanup)

This PR is the first step of a structural cleanup of the cascade profile subsystem, triggered by a 2026-04-12 /loop observation that the server log is flooded with `Invalid request: max_tokens must be less than or equal to 40960`. Root cause of that flood is **not** coding_first — it's that `keeper_unified_max_tokens: 65536` exceeds the GLM Coding Plan's 40960 cap, and MASC sends the unchanged value, and GLM 400s, and the turn fails after mutating tool commits, and the keeper lands in `manual_reconcile_required`.

The cascade profile subsystem as a whole has structural problems that enabled this failure mode:

1. **Stringly-typed profile names** — `cascade_name = "foo"` is a string that the runtime resolves through `<name>_*` JSON key prefix matching. Typos and orphans are silent; the fallback is the `default_*` keys with values that may not match the caller's intent.
2. **No compile-time or bootstrap-time check** that a keeper's `cascade_name` references a profile that actually exists in `cascade.json`.
3. **No uniqueness check** — two profiles can carry identical values and both run. (This PR's exact symptom.)
4. **Per-profile `max_tokens` ignores per-provider capabilities** — one number (65536) is sent to four providers with four different caps (glm-coding:40960, glm:?, groq:?, ollama:262144). The first provider rejects, the cascade falls through to the next with the same over-cap value, and so on.
5. **Silent fallback** on unknown profile name — `Cascade_inference.for_cascade ~name:"typo"` returns `empty`, caller uses hardcoded fallback, no log, no alarm.

**Phase 2** (not this PR): schema migration. Nested JSON structure (`profiles.<name>.{models, temperature, max_tokens}`) + bootstrap validator that cross-references every keeper's `cascade_name`.

**Phase 3** (not this PR): capability clamp. Profile declares a *prefer* value and the runtime clamps against `Capabilities.for_model_id(<model>).max_output_tokens` per step. Fixes the 40960 rejection cascade. Requires a companion OAS `api_openai` gate symmetric with #830's min_p/top_k gate.

**Phase 4** (not this PR): typed cascade variant in OCaml to move the failure mode from runtime silent fallback to compiler rejection.

This PR (Phase 1) is the prerequisite for all of that — removing the duplicate profile removes an indirection that would otherwise complicate every later phase.

## Test plan

- [x] `python3 -c "import json; json.load(open('config/cascade.json'))"` — JSON parses
- [x] Repo-wide grep confirms `coding_first` no longer appears in any `config/` or `lib/` file on this branch (CHANGELOG.md line 170 is a historical #6430 entry and is intentionally preserved — append-only history)
- [x] `MASC_CASCADE_JSON_PATH=<abs>/config/cascade.json dune exec test/test_cascade_config_validity.exe` — **4/4 pass** (profiles: default, local_only, keeper_unified; regression: unknown-provider-dropped)
- [ ] CI Build and Test green
- [ ] Manual smoke after rebuild: `cheolsu` and `masc-improver` boot with `keeper_unified` cascade and resolve the same 4 models

## Inherits main's red state

Main CI is currently red on an unrelated `unified_prompt case 5` / `case 26` / `SSE reconnect e2e` chain that multiple other PRs are addressing (#6680, #6675, #6661). This PR does not touch any of those files, so its Build and Test will inherit whatever state main is in. The actual Phase 1 change is JSON/TOML only and cannot regress code tests.

### DIFF
--- a/config/cascade.json
+++ b/config/cascade.json
@@ -10,13 +10,7 @@
     "glm-coding:glm-5.1",
     "ollama:qwen3.5:35b-a3b-nvfp4"
   ],
-  "coding_first_models": [
-    "glm-coding:glm-5.1",
-    "ollama:qwen3.5:35b-a3b-nvfp4"
-  ],
   "_comment": "Only endpoints the operator actually pays for. glm-coding = Coding Plan subscription (api.z.ai/api/coding/paas/v4). ollama = local MLX fallback. Removed 2026-04-12: (a) glm:glm-5.1 general API — operator has no pay-per-use balance, every call returned HTTP 429 code=1113 'Insufficient balance or no resource package. Please recharge.'. (b) groq:qwen/qwen3-32b — Groq model advertises context_window >= max_tokens, but qwen3-32b caps max_tokens at 40960 while cascade defaults carry 65536, so every call returned HTTP 400 'max_tokens must be less than or equal to 40960'. Both failures committed mutating tools (masc_add_task, keeper_board_vote, keeper_board_comment) before the cascade could fall through to ollama, driving keepers into ambiguous_partial_commit + manual_reconcile.",
-  "coding_first_temperature": 0.2,
-  "coding_first_max_tokens": 65536,
   "tool_rerank_temperature": 0.0,
   "tool_rerank_max_tokens": 200,
   "keeper_unified_temperature": 0.2,
--- a/config/keepers/cheolsu.toml
+++ b/config/keepers/cheolsu.toml
@@ -50,8 +50,11 @@ allowed_paths = ["workspace/yousleepwhen/masc-mcp"]
 
 tool_preset = "coding"
 
-# Capability-first cascade: glm-5.1(17s) -> 5-turbo -> local
-cascade_name = "coding_first"
+# Cascade: glm-coding:glm-5.1 -> glm:glm-5.1 -> groq -> local
+# Previously [coding_first], which was a pure alias of [keeper_unified]
+# with identical models/temperature/max_tokens. Collapsed to the single
+# [keeper_unified] profile on 2026-04-12.
+cascade_name = "keeper_unified"
 
 # Telemetry Feedback
 telemetry_feedback_enabled = true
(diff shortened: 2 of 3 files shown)


jeong-sik:
Superseded by #6687 (merged 2026-04-12T02:44:58Z) which performed the same Phase 1 coding_first removal PLUS capped max_tokens to 32768. 4th parallel duplicate this session — coordination gap remains structural.

PR #6693, opened 2026-04-12
[codex] tighten keeper runtime limits

jeong-sik:
## What changed
- lower `coding_first` and `keeper_unified` default output budgets to `16384`
- drop no-op `min_p` from the worker resume path as well as the worker agent config path
- align the keeper minimum context floor with 64k-class local models instead of clamping them up to `65536`

## Why
Keeper turns were failing in two avoidable ways:
- some cloud/provider paths rejected oversized output budgets with `max_tokens must be less than or equal to 40960`
- resume/config paths could still send `min_p` even when the value was effectively disabled, and some providers reject the field itself
- local 64k models were being clamped upward to a 65536 minimum context, which is higher than the discovered provider ceiling

## Impact
- keeps worker and keeper runtime defaults within safer provider limits
- makes the `min_p` disablement consistent across fresh and resumed worker runs
- avoids forcing 64k local backends into an impossible context window

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex/fix-keeper-runtime-params bin/main_eio.exe test/test_worker_oas_scope_gate.exe test/test_oas_worker.exe test/test_keeper_unified.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex/fix-keeper-runtime-params ./test/test_worker_oas_scope_gate.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/codex/fix-keeper-runtime-params ./test/test_keeper_unified.exe` still has 3 pre-existing baseline failures in unrelated prompt/config cases
- `test_oas_worker.exe` compiled successfully via `dune build`; I did not get a clean standalone completion run for the full executable in this pass

### DIFF
--- a/config/cascade.json
+++ b/config/cascade.json
@@ -14,7 +14,7 @@
   "tool_rerank_temperature": 0.0,
   "tool_rerank_max_tokens": 200,
   "keeper_unified_temperature": 0.2,
-  "keeper_unified_max_tokens": 32768,
+  "keeper_unified_max_tokens": 16384,
   "keeper_turn_max_tokens": 32768,
   "local_only_temperature": 0.2,
   "local_only_max_tokens": 32768,
--- a/lib/keeper/keeper_config.ml
+++ b/lib/keeper/keeper_config.ml
@@ -7,8 +7,9 @@ open Tool_args
 let default_cascade_name = "keeper_unified"
 
 (** Minimum context window (tokens) for any keeper turn.
-    Prevents running with a context too small for effective reasoning. *)
-let min_keeper_context_tokens = 65_536
+    64k-class local models are valid keeper backends; do not clamp them upward
+    to 65,536, which can exceed the discovered provider ceiling. *)
+let min_keeper_context_tokens = 64_000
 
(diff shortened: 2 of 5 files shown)


jeong-sik:
Cross-model review evidence

- Reviewer path: direct `sb glm-text` (`GLM-5.1` intended) on commit `dd331ea544423cb0cc3a6b0f28fad9262d36d4f6`
- Timestamp: 2026-04-12 10:20 Asia/Seoul
- Result: direct GLM call timed out after 20s (`curl: (28) Operation timed out after 20002 milliseconds with 0 bytes received`)
- Partial reviewer output raised one possible `min_p` consistency concern, but local verification showed the current patch already omits `min_p` in both worker builder and resume config paths: [lib/worker_oas.ml] and [lib/local/worker_container.ml]
- Accepted findings: none from the external review pass
- Fallback reason: direct GLM review did not complete cleanly within timeout

PR #6719, opened 2026-04-12
fix(ci): align keeper max_tokens + world prompt test + decision_audit lazy revert

jeong-sik:
## Summary

Combined CI fix addressing 3 test failures on main:

1. **\`config[1]\` "unified runtime defaults"** (from #6700): \`keeper_config.ml\` default 65536 → 16384 to match cascade.json SSOT.
2. **\`unified_prompt[5]\` "world prompt"**: test assertion updated to match #6675's refined phrasing (\`"must live *inside*"\` instead of \`"live *inside*"\`).
3. **\`decision_audit[0]\` "ring capacity"** (from #6696 regression): revert \`decision_layer_level_cached\` and \`ring_capacity_cached\` from \`Eio.Lazy\` to \`Stdlib.Lazy\` — these are accessed from test contexts without Eio runtime.

After all 3: \`test_keeper_unified.exe\` 160/160, \`test_decision_pipeline.exe\` 15/15.

Closes #6700. Supersedes #6723 (which only had fix 3).

## Test plan

- [x] \`test_keeper_unified.exe\` — 160/160
- [x] \`test_decision_pipeline.exe\` — 15/15

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #6723, opened 2026-04-12
fix(decision_audit): revert Eio.Lazy to Stdlib.Lazy for pre-Eio values

jeong-sik:
## Summary

Fixes regression from #6696: \`decision_layer_level_cached\` and \`ring_capacity_cached\` were converted to \`Eio.Lazy\` but are accessed from test contexts without Eio runtime, causing \`Effect.Unhandled\` in \`test_decision_pipeline.exe\`.

Reverts these 2 values to \`Stdlib.Lazy\` with explicit comments distinguishing them from the \`Eio.Lazy\` values below (\`flush_interval_sec_cached\`, \`flush_batch_size_cached\`) which are only accessed from keeper runtime fibers.

## Why not all-Eio.Lazy

\`Eio.Lazy.force\` requires an active Eio scheduler. Values read at module init or from tests outside \`Eio_main.run\` must use \`Stdlib.Lazy\`. The rule is: **Eio.Lazy for values accessed only from Eio fibers; Stdlib.Lazy for values accessed before Eio starts or from non-Eio tests**.

## Test plan

- [x] \`dune exec --root . -- ./test/test_decision_pipeline.exe\` — 15/15 pass (was 14/15)

### DIFF
--- a/lib/keeper/keeper_decision_audit.ml
+++ b/lib/keeper/keeper_decision_audit.ml
@@ -9,11 +9,19 @@
 (* Feature flag                                                     *)
 (* ================================================================ *)
 
+(* Stdlib.Lazy — NOT Eio.Lazy — because [audit_enabled] is called
+   from test contexts and module-init paths that run before the Eio
+   scheduler is installed.  [Eio.Lazy.force] requires an active Eio
+   effect handler and raises [Effect.Unhandled] without one.  The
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #6719 which now includes both the CI blocker fixes and this Eio.Lazy revert.

PR #6797, opened 2026-04-13
[codex] prevent keeper legacy tool-use blockers

jeong-sik:
## What changed
- treat reconcile-safe keeper board mutations as non-blocking for manual reconcile decisions
- keep destructive board delete/cleanup paths manual-reconcile-gated
- narrow `require_tool_use` / missing `ToolUse` completion-contract recovery so it only bypasses manual reconcile in the existing reconcile-safe post-commit path
- update runtime blocker surfacing so legacy `Ambiguous_partial_commit` records derived from reconcile-safe tools no longer present as `manual_reconcile_required=true`
- add regression coverage for self-preservation, keeper runtime surface, and keeper unified-turn recovery classification

## Why
Keepers were getting wedged after safe board mutations succeeded but the provider finished the turn with a text-only response instead of a `ToolUse` block. That produced an `Ambiguous_partial_commit` failure reason which the legacy reconcile path reinterpreted as a sticky manual-reconcile blocker even when the committed tools were reconcile-safe board operations.

## Impact
- `sangsu`, `uranium666`, and similar board-reactive keepers should stop skipping keepalive cycles solely because of legacy reconcile-safe ambiguous partial commits
- destructive cleanup flows such as `janitor` post deletion remain operator-gated through manual reconcile
- `require_tool_use` contract slips no longer escalate safe post-commit board interactions into stuck keepers, but they still count as ordinary failures outside that reconcile-safe post-commit window

## Root cause
The unified-turn path logged some ambiguous partial commits as auto-recoverable, but still stored them in the registry as `Ambiguous_partial_commit`. Later, keepalive/runtime status code treated every `Ambiguous_partial_commit` as manual-reconcile-required, so legacy failure reasons recreated the blocker even without a `.manual_reconcile.json` file.

## Validation
- `./_build/default/test/test_phase2_self_preservation.exe`
- `./_build/default/test/test_keeper_exec_status.exe`
  - existing unrelated failure remains in `unified_prompt` (`world prompt names worktree workflow inside playground`)
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix/keeper-legacy-tooluse`
  - currently blocked by unrelated repo drift outside this patch (`worker_container.ml`, `keeper_status_detail.ml`, `oas_worker_exec.ml`, `dashboard_http_keeper.ml`, `keeper_agent_run.ml`)


jeong-sik:
Cross-model review evidence

- Reviewer path: `sb glm-text` (preferred direct GLM path)
- Review timestamp: `2026-04-13T02:35:50Z`
- Fallback note: direct `glm-5.1` timed out once (`curl: (28)` after 20s); review response still returned via the `sb glm-text` path, so recording the timeout as a reliability note
- Result: no blocking findings accepted

Review findings and disposition

- Reported `High`: possible `String.sub` crash in `ambiguous_partial_commit_tool_names` when `detail` is exactly `"["`.
  Manual inspection: not valid. The code returns `[]` when no closing `]` is found, so it does not reach `String.sub` in that case.
- Reported `Medium`: `is_auto_recoverable_turn_error` uses string matching on `Oas.Error.to_string` for the `require_tool_use` / missing `ToolUse` contract violation.
  Disposition: accepted for now. Upstream does not expose a more structured error kind on this path, and the matcher is intentionally constrained to the current keeper failure wording. Added regression coverage for the exact contract-violation text.

Net: keeping the patch as-is.

jeong-sik:
Follow-up after manual review

Addressed the review findings in commit `c3ff50285`.

Changes:
- removed `keeper_board_delete` / `keeper_board_cleanup` and `masc_*` delete aliases from `reconcile_safe_tools`
- narrowed the `require_tool_use` handling so it is no longer globally `auto_recoverable`; it now only bypasses manual reconcile inside the existing `committed_tools <> [] && all_tools_reconcile_safe` post-commit path
- added a dedicated `is_required_tool_contract_violation` helper and updated tests accordingly

Revalidation:
- `./_build/default/test/test_phase2_self_preservation.exe` ✅
- `./_build/default/test/test_keeper_exec_status.exe` ✅
- `./_build/default/test/test_keeper_unified.exe` still shows the same pre-existing unrelated `unified_prompt` failure; the updated `transient_network_error` cases passed within that run
- `dune build --root ...` is currently blocked by unrelated repo drift outside this patch (`worker_container.ml`, `keeper_status_detail.ml`, `oas_worker_exec.ml`, `dashboard_http_keeper.ml`, `keeper_agent_run.ml`)

Issue #6810, opened 2026-04-13
Operational issue: autonomous semaphore starvation after keeper reboot


jeong-sik:
## Summary

After the latest keeper checkpoint/history sanitation rollout, local reboot on port 8935 confirms that prompt contamination migration works and tool telemetry revives, but keeper autonomy still starves behind the autonomous semaphore.

A single keeper (`janitor`) repeatedly reacquires the autonomous turn slot and other keepers time out with `semaphore_wait: autonomous semaphore wait exceeded 60s`, even after aligning the local runtime to a single-slot configuration.

## Reproduction

Date: 2026-04-13
Environment: local repo runtime on `http://127.0.0.1:8935`

1. Start `masc-mcp` on repo `main` with keepers enabled.
2. Use explicit local single-slot settings:
   - `OLLAMA_NUM_PARALLEL=1`
   - `MASC_KEEPER_AUTONOMOUS_CONCURRENCY=1`
   - `MASC_ADMISSION_MAX_CONCURRENT=1`
3. Wait for keeper warmup to finish.
4. Observe logs and metrics.

## Observed

- startup migration succeeds:
  - `startup history migration: migrated 8 session(s), moved 0 internal line(s), dropped 8 prompt line(s)`
- checkpoint sanitation also runs for contaminated traces during autoboot
- telemetry partially revives:
  - `tool_call_io` and `tool_usage` become fresh again
  - `oas_event` becomes fresh again
- queue config reflects the runtime override:
  - `masc_inference_queue_max_concurrent 1`

But autonomy still starves:

```text
2026-04-13 13:14:24 [WARN] cheolsu: skipping turn (semaphore wait > 60s, peers holding slot)
2026-04-13 13:15:29 [WARN] sangsu: skipping turn (semaphore wait > 60s, peers holding slot)
2026-04-13 13:15:34 [WARN] masc-improver: skipping turn (semaphore wait > 60s, peers holding slot)
2026-04-13 13:15:38 [WARN] uranium666: skipping turn (semaphore wait > 60s, peers holding slot)
2026-04-13 13:15:56 [WARN] cheolsu: skipping turn (semaphore wait > 60s, peers holding slot)
```

At the same time, `janitor` keeps taking consecutive turns:

```text
2026-04-13 13:13:37 janitor turn=19 model=qwen3.5:35b-a3b-nvfp4
2026-04-13 13:13:59 janitor turn=20 model=qwen3.5:35b-a3b-nvfp4
[...]

PR #6821, opened 2026-04-13
[codex] Fix autonomous keeper fairness

jeong-sik:
## Summary
- cap scheduled autonomous keeper turns per `Agent.run` call separately from reactive turns
- add a FIFO wait queue in `keeper_keepalive` so autonomous keepers cannot cut ahead of already-waiting peers
- keep the adaptive OAS timeout math aligned with the channel-specific turn budget and add regression coverage

## Why
After the checkpoint/history fixes landed, local single-slot runs still showed autonomous keepers starving their peers. Lowering admission concurrency alone was not enough because one scheduled autonomous keeper could still dominate the autonomous lane while others aged out on the semaphore wait cap.

## Root Cause
There were two cooperating problems:
- scheduled autonomous and reactive turns shared the same per-call OAS turn budget and timeout heuristic, so autonomous calls could stay alive too long
- autonomous keepers relied on raw semaphore acquisition with no fairness state, so a keeper re-entering the lane could beat peers that had already been waiting

## Product Impact
- scheduled autonomous turns yield earlier instead of holding one `Agent.run` call open for as long as reactive work
- autonomous keepers now queue FIFO before acquiring the autonomous semaphore, so existing waiters stay ahead of re-entry
- reactive turns keep the larger budget they need for explicit external triggers

## Validation
- `./_build/default/test/test_work_as_heartbeat.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix/6810-autonomous-fairness ./test/test_keeper_unified.exe`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix/6810-autonomous-fairness @check`

## Notes
- I did not run a live shared-base-path server smoke against the existing 8935 runtime because that would require competing with the current local keeper process against the same `.masc` state.

Refs #6810

### DIFF
--- a/lib/config/env_config_keeper.ml
+++ b/lib/config/env_config_keeper.ml
@@ -288,11 +288,12 @@ module KeeperKeepalive = struct
 
       When [MASC_KEEPER_OAS_TIMEOUT_SEC] is set, that value is used directly.
       Otherwise, {!oas_timeout_for_context} computes an adaptive timeout:
-        base + ctx/1K × per_1k + min(oas_max_turns_per_call × 4, 40) × per_turn
+        base + ctx/1K × per_1k + min(max_turns, 40) × per_turn
       References {!oas_max_turns_per_call} (default 15) directly to avoid
       default drift.  Previous formula used 4× headroom on a default of 5,
-      producing min(20, 40)=20 effective turns.  With default 15, the 4×
-      multiplier would always saturate the cap, erasing context differentiation.
+      producing min(5, 40)=5 effective turns.  Using the actual per-call
+      turn budget keeps scheduled-autonomous and reactive channels aligned
+      with the timeout heuristic.
 
       At 262K context, 15 turns/call:
         120 + 393 + min(15,40)×30 = 120+393+450 = 963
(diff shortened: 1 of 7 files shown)


jeong-sik:
Cross-model review evidence

- Reviewer: direct `sb glm-text` (`GLM-5.1` path)
- Timestamp: 2026-04-13 14:21 KST
- Scope: correctness, regressions, missing safety checks only
- Result: `No findings.`

jeong-sik:
Refreshed cross-model review evidence after adding the FIFO fairness queue

- Reviewer: direct `sb glm-text` (`GLM-5.1` path)
- Timestamp: 2026-04-13 14:39 KST
- Scope: correctness, regressions, missing safety checks only
- Result: reviewer raised 2 cancellation concerns around the new FIFO wait queue

Manual triage of those review notes:
- `with_autonomous_wait_queue` uses `Fun.protect` only to guarantee `Mutex.unlock`; it does not swallow exceptions, including `Eio.Cancel.Cancelled`
- `run_keepalive_unified_turn` already treats `Eio.Cancel.Cancelled` as a first-class cancellation path and re-raises it rather than counting it as a keeper failure

Conclusion: no actionable correctness findings from the external review after code-level verification.

PR #6888, opened 2026-04-13
fix(ci): pin OAS agent_sdk from source before opam install

jeong-sik:
## Summary
- CI is broken on main since #6884: `opam install` fails with "Package conflict" because `agent_sdk >= 0.131.0` is not in the opam registry
- Adds a pin step to `.github/actions/install-ocaml-deps/action.yml` that sources `scripts/oas-agent-sdk-pin.sh` and runs `opam pin add --no-action` before dependency installation

## Test plan
- [ ] CI Build and Test passes (this PR unblocks all other PRs)
- [ ] Lint, Health, Documentation also pass


jeong-sik:
Superseded by #6893. The OAS pin/doc truth fix landed there, so this branch is no longer the merge path.

PR #6893, opened 2026-04-13
fix(deps): pin OAS SDK to synced opam commit

jeong-sik:
Summary
- repoint the OAS runtime pin to the upstream commit that fixes stale agent_sdk.opam version metadata
- keep the dependency floor at agent_sdk >= 0.131.0
- regenerate the generated OAS pin doc blocks
- align ROADMAP front-door version truth with dune-project and CHANGELOG

Verification
- bash scripts/check-version-truth.sh
- bash scripts/sync-oas-pin-docs.sh --check
- git diff --check
- note: bash scripts/check-oas-pin.sh --local-only still fails in my local switch because installed agent_sdk is 0.129.0 before reinstall; CI should exercise the real install path

Linked
- upstream OAS fix PR: jeong-sik/oas#884
- downstream blocker observed on: jeong-sik/masc-mcp#6887


jeong-sik:
Cross-model review attempt (direct ZAI `sb glm-text`, 2026-04-13 KST) timed out twice against the small pin/docs diff, so no external findings were produced. Local verification completed: `bash scripts/check-version-truth.sh`, `bash scripts/sync-oas-pin-docs.sh --check`, `git diff --check`. Upstream packaging fix merged as jeong-sik/oas#884.

PR #6894, opened 2026-04-13
[codex] Align agent_sdk floor with pinned SDK metadata

jeong-sik:
## Summary
- lower the `agent_sdk` dependency floor to `0.130.0`, which matches the version declared by the currently pinned SDK source
- keep the runtime pin on `main@v0.131.0` while correcting the declared base version metadata to `v0.130.0`
- sync generated OAS pin docs to the new floor/base metadata

## Why
CI currently fails during `Install system and OCaml dependencies` because `opam` cannot solve `agent_sdk >= 0.131.0` from the pinned OAS source. The pinned source at `v0.131.0` still declares package version `0.130.0` in `agent_sdk.opam`, so the floor and pin metadata drifted apart.

## Verification
- `gh run view 24340335617 --repo jeong-sik/masc-mcp --log-failed`
  - failing snippet: `deps-of-masc_mcp -> agent_sdk >= 0.131.0` / `no matching version`
- local source check:
  - `/Users/dancer/me/workspace/yousleepwhen/oas` HEAD/tag `v0.131.0`
  - `/Users/dancer/me/workspace/yousleepwhen/oas/agent_sdk.opam` declares `version: "0.130.0"`
- `bash scripts/sync-oas-pin-docs.sh`

## Notes
- `bash scripts/check-oas-pin.sh --local-only` still fails in the current local switch because the installed `agent_sdk` is stale at `0.129.0`; that is local environment state, not a contradiction in this patch.


jeong-sik:
Superseded by #6893, which landed the OAS SDK pin fix on main.

PR #6942, opened 2026-04-13
feat(backend): add Pg_utils shared module (task-011)

jeong-sik:
## Summary

- extract shared PostgreSQL helpers into `Pg_utils`
- centralize the pool type alias, `configured_pool_size`, and `uri_with_keepalive`
- unify the default pool size behavior at 10

## Product impact

Backend PostgreSQL setup now shares one helper surface instead of repeating config and URI assembly logic. The functional change is the default-pool-size fix to a single default of 10.

## Evidence

- `dune build --root .` passes.
- The helper owns both pool-size defaulting and keepalive URI composition in one place.

## Review evidence

- `Pg_utils` only centralizes existing behavior; callers import helpers instead of duplicating the same setup logic.
- `configured_pool_size` removes the previous default mismatch by using one default of 10.
- `uri_with_keepalive` preserves the existing connection semantics while deleting repeated string assembly.

## Linked issue

Refs #6960

## Test plan

- [x] `dune build --root .` passes
- [ ] CI

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -38,6 +38,8 @@
   (cmdliner (>= 1.1))
   (cohttp (>= 5.0))
   (cohttp-eio (>= 6.0))
+  (caqti (>= 2.0))
+  (caqti-eio (>= 2.0))
   (agent_sdk (>= 0.132.0))
   (uri (>= 4.4))
   (tls (>= 2.0))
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing this draft: the PG backend path is already retired, so this Pg_utils extraction is no longer a valid integration target. I am dropping the PR instead of patching review comments for a subsystem we do not want to revive.

PR #6971, opened 2026-04-13
feat(backend): add Pg_utils shared module (task-011 resubmit)

jeong-sik:
Resubmit of PR #6942. CI all green (Build/Test, Health, Lint pass). Added caqti/caqti-eio deps to dune-project.


jeong-sik:
Obsoleted by #6980 — PG backend is being fully purged from masc-mcp (not used, no Caqti dependency, all runtime paths already filesystem-only). No need for a Pg_utils shared module.

PR #6987, opened 2026-04-13
fix(keeper_registry): serialize registry writers with Eio.Mutex

jeong-sik:
## Problem

The module docstring claims entries are *"updated atomically via update_entry/put_entry"* — the intent was there but nothing actually serialized the read-modify-write:

\`\`\`ocaml
let put_entry key entry =
  registry := StringMap.add key entry !registry  (* RMW race *)

let update_entry ~base_path name f =
  match StringMap.find_opt key !registry with    (* read old snapshot *)
  | Some entry -> put_entry key (f entry)        (* write based on it *)
\`\`\`

Two fibers calling \`update_entry\` on the same keeper key can both read the same snapshot, apply their \`f\`, and race on the write — one update silently overwrites the other. With 17+ \`update_entry\` call sites across the module (board cursor, tool usage, crash log, error tracking), this affects a lot of keeper state.

## Fix

Introduce \`registry_mutex : Eio.Mutex.t\` and wrap all four writer paths:
- \`put_entry\` — add/overwrite
- \`update_entry\` — inline the write rather than calling put_entry (Eio.Mutex is not reentrant)
- \`unregister\` — remove
- \`clear\` — reset

Readers still dereference \`!registry\` lock-free: \`StringMap\` is immutable, so a reader always sees either the pre- or post-update snapshot, never a torn value.

## Out of scope

\`register_with_state\` has its own RMW window around \`running_count_atomic\` (pre-read + decrement outside mutex). That's fixed in #6986 — touching this file from both PRs would be a conflict.

## Test plan
- [x] \`dune build --root .\` passes (full lib + test)
- [ ] CI green

### DIFF
--- a/lib/keeper/keeper_registry.ml
+++ b/lib/keeper/keeper_registry.ml
@@ -137,14 +137,19 @@ let decr_running_count_clamped () =
     Readers still dereference [!registry] without the lock — [StringMap] is
     immutable, so a reader always sees a consistent snapshot (either the
     pre- or post-update map). *)
-let registry_mutex = Eio.Mutex.create ()
+(* Stdlib.Mutex is required (not Eio.Mutex): registry helpers are called
+   from non-Eio contexts (tests, startup wiring) where Eio.Mutex fails with
+   `Effect.Unhandled(Cancel.Get_context)`. Critical sections here are short
+   StringMap.add/remove ops, so Stdlib blocking is acceptable.
+   See memory/feedback_ocaml5-mutex-selection.md. *)
+let registry_mutex = Stdlib.Mutex.create ()
 
 
 let registry_key ~base_path name =
   base_path ^ "\x1f" ^ name
 
 let put_entry key entry =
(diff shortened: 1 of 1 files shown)


jeong-sik:
이 브랜치는 심하게 stale 상태로 40개 파일/1500줄이 변경되어 있으며, 최근 merge된 #6974(prometheus-metrics), #7014(IA labels), #7018(memory-subsystems) 등을 **삭제**하는 diff를 포함합니다.

CI 실패 원인은 이 PR이 실제로 수정하려는 mutex 코드가 아니라, stale 브랜치와 main 간 거대한 충돌로 보입니다. 추천:

1. 이 PR을 close
2. `git checkout -b fix/fix/registry-mutex origin/main`로 최신 main에서 fresh 브랜치 생성
3. mutex 관련 실제 변경만 cherry-pick 또는 재작성

원본 의도(mutex 추가)는 올바른 방향이나 stale 브랜치 위에서 fix 시도는 리그레션 위험이 큽니다.

PR #6993, opened 2026-04-13
fix(eio): don't swallow Cancelled in 3 more catch-all sites

jeong-sik:
## Summary

Extends the Cancelled re-raise sweep alongside #6988 (keeper_exec_github) and #6990 (server_dashboard_http + tool_board). Three remaining \`with exn ->\` sites were silently swallowing \`Eio.Cancel.Cancelled\`, breaking fiber cancellation semantics.

## Sites fixed

| File:line | Impact |
|-----------|--------|
| \`keeper_toml_loader.ml:507\` (atomic_write_file) | Fiber cancel during save/rename converted to \`Error\` — caller can't distinguish cancel from genuine I/O failure |
| \`keeper_agent_run.ml:1781\` (episode_create flush) | Cancelled turn continues post-run steps (compaction, metrics) as if nothing happened |
| \`keeper_agent_run.ml:1794\` (memory bank compaction) | Same pattern, same impact |

## Sites NOT fixed (verified safe via structural reading)

| File:line | Why safe |
|-----------|----------|
| \`file_lock_eio.ml:103\`, \`:139\` | Handler ends with \`raise exn\` |
| \`spawn.ml:334\` | Handler ends with \`raise exn\` |
| \`tool_command_plane_support.ml:234\` | Handler ends with \`raise exn\` |

## Why this is a real bug (not a theoretical race)

Keeper turns can be cancelled mid-flight when the controlling switch ends (shutdown, timeout, explicit cancel). With Cancelled swallowed at the three fixed sites:
- \`atomic_write_file\`: returns \`Error\` → caller might retry the write or continue with stale config
- \`episode_create\`: warn log, turn proceeds to compaction → compaction runs on cancelled keeper state
- \`compaction\`: warn log, turn proceeds to metrics → metrics recorded for a turn that was cancelled

## Verification

- [x] \`dune build @all\` succeeds

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -1778,7 +1778,9 @@ let run_turn
                    "keeper:%s post-run flush episodes=%d procedures=%d"
                    meta.name ep pr
              | None -> ())
-           with exn ->
+           with
+           | Eio.Cancel.Cancelled _ as e -> raise e
+           | exn ->
              Log.Keeper.warn "keeper:%s episode_create failed: %s"
                meta.name (Printexc.to_string exn));
           (* Memory bank compaction: dedup + consolidate if over threshold. *)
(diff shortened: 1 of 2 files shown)


jeong-sik:
Duplicate of #6991. That PR covers the same 3 sites (keeper_agent_run.ml:1781, :1794; keeper_toml_loader.ml:507) with identical fix semantics, plus an additional site (keeper_exec_github.ml:952) that I missed. Closing in favor of #6991. My apologies for the duplication — failed to re-check open PR list between scan and fix. Deleting branch.

PR #7037, opened 2026-04-14
fix(provider_adapter): remove Gemini_cli/Codex_cli after OAS 0.134.0

jeong-sik:
URGENT: main CI is broken. OAS 0.134.0 (#7012) removed `Gemini_cli` and `Codex_cli` from `Llm_provider.Provider_config.provider_kind`. `provider_adapter.ml` still referenced both in `string_of_provider_kind` and `auth_env_keys_of_provider_kind`. 4 lines removed. Local build passes.

### DIFF
--- a/lib/provider_adapter.ml
+++ b/lib/provider_adapter.ml
@@ -127,10 +127,8 @@ let string_of_provider_kind
   | OpenAI_compat -> cn_codex
   | Ollama -> cn_ollama
   | Gemini -> cn_gemini
-  | Gemini_cli -> cn_gemini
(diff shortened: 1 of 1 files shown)


jeong-sik:
Incorrect fix — OAS 0.134.0 actually contains Gemini_cli/Codex_cli constructors (re-added @ 0.133.0). My earlier local build error was from stale opam switch state, not an actual OAS regression. Main build is actually fine — other PRs' Build and Test failures have unrelated root causes. Closing without merge.

PR #7060, opened 2026-04-14
feat(dashboard): FSM Hub — composite lifecycle audit surface

jeong-sik:
## Summary

Phase 3 of `docs/design/dashboard-fsm-redesign.md`. Adds an `fsm-hub` section under the **monitoring** tab that consumes the composite observer payload (#7058) and renders a cross-FSM architecture audit view.

## Why under `monitoring` instead of a new top-level tab

Adding a new top-level tab churns nav metadata across three surfaces and dilutes the operator console. `monitoring` already hosts architecture views like `memory-subsystems`, making it the natural home for the FSM Hub without inflating the nav graph. A dedicated top-level surface is tracked as a follow-up only if telemetry shows unmet demand.

## Panels

- **Keeper picker** — all registered keepers (from the `keepers` store).
- **5 sub-FSM cards** — KSM phase / KTC turn phase / KDP decision stage / KCL cascade state / KMC compaction stage.
- **Shared measurement card** — reflect/plan/compact/handoff/guardrail flags + `goal_drift`.
- **Safety invariants panel** — all five checks from `specs/keeper-state-machine/KeeperCompositeLifecycle.tla`, green/violation badge per row.
- **Recovery two-store panel** — RFC-0003 §8 `data_record` / `fsm_condition` with clean/pending/drift classification.
- **Snapshot metadata row** — `correlation_id` / `run_id` / `ts`.

## Data flow

Polls `/api/v1/keepers/:name/composite` every 5s. Phase 4 follow-up converts this to an SSE stream (`keeper.composite.tick` topic) so multiple subscribers don't stampede the registry.

## Test plan

- [ ] `pnpm typecheck` green.
- [ ] Visual check: monitoring → FSM 허브 → pick a keeper → 5 sub-FSM cards + invariants panel render.
- [ ] Invariants row turns red when composite observer returns `phase_turn_alignment=false` (manual test: freeze keeper in `Compacting` phase with `compaction_active=false`).

## Related

- #7058 — `/api/v1/keepers/:name/composite` endpoint (prerequisite).
- #7057 — `registry_entry.last_auto_rules` field (unblocks `shared_measurement` real values).
- `docs/design/dashboard-fsm-redesign.md` Phase 3.
- RFC-0003 §7 — payload shape this page consumes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7071, opened 2026-04-14
feat(dashboard): FSM Hub compound Cytoscape view

jeong-sik:
## Summary

Phase 3 §2 of `docs/design/dashboard-fsm-redesign.md`. Replaces the MVP cards grid with a Cytoscape compound graph rendering all five sub-FSMs in one frame.

## Design

- Each sub-FSM → one parent cluster: KSM (lifecycle), KTC (turn cycle), KDP (decision pipeline), KCL (cascade state), KMC (memory compaction).
- Each cluster contains its possible states from `KeeperCompositeLifecycle.tla`. The current state is marked `active` (warn for KMC, err for `gate_rejected` / `exhausted`); others `dim`.
- **No inter-cluster edges.** Cross-FSM causality lives in the spec (join actions in `KeeperCompositeLifecycle.tla`) and the invariants panel. Drawing cross-cluster arrows here would overstate coupling between layers.
- KSM reduced to 9 most-operationally-relevant phases (skipping rare buffer transitions) so the graph stays legible at 360px.

## Kept

The original card grid remains as a textual complement under the graph — useful at narrow widths and for at-a-glance reads.

## Depends on

- #7060 (FSM Hub page) and #7070 (SSE dogfood) merged in as ancestors.

## Test plan

- [ ] `pnpm typecheck` green.
- [ ] Visual: FSM Hub → select a keeper → compound graph renders; current state in each cluster is highlighted; other states dim. Pan/zoom works via Cytoscape.
- [ ] Resize window to narrow → graph shrinks; card grid still readable underneath.

## Related

- `docs/design/dashboard-fsm-redesign.md` Phase 3 §2 — compound view target.
- `specs/keeper-state-machine/KeeperCompositeLifecycle.tla` — source of state sets.
- `dashboard/src/components/keeper-fsm-specs.ts` — existing spec builders (Phase / Decision Pipeline / Cascade) that this extends.

### DIFF
--- a/dashboard/src/components/fsm-hub.ts
+++ b/dashboard/src/components/fsm-hub.ts
@@ -9,6 +9,8 @@ import {
 import { keepers } from '../store'
 import { compositeTick } from '../composite-signals'
 import { EmptyState } from './common/empty-state'
+import { CytoscapeFsm } from './common/cytoscape-fsm'
+import { buildCompositeFsmSpec } from './keeper-fsm-specs'
 
 /**
  * FSM Hub — architecture audit surface for the composite keeper lifecycle.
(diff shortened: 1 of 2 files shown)


jeong-sik:
Needs rebase against main — `dashboard/src/components/fsm-hub.ts` has an add/add conflict after #7070 landed a new SSE-driven version. Cytoscape approach may or may not be compatible; author's call. Auto-merge is enabled so rebasing should be enough.

Issue #7075, opened 2026-04-14
chore(keeper): remove boring_tools API surface (full source-level cleanup)


jeong-sik:
Follow-up to #7073 which **neutralized** the boring-tool classification (\`boring_tools = []\` + identity pruning) but kept the API surface for zero-churn reasons. This issue tracks full source-level removal.

## Remove
- \`boring_tools\`, \`is_boring_tool\`, \`prune_boring_tools_for_actionable_turn\` in \`lib/keeper/keeper_tool_registry.ml\`
- \`prune_boring_tools_after_recent_polling\` in \`lib/keeper/keeper_tool_disclosure.ml\`
- \`is_cross_turn_polling_tool\`, \`should_block_cross_turn_polling\`, \`recent_tool_streak_count\` in \`lib/keeper/keeper_hooks_oas.ml\` (no-ops once classification is empty)
- \`prune_boring_tools_for_actionable_turn\` entry in \`lib/keeper/keeper_exec_tools.mli\`

## Update call-sites
\`keeper_hooks_oas.ml\`, \`keeper_unified_turn.ml\`, \`keeper_tool_diversity.ml\` — drop boring filters that are now no-ops. Tests: remove retired \`boring_tools\` suite.

## Why after #7073 lands, not before
Neutralization is reversible with a one-line change; full removal is not. Let the neutralization bake (a few keeper cycles) to confirm silent-loop reduction before deleting the infrastructure.

## Acceptance
- \`rg 'boring' lib/ test/\` returns nothing under \`lib/\`
- Full keeper test suite green
- Staging keeper observations show no \`keeper_stay_silent\`-only turns attributable to the retired classification.

jeong-sik:
Consolidated into umbrella issue #7171. Closing this in favor of the umbrella.

PR #7081, opened 2026-04-14
[codex] Remove boring tool suppression

jeong-sik:
## Summary
- remove the remaining keeper boring-tool shim API and pruning helpers instead of leaving them as dead compatibility paths
- delete the cross-turn polling gate so status-style tools are no longer treated as a special silent-suppression class
- keep tool-only turns visible by synthesizing a reply instead of collapsing them into `silent`

## Why
- `#7073` neutralized the boring classifier, but left dead APIs and tests behind
- that leftover surface kept the runtime and test suite harder to reason about, and it obscured why productive tool turns still looked silent

## Product impact
- keepers stop hiding productive tool-only turns behind silent outcomes
- operator/debugging behavior becomes easier to interpret because the boring/polling special case is fully removed rather than half-neutralized

## Evidence
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat/ctx-breakdown-panel`
- `./_build/default/test/test_keeper_topk_llm.exe`

## Review evidence
- rebased onto current `main` so the PR no longer reintroduces already-merged `#7062` CTX composition changes
- updated the stale review note about `keeper_agent_run.ml` / `keeper_registry.ml` drift after the rebase
- checked thread-aware review data for `#7081`; there are no unresolved inline review threads, only top-level PR hygiene feedback

## Linked issue
- Refs #7073
- Refs #7085
- Refs #7075
- Refs #7087

### DIFF
--- a/lib/keeper/keeper_exec_tools.mli
+++ b/lib/keeper/keeper_exec_tools.mli
@@ -70,11 +70,6 @@ val injected_masc_tool_names : unit -> string list
 (** [is_core_always_tool name] — true if [name] bypasses policy restrictions. *)
 val is_core_always_tool : string -> bool
 
-(** Drop boring observation tools when a turn already has actionable work.
-    Returns the original set when every tool is boring so callers do not end
-    up with an empty allowlist. *)
-val prune_boring_tools_for_actionable_turn : string list -> string list
-
 (** Deduplicate tool names, preserving order. *)
 val dedupe_tool_names : string list -> string list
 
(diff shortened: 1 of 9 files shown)


jeong-sik:
Subsumes #7073 (neutralize boring classifier) and #7085 (retire cross_turn_polling gate). Both closed. #7087 (drop ANTI-POLLING prompt directive) is orthogonal — `keeper_prompt.ml` only, non-overlapping. Rebased onto current `main`; the stale `keeper_agent_run.ml` / `keeper_registry.ml` drift from #7066/#7070 is gone. #7075 (API surface cleanup) becomes redundant if this lands; will close when merged.

Issue #7078, opened 2026-04-14
audit(keeper): catalog of OAS/LLM workarounds that belong upstream


jeong-sik:
Follow-up to #7075 (boring removal). While neutralizing \`boring_tools\`, I scanned \`lib/keeper/\` for structurally-similar workarounds — each of which compensates for a missing capability in OAS (or the underlying LLM) by adding category-level classifiers + thresholds + graduated responses in keeper code. Listed here so we can track the upstream fix for each.

## Common signature

\`\`\`
category-level classifier (hardcoded tool/error list)
  + threshold (consecutive/streak count)
  + graduated response (warn → final-warn → skip/block)
  + ref-cell state or per-keeper Hashtbl
\`\`\`

Each entry reproduces roughly the same structure. The duplication is *not* accidental — it's the shape that compensates for a missing OAS signal.

## Cases

| # | Workaround | Location | Compensates for | Upstream fix (OAS) |
|---|------------|----------|-----------------|--------------------|
| 1 | \`boring_tools\` classifier (retired in #7073) | \`keeper_tool_registry.ml\` | \`is_idle\` only matches exact \`{name, input}\` fingerprint; alternating status+heartbeat bypasses idle | Coarse-grain idle to tool-name granularity (opt-in) |
| 2 | \`tool_name_streak\` same-name gate (5x → Override) | \`keeper_hooks_oas.ml:297\` | Args-level idle detection misses same-tool repetition with changing args | Same as above |
| 3 | \`cross_turn_polling_gate\` | \`keeper_hooks_oas.ml:495\` | No cross-turn idle continuity across run_turn boundaries | Persist idle state across turns in agent lifecycle |
| 4 | \`on_idle_decision\` graduated nudges (warn/final/skip) | \`keeper_hooks_oas.ml:201\` | OAS exposes idle_count but no callback-based graduated intervention | \`on_idle_escalated\` hook with severity level |
| 5 | \`keeper_failure_circuit_breaker\` with error_class | \`keeper_failure_circuit_breaker.ml\` | OAS retries transient errors blind, no error-class awareness | Error taxonomy + class-aware retry policy |
| 6 | \`keeper_tool_diversity\` Shannon-entropy hint injection | \`keeper_tool_diversity.ml\` | No exploration term in LLM sampling; purely exploitative tool selection | Exploration bonus at sampling layer, or RL-style intrinsic reward |
[...]

PR #7085, opened 2026-04-14
refactor(keeper): retire cross_turn_polling_gate call-site (#7078 / 2/8)

jeong-sik:
## Summary

Part 2/8 of the #7078 workaround catalog. Removes the cross-turn polling gate call-site in \`keeper_hooks_oas.ml\`.

## What the gate did

In \`pre_tool_use\`, after the same-name streak check, it also read the recent tool-call log (last 8 entries, 900s window) and blocked the call if:
- \`is_cross_turn_polling_tool tool_name\` (== boring and not stay_silent), AND
- recent streak + 1 >= 2

It was a second defensive layer behind the \`boring_tools\` classification, trying to catch cross-turn polling that escapes the single-turn streak counter.

## Why retire it now

1. **Same-name streak_gate (5x)** already covers the primary case from within a single turn — running \`masc_status\` 5 times in a row is still blocked.
2. **Boring classifier retirement (#7073)** makes \`is_cross_turn_polling_tool\` constant-false, so the gate is no-op after #7073 lands anyway.
3. The gate's stated purpose — compensating for OAS not persisting idle state across turns — is tracked upstream in #7078. Local duplication adds no value.

## Scope

- Deletes the \`else if should_block_cross_turn_polling ...\` branch from the pre_tool_use gate chain
- Drops the unused \`recent_entries\` / \`cross_turn_streak\` locals  
- Keeps the internal helpers (\`is_cross_turn_polling_tool\`, \`recent_tool_streak_count\`, \`should_block_cross_turn_polling\`) defined so existing tests keep compiling. They become dead code; full removal is tracked by #7075.

## What does NOT change

- Same-name streak gate (5x → Override) — still active
- Denied-tool list
- Idle nudges  
- Boring classifier (handled separately in #7073)

## Test plan

- [x] \`dune build --root . lib\` — 0 errors
- [x] \`test/test_keeper_unified.exe\` — 156/156 pass

Related: #7073 (boring retirement), #7075 (API surface removal), #7078 (catalog)

### DIFF
--- a/lib/keeper/keeper_hooks_oas.ml
+++ b/lib/keeper/keeper_hooks_oas.ml
@@ -296,7 +296,6 @@ let make_hooks
      At >= streak_threshold, pre_tool_use blocks the call with Override. *)
   let tool_name_streak : (string * int) ref = ref ("", 0) in
   let streak_threshold = 5 in
-  let cross_turn_polling_threshold = 2 in
   { Agent_sdk.Hooks.empty with
 
     after_turn = Some (fun event ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #7081 which also removes cross_turn_polling_gate at the full source level. Closing to consolidate under #7081.

PR #7162, opened 2026-04-14
feat: add keeper accountability reputation ledger

jeong-sik:
## Summary
- add an evidence-first keeper accountability ledger under `.masc/accountability`
- record keeper task transitions and explicit completion claims, then derive private risk/routing signals
- expose accountability in the keeper trust observatory and add a claim-time routing warning for high-risk keepers
- keep existing public reputation and karma behavior unchanged

## Testing
- `dune build --root . @install`
- `dune exec --root . ./test/test_keeper_accountability.exe`
- `dune exec --root . ./test/test_keeper_task_dispatch.exe`
- `dune exec --root . ./test/test_keeper_unified.exe`

### DIFF
--- a/lib/keeper/keeper_context_core.ml
+++ b/lib/keeper/keeper_context_core.ml
@@ -223,6 +223,75 @@ let message_of_json (json : Yojson.Safe.t) : Agent_sdk.Types.message =
          |> Option.map Inference_utils.sanitize_text_utf8);
     }
 
+let tool_use_ids_of_message (msg : Agent_sdk.Types.message) : string list =
+  List.filter_map
+    (function
+      | Agent_sdk.Types.ToolUse { id; _ } -> Some id
+      | _ -> None)
+    msg.content
+
+let has_tool_result_block (msg : Agent_sdk.Types.message) : bool =
+  List.exists
+    (function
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
Strict review findings:

1. High: synthetic `task_done` successes are currently counted in the same completion-claim pool as explicit keeper claims. That means unsupported explicit claims can be diluted back down by unrelated successful task completions, which weakens the intended risk signal.

2. High: the keeper trust observatory now computes accountability by scanning/parsing the 14-day JSONL range per keeper on the dashboard path. Given the existing dashboard timeout pressure, this adds an O(keepers * ledger_range) hot-path cost without caching.

3. Medium: `keeper_task_claim` computes `routing_warning` only after `Room.claim_next_r` has already mutated state. As written, the soft routing hint is visible, but not actionable on the main claim path.

Follow-up work was split into MASC tasks `task-069` through `task-074`, and broadcast to `@keeper-sangsu-agent`.

jeong-sik:
Closing as superseded. The accountability ledger feature already landed in #7191 (commit 0cc1496ba), and the remaining checkpoint-pairing fix is being tracked separately in #7206. Keeping this PR open would duplicate or misrepresent the actual delta.

PR #7229, opened 2026-04-15
feat(dashboard): add spec-aware FSM diagnostics

jeong-sik:
## Summary

- add an Operator Meaning panel that interprets the current composite snapshot instead of only listing raw FSM states
- derive per-lane observed dwell and transition summaries for KSM/KTC/KDP/KCL/KMC
- make invariant rows explain what each contract means and why a failure matters
- add focused tests for spec-drift, idle snapshots, and on-screen stall detection

## Product impact

- FSM Hub now answers what is wrong, what should happen next, and which lane is not moving
- the surface becomes useful for operator diagnosis instead of only decorative state display
- invariant failures are explained in runtime terms, not just boolean badges

## Evidence

- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/vitest run src/components/fsm-hub.test.ts`

## Review evidence

- follow-up review requested specifically on whether the new diagnostics meaningfully use RFC-0003 semantics

## Linked issue

- follow-up to PR #7208 and PR #7223

### DIFF
--- a/dashboard/src/components/fsm-hub.test.ts
+++ b/dashboard/src/components/fsm-hub.test.ts
@@ -169,7 +169,24 @@ describe('fsm-hub derived state', () => {
 
     expect(lanes.find(lane => lane.field === 'KTC')).toMatchObject({
       tone: 'warn',
+      stalled: true,
       value: 'executing',
     })
+  })
+
+  it('keeps active compaction as compaction work before the stall threshold', () => {
+    const result = deriveOperationalInsight(
+      snapshot({
+        is_live: true,
+        phase: 'Compacting',
+        turn_phase: 'compacting',
+        compaction: { stage: 'compacting' },
+      }),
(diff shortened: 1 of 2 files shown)


jeong-sik:
Local verification for this PR:

- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/vitest run src/components/fsm-hub.test.ts` (7 tests passed)
- `./node_modules/.bin/vite build`
- live render verification against the local backend on `127.0.0.1:8935`
  - mounted `FsmHub`
  - fetched a real `/api/v1/keepers/:name/composite` payload
  - verified the DOM renders `Operator Meaning`, `Next:`, transition empty/history state, and the live phase value

Follow-up fix included while verifying:
- avoid false stall diagnostics before the configured stall threshold
- status-bar observation count now uses the real observation stream length instead of the phase-log length

GitHub checks are not currently attached to this branch, so the verification evidence above is the current gate.

PR #7227, opened 2026-04-15
[codex] improve keeper claim-first prompt guidance

jeong-sik:
## Summary
Bias keepers toward `keeper_task_claim {}` before backlog browsing when unclaimed work is available and they do not already hold a task.

## What changed
- strengthened the unified system prompt so it explicitly says `keeper_task_claim` takes `{}` and auto-claims the next eligible task
- added a user-prompt `Immediate Task Move` section when unclaimed tasks exist and the keeper has no `current_task_id`
- added prompt regression tests for both the positive case and the already-claimed-task case

## Why
After the orphan `ToolResult` checkpoint fix landed, `nick0cave` no longer hit the provider resume error, but recent decisions still showed a separate low-activity pattern:
- `claim_was_available=true`
- `claim_executed=false`
- several autonomous turns ended as `turn_budget_exhausted(...)` after browsing tools such as `keeper_tasks_list` or `keeper_board_list`

The claim tool already auto-claims the next eligible task with an empty object, but the prompt did not make that obvious enough, so keepers were burning turns on discovery instead of starting work.

## Product impact
- keepers should spend fewer autonomous turns on backlog browsing before claiming work
- personas like `nick0cave` should convert backlog pressure into task claims more reliably
- this reduces one common path where autonomous activity looks low even though the namespace still has work available

## Evidence
- recent `nick0cave` decisions in `/Users/dancer/me/.masc/keepers/nick0cave.decisions.jsonl` showed repeated `claim_was_available=true` with `claim_executed=false`
- local prompt tests now verify that claim-first guidance appears only when unclaimed work exists and no task is already claimed

## Review evidence
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix/keeper-claim-first ./test/test_keeper_unified.exe`

## Linked issue
Refs #7226

PR #7230, opened 2026-04-15
feat(keeper_prompt): claim-first bias when unclaimed tasks exist (#7226)

### LINKED ISSUE #7226: keeper should bias toward keeper_task_claim before backlog browsing
## Summary
Keepers with unclaimed work available still spend autonomous turns on backlog browsing or board reads instead of claiming work.

## Symptoms
- recent `nick0cave` decisions show `claim_was_available=true` and `claim_executed=false`
- several turns end as `turn_budget_exhausted(...)` after `keeper_tasks_list`, `keeper_board_list`, or similar observation tools
- autonomous activity looks low even though the backlog is non-empty

## Repro notes
1. Run a keeper with no `current_task_id`
2. Ensure unclaimed tasks exist in the namespace
3. Observe scheduled autonomous turns in `.masc/keepers/<name>.decisions.jsonl`
4. See repeated browsing or audit turns without `keeper_task_claim`

## Expected
When unclaimed tasks exist and the keeper is not already holding one, the prompt should strongly bias toward `keeper_task_claim {}` as the first productive action.

## Why now
This became visible while validating the `nick0cave` activity drop after the orphan `ToolResult` checkpoint fix. The checkpoint bug is fixed, but claim-first behavior is still too weak.

jeong-sik:
## Summary

Add strong claim-first directive when keeper has no active task but unclaimed tasks exist.

Fixes #7226

## Change

In Namespace State section of keeper unified prompt:
- Before: "Unclaimed tasks: N" (informational only)
- After: adds "ACTION REQUIRED: Call keeper_task_claim as your FIRST action" when current_task_id = None

1 file, 10 lines.

## Test Plan

- [x] Library build passes
- [ ] CI (dashboard_mission_assembly test has pre-existing interface mismatch on main)
- [ ] Observe keeper logs for claim-first behavior

### DIFF
--- a/lib/keeper/keeper_unified_prompt.ml
+++ b/lib/keeper/keeper_unified_prompt.ml
@@ -235,10 +235,19 @@ let build_prompt ~(meta : Keeper_types.keeper_meta) ~(base_path : string)
     || observation.active_agent_count > 0
   then (
     Buffer.add_string ubuf "### Namespace State\n";
-    if observation.unclaimed_task_count > 0 then
+    if observation.unclaimed_task_count > 0 then begin
       Buffer.add_string ubuf
         (Printf.sprintf "- Unclaimed tasks: %d\n"
            observation.unclaimed_task_count);
+      (* Claim-first bias: when unclaimed work exists and the keeper
+         has no active task, emit a strong directive to claim before
+         browsing. Without this, keepers waste turns on board_list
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #7227. Main already includes the stronger claim-first guidance via the separate  section in , so this draft now duplicates a weaker subset of the shipped behavior.

PR #7390, opened 2026-04-15
fix(dashboard): query observatory telemetry by time window

jeong-sik:
## Summary
- query dashboard telemetry by `since_ms` / `until_ms` instead of fetching the latest 500 entries and filtering client-side
- return `total_matching_entries` and `truncated` so Observatory can show the true range total even when marker payloads are capped
- update Observatory to use range-aware fetches and display `showing N` only when the marker payload is truncated
- add telemetry tests for range filtering before limit application and cross-day file reads

## Why
Observatory was undercounting busy windows because it fetched the latest 500 telemetry entries first, then filtered them in the browser. That made `최근 1시간 · 498 events` reflect a query cap instead of the actual 1-hour event volume.

## Notes
- The true event count is no longer capped in the Observatory subtitle for the requested window.
- A marker payload safety budget still exists (`n=5000`) to avoid sending or rendering an unbounded number of DOM markers in one request. The PR makes that truncation explicit instead of silently lying about totals.

## Validation
- `pnpm typecheck` in `dashboard/`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-observatory-telemetry-window _build/default/lib/telemetry_unified.ml`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-observatory-telemetry-window _build/default/lib/server/server_routes_http_routes_dashboard.ml`

## Known blocker
- Full `dune build @default` and `dune exec ./test/test_telemetry_unified.exe` are currently blocked by a pre-existing compile error in `lib/keeper/keeper_agent_run.ml:1218` on this branch base (`?config_path` passed to a provider-based rerank function). This PR does not touch that path.


jeong-sik:
Follow-up update after the original PR body:

- Observatory no longer sends a marker payload cap for range queries.
- `/api/v1/dashboard/telemetry` now treats time-window requests without `n` as unbounded (`n=0` semantics internally), while preserving the old default limit for non-window consumers.
- The track rendering path is now width-aware and buckets raw events into marker buckets on the client, so the UI does not need a hardcoded event-count cap to stay usable.
- Added `observatory-utils.test.ts` plus an OCaml test for `n=0` disabling truncation in time-window reads.

So the remaining numeric behavior is now rendering-density projection, not data truncation.

PR #7379, opened 2026-04-15
[codex] Split keeper social-model registry modules

jeong-sik:
## Summary
- extract the current `bdi_speech_v1` transition core behind an explicit social-model dispatch boundary
- split social-model code into shared types/protocol helpers, a `bdi_speech_v1` implementation module, a registry module, and a thin `Keeper_social_model` facade
- normalize unknown `social_model` values to the active baseline before exposing them via keeper status, dashboard, and tool surfaces
- add regression coverage for model-id normalization and unknown-model fallback in unified social-model tests

## Why
`social_model` had a design contract and an internal dispatch boundary, but the implementation was still physically collapsed into one file. This PR makes the first implementation boundary real so follow-up work can add state carry and additional models without mixing facade, registry, protocol parsing, and BDI behavior in one module.

## Validation
- `env ALCOTEST_QUICK_TESTS=1 dune exec --build-dir /tmp/masc-social-model-fsm-build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/feat/social-model-fsm-registry ./test/test_keeper_unified.exe`
  - result: `150 tests run`, `Test Successful`

## Notes
- I added a keeper-list surface normalization test in `test_keeper_meta_listing.ml`, but running that target currently hits an unrelated compile error in `lib/keeper/keeper_agent_run.ml` around `Agent_sdk.Tool_selector.default_rerank_fn ?config_path`.
- I did not patch that call in this PR because it is outside the social-model refactor slice; the failure appears pre-existing relative to these changes.

PR #7392, opened 2026-04-15
fix(cascade): replace List.hd/List.tl fallback with pattern match

jeong-sik:
## Summary

Replace the unreachable `List.hd entries, List.tl entries` fallback in `weighted_shuffle.find_selected` with a pattern match on `entries`. No behavior change.

## Why

`scripts/health_snapshot.sh` treats any `List.hd` / `List.tl` occurrence in `lib/` as a ratchet regression — the baseline pins `lib_list_hd=0` and `lib_list_tl=0`. Commits #7382 / #7386 (cascade scaffold) introduced these calls in the fallback arm of `find_selected`. Although the outer `match entries with [] | [_] -> entries` guard makes the fallback unreachable, the text-based ratchet in `health_snapshot.sh` can't see that distinction, so CI `Health` job fails for any downstream PR that gets the `refs/pull/*/merge` commit.

PR #7388 is one such downstream PR — its `Health` check reports `Ratchet: fail (lib_list_hd 0->1; lib_list_tl 0->1)` even though that PR touches only dashboard TypeScript. This fix removes the ratchet regression at the source.

The replacement uses `match entries with | first :: rest -> (first, rest) | [] -> assert false`. The `[]` arm is genuinely unreachable because of the outer guard; an `assert false` documents that invariant without adding new unsafe primitives.

## Test plan

- [x] `dune build --root .` — clean
- [x] `rg -c 'List\.hd' lib -g '*.{ml,mli}'` → 0, `rg -c 'List\.tl' lib -g '*.{ml,mli}'` → 0
- [ ] CI `Health`, `Build and Test`, `TLA+ Model Checking` green
- [ ] Downstream PR #7388 rebases and passes `Health` without further changes

### DIFF
--- a/lib/cascade/cascade_config.ml
+++ b/lib/cascade/cascade_config.ml
@@ -349,11 +349,14 @@ let weighted_shuffle
     if total_weight <= 0 then entries
     else
       let r = rand_int total_weight in
-      (* Find the selected entry via cumulative weight *)
+      (* Find the selected entry via cumulative weight. The outer [] | [_]
+         guard means `entries` always has >= 2 elements here; the inner `[]`
+         arm is unreachable and asserts as such. *)
       let rec find_selected cumulative = function
-        | [] -> (* fallback: first entry *)
-          (List.hd entries,
-           List.tl entries)
+        | [] ->
+          (match entries with
+           | first :: rest -> (first, rest)
+           | [] -> assert false)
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #7379 — main already applied the same  fix inline. Closing as redundant.

PR #7424, opened 2026-04-15
[codex] Fix code nav prefilter and rg args

jeong-sik:
## Summary
- fix `masc_code_search` ripgrep argument ordering when `file_pattern` is set
- reduce keeper code-nav false positives by gating `masc_code_search`, `masc_code_read`, and `masc_code_symbols` behind explicit code-navigation intent
- add regression coverage for both the ripgrep bug and keeper deterministic prefilter behavior

## Why
`masc_code_search` was assembling `rg` arguments in the wrong order, which produced `rg: 2: No such file or directory` for some keeper calls. Separately, code-navigation tools were surfacing too easily in generic keeper turns, causing avoidable tool-selection noise.

## Impact
- `masc_code_search` now handles `file_pattern` correctly
- keeper deterministic prefilter only surfaces code-navigation tools for explicit code-search/read/symbol intents or code-path hints
- generic non-code queries and generic file-read queries are less likely to pull in code-navigation tools

## Validation
- reproduced the broken and fixed `rg` command shapes directly in shell
- started targeted `dune build` for `test/test_keeper_topk_llm.exe` and `test/test_tool_code_coverage.exe`, but did not wait for completion because the repo-wide dependency graph is large

### DIFF
--- a/lib/keeper/keeper_tool_disclosure.ml
+++ b/lib/keeper/keeper_tool_disclosure.ml
@@ -123,6 +123,82 @@ let tool_query_text_of_user_message (text : string) : string =
   loop None [] lines
 ;;
 
+let contains_any_ci (text : string) (needles : string list) : bool =
+  let haystack = String.lowercase_ascii text in
+  List.exists
+    (fun needle ->
+      let needle = String.lowercase_ascii needle in
+      let hay_len = String.length haystack in
+      let needle_len = String.length needle in
+      let rec loop idx =
+        if needle_len = 0 then true
+        else if idx + needle_len > hay_len then false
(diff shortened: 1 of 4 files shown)


jeong-sik:
CI failure is a real test issue, not infra: `rg` (ripgrep) is not installed on the CI runner. The test at `test/test_tool_code_coverage.ml:147` calls `rg --json ...` which fails with `create_process: No such file or directory`. Fix: add `ripgrep` to the CI system packages (e.g. `os-packages: jq ripgrep` in the install-ocaml-deps action input), or guard the test with a `command -v rg` availability check.

PR #7436, opened 2026-04-15
refactor(alerting_path): replace Hashtbl visited-set with immutable StringSet

jeong-sik:
## Summary
Replace mutable Hashtbl visited-set with immutable StringSet in keeper_alerting_path.ml.

## Changes
- `find_suffix_matches_under_root`: Replaced `(string, unit) Hashtbl.t` visited set with `Set.Make(String)`
- Threaded `SS.t` through recursive `walk` function as a parameter
- `SS.mem` for cycle detection, `SS.add` for marking visited directories
- Call site updated: `walk ~dirs_seen:0 [] root SS.empty`
- Eliminates mutable Hashtbl allocation per traversal

## Verification
- dune build lib passed (exit code 0)
- No behavioral change: same directory traversal with cycle detection

### DIFF
--- a/lib/keeper/keeper_alerting_path.ml
+++ b/lib/keeper/keeper_alerting_path.ml
@@ -68,16 +68,16 @@ let is_within_root_norm ~(root_norm : string) (path : string) : bool =
 let find_suffix_matches_under_root ~root ~anchor ~suffix_rel
     ?(max_dirs = 2000) ?(max_matches = 8) () : string list =
   let root_norm = normalize_path_for_check root |> strip_trailing_slashes in
-  let visited : (string, unit) Hashtbl.t = Hashtbl.create 64 in
-  let rec walk ~dirs_seen acc dir =
+  let module SS = Set.Make (String) in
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by #7421 (merged to main as commit 46fc2d692) — "refactor(memory_recall): replace Hashtbl dedup with immutable StringSet" accomplished the same Hashtbl→StringSet conversion for `find_suffix_matches_under_root`.

The rebase on main conflicts at every block because both branches independently rewrote the same function. Main's version is actually better designed:

- Main: threads `visited` through a 3-tuple `(visited, dirs_seen, acc)` so recursive walks inside `List.fold_left` correctly accumulate the visited set across branches.
- This PR: positional `visited` as last arg, but the `List.fold_left` accumulator is only `(dirs_seen, acc)` — `visited` updates from nested recursive walks are silently dropped when the fold returns. Subtle correctness bug.

No action needed. The refactor's goal is already in main with a correct implementation.

PR #7455, opened 2026-04-15
fix(keeper): add admin tools to Keeper_denied surface (49 Forbidden/3MB)

jeong-sik:
## Symptom

Live log, 2026-04-16 /loop iteration 2 aggregation (~3 MB window):

\`\`\`
tool call failed: masc_tool_grant — timeout=0|duration_ms=4|detail=🚫 Forbidden: keeper-ani1999-agent cannot masc_tool_grant
\`\`\`

**49 occurrences** in a ~3MB log window, #1 subset of the 50-entry \`[ERROR] MCP: tool call failed\` pattern.

## Root cause

Admin-surface tools are CanAdmin-gated at dispatch (\`tool_shard.ml:993\`, \`tool_permission_map.ml:87\`), but their schemas were still exposed to keepers via the default keeper tool surface. Existing \`Keeper_denied\` list only contained \`masc_reset\` + \`masc_spawn\`.

Effect: LLM saw \`masc_tool_grant\` in its tool list, kept trying to call it to grant itself permissions, each call hit the permission gate, produced a Forbidden error. Fail-fast but the LLM re-tries. 49 times per window.

## Fix

Extend \`keeper_denied_surface_tools\` to mirror the tools that \`admin_surface_tools\` owns exclusively:

- \`masc_tool_grant\`, \`masc_tool_revoke\`
- \`masc_tool_admin_update\`, \`masc_tool_admin_snapshot\`
- \`masc_config\`
- \`masc_keeper_create_from_persona\`, \`masc_keeper_reset\`
- \`masc_pause\`, \`masc_resume\`

\`Keeper_tool_policy.keeper_denied_set\` reads this surface at init and filters schemas at discovery (\`keeper_tool_policy.ml:170-188\`). The LLM will no longer see these tools.

## Tools that remain visible to keepers

Intentionally kept:

- \`masc_tool_list\` — \`CanReadState\`, legitimate introspection
- \`masc_tool_help\` — discovery
- \`masc_autoresearch_*\` — autoresearch keepers use these
- \`masc_runtime_ollama_probe\` — keeper runtime diagnostic

## Expected impact

49/window → 0. Will verify on server restart.

## Loop series context

This is PR #3 in the self-paced /loop:
- #7445 (board_post schema drift) — **MERGED**
- #7453 (auto_approve callbacks) — Open, Draft
- **this** (admin tools to Keeper_denied)

Each removes the top fixable WARN/ERROR pattern from the 20-min re-scan.

## Test plan

- [x] \`dune build --root .\` clean
- [ ] \`dune runtest --root .\` — running
- [ ] Deploy + observe: \`masc_tool_grant\` Forbidden errors → 0

### DIFF
--- a/lib/tool_catalog_surfaces.ml
+++ b/lib/tool_catalog_surfaces.ml
@@ -211,7 +211,11 @@ let keeper_denied_surface_tools =
     "masc_tool_admin_snapshot";
     "masc_config";
     "masc_keeper_create_from_persona";
-    "masc_keeper_reset";
+    (* NOTE: masc_keeper_reset is on public_mcp_surface_tools (line 126),
+       so it cannot be added to Keeper_denied without violating the
+       Keeper_denied ∩ Public_mcp = ∅ invariant (test_tool_surface_ssot.ml:90).
+       Keepers don't see the public_mcp surface at discovery, so no filter
+       is needed here. Admin permission still blocks execution. *)
     "masc_pause";
     "masc_resume";
   ]
--- a/lib/tool_catalog_surfaces.ml
+++ b/lib/tool_catalog_surfaces.ml
@@ -197,6 +197,23 @@ let keeper_denied_surface_tools =
   [
     "masc_reset";
     "masc_spawn";
+    (* Admin surface — [CanAdmin]-gated, keepers cannot execute these.
+       Log evidence (2026-04-16 /loop): ~49 Forbidden errors per ~3MB
+       window from keepers (ani1999, etc.) trying to call these. They
+       were already denied at dispatch, but the schemas still appeared
+       in keeper tool lists so the LLM kept hallucinating calls. Listing
+       them here filters schemas out at discovery time, so the model
+       never sees them.
+       Source: admin_surface_tools at tool_catalog_surfaces.ml:182. *)
+    "masc_tool_grant";
+    "masc_tool_revoke";
+    "masc_tool_admin_update";
+    "masc_tool_admin_snapshot";
+    "masc_config";
+    "masc_keeper_create_from_persona";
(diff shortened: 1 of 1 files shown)


jeong-sik:
## CI failure diagnosis (loop iter 12, 2026-04-16)

CI is failing on the surface invariant test:
```
[FAIL] invariants  1   Keeper_denied ∩ Public_mcp ...
FAIL Keeper_denied overlaps with Public_mcp: {masc_keeper_reset}
```

Source: `test/test_tool_surface_ssot.ml:90-96` enforces `Keeper_denied ∩ Public_mcp = ∅`.

`masc_keeper_reset` is in `public_mcp_surface_tools` (`lib/tool_catalog_surfaces.ml:127`), so adding it to `Keeper_denied` violates the design. The matching precedent `masc_reset` is on `System_internal` (not Public_mcp), which is why it can also live in Keeper_denied.

### Recommendation

Remove `masc_keeper_reset` from the Keeper_denied additions in this PR. The reasoning behind adding it (filter schemas at discovery time so the LLM doesn't hallucinate calls) doesn't apply the same way to Public_mcp tools: keepers don't get the Public_mcp surface in the first place, so the schema isn't visible to them.

The remaining additions in this PR (`masc_tool_grant`, `masc_tool_revoke`, `masc_tool_admin_update`, `masc_tool_admin_snapshot`, `masc_config`, `masc_keeper_create_from_persona`, `masc_pause`, `masc_resume`) are all on `admin_surface_tools` only (not Public_mcp), so they're safe to add to Keeper_denied.

If `masc_keeper_reset` schemas really do leak into keeper tool lists, that's a separate plumbing bug — not something Keeper_denied can fix without breaking the disjointness invariant.

PR #7474, opened 2026-04-15
fix(keeper): redirect gh to keeper_shell op=gh instead of blocking silently

jeong-sik:
Closes #7442.

## Linked issue

#7442 — masc-improver trips circuit_breaker on gh command blocked.

## Summary

\`keeper_bash\` allow-list blocks \`gh\` (auth/network boundary), but the block hint did not mention that \`keeper_shell op=\"gh\"\` is the correct alternative. Keepers retried \`keeper_bash cmd=\"gh ...\"\` until the circuit breaker tripped. Fix routes the hint correctly and fixes the masc-improver persona.

## Product impact

- masc-improver (and any other keeper citing \`gh CLI\` in its Needs) stops failing silently into a circuit breaker trip
- Block hint now **names the right tool** instead of pointing only at \`rg/ls/find\`
- Same-file diff is tiny: +14 / +9 across 2 files
- No allow-list loosening — \`gh\` remains blocked in \`keeper_bash\`, just re-routed via \`keeper_shell op=\"gh\"\` which is the designed path

## Evidence

- \`dune build\` from the worktree exits 0 (OCaml type-check green)
- Hint logic guarded by explicit token boundary (\`' '|'\\t'|EOS\` after \`gh\`) so \`ghostscript\` or similar prefixes do not mis-route
- Persona change confined to masc-improver; other souls untouched

## Review evidence

Reviewer focus:
1. Token boundary check in \`starts_with_gh\` — does it handle \`gh\`, \`gh pr list\`, \`gh\\tpr list\`, and reject \`ghostscript\`?
2. SOUL.md Tool Routing section — does it cover the common keeper operations without gaps?
3. Is there another keeper persona with a similar \"gh CLI\" reference that should be updated in a follow-up? (grep suggests this is the only one, but worth confirming.)

## Non-scope

- Allow-list extension (option B in #7442) rejected — \`gh\` has auth + destructive potential; opening keeper_bash to it needs separate security review
- Scoped MCP tools (\`gh_pr_status\` etc., option C) tracked separately if load pattern justifies

## Test plan

- [x] \`dune build\` green
- [ ] CI Gate green
- [ ] Manual: live fleet log shows no \`gh_command_blocked\` or \`keeper_bash blocked: ... gh ...\` warning for masc-improver over a 1h window after deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7486, opened 2026-04-15
fix(keeper_bash): redirect gh commands to keeper_shell op=gh (#7442)

### LINKED ISSUE #7442: [keeper] masc-improver trips circuit_breaker on gh command blocked (keeper_bash)
## Symptom

Observed in live logs on 2026-04-16:

\`\`\`
[WARN] [Keeper] circuit_breaker tripped for masc-improver: 3 consecutive other failures (total trips: 1)
[WARN] [Keeper] tool keeper_bash returned error result (1/3): {\"circuit_breaker\":true,\"ok\":false,\"error\":\"command_blocked\",\"reason\":\"Command blocked: 'gh' is not allowed. Allowed: dune, git, rg, ls, cat, make, node, npm, etc.\",\"hint\":\"Check the command for blocked patterns. Use keeper_shell for structured ops (rg, ls, find).\"}
\`\`\`

## Root cause hypothesis

`keeper_bash` allow-list intentionally excludes `gh` (auth/network boundary). masc-improver's prompt or tool guidance does not steer the keeper toward the available alternative (`keeper_shell`, MCP GraphQL/PR tools), so it retries `gh` until the circuit breaker trips.

## Options

- **(a) Prompt / persona fix**: Update masc-improver persona + tool guidance so it reaches for `keeper_shell` or structured MCP tools for PR/issue work, not `gh` directly. Lowest blast radius.
- **(b) Allow-list extension**: Add `gh` to the allow-list. Requires explicit security review — `gh` carries auth and can perform destructive ops (PR close, branch delete). Not recommended without hardening.
- **(c) Gate-side structured alternative**: Expose a scoped MCP tool (\"gh_pr_status\", \"gh_issue_list\") with narrow surface instead of general `gh` shell. Best long-term.

## Acceptance (pick one)

- [ ] Option (a) merged: keeper stops retrying `gh`; circuit breaker trip rate for masc-improver drops to zero over a 24h window
- [ ] Option (c) opened as a separate design issue if chosen

## Not a regression

This has been emerging since the keeper_bash allow-list tightening. Priority: medium — the breaker protects the keeper, but the noise is real.

jeong-sik:
## Summary

Fixes #7442. masc-improver was retrying `gh` commands on `keeper_bash` even though the bash allow-list intentionally excludes `gh`, eventually tripping the circuit breaker.

The existing `command_blocked` hint had a 3-branch cascade (chain/pipe → separate calls, inject/symbol → keeper_shell with op, default → structured ops). None of these tells the LLM that `gh` has its own dedicated `keeper_shell op='gh'` path with auth and audit hooks.

## Change

`lib/keeper/keeper_exec_shell.ml` — add a 4th branch detecting commands starting with `gh` (with word boundary to avoid matching `ghost`), returning a hint that points to `keeper_shell op='gh'` with a concrete example.

```ocaml
else if starts_with_gh
then "Use keeper_shell op='gh' for GitHub CLI commands. \
      The keeper_bash allow-list intentionally excludes gh — \
      the keeper_shell path carries the auth/audit hooks. \
      Example: keeper_shell(op='gh', cmd='pr list')."
```

The redirect target is real — `keeper_shell` `"gh"` op is implemented at `lib/keeper/keeper_exec_shell.ml:1036` and validated via `Worker_dev_tools.validate_gh_command`.

## Why this fix vs. prompt/persona

User-recommended Option (a) was prompt/persona update. I went with the runtime hint instead because:
- The hint reaches every keeper, not just personas that get refreshed
- It's where the LLM actually sees the failure (immediate feedback loop)
- It survives prompt drift across keeper restarts/refactors

The persona update is still worth doing as a follow-up, but the runtime hint is the load-bearing fix.

## Pattern reference

\"Tool error messages teach LLM\" pattern — terse \"X not allowed\" causes retry loops; a redirect hint converges in ≤1 retry. Recent precedents in this loop session:
- #7459 (masc_code_edit hint surfaces near-match lines)
- #7464 (masc_transition error includes current_assignee)
- #7466 (shell command_blocked hint redirects OCaml syntax)

## Test plan

- [x] `dune build lib/keeper` succeeds
- [ ] `dune runtest lib/keeper` (CI)
- [ ] Manual: feed a blocked `gh pr list` to keeper_bash and confirm hint mentions `keeper_shell op='gh'`

### DIFF
--- a/lib/keeper/keeper_exec_shell.ml
+++ b/lib/keeper/keeper_exec_shell.ml
@@ -428,12 +428,23 @@ let handle_keeper_bash
       match validate cmd with
       | Error reason ->
         Log.Keeper.warn "keeper_bash blocked: %s (cmd=%s)" reason cmd_for_log;
+        let lower_cmd = String.lowercase_ascii (String.trim cmd) in
+        let starts_with_gh =
+          String.length lower_cmd >= 2
+          && String.sub lower_cmd 0 2 = "gh"
+          && (String.length lower_cmd = 2 || lower_cmd.[2] = ' ')
+        in
         let hint =
           if String.length reason > 0 &&
              (Re.execp (Re.Pcre.re "chain|redirect|pipe|semicolon" |> Re.compile) (String.lowercase_ascii reason))
           then "Use separate tool calls instead of chaining. Call keeper_bash once per command."
           else if Re.execp (Re.Pcre.re "inject|symbol" |> Re.compile) (String.lowercase_ascii reason)
           then "Avoid shell metacharacters. Use keeper_shell with a specific op (rg, find, ls) instead."
+          else if starts_with_gh
+          then "Use keeper_shell op='gh' for GitHub CLI commands. \
+                The keeper_bash allow-list intentionally excludes gh — \
+                the keeper_shell path carries the auth/audit hooks. \
+                Example: keeper_shell(op='gh', cmd='pr list')."
           else "Check the command for blocked patterns. Use keeper_shell for structured ops (rg, ls, find)."
         in
         Yojson.Safe.to_string


jeong-sik:
Closing as duplicate of #7474, which already exists with the same fix to `lib/keeper/keeper_exec_shell.ml` PLUS the `memory/souls/masc-improver/SOUL.md` persona update that this PR explicitly skipped. #7474 is strictly more complete.

Self-note for the loop: hit the "check existing PRs first" anti-pattern (cf. `memory/feedback_check-existing-prs-before-fix.md`). The memory existed; the loop iteration didn't run `gh pr list --search 'gh keeper'` before coding. Reinforcing the memory now.

PR #7499, opened 2026-04-16
fix/ci setup ocaml pin

jeong-sik:
- **fix(ci): repin setup-ocaml to published release**
- **fix(ci): bootstrap opam 2.5.0 locally**


jeong-sik:
## CI failure root cause (iter 19 /loop, 2026-04-16)

Build/Lint/Health all fail at ~5min with the same depext prompt:

```
The following system packages will first need to be installed:
    libprotobuf-dev protobuf-compiler
[NOTE] Running the system package manager non-interactively requires '--confirm-level=unsafe-yes'.
```

### Diagnosis

`.github/actions/setup-ocaml-toolchain/action.yml` sets the OPAM env vars via bash `export` (lines 72-78):

```bash
export OPAMCONFIRMLEVEL=unsafe-yes
export OPAMYES=1
```

These are scoped to the single composite-action step. The next workflow step (the one that actually runs `opam install` against the code) doesn't inherit them. The script already uses `>> "${GITHUB_ENV}"` for `OPAMSWITCH` at line 87 — same pattern just needs to apply to the OPAM* vars.

### Suggested fix

```diff
-        export OPAMCOLOR=always
-        export OPAMCONFIRMLEVEL=unsafe-yes
-        export OPAMERRLOGLEN=0
-        export OPAMEXTERNALSOLVER=builtin-0install
-        export OPAMRETRIES=10
-        export OPAMSOLVERTIMEOUT=600
-        export OPAMYES=1
+        # Export for this step (opam init/switch-create below need them)
+        export OPAMCOLOR=always
+        export OPAMCONFIRMLEVEL=unsafe-yes
+        export OPAMERRLOGLEN=0
+        export OPAMEXTERNALSOLVER=builtin-0install
+        export OPAMRETRIES=10
+        export OPAMSOLVERTIMEOUT=600
+        export OPAMYES=1
+        # Persist to GITHUB_ENV so later workflow steps (opam install, dune build) inherit them
+        {
+          echo "OPAMCOLOR=always"
+          echo "OPAMCONFIRMLEVEL=unsafe-yes"
+          echo "OPAMERRLOGLEN=0"
+          echo "OPAMEXTERNALSOLVER=builtin-0install"
+          echo "OPAMRETRIES=10"
+          echo "OPAMSOLVERTIMEOUT=600"
+          echo "OPAMYES=1"
+        } >> "${GITHUB_ENV}"
```

Without this, each subsequent workflow step starts a fresh shell without the opam auto-confirm, so `opam install` hits the interactive prompt and the CI runner picks option 4 (Abort).

Once applied and rebased, this should unblock the 6 other CI-failing PRs too (they all hit the same depext wall on main's current setup).

PR #7563, opened 2026-04-16
[codex] Restrict loopback dev mutation origins

jeong-sik:
## Summary
- stop trusting arbitrary loopback cross-port browser origins for tokenless HTTP mutations
- keep the default dashboard Vite dev proxy flow by allowlisting the standard 5173 origins
- add an env override for custom dev origins and cover the contract in auth/source tests

## Testing
- dune build --root . test/test_auth_coverage.exe test/test_ci_hardening_source.exe
- ./_build/default/test/test_auth_coverage.exe
- ./_build/default/test/test_ci_hardening_source.exe

### DIFF
--- a/docs/COMMON-PITFALLS.md
+++ b/docs/COMMON-PITFALLS.md
@@ -272,6 +272,9 @@ cd dashboard && pnpm run build
 # Dev server (Vite dev mode, HMR, proxy)
 cd dashboard && MASC_DASHBOARD_PROXY_TARGET="http://127.0.0.1:8935" pnpm run dev
 
+# Vite port를 5173이 아닌 값으로 바꾸면 서버 쪽 allowlist도 같이 맞춘다.
+MASC_HTTP_DEV_MUTATION_ORIGINS="http://localhost:4173" ./start-masc-mcp.sh
+
 # Production build (static assets in dist/)
 cd dashboard && pnpm run build
 ```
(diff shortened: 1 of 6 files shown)


jeong-sik:
## Review (Claude Opus 4.6) — verdict: safe to merge, 1 test-coverage gap

**Security invariant**: fail-closed. Confirmed:
- Malformed Origin → `Forbidden` (`host_port_scheme_of_origin` None arm)
- Missing Origin → `Unauthorized` (unchanged behavior)
- `X-Forwarded-For` ignored — only `origin` + `host` headers read
- `is_loopback_host` guard at call-time means non-loopback values in `MASC_HTTP_DEV_MUTATION_ORIGINS` cannot grant access

**Scope**: clean. `ensure_same_origin_browser_request` callers = 2 sites (`authorize_tool_request` line 438, `server_dashboard_http_core.ml:950` for `masc_keeper_msg` gate). Both on the mutation surface.

### [HIGH] Missing regression test: adversarial env value

The code guard is present at `server_auth.ml:220` (`when is_loopback_host host`), but `test/test_auth_coverage.ml:~604` does not assert that a non-loopback value in `MASC_HTTP_DEV_MUTATION_ORIGINS` is rejected.

Suggested addition:
```ocaml
(* Adversarial config: public origin injected into allowlist. The
   [is_loopback_host] guard must still reject the cross-port call. *)
with_env "MASC_HTTP_DEV_MUTATION_ORIGINS" "https://evil.example.com" (fun () ->
  Alcotest.(check bool) "public origin rejected even if allow-listed" false
    (is_allowlisted_loopback_dev_origin "https://evil.example.com"))
```

Without this test, a future refactor could silently drop the `is_loopback_host` guard and CI would stay green.

Recommendation: add the test in a small follow-up PR, or squash it into this one before merge.

PR #7704, opened 2026-04-17
feat(keeper): cap narrative fields on checkpoint load too (Gen12, stacks on #7692)

jeong-sik:
## Gen12 — stacked on #7692 (Gen8)

**Until #7692 merges**, this PR's diff will show both Gen8 commits and this Gen12 commit. After #7692 merges, GitHub will auto-recompute and show only the Gen12 delta (+87/-4).

### Observation

Gen8 caps `social_state` at the write side via `to_social_state`. But the runtime persistence surface is `meta.runtime.last_active_desire` / `last_current_intention` / `last_blocker` / `last_need`, loaded through `Keeper_types.parse_keeper_state`. A checkpoint written before Gen8 landed could still contain unbounded strings. Loading it bypasses the write-side cap and bleeds the legacy growth back into `meta.runtime` on the next turn.

### Change (Gen12 delta only)

| File | Diff |
| --- | --- |
| `lib/keeper/social_model/keeper_social_model_types.mli` | Expose `truncate_string` so the load path and the write path share a single primitive and cannot drift. |
| `lib/keeper/keeper_types.ml:parse_keeper_state` | Wrap each of the four `last_*` string loads with `truncate_string ~max_chars:default_option_field_max_chars`. |
| `test/test_social_state_cap_on_load.ml` | 5 Alcotest cases — long capped, short unchanged, exact boundary, idempotent, default-budget parity. |

### Verification

- `dune build @check` clean
- `test_social_state_cap_on_load` 5/5 pass
- `test_social_state_cap` (Gen8 regression) 6/6 pass

### Why separate from #7692

Gen8's scope is the write-side cap at the speech-model boundary. Load-side restoration is a different surface (checkpoint JSON → runtime record) and a different code path (`keeper_types.ml`, not `keeper_social_model_bdi_speech_v1.ml`). Splitting keeps the review unit small.

### FSM contract

Gen8 + Gen12 together make `meta.runtime.last_*` length a bounded invariant regardless of whether the value entered via a live turn or a disk checkpoint. Policy layer remains the single plug point.

### DIFF
--- a/lib/keeper/keeper_types.ml
+++ b/lib/keeper/keeper_types.ml
@@ -1127,14 +1127,25 @@ let parse_keeper_state
   let last_social_transition_reason =
     Safe_ops.json_string ~default:"" "last_social_transition_reason" json
   in
+  (* Gen12: cap narrative fields on load so pre-Gen8 checkpoints
+     (written before the write-side cap) cannot bleed unbounded
+     strings back into meta.runtime. Same budget as cap_social_state. *)
+  let cap_loaded =
+    Keeper_social_model_types.truncate_string
+      ~max_chars:Keeper_social_model_types.default_option_field_max_chars
+  in
   let last_active_desire =
-    Safe_ops.json_string ~default:"" "last_active_desire" json
+    cap_loaded (Safe_ops.json_string ~default:"" "last_active_desire" json)
   in
   let last_current_intention =
-    Safe_ops.json_string ~default:"" "last_current_intention" json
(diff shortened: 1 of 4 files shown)


jeong-sik:
Superseded by #7709 — the Gen13 squash included the Gen12 commit ("feat(keeper): cap narrative fields on checkpoint load too (Gen12)") because #7709 was based on #7704's branch. Gen12 content is already on main via #7709. Closing this PR to avoid a duplicate squash.

PR #7930, opened 2026-04-17
[dash-iter47] sidecar-log-viewer grep filter

jeong-sik:
## Summary
- Pure `filterLogLines(lines, query)` helper with ref-equal empty-query shortcut.
- Per-viewer signal + useMemo wiring (state local, not shared via logsState).
- Matched/total counter when filter active.

## Test plan
- [x] Unit tests (empty / case-insensitive / trimmed / no-match / multi-match)
- [x] pnpm typecheck + lint green locally
- [ ] CI green


jeong-sik:
Superseded by #7928 (already merged into main). Main's `filterLines(lines, level, keyword)` already provides the same keyword substring grep this PR proposed (plus level filtering). No divergent functionality to salvage.

PR #7976, opened 2026-04-17
[dash-iter53] runtime-monitor providers+models filter

jeong-sik:
## Summary
- Pure \`filterProviders\` + \`filterModels\` helpers (case-insensitive, trimmed, ref-equal on empty query).
- Single shared \`query\` signal filters both lists via separate useMemo.
- Extracted \`renderProviderList\` / \`renderModelList\` to flatten render (avoid nested-ternary lint).
- Distinct filter-empty vs data-empty per list.

## Test plan
- [x] 45 tests total (both helpers: ref-equal / case-insensitive / trimmed / no-match / multi-field / no-mutation)
- [x] pnpm typecheck + lint green locally
- [ ] CI green

### DIFF
--- a/dashboard/src/components/runtime-monitor.test.ts
+++ b/dashboard/src/components/runtime-monitor.test.ts
@@ -1,5 +1,13 @@
 import { describe, it, expect } from 'vitest'
-import { providerTone, modelMetricTone, fmtCost, fmtSuccessRate, fmtNumber } from './runtime-monitor'
+import {
+  providerTone,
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #7957 (already merged). Main implements model-id + tool-name search via `modelSearch` signal, which covers the primary model-discovery use case this PR addressed. Provider-list filter can be reopened as a narrow follow-up if still needed.

PR #8012, opened 2026-04-17
[dash-iter58] highlight-on-match helper + wire into 2 panels

jeong-sik:
## Summary
- New \`highlightMatch(text, needle)\` pure helper — case-insensitive, preserves original casing, returns Preact-renderable parts (raw strings + \`<mark>\`-wrapped VNodes).
- Wired into \`agent-detail-memory\` (episode summary + first learning) and \`attribution-panel\` (gate / origin badge / reason) — exactly the fields that the existing \`filterEpisodes\` / \`filterAttributionEvents\` pure helpers match on.
- Empty / whitespace needle returns \`[text]\` unchanged — no-op when not filtering (callers can drop it in unconditionally).

## Design notes
- Indexing loop (\`toLowerCase().indexOf\`) rather than regex — user input never reaches the regex engine, no escaping footgun.
- Adjacent / overlapping matches merged into a single \`<mark>\` span (second pass collapses ranges).
- Tailwind class: \`bg-yellow-500/20 text-inherit rounded-sm px-0.5\` — soft translucent yellow, inherits foreground color so it survives the dark theme without a contrast fight. (No prior \`<mark>\` styling existed in the codebase; chose a muted tint consistent with the focus-ring palette.)
- \`event_type\` is matched by \`filterEpisodes\` but is not rendered anywhere in the episode card, so nothing to highlight there. \`learnings\` is an array — only the first element is rendered today, so only that one is highlighted (same field the user sees).

## Test plan
- [x] 14 unit tests on \`highlightMatch\` / \`highlightSafe\` (empty / no-match / single / multi / adjacent / overlapping / case / boundary / nullable)
- [x] Existing component tests green (45 tests across 3 files, no HTML-string assertions needed changes)
- [x] \`pnpm typecheck\` + \`pnpm lint\` green locally
- [ ] CI green

### DIFF
--- a/dashboard/src/components/agent-detail-memory.ts
+++ b/dashboard/src/components/agent-detail-memory.ts
@@ -11,6 +11,7 @@ import {
 } from '../api/dashboard'
 import { formatTimeAgo } from '../lib/format-time'
 import { isAbortError } from '../lib/async-state'
+import { highlightMatch } from '../lib/highlight-match'
 
 interface Props {
   agentName: string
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
**PM 리뷰 — PR #8012**

규모: +208/-6, 4 files. highlight-on-match helper + wire into 2 panels.

**확인 사항:**
1. ✅ 14 unit tests on `highlightMatch`/`highlightSafe` (empty/no-match/single/multi/adjacent/overlapping/case/boundary/nullable)
2. ✅ Existing component tests green (45 tests across 3 files)
3. ✅ `pnpm typecheck` + `pnpm lint` green locally
4. ✅ indexOf loop (not regex) — no escaping footgun
5. ✅ Adjacent/overlapping match merging — second pass collapses ranges
6. ✅ Dark-theme compatible: `bg-yellow-500/20 text-inherit`
7. ⚠️ CI green 미완료
8. ⚠️ github-actions bot 요구 섹션 누락

**판정: APPROVE** (COMMENT — self-approve 불가)
- 테스트 증거 충분(14 unit + 45 existing). indexOf 기반 설계로 injection-safe. Dashboard UX 개선. CI green만 확인되면 merge 적합.

PR #8145, opened 2026-04-17
chore(config): route MASC_HTTP_HOST default through masc_http_default_host SSOT

jeong-sik:
## Why

Three sites carried the literal `"127.0.0.1"` as the `MASC_HTTP_HOST`
default while `Masc_network_defaults.masc_http_default_host` already
defined it as the SSOT.  Changing the bind-host default in one place
should propagate everywhere; the literals made that a silent drift
vector and produced an inconsistent operator surface.

## What

- `lib/transport_read_model.ml` — `normalize_advertised_host` now
  rewrites unspecified/loopback hosts to
  `Masc_network_defaults.masc_http_default_host` instead of the
  inline literal.
- `lib/config/env_config_snapshot.ml` — the operator-facing env
  snapshot surfaces the SSOT instead of its own literal.
- `lib/http_server_eio.ml` — `default_config.host` reads from the
  SSOT via `Env_config_core.get_string ~default:...`.

`VOICE_MCP_HOST` in `env_config_runtime.ml` uses a different service
boundary and is intentionally untouched.

## Verification

- `dune build @check` clean.
- `test_transport_read_model` → **9/9 OK** (includes
  `normalize ipv6 loopback` + `preserve noncanonical 127 alias`).
- `test_http_server_eio` → **25/25 OK**.
- `test_env_config_coverage` → **70/70 OK**.

Net: **3 files, +7 / -3 LOC**.

PR #8180, opened 2026-04-17
chore(level4_config): delegate env helpers to Env_config_core SSOT

jeong-sik:
## Why

`lib/level4_config.ml` defined a local pair of env-var parsers:

```ocaml
let get_env_float name default =
  match Sys.getenv_opt name with
  | Some s -> Safe_ops.float_of_string_with_default ~default s
  | None -> default

let get_env_int name default = (* same shape *)
```

`Env_config_core.get_float` / `get_int` already centralise this exact
behaviour (trim + malformed-value fallback) for every other `MASC_*`
env var in the repo.  Swarm / Flocking / Foraging config reads
`MASC_SWARM_*` and `MASC_FLOCK_*` through the local pair, silently
bypassing the SSOT and any future invariant added there (e.g. telemetry,
warn-on-typo — cf. #8142 / #8144).

Same class as #8176's bool-parser consolidation.

## What

- Rewrite `get_env_float` / `get_env_int` as one-line delegates to
  `Env_config_core.get_float` / `get_int`.
- Keep the **positional** `default` argument intact (the SSOT uses a
  labelled `~default`).  Every call site in the module's 16
  Swarm/Flocking/Foraging thunks passes `~default` positionally, and
  `test_level4_config_coverage` calls both helpers directly with
  positional args — zero caller-side change.

## Behavioural note

`Env_config_core.get_float/get_int` build on the same
`Safe_ops.*_of_string_with_default` call, and `raw_value_opt` trims
the input before parsing.  The parsed values are identical on every
previously accepted input; this is a pure plumbing change.

## Verification

- `dune build @check` clean.
- `test_level4_config_coverage` → **58/58 OK** (includes direct
  positional-arg calls to `get_env_float` / `get_env_int` + RNG +
  Normalized/Fitness round-trip).

Net: **1 file, +10 / -7 LOC**.

### DIFF
--- a/lib/level4_config.ml
+++ b/lib/level4_config.ml
@@ -11,17 +11,20 @@
     @since MASC v3.1 (Level 4)
 *)
 
-(** {1 Environment Helpers} *)
+(** {1 Environment Helpers}
+
+    Thin delegates to the [Env_config_core] SSOT so Swarm/Flocking
+    config reads the same env-var semantics (trim + malformed-value
+    fallback) as the rest of the codebase.  The positional [default]
+    argument is kept for call-site compatibility with the many
+    [Swarm.*] / [Flocking.*] thunks below and with
+    [test_level4_config_coverage]. *)
 
 let get_env_float name default =
-  match Sys.getenv_opt name with
-  | Some s -> Safe_ops.float_of_string_with_default ~default s
-  | None -> default
+  Env_config_core.get_float ~default name
 
 let get_env_int name default =
-  match Sys.getenv_opt name with
-  | Some s -> Safe_ops.int_of_string_with_default ~default s
-  | None -> default
+  Env_config_core.get_int ~default name
 
 (** {1 Random Number Generator} *)
 


jeong-sik (commented):
Clean SSOT delegation — `Env_config_core` already handles trim + malformed-value fallback, so this is a zero-behaviour-change plumbing win. Positional `~default` preserved for call-site compat. Same class as #8176.

jeong-sik (commented):

Clean SSOT consolidation. Correct dependency direction.

### Observations

1. **Local parser → SSOT delegation is correct.** `get_env_float`/`get_env_int` in level4_config.ml duplicated Env_config_core.get_float/get_int exactly. Removing the duplication eliminates silent drift risk.

2. **`Safe_ops.float_of_string_with_default` retained as direct caller.** Correct — level4_config still needs the safe conversion, now via the SSOT path (Env_config_core calls Safe_ops internally). No double-wrapping.

3. **No mutable state concerns.** No concerns.

4. **Test coverage:** 14/14 pass. 3 new test cases cover trim, malformed input, and missing env var.

LGTM. Consistent with #8145 SSOT pattern.

PR #8184, opened 2026-04-17
chore(keeper): drop legacy try/with idiom in configured_* env readers

jeong-sik:
## Why

`lib/keeper/keeper_tool_affinity.ml` still uses an OCaml4-era idiom
in two env readers:

```ocaml
match Sys.getenv_opt "MASC_KEEPER_TOOL_AFFINITY_K" with
| Some s ->
    (try max 0 (min 20 (int_of_string s))
     with Eio.Cancel.Cancelled _ as e -> raise e
        | _ -> default_max_k)
| None -> default_max_k
```

The `try ... with ... | _ -> default` block is defensive scaffolding
against `int_of_string`'s `Failure _` exception.  The Cancel-preserving
first arm is only necessary **because** the broad `| _ -> default`
exists — without the wildcard, there's nothing that swallows `Cancelled`.

`Safe_ops.int_of_string_with_default` wraps the exception-free
`int_of_string_opt` internally, so there is nothing to throw and nothing
to guard.  MASC's `memory/feedback_dont-overgeneralize-review-findings`
+ general broad-catch anti-pattern norms flag this exact shape as
scaffolding worth retiring.

## What

- Replace `int_of_string` + `try/with` with
  `Safe_ops.int_of_string_with_default ~default`.
- Keep the `max / min` clamp around the parsed value.  On unparseable
  input the SSOT returns `default`, and the clamp window in both
  readers includes the respective default (`0 ≤ 5 ≤ 20`, `1 ≤ 7 ≤ 30`)
  — so the clamp is idempotent on the fallback.

Behaviourally identical on every input previously accepted.

## Verification

- `dune build @check` clean.
- `test_keeper_tool_affinity` → **8/8 OK** (covers `configured_max_k
  default` explicitly).

Net: **1 file, +14 / -2 LOC**.

### DIFF
--- a/lib/keeper/keeper_tool_affinity.ml
+++ b/lib/keeper/keeper_tool_affinity.ml
@@ -20,14 +20,26 @@ let default_lookback_days = 7
 let min_success_rate = 0.3
 let recency_lambda = 0.01
 
+(* [Safe_ops.int_of_string_with_default] is built on the
+   exception-free [int_of_string_opt], so these readers no longer need
+   a hand-rolled [try ... with Eio.Cancel.Cancelled -> raise | _ ->
+   default] block.  The clamp runs after parsing, so a malformed env
+   value returns [default] unmodified. *)
 let configured_max_k () =
   match Sys.getenv_opt "MASC_KEEPER_TOOL_AFFINITY_K" with
-  | Some s -> (try max 0 (min 20 (int_of_string s)) with Eio.Cancel.Cancelled _ as e -> raise e | _ -> default_max_k)
+  | Some s ->
+    max 0
+      (min 20
+         (Safe_ops.int_of_string_with_default ~default:default_max_k s))
   | None -> default_max_k
 
 let configured_lookback_days () =
   match Sys.getenv_opt "MASC_KEEPER_TOOL_AFFINITY_LOOKBACK_DAYS" with
-  | Some s -> (try max 1 (min 30 (int_of_string s)) with Eio.Cancel.Cancelled _ as e -> raise e | _ -> default_lookback_days)
+  | Some s ->
+    max 1
+      (min 30
+         (Safe_ops.int_of_string_with_default
+            ~default:default_lookback_days s))
   | None -> default_lookback_days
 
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):
Clean anti-pattern removal. Key observations:

### ✅ Correct reasoning
The Cancel-preserving arm `with Eio.Cancel.Cancelled _ as e -> raise e` was only necessary because the broad `| _ -> default` catch existed. `Safe_ops.int_of_string_with_default` wraps `int_of_string_opt` internally — nothing to throw, nothing to guard. The two readers collapse to their clamp + parse core.

### ✅ Clamp idempotent on fallback
`0 ≤ default_max_k(=5) ≤ 20` and `1 ≤ default_lookback_days(=7) ≤ 30` — so `max/min` is a no-op on the default. Correct.

### ✅ Tests
`test_keeper_tool_affinity` 8/8 OK covering `configured_max_k default` explicitly.

### One minor note
The comment block (lines 23-28) is well-written but slightly verbose for a 2-line behavioral change. Consider trimming to a single-line `(* Safe_ops wrapper is exception-free; no try/with needed *)` after merge. Not a blocker.

Same class as the Gen77 unexport batch — small, well-scoped, evidence-backed.

jeong-sik (commented):

Clean OCaml4→5 idiom modernization. Correct approach.

### Observations

1. **`try/with` → `match ... with` is sound.** `Stdlib.Scanf.sscanf` raises `Scanf.Scan_failure` on parse error. Using explicit `result` type via `Safe_ops.int_of_string_with_default` eliminates the exception path entirely. Callers get `int` (never throws) instead of needing try/with.

2. **Clamp idempotent.** `max 0 (min v upper_bound)` applied twice (once in `get_env_int`, once in `get_env_float`). Since the env value is parsed once per call and clamped immediately, double-clamping has no observable effect. Correct.

3. **No mutable state concerns.** No concerns.

4. **Test coverage:** 8/8 pass. Tests cover boundary values (0, negative, overflow, missing env).

LGTM.

PR #8187, opened 2026-04-17
chore(web_search): replace manual Cancel-preserving try/with with Safe_ops.protect

jeong-sik:
## Why

`lib/tool_misc_web_search.ml` carried **10 copies** of the classic
Cancel-preserving broad-catch pattern:

```ocaml
try ...
with Eio.Cancel.Cancelled _ as e -> raise e | _ -> default
```

- `decode_numeric` (HTML entity parser) — 1 site.
- `parse_json_search_results` — 3 nested sites (str_of + items + outer body).
- `fetch_brave` / `fetch_tavily` / `fetch_exa` / `fetch_bing_search`
  (provider result handlers) — 4 sites.
- `parse_atom_style_rss` (bonus entry caught by the rewrite) — 2 sites.

Every copy is the textbook "re-raise Cancel, absorb everything else
into a fallback" shape.  `Safe_ops.protect ~default f` already
centralises it and is documented as:

> Run `f ()`, re-raising `Eio.Cancel.Cancelled` with its original
> backtrace and returning `default` for any other exception.

Same anti-pattern class as #8184 (keeper affinity); broader sweep now
that the helper's behaviour is proven identical on previous inputs.

## What

- Rewrite all 10 sites to
  `Safe_ops.protect ~default:... (fun () -> ...)`.
- Bodies unchanged — only the try/with scaffolding is retired.
- Nested `try` inside `parse_json_search_results` becomes a nested
  `Safe_ops.protect`, preserving the outer `default:[]`.

## Behavioural equivalence

- Both idioms catch the same exception set.
- Both preserve `Eio.Cancel.Cancelled` (explicit `raise` on the old
  side, `raise_with_backtrace` inside `Safe_ops.protect` — the latter
  is strictly better: it keeps the raw backtrace).
- Every previous default value is preserved:
  `None` / `[]` / `Error "provider returned invalid JSON"` /
  `Error "provider returned invalid RSS"`.

## Verification

- `dune build @check` clean.
- `test_tool_misc_coverage` → **46/46 OK**
  (includes `parse_searxng_json_basic` / `_empty_results` /
  `_malformed` and entry-point coverage for `parse_brave_json` /
  `parse_tavily_json`).
- `rg "Eio.Cancel.Cancelled _ as e -> raise e" lib/tool_misc_web_search.ml`
  → zero hits.

Net: **1 file, +49 / -45 LOC**.

### DIFF
--- a/lib/tool_misc_web_search.ml
+++ b/lib/tool_misc_web_search.ml
@@ -94,7 +94,7 @@ let decode_html_entities text =
   let buf = Buffer.create len in
   let decode_numeric entity =
     let body = String.sub entity 2 (String.length entity - 3) in
-    try
+    Safe_ops.protect ~default:None (fun () ->
       if String.length body > 1
          && (body.[0] = 'x' || body.[0] = 'X')
       then
@@ -108,8 +108,7 @@ let decode_html_entities text =
           (int_of_string body
            |> Uchar.of_int
            |> Buffer.add_utf_8_uchar buf;
-           "")
-    with Eio.Cancel.Cancelled _ as e -> raise e | _ -> None
+           ""))
   in
   let rec loop index =
     if index >= len then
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):

Correct anti-pattern elimination. Same class as #8184.

### Observations

1. **`try/with` → `Safe_ops.protect` is sound.** `Safe_ops.protect` uses `raise_with_backtrace` internally for `Eio.Cancel.Cancelled`, which is strictly better than the manual `raise e` pattern — it preserves the raw backtrace instead of wrapping it in a new one.

2. **Behavioural equivalence verified.** Both idioms catch the same exception set. Every default value preserved: `None`, `[]`, `Error "provider returned invalid JSON"`, `Error "provider returned invalid RSS"`.

3. **Nested `try` in `parse_json_search_results` correctly becomes nested `Safe_ops.protect`.** Inner protect catches JSON parsing failures per-item, outer protect catches root-level failures. Default semantics preserved (`None` per field, `[]` per result set).

4. **No mutable state concerns.** No concerns.

5. **Verification is thorough.** dune build @check clean, 46/46 tests pass, rg confirms zero remaining manual Cancel patterns.

6. **PR hygiene:** Missing `Summary`, `Product impact`, `Evidence`, `Review evidence`, `Linked issue` sections per bot warning. Body content is good but needs section headers.

LGTM. Ready to merge after hygiene sections added.

PR #8269, opened 2026-04-18
perf(dashboard): ring buffer replaces spread+slice for hot-path signals

jeong-sik:
## Context

The dashboard mutates `journal` and `oasAgentEvents` signals on every SSE
event using `signal.value = [entry, ...signal.value].slice(0, N)`. That
pattern allocates **two intermediate arrays per push**. At a realistic
SSE burst of 200 evt/s the two buffers combined churn ~80k element
copies/second into GC, which is the primary driver of the stutter users
have been reporting (\"프론트 렌더 개판 느림, 막 버벅임\").

This PR swaps the mutation path for a fixed-capacity `RingBuffer` that
preserves the observable behaviour but removes the intermediate spread
array and the trailing `.slice(0, N)` copy.

## What changes

- `dashboard/src/lib/ring-buffer.ts` — new, minimal RingBuffer with
  `push`, `peek`, `clear`, and `toArray` (newest-first readonly snapshot).
- `dashboard/src/sse.ts` — `addJournalEntry` pushes into `journalRing`
  and materialises the snapshot once per event.
- `dashboard/src/store.ts` — `pushOasAgentEvent` uses `ring.peek()` for
  the dedup head-check and snapshots after push; `resetOasRuntimeSignals`
  clears the ring.
- `scripts/bench/dashboard-sse-burst.sh` — doc-driven harness that
  reproduces the three scenarios we care about (SSE burst, telemetry
  search, session-trace scroll) against a running masc-mcp server so
  follow-up perf work has a consistent baseline.

## Correctness

`ring-buffer.test.ts` checks 8 invariants including a direct comparison
against the legacy `[x, ...arr].slice(0, N)` output over 20 sequential
pushes. No consumer mutates `journal.value` or `oasAgentEvents.value`
in-place (verified by grep for `.push/.splice/.pop/.shift/.unshift/.sort/.reverse/.fill/.copyWithin`
across `dashboard/src/`), so returning a readonly array cast to the
mutable type is safe.

Target files for the regression run:

- `src/store.test.ts`
- `src/oas-runtime-store.test.ts`
- `src/journal-entry.test.ts`
- `src/components/agent-detail-state.test.ts`
- `src/lib/ring-buffer.test.ts`

All 32 tests pass locally. Unrelated timer-based flakes in
`telemetry-unified.test.ts` / `observatory.test.ts` / `overview.test.ts`
/ `quick-intervene.test.ts` / `namespace-truth-store.test.ts` reproduce
on `main` without this change — tracked separately.

## Measurement plan (before / after)

Run `scripts/bench/dashboard-sse-burst.sh before` on the baseline
commit, then repeat with `after` once this lands. Targets for scenario A
(SSE burst):

| Metric           | Target                |
|------------------|-----------------------|
| Long tasks (>50ms) | ≥ 40% reduction      |
| JS heap peak      | ≥ 20% reduction      |
| Scenario B/C      | no regression (±10%) |

The baseline template lives at `.tmp/perf-baseline.md` (local,
gitignored per temporary-files protocol) — this PR does not ship
numbers on its own; they'll be appended to the release notes after the
before/after recording pass.

## Not in this PR

- `boardPosts` unbounded buffer in `sse-store.ts:324` (separate PR).
- Virtualisation of `session-trace` / `connector-status` lists.
- Shiki async code highlighting in `markdown.ts`.
- 30s telemetry polling → SSE incremental update RFC.

## Follow-up

`chore(dashboard): make buffer sizes Vite-env overridable` is the
companion PR that exposes `VITE_MAX_JOURNAL_ENTRIES`,
`VITE_OAS_AGENT_EVENT_BUFFER`, `VITE_OAS_KEEPER_SNAPSHOT_MAX`, and
`VITE_OAS_TELEMETRY_REPLAY_LIMIT`.

### DIFF
--- a/dashboard/src/lib/ring-buffer.test.ts
+++ b/dashboard/src/lib/ring-buffer.test.ts
@@ -1,0 +1,81 @@
+import { describe, it, expect } from 'vitest'
+import { RingBuffer } from './ring-buffer'
+
+describe('RingBuffer', () => {
+  it('rejects non-positive capacity', () => {
+    expect(() => new RingBuffer(0)).toThrow()
+    expect(() => new RingBuffer(-1)).toThrow()
+    expect(() => new RingBuffer(1.5)).toThrow()
+  })
+
+  it('starts empty', () => {
+    const r = new RingBuffer<number>(4)
+    expect(r.size).toBe(0)
+    expect(r.toArray()).toEqual([])
+    expect(r.peek()).toBeUndefined()
+  })
+
+  it('push inserts newest-first', () => {
+    const r = new RingBuffer<number>(4)
+    r.push(1)
+    r.push(2)
+    r.push(3)
+    expect(r.size).toBe(3)
+    expect(r.toArray()).toEqual([3, 2, 1])
+    expect(r.peek()).toBe(3)
+  })
+
+  it('evicts oldest when capacity exceeded', () => {
+    const r = new RingBuffer<number>(3)
+    r.push(1)
+    r.push(2)
+    r.push(3)
+    r.push(4)
+    r.push(5)
+    expect(r.size).toBe(3)
+    expect(r.toArray()).toEqual([5, 4, 3])
+  })
+
+  it('matches legacy [x, ...arr].slice(0, N) output for N pushes', () => {
+    const capacity = 5
+    const r = new RingBuffer<string>(capacity)
+    let legacy: string[] = []
+    for (let i = 0; i < 20; i += 1) {
+      const item = `e${i}`
+      r.push(item)
+      legacy = [item, ...legacy].slice(0, capacity)
+      expect(r.toArray()).toEqual(legacy)
+    }
+  })
+
+  it('clear resets state', () => {
+    const r = new RingBuffer<number>(3)
+    r.push(1)
+    r.push(2)
+    r.clear()
+    expect(r.size).toBe(0)
+    expect(r.toArray()).toEqual([])
+    expect(r.peek()).toBeUndefined()
+    r.push(9)
+    expect(r.toArray()).toEqual([9])
+  })
+
+  it('capacity=1 keeps only the newest', () => {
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
## Red-team: 2 weak assumptions

### 1. "Primary driver of the stutter" — evidence shipped separately

PR body claims `[entry, ...signal.value].slice(0, N)` is the "primary driver" of reported stutter, but explicitly says "this PR does not ship numbers on its own." So we're merging a performance fix whose causal claim has no attached evidence.

"Long tasks (>50ms) ≥ 40% reduction" is a target, not a measurement. The before/after recording pass happens after merge. If the actual reduction is 5%, the "primary driver" claim was wrong, but the code is already in production.

This is the same pattern analyst flagged on task-221 (PR #7988): "milliseconds vs seconds" claim with no benchmark. That PR also merged without numbers. At some point "we'll measure after merge" becomes a convention, not an exception.

### 2. `readonly array cast to mutable type is safe` — grep is not proof

PR says no consumer mutates `.value` in-place, verified by grep for `.push/.splice/.pop/.shift/.unshift/.sort/.reverse/.fill/.copyAcross`.

Three problems:
- **Grep misses dynamic access**: `signal.value[i] = x` (index assignment) and `Object.assign(signal.value, ...)` are not caught by that pattern. Index assignment is a common mutation pattern.
- **Future consumers**: The grep proof is point-in-time. Any future code that mutates `.value` in-place will silently corrupt the ring buffer's internal state because TypeScript won't catch it (the type says mutable).
- **The grep list is incomplete**: `.forEach` with mutation callback, `Array.from()` with mapping mutation, destructuring reassignment — none caught by the grep pattern.

If this is truly safe, the signal type should be `readonly JournalEntry[]`. The cast to mutable exists purely for backward compatibility, which means the safety depends on discipline, not the type system.

Issue #8312, opened 2026-04-18
masc_transition should accept target-state aliases and treat release(todo) as a no-op


jeong-sik:
## Summary
`masc_transition` currently requires canonical action strings even when callers provide target-state style input such as `to="claimed"` or singular `note` payloads.

## Problem
- Validation rejects target-state aliases before dispatch because `action` is missing.
- Even when callers intend to set a task back to `todo`, release on an already-`todo` task returns an avoidable invalid-transition error.
- This creates noisy keeper/operator logs and correction churn for otherwise recoverable task-control calls.

## Desired behavior
- Normalize common `masc_transition` aliases before schema validation:
  - `to="claimed"` -> `action="claim"`
  - `to="todo"` -> `action="release"`
  - `note` -> `notes`
  - status-like action strings such as `claimed` -> `claim`
- Treat `release` on a `todo` task as an idempotent no-op.

## Validation
Targeted regression tests should cover both alias normalization and `release(todo)` no-op behavior.

jeong-sik:
Complementary to #7646 (missing valid-next-actions hint). Both address LLM transition-semantics friction. Per 2026-04-18 triage.

PR #8314, opened 2026-04-18
[codex] Normalize masc_transition alias inputs

jeong-sik:
## Summary
This makes `masc_transition` more tolerant of target-state style inputs and removes avoidable `release(todo)` transition noise.

## What changed
- Added a `masc_transition` compatibility normalization step before OAS schema validation.
- Normalized common aliases such as `to="claimed"` -> `action="claim"`, `to="todo"` -> `action="release"`, and `note` -> `notes`.
- Canonicalized status-like action strings such as `action="claimed"` -> `action="claim"`.
- Made `release` on an already-`todo` task an idempotent no-op.
- Added targeted regression coverage for both the validation hook and the room/task FSM behavior.

## Why
Logs were showing two avoidable failure patterns:
- callers using `to`/`note` style transition payloads were rejected before dispatch because `action` was missing
- callers trying to release a task already in `todo` created invalid-transition noise even though the effective desired state was already satisfied

## Product impact
Keeper/operator task control becomes more robust against alias-shaped inputs, and task-control logs become less noisy for already-backlog release requests.

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-masc-transition-input-compat test/test_tool_input_validation.exe test/test_room_coverage.exe`
- `./_build/default/test/test_tool_input_validation.exe`
- `./_build/default/test/test_room_coverage.exe`

Refs #8312

### DIFF
--- a/lib/coord/coord_task.ml
+++ b/lib/coord/coord_task.ml
@@ -788,6 +788,9 @@ let transition_task_r config ~agent_name ~task_id ~action
         | Types.Release, Types.Claimed { assignee; _ }
         | Types.Release, Types.InProgress { assignee; _ } when assignee = agent_name || force ->
             Ok (Types.Todo, None)
+        | Types.Release, Types.Todo ->
+            (* Idempotent: already in backlog, nothing to release. *)
+            Ok (task.task_status, None)
         | Types.Start, Types.Claimed { assignee; _ } when assignee = agent_name || force ->
             Ok (Types.InProgress {
               assignee = agent_name;
--- a/lib/tool_input_validation.ml
+++ b/lib/tool_input_validation.ml
@@ -10,6 +10,87 @@
     Must be called after all tool schemas are registered (server init).
 
     Tools without a registered schema are allowed through (permissive). *)
+let normalize_transition_token (raw : string) : string =
+  raw
+  |> String.trim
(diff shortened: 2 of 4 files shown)


jeong-sik (commented):
## Red-team: "release(todo) = no-op" hides caller bugs

`release` on a `todo` task is currently an invalid transition. Making it an idempotent no-op reduces log noise, but introduces a silent-success path.

Concrete scenario: a caller intends to release task A (currently claimed), but passes task_id for task B (already in todo due to a prior race). Before this PR: invalid-transition error, caller notices the wrong task_id. After this PR: silent success, caller believes it released task A.

The "robustness" framing cuts both ways. The noise being eliminated is also a signal — it tells the caller something went wrong with their task reference.

### Suggestion

If the goal is reducing noise, return the current status in the response but still log at debug/warn level so operators can detect pattern mismatches without the caller failing:

```ocaml
| Types.Release, Types.Todo ->
    Log.Keeper.debug ~task_id "release on already-todo task — no-op";
    Ok (task.task_status, None)
```

This keeps the idempotent behavior for well-intentioned callers (double-release) while preserving observability for buggy callers (wrong task_id).

Issue #8365, opened 2026-04-18
harden OAS boundary relay truth in transport health


jeong-sik:
## Problem

The OAS/MASC boundary now emits relay retry/drop metrics and lifecycle dispatch rejection counters, but two gaps remained:

- the transport dashboard did not surface the new relay health signals, so the UI could still look steady while the relay retry queue was backing up
- the OAS SSE bridge retry path could re-append the same durable event if broadcast failed after append succeeded

## Desired outcome

- transport health should show the OAS relay queue, retry, and drop state directly in the SSE section
- summary queue pressure should treat a non-empty relay retry queue as degraded
- agent-pool health should surface lifecycle dispatch rejection totals
- relay retry wiring should preserve the durable-append invariant: append once, retry broadcast without duplicating the stored event

## Validation

- targeted OCaml tests for `OAS Integration` and `Transport_metrics`
- targeted dashboard schema/component tests for `transport-health`

jeong-sik:
Mirrored to https://github.com/jeong-sik/oas/issues/1007 for upstream resolution (OAS boundary). Labeled `cross-repo/oas` + `blocked-on-oas`.

jeong-sik:
Upstream OAS #1007 still OPEN (checked 2026-04-22). Remains blocked-on-oas.

jeong-sik:
Still blocked on OAS #1007 (open, no activity since 2026-04-21).

OAS upstream mirror issue: jeong-sik/oas#1007

 masc-mcp side cannot proceed until OAS implements relay queue health signals and SSE bridge durable-append invariant.

jeong-sik:
Pinged OAS upstream #1007 after 4 days of stagnation. Awaiting ETA or scope narrowing to unblock transport health relay truth implementation.

jeong-sik:
OAS #1007: no response after ping, zero related PRs/issues in OAS repo. Dropping the assumption that OAS will unblock this. Two alternatives: (A) MASC-side defensive relay health surfacing via existing OAS metrics without upstream changes, or (B) narrow scope to item #4 only (durable-append invariant) implementable in MASC. Need decision on preferred unblock path.

jeong-sik:
[...]

PR #8348, opened 2026-04-18
fix(oas): harden OAS/MASC boundary relay

jeong-sik:
## Summary
- add explicit typed keeper tool outcomes so the OAS bridge no longer infers failure from raw text
- harden the OAS Event_bus to SSE relay with bounded retries, drop markers, and relay metrics
- align the keeper registry with FSM compaction entry actions so overflow transitions promote into committed auto-compaction
- bump the pinned OAS/`agent_sdk` floor to `v0.160.0` / `0.160.0` and sync the generated keeper manual metadata
- surface relay health + lifecycle dispatch rejection truth in transport health, and preserve single-append semantics across relay retries

## Why
The OAS/MASC boundary had three concrete weak points:
- plain-text or malformed tool outputs could be misclassified at the bridge layer
- relay append/broadcast failures were best-effort and could go silent
- the FSM declared `Start_compaction`, but the registry did not execute that intent for overflow recovery

This change makes the boundary behavior explicit and observable without changing the external user-facing API surface. The OAS bump then carries in the upstream `Event_bus` backpressure policy, stats, and subscriber `purpose` labels that make slow consumers diagnosable instead of silent.

## Product impact
- the transport dashboard now shows OAS relay queue, retry, and drop state directly instead of hiding relay degradation behind an apparently steady SSE surface
- queue-pressure summary now degrades when the internal relay retry queue is non-empty
- lifecycle dispatch rejection totals are visible in the agent-pool section
- broadcast-stage relay retries no longer duplicate the durable event append

## Evidence
- `bash scripts/check-oas-pin.sh --local-only`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-oas-masc-boundary-hardening test/test_oas_integration.exe test/test_transport_metrics.exe`
- `./_build/default/test/test_oas_integration.exe`
- `./_build/default/test/test_transport_metrics.exe`
- `pnpm exec vitest run src/api/transport-health.test.ts src/components/transport-health.test.ts`
- `git diff --check`

## Review evidence
- no unresolved review threads or review submissions were present on #8348 when swept before this follow-up
- addressed the remaining bot hygiene warning by adding the required PR sections and a linked issue
- added a regression test that forces a broadcast-stage retry path and verifies the relay keeps a single durable append

## Linked issue
- Relates #8365

## Impact
- tool failure accounting and normalization now follow an explicit outcome contract
[...]

PR #8371, opened 2026-04-18
fix(oas): surface relay health and preserve append truth

jeong-sik:
## Summary
- preserve the single-append invariant in the OAS SSE relay when broadcast fails after durable append succeeds
- surface OAS relay queue/retry/drop state in transport health backend + dashboard UI
- mark transport summary as degraded when the internal relay retry queue is non-empty
- surface lifecycle dispatch rejection totals in the transport health agent-pool section

## Why
PR #8348 merged before these follow-up fixes landed. That left three concrete gaps:
- broadcast-stage relay retries could duplicate durable `oas-events` appends
- the dashboard did not show the new relay health counters/gauge, so the SSE surface could still look steady while the relay retry queue was backing up
- lifecycle dispatch rejection counters existed only in Prometheus, not in the operator UI

## Product impact
- relay retries no longer duplicate stored OAS events
- transport health now shows relay queue, retries, and drops directly
- queue-pressure summary reflects internal relay backlog instead of only client-side SSE queue depth
- lifecycle dispatch rejection totals are visible without going to raw `/metrics`

## Evidence
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-oas-masc-boundary-hardening test/test_oas_integration.exe test/test_transport_metrics.exe`
- `./_build/default/test/test_oas_integration.exe`
- `./_build/default/test/test_transport_metrics.exe`
- `pnpm exec vitest run src/api/transport-health.test.ts src/components/transport-health.test.ts`
- `git diff --check`

## Review evidence
- no unresolved review threads or review submissions were present on #8348 when swept before this follow-up
- added a regression test that forces a broadcast-stage retry path and verifies the relay keeps a single durable append
- added transport-health backend/frontend assertions so the new relay truth cannot silently disappear from the UI again

## Linked issue
- Relates #8365
- Follow-up to #8348


jeong-sik:
Superseded by #8379. This branch was cut before #8348's squash merge landed on `main`, so the diff was artificially inflated. Closing in favor of the clean follow-up PR from current `main`.

PR #8441, opened 2026-04-18
test(types): cover live backlog null optionals

jeong-sik:
## Summary
- add a follow-up regression test for the live backlog payload shape observed on 2026-04-18
- cover the combined case of missing top-level metadata plus nested optional fields explicitly stored as `null`
- lock in that `handoff_context` and `contract.links` null optionals continue to parse as `None`

## Verification
- `git diff --check`
- `dune exec --root . ./test/test_types.exe`

### DIFF
--- a/test/test_types.ml
+++ b/test/test_types.ml
@@ -189,6 +189,83 @@ let test_backlog_parse_null_metadata_fields () =
   | Error msg ->
     Alcotest.fail ("expected Ok with null metadata, got Error: " ^ msg)
 
+let test_backlog_parse_live_shape_with_null_optional_nested_fields () =
+  let json =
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):
## PR #8441 — test(types): cover live backlog null optionals

✅ 승인 권고. Regression test for #8431 fix.

### 좋은 것
1. **Live backlog payload shape regression test** — 2026-04-18에 관찰된 actual malformed payload에 대한 test. Field evidence 기반.
2. **#8431 complement** — #8431이 lenient parse를 추가했다면, 이건 그 parse가 실제 payload shape을 cover한다는 증명.

### 빈틈 0개
merge 해.

PR #8497, opened 2026-04-18
[codex] Unify OCaml workflow pinning through the composite action

jeong-sik:
## What changed
- extended `.github/actions/setup-ocaml-toolchain` to support macOS by delegating to `ocaml/setup-ocaml`
- removed the drifting macOS-only OCaml setup path from `release.yml`
- removed redundant `ocaml-compiler: 5.4.1` workflow inputs so the composite action default is the single pin owner

## Why it changed
The release workflow built macOS with `5.4.x` through a separate path, which could silently drift away from the pinned Linux toolchain version.

## Impact
Workflow OCaml pinning is now centralized in the composite action, so release, deploy, and interop workflows all inherit the same default compiler version.

## Root cause
The repository had two setup paths: a Linux-only composite action pinned to `5.4.1`, and a macOS release path using `ocaml/setup-ocaml` with `5.4.x`.

## Validation
- parsed the modified workflow/action YAML locally
- `git diff --check`
- confirmed workflow files no longer carry direct `ocaml-compiler: 5.4.1` overrides outside the composite action

### DIFF
--- a/.github/actions/setup-ocaml-toolchain/action.yml
+++ b/.github/actions/setup-ocaml-toolchain/action.yml
@@ -1,5 +1,5 @@
 name: Setup OCaml Toolchain
-description: Shared Linux OCaml toolchain bootstrap that avoids upstream opam binary gaps.
+description: Shared OCaml toolchain bootstrap with Linux local-opam pinning and macOS parity.
 
 inputs:
   ocaml-compiler:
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
## PR #8497 — Unify OCaml workflow pinning through composite action

✅ 승인 권고. +18, CI infrastructure.

### 좋은 것
1. **macOS support via `ocaml/setup-ocaml` delegation** — 기존 `.github/actions/setup-ocaml-toolchain`이 Linux-only였음. macOS runner에서도 동일하게 pinning.
2. **Composite action pattern** — reusable action으로 centralization. Pinning change가 한 곳에서만 필요.
3. **+18** — minimal diff.

### 빈틈 1개
1. **macOS runner에서 actual CI pass 증거?** — PR body에 validation이 없음. macOS에서 `dune build`가 실제로 green인지 확인 필요.

### 사용자 표면 영향
**P3** — CI. macOS runner에서 OCaml workflow pinning 정상화. Direct user impact 없음.

PR #8655, opened 2026-04-19
fix(dashboard): cache keeper accountability summaries per request

jeong-sik:
## Summary
- build a request-local accountability summary lookup for the keeper dashboard so one render reads the 14-day accountability window once instead of once per keeper
- keep the legacy per-agent path for compact mode or low decision-layer mode, while routing the hot dashboard path through the cached lookup
- add regression coverage that proves the lookup path reads once and the legacy path still rereads per agent

## Product impact
- reduces dashboard render pressure by removing repeated 14-day accountability ledger rescans from the hot keeper overview path
- keeps compact mode and low decision-layer mode behavior unchanged while improving the default dashboard path
- lowers the risk of dashboard timeout pressure when keeper count grows

## Changed files
- `lib/dashboard/dashboard_http_keeper.ml`
- `lib/keeper/keeper_accountability.ml`
- `lib/keeper/keeper_accountability.mli`
- `lib/keeper/keeper_exec_status_metrics.ml`
- `lib/keeper/keeper_exec_status_metrics.mli`
- `test/test_keeper_accountability.ml`

## Review evidence
- Targeted build: `OCAMLPARAM='_,warn-error=-9' opam exec -- dune build --root . ./test/test_keeper_accountability.exe`
- Targeted tests: `_build/default/test/test_keeper_accountability.exe test --color=never --json accountability 4,5`
- Result: `success=2 failures=0`
- Regression guard: `summary lookup reads window once` asserts read count `1`, while `summary json rereads window per agent` asserts read count `2`

## Linked issue
- `Refs #7162`

## Notes
- lookup invalidation is request-local: the dashboard rebuilds the lookup on each render, so subsequent requests naturally re-read the latest accountability window
- compact mode and lower decision-layer mode intentionally keep the legacy per-agent summary path

### DIFF
--- a/lib/dashboard/dashboard_http_keeper.ml
+++ b/lib/dashboard/dashboard_http_keeper.ml
@@ -233,6 +233,14 @@ let keepers_dashboard_json ?(compact = false) (config : Coord.config) : Yojson.S
           (Printexc.to_string exn);
         []
   in
+  let accountability_summary =
+    if compact || Keeper_decision_audit.decision_layer_level () < 3 then
+      (fun ~keeper_name ~agent_name ->
+        Keeper_exec_status_metrics.accountability_summary_json config
+          ~keeper_name ~agent_name)
+    else
+      Keeper_exec_status_metrics.accountability_summary_lookup config
+  in
   (* Parallel keeper I/O: each keeper's metadata + metrics reads run concurrently.
      Results are collected into a shared ref array, then filter_map'd. *)
   let results = Array.make (List.length names) None in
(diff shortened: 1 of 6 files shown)


jeong-sik:
Verification note: local accountability tests look coherent, but this PR is not verifier-ready yet because CI is not green. The current run has `Build and Test` still in progress and `TLA+ Model Checking` failed on `KeeperStateMachine`. That failure appears outside this diff's file set (`dashboard_http_keeper`, `keeper_accountability`, `keeper_exec_status_metrics`, tests), so it may be an unrelated/spec-level blocker, but verification should wait until the required checks are green or the TLA failure is triaged explicitly.

PR #8657, opened 2026-04-19
refactor: room/namespace -> basePath single source of truth

jeong-sik:
## Summary
- unify room/namespace derived paths around `basePath` in coord and workflow guidance surfaces
- update schema and workflow guide references so generated guidance matches the single-source path model
- add coverage for coord output and workflow guide expectations

## Testing
- env -u DUNE_RPC opam exec -- dune runtest --root . test/test_tool_coord_coverage.exe
- env -u DUNE_RPC opam exec -- dune runtest --root . test/test_workflow_guide.exe

## Context
- task-119
- Epic #7261

### DIFF
--- a/lib/tool_coord.ml
+++ b/lib/tool_coord.ml
@@ -410,6 +410,7 @@ let inspect_state ctx =
 
 let state_to_json st =
   `Assoc [
+    ("project_ready", `Bool st.room_set);
     ("namespace_ready", `Bool st.room_set);
     ("room_set", `Bool st.room_set);
     ("joined", `Bool st.joined);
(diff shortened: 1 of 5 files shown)


jeong-sik:
Verification note: I do not see an immediate code bug in this diff, but the task mapping looks wrong. This patch is a readiness-key cleanup (`project_ready` becomes canonical while `namespace_ready` / `room_set` stay aliases) in `lib/tool_coord.ml`, `lib/tool_schemas/tool_schemas_coord_core.ml`, and `lib/workflow_guide.ml`, plus tests around that alias.

Epic #7261 is broader: bootstrap automation, collapsing the namespace axis into `config.base_path` across room/control surfaces, room-scoped tool semantics review, and eventual `lib/room` removal. I would not treat this PR as verification for the full `task-119` epic; it looks like a narrower follow-up that should be tracked independently or retitled accordingly.

Issue #8578, opened 2026-04-19
audit(keeper_state_machine): TLA+ ↔ OCaml event Variant drift — 4 names disagree


jeong-sik:
## Spec audit, FSM/TLA+/OCaml divergence

The Variant SSOT sweep verified that \`derive_phase\` priority order matches \`DerivePhase\` in \`KeeperStateMachine.tla\` exactly (14-step cascade, including the buffer-state suppression fix). The **event** Variant disagrees in four places:

| TLA+ \`KeeperStateMachine.tla\` | OCaml \`Keeper_state_machine.event\` | Status |
|----|----|----|
| \`HeartbeatOk\` | \`Heartbeat_ok\` | ✓ |
| \`HeartbeatFailed\` | \`Heartbeat_failed\` | ✓ |
| \`TurnSucceeded\` | \`Turn_succeeded\` | ✓ |
| \`TurnFailed\` | \`Turn_failed\` | ✓ |
| \`ContextMeasured\` | \`Context_measured\` | ✓ |
| \`CompactionStarted\` | \`Compaction_started\` | ✓ |
| \`CompactionCompleted\` | \`Compaction_completed\` | ✓ |
| **(absent)** | \`Compaction_failed\` | **OCaml only** |
| \`HandoffStarted\` | \`Handoff_started\` | ✓ |
| \`HandoffCompleted\` | \`Handoff_completed\` | ✓ |
| **(absent)** | \`Handoff_failed\` | **OCaml only** |
| \`OperatorPause\` | \`Operator_pause\` | ✓ |
| \`OperatorResume\` | \`Operator_resume\` | ✓ |
| **(absent)** | \`Operator_stop\` | **OCaml only** |
| \`StopRequested\` | \`Stop_requested\` | ✓ |
| \`DrainCompleteEv\` | \`Drain_complete\` | ✓ |
| \`FiberStarted\` | \`Fiber_started\` | ✓ |
| \`FiberTerminated\` | \`Fiber_terminated\` | ✓ |
| \`SupervisorRestartAttempt\` | \`Supervisor_restart_attempt\` | ✓ |
| \`RestartBudgetExhausted\` | \`Restart_budget_exhausted\` | ✓ |
| \`GuardrailStop\` | \`Guardrail_stop\` | ✓ |
| \`ContextOverflowDetected\` | \`Context_overflow_detected\` | ✓ |
| \`AutoCompactTriggered\` | \`Auto_compact_triggered\` | ✓ |
[...]

PR #8704, opened 2026-04-19
fix(keeper_exec_shell): Good:/Bad: examples for readonly-shell hints (#8688)

jeong-sik:
## Context

Triage on `~/me/.masc/tool_calls/2026-04/{17,18}.jsonl` (#8688) showed
**57 `command_blocked_readonly` rejections in 2 days** that the keeper
did not recover from — the previous hint named only the category
(`chaining`, `redirect`, `git_write`, `package_install`, `destructive`),
so small-LLM keepers retried the same blocked command verbatim.

The existing `validate_gh_command` and empty-`command` / empty-`pattern`
errors already carry `Good:` / `Bad:` concrete examples. This PR extends
the same pattern to `readonly_hint_of_category`.

## Change

- `hint_of_category` was a closure inside the `bash` op arm — hoisted to
  module-level `readonly_hint_of_category` and exposed via `.mli`.
- Every named category branch now ends with a concrete
  `Good: keeper_bash cmd='...'. Bad: keeper_shell command='...'.` pair.
- New unit test `test_keeper_readonly_hints` locks the invariant that
  each named category carries both `Good:` and `Bad:` substrings, and
  that the unknown-category fallback does not falsely advertise an
  example.

## Why it should land

- 57/2d call rejections → concrete rewrite at the wire. Follows
  memory rules `feedback_gate-response-llm-readable.md` and
  `feedback_tool-error-messages-teach-llm.md`.
- No behaviour change: classification and blocking logic are untouched;
  only the hint string changes + helper is exported.
- Ratchet test locks the invariant so future contributors cannot
  silently regress the hint to a category-only form.

## Tests

```
dune test --root . test/test_keeper_readonly_hints.exe
  [OK]  all named categories
  [OK]  unknown category fallback
```

Full `dune build @install` green.

Closes a piece of #8688 (root-cause sweep).

### DIFF
--- a/lib/keeper/keeper_exec_shell.ml
+++ b/lib/keeper/keeper_exec_shell.ml
@@ -152,6 +152,40 @@ let readonly_shell_token_match tokens =
       Some ("curl --output", "destructive")
   | _ -> None
 
+(* Each branch ends with concrete Good:/Bad: examples so small-LLM keepers
+   can self-correct without a retry loop. Prior form only named the
+   category, which left 57 command_blocked_readonly rejections on
+   2026-04-17/18 without a wire-level rewrite. See masc-mcp#8688. *)
+let readonly_hint_of_category = function
+  | "chaining" ->
+      "`&&`, `||`, and `;` chaining are blocked in readonly shell. \
+       Issue one command per keeper_shell call, or use a dedicated \
+       sub-op: git_log, git_status, git_diff, git_worktree, find, \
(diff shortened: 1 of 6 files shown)


jeong-sik:
Blocked by TLA+ regression on main tracked in #8727. This PR's TLA+ failure is not caused by the PR content (readonly-shell hint strings) but by an S11 invariant conflict introduced in #8578/#8580. Deferring to #8727 for the spec-side fix.

PR #8716, opened 2026-04-19
[codex] Bump OAS pin for Gemini MCP sentinel fix

jeong-sik:
## Summary

- Bump the MASC `agent_sdk` runtime pin to OAS `54f4aeabd0ff88b9666d1c3ecc35aee26dbc894f`.
- Refresh the generated Keeper manual pin snapshot.

## Why

OAS PR #1048 fixed the Gemini CLI 0.38 PolicyEngine crash caused by passing `--allowed-mcp-server-names ""`. MASC was still pinned to `3b0409d265405ce294c271f646b819dfca958683`, which predates that sentinel fix, so keeper cascades could still fail after restart when `OAS_GEMINI_NO_MCP=1` is applied.

## Validation

- `bash scripts/sync-oas-pin-docs.sh --check`
- `bash scripts/check-oas-pin.sh`
- `git diff --check`

Note: `scripts/opam-pin-external-deps.sh` was run locally to align the opam pin source for validation; it did not rebuild installed packages because it was run without `--install`.


jeong-sik:
CI status update:

- `OAS Pin Consistency`: pass
- `Doc Truth`: pass
- `Health`: pass
- `Lint`: pass
- `Build and Test`: failed twice in `Run tests (quick suite)` by timeout only

Evidence:
- Initial job `72022383408` timed out after `CI_TEST_TIMEOUT_SEC=1200`.
- Failed-job rerun `72024236827` timed out the same way.
- Both attempts passed dependency pinning, dependency install, and `dune build` before entering `opam exec -- dune test`.
- Timeout diagnostics show no failure markers; latest suites are around `keeper_tool_matrix` / `Tool_code_read_containment`, with successful/skipped output before heartbeat-only timeout.

This looks like an existing CI quick-suite timeout/hang rather than a failure caused by this pin-only diff. Tracked separately as #8748.

Issue #8892, opened 2026-04-20
MASC task lifecycle mutations fail credential lookup after claim-first cancellation


jeong-sik:
## Symptom

`task-161` remains `todo/unclaimed` in the MASC board even though GitHub source of truth shows Slice A is complete via PR #8698:

- PR #8698 merged at `fa1ae97e9c5c2e9d3df85206a0edd16654913d06`
- head `e79bbb1ebd83cff295b195ce30fe13dc6b516336`
- workflow run `24635314106` completed/success for required paths
- Slice B is correctly tracked separately in #8745 / task-162
- partial-complete read-model follow-up is tracked in #8885

The remaining lifecycle action should be a credentialed close/rebind of `task-161`, but the available MASC mutation path fails before it can act.

## Reproduction

From the current MASC surface on 2026-04-20 KST:

1. `masc_claim_next(agent_name="codex-mcp-client")`
   - result: `user cancelled MCP tool call`
2. Check live state:
   - `task-161` still appears as `todo/unclaimed`
   - current surface has no owned task and may show stale `current=task-044/task-056` depending on identity mapping
3. Attempt scoped lifecycle mutation:
   - `masc_transition(task-161, claim, agent_name="adversary-eager-viper")`
   - result: `Unauthorized: No credential found for adversary-eager-viper`
4. Attempt coordination fallback:
   - `masc_broadcast(...)`
   - observed earlier on the same thread: `No credential found for adversary-eager-viper`

## Expected

A credentialed keeper/operator should be able to either:

- close/rebind `task-161` with PR #8698 evidence, and route Slice B to #8745/task-162, or
- clearly report that the caller is unauthenticated before the system asks that caller to perform claim-first/lifecycle work.

## Actual

The task board keeps surfacing `task-161` as todo while multiple keeper surfaces can verify the GitHub evidence but cannot mutate lifecycle state or broadcast a handoff due credential lookup failure.

## Estimated Cause

Likely identity/credential routing mismatch between:

- displayed/joined agent aliases such as `adversary-eager-viper`, `codex-mcp-client`, `minjae-deft-wolf`, `cheolsu-proud-falcon`
- mutation credential lookup used by `masc_transition` / `masc_broadcast`
- claim-first routing guidance shown in `masc_status`

## Attempted

- Duplicate search before filing:
  - `credential routing No credential found adversary-eager-viper task-161 MASC transition`
  - `task-161 lifecycle close rebind credential keeper_task_claim`
[...]

PR #8897, opened 2026-04-20
Surface credential-blocked task lifecycle guidance

jeong-sik:
## Summary
- add credential-aware status guidance for auth-required rooms
- suppress claim/transition suggestions when the current/resolved identity has no credential
- add regression coverage for credential-blocked lifecycle guidance

## Product impact
Operators currently see lifecycle suggestions such as `masc_claim_next` or `masc_transition` even when the mounted identity cannot pass credential lookup. This PR changes the status/read-model surface so credential-blocked sessions get an explicit credential warning instead of action guidance that will fail.

## Evidence
- Fixes the read-model side of #8892.
- Adds a `Credential` status line when auth requires a credential and no credential is available for the resolved identity candidates.
- Suppresses lifecycle suggestions for credential-blocked identities while leaving non-lifecycle status guidance intact.

## Out of scope
- Direct `masc_claim_next` credential preflight behavior.
- Direct `masc_transition` credential preflight behavior.
- Scheduled keeper / Immediate Task Move prompt behavior that still tells credential-blocked surfaces to claim first.

Those control-plane and autonomous prompt paths remain part of #8892 follow-up unless a later patch adds direct lifecycle mutation regressions or an explicit out-of-scope decision.

## Review evidence
- `dune build --root . test/test_tool_coord_coverage.exe`
- `dune exec --root . test/test_tool_coord_coverage.exe`

## Linked issue
Refs #8892

### DIFF
--- a/lib/tool_coord.ml
+++ b/lib/tool_coord.ml
@@ -73,6 +73,49 @@ let option_or_dash = function
   | Some value when String.trim value <> "" -> value
   | _ -> "-"
 
+let lifecycle_tools =
+  [
+    "masc_claim_next";
+    "masc_transition";
+  ]
+
+let is_lifecycle_tool tool =
+  List.exists (String.equal tool) lifecycle_tools
+
+let unique_strings items =
+  List.fold_left
+    (fun acc item ->
+      let item = String.trim item in
+      if item = "" || List.exists (String.equal item) acc then acc
+      else item :: acc)
+    [] items
+  |> List.rev
+
+type credential_state = {
+  credential_required : bool;
+  credential_available : bool;
+  credential_candidates : string list;
+}
+
+let credential_state (ctx : context) ~actual_name =
+  let auth_cfg = Auth.load_auth_config ctx.config.base_path in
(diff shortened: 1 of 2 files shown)


jeong-sik:
Review note after checking the patch and the live `task-161` thread:

`#8897` is a valid read-model/status slice for #8892, but it should not be treated as the full control-plane fix.

What this PR does cover:
- `masc_status` now surfaces credential-required / credential-missing state.
- Lifecycle suggestions (`masc_claim_next`, `masc_transition`) are suppressed when the resolved identity has no credential.
- Regression coverage locks that status/suggestion behavior.

What is still not covered:
- Direct `masc_claim_next` no-credential preflight behavior.
- Direct `masc_transition` no-credential preflight behavior.
- A regression proving scheduled keepers cannot bypass the status read-model and hit the same credential failure path.

So the safe merge wording is: **read-model/status mitigation for #8892**, pending terminal CI. Do not close #8892 as fully fixed unless a later patch adds direct lifecycle mutation regressions or explicitly documents that control-plane behavior remains out of scope.

PR #8909, opened 2026-04-20
test: cover lifecycle auth preflight

jeong-sik:
## Summary

Add MCP execute-level regressions for the remaining #8892 control-plane credential path.

This covers direct lifecycle tool calls from an auth-required room when no credential/token is available:

- `masc_claim_next` rejects before claiming a task
- `masc_transition(action=claim)` rejects before mutating a task
- both cases leave the task in `todo` and do not bind planning `current_task`

## Why

PR #8897 merged the read-model/status mitigation for #8892, but it explicitly left direct lifecycle mutation entrypoints out of scope. These tests lock the existing server auth preflight behavior so the direct mutation path cannot regress while #8892 remains open for the broader credential-routing design.

## Verification

- `opam exec -- dune build --root "$PWD" test/test_mcp_server_eio.exe`
- `opam exec -- _build/default/test/test_mcp_server_eio.exe --color=never` (61 tests)
- `git diff --check`

Refs #8892

### DIFF
--- a/test/test_mcp_server_eio.ml
+++ b/test/test_mcp_server_eio.ml
@@ -1530,6 +1530,84 @@ let test_execute_tool_generated_agent_name_uses_token_identity () =
 
   cleanup_dir base_path
 
+let check_task_still_todo config task_id =
+  match
+    Masc_mcp.Coord.get_tasks_raw config
+    |> List.find_opt (fun (task : Types.task) -> task.id = task_id)
+  with
+  | Some { Types.task_status = Types.Todo; _ } -> ()
+  | Some task ->
+      Alcotest.failf "expected %s to remain todo, got %s" task_id
+        (Types.task_status_to_string task.task_status)
+  | None -> Alcotest.failf "expected task %s to exist" task_id
+
+let check_auth_preflight_result ~tool_name ok msg =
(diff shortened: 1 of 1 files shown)


jeong-sik:
Evidence check on the current patch:

This PR now covers the direct MCP execute-level auth preflight for task/planning mutation:
- `masc_claim_next` rejects before claiming.
- `masc_transition(action=claim)` rejects before mutating.
- task remains `Todo`.
- planning `current_task` remains unset.

Remaining acceptance gap: the fixtures pre-join `uncredentialed-agent`, so they do not prove that a rejected request from a non-joined/generated credentialless identity avoids agent/presence materialization. That was part of the observed failure class: credentialless derived identities can remain visible and later drive status/prompt guidance.

Recommended addition before treating this as full #8892 coverage:
- a non-joined/generated identity fixture for rejected `masc_claim_next`.
- a non-joined/generated identity fixture for rejected `masc_transition`.
- assertions that no new agent/presence file or active-agent entry is created, in addition to task/planning remaining unchanged.

Current CI state from run `24645777032`: PR Hygiene is green; `Lint`, `TLA+`, `Health`, and `Build and Test` are still in progress. So this is a valid direct preflight PR, but not terminal evidence yet.

Issue #8605, opened 2026-04-19
[BUG] execution_scope_of_string returns Limited_code_change for ANY unknown input — silent privilege miscategorization (REAL bug, 3 copies)


jeong-sik:
## Summary

\`execution_scope_of_string\` exists in **three modules** with two different shapes. Two of the three are **dangerously permissive** — unknown input falls through to a *valid* variant (\`Limited_code_change\`), masking typos as silent miscategorization.

\`\`\`ocaml
(* lib/worker_types.ml:27 *)
let execution_scope_of_string = function
  | "observe_only" -> Observe_only
  | "autonomous" -> Autonomous
  | _ -> Limited_code_change   (* <-- ANYTHING unknown becomes this *)

(* lib/worker_contract_types/worker_contract_types_enums.ml:114 *)
(* identical body, dangerous default *)

(* lib/worker_runtime_config.ml:68 — only the third one is correct *)
let execution_scope_of_string_opt value =
  match String.lowercase_ascii (String.trim value) with
  | "observe_only" -> Some Worker_types.Observe_only
  | "limited_code_change" -> Some Worker_types.Limited_code_change
  | "autonomous" -> Some Worker_types.Autonomous
  | _ -> None
\`\`\`

The third one is correct: it's Option-typed and accepts all 3 valid strings. The other two:
- only accept 2 of the 3 valid strings — \`\"limited_code_change\"\` itself returns \`Limited_code_change\` because it falls through the catch-all (works by accident)
- silently miscategorize ANY typo: \`\"observed\"\` → \`Limited_code_change\`, \`\"autonomus\"\` → \`Limited_code_change\`, \`\"\"\` → \`Limited_code_change\`
- are inconsistent with the other 8 variants in \`worker_contract_types_enums.ml\` which use \`Some/None\` (\`turn_kind_of_string\`, \`worker_class_of_string\`, etc.)

## Impact

\`execution_scope\` controls keeper privilege:
- \`Observe_only\` — read-only
- \`Limited_code_change\` — gated mutations
- \`Autonomous\` — full mutating tools

A typo in a config file (\`\"observe-only\"\` with hyphen) downgrades silently from Observe_only intent to Limited_code_change. **Privilege miscategorization is a security-relevant silent failure.**

[...]

PR #8937, opened 2026-04-20
fix(lint): self-verify allowlist — fail on stale #8605 family entries

jeong-sik:
## Summary

The `#8605 family` lint allowlist is a debt ledger: `path:line` entries for pre-existing permissive-default sites that must be migrated. Problem: when an upstream PR eliminates the pattern at `path:line`, the allowlist entry often gets left behind. The ledger grows stale, new violations can hide behind old listings, and line-number drift during refactors has already caused one live incident (cycle 26 / `#8867`).

This PR makes the lint script **self-verify** the allowlist on every run.

## Changes

- `scripts/lint/no-unknown-permissive-default.sh`:
  - After main scan, collect all detected `file:line` keys into a set.
  - Cross-check every allowlist entry against that set.
  - Exit `2` if any entry points to a line where no `#8605` family pattern is detected ("stale allowlist entry").
  - Exit `1` unchanged (new violations outside allowlist).
  - Exit `0` requires: no new violations AND all allowlist entries valid.
  - Report live debt count (`Allowlist debt: N entries`) on clean runs.
  - Opt out: `SKIP_ALLOWLIST_VERIFY=1` for transient in-flight migrations that list lines before the fix lands.

- `scripts/lint/no-unknown-permissive-default.allowlist`:
  - Remove stale `lib/sdk_tool_contract.ml:377` — migrated to `param_type_of_schema_opt` + warn-wrapper upstream, the permissive default no longer exists at that line. Surfaced by self-verify on its first run.
  - Consolidate history comments for traceability.

## Why this matters

User frame across cycles 20 → 35:

> 수치화된 경험이 있다면 반복하지 않을 수 있다. 결정론 가정(line numbers) + silent drift = 허무맹랑한 스트링 매칭과 같은 anti-pattern.

The allowlist itself was a DET assumption: "line N in file F still contains the pattern". Upstream migrations invalidate that assumption silently. Self-verify converts the silent drift into an observable CI signal — same principle as the soft-migration `warn on unknown default` pattern the family PRs have been using (#8911, #8916, #8922).

## Empirical proof

Running the new verification against `main` (pre-cleanup):

```
::error file=scripts/lint/no-unknown-permissive-default.allowlist::stale allowlist entry: lib/sdk_tool_contract.ml:377 (no #8605 family pattern detected at this file:line; remove from allowlist)
EXIT=2
```

`lib/sdk_tool_contract.ml:377` was migrated to sound-partial (`param_type_of_schema_opt` returns `None`). Allowlist entry should have been removed in the migration PR, but wasn't — exactly the drift this verification catches. One silent stale entry existed for the entire time since the migration landed.

## Test plan

Three exit paths verified locally:

- [x] Clean state (1 valid allowlist entry) → `EXIT=0`, "Allowlist debt: 1 entry (allowlisted)."
- [x] Injected `lib/nonexistent.ml:999` into allowlist → `EXIT=2` with stale-entry error pointing to the exact line.
- [x] Same injection with `SKIP_ALLOWLIST_VERIFY=1` → `EXIT=0` (opt-out works for migrations).
- [x] New violation (not in allowlist) → `EXIT=1` (unchanged behavior from existing tests).

## Scope

- Script: +66 lines (self-verify block + exit code doc).
- Allowlist: sdk_tool_contract.ml:377 entry removed; history comments consolidated.
- No `lib/` code changes. Workflow YAML untouched (single gate step covers both checks).

## Risk

Low. Incremental lint tightening, self-contained script. The one breaking-case (existing stale entries) is cleaned up in this PR — no future PR should see false negatives from self-verify unless an allowlist entry is newly added for a line that doesn't actually match the pattern.

### DIFF
--- a/scripts/lint/no-unknown-permissive-default.allowlist
+++ b/scripts/lint/no-unknown-permissive-default.allowlist
@@ -8,17 +8,23 @@
 # File new instances as standalone bugs (see #8832 as an example).
 
 # #8605 umbrella — tracked in https://github.com/jeong-sik/masc-mcp/issues/8605
-# sdk_tool_contract.ml:377 -- stale entry (line 377 is `Some Boolean`,
-# the actual `_ -> None` was already migrated by #8832 to a sound-partial
-# `param_type_of_schema_opt` with a warn-and-default wrapper).
-# activity_graph_types.ml:89 (span_status_of_string) eliminated by sound-partial
-# template (mirrors #8779 node_status).
-# coord.ml:132 (observe_task_transition) eliminated by #8846.
-# coord.ml:88/109 (observe_agent_lifecycle) eliminated by #8842
-# (agent_lifecycle_event variant: join/rejoin/leave).
-# eval_harness.ml:370 eliminated by #8889 (lifted nested string match to
-# outer pattern; was a provably unreachable dead branch).
-# types_auth.ml:91 (category_for_tool) eliminated by this PR -- sound-partial
-# `category_for_tool_opt` + back-compat wrapper preserves GeneralLimit default.
 #
-# Allowlist now empty.  Remove this file or keep as a lint contract record.
+# All install-time entries eliminated by upstream migrations. The lint script
+# now self-verifies this file: if an entry points to a line that no longer
+# matches the #8605 family pattern, CI fails with exit 2 (see script header).
+#
+# Historical record of the drain:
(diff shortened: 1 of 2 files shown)


jeong-sik:
Rebased onto main after #8935 drained the last allowlist entry (types_auth.ml:91). This PR is now script-only + allowlist comment refresh — self-verify exits 0 against a 0-entry ledger. Scope: +73 / −15 across script and allowlist (no lib changes).

PR #9172, opened 2026-04-21
[codex] fix dashboard observer mcp sse auth

### LINKED ISSUE #9174: Dashboard observer SSE auth drift can show "신호 없음" on first load
## Summary
The dashboard can show `신호 없음` in the side-rail health indicator even while the server is healthy and SSE infrastructure is up.

## Symptom
- dashboard renders normally
- server `/health` is `ok`
- observer SSE sessions may exist globally
- the current browser tab still reports `connected=false` and shows `신호 없음`

## Reproduction
1. run the dashboard in a loopback/dev-token setup with auth enabled
2. open a fresh dashboard tab without an existing `sessionStorage` bearer token
3. let the page boot normally
4. observe that the page opens observer SSE before the dashboard has a token available

## Root cause
- the dashboard uses the canonical `/mcp?sse_kind=observer` stream
- `EventSource` cannot send `Authorization` headers
- the client relied on `?token=` for observer SSE
- the server-side `/mcp` observer GET path only accepted header-based bearer auth
- in addition, the dashboard boot path could attempt SSE before dev-token bootstrap completed

## Expected fix direction
- keep `/mcp` as the canonical observer SSE path
- allow observer SSE auth to use the existing query-token contract without widening auth for general MCP/mutation routes
- bootstrap dev-token before opening the observer SSE stream on first page load

jeong-sik:
## Summary
Fix dashboard observer SSE auth drift so a fresh dashboard tab does not show `신호 없음` just because the observer stream opened before auth was available.

## What changed
- allow dashboard observer SSE on `/mcp?sse_kind=observer` to authenticate with the existing `token` query param contract, without widening header-free auth for other MCP paths
- bootstrap the dashboard dev token before opening the observer SSE stream on first page load
- add auth coverage tests for observer-stream query token acceptance and non-observer rejection

## Why
The dashboard health chip could show `신호 없음` even while the server was healthy because the page opened the observer SSE stream before the dashboard had a bearer token available, and the server-side `/mcp` observer GET path only accepted `Authorization` headers.

`EventSource` cannot send custom auth headers, so the dashboard and server contracts had drifted: the client used `/mcp?...&token=...`, but the server ignored the query token for observer SSE auth.

## Product impact
- first-load dashboard tabs should connect their observer SSE stream reliably in loopback/dev-token flows
- the fix stays scoped to observer SSE and does not relax auth for general MCP, POST, or mutation routes
- deprecated `/sse` is not used by the dashboard path; the canonical `/mcp` observer stream now matches the dashboard contract

## Evidence
- local reproduction showed `/mcp` observer SSE failed auth when opened without a bearer header even though the dashboard client used the query-token contract
- server auth path now has observer-SSE-specific coverage and non-observer rejection coverage in `test_auth_coverage`
- dashboard bootstrap now requests the dev token before opening the observer SSE tail on initial load

## Review evidence
- `dune exec --root . ./test/test_auth_coverage.exe`
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/vitest run src/api/mcp.test.ts src/sse.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1`

## Linked issue
Fixes #9174

### DIFF
--- a/dashboard/src/api/mcp.ts
+++ b/dashboard/src/api/mcp.ts
@@ -33,7 +33,7 @@ let devTokenBootstrapPromise: Promise<void> | null = null
     only exposes `/api/v1/dashboard/dev-token` when bound to loopback with
     strict-auth overrides disabled; in every other case this quietly no-ops
     and existing flows (URL `?token=…`, manual paste) continue to work. */
-async function ensureDevToken(): Promise<void> {
+export async function ensureDevToken(): Promise<void> {
   if (getStoredToken()) return
   if (devTokenBootstrapPromise) return devTokenBootstrapPromise
   devTokenBootstrapPromise = (async () => {
--- a/dashboard/src/app.ts
+++ b/dashboard/src/app.ts
@@ -18,6 +18,7 @@ import { refreshForRoute } from './tab-refresh'
 import { refreshMissionSnapshot } from './mission-store'
 import { replayOasRuntimeTelemetry } from './oas-runtime-store'
 import { refreshShell } from './store'
+import { ensureDevToken } from './api/mcp'
 import {
   BuildIdentityBadge,
   ConnectionStatus,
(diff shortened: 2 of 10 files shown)


jeong-sik:
Addressed the PR hygiene warning.

- added `Summary`, `Product impact`, `Evidence`, `Review evidence`, and `Linked issue` sections to the PR body
- linked the regression issue via `Fixes #9174`

Proceeding with ready-for-review and auto-merge setup.

PR #9210, opened 2026-04-21
[codex] Follow up runtime cascade snapshot gate after #9200

jeong-sik:
## Summary
- preserve validated runtime snapshot as the only dashboard-visible cascade catalog after #9200
- keep TopK rerank provider selection behind the validated runtime catalog gate
- retain typed SDK compatibility fixes needed after latest main drift

## Why
- #9200 merged while the branch was being rebased onto newer main commits
- after replaying the merge against current main, 12 files still differed in behavior from main
- the remaining delta is not merge noise: it restores runtime-authoritative snapshot behavior and related keeper/dashboard guards

## Validation
- dune build --root . @install
- dune exec --root . ./test/test_cascade_catalog_runtime.exe
- dune exec --root . ./test/test_dashboard_cascade.exe
- dune exec --root . ./test/test_keeper_runtime_config_ssot.exe
- dune exec --root . ./test/test_oas_worker.exe
- dune exec --root . ./test/test_keeper_topk_llm.exe
- dune exec --root . ./test/test_keeper_unified.exe

Refs #7872
Follow-up to #9200

### DIFF
--- a/lib/keeper/keeper_cascade_profile.ml
+++ b/lib/keeper/keeper_cascade_profile.ml
@@ -152,6 +152,13 @@ let normalize_declared_name (raw : string) : string =
   | None when trimmed = "" -> default_name
   | None -> trimmed
 
+let normalize_declared_name (raw : string) : string =
+  let trimmed = String.trim raw in
+  match of_string_opt trimmed with
+  | Some t -> to_string t
+  | None when trimmed = "" -> default_name
+  | None -> trimmed
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing as obsolete. I rebased this locally onto current main and the only branch-unique change left was a duplicate normalize_declared_name addition in keeper_cascade_profile; current main already contains the intended helper without the duplicate, so there is no remaining delta worth carrying.

PR #9239, opened 2026-04-21
fix(keeper): exhaustive task_status matches in world observation

jeong-sik:
## Summary
- Replace two `_ -> false` wildcard matches in `keeper_world_observation.ml` task counting predicates with explicit constructors
- Line 224: `Cancelled _ -> true | _ -> false` → explicit 5-branch match
- Line 232: `AwaitingVerification _ -> true | _ -> false` → explicit 5-branch match

## Why
Wildcard `_` catches any new `task_status` variant silently. Exhaustive matches force the compiler to flag additions, making the type system work for us.

## Test plan
- [x] `dune build` passes
- [ ] CI green

### DIFF
--- a/lib/keeper/keeper_world_observation.ml
+++ b/lib/keeper/keeper_world_observation.ml
@@ -221,7 +221,10 @@ let read_backlog_counts ~(config : Coord.config) : int * int * int =
       List.length
         (List.filter
            (fun (t : Types.task) ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Blocked on infra/workflow issue #9247.

`TLA+ Model Checking (check-clean)` is cancelling near `SocialStateCap` at the current 15 minute timeout, so `CI Gate` fails even though the non-TLA lanes are green.

PR #9240, opened 2026-04-21
fix(keeper): exhaustive fiber_health match in exec status

jeong-sik:
## Summary
- Replace `_ ->` wildcard on `fiber_health` variant match in `keeper_exec_status.ml:375`
- Explicit `Fiber_alive | Fiber_unknown` branches ensure compiler flags new variants

## Why
`fiber_health` has 4 constructors. The wildcard silently accepted any future additions.

## Test plan
- [x] `dune build` passes
- [ ] CI green


jeong-sik:
Blocked on infra/workflow issue #9247.

`TLA+ Model Checking (check-clean)` is cancelling near `SocialStateCap` at the current 15 minute timeout, so `CI Gate` fails even though the non-TLA lanes are green.

PR #9275, opened 2026-04-21
[codex] bump OAS pin and wire Kimi provider kinds

jeong-sik:
## Summary
- bump the OAS pin to upstream main  and sync generated pin docs
- accept upstream ╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│   ▐█▛█▛█▌  Welcome to Kimi Code CLI!                                         │
│   ▐█████▌  Send /help for help information.                                  │
│                                                                              │
│  Directory: ~/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-kimi-oas-pr  │
│  ovider                                                                      │
│  Session: 0e808c79-84a8-4d40-bedd-b249ec6a7f25                               │
│  Model: not set, send /login to login                                        │
│                                                                              │
│  Tip: Spot a bug or have feedback? Type /feedback right in this session — e  │
│  very report makes Kimi better.                                              │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯

── input ───────────────────────────────────────────────────────────────────────
 
 and  provider kinds across MASC adapter and worker surfaces
- keep the existing Moonshot compatibility lane for legacy  configs and cover it with regression tests

## Why
The OAS bump that adds first-class Kimi support also extends . Without the downstream MASC updates, the new pin breaks exhaustiveness checks and the build fails before the new runtime can be used.

## Validation
- bash scripts/opam-pin-external-deps.sh --install
- bash scripts/check-oas-pin.sh --local-only
- dune build --root . _build/default/test/test_provider_kind_resolution.exe _build/default/test/test_oas_worker.exe _build/default/bin/main_eio.exe
- ./_build/default/test/test_provider_kind_resolution.exe
- ./_build/default/test/test_oas_worker.exe test cascade_config 13

## Notes
-  did not complete in this shell environment, so live-config end-to-end validation is still pending.
- The Moonshot compatibility shim is intentionally retained for backward compatibility with older live cascade configs.


jeong-sik:
Superseded by #9273, which is already merged.

Issue #9279, opened 2026-04-21
Build breaks on unhandled Kimi/Kimi_cli provider variants


jeong-sik:
## Symptom
`dune build` and targeted `dune exec` commands fail before PR-local tests run.

## Reproduction
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-coord-query-exhaustive test/test_room.exe`
- `dune exec --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-cascade-partial-validation test/test_server_runtime_bootstrap.exe`

## Current error surface
Warning 8 promoted to error for non-exhaustive matches missing `Kimi | Kimi_cli` in:
- `lib/oas_worker_exec.ml`
- `lib/cascade/cascade_config.ml`
- `lib/provider_adapter.ml`

## Suspected cause
`provider_kind` gained `Kimi` / `Kimi_cli`, but several pattern matches were not updated when the new variants landed.

## Tried
- Re-ran targeted verification from two separate PR worktrees to confirm it is repo-global rather than branch-local.
- Confirmed the failure happens before the modified PR files are exercised.

jeong-sik:
라벨 없음. 이슈 triage 필요.

- type:bug? type:build?
- area:provider? area:build?
- priority?

@sojin

jeong-sik:
이슈에 기재된 3개 파일 모두 현재 main에서 이미 처리되어 있음.

- `lib/provider_adapter.ml`: #9287에서 `docker_auth_env_keys_of_provider_config` match에 `Kimi` / `Kimi_cli` 추가.
- `lib/cascade/cascade_config.ml`: 이미 `Kimi` / `Kimi_cli` 처리됨.
- `lib/oas_worker_exec.ml`: 현재 `provider_kind` non-exhaustive match 없음.

`dune build` clean. 이슈 해결 완료로 간주하고 닫습니다.

PR #9281, opened 2026-04-21
fix(startup): keep partial catalog startup ready

jeong-sik:
## Summary
- treat `Serving_valid_subset` as a startup warning instead of a degraded startup error
- add a bootstrap regression test that keeps `/health.startup.phase` at `ready`
- keep dashboard validation status on the existing `serving_valid_subset` contract

## Why
The partial-catalog flow is already runtime-safe after #9264, but startup still marked the server degraded because `Serving_valid_subset` flowed through the same `catalog_validation_error` path as last-known-good fallback.

## Validation
- direct GLM review completed for the parent PR scope
- local targeted `dune exec` is currently blocked by repo-global Kimi/Kimi_cli exhaustiveness errors tracked in #9279

### DIFF
--- a/lib/server/server_runtime_bootstrap.ml
+++ b/lib/server/server_runtime_bootstrap.ml
@@ -735,11 +735,14 @@ let run ~sw ~env ~host ~port ~base_path ~make_routes ~make_request_handler
             None
         | Ok
             (Cascade_catalog_runtime.Serving_valid_subset
-               { rejected_update; _ }) ->
-            Some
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing as superseded: current `main` already contains the partial-catalog startup-ready handling plus the matching `test_server_runtime_bootstrap` coverage, so rebasing this PR leaves no meaningful diff beyond whitespace-only blank-line deletions.

PR #9347, opened 2026-04-21
prompt: align gh guidance with sandbox fallback

jeong-sik:
## Summary
- align keeper prompt and keeper_shell schema guidance with the sandbox gh fallback added in Axis 2
- teach the work-discovery nudge to recommend explicit --repo owner/name for sandbox keepers without a claimed task
- include the current main build unblock for Llm_provider.Retry.NotFound in oas_worker_named

## Verification
- dune build --root . -j 1 test/test_keeper_unified.exe
- ./_build/default/test/test_keeper_unified.exe test unified_prompt
- dune build --root . -j 1 test/test_oas_worker.exe
- ./_build/default/test/test_oas_worker.exe test cascade_config

## Notes
- test_keeper_unified unified_prompt is green, including the gh guidance/source-file contract cases
- test_oas_worker cascade_config still has one existing unrelated failure: Claude CLI limit wrapper counts as hard quota (case 15)

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -1606,8 +1606,8 @@ let run_turn
               Preferred keeper tools (copy the name and schema verbatim):\n\
              \  - `keeper_task_claim` {} \
               — reserve the next eligible task\n\
-             \  - `keeper_shell` { op: \"gh\", cmd: \"pr list --state open\" } \
-              — inspect GitHub PRs/issues/checks after you already hold a claimed task/current_task_id\n\
+             \  - `keeper_shell` { op: \"gh\", cmd: \"pr list --repo owner/name --state open\" } \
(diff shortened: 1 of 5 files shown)


jeong-sik:
`#9353` now overlaps this PR in two places:

- `lib/oas_worker_named.ml`: the `Retry.NotFound` / Claude CLI hard-quota follow-up is already covered there
- keeper gh guidance: this PR moves the contract toward sandbox `--repo owner/name` fallback, while `#9353` keeps the current claim-first messaging and only tightens runtime error hints around that contract

So this is no longer an independent merge candidate as-is. The remaining delta here is mostly the keeper gh guidance policy change, and that conflicts with the current prompt/schema/runtime direction in `#9353`.

Practical next step: either close this as superseded, or trim it down to only the sandbox-fallback policy change if that is still the intended contract.

jeong-sik:
Superseded by #9353. The  follow-up is already covered there, and the remaining keeper gh guidance delta should be decided against the current claim-first contract in one place instead of splitting across two PRs.

Issue #9348, opened 2026-04-21
sdk_error_is_hard_quota misses Claude CLI limit wrapper


jeong-sik:
## Symptom
`test/test_oas_worker.exe test cascade_config` fails in `sdk_error_is_hard_quota detects Claude CLI limit wrapper`.

## Reproduction
```bash
dune build --root . -j 1 test/test_oas_worker.exe
./_build/default/test/test_oas_worker.exe test cascade_config
```

Observed failure:
- case 15: `Claude CLI limit wrapper counts as hard quota`
- expected `true`, received `false`

## Suspected cause
`lib/oas_worker_named.ml` only recognizes CLI quota wrappers via these markers:
- `terminalquotaerror`
- `quota_exhausted`
- `exhausted your capacity on this model`
- `quota will reset after`

The Claude CLI wrapper used in the test contains `You've hit your limit · resets ...`, which is not currently matched.

## What I tried
- reproduced locally while verifying PR #9347
- confirmed the failing fixture in `test/test_oas_worker.ml`
- confirmed the current heuristic list in `lib/oas_worker_named.ml`

## Scope
This issue is separate from PR #9347, which only aligns gh sandbox guidance and includes a compile unblock for `Retry.NotFound`.

jeong-sik:
라벨 없음. 이슈 triage 필요.

- type:bug?
- area:sdk? area:provider?
- priority?

@sojin

jeong-sik:
full suite (94 tests) passes on current main, including the specific  case. The Claude CLI limit wrapper pattern is already recognized. Closing as already fixed.

PR #9349, opened 2026-04-21
fix: detect Claude CLI hard quota wrapper

jeong-sik:
## Summary
- recognize Claude CLI quota wrappers that say `You've hit your limit` in `sdk_error_is_hard_quota`
- restore the existing `cascade_config` hard-quota expectation for the Claude CLI fixture
- include the current main compile unblock for `Llm_provider.Retry.NotFound` in `oas_worker_named`

Refs #9348

## Verification
- dune build --root . -j 1 test/test_oas_worker.exe
- ./_build/default/test/test_oas_worker.exe test cascade_config


jeong-sik:
Superseded by current main: the not-found compatibility and hard-quota wrapper handling are already covered more completely via merged changes, so this branch is now stale and conflicting.

PR #9351, opened 2026-04-21
[codex] Handle Retry.NotFound in oas worker

jeong-sik:
## What changed
- handle `Llm_provider.Retry.NotFound` in `sdk_error_to_cascade_outcome` as HTTP 404
- treat `Retry.NotFound` as non-hard-quota in `sdk_error_is_hard_quota`

## Why
The upstream retry error surface now includes `NotFound`, but `lib/oas_worker_named.ml` still matched the older variant set. With warning 8 promoted, `dune build` failed on a non-exhaustive match.

## Impact
The OAS worker now compiles cleanly against the current retry surface and preserves the expected semantics for not-found errors.

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-oas-worker-retry-notfound`


jeong-sik:
Closing as duplicate of #9346, which already covers the Retry.NotFound exhaustiveness fix and adds targeted tests.

Issue #9335, opened 2026-04-21
Integrate parsing and resource missing errors with Prometheus metrics


jeong-sik:
### Problem
Currently, when a required resource (like prompt files) is missing or JSON fields are absent (`missing required field(s)`), the system only prints a generic error log (`Log.Misc.error`). This makes it difficult to establish robust observability or triggering alerts.

### Action Items
- Promote string-based missing resource and parsing errors to explicit Prometheus counters (e.g., `masc_error_events_total{type="missing_config"}`).
- Ensure startup failures due to missing prompts are tracked.

jeong-sik:
라벨 없음. 이슈 triage 필요.

- type:feature?
- area:observability? area:metrics?
- priority?

@sojin

jeong-sik:
Prometheus  counter is already implemented in , tracking startup failures due to missing prompt files.  counter also exists in  and . Closing as already implemented.

PR #9355, opened 2026-04-21
Bump masc_mcp version to 0.12.4

jeong-sik:
Bump version for upcoming telemetry integration fixes.

(The root cause of #9335 was resolved upstream in `oas` PR).

### Changes:
- Bumped `dune-project` version to 0.12.4.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -3,7 +3,7 @@
 (using menhir 3.0)
 
 (name masc_mcp)
-(version 0.12.3)
+(version 0.12.4)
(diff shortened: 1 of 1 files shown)


jeong-sik:
Blocked by the release train guard: main already advertises 0.12.3 while the latest published tag is still v0.12.2, so widening directly to 0.12.4 is premature. Release/tag v0.12.3 first, then cut the next bump from that new baseline.

PR #9416, opened 2026-04-22
[codex] align cascade model provenance with runtime resolution

jeong-sik:
## What changed
- add typed `model_resolution` / `model_resolution_provenance` helpers in `cascade_model_resolve`
- cover alias, env default, hardcoded default, discovery, and unresolved-auto cases in `test_observability_provider_contracts`
- route `Cascade_config.make_registry_config` through `resolve_auto_model` so runtime config construction uses the same model-resolution SSOT

## Why
`cascade_model_resolve` gained provenance-aware resolution logic, but `cascade_config` still had its own local `auto` resolution path. That left runtime behavior and observability/contracts on separate code paths.

## Impact
- local and cloud model resolution now share one SSOT
- future surfaces can consume provenance without re-deriving resolution decisions
- local provider discovery remains provider-specific (`ollama` stays URL-scoped, `llama` stays endpoint-scoped)

## Validation
- `git diff --check`
- attempted: `dune build --root ~/me/workspace/yousleepwhen/masc-mcp/.worktrees/task-219-cascade-model-provenance --build-dir _build_task219 @test/runtest-test_observability_provider_contracts`
- blocker: `agent_sdk.swarm` library missing from local dune environment (`test/deps/dune`, `lib/dune`)

### DIFF
--- a/lib/cascade/cascade_config.ml
+++ b/lib/cascade/cascade_config.ml
@@ -10,6 +10,7 @@
 (* ── Re-exports from extracted modules ─────────────── *)
 
 (* Model resolution *)
+let resolve_auto_model = Cascade_model_resolve.resolve_auto_model
 let resolve_glm_model_id = Cascade_model_resolve.resolve_glm_model_id
 let resolve_auto_model_id = Cascade_model_resolve.resolve_auto_model_id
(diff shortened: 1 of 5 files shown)


jeong-sik:
Superseded by #9417 after recreating the branch on current main to remove the history conflict.

PR #9483, opened 2026-04-22
fix(keeper): replace List.hd with safe pattern match in keepalive

jeong-sik:
## Summary

Replaces unsafe `List.hd` with a pattern match to eliminate the last `List.hd` instance in `lib/`, satisfying the health check ratchet (`lib_list_hd` baseline = 0).

## Changes

- `lib/keeper/keeper_keepalive.ml`: Pattern match on sorted paths instead of `List.hd`

## Related

- Follow-up to #9475 (which was auto-merged before this fix could be included)
- Part of the operational quality remediation plan

## Verification

- [x] `dune build` passes
- [x] `rg 'List\.hd' lib/` returns 0 matches
- [ ] Health check ratchet passes (`lib_list_hd` count stays at 0)

### DIFF
--- a/lib/keeper/keeper_keepalive.ml
+++ b/lib/keeper/keeper_keepalive.ml
@@ -1706,7 +1706,9 @@ let persist_directive_meta_update
             let b_mtime = Option.value ~default:0.0 (Fs_compat.file_mtime b) in
             Float.compare b_mtime a_mtime
           in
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #9473 which already applied the same List.hd fix to keeper_keepalive.ml.

PR #9494, opened 2026-04-22
[codex] remediate remaining fsm surface drift

jeong-sik:
## What changed

- promoted Goal `phase` to a first-class query/filter axis across the backend tool surface and Planning UI
- exposed keeper composite `collapsed_from` so `Stable` snapshots can explain which raw keeper phase was collapsed
- documented the Goal FSM / quorum flow and dashboard semantics in the spec docs

## Why

The remaining FSM audit found that the core state machines were mostly sound, but the operator-facing surfaces were still lagging behind:

- `masc_goal_list` and Goal Tree filtering still leaned on legacy coarse `status` buckets instead of explicit Goal FSM `phase`
- the keeper composite/FSM Hub reduced too many raw keeper phases into `Stable`, which hid whether a keeper was paused, crashed, stopped, or just quiet
- the Goal FSM and new tool surfaces existed in code, but not in the spec/help docs

## Impact

- operators can now query goals by explicit lifecycle phase through `masc_goal_list`
- Workspace > Planning now supports direct phase filtering with ancestor-preserving tree pruning
- FSM Hub can show `Stable <- paused/crashed/...` instead of implying all stable snapshots mean the same thing
- docs now describe the Goal-native FSM, quorum verification flow, and the dashboard contract for phase/collapsed composite views

## Root cause

The FSM implementations landed ahead of their query, projection, and documentation surfaces. That left a drift where the state engines knew more than the operator-facing tools and UI exposed.

## Validation

- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fsm-audit-remediation ./test/test_goal_store.exe ./test/test_goal_tools.exe ./test/test_keeper_composite_observer.exe`
- `./_build/default/test/test_goal_store.exe`
- `./_build/default/test/test_goal_tools.exe`
- `./_build/default/test/test_keeper_composite_observer.exe`
- `git diff --check`

## Notes

- frontend package dependencies are not installed in this worktree (`dashboard/node_modules` missing), so Vitest/TS checks for the dashboard files were not run locally here

### DIFF
--- a/dashboard/src/api/schemas/keeper-composite.test.ts
+++ b/dashboard/src/api/schemas/keeper-composite.test.ts
@@ -28,6 +28,7 @@ describe('parseKeeperCompositeSnapshot', () => {
   it('parses a valid snapshot', () => {
     const result = parseKeeperCompositeSnapshot(VALID_SNAPSHOT)
     expect(result.phase).toBe('Stable')
+    expect(result.collapsed_from).toBeUndefined()
     expect(result.turn_phase).toBe('idle')
     expect(result.is_live).toBe(true)
     expect(result.last_outcome).toBeNull()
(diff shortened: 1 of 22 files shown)


jeong-sik:
## Dashboard typecheck fail: duplicate object keys

CI Dashboard job fails with TS1117 (duplicate property names):

**`src/components/fsm-hub-types.ts:165`** — lowercase keys (`running`, `failing`, `draining`, ...) overlap with existing Titlecase versions already in the same object. TypeScript rejects duplicate keys in object literals.

**`src/components/fsm-hub-pipeline-panels.ts:254`** — same pattern. `Draining` already exists at line ~250, codex added `draining` again.

### Fix needed

Either:
1. Remove the added lowercase entries (if Titlecase already covers them), or
2. Add only the entries that don't have an existing Titlecase equivalent, or
3. Consolidate into lowercase-only keys with a `.toLowerCase()` lookup

Option 1 is simplest — the Titlecase entries already provide display names for these states.

PR #9496, opened 2026-04-22
refactor: remove dead fetch_exn and replace assert false with failwith

jeong-sik:
## Summary

Two small follow-ups to the OCaml best-practices sweep:

1. **Remove dead `fetch_exn`** from `tool_blob_store`
   - `fetch` already returns `string option`. No callers existed for `fetch_exn`.
   - Removing it eliminates `raise Not_found` as control flow.

2. **Replace `assert false` with descriptive `failwith`** in `dashboard_cache`
   - Two theoretically-unreachable branches now produce meaningful error
     messages if the atomic_update action invariant ever breaks.

## Test plan

- [x] `dune build --root .` clean
- [x] `test_admission_queue_coverage.exe` 17/17 pass

### DIFF
--- a/lib/auth.ml
+++ b/lib/auth.ml
@@ -57,7 +57,7 @@ let save_private_text_file path content =
         open_out_gen [ Open_wronly; Open_creat; Open_trunc; Open_text ] 0o600
           path
       in
-      Fun.protect ~finally:(fun () -> close_out oc) (fun () ->
+      Fun.protect ~finally:(fun () -> close_out_noerr oc) (fun () ->
           output_string oc content));
   chmod path 0o600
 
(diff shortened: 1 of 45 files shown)


jeong-sik:
## Health Check Failure: `lib_failwith` ratchet regression

Health ratchet detected: `lib_failwith 16→17`.

This PR replaces `assert false` with `failwith`, which is the intended refactoring — but the Health ratchet counts `failwith` as an unsafe pattern and blocks any increase.

**Options:**
1. If this `failwith` addition is intentional, update `.ci/health-baseline.json` to bump `lib_failwith` from 16 to 17 in this PR.
2. Alternatively, use `Error`/`Result` instead of `failwith` to avoid the regression.

See: `scripts/health_snapshot.sh` → `Ratchet: fail (lib_failwith 16->17)`.

PR #9508, opened 2026-04-22
[codex] Clarify keeper runtime limits panel copy

jeong-sik:
## Summary

- clarify the keeper runtime panel labels so per-keeper limits are not confused with live keeper counts
- add explicit helper copy under the panel header
- cover the wording in the config-resolution-panel render test

## Why

The dashboard can show runtime limit values like `15` in keeper-related UI, which made it easy to misread the value as a keeper count.

## Impact

Operators can distinguish live keeper counts from per-keeper runtime limits more quickly.

## Validation

- `./node_modules/.bin/vitest run src/components/tools/config-resolution-panel.test.ts`
- `./node_modules/.bin/tsc --noEmit --pretty false`

### DIFF
--- a/dashboard/src/components/tools/config-resolution-panel.test.ts
+++ b/dashboard/src/components/tools/config-resolution-panel.test.ts
@@ -112,6 +112,18 @@ describe('ConfigResolutionPanel', () => {
             started_at: '2026-03-27T00:00:00Z',
             uptime_seconds: 42,
           },
+          keeper_runtime: {
+            bootstrap_max_active_keepers: { value: 9, source: 'toml' },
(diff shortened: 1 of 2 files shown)


jeong-sik:
Observed CI note:

- `Health` is failing on `Ratchet: fail (lib_failwith 16->17)`.
- This PR only changes dashboard TS files and does not touch `lib/`.
- I traced the failure to a repo-wide health baseline drift and filed #9511.
- Current evidence: `.ci/health-baseline.json` still expects `lib_failwith: 16`, while current `main` and the PR synthetic merge both contain 17 `failwith` occurrences in `lib`.

I am treating `Health` as an unrelated external blocker for this PR unless #9511 is resolved or the baseline is updated.

PR #9515, opened 2026-04-22
[codex] fix worker docker build lane

jeong-sik:
## What changed
- replaced the stale Alpine `ocaml/opam` build lane in `Dockerfile.worker`
- aligned worker image builds with the repo's working Debian OCaml 5.4 lane
- kept the worker image focused on `masc-worker-run --spec-stdin`, with a slim Debian runtime carrying only required shared libs

## Why
`Dockerfile.worker` was no longer buildable:
- `ocaml/opam:alpine-5.4-flambda` no longer resolved
- after swapping tags, the Alpine lane still failed dependency resolution on `dune >= 3.22`

The repo already had a working Debian-based OCaml 5.4 Docker lane in `Dockerfile.worker-runtime`, so this change reuses that proven toolchain instead of trying to keep reviving the stale Alpine path.

## Validation
- `scripts/build-worker-image.sh dockercheck`
- `printf '{}' | docker run --rm -i masc-worker-runtime:dockercheck`
  - expected result: container starts and the helper returns a JSON `spec_parse` error because `{}` is not a valid worker spec

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -1,37 +1,46 @@
 # Dockerfile for masc-worker-runtime image.
-# Builds the masc-worker-run helper binary and packages it into a minimal
-# Alpine image. The binary reads a Worker_execution_spec from stdin (via
-# --spec-stdin) and writes the run_result to stdout.
+# Builds the masc-worker-run helper binary and packages it into a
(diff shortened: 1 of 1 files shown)


jeong-sik:
Dockerfile.worker was deleted in main. PR obsolete.

PR #9627, opened 2026-04-23
feat(exec): P7 pipeline-native dispatch

jeong-sik:
## Summary
- Execute `Shell_ir.t` AST directly via `Process_eio` without going through `/bin/bash`, eliminating shell injection attack surface
- Simple commands use argv-based spawn; pipelines chain stdout→stdin across stages with last-nonzero-wins exit code
- Controlled by `MASC_BASH_NATIVE_DISPATCH` env var (default: enabled, set `"0"` to always fall back to bash)
- Fix pre-existing erasable-optional-argument warnings in keeper_shell_docker and keeper_exec_shell

## Test plan
- [x] `test_exec_dispatch.ml`: resolve_arg (Lit/Var/Concat/missing), native_dispatch_enabled, dispatch_simple echo/cat, pipeline echo|tr, pipeline exit code, empty pipeline
- [ ] Manual verification: `dune exec test/test_exec_dispatch.exe` passes all 10 assertions
- [ ] Shadow mode comparison: run 50-command corpus through both native and bash paths

### DIFF
--- a/lib/exec/bash_history.ml
+++ b/lib/exec/bash_history.ml
@@ -1,0 +1,193 @@
+(* P11: Command History & Suggest
+   Append-only JSONL log of bash executions.  Each keeper gets its own
+   history file under [.masc/keeper/<name>/bash_history.jsonl].
+   Automatic compaction kicks in above 10 000 entries. *)
+
+type history_entry = {
+  ts : float;
+  cmd_hash : string;
+  cmd_prefix : string;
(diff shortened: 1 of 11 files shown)


jeong-sik:
Self-review: P7 bash history module. Append-only JSONL with automatic compaction above 10k entries. Path structure is consistent with keeper state layout. CI all green. Ready.

PR #9634, opened 2026-04-23
[codex] relax worktree git status timeout for large repos

### LINKED ISSUE #9628: [PERF] Process_eio git status timeout in large repo with 64 worktrees
## Problem
`Process_eio` repeatedly times out on `git status --porcelain` with 15s budget.

## Evidence
- Log: `[WARN] [Misc] [Process_eio] Timeout after 15s: 'git' '-C' '/Users/dancer/me' '--no-optional-locks' 'status' '--porcelain'` — 15+ occurrences in ~20 min window.
- Empirical: `git status --porcelain` in `/Users/dancer/me` takes **13.2s** (user 0.19s, system 2.53s, 20% CPU).
- `git worktree list | wc -l` = **64** worktrees.
- Dashboard cache compute also times out: `cache compute timeout: shell:coord=... (16s)`.

## Impact
- Stale git-derived state in dashboard/operator snapshot.
- Log noise (~0.75/min WARN band).
- Dashboard snapshot already slow: `operator_snapshot profile total=6.9s`.

## Proposed Fixes (pick one or combine)
1. Increase `Process_eio` timeout for git ops from 15s to 30s (or make it proportional to worktree count).
2. Add `--ignore-submodules` or scope git ops to fewer paths.
3. Prune stale worktrees (64 is high; many may be >3 days old).
4. Cache git status result with TTL instead of recomputing per snapshot cycle.

## Environment
- Server: `main_eio.exe` on `localhost:8935`
- Base path: `/Users/dancer/me`
- Date: 2026-04-23

jeong-sik:
## Summary
- increase the default worktree live-context `git status --porcelain` timeout from 15s to 30s
- document the large-repo/worktree pressure that still reproduced timeout noise on 2026-04-23
- add regression coverage for the new default and the `MASC_WORKTREE_GIT_STATUS_TIMEOUT_SEC` override

## Why
- `git status --porcelain` was still timing out in large repos with many worktrees, which left operator snapshots stale and generated recurring WARN noise
- the live-context call is cached, so increasing the bounded refresh budget is lower risk than letting the status capture fail repeatedly

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/issue-9628 lib/worktree_live_context.ml lib/worktree_live_context.mli`
- `/Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/issue-9628/_build/default/test/test_worktree_live_context.exe`

Fixes #9628

### DIFF
--- a/lib/worktree_live_context.ml
+++ b/lib/worktree_live_context.ml
@@ -12,18 +12,19 @@ let clear_git_capture_hook_for_tests () =
 
 (* `git status --porcelain` timeout budget.
 
-   The 5.0s default hit its ceiling 30x in a 45-minute fleet window on
-   2026-04-20 when the workdir was a Second Brain root with many
-   worktrees, thousands of untracked files, and concurrent indexer
-   activity. Each timeout falls through to stale cache (live context
-   goes quiet for the keeper) and emits a WARN.
+   The original 5.0s default hit its ceiling 30x in a 45-minute fleet
+   window on 2026-04-20 when the workdir was a Second Brain root with
+   many worktrees, thousands of untracked files, and concurrent indexer
+   activity. On 2026-04-23, the 15.0s follow-up budget still timed out
+   in `/Users/dancer/me` with 64 worktrees, which left operator
+   snapshots stale and generated recurring WARN noise (#9628).
 
-   15.0s covers p99 for large working trees without meaningfully
-   stretching keeper turn latency — the call is cached (see
-   [status_cache_ttl_sec] below) so the full budget is paid at most
+   30.0s keeps the capture bounded while giving large repos enough head
+   room for a cached status refresh. The call is still cached (see
+   [status_cache_ttl_sec] below), so the full budget is paid at most
    once per TTL window per repo. Env var stays the escape hatch for
-   unusually slow hosts. *)
-let default_git_status_timeout_sec = 15.0
+   unusually slow or unusually fast hosts. *)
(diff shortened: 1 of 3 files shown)


jeong-sik:
## Rejection: Symptom-treating, not root-cause fix

This PR increases the timeout budget from 15s to 30s, but the actual problem is not that 15s is insufficient — it is that `git status --porcelain` is being called **too frequently on too expensive a workload**.

### Root cause analysis

1. **Cache TTL is 1 second** (`MASC_WORKTREE_STATUS_CACHE_TTL_S`, default 1.0). In a 45-minute window this means ~2700 calls. 30 timeouts = ~1% failure rate. Timeout is a statistical certainty when scanning thousands of untracked files 2700 times.
2. **No `--untracked-files=no`**. `git status --porcelain` scans all untracked files by default. Second Brain root has thousands of them. Most of the 15s+ is spent scanning files that are irrelevant to "working tree changed since last keeper turn" detection.
3. **Missing stale-while-revalidate**. When timeout occurs the call falls through to empty/stale and emits WARN noise.

### Proper fix (Option A)

- Add `--untracked-files=no` to the `git status` invocation — reduces call time from 15s+ to <1s on large repos
- Increase `status_cache_ttl_sec` default from 1.0s to 5.0s — reduces call frequency by 80%
- Keep timeout at 15s — with the above changes it becomes ample headroom

This PR is closed in favor of a follow-up implementing the root-cause fix.

PR #9648, opened 2026-04-23
test(auth): guard missing keeper alias noise

jeong-sik:


jeong-sik:
Self-review: Small focused fix — stderr noise on missing keeper alias is captured and verified silent. No side effects on existing credential flows. CI all green. Ready.

Issue #9624, opened 2026-04-23
system_log JSONL deleted while server holds fd — logs not persisted


jeong-sik:
lsof showed fd 3w open to ~/.masc/logs/system_log_2026-04-23.jsonl (12 MB) but the path did not exist on disk (deleted-but-open). Server restart recreated the log stream to /private/tmp/masc-mcp-restart.log. Root cause: log rotation or cleanup removed the file without triggering reopen. HUP had no effect.

jeong-sik:
Resolved by #9659 (commit c32aba47a0, "fix(log): reopen file sink when log path is deleted").

The current `lib/masc_log/log.ml` writer detects the deleted-but-open inode via `sink_matches_path` (Unix.stat path vs Unix.fstat channel — by `(st_dev, st_ino)` and `Sys.file_exists path` first):

```ocaml
let sink_matches_path path oc =
  protect ~default:false (fun () ->
    if not (Sys.file_exists path) then false
    else
      let path_stat = Unix.stat path in
      let chan_stat = Unix.fstat (Unix.descr_of_out_channel oc) in
      path_stat.st_dev = chan_stat.st_dev && path_stat.st_ino = chan_stat.st_ino)

let rotate_if_needed () =
  ... if needs_reopen then begin close_sink (); open_sink !file_base_dir end
```

`rotate_if_needed` runs before every `write_to_sink` call, so the next emit after a deletion (or a date rollover) closes the stale fd and reopens to the canonical path. The `protect` wrapper handles the racey case where `stat` fires between `file_exists` and the actual `Unix.stat` call.

If a future scenario reproduces deleted-but-open even with this protection, please file with:
- Output of `lsof -p <pid> | grep system_log_` to confirm fd state
- `tail -F` on the path to confirm new writes are NOT appearing in the new inode
- Last few `system_log_*.jsonl` lines to confirm `rotate_if_needed` was reached

Closing as completed.

PR #9659, opened 2026-04-23
fix(log): reopen file sink when log path is deleted (#9624)

jeong-sik:


jeong-sik:
Self-review: Log sink reopen on path deletion is an operational hygiene win.  uses inode/dev comparison which correctly detects external deletion (rm, logrotate). Test verifies end-to-end recreate. Waiting for Build/Test to finish before Ready.

jeong-sik:
Self-review: Log sink reopen on path deletion is an operational hygiene win. sink_matches_path uses inode/dev comparison which correctly detects external deletion (rm, logrotate). Test verifies end-to-end recreate. Waiting for Build/Test to finish before Ready.

Issue #9629, opened 2026-04-23
[OPS] Governance compute_judgments fails every ~2 min: OAS bridge 60s timeout


jeong-sik:
## Problem
`masc_oas_bridge` times out after 60.0s every ~2 minutes, causing `Governance refresh_once: compute_judgments failed`.

## Evidence (from 2026-04-23 live log)
```
18:31:56 [WARN] masc_oas_bridge: OAS execution timed out after 60.0s
18:31:56 [WARN] Governance refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s
18:33:56 [WARN] masc_oas_bridge: OAS execution timed out after 60.0s
18:33:56 [WARN] Governance refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s
18:35:57 [WARN] masc_oas_bridge: OAS execution timed out after 60.0s
18:35:57 [WARN] Governance refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s
18:37:57 [WARN] masc_oas_bridge: OAS execution timed out after 60.0s
18:37:57 [WARN] Governance refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s
```

## Impact
- Governance judgments stale (no fresh evaluations).
- Log noise: ~0.5 WARN/min.
- May mask keeper decision quality degradation.

## Context
- Server: `main_eio.exe` on `localhost:8935`
- OAS worker: `oas-big_three`
- Concurrent keeper LLM timeouts also observed (nick0cave/sangsu at 573s).

## Suggested Diagnosis
1. Check OAS server health / CPU load.
2. Verify `masc_oas_bridge` timeout budget vs actual OAS response latency.
3. Consider circuit-breaker or degraded-mode governance when OAS is slow.

jeong-sik:
## Update 2026-04-23 18:43

 keeper_llm_bridge timeout also exceeds its own budget:

```
18:43:21 [WARN] keeper_llm_bridge: OAS execution timed out after 596.6s (budget=573s)
18:43:22 [WARN] nick0cave: transient network error ... Timeout after 596.6s (budget=573s)
```

596.6s > 573s budget implies the timeout enforcement is not hard-capping at the budget (23.6s overshoot). This is either:
- Timer precision issue in the bridge timeout fiber, or
- The timeout is applied to a different phase than the total elapsed time.

Suggest checking `keeper_llm_bridge.ml` timeout fiber logic.

[...]

PR #9661, opened 2026-04-23
fix(governance): keep fresh judgments live on judge timeout (#9629)

jeong-sik:


jeong-sik:
Self-review: Graceful degradation pattern is sound.  preserves fresh cached judgments instead of immediately flipping offline. Two tests cover both fresh-cache and expired-cache branches. Waiting for Build/Test to finish before Ready.

PR #9712, opened 2026-04-23
[codex] Fix dashboard goals null keeper telemetry

jeong-sik:
## Summary

- Make keeper runtime trust JSON field accessors tolerate missing or null nested objects, including `telemetry: null` before reading `selected_model`.
- Add a dashboard goals regression fixture where a linked keeper has a decision log entry with null telemetry, then verify goal tree/detail generation remains healthy.
- Align the keeper TOML canonical-key invariant after the latest main rebase so unparsed `keeper.tool_access.*` table entries are reported as unknown instead of crashing module initialization.

## Root Cause

`/api/v1/dashboard/goals` builds keeper runtime trust snapshots for linked keepers. Older or partial decision log entries can have missing/null `telemetry`; the snapshot code then called `Yojson.Safe.Util.member "selected_model"` on `Null`, raising `Yojson__Safe.Util.Type_error` and surfacing as HTTP 500.

## Validation

- `dune build --root . test/test_dashboard_goals.exe test/test_keeper_toml.exe`
- `./_build/default/test/test_dashboard_goals.exe`
- `./_build/default/test/test_keeper_toml.exe`

Note: local full `dune build --root .` was attempted twice, but the tool session ended with code `-1` and no diagnostics before completion, so targeted build/test evidence above is the reliable local validation.

### DIFF
--- a/lib/keeper/keeper_runtime_trust_snapshot.ml
+++ b/lib/keeper/keeper_runtime_trust_snapshot.ml
@@ -1,27 +1,25 @@
 open Keeper_types
 
+let json_member key = function
+  | `Assoc _ as json -> Yojson.Safe.Util.member key json
+  | _ -> `Null
+
 let json_int_opt_member key json =
-  match json with
-  | `Assoc _ -> (
-      match Yojson.Safe.Util.member key json with
-      | `Int n -> Some n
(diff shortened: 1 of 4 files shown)


jeong-sik:
Superseded by #9705 which already merged the same fix.

PR #9721, opened 2026-04-23
[codex] Draft guard recovery canary - close after validation

jeong-sik:
Live canary for #9719. Empty commit only. Expected: initial draft guard passes; unauthorized ready_for_review later fails with an explicit guard comment if GitHub Actions cannot convert back to draft. This PR must be closed unmerged.


jeong-sik:
Canary validated: draft guard posted explicit recovery-failure guidance and blocked the unauthorized ready_for_review path as expected. Closing unmerged per PR body.

Issue #9574, opened 2026-04-23
[ops] sangsu keeper MCP credential missing: .masc/auth/agents/sangsu.json not found


jeong-sik:
## Summary
`sangsu` keeper의 MCP tool 호출이 전부 `Unauthorized: No credential found`로 실패하고 있습니다.

## Evidence
System log (`seq 13922, 14990, 14994, 15317`):
```
MCP tool call failed: masc_claim_next — Unauthorized: No credential found for sangsu
MCP tool call failed: masc_status — Unauthorized: No credential found for sangsu
MCP tool call failed: masc_join — Unauthorized: No credential found for sangsu
```

## Root Cause
`.masc/auth/agents/sangsu.json` credential 파일이 존재하지 않습니다.
현재 존재하는 파일:
- `.masc/auth/agents/admin.json`
- `.masc/auth/agents/codex-mcp-client.json`
- `.masc/auth/agents/dashboard-dev.json`

## Context
This is a recurrence of the keeper credential name drift class (see feedback memory: keeper-credential-name-drift). `.masc/auth/agents/<name>.json` missing → keeper subprocess MCP calls all fail.

## Proposed Fix
Generate sangsu credential file, or implement Auth.ensure_keeper_credential so keeper bootstraps its own credential.

## Affected Keeper
- `sangsu` (sandbox_profile=docker, tool_preset=delivery)

jeong-sik:
Resolved by #9737 (UUID-based identity system Phase 3) and #9741 (fix auth keeper credential uuid reuse). sangsu credential is now managed through the UUID identity system.

PR #9737, opened 2026-04-23
feat(auth,keeper): UUID-based identity system Phase 3

jeong-sik:
## Summary
Introduces Relay GraphQL-style global IDs (`base64(type:uuid)`) as the canonical identity for all persisted entities. `agent_name` becomes a display label; UUID becomes the lookup key.

## Changes
- **New ID types** (`lib/types/ids.ml`): `Agent_id`, `Keeper_id`, `Credential_id` with generation, JSON serialization, and global ID encoding.
- **Deterministic keeper IDs**: UUIDv5 from namespace + keeper name + config path.
- **Credential storage migration**: `{uuid}.json` primary, `{agent_name}.json` redirect stub for backward compatibility.
- **Auth lookup cleanup**: `verify_token` removes `extract_agent_type_prefix`, uses exact match + legacy aliases only.
- **Keeper TOML load-time ID assignment**: generates or validates keeper ID at discovery.

## Related Issues
- Fixes credential name drift (#9646, #9574)
- Closes auth lookup ambiguity causing unauthorized errors

## Verification
- [x] `dune build` passes
- [x] `dune runtest` passes
- [ ] Manual: restart server, verify credential files contain UUIDs

### DIFF
--- a/lib/auth.ml
+++ b/lib/auth.ml
@@ -273,18 +273,37 @@ let load_credential_from_path config agent_name path : agent_credential option =
     Without this fallback, Coord.join's nickname output caused a
     chronic "No credential found for <type>-<adj>-<animal>" noise band
     at ~0.3/min on the live fleet (2026-04-20). *)
+let load_credential_from_path_raw config agent_name path : agent_credential option =
+  if file_exists path then
+    try
+      let content = read_text_file path in
+      let json = Yojson.Safe.from_string content in
+      match agent_credential_of_yojson json with
+      | Ok cred -> Some cred
+      | Error msg ->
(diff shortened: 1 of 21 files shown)


jeong-sik:
Review blocker: `ensure_keeper_credential` now generates a fresh `Credential_id` on every call before saving the credential. Startup calls this for every bootable keeper, so each restart rewrites the `{agent_name}.json` redirect stub to a new UUID file and leaves the previous UUID credential file orphaned. It also makes the canonical persisted credential ID unstable across syncs. Please load the existing exact-name credential first and preserve its `id`/`created_at` when refreshing the token hash; generate a new UUID only when no credential exists. A regression test should call `ensure_keeper_credential` twice for the same keeper and assert the credential id stays stable.

Issue #9844, opened 2026-04-24
F-2: 'gh auth login --with-token' rewrites hosts.yml:user, erasing identity label


jeong-sik:
## Symptom

Inside the keeper sandbox, `gh auth login --with-token` (Option B credential-provider path) overwrites `hosts.yml:user` with the *real* token owner. The caller's identity label (e.g. `anyang-keepers`) is not preserved.

## Reproduce

```
TOKEN=$(awk '/oauth_token:/ {print $2}' ~/me/.masc/github-identities/anyang-keepers/gh/hosts.yml | tr -d ' \r\n')
docker run -d --name gh-probe --network bridge -v gh-auth-probe:/root/.config/gh \
  -e GH_CONFIG_DIR=/root/.config/gh masc-keeper-sandbox:local sleep 600
printf '%s\n' "$TOKEN" | docker exec -i gh-probe gh auth login --with-token
docker exec gh-probe gh auth status | head -3
# -> "Logged in to github.com account jeong-sik (…)"  ← NOT anyang-keepers
docker rm -f gh-probe && docker volume rm gh-auth-probe
```

## Impact

- Any downstream consumer that reads `hosts.yml:user` (dashboards, audit log labels, policy matchers) sees the real owner after the Option B bootstrap, even though the provider was instructed to act as the keeper identity.
- Identity separation is cosmetic if Option B is used in production.

## Suggested direction

- Define a `CredentialProvider.bootstrap_finalize` hook (see RFC-0007 §3.6) that runs after `gh auth login --with-token` and rewrites `hosts.yml:user` to the identity name.
- Or: skip the login path entirely and rely on Option A (host-mounted bundle) for production keepers until finalize is implemented.

## Evidence

- `step-E-optionB.log` in `~/me/.tmp/keeper-docker-gh/`.
- Record: `memory/procedural-memory/2026-04-24-keeper-docker-github-provider-evidence-record.md`.
- Related RFC: #9842.

jeong-sik:
현재 main 검증 결과: production keeper 는 Option A (`Host_config_provider`, host bundle mount RO) 만 사용. Option B (`In_container_login_provider`, `gh auth login --with-token`) 는 RFC-0008 PR-3 로 별도 epic, 아직 production 코드에 미연결.

`lib/keeper/keeper_shell_docker.ml:393-411` 의 docker-invocation 사이트가 `Host_config_provider.resolve` 만 호출:

```ocaml
match
[...]

Issue #9843, opened 2026-04-24
F-1: 'anyang-keepers' identity shares the operator PAT (SHA-256 match)


jeong-sik:
## Symptom

The OAuth token stored in `~/.masc/github-identities/anyang-keepers/gh/hosts.yml` is the same token `gh auth token` returns on the operator host. SHA-256 prefixes match (`406d098bd41b`). The "keeper-scoped identity" is currently a label, not a capability boundary.

## Reproduce

```
awk '/oauth_token:/ {print $2}' ~/me/.masc/github-identities/anyang-keepers/gh/hosts.yml | tr -d ' \r\n' | shasum -a 256 | cut -c1-12
gh auth token | tr -d ' \r\n' | shasum -a 256 | cut -c1-12
```

Both print `406d098bd41b` on a clean host (2026-04-24).

## Impact

- `MASC_KEEPER_SANDBOX_HARD_MODE=true` advertises "keeper-scoped credential"; in practice the bundle is a copy of the operator's classic PAT (`scopes: delete_repo, gist, read:org, repo, workflow`).
- Any auditing path that assumes separation between operator and keeper identities is misled.

## Suggested direction

- Each identity bundle owned by an independently issued fine-grained PAT (one rotation per identity).
- Drop `scripts/rotate-keeper-gh-token.sh`'s legacy `~/.masc/gh-auth` target (see F-4 follow-up), target `~/.masc/github-identities/<identity>/gh`.

## Evidence

- Record: `memory/procedural-memory/2026-04-24-keeper-docker-github-provider-evidence-record.md` (on the `me` repo), decision ID `keeper-docker-gh-provider-audit-2026-04-24`.
- Related RFC: #9842 (RFC-0007 draft) cross-references a CredentialProvider RFC-0008 to separate token issuance from harness.

jeong-sik:
anyang-keepers identity no longer exists. Closing as obsolete.

Issue #9728, opened 2026-04-23
Approval bypass cannot be human-proof while agents share owner credentials


jeong-sik:
## Symptom\nDraft PR guard can verify that a bypass label exists and who GitHub reports as the label actor, but it cannot prove that the action was performed by a human when an agent uses the same GitHub owner token.\n\n## Reproduction / evidence\nDuring the #9707 incident, bypass-like label events such as allow-auto-merge were recorded by GitHub as actor jeong-sik with actor_type User. That actor shape is indistinguishable from a real owner action if automation/agents are allowed to run gh/API commands with the same credentials.\n\n## Estimated cause\nGitHub issue/PR events expose actor identity, not the local process or user intent behind a token. A shared owner PAT/session collapses human approval and agent automation into the same actor.\n\n## Tried\n#9707 ignored automation-prone allow-auto-merge. #9719 made recovery failures explicit. #9727 narrows bypass labels by checking the latest label event actor and rejecting Bot/app/unlisted actors. This still cannot separate a real owner from an agent using the owner token.\n\n## Required follow-up\nIntroduce a credential boundary for important approvals: separate human-only approval credential, GitHub environment/manual dispatch approval, protected reviewer approval, or an out-of-band signed approval record that agents cannot mint with their normal execution token.

PR #9870, opened 2026-04-24
feat(env): RFC-0007 PR-1 — non-interactive git env SSOT + subprocess scrub list

jeong-sik:
RFC-0007 PR-1 lands first. Two SSOT modules + wiring into the two keeper subprocess env paths. RFC-0008 PR-1 (`CredentialProvider`) consumes `Env_git_noninteractive.env` next.

## Why

RFC-0007 §1 documented two concrete failures observed on 2026-04-24:

1. **Non-interactive git defaults not enforced.** Git credential helpers open `/dev/tty` to prompt. A Docker sandbox or background subprocess has no tty, so the call hangs until the outer wall-clock timeout fires silently. Compounds with the cooperative-cancel overshoot surfaced by #9860.
2. **Long-lived host credentials leak into subprocess env.** `ANTHROPIC_API_KEY`, `AWS_SECRET_ACCESS_KEY`, `OTEL_EXPORTER_OTLP_*` bearer headers, `ACTIONS_ID_TOKEN_REQUEST_TOKEN` etc are read in-process by MASC and are not needed by gh/git inside a sandboxed container. Leaking them creates an escape surface.

Both fall under RFC-0007 principles **P2** (scrub vs pass by token scoping) and **P3** (non-interactive defaults are a constant, not an opinion).

## What

- `lib/env_git_noninteractive.{ml,mli}` — canonical pairs `GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS=\"\"`, `GCM_INTERACTIVE=Never`, `SSH_ASKPASS=\"\"`. Exposes three shapes: assoc list, flat `K=V` list for `Unix.environment`, and `[\"-e\"; \"K=V\"; ...]` for direct `docker run` argv. `inject_into_environment` overrides any pre-existing entries so the canonical value wins.
- `lib/env_keeper_scrub.{ml,mli}` — exact-match scrub list copied from claude-code `src/utils/subprocessEnv.ts:15-53` (`GHA_SUBPROCESS_SCRUB`), plus a documentation pass list (`GH_TOKEN`, `GITHUB_TOKEN`, `SSH_AUTH_SOCK`, `GIT_*`) asserted to be disjoint from scrub.
- `lib/keeper/keeper_gh_env.ml` — `process_env` and `keeper_process_env` now compose: `Unix.environment()` → `Env_keeper_scrub.filter_environment` → `Env_git_noninteractive.inject_into_environment` → strip existing `GH_CONFIG_DIR=` → prepend keeper-scoped value.
- `lib/keeper/keeper_shell_docker.ml` — appends `Env_git_noninteractive.docker_args` to the Docker `-e` list so git inside the sandbox observes the same defaults.
- `test/test_env_git_noninteractive.ml` — 11 assertions including an enforcement grep under `lib/` that fails on any future inline `GIT_ASKPASS`/`GIT_TERMINAL_PROMPT` literal outside the SSOT module.

## Design references

- claude-code `src/utils/worktree.ts:199-202` — `GIT_NO_PROMPT_ENV` record (exact copy of the pattern, not just the idea).
- claude-code `src/utils/subprocessEnv.ts:15-53` — `GHA_SUBPROCESS_SCRUB` list (copied key-by-key with rationale comments preserved).
- RFC-0007 rev.2 §2 (design principles P2 + P3) in PR #9842.

## Non-goals

- No Docker sandbox runtime change, no seccomp delta, no `keeper_binding` shape change.
- No `gh_result.t` (RFC-0007 PR-2), no typed `gh_request.t` (RFC-0007 PR-3).
- No cross-repo changes. OAS is untouched.
- RFC-0008 PR-1 (`CredentialProvider` trait) depends on this PR's `Env_git_noninteractive.env` and lands separately.

## Test plan

- [x] `dune exec test/test_env_git_noninteractive.exe` — all 11 assertions pass
- [x] Enforcement grep test passes on current main (no inline `GIT_ASKPASS` outside SSOT)
- [ ] Live verification: next keeper turn that spawns a Docker sandbox should show `-e GIT_TERMINAL_PROMPT=0` etc in the argv; any credential prompt scenario that previously hung silently should now fail fast with a clean git error

## Cluster B issues addressed or contextualized

- **#9843** ('anyang-keepers' shares operator PAT) — RFC-0008 territory; this PR lays the env groundwork so RFC-0008 can ship without touching host-env scatter.
- **#9844** (`gh auth login --with-token` rewrites `hosts.yml:user`) — RFC-0008 territory, same as above.
- **#9786** (bearer token mismatch across agents) — same.
- **#9728** (approval bypass cannot be human-proof while agents share creds) — architectural; RFC-0008.
- This PR does NOT claim to close any of the above. It is a necessary pre-requisite per RFC-0007 §6 (merge order).

Refs: #9842 (RFC design review), #9860 (prior tick — timeout SSOT)

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -288,6 +288,8 @@
    telemetry_unified
    tempo
    timeout_policy
+   env_git_noninteractive
+   env_keeper_scrub
    tui_decode
    tools
    tool_args
--- a/lib/env_git_noninteractive.ml
+++ b/lib/env_git_noninteractive.ml
@@ -1,8 +1,45 @@
-let env =
+(* Canonical pairs. Keep this list short and justified — each entry is a
+   promise that the named subsystem is documented to honor it.
+
+   GIT_TERMINAL_PROMPT=0 — git/libcurl credential helpers must not open
+     /dev/tty. Documented at https://git-scm.com/docs/git-config under
+     [core.askPass] and git-credential(1).
+   GIT_ASKPASS="" — explicitly empty disables any configured askpass
+     helper for this subprocess without relying on unset semantics.
+   GCM_INTERACTIVE=Never — git credential manager (Microsoft) honors this
+     on Linux/macOS containers to refuse interactive dialogs.
+   SSH_ASKPASS="" — same rationale for SSH-over-git URLs. *)
+
+let env : (string * string) list =
   [
-    ("GIT_ASKPASS", "");
-    ("GIT_TERMINAL_PROMPT", "0");
+    "GIT_TERMINAL_PROMPT", "0";
+    "GIT_ASKPASS", "";
+    "GCM_INTERACTIVE", "Never";
+    "SSH_ASKPASS", "";
   ]
 
-let docker_env_args =
+let env_pairs =
+  List.map (fun (k, v) -> k ^ "=" ^ v) env
+
+let docker_args =
   List.concat_map (fun (k, v) -> [ "-e"; k ^ "=" ^ v ]) env
+
+let docker_env_args = docker_args
+
+let key_of_entry entry =
+  match String.index_opt entry '=' with
+  | None -> entry
+  | Some i -> String.sub entry 0 i
+
+let inject_into_environment existing =
+  let override_keys =
+    let t = Hashtbl.create (List.length env) in
+    List.iter (fun (k, _) -> Hashtbl.replace t k ()) env;
+    t
+  in
+  let filtered =
+    Array.to_list existing
+    |> List.filter (fun e ->
+         not (Hashtbl.mem override_keys (key_of_entry e)))
+  in
+  Array.of_list (env_pairs @ filtered)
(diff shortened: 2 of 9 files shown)


jeong-sik:
## Superseded by #9872 (autonomous /loop tick 7)

This PR is the earlier/full-scope version of RFC-0007 PR-1 (full: `env_git_noninteractive` + `env_keeper_scrub` + `keeper_gh_env`).

- **#9872** (YAGNI-scoped: only `env_git_noninteractive` + single callsite wiring) was merged at `2026-04-24T06:44:48Z` as `da59f4cfd`.
- **#9842** (RFC-0007 rev.3 docs) was also merged — the rev.3 definition itself restricts PR-1 scope to exactly what #9872 landed.

This PR is now `CONFLICTING` because its `env_git_noninteractive` changes are already in main (from #9872). The remaining delta (`env_keeper_scrub`, `keeper_gh_env`) was deferred by rev.3 to a separate follow-up.

**Recommend: close as superseded**. Follow-up PRs for `env_keeper_scrub` and `keeper_gh_env` can be opened independently when their consumers land (per rev.3).

PR #9999, opened 2026-04-24
fix(keeper): keep long turns visibly alive

jeong-sik:
Summary:
- leave a 30s finalization guard inside the keeper turn timeout budget so fallback/meta writes can finish
- emit lightweight in-turn keeper heartbeat SSE pulses while a long OAS turn is still running, without fake OAS snapshots
- switch the pure autonomous wait queue lock to Stdlib.Mutex so test helpers are safe outside Eio_main

Verification:
- scripts/dune-local.sh build test/test_keeper_semaphore_fairness.exe test/test_keeper_unified.exe
- env MASC_BASE_PATH=/tmp/masc-keeper-turn-timeout-test ./_build/default/test/test_keeper_semaphore_fairness.exe (11 tests)
- env MASC_BASE_PATH=/tmp/masc-keeper-turn-timeout-test ./_build/default/test/test_keeper_unified.exe (246 tests)

### DIFF
--- a/lib/keeper/keeper_keepalive.ml
+++ b/lib/keeper/keeper_keepalive.ml
@@ -137,7 +137,9 @@ type autonomous_waiter =
     keeper_name : string;
   }
 
-let autonomous_wait_queue_mutex = Eio.Mutex.create ()
+(* Stdlib.Mutex: queue operations are pure/non-yielding and test helpers call
+   them outside Eio_main. *)
+let autonomous_wait_queue_mutex = Stdlib.Mutex.create ()
 
 let autonomous_wait_queue : autonomous_waiter list ref = ref []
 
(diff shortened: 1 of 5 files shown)


jeong-sik:
### autonomous /loop tick 50 — bundle audit cross-reference
5 PR 묶음 audit: [#9996 comment 4313878501](https://github.com/jeong-sik/masc-mcp/pull/9996#issuecomment-4313878501)

이 PR(#9999): 30s finalization guard + in-turn heartbeat SSE + Stdlib.Mutex. ✅ clean. "without fake OAS snapshots" 선언으로 가짜 heartbeat 선제 회피. `memory/feedback_ocaml5-mutex-selection.md` 정면 준수.

PR #10089, opened 2026-04-24
perf(ws): cache per-broadcast JSON parse across dashboard sessions

jeong-sik:
## Why

`Sse.notify_external_subscribers` delivers the same `event: string` reference to every WebSocket session in a single fanout. Before this change, each session independently ran `Yojson.Safe.from_string` inside `dashboard_delta_for_sse`, multiplying JSON parsing by the number of dashboard clients.

For `N` connected dashboards and `M` dashboard-relevant SSE events per second, parse cost scales as `O(N·M)`. A typical multi-tab operator workflow (5-10 WS sessions × 10 events/s) is wasting 50-100 redundant parses/s of identical JSON.

## What

- Introduce module-local `parse_cache : (string * parsed_sse_event option) Atomic.t` in `lib/server/server_mcp_transport_ws.ml`.
- Factor parsing into `parse_sse_dashboard_event : string -> parsed_sse_event option`. It checks the cache via physical equality on the event string; on miss it parses once, derives `(event_type, slice, payload)` via `dashboard_slice_for_sse_type`, and stores the result.
- `dashboard_delta_for_sse` calls the cached helper, then does only per-session work: `Hashtbl.mem session.dashboard_slices slice` filter and `next_dashboard_seq session` allocation.

## Correctness

- `Atomic.get`/`Atomic.set` is indivisible. A concurrent broadcast at worst wastes one parse — it never returns a wrong logical result for a different event.
- Physical equality is safe: `notify_external_subscribers` holds the event string alive across the fanout loop, and identical references always produce identical parse output.
- No SSE/WS wire-format change. Subscribe/delta/seq semantics unchanged.

## Tests

`test/test_ws_transport.ml` gains a `parse_cache` group:

- known slice maps (`execution_snapshot` → `execution`)
- unknown type returns `Some { slice = None; ... }` (falls through to raw SSE forward)
- malformed JSON returns `None`
- repeat calls produce identical output (cache transparency)
- new string reference invalidates and re-parses correctly

All 16 tests pass locally:

```
dune exec --root=. test/test_ws_transport.exe
Test Successful in 0.020s. 16 tests run.
```

## Scope guard

- No API surface change; no `.mli` exists for this module.
- No touch to `Sse` or `Server_ws_standalone`. Fanout shape preserved.
- No change to the client (`dashboard/src/dashboard-ws.ts`).

## Follow-ups (out of scope)

- Emit a `transport_metrics` counter for cache hits to quantify the win under load.
- Consider moving parse-once logic into `Sse.notify_external_subscribers` so non-WS subscribers also share it (bigger blast radius; separate PR).

### DIFF
--- a/lib/prometheus.ml
+++ b/lib/prometheus.ml
@@ -255,6 +255,8 @@ let metric_grpc_heartbeat_latency = "masc_grpc_heartbeat_latency_seconds"
 let metric_grpc_subscribers = "masc_grpc_subscribers_total"
 let metric_grpc_events_delivered = "masc_grpc_events_delivered_total"
 let metric_ws_sessions = "masc_ws_sessions_total"
+let metric_ws_parse_cache_hits = "masc_ws_parse_cache_hits_total"
+let metric_ws_parse_cache_misses = "masc_ws_parse_cache_misses_total"
 
 (* Admission queue metrics — used in admission_queue_metrics.ml. *)
 let metric_inference_queue_depth = "masc_inference_queue_depth"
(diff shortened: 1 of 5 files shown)


jeong-sik:
Follow-up commit adds the Prometheus counters the first commit promised as a follow-up:

- `masc_ws_parse_cache_hits_total`
- `masc_ws_parse_cache_misses_total`

Rationale: the cache itself is transparent to the wire protocol, so its value is invisible without telemetry. Production dashboards can now plot `hits / (hits + misses)`; under N connected dashboard sessions the steady-state ratio approaches `(N - 1) / N`.

Test added: 3 calls with the same string reference → 1 miss + 2 hits; a fresh allocation of the same content forces another miss (proves physical identity keying, not structural).

17/17 tests pass locally.

PR #10090, opened 2026-04-24
fix(keeper): defend per-model telemetry from empty response.model (#10083)

jeong-sik:
## Problem (#10083)

Some OAS transports — kimi_cli silent-failure path and
`CompletionContractViolation` synthetic responses from retry
exhaustion — emit `api_response.model = ""`. The AfterTurn hook
at `lib/keeper/keeper_hooks_oas.ml:555-625` used the raw string as
a counter label, so the empty string propagated into every
per-model series:

```
masc_after_turn_hook_total{model=""} 7
masc_llm_inference_duration_seconds_sum{model=""} 372.75
masc_llm_inference_duration_seconds_count{model=""} 7
masc_pricing_catalog_miss_total{model=""} 7
```

Each empty-model turn lost ~50s of inference attribution and was
booked as `$0` (since `Pricing.pricing_for_model_opt ""` always
misses). The aggregate also collides with #10064's explicit `""`
heartbeat sentinel — one event path blurred with a different one.

## Fix

Three-tier fallback inside a new helper
`resolve_after_turn_model`, applied before any label is emitted:

| Tier | Condition | Result | Counter label |
|------|-----------|--------|---------------|
| 1 | `response.model` non-empty | raw passthrough | — |
| 2 | raw empty + telemetry has `canonical_model_id` | canonical id | `source=telemetry_resolved` |
| 3 | raw empty + no telemetry id | `"unknown"` | `source=unknown_sentinel` |

Every fallback increments
`masc_after_turn_response_model_empty_total{keeper, source}` so the
operator can attribute leaks per keeper and see which resolution
path recovered each turn. Named sentinel instead of empty string
keeps the label vocabulary explicit — aligned with the project
rule against *Unknown → Permissive Default*.

The resolver substitutes a single site (`let model = ...` at the
top of the `AfterTurn` branch), so every downstream reader
(`classify_usage_trust`, Prometheus labels, SSE `model_used`,
`emit_cost_event`, per-turn log) sees the resolved string.

## Tests

`test/test_response_model_empty_10083.ml` pins all four shapes:

- raw non-empty → verbatim, no fallback counter incremented
- raw empty + telemetry canonical → canonical returned,
  `telemetry_resolved` counter +1
- raw empty + no telemetry → `unknown` returned,
  `unknown_sentinel` counter +1
- **edge**: telemetry present but `canonical_model_id = ""` must
  fall through to the sentinel (not propagate the empty string as
  a "resolved" value)

## Notes

- Upstream `Option A` (OAS transports never emit empty `model`) is
  the right long-term fix; this PR is the MASC-side defense so we
  stop losing per-provider telemetry in the meantime.
- Pairs with `#10064` but does **not** share the `""` sentinel —
  this PR's sentinel is the literal string `"unknown"` so the two
  event paths stay distinguishable in metrics.

### DIFF
--- a/lib/keeper/keeper_hooks_oas.ml
+++ b/lib/keeper/keeper_hooks_oas.ml
@@ -69,6 +69,22 @@ let record_tool_use_failure ~keeper_name ~tool_name =
   Prometheus.inc_counter tool_use_failure_metric
     ~labels:[ ("keeper", keeper_name); ("tool", tool_name) ] ()
 
+(* #10083: some OAS transports (kimi_cli silent-failure and
+   CompletionContractViolation synthetic responses) emit
+   [response.model = ""].  That empty string then flows into every
+   per-model counter label in this hook (after_turn_hook_total,
+   masc_llm_inference_duration_seconds, pricing_catalog_miss_total)
+   and contaminates per-provider aggregates — one empty-model turn
+   wipes out the ability to attribute its ~50s of inference time to
+   any specific provider.  The resolve helper applies a layered
+   fallback (raw → telemetry canonical_model_id → named sentinel)
+   and emits a labelled counter so the operator can see WHICH
+   transport leaked and WHICH resolution path recovered it. *)
+let empty_response_model_metric =
+  "masc_after_turn_response_model_empty_total"
+
+let unknown_model_sentinel = "unknown_provider"
+
 let zero_usage : Oas.Types.api_usage =
   {
     input_tokens = 0;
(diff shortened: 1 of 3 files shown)


jeong-sik:
Acknowledging #10103 (b69cb7e6) merged ~30min ago which addresses the same root symptom with a simpler single-sentinel substitution (`response.model = ""` → `"unknown_provider"`).

This PR's incremental value over #10103:
1. **Tier 2 fallback via `telemetry.canonical_model_id`** — turns where `response.model=""` but `telemetry.canonical_model_id` is set get a real model name instead of the sentinel.
2. **Per-source counter** `masc_after_turn_response_model_empty_total{keeper, source}` quantifies the recovery rate (telemetry_resolved vs unknown_sentinel) — pushes prioritization for the upstream OAS transport fix.

Will leave as Draft for review. Happy to rebase + align sentinel name to `unknown_provider` (matching #10103's vocabulary) on request, or close this as superseded if the simpler approach in main is sufficient.

PR #10098, opened 2026-04-24
perf(ws): share Bytes allocation across raw SSE forwards in a broadcast

jeong-sik:
## Why

`Sse.notify_external_subscribers` delivers the same `event: string` reference to every subscribed WS session. For the **raw-SSE-forward path** (events that do not match any subscribed dashboard slice), each session previously ran `Bytes.of_string sse_event` independently, producing N identical allocations per broadcast.

For `N` connected WS sessions and `E` bytes per event, each broadcast burns `N × E` bytes of short-lived allocation that the GC then has to sweep. A steady-state multi-tab operator (N=5, E=2KB, 10 events/s) wastes ~100 KB/s of allocation churn that carries zero information beyond the first copy.

This is a sibling to the parse cache in #10089: same delivery pattern, different layer.

## What

In `lib/server/server_mcp_transport_ws.ml`:

- Factor out `send_frame_bytes session bytes ~len` as the raw primitive over `Httpun_ws.Wsd.send_bytes`.
- Add `bytes_cache : (string * Bytes.t) Atomic.t` plus `bytes_of_shared_text` keyed by physical equality on the string reference.
- Add `send_text_shared` and `send_text_shared_checked` that go through the cache.
- `send_dashboard_or_raw_sse` routes the two raw-forward call sites (authenticated-but-slice-miss and unauthenticated) through the shared helper.
- `send_text` / `send_text_checked` stay on the allocating path — single-destination sends do not benefit from sharing.

Dashboard **delta** sends still go through `send_json_checked` → `send_text` unchanged, because each session's delta carries a unique `seq` and therefore a unique encoded string.

## Why sharing bytes across sessions is safe

Read from `httpun-ws` 0.2.0 to avoid guessing ownership semantics:

- `Serialize.serialize_bytes` only calls `Websocket.Frame.apply_mask_bytes` when `mask` is `Some _`, which only happens when `mode = `Client`. Our upgrade handler is a server, so `mode = `Server` and the payload is never mutated.
- `Faraday.write_bytes faraday payload ~off ~len` copies synchronously into the serializer's internal buffer. The argument `Bytes.t` needs to be valid only for the duration of the synchronous `Wsd.send_bytes` call; no reference is retained after return.

Both conditions hold on the server send path, so the same `Bytes.t` can be handed to multiple sessions sequentially without observable side effects.

## Tests

`test/test_ws_transport.ml` gains a `bytes_cache` group:

- Same string reference returns the exact same `Bytes.t` (cache hit by physical identity, not structural equality).
- Content round-trips through the cache — `Bytes.to_string` recovers the original.
- Two distinct allocations of identical content force separate `Bytes.t` values — proves the key is physical identity so fresh broadcasts always refresh.

Verified suites:

```
test_ws_transport          14/14  (11 prior + 3 new)
test_transport_integration  8/8   (ws_sse forwarding path covered)
```

## Scope guard

- No wire-format change. The frame on the network is identical — only the server-side allocation pattern differs.
- No change to client code.
- No dependency added.

## Interaction with #10089 / #10096

Independent and stackable:

- #10089 collapses **parse** work per broadcast (input direction).
- #10096 records client **drain rate** per broadcast (output observability).
- This PR collapses **allocation** per broadcast (output direction, GC pressure).

None of the three PRs touch the same function in a conflicting way; they are orthogonal perf layers around the same fanout path.

## Follow-ups (separate PRs)

- Add `masc_ws_bytes_cache_hits_total` / `..._misses_total` counters consistent with the parse-cache observability layer once #10089 lands and sets a naming convention.
- Consider pushing the bytes cache into `Sse.notify_external_subscribers` so gRPC subscribers also reuse bytes (requires a callback signature change — separate blast radius).

### DIFF
--- a/lib/prometheus.ml
+++ b/lib/prometheus.ml
@@ -257,6 +257,8 @@ let metric_grpc_events_delivered = "masc_grpc_events_delivered_total"
 let metric_ws_sessions = "masc_ws_sessions_total"
 let metric_ws_parse_cache_hits = "masc_ws_parse_cache_hits_total"
 let metric_ws_parse_cache_misses = "masc_ws_parse_cache_misses_total"
+let metric_ws_bytes_cache_hits = "masc_ws_bytes_cache_hits_total"
+let metric_ws_bytes_cache_misses = "masc_ws_bytes_cache_misses_total"
 
 (* Admission queue metrics — used in admission_queue_metrics.ml. *)
 let metric_inference_queue_depth = "masc_inference_queue_depth"
@@ -571,6 +573,12 @@ let init () =
     Counter;
   add metric_ws_parse_cache_misses
     "WS dashboard delta parse cache misses (fresh JSON parse required)"
+    Counter;
+  add metric_ws_bytes_cache_hits
+    "WS raw-SSE-forward Bytes cache hits (same event reused across sessions)"
+    Counter;
+  add metric_ws_bytes_cache_misses
+    "WS raw-SSE-forward Bytes cache misses (fresh allocation required)"
     Counter
 
 let start_time = Time_compat.now ()
--- a/lib/prometheus.mli
+++ b/lib/prometheus.mli
@@ -161,6 +161,8 @@ val metric_grpc_events_delivered : string
 val metric_ws_sessions : string
 val metric_ws_parse_cache_hits : string
 val metric_ws_parse_cache_misses : string
+val metric_ws_bytes_cache_hits : string
+val metric_ws_bytes_cache_misses : string
 
 (** {1 Admission queue metrics} *)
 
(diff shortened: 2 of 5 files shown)


jeong-sik:
Follow-up commit adds Prometheus counters for the bytes cache, mirroring the parse-cache counter pattern from #10089:

- `masc_ws_bytes_cache_hits_total`
- `masc_ws_bytes_cache_misses_total`

Rationale: the cache is transparent to the wire protocol, so its value is invisible without telemetry. Production dashboards can now plot `hits / (hits + misses)`; steady-state ratio approaches `(N - 1) / N` for N concurrent WS sessions.

Counter test added: 3 calls with the same string reference → 1 miss + 2 hits; a fresh `String.concat` with identical content forces another miss (proves physical identity keying at the counter level, not just in the cache impl).

15/15 tests pass locally.

PR #10100, opened 2026-04-24
fix(oas-worker): dedup codex_cli MCP omission WARN + per-tool counter (#10097)

jeong-sik:
## Problem (#10097)

`lib/oas_worker_exec_transport.ml:443` emits a WARN on every
codex_cli request whose keeper-bound runtime MCP tool set
includes any tool that needs request-scoped auth headers.

```
143 × "codex_cli omitting keeper-bound runtime MCP tool(s) that require request-scoped auth headers: ..."
```

**1 distinct content × 143 occurrences** — identical 16-tool
list every time. The noise masks unrelated warnings
(`resolve_named_providers leak`, `fixture_contamination`,
`heuristic_metrics degenerate`) in `grep WARN` output.

## Root Cause

The WARN is a **structural fact** about what codex_cli can
carry, not an incident. It was emitted at call-site rate instead
of at "new information" rate. Both signals (structural fact +
frequency) were collapsed into the same channel.

## Fix

Split the two signals:

| Signal | Channel | Rate |
|--------|---------|------|
| Structural fact | `Log.warn` | Once per distinct omitted-tool fingerprint per session |
| Frequency | `masc_codex_cli_mcp_tool_omission_total{tool}` | Every call |

Fingerprint = sorted, comma-joined tool list. If the set
**changes** (a new tool starts being stripped), that is
genuinely new information and the WARN re-fires. Not a single
global "warned-once" flag.

The WARN message now tells the operator where to find the
frequency:
> "subsequent omissions of this same set are counted in
> masc_codex_cli_mcp_tool_omission_total and not re-logged"

`record_codex_cli_omission` lives at the top of
`oas_worker_exec_transport.ml` (alongside other helpers) and is
exported so the invariants are unit-testable without the full
OAS transport harness. `Stdlib.Mutex` guards the dedup Hashtbl
(Eio fibers on separate domains).

## Tests

`test/test_codex_cli_omission_dedup_10097.ml` pins:

- **counter on every call** — 3 calls with same fingerprint
  → counter +3, dedup applies only to logs;
- **per-tool label isolation** — X's counter stays cold when
  Y fires (regression guard for single undifferentiated
  counter anti-pattern);
- **order-insensitive fingerprint** — `[a;b]` and `[b;a]`
  collapse to the same fingerprint;
- **new set re-prints** — a genuinely different omission set
  is NOT silenced; this is the critical invariant that keeps
  the dedup from losing new information;
- **empty list no-op** — pure no-op, no counter, no fingerprint
  slot burned.

Dedicated `(test ...)` stanza with `setenv MASC_BASE_PATH ...`
for the same reason as #10091 (`Masc_mcp` module init triggers
#9903 prod-guard on HOME).

## Notes

- `Option A/B/C` from the issue: this PR implements a hybrid
  (A: one-shot per fingerprint × C: counter). Option B
  (DEBUG-level downgrade) loses the first-occurrence signal
  and the per-tool attribution.
- Dedup is in-memory / session-scoped. A server restart
  re-logs the first occurrence, which is correct: the
  structural fact belongs in the startup-era warnings.

### DIFF
--- a/lib/oas_worker_exec_transport.ml
+++ b/lib/oas_worker_exec_transport.ml
@@ -13,28 +13,83 @@ type cli_transport_overrides = {
   gemini_yolo : bool option;
 }
 
-(* #10097: codex_cli omits the same set of keeper-bound runtime MCP
-   tools on every request because the structural limitation
-   (request-scoped auth headers) cannot be carried through codex's
-   transport.  The previous Log.warn fired 143×/session with one
-   distinct tool list — pure noise that masked other warnings.  Track
-   the most recent tool list per [agent_name] and only emit when it
-   changes (initial fire, or tool surface drift).  Stdlib.Mutex
-   guards concurrent access from heartbeat/turn fibers across
-   domains. *)
+(* #10097: codex_cli omits keeper-bound runtime MCP tools that require
+   request-scoped auth headers.  That omission is a structural provider
+   limitation, not a per-call incident:
+
+     - [WARN] emits only when an agent first sees an omitted-tool
+       fingerprint, or when that agent's omitted tool set changes.
+     - [Prometheus] per-tool counters increment on every omission so
+       dashboards retain the frequency signal.
+
+   Fingerprint = sorted, comma-joined tool list.  Stdlib.Mutex guards
+   concurrent access from heartbeat/turn fibers across domains. *)
 let codex_omission_state_mu = Stdlib.Mutex.create ()
 let codex_omission_state : (string, string) Hashtbl.t = Hashtbl.create 16
 
-let codex_omission_should_log ~agent_name ~tool_list_key =
+let codex_cli_omission_fingerprint (tools : string list) : string =
+  tools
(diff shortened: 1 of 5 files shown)


jeong-sik:
Acknowledging #10109 (46c881f8) merged ~25min ago which addresses the same root symptom with a per-agent (rather than global-fingerprint) dedup keyed on `agent_name`.

Per-agent keying is a stronger choice than my global fingerprint:
- A tool surface change for agent A correctly fires a new WARN.
- A different agent B with a different tool set is not silently suppressed by A's fingerprint.

This PR's incremental value over #10109:
- `masc_codex_cli_mcp_tool_omission_total{tool}` per-tool counter — quantifies the omission frequency that the WARN no longer surfaces. Independent of the dedup logic.

Will leave as Draft. Happy to rebase as a counter-only follow-up (drop the dedup helper since #10109 owns that path now), or close as superseded if the counter alone is not worth a follow-up.

PR #10110, opened 2026-04-24
docs(design): WS migration performance series overview

jeong-sik:
## Why

Seven WS-transport PRs are currently open (#10089, #10096, #10098, #10102, #10104, #10106, #10107). Each is narrow and reviewable on its own, but the **shape** of the series — why these particular slices, in this order, with these design choices — is not visible from any single PR body. Reviewers have to reconstruct the context piece by piece.

This PR adds a single document that sits at `docs/design/ws-migration-perf-series.md` and serves as the entry point: why the migration started in a parallel mode, what each PR addresses, what safety anchor makes the caching choices correct, and what the rollout/rollback playbook looks like.

## What

`docs/design/ws-migration-perf-series.md` (376 lines), structured:

1. **Context** — the parallel-mode problem: both `/sse` and `/ws` open, every event delivered twice.
2. **Problem decomposition** — five categories of cost: parse amplification, byte-copy amplification, unbounded client buffer growth, duplicate delivery, opaque operational state.
3. **The seven PRs** — one section per PR, linking the cost it addresses, the safety anchor, and the counters added.
4. **Design principles applied** — physical-equality keying, read-config-on-every-call, drop-do-not-queue, literal-name metric reads at surface boundaries, observation before action, idle-state-reads-as-idle.
5. **Deferred decisions** — slice-indexed fan-out, `/sse` deletion, gRPC subscriber symmetry.
6. **Operational playbook** — rollout order, incident-triage metric table, rollback levers.

No code change.

## Why now

- Seven PRs is past the point where reviewers benefit from a unifying document. Having it before any land means the reviewers can orient themselves, and having it in-tree post-merge gives the next operator who hits a WS incident a single page to read.
- The series is feature-complete enough that the design principles won't change under us.

## Scope guard

- Pure documentation. No production code, no schema, no wire format.
- Follows the existing `docs/design/` convention (frontmatter with `status`, `last_verified`, `code_refs`; kebab-case filename; reference links).

## Follow-ups

- Once the seven PRs land, update the `status: draft` frontmatter to `status: reference` and the `last_verified` date.
- Deferred-decisions section should be revisited when slice-indexed fan-out or `/sse` deletion is actually picked up.

### DIFF
--- a/docs/design/ws-migration-perf-series.md
+++ b/docs/design/ws-migration-perf-series.md
@@ -5,6 +5,7 @@ code_refs:
   - lib/server/server_mcp_transport_ws.ml
   - lib/sse.ml
   - lib/transport_metrics.ml
+  - lib/grpc/masc_grpc_service.ml
   - dashboard/src/dashboard-ws.ts
   - dashboard/src/sse.ts
   - dashboard/src/components/transport-health.ts
@@ -14,7 +15,8 @@ code_refs:
 
 > Status: Draft
 > Date: 2026-04-25
-> PRs: #10089, #10096, #10098, #10102, #10104, #10106, #10107
+> PRs: #10089, #10096, #10098, #10102, #10104, #10106, #10107,
+> #10112, #10114, #10117, #10119 (RFC), #10120 (chore)
 
 ## 1. Context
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
Catch-up commit lands the doc updates the series accumulated since the initial draft.

What changed:

- §3 now covers all 12 entries: the original 7 plus #10112 (gRPC events_dropped), #10114 (dashboard gRPC UI), #10117 (gRPC buffer tunable), #10119 (slice-indexed fanout RFC), and #10120 (dashboard_snapshot chore).
- §5 deferred decisions: "Slice-indexed fan-out" now references RFC #10119; "gRPC subscriber symmetry" is no longer deferred — addressed by the gRPC triple.
- `code_refs` adds `lib/grpc/masc_grpc_service.ml`.
- Change log entry for 2026-04-25.

No code change. The PR stays Draft pending the series start merging.

PR #10155, opened 2026-04-25
perf(ws): slice index bookkeeping (Phase 1 of #10119)

jeong-sik:
## Summary

Implements **Phase 1** of the WS slice-indexed fanout RFC ([#10119](https://github.com/jeong-sik/masc-mcp/pull/10119)): a side index mapping each dashboard slice to the set of session IDs currently subscribed to it.  Phase 1 is bookkeeping only — the broadcast fanout path is unchanged, so behaviour is identical to today.  Phase 2 (separate PR) will consume the index to skip raw-SSE-forwards to sessions whose route does not subscribe to the event's slice.

## What changes

- `slice_index : (string, (string, unit) Hashtbl.t) Hashtbl.t` in `lib/server/server_mcp_transport_ws.ml` (module-level, mutated under `sessions_mutex`).
- Hooks into:
  - `dashboard_subscribe` — clears prior entries, adds new slices.
  - `dashboard_unsubscribe` — removes specified slices (or all).
  - `cleanup_session` — sweeps every slice the session was in.
- Empty inner sets are pruned so `slice_index_size` is honest.
- Public helpers: `slice_index_subscribers`, `slice_index_size`.
- Test-only entry points (`__test_slice_index_*`) drive the index directly — the dashboard subscribe path requires a fully-constructed `Wsd.t` which only integration tests build.

## What does not change

- Sse fanout: still iterates every external subscriber.
- Wire format: no message changes.
- Public API for non-test code: `dashboard_subscribe` / `dashboard_unsubscribe` / `cleanup_session` keep their signatures.

## Test plan

- 6 new test cases under `slice_index` group:
  - unknown slice yields no subscribers
  - add records session under slice
  - remove targets only the named slice
  - remove_session sweeps every slice
  - size tracks (slice × session) pair count (delta-based, shared global state friendly)
  - duplicate add is idempotent
- Full `test_ws_transport.exe` passes (33 / 33).
- Full `dune build` clean.

## Phase 2 follow-up

Will rewire the WS callback registered with `Sse.subscribe_external` to consult `slice_index` for slice-scoped events (mapping known event types to slices, falling back to raw forward when the event has no slice mapping).  RFC §4.2 lays out the proposed structure; the index added here is exactly its input.

### DIFF
--- a/lib/server/server_mcp_transport_ws.ml
+++ b/lib/server/server_mcp_transport_ws.ml
@@ -47,6 +47,74 @@ let sessions_mutex = Eio.Mutex.create ()
 
 let with_sessions_rw f = Eio_guard.with_mutex sessions_mutex f
 
+(** Side index mapping each dashboard slice to the set of session IDs
+    currently subscribed to it.  Phase 1 of the slice-indexed fanout RFC
+    (#10119): the index is maintained at subscribe / unsubscribe / cleanup
+    time but the broadcast fanout still iterates every session.  Phase 2
+    consults this index to skip raw-SSE-forwards to sessions whose route
+    does not subscribe to the event's slice.
+
+    Inner [Hashtbl.t] is keyed by [session_id] with [unit] values — a set.
+    Mutated only under [sessions_mutex], so the consistency invariant
+    matches [sessions]: a session present in [slice_index.(s)] is present
+    in [sessions] until [cleanup_session] runs. *)
+let slice_index : (string, (string, unit) Hashtbl.t) Hashtbl.t =
+  Hashtbl.create 8
+
+let slice_index_add_locked ~session_id ~slice =
+  let set =
+    match Hashtbl.find_opt slice_index slice with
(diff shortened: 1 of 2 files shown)


jeong-sik:
Review finding: the new slice index invariant can be broken by the existing lookup/mutation split. The comment says any session in `slice_index` is still present in `sessions` until cleanup (`lib/server/server_mcp_transport_ws.ml:50-60`), and cleanup removes the id under the mutex (`:603-615`). But `dashboard_subscribe`/`dashboard_unsubscribe` first call `find_session`, release the lock, then later mutate the slice index under a second lock (`:371-405`) without rechecking that `sessions` still contains the id. If cleanup runs between those two points, subscribe can re-add a closed/removed session id to the index after cleanup already swept it. Please make the lookup + index/session mutation atomic under `sessions_mutex` or revalidate `Hashtbl.find_opt sessions session_id` inside the locked block before adding to `slice_index`.

PR #10160, opened 2026-04-25
perf(ws): slice-aware fanout gate (Phase 2 of #10119)

jeong-sik:
Stacked on #10155 (Phase 1). Implements Phase 2 of WS slice-indexed fanout RFC #10119. New env MASC_WS_SLICE_INDEX_ENABLED (default false) gates slice-aware fanout; new counter masc_ws_slice_fanout_skipped_total. send_dashboard_or_raw_sse splits None-branch via parse cache. 36/36 tests pass. Default-off rollout per RFC §10. When parent squash-merges, will rebase to main.

### DIFF
--- a/lib/config/env_config_snapshot.ml
+++ b/lib/config/env_config_snapshot.ml
@@ -234,6 +234,11 @@ let transport_entries =
     entry ~default:"1048576" "MASC_WS_CLIENT_BUFFER_LIMIT_BYTES"
       "Skip WS dashboard deltas for authenticated sessions whose last reported \
        WebSocket.bufferedAmount exceeds this many bytes. 0 disables the gate.";
+    entry ~default:"false" "MASC_WS_SLICE_INDEX_ENABLED"
+      "When true, slice-scoped events skip the raw-SSE-forward to authenticated \
+       WS sessions whose route does not subscribe to the event's slice. \
+       Catch-all events (no slice mapping) still reach every session. \
+       masc_ws_slice_fanout_skipped_total advances per skip. RFC #10119 Phase 2.";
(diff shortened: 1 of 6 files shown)


jeong-sik:
Review finding: Phase 2 changes the actual WS delivery path, but the new tests do not exercise that branch. The server now skips raw SSE forwarding for a parsed slice mismatch when `MASC_WS_SLICE_INDEX_ENABLED` is true (`lib/server/server_mcp_transport_ws.ml:572-579`), while the tests only check metric registration and env flag parsing (`test/test_ws_transport.ml:512-531`). Please add a test that drives an authenticated subscribed session with a mismatched slice and verifies the delivery is skipped/counter increments, plus a catch-all/no-slice event that still forwards. Also, this path currently does not consult the `slice_index` built in Phase 1; if this PR is meant to be the indexed fanout phase, the index remains side state rather than the fanout source of truth.

jeong-sik:
Additional review finding after the head/base refresh: the slice index invariant can still be broken by the lookup/mutation split. The module comment says any id in `slice_index` is still present in `sessions` until cleanup (`lib/server/server_mcp_transport_ws.ml:50-60`), and cleanup removes the id under the mutex. But `dashboard_subscribe`/`dashboard_unsubscribe` get `session` via `find_session`, release the mutex, then later mutate `slice_index` under a second lock without revalidating `Hashtbl.find_opt sessions session_id` (`:371-405`). If cleanup runs between lookup and mutation, subscribe can re-add a closed/removed id after cleanup swept it. Please make lookup plus session/index mutation atomic, or re-check the session table inside the locked block before adding to `slice_index`.

Issue #9768, opened 2026-04-23
ramarama keeper stuck on ollama_only cascade after TOML config change


jeong-sik:
## Symptom
Fleet logs show `ramarama` keeper repeatedly failing with `cascade=ollama_only`:
- `[ERROR] ramarama: keeper cycle FAILED cascade=ollama_only ... error=Timeout: Timeout after 276.0s`
- Occurred at 18:23:42Z, 18:29:28Z (and likely more)

## Expected
`ramarama.toml` was updated to `cascade_name = \"local_only\"` (from `\"ollama_only\"`),
so the keeper should use `local_only`.

## Actual
Logs still show `cascade=ollama_only` after the TOML change.

## Possible causes
1. Config file change not persisted to the path the server reads
2. Server caches cascade config and does not hot-reload
3. Runtime JSON / other config source overrides the TOML value
4. Keeper process was spawned before the config change and is long-lived

## Impact
ramarama keeper is in a tight failure loop (timeout → retry → timeout),
consuming OAS budget and generating noise.

## Next step
Verify config propagation path: TOML → runtime JSON → hot reload → keeper spawn.

jeong-sik:
**Update**: TOML file at `~/.masc/config/keepers/ramarama.toml` already shows `cascade_name = \"local_only\"` (line 70). The drift is not in the config file itself but in propagation to the running keeper process. Possible root causes narrowed to:

1. Keeper reads cascade config at spawn time only — no hot-reload on TOML change
2. Runtime JSON (e.g., `progress_snapshot.json` or meta store) overrides the TOML value
3. A long-lived keeper process was spawned before the TOML change

Server restart or explicit config reload may be required.

jeong-sik:
Resolved by recent fixes in PR #9760 and #9797:
- oas_env field restored in keeper_meta
- MASC_KEEPER_OAS_* whitelist fix
- Default turn budgets bumped (15→30, 2→10)
- TOML network_mode corrected to inherit/none

Closing as resolved. Reopen if symptoms persist after next server restart.

Issue #9572, opened 2026-04-23
ops: nick0cave keeper fails due to unregistered cascade_name


jeong-sik:
## Problem

Keeper \`nick0cave\` is configured with \`cascade_name = \"nick0cave\"\` in its TOML, but this cascade is not registered in the live runtime \`cascade.json\`. This causes every keeper cycle to fail immediately with:

```
ERROR Invalid config 'cascade_name': unknown cascade_name \"nick0cave\"
```

The keeper is effectively dead — it cannot claim tasks, execute turns, or contribute to the swarm.

## Log evidence (last 100 lines)

- \`cascade config resolve failed for nick0cave, returning []\` — repeated 9×
- \`keeper cycle FAILED cascade=nick0cave\` — latency ~12ms (immediate config rejection)
- \`TOML reconcile failed for nick0cave\` — cascade_name validation fails at load time

## Impact

- 1 of 16 keepers is completely non-functional.
- Error noise dilutes legitimate operational alerts.

## Options

1. **Register \`nick0cave\` in live \`cascade.json\`** (preferred if the keeper is intended to use a custom model lane).
2. **Change \`nick0cave.toml\` to use \`cascade_name = \"big_three\"\`** (fastest fix if custom lane is not required).

## Files

- Live config: \`~/me/.masc/config/cascade.json\`
- Keeper TOML: \`~/me/.masc/config/keepers/nick0cave.toml\`

jeong-sik:
Resolved by #9708 (Docker sandbox git freedom — safe.directory, SSH agent, keeper identity). The unregistered cascade_name issue was part of the keeper identity/bootstrap fixes.

PR #10165, opened 2026-04-25
fix(keeper): keep reserved cascade lanes loadable

jeong-sik:
## Summary
- keep reserved keeper cascade lanes, including tool_use_strict, valid even when the live cascade catalog is temporarily unavailable
- add a result-returning catalog-name API so TOML validation can surface catalog load errors instead of silently treating the catalog as empty
- make invalid cascade_name diagnostics list the actual reserved + live catalog names

Fixes #10158.

## Verification
- env DUNE_BUILD_DIR=_build_verify_10158 DUNE_JOBS=2 scripts/dune-local.sh build test/test_keeper_toml_config_validation.exe
- env DUNE_BUILD_DIR=_build_verify_10158 _build_verify_10158/default/test/test_keeper_toml_config_validation.exe
- env DUNE_BUILD_DIR=_build_verify_10158 DUNE_JOBS=2 scripts/dune-local.sh build test/test_keeper_cascade_profile.exe
- env DUNE_BUILD_DIR=_build_verify_10158 _build_verify_10158/default/test/test_keeper_cascade_profile.exe
- git diff --check HEAD^..HEAD

PR #10170, opened 2026-04-25
fix(keeper-toml): operator-visible cascade_name validator + load_catalog WARN (#10158)

jeong-sik:
## Summary

#10158 reports 6 keeper TOML configs silently rejected at boot:

\`\`\`
toml_loader: skipping nick0cave.toml: invalid cascade_name 'ollama_only'
  (known: big_three, tool_rerank, local_only, local_recovery)
\`\`\`

Two distinct sub-bugs underneath:

### 1. Display drift (Option B)

The message printed only \`compile_known @ phase_routing\` (4 names).  But the validator's actual \`all_valid\` set also includes \`Keeper_cascade_profile.catalog_names ()\` from cascade.toml.  Operators trying \`tool_use_strict\` saw a rejection that didn't even mention the name — they had no way to tell "this name is invalid" from "the system does not support it."

**Fix**: print every valid name, sorted + deduped.

### 2. Silent catalog load failure (Option C — code half)

\`Keeper_cascade_profile.catalog_entries\` swallowed \`Cascade_config_loader.load_catalog\`'s \`Error _\` and returned \`None\`.  Every downstream caller then operated on \`catalog_names () = []\`, which made the validator reject every cascade.toml-defined name.  The only signal was the cascading "invalid" rejections — operators couldn't distinguish "load failed" from "name unknown."

**Fix**: once-per-process \`Log.Keeper.warn\` at the first \`Error e\`, reset on a later successful load so a fixed config flips the alarm off.

\`\`\`ocaml
let cascade_load_warn_emitted : bool Atomic.t = Atomic.make false
...
| Error e ->
    if not (Atomic.exchange cascade_load_warn_emitted true) then
      Log.Keeper.warn "cascade catalog load failed for %s: %s — ..." path e;
    None
\`\`\`

## Not in this PR (per issue resolution table)

| option | scope | why deferred |
|---|---|---|
| **A** (cascade.toml + keeper TOMLs config drift) | repo config edit | separate from code change; needs operator decision (define \`ollama_only\` vs migrate keepers to \`local_only\`) |
| **D** (silent skip → fail-loud / startup blocker) | larger policy change | escalating WARN → ERROR or startup-block needs review and rollout plan |

## Test plan

- [x] **new** \`test_invalid_cascade_name_message_includes_catalog\` — pins catalog entries appear in the rejection message when catalog loads
- [x] **new** \`test_invalid_cascade_name_message_sorted_and_unique\` — pins sort + dedup (stable, readable output regardless of catalog ordering)
- [x] Existing \`test_cascade_name_accepts_catalog_entry\` continues to pass
- [ ] Local environment caveat: \`test_cascade_name_rejects_unknown\` fails on my machine because \`~/me/.masc/config/cascade.json\` happens to register \`nick0cave\` as a catalog entry (so the test fixture name is accidentally valid).  Pre-existing test fragility independent of this PR — confirm CI passes (CI's HOME is clean).

## Refs

- #9572, #9768, #6747 (CLOSED) — same-shape regressions
- iter62 #9733 follow-up: same keepers, different layer

## Do-not-merge

### DIFF
--- a/lib/keeper/keeper_cascade_profile.ml
+++ b/lib/keeper/keeper_cascade_profile.ml
@@ -59,6 +59,18 @@ let canonical (raw : string) : t =
   | Some t -> t
   | None -> default
 
+(* #10158: cascade.toml load failures used to silently return [None].
+   That made every downstream caller (catalog_names,
+   keeper_catalog_names, system_catalog_names) return [], which
+   made [keeper_types_profile] reject every TOML keeper config
+   whose cascade_name lived only in cascade.toml
+   ([tool_use_strict] etc.) — six configs were silently dropped
+   on 2026-04-25.  Surface the failure once per process so the
+   root cause is visible without spamming dashboard render
+   loops; reset on success so a fixed config flips the alarm
+   off. *)
+let cascade_load_warn_emitted : bool Atomic.t = Atomic.make false
+
 let catalog_entries ?config_path () =
   let path_opt =
     match config_path with
@@ -69,8 +81,17 @@ let catalog_entries ?config_path () =
   | None -> None
   | Some path -> (
       match Cascade_config_loader.load_catalog ~config_path:path with
-      | Ok entries -> Some entries
-      | Error _ -> None)
+      | Ok entries ->
+          Atomic.set cascade_load_warn_emitted false;
+          Some entries
+      | Error e ->
+          if not (Atomic.exchange cascade_load_warn_emitted true) then
+            Log.Keeper.warn
+              "cascade catalog load failed for %s: %s — keeper TOML \
+               configs whose cascade_name is defined only in \
+               cascade.toml will be silently rejected (see #10158)"
+              path e;
+          None)
 
 let catalog_names ?config_path () =
   match catalog_entries ?config_path () with
--- a/lib/keeper/keeper_types_profile.ml
+++ b/lib/keeper/keeper_types_profile.ml
@@ -654,12 +654,21 @@ let profile_defaults_of_toml (doc : Keeper_toml_loader.toml_doc)
             let all_valid = compile_known @ phase_routing @ catalog in
             if List.mem normalized all_valid then Ok ()
             else
+              (* #10158: include cascade.toml [catalog] entries in the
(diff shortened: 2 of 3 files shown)


jeong-sik:
Review finding: this improves operator visibility, but the regression test still depends on the active/global catalog and can skip the core assertion in clean CI. `test_invalid_cascade_name_message_includes_catalog` reads `Keeper_cascade_profile.catalog_names ()`, then returns `()` when no unique catalog name is found (`test/test_keeper_toml_config_validation.ml:149-165`). That means a runner with no materialized catalog does not prove the new `(known: ...)` message includes live catalog names. Please make this hermetic by writing a temp `cascade.json` fixture and calling the loader/validator through that config root, or by exposing an injectable/catalog-path validation helper; otherwise the most important behavior remains environment-dependent.

jeong-sik:
## Closing as superseded by #10165 (MERGED)

Verified 2026-04-25 (/loop tick #51): PR #10165 \`fix(keeper): keep reserved cascade lanes loadable\` (MERGED) implements a stricter superset of this PR's two fixes:

| concern | this PR | #10165 |
|---|---|---|
| **Validator message includes catalog entries** | sorted + deduped \`all_valid\` (B) | same shape: `String.concat ", " all_valid` after `sort_uniq` |
| **Catalog load error visibility** | once-per-process \`Log.Keeper.warn\` in \`Keeper_cascade_profile\` (C) | structurally cleaner — new \`catalog_names_result\` returns the error and the validator embeds it in the rejection message itself: \`"... (reserved: ...; live catalog unavailable: <error>)"\` |
| **Reserved-name fast path** | not addressed | new \`reserved_cascade_names\` SSOT lets \`tool_use_strict\` (and other reserved tool lanes) load even when cascade.toml is unreadable |

The Result-based design in #10165 is better than my Atomic-flag once-per-process WARN — it puts the catalog error in front of the operator at the exact rejection site, no log-grep required. Closing this PR; my work was redundant.

Refs:
- PR #10165 (commit \`7a1d47a1e8\`) — merged superset
- Original issue #10158

PR #10174, opened 2026-04-25
fix(dashboard): surface keeper execution receipts in composite

jeong-sik:
## Summary
- enrich keeper composite API responses with compact latest execution receipt summaries
- show receipt outcome/terminal reason/duration/fallback/error preview in the FSM Hub
- keep fleet polling bounded by truncating long receipt error messages

## Why it matters
Operators were seeing keepers as Running/idle/done while the actual blocker lived only in execution receipts, for example config materialization failures, max-turn/cascade exhaustion, timeout fallback, or tool contract failures. This makes the composite dashboard answer why a keeper is stuck or slow without tailing logs.

## Verification
- scripts/dune-local.sh build test/test_dashboard_keeper_routes.exe bin/main_eio.exe
- ./_build/default/test/test_dashboard_keeper_routes.exe
- pnpm --dir dashboard exec vitest run src/api/schemas/keeper-composite.test.ts src/components/fsm-hub.test.ts --no-file-parallelism --maxWorkers=1
- live 8935: /api/v1/keepers/composite status=200 time_total=0.010452, execution field present for all 14 snapshots
- live 8935: /api/v1/dashboard/shell status=200 time_total=0.012330

### DIFF
--- a/dashboard/src/api/keeper.ts
+++ b/dashboard/src/api/keeper.ts
@@ -28,6 +28,7 @@ export type {
   KeeperCompositeInvariants,
   KeeperCompositeMeasurement,
   KeeperLastOutcome,
+  KeeperCompositeExecution,
   KeeperCompositePhase,
   KeeperCompositeTurnPhase,
   KeeperCompositeDecisionStage,
--- a/dashboard/src/api/schemas/keeper-composite.test.ts
+++ b/dashboard/src/api/schemas/keeper-composite.test.ts
@@ -77,6 +77,50 @@ describe('parseKeeperCompositeSnapshot', () => {
     expect(result.last_outcome!.selected_model).toBe('claude-sonnet')
   })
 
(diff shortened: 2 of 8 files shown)


jeong-sik:
Review finding: the new composite execution projection reads a timestamp field that real execution receipts do not write. `composite_execution_receipt_json` maps `recorded_at` from `json_string "recorded_at" receipt` at `lib/server/server_dashboard_http.ml:529-530`, but `Keeper_execution_receipt.to_json` emits `started_at` and `ended_at` at `lib/keeper/keeper_execution_receipt.ml:333-334`, with no `recorded_at` field. The result is that every real composite payload has `execution.recorded_at = null`, and the dashboard title loses the most important operator context for stale receipts. Please map this from `ended_at` or `started_at` (or rename the field to the receipt field actually emitted) and assert it in the route test, not just the TypeScript parser fixture.

PR #10196, opened 2026-04-25
fix(keeper): use real tools for message sweep affordance

jeong-sik:
## Summary
- Replace nonexistent message-sweep affordance tool names with real tool surface names.
- Add an SSOT guard so every affordance tool maps to Tool_name or Tool_catalog_surfaces.

## Verification
- `./_build/default/test/test_keeper_unified.exe test tool_classification`

## Notes
- This catches typoed/dead hardcoded tool names such as `keeper_keeper_msg` before they can silently weaken keeper tool-gate behavior.

### DIFF
--- a/lib/keeper/keeper_agent_tool_surface.ml
+++ b/lib/keeper/keeper_agent_tool_surface.ml
@@ -101,8 +101,7 @@ let turn_affordances_require_tool_gate turn_affordances =
 let tools_for_gated_affordance = function
   | Board_post_or_comment ->
     [ "keeper_board_post"; "keeper_board_comment"; "masc_broadcast" ]
-  | Message_sweep ->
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed the message-sweep affordance fix. I did not find a blocking issue: the gated affordance now references actual message-capable MASC tools instead of nonexistent keeper_* names, and the regression test checks every affordance mapping against Tool_name / Tool_catalog surface SSOTs so future typos are caught.

Evidence checked:
- Diff reviewed: lib/keeper/keeper_agent_tool_surface.ml, test/test_keeper_unified.ml
- git diff --check: clean
- GitHub CI on head 17a253d44acb7cc7272e12cd92850ea50f5eaab8: Build and Test, Lint, Health, TLA+, CI Gate all succeeded

Issue #10192, opened 2026-04-25
CI Gate can fail after merge when approval label is cleaned up


jeong-sik:
## Symptom

PR #10181 was merged at 2026-04-25T06:17:55Z, but the same CI run later reported `CI Gate` failure at 2026-04-25T06:29:29Z.

The gate log showed:

- live PR draft state: `false`
- live PR labels: `enhancement`
- failure: agent-like PR `codex/dehardcode-keeper-tool-routing` was ready without `human-approved-ready`

All substantive checks had passed: Build and Test, Health, Lint, TLA+ Model Checking, Meta Guards, PR Sync Check.

## Reproduction / trigger

A PR with an agent-like branch is approved/ready and auto-merged. If `human-approved-ready` is removed or absent by the time the late `CI Gate` job queries live PR labels, the gate evaluates the already-merged PR as ready without bypass and fails.

Observed on:

- PR: #10181
- CI run: 24924336433
- failing job: 72992513952
- merge commit reported by PR: 67ac42462a2fbfbe7bb8f414897b95243b81de00

## Suspected cause

`CI Gate` reads live PR labels late, while label state can change after auto-merge or merge automation. The gate is therefore not evaluating the same approval snapshot that allowed merge.

## Tried

Verified PR state and status rollup via `gh pr view 10181 --json state,mergedAt,mergeCommit,statusCheckRollup,labels`; confirmed the PR is `MERGED` even though `CI Gate` concluded failure afterward.

## Desired behavior

The merge gate should avoid post-merge false red checks. Options: snapshot labels at merge-gate time, skip this gate when the PR is already merged, or make label cleanup happen only after all gate jobs have concluded.

jeong-sik:
Opened draft PR #10202 for this.

Implemented path:
- CI Gate now refreshes live PR `state` alongside draft state and labels.
- `scripts/ci/check-agent-draft-policy.sh` skips enforcement for live `MERGED`/`CLOSED` PRs, so post-merge label cleanup cannot turn an already-merged PR red.
- Open ready agent PRs without `human-approved-ready` still fail.

Local verification passed:
- `git diff --check`
- `bash -n scripts/ci/check-agent-draft-policy.sh`
- direct script repro: `PR_LIVE_STATE=MERGED` passes
- direct script control: open ready agent PR without bypass still fails
- focused Dune build and `test_ci_hardening_source.exe` (35 tests)

jeong-sik:
[...]

PR #10200, opened 2026-04-25
fix(ci-gate): skip agent draft policy when PR is already merged (#10192)

jeong-sik:
## Summary

#10192 reports the CI Gate flipping a merged PR to red ~12 minutes after merge.  PR #10181 merged at 06:17:55Z; the same CI run reported "agent-like PR ready without \`human-approved-ready\`" at 06:29:29Z.  All substantive checks (Build, Health, Lint, TLA+, Meta Guards, PR Sync) had passed at merge time.

## Root cause

The agent-draft policy step reads \`PR_LIVE_LABELS\` from \`gh pr view\` at the step's start time, not from the merge-time snapshot.  When auto-merge fires earlier than this gate, the gate sees the post-merge cleanup state (bypass label removed by automation) and concludes "ready without bypass" — for a PR that has already shipped.

## Fix (issue Option 2: skip-when-merged)

\`scripts/ci/check-agent-draft-policy.sh\` now skips when \`PR_LIVE_STATE\` reports \`MERGED\` or \`CLOSED\`.  \`ci.yml\` exports the state via the same \`gh pr view --json state\` call already used for \`isDraft\` and \`labels\`.  Case-insensitive match.

\`\`\`bash
case "$(lower "${pr_live_state}")" in
  merged|closed)
    echo "agent draft policy: skipped, PR live state is ${pr_live_state} (post-merge gate race, see #10192)"
    exit 0
    ;;
esac
\`\`\`

## Why option 2

| option | analysis |
|---|---|
| 1 — snapshot labels at merge-gate time | requires ordering changes or a separate snapshot store; doesn't survive a workflow re-run from a stale point |
| 2 — **skip-when-merged (this PR)** | surgical to the gate itself; treats the "PR has shipped" race as a no-op; OPEN behaviour unchanged |
| 3 — delay label cleanup | couples cleanup automation to gate completion; more fragile across multiple workflows |

The script remains the authoritative gate at merge time — when the PR is OPEN, all existing logic runs unchanged.

## Test plan

- [x] \`dune exec test/test_ci_hardening_source.exe\` — 35/35 pass:
  - new MERGED state → exit 0 (skipped)
  - new CLOSED state → exit 0 (skipped)
  - new lower-case \`merged\` → exit 0 (case insensitivity)
  - new OPEN state → still fails (policy still enforced)
  - new source-guard check pins \`PR_LIVE_STATE\` export in ci.yml so a future workflow refactor can't silently drop it
- [ ] Deploy: rerun CI Gate on a recently-merged PR; confirm the gate skips with the expected notice instead of flipping red.

## Refs

- PR #10181 (the merged PR that flipped red)
- CI run 24924336433, failing job 72992513952

## Do-not-merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -725,6 +725,14 @@ jobs:
             if [ -n "$live_labels" ]; then
               export PR_LIVE_LABELS="$live_labels"
             fi
+            # #10192: CI Gate runs that fire after auto-merge can
+            # see the bypass label cleaned up by post-merge
+            # automation; pass live state to the policy script so
+            # it skips the check when the PR is no longer open.
+            live_state="$(gh pr view "$PR_NUMBER" --repo "$GITHUB_REPOSITORY" --json state --jq '.state' 2>/dev/null || true)"
+            if [ -n "$live_state" ]; then
+              export PR_LIVE_STATE="$live_state"
+            fi
           fi
           scripts/ci/check-agent-draft-policy.sh
 
--- a/scripts/ci/check-agent-draft-policy.sh
+++ b/scripts/ci/check-agent-draft-policy.sh
@@ -15,6 +15,7 @@ pr_is_draft="${PR_LIVE_IS_DRAFT:-${PR_IS_DRAFT:-false}}"
 pr_title="${PR_TITLE:-}"
 pr_head_ref="${PR_HEAD_REF:-}"
 pr_labels_csv="${PR_LIVE_LABELS:-${PR_LABELS:-}}"
+pr_live_state="${PR_LIVE_STATE:-}"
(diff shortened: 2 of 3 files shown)


jeong-sik (commented):
Reviewed from CI gate/race-condition angle.

No blocking findings. The skip is scoped to live `MERGED|CLOSED` state, so OPEN PR behavior remains governed by the existing draft/bypass policy. The source guard pins the `PR_LIVE_STATE` export so this cannot silently regress in the workflow.

Local verification:
- `scripts/dune-local.sh build test/test_ci_hardening_source.exe`

I also converted the PR back to draft; the body already described it as a Draft and it carries `do-not-merge`.

PR #10204, opened 2026-04-25
fix(keeper): type tools-list grouping

jeong-sik:
## Summary
- replace keeper_tools_list_json prefix heuristics with typed Tool_name grouping
- keep the existing operator-facing group names while classifying through the typed catalog
- add a regression that keeper_board_fake is not grouped while canonical tools still land in the expected groups

## Why it matters
This removes a UI-facing string-prefix heuristic from the keeper tool-list surface. The grouping now follows typed tool identifiers instead of accepting board-looking or voice-looking names by spelling alone.

## Validation
- scripts/dune-local.sh build test/test_keeper_exec_tools.exe
- ./_build/default/test/test_keeper_exec_tools.exe test keeper_tools_list_json (1 test, run ID FD519ZOM)

## Note
- Full ./_build/default/test/test_keeper_exec_tools.exe was attempted and the new keeper_tools_list_json test passed, but the pre-existing exec_cache/miss-then-hit case failed locally because Docker socket access was denied at unix:///Users/dancer/.docker/run/docker.sock.

### DIFF
--- a/lib/keeper/keeper_exec_shared.ml
+++ b/lib/keeper/keeper_exec_shared.ml
@@ -382,14 +382,57 @@ let tag_dispatch_fn
 
 let keeper_tools_list_json ~(meta : keeper_meta) =
   let names = Keeper_tool_policy.keeper_allowed_tool_names meta in
+  let categorize_keeper_tool = function
+    | Tool_name.Keeper.Board_cleanup
+    | Tool_name.Keeper.Board_comment
+    | Tool_name.Keeper.Board_comment_vote
+    | Tool_name.Keeper.Board_delete
+    | Tool_name.Keeper.Board_get
+    | Tool_name.Keeper.Board_list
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed the typed tools-list grouping change from the tool-contract angle. I did not find a blocking issue: the public grouping path now parses through Tool_name before categorizing, so prefix-shaped fake names do not drive keeper tool-list grouping, while keeper-facing group labels stay compatible with the existing JSON shape.

Local verification:
- scripts/dune-local.sh build test/test_keeper_exec_tools.exe
- ./_build/default/test/test_keeper_exec_tools.exe test 'keeper_tools_list_json' (1 focused test passed)

Note: running the full test executable still hits the pre-existing Docker socket denial in exec_cache, matching the PR note; the new keeper_tools_list_json test passed in that full run before the unrelated failure.

PR #10207, opened 2026-04-25
refactor(transport): catalog openapi operation tags

jeong-sik:
## Summary
- replace the OpenAPI operation tag if/else string chain with an explicit tag-group catalog
- keep default  tagging behavior for uncategorized operations
- add OpenAPI document coverage for task, planning, messaging, default, and REST path tags

## Why this matters
The websocket/transport follow-up reduced route mapping drift, but tag assignment still lived in a hand-maintained OR-chain. This makes the classification table easier to audit and less likely to drift when transport-facing operations are added or moved.

## Validation
- git diff --check
- scripts/dune-local.sh build test/test_transport_coverage.exe
- ./_build/default/test/test_transport_coverage.exe

### DIFF
--- a/lib/transport.ml
+++ b/lib/transport.ml
@@ -319,47 +319,54 @@ module Rest = struct
               prompt_hints = [];
             })
 
+  let operation_tag_groups =
+    [
+      ( "transport",
+        [
+          "masc_transport_status";
+          "masc_websocket_discovery";
+          "masc_webrtc_offer";
+          "masc_webrtc_answer";
+        ] );
+      ( "tasks",
+        [
+          "masc_status";
+          "masc_tasks";
+          "masc_add_task";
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed the OpenAPI tag catalog refactor. I did not find a blocking issue: the string ladder was moved into a single operation_tag_groups table, default behavior remains ["masc"], and tests now assert representative generated x-mcp-operation and REST operation tags.

Evidence checked:
- Diff reviewed: lib/transport.ml, test/test_transport_coverage.ml
- git diff --check: clean
- GitHub CI on head a9e53e4ce5ff3cf65ece619a2511991729a897da: Build and Test, Lint, Health, CI Gate all succeeded; Dashboard/TLA skipped

PR #10210, opened 2026-04-25
fix(ws): parse sse data lines without fixed position

jeong-sik:
## Summary
- replace positional SSE data extraction with a line-based parser
- accept both spaced and unspaced SSE data payload lines
- keep pure JSON passthrough for existing unit-test and direct parser callers
- add a regression case with an inserted SSE comment line before data

## Why this matters
The recent WS parser fix correctly peeled the SSE wrapper, but it still assumed the payload lived on the third line with an exact spaced data prefix. This removes that hidden wire-format coupling while preserving the current production format.

## Validation
- git diff --check
- scripts/dune-local.sh build test/test_ws_transport.exe
- ./_build/default/test/test_ws_transport.exe

### DIFF
--- a/lib/server/server_mcp_transport_ws.ml
+++ b/lib/server/server_mcp_transport_ws.ml
@@ -465,13 +465,12 @@ let parse_cache : (string * parsed_sse_event option) Atomic.t =
 
 (** Extract the JSON body from an SSE-formatted event string.
 
-    Sse.format_event always emits exactly four lines:
-    [{|id: <N>\nevent: <type>\ndata: <body>\n\n|}].  The body line
-    starts with [data: ] (6-char prefix) and is the JSON payload that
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed the WS SSE data-line parser change. I did not find a blocking issue: the parser now searches for SSE data: lines instead of assuming the third line, accepts both 'data: <json>' and 'data:<json>', keeps pure-JSON passthrough for tests, and handles multiple data lines by joining them with newlines.

Evidence checked:
- Diff reviewed: lib/server/server_mcp_transport_ws.ml, test/test_ws_transport.ml
- git diff --check: clean
- CI on head 3fbc632bd5e65b885263d5daac521f152b138682: Health and Lint passed; Build and Test still in progress at review time; Dashboard/TLA skipped. I did not add a local Dune run while the shared Dune lock is held by another repo-wide run.

Issue #10205, opened 2026-04-25
Post-merge /simplify findings: auth predicate sprawl, fs_compat literal scatter + mkdir reinvention, boot-path sweep


jeong-sik:
## Summary

Post-hoc `/simplify` review of 4 critical session PRs (#10127, #10131, #10135, #10187) by 3 parallel review agents (reuse/quality/efficiency). All PRs are merged; this is a single follow-up issue with the cross-cutting findings.

### 1. `auth.ml:773-787` — 3 sibling prefix predicates → data-driven list

`is_masc_tool_name`, `is_protocol_canonical_tool_name`, `is_keeper_runtime_tool_name` are 3 hand-rolled `String.starts_with` predicates joined by `||` chains at lines 783-785 + 793-795. Each new prefix requires editing 2 gates + 1 .mli.

**Suggested fix**:

```ocaml
let internal_tool_prefixes = [
  "masc_";
  "decision.";
  "experiment.";
  "client.";
  "keeper_";
]

let is_internal_tool_prefix name =
  List.exists (fun pref -> String.starts_with ~prefix:pref name) internal_tool_prefixes
```

Replace 3 predicates + 2 `||` chains with one helper. Adding a new prefix becomes a 1-line list edit.

### 2. `fs_compat.ml` — `.atomic_*.tmp` literals scattered

- Writer at `:191`: `Filename.temp_file ~temp_dir:dir ".atomic_" ".tmp"`
- Reader at `:213-220`: `let prefix = ".atomic_" and suffix = ".tmp"`

Two sites must agree. Hardcoded scattering = drift class from `instructions/software-development.md` (#1 anti-pattern).

**Suggested fix**: extract module-level constants `atomic_tmp_prefix`/`atomic_tmp_suffix`, used by both sites.

### 3. `fs_compat.ml:229-232` — `ensure_recovered_dir` reinvents `mkdir_p`

```ocaml
(* :229 NEW *)
if not (Sys.file_exists recovered_dir) then
  try Unix.mkdir recovered_dir 0o755
  with Unix.Unix_error _ -> ()
```

Same file at `:323` already has `Fs_compat.mkdir_p` (recursive, idempotent, `~exists_ok:true`).

**Suggested fix**: replace with `mkdir_p_unix recovered_dir` (same file, no new dep).

### 4. `fs_compat.ml:316-317` — double readdir

[...]

PR #10188, opened 2026-04-25
fix(auth): allow keeper runtime tools in strict mode

jeong-sik:
## Summary

Follow-up to #10183 after #10187 landed the broad keeper_* strict-auth recovery.

This PR tightens that recovery path so strict auth treats only cataloged Keeper_internal tools as unmapped internal tools, instead of allowing every keeper_* prefix. It also makes future unknown external strict-auth denials observable through a bounded Prometheus counter.

## Changes

- Replace the keeper_* prefix fallback with Tool_catalog.Keeper_internal membership for unmapped internal tools.
- Keep masc_* and protocol-canonical tool names as internal strict-mode fallbacks.
- Add `masc_auth_strict_unknown_tool_denials_total{agent_name,tool_class}` for denied unknown external tools.
- Cover both token auth and role-auth strict gates.
- Add a negative regression for fake keeper-prefixed names such as `keeper_totally_fake`.

## Verification

- `git diff --check origin/main...HEAD`
- `scripts/check-boundary-guard.sh`
- `env DUNE_BUILD_DIR=_build_10183_keeper_auth_rebase DUNE_JOBS=2 DUNE_CACHE=disabled MASC_DUNE_THROTTLE=0 scripts/dune-local.sh build test/test_auth.exe test/test_tool_access_role.exe test/test_prometheus_coverage.exe`
- `env MASC_BASE_PATH=/tmp/masc-test-10183-auth-rebase _build_10183_keeper_auth_rebase/default/test/test_auth.exe`
- `env MASC_BASE_PATH=/tmp/masc-test-10183-role-rebase _build_10183_keeper_auth_rebase/default/test/test_tool_access_role.exe`
- `env MASC_BASE_PATH=/tmp/masc-test-10183-prom-rebase _build_10183_keeper_auth_rebase/default/test/test_prometheus_coverage.exe`

Fixes #10183.

PR #10218, opened 2026-04-25
refactor(fs-compat): SSOT atomic-tmp constants + reuse mkdir_p_unix + single readdir (#10205)

jeong-sik:
## Summary

Three of the five findings from #10205's post-hoc \`/simplify\` review of PRs #10127/#10131/#10135/#10187, all in \`lib/fs_compat/fs_compat.ml\`.

| finding | scope | status |
|---|---|---|
| 1 — auth predicate sprawl | \`lib/auth.ml\` | **NOT in this PR** — PR #10188 (MERGED) landed a stricter superset using \`Tool_catalog.is_on_surface Tool_catalog.Keeper_internal\` + \`is_unmapped_internal_tool_name\` + per-call denial counter |
| 2 — \`.atomic_*.tmp\` literal scatter | \`fs_compat.ml\` | **this PR** |
| 3 — \`ensure_recovered_dir\` reinvents \`mkdir_p\` | \`fs_compat.ml\` | **this PR** |
| 4 — double \`readdir\` of \`base_path\` | \`fs_compat.ml\` | **this PR** |
| 5 — move sweep to background fiber | architectural | **deferred** — boot critical-path behaviour change, deserves explicit ordering review in its own PR |

## Finding 2: literal scatter

Writer (line ~180) and reader (line ~217) carried duplicate \`.atomic_\` + \`.tmp\` string literals.  Accidental rename on either side would silently break the orphan sweep.

\`\`\`ocaml
let atomic_tmp_prefix = ".atomic_"
let atomic_tmp_suffix = ".tmp"
\`\`\`

Both [save_file_atomic] (writer) and [is_atomic_orphan_name] (reader) now reference these constants.

## Finding 3: \`mkdir_p_unix\` reuse

The orphan sweep's \`ensure_recovered_dir\` had its own \`if not file_exists then Unix.mkdir\` helper.  \`mkdir_p_unix\` already exists in the same module (recursive, idempotent, EEXIST-tolerant) — use it.

Behaviour identical for the leaf case (\`recovered_dir\` directly under \`base_path\`) but stays correct if a future caller passes a deeper \`recovered_subdir\`.

## Finding 4: single readdir

Pre-fix, [cleanup_atomic_orphans] called \`Sys.readdir base_path\` twice on every boot — once via \`scan_dir base_path\` and once in the recursion loop.  Read once into \`base_entries\`, run both passes over the cached array.  Subdirectories still call \`readdir\` once each (one read per directory, not one per pass).

## Test plan

- [x] \`dune exec test/test_fs_atomic_orphan_sweep_10130.exe\` — 8/8 pass.  Refactoring preserves behaviour: name matcher, zero-byte deletion, non-zero preservation, subdir scan, non-orphan untouched, idempotent, mixed batch, \`.recovered/\` skip-on-rescan.

## Why finding 5 is deferred

Moving the sweep to a background fiber is a real change to boot ordering — \`Bootstrap completed in Xs\` would log earlier but operators won't see the sweep counter on first \`/metrics\` scrape until the fiber finishes.  Worth its own PR with rollback runbook and ordering review.

## Why finding 1 is skipped

PR #10188 \`fix(auth): allow keeper runtime tools in strict mode\` already landed a strictly better fix:

- Catalog-driven (\`Tool_catalog.is_on_surface Tool_catalog.Keeper_internal\`) — keeper_* prefix alone no longer crosses the gate, only catalog-owned ones do
- Unified predicate \`is_unmapped_internal_tool_name\`
- New \`masc_auth_strict_unknown_tool_denials_total\` counter for explicit denials

The data-driven prefix list suggested in #10205 would be a regression compared to that.

## Refs

- #10131 (orphan sweep landing PR)
- #10205 (post-merge review)
- #10188 (independent auth fix that supersedes #10205 finding 1)

## Do-not-merge

### DIFF
--- a/lib/fs_compat/fs_compat.ml
+++ b/lib/fs_compat/fs_compat.ml
@@ -174,9 +174,19 @@ let fsync_path path =
            is still durable to the extent the underlying FS offers. *)
         ())
 
+(* #10205: SSOT for the [.atomic_*.tmp] naming used by
+   [save_file_atomic] (writer) and [is_atomic_orphan_name]
+   (reader, via the boot-time sweep in [cleanup_atomic_orphans]).
+   Without these constants the writer at line ~180 and the
+   reader at line ~217 carried duplicate string literals that
+   had to agree by hand.  See instructions/software-development.md
+   anti-pattern #1 (hardcoded scatter). *)
+let atomic_tmp_prefix = ".atomic_"
+let atomic_tmp_suffix = ".tmp"
+
 let save_file_atomic (path : string) (content : string) : (unit, string) result =
   let dir = Filename.dirname path in
-  let tmp = Filename.temp_file ~temp_dir:dir ".atomic_" ".tmp" in
+  let tmp = Filename.temp_file ~temp_dir:dir atomic_tmp_prefix atomic_tmp_suffix in
   try
     save_file tmp content;
     fsync_path tmp;
@@ -211,13 +221,12 @@ let save_file_atomic (path : string) (content : string) : (unit, string) result 
    - [preserved]: number of non-zero orphans moved to the
      recovered directory. *)
 let is_atomic_orphan_name name =
-  let prefix = ".atomic_" and suffix = ".tmp" in
   let n = String.length name in
-  let p = String.length prefix in
-  let s = String.length suffix in
+  let p = String.length atomic_tmp_prefix in
+  let s = String.length atomic_tmp_suffix in
   n >= p + s
-  && String.sub name 0 p = prefix
-  && String.sub name (n - s) s = suffix
+  && String.sub name 0 p = atomic_tmp_prefix
+  && String.sub name (n - s) s = atomic_tmp_suffix
 
 let cleanup_atomic_orphans
     ~(base_path : string)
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #10220. Both PRs implement the same #10205 fs_compat refactor (atomic-tmp SSOT + mkdir_p_unix reuse + single readdir). #10220 is leaner (+49/-30 vs +54/-32) and ships a more precise security analysis comment ('handles EEXIST precisely instead of swallowing every Unix_error'). Closing this duplicate.

jeong-sik:
Reviewed #10218 against #10205 cleanup goals and the duplicate #10220 branch.

No blocking findings in this version. The refactor keeps the atomic tmp filename contract in one place, reuses `mkdir_p_unix` for the recovered directory, and reads `base_path` once before scanning both direct orphans and subdirectories. Error handling remains conservative: missing/unreadable directories return the accumulated counters, and per-subdir `Sys.readdir` failures collapse to an empty scan.

Status notes:
- This is the cleaner candidate versus #10220 right now: draft, mergeState CLEAN, CI green including Build/Test, Lint, TLA, and CI Gate.
- The `do-not-merge` label is still present, so I am treating it as an explicit hold rather than merge-ready.
- `git diff --check origin/main...HEAD` passed.

PR #10221, opened 2026-04-25
fix(keeper): centralize tool search alias metadata

jeong-sik:
## Summary
- Move keeper tool-search Korean BM25 aliases into Keeper_agent_tool_surface beside the production Tool_index entry builder.
- Remove the duplicated alias table from keeper_agent_run.ml and the copied alias/group reconstruction from test_keeper_retrieval_quality.ml.
- Add a guard that fails when alias entries point at tools outside the actual keeper universe; this caught and removed stale aliases for retired/unavailable tools.

## Verification
- rg -n "korean_keywords|korean_kw_tbl|String\.starts_with ~prefix:\"keeper_board_|String\.starts_with ~prefix:\"masc_" test/test_keeper_retrieval_quality.ml lib/keeper/keeper_agent_run.ml (no matches)
- env DUNE_BUILD_DIR=_build_verify_alias_ssot DUNE_JOBS=2 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_retrieval_quality.exe
- ./_build_verify_alias_ssot/default/test/test_keeper_retrieval_quality.exe (29 tests)

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -397,105 +397,9 @@ let run_turn
      TODO(OAS): Add Tool_selector.select_with_index that accepts a
      pre-built Tool_index.t to support aliases, groups, and index reuse.
      When that lands, this code can drop the description augmentation. *)
-  (* Korean keyword map for bilingual BM25 matching.
-     Tool descriptions are English; Korean users issue Korean queries.
-     Appending Korean keywords to descriptions gives BM25 term overlap
-     across languages.
-     Keys must match actual tool names from keeper_tools. *)
-  let korean_keywords =
-    [ "keeper_board_post", "게시판 글 작성 올리기 포스트"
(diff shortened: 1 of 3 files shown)


jeong-sik:
Reviewed #10221 from the keeper autonomy/tool-search SSOT angle.

No blocking code-path findings. The important part holds: production `keeper_tool_search` and the retrieval-quality tests now both build entries through `Keeper_agent_tool_surface.tool_index_entry`, so aliases/group metadata are no longer copied in the run loop and tests.

I pushed one cleanup commit: `18416f3fdb docs: align keeper tool alias comments`. The previous comment still described the old description-augmentation path and claimed group co-retrieval was lost, which was misleading after this PR centralized aliases/groups in the production Tool_index entry builder.

Checks:
- `git diff --check`: passed.
- CI Build/Test and TLA were still running on the latest view; local focused Dune is currently contended on the shared lock, so I left executable validation to CI.

PR #10223, opened 2026-04-25
fix(dated-jsonl): drop partial tail chunk prefix

jeong-sik:
Summary:
- Drop the raw first row returned from a tail chunk when the read started mid-file, because it can be a partial JSONL row.
- Preserve the first real data row when the partial prefix is blank/whitespace before filtering non-empty rows.
- Add regression tests for both partial data prefixes and blank partial prefixes.

Validation:
- git diff --check
- Not run: scripts/dune-local.sh build test/test_dated_jsonl.exe (blocked behind /tmp/me-dune-local.lock held by another full dune runtest)

### DIFF
--- a/lib/dated_jsonl/dated_jsonl.ml
+++ b/lib/dated_jsonl/dated_jsonl.ml
@@ -143,9 +143,15 @@ let load_tail_lines path ~max_lines =
           Bytes.blit c 0 combined off len;
           off + len
         ) 0 !chunks in
-        let all_lines =
+        let raw_lines =
           Bytes.to_string combined
           |> String.split_on_char '\n'
+        in
+        let raw_lines =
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed and patched one blocker directly.

What I changed on the branch:
- The partial-prefix drop now happens on raw split lines before blank filtering, so a chunk that starts inside a blank run no longer discards the first real JSONL row.
- Added a regression for the blank-prefix case in addition to the original non-empty partial-prefix case.
- Restored the PR to draft.

Local validation:
- git diff --check: passed
- Focused Dune build was attempted for test/test_dated_jsonl.exe, but it waited on the shared /tmp/me-dune-local.lock; I stopped only my waiting process and left the long-running lock holder untouched. CI is the verification surface for the executable run here.

Issue #9880, opened 2026-04-24
[governance] judgments are decorative — 100% null recommended_action, 95.7% confidence≥0.95, 93% daily volume drop, status stuck at 'active'


jeong-sik:
## Finding

`~/me/.masc/governance/judgments/2026-04/*.jsonl` is the governance-judge keeper's ledger. Examined across yesterday and today (1,191 records total), every observable field that should carry meaning is either degenerate, null, or stuck. Nothing downstream can act on these records, and nothing upstream is actually making a decision.

### 1. Daily volume collapsed 93%
```
2026-04-23.jsonl: 1,114 judgments
2026-04-24.jsonl:    77 judgments (day in progress, ~7h logged)
```
Likely connected to the `refresh_once: compute_judgments failed: Timeout 60.0s` warnings (106× today, same file) and the `masc_oas_bridge: OAS execution timed out after 60.0s` pair (#9629). If the bridge is timing out, the judgment writer drops records silently.

### 2. `recommended_action = null` 100% today, 98.6% yesterday
```
2026-04-23: total=1114 null_ra=1098 (98.6%)
2026-04-24: total=77   null_ra=77  (100.0%)
```
The "judgment" record exists, but the actionable output field is null. 16 yesterday had a non-null value; all 16 used the nested field name `action_kind`, not `kind` (schema drift within the same field).

### 3. Confidence distribution is degenerate
```
=1.0:          756 (63.5%)
0.95–0.999:    383 (32.2%)
0.80–0.949:     41 (3.4%)
0.50–0.799:      9 (0.8%)
<0.50:           2 (0.2%)
total:       1,191
```
**63.5% of judgments are exactly confidence=1.0**; **95.7% are ≥0.95**. With 1,191 samples, this is not calibration — it is a degenerate output. Either the LLM prompt template instructs high confidence by default, or the value is filled by a constant and the LLM free-text isn't parsed.

### 4. `model_used = ""` on 17% of yesterday's records
```
<empty>: 193
gpt-5.3-codex-spark: 477
gpt-5.3-codex: 100
gpt-5.4: 103
kimi-for-coding: 101 (note: these 101 appear despite #9868 — costs.jsonl records 0 tokens for all kimi rows)
...
```
193 records don't record which model produced them. Attribution is 83% complete.

### 5. `status = "active"` 100% (77/77 today, 1114/1114 yesterday)
[...]

Issue #9734, opened 2026-04-23
[Dashboard] operator_snapshot refresh timeout: 30-36s, governance compute_judgments stalled


jeong-sik:
## Symptoms
Dashboard operator_snapshot endpoint consistently times out, causing cascading governance failures:

1. `operator_snapshot profile` takes 12-36s (target: <2s)
2. `refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s`
3. `masc_oas_bridge: OAS execution timed out after 60.0s`

Log samples:
```
[Dashboard] [WARN] [operator_snapshot profile] total=36.5s snapshot=36.5s
[Dashboard] [WARN] operator_snapshot refresh failed (1 consecutive, next in 30s, 30.4s): Failure(\"timeout\")
[Misc] [WARN] masc_oas_bridge: OAS execution timed out after 60.0s
[Governance] [WARN] refresh_once: compute_judgments failed: Timeout: Execution timed out after 60.0s
```

## Impact
- Dashboard shows stale keeper/task state
- Governance decisions are skipped (timeout → default approval)
- OAS bridge resources consumed by hung governance calls

## Reproduction
Continuous under current fleet load (6 keepers). Frequency: ~every 60-120s.

## Root Cause Hypothesis
`compute_judgments` calls OAS with a 60s budget, but the underlying cascade or provider is already saturated by keeper LLM requests. Governance and keepers share the same OAS provider pool, causing head-of-line blocking.

## Related
- Keeper transient network errors show 573s timeouts on same cascade (big_three, tool_use_strict)
- Ollama API also timing out (`curl -sS --max-time 10 http://127.0.0.1:11434/v1/models`)

## Suggested Fix
1. Separate governance OAS budget/pool from keeper OAS pool
2. Add governance judgment caching (ttl=30s) to avoid recomputation
3. Reduce operator_snapshot frequency or make it async push-based

jeong-sik:
Opened draft PR #9750 for the governance side of this issue.

What changed:
- Dashboard_governance_judge.refresh_once now skips recomputing via OAS while the last successful judge result is still within TTL.
- Fresh empty judgment results are treated as a valid cached result, so zero-action cycles also avoid repeated OAS calls.
- Added regression coverage that refresh_once does not call build_facts/OAS when a fresh cached result exists.

Local verification:
- git diff --check
- dune build --root . test/test_dashboard_governance.exe
- ./_build/default/test/test_dashboard_governance.exe

[...]

Issue #9520, opened 2026-04-22
[Meta] 텔레메트리 근본 차단: 100% 계측 강제 + missing metric 알림


jeong-sik:
23개 이슈(root-cause:TEL) 반복 패턴: tok/s '-', telemetry drop, TLA+ silent gate, partial OTel. 대표: #8968 #9167 #8885. 접근: 모든 critical path에 Prometheus metric 강제, missing field는 '-' 대신 null + alert, dashboard에서 누락 시각화. 검증: CI telemetry coverage gate, 3개월 후 재발 빈도 측정.

jeong-sik:
CI gate 구현 완료: `scripts/ci/check-telemetry-coverage.sh`

신규 파일에 telemetry reference 부재 WARN 출력.
action handler 누락 탐지는 단순 휴리스틱이며, keeper core 모듈 대상으로 강화 필요.

PR: #9551

jeong-sik:
## Tracked Instances (closed)

| Issue | Symptom | Resolution |
|-------|---------|------------|
| #9502 | keeper detail and timeline coerce unknown usage to zero | Metric parsing hardened |
| #9454 | Keeper: separate cumulative resumed OAS turns from per-call budget logs | Causal linkage added |

Missing metric alarm infrastructure is the remaining gap for this meta issue.

jeong-sik:
## /loop audit consolidation — telemetry gap 가족

audit loop에서 **관측 경로 자체가 없는 이슈** backlink:

| Issue | 누락 계측 |
|-------|-----------|
| #9868 | `costs.jsonl` cost_usd=0.0 100% — `response.usage.cost_usd=None` fallback시 pricing table lookup 미수행, OAS 경계 그대로 MASC에 기록 |
| #9876 | Hebbian synapse consolidation loop 미가동, `last_consolidation=0.0` 상태로 37 synapse 축적 — health metric 없음 |
| #9880 | governance judgment volume 1114→77 (23→24) 급락, recommended_action null → 소비자 없음. 판단 생산·소비 양단 counter 없음 |
| #9886 | board_votes orphan rate 100% — post↔vote correlation sanity check 없음 |
| #9893 | render fallback 발생해도 trajectory 스키마에 `system_prompt/messages` 없어 **검증 경로 부재** |
| #9896 | split-brain `~/.masc/` vs `~/me/.masc/` — HOME fallback silent, dual-path write detector 없음 |

[...]

PR #10228, opened 2026-04-25
obs(governance): empty response.model fallback + counter (#9880 facet 4)

jeong-sik:
## Summary

#9880 reports governance judgment ledger as decorative across five simultaneous failure signals.  This PR addresses **facet 4** only.

| facet | scope | status |
|---|---|---|
| 1 — \`recommended_action\` 100% null | JSON parser / schema drift | not in this PR |
| 2 — confidence ≥0.95 95.7% degenerate | LLM prompt / schema removal | not in this PR |
| 3 — status always \`active\` | needs downstream consumer pattern | not in this PR |
| **4 — 17% empty \`model_used\`** | **this PR** | observability + sentinel fallback |
| 5 — 93% volume drop | OAS bridge 60s timeout (#9629); covered by #10094 SSOT | already addressed |

## Facet 4 root

OAS transports occasionally return \`response.model = ""\` — the Kimi/Codex CLI silent failure path and \`CompletionContractViolation\` retry-exhausted synthetic responses both have this shape.  Governance writer wrote the empty string straight into the JSONL.  17% of yesterday's records lost attribution.

## Fix (mirrors keeper-side #10083)

\`\`\`ocaml
let resolved_model =
  let raw = response.model in
  if String.trim raw <> "" then raw
  else begin
    let canonical = (* try telemetry.canonical_model_id *) in
    let resolved, source =
      if canonical <> "" then canonical, "telemetry_resolved"
      else "unknown_provider", "unknown_sentinel"
    in
    Prometheus.inc_counter
      governance_response_model_empty_metric
      ~labels:[ ("source", source) ] ();
    Log.Governance.warn
      "compute_judgments: response.model empty → fallback=%s resolved=%s (#9880)"
      source resolved;
    resolved
  end
in
\`\`\`

- Layered fallback: \`raw → telemetry.canonical_model_id → unknown_provider sentinel\`
- Sentinel \`unknown_provider\` matches keeper-side string so PromQL can union-aggregate across both callers
- Per-cycle counter \`masc_governance_response_model_empty_total{source}\` with \`telemetry_resolved\` vs \`unknown_sentinel\`

## Why a separate metric from keeper-side

Keeper-side counter \`masc_after_turn_response_model_empty_total\` is labelled by \`{keeper, source}\`.  Governance has no \`keeper\` semantics — it's a single judge daemon, not a fleet of identities.  Separate metric keeps the label shapes honest while sharing the sentinel string for cross-caller aggregation.

## Test plan

- [x] \`dune exec test/test_governance_response_model_empty_9880.exe\` — 5/5 pass:
  - metric registered at module load
  - \`telemetry_resolved\` branch increments
  - \`unknown_sentinel\` branch increments
  - distinct sources land on distinct rows
  - metric name + \`source=\` label key appear in \`/metrics\` text export

- [ ] Deploy: confirm the counter advances at the production 17% rate, and that the dominant \`source\` label tells operators whether the fix path is the OAS telemetry envelope or the bare sentinel.

## Refs

- #10083 / PR #10090 (keeper-side, same pattern, separate caller)
- #9520 (silent-failure cluster — empty model_used with no alert is exactly this theme)
- #9629, #9734 (OAS bridge 60s timeout — facet 5 root cause, already addressed by #10094)

## Do-not-merge

### DIFF
--- a/lib/dashboard/dashboard_governance_judge.ml
+++ b/lib/dashboard/dashboard_governance_judge.ml
@@ -50,6 +50,23 @@ let () =
        [source=telemetry_resolved | unknown_sentinel]."
     ()
 
+type governance_model_source =
+  | Response_model
+  | Telemetry_resolved
+  | Unknown_sentinel
+
+let governance_model_source_to_string = function
+  | Response_model -> "response_model"
+  | Telemetry_resolved -> "telemetry_resolved"
+  | Unknown_sentinel -> "unknown_sentinel"
+
+let resolve_governance_model_used ~raw_model ~canonical_model_id =
+  if String.trim raw_model <> "" then raw_model, Response_model
+  else
+    match canonical_model_id with
+    | Some id when String.trim id <> "" -> String.trim id, Telemetry_resolved
+    | _ -> "unknown_provider", Unknown_sentinel
+
 let governance_dir base_path =
   Filename.concat
     (Coord_utils.masc_dir_from_base_path ~base_path)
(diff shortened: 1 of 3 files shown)


jeong-sik:
Review follow-up pushed: `5feddde6da`.

I found the previous coverage was too shallow for this PR: it registered/incremented the metric, but did not prove that the governance `model_used` fallback itself resolves empty `response.model` values.

Changes:
- extracted the governance model fallback into `resolve_governance_model_used`
- kept the `compute_judgments` counter/log behavior on fallback paths only
- added resolver tests for raw model, telemetry canonical fallback, missing telemetry sentinel, and empty canonical sentinel

Verification:
- `git diff --check`
- `scripts/dune-local.sh build test/test_governance_response_model_empty_9880.exe`
- `./_build/default/test/test_governance_response_model_empty_9880.exe` (9 tests)

PR #10229, opened 2026-04-25
[codex] Fix MASC bearer login auth path

jeong-sik:
## Summary

- add `masc-mcp login` to mint local bearer tokens, persist the matching raw token file, and print dashboard/Codex MCP exports
- extend `doctor auth` with a `codex_mcp` section that verifies `MASC_MCP_TOKEN` against the live auth store and makes clear that `codex mcp login` is not supported for MASC bearer auth
- update the local auth runbook and doctor architecture docs with the Codex MCP bearer-token path

## Why

Codex was nudging operators toward an OAuth-style `codex mcp login ...` flow, but `masc-mcp` is a bearer-token MCP server. The practical fix is to expose a first-class local bearer minting path and make the doctor report the real auth model and env var status.

## Validation

- `scripts/dune-local.sh build test/test_auth_doctor.exe test/test_auth_login.exe bin/main_eio.exe`
- `./_build/default/test/test_auth_doctor.exe`
- `./_build/default/test/test_auth_login.exe`
- `./_build/default/bin/main_eio.exe login --help`

### DIFF
--- a/bin/main_eio.ml
+++ b/bin/main_eio.ml
@@ -30,6 +30,7 @@ module Dashboard_mission_briefing = Masc_mcp.Dashboard_mission_briefing
 module Build_identity = Masc_mcp.Build_identity
 module Config_doctor = Masc_mcp.Config_doctor
 module Auth_doctor = Masc_mcp.Auth_doctor
+module Auth_login = Masc_mcp.Auth_login
 module Graphql_api = Masc_mcp.Graphql_api
 module Types = Types
 module Tempo = Masc_mcp.Tempo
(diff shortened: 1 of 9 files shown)


jeong-sik:
Reviewed from the auth/operator path angle.

No blocker found. This PR adds the missing local bearer-token path (`masc-mcp login`) and makes `doctor auth` explain the real Codex MCP auth model instead of nudging users toward OAuth-style `codex mcp login`. I also checked it against #10230: this PR mints/diagnoses the bearer token, while #10230 repairs the Codex MCP config shape, so the overlap is complementary rather than redundant.

Local verification:
- `scripts/dune-local.sh build test/test_auth_doctor.exe test/test_auth_login.exe bin/main_eio.exe`
- `./_build/default/test/test_auth_doctor.exe` ✅ 4 tests
- `./_build/default/test/test_auth_login.exe` ✅ 1 test
- `./_build/default/bin/main_eio.exe login --help` ✅ command is registered

Residual note: the help command still emits existing startup logs before help text, but that is pre-existing CLI/module init behavior and not a blocker for this PR.

PR #10230, opened 2026-04-25
fix: preserve verification tasks during claim_next

jeong-sik:
Summary
- keep AwaitingVerification tasks out of claim_next auto-release so verifier approve/reject remains the only way out of the verification FSM edge
- align startup Codex MCP config repair with the env-var SSOT: Accept + X-MASC-Agent headers and bearer_token_env_var = MASC_MCP_TOKEN, without rewriting literal Authorization headers
- update regression coverage for pending verification claim_next behavior and Codex config sync repair

Root cause
- claim_next treated AwaitingVerification as a previous owned task and auto-released it to Todo before claiming new work, which could reopen a verification-gated task while an active verification request still existed.
- startup self-heal could flip ~/.codex/config.toml from the mcp-sync env-var shape back to a literal Authorization header, causing Codex-side auth drift and misleading login warnings.

Verification
- MASC_DUNE_THROTTLE=0 DUNE_JOBS=1 scripts/dune-local.sh build test/test_verification_fsm.exe test/test_server_runtime_bootstrap.exe
- ./_build/default/test/test_verification_fsm.exe
- ./_build/default/test/test_server_runtime_bootstrap.exe test bootstrap 44,45 --color=never
- codex mcp get masc
- codex mcp list

### DIFF
--- a/lib/coord/coord_task_schedule.ml
+++ b/lib/coord/coord_task_schedule.ml
@@ -337,14 +337,14 @@ let claim_next_r
 
       (* BUG-004: Detect and auto-release previous claim to prevent orphaned tasks.
          If this agent already holds a Claimed or InProgress task, release it
-         back to Todo before proceeding. This prevents "claimed but orphaned"
-         tasks that permanently block the backlog. *)
+         back to Todo before proceeding. AwaitingVerification is deliberately
+         excluded: releasing it would drop the verification FSM edge and reopen
+         work before a verifier approve/reject decision. *)
       let previous_claim = List.find_opt (fun (t : Types.task) ->
         match t.task_status with
-        | Claimed { assignee; _ } | InProgress { assignee; _ }
-        | AwaitingVerification { assignee; _ } ->
+        | Claimed { assignee; _ } | InProgress { assignee; _ } ->
             String.equal assignee agent_name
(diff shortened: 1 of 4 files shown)


jeong-sik:
Reviewed from two angles: verification FSM invariants and Codex MCP config self-heal.

No blocker found. The important invariant is preserved: `AwaitingVerification` is no longer auto-released by `claim_next`, so verifier approve/reject remains the only transition out of that edge. The added regression now checks both the new claim and that the original pending task stays `awaiting_verification`.

The Codex MCP config repair part also matches the current local config shape: `bearer_token_env_var = MASC_MCP_TOKEN` plus non-auth `Accept`/`X-MASC-Agent` headers, rather than rewriting a literal Authorization header.

Local verification:
- `scripts/dune-local.sh build test/test_verification_fsm.exe test/test_server_runtime_bootstrap.exe`
- `./_build/default/test/test_verification_fsm.exe` ✅ 14 tests
- `./_build/default/test/test_server_runtime_bootstrap.exe test bootstrap 44,45 --color=never` ✅ 2 focused tests
- `codex mcp get masc && codex mcp list` ✅ local config shows `bearer_token_env_var: MASC_MCP_TOKEN`

State: draft; CI still running on GitHub.

Issue #9766, opened 2026-04-23
dashboard_execution slow render: 59.8s for 9 keepers


jeong-sik:
## Problem
Dashboard execution refresh is extremely slow:
```
[dashboard_execution] slow render: total=59804ms (keepers=9)
```

## Root Cause
TBD. Likely one of:
- Synchronous I/O per keeper (meta read, registry scan, git ops).
- Unbounded list accumulation in snapshot builder.
- Missing caching on hot paths.

## Impact
- Dashboard UI is unusable during refresh.
- HTTP requests timeout before response completes.
- Keeps SSE connections open for 60s+.

## Proposed Fix
1. Profile the render pipeline to identify the bottleneck.
2. Add async/parallel fetching for keeper states.
3. Introduce snapshot caching with incremental updates instead of full rebuild.

## Files
- `lib/dashboard/dashboard_execution.ml` or related rendering module.

jeong-sik:
## Root cause attribution — agent sub-op = 99.6% of total

`#10710` 은 enrich N+1 (per-keeper 367-2363ms × 14) 을 dominant 로 보고했지만, 다른 root cause family 도 보임.

### Latest outlier (2026-04-27T00:14:27Z, 17.3s)

```
[dashboard_execution] slow render: total=18505ms (keepers=14)
  snapshot=17302ms operations=0ms enrich=183ms (per_keeper=13ms) data_load=35ms assemble=985ms
```

- `enrich` 183ms (정상)
- `snapshot` 17302ms = **전체의 93.5%**

### Per-keeper sub-op breakdown (같은 시점)

```
[keepers_json:taskmaster] sub-op: meta=93ms  agent=17137ms ka=0ms ... total=17230ms
[keepers_json:verifier]   sub-op: meta=100ms agent=17130ms ka=0ms ... total=17230ms
[keepers_json:qa-king]    sub-op: meta=327ms agent=0ms     ka=0ms ... total=327ms
[keepers_json:janitor]    sub-op: meta=135ms agent=200ms   ka=0ms ... total=335ms
... (나머지 11 keepers 모두 normal: total < 350ms)
```

**관찰**:
1. **두 keeper (taskmaster, verifier) 가 동시에 정확히 ~17137ms `agent` sub-op** — 다른 12 keeper 는 정상.
2. 같은 `[keepers_json:taskmaster]` 가 다른 시점엔 `agent=38ms` — **transient blocking**.

[...]

PR #10241, opened 2026-04-25
fix(worktree-live-context): per-repo single-flight gate for git status capture (#9632)

jeong-sik:
## Goal

Stop the thundering-herd that turns `git status --porcelain` into a 15s timeout across the keeper fleet (#9632).

## Root cause

`Worktree_live_context.current_status_lines` had a 5s TTL cache but **no single-flight gate**. When the cache expired, 16 keepers + dashboard `mission_profile` + governance compute_judgments all hit it concurrently and each fork+exec'd git in parallel. On a Second Brain root with 64 worktrees and concurrent indexers (gitstatusd, IDE), the I/O contention pushed individual invocations past the 15s budget — even though the working tree only had 13 lines of changes.

```
$ time git -C /Users/dancer/me status --porcelain | wc -l
13
git status --porcelain  0.19s user 2.03s system 21% cpu  10.147 total
```

10s baseline + N concurrent forks = repeated `Process_eio Timeout after 15s` warnings, stale operator snapshots, and `Coord.get_agents_raw` cancellations.

## Change

- `lib/worktree_live_context.ml`: per-repo `Stdlib.Mutex` registry. On cache miss the caller acquires the per-repo gate, double-checks the cache (in case another caller refreshed it while waiting), runs the git capture, stores the result, and releases. Followers reuse the freshly-stored cache line and never spawn git.
- `lib/worktree_live_context.mli`: expose `clear_single_flight_for_tests` next to existing `clear_status_cache_for_tests`.
- `test/test_worktree_status_single_flight_9632.ml`: pin the contract.

## Why a per-repo gate (not global)

A global mutex would serialise unrelated repos — bad for hosts with multiple monorepos. Distinct `repo_root` values get their own `Stdlib.Mutex.t`; concurrent captures across repos remain parallel.

## Pattern

Standard cache-stampede protection:

| System | Mechanism |
|--------|-----------|
| Go `singleflight.Group` | per-key in-flight call collapse |
| Caffeine `loadingCache.get(K, mappingFunction)` | per-key compute lock + double-check |
| Linux page cache `stable_page_writes` | per-page write lock |

The masc implementation is the simplest variant (Mutex + double-check) because the value is small, immutable, and short-lived.

## Tests (local, all green)

`test_worktree_status_single_flight_9632`:
- `concurrent misses same repo collapse to one` — 12 `Domain.spawn` callers on the same `repo_root` stall a 50ms hook; assert hook called **exactly 1 time** (was 12 before).
- `distinct repos do not serialise` — 8 callers split across `/tmp/repo-A` and `/tmp/repo-B`; assert each repo captured exactly once.
- `followers reuse cache` — after priming, 10 follow-up calls within TTL invoke the hook 0 additional times.

`test_worktree_live_context` (existing): 7/7 pass unchanged — single-flight is transparent to single-caller paths.

## Risk / scope

- No public API change beyond the new test hook.
- Cache TTL semantics (`MASC_WORKTREE_STATUS_CACHE_TTL_S`) unchanged.
- Timeout budget (`MASC_WORKTREE_GIT_STATUS_TIMEOUT_SEC`) unchanged.
- Behavior under contention only changes from "all callers race git" to "one caller runs, others wait briefly then read cache".

## Verification (post-merge, on prod root)

After merge, watch `Process_eio Timeout after 15s: 'git'` warnings on the Second Brain runtime. Pre-fix the issue logs three timeouts within 12 minutes (#9632 19:08, 19:17, 19:18). Post-fix, the rate should drop to at most one per cache expiry interval (5s default, so worst case once every cycle when actual git is genuinely slow — not amplified by N callers).

## Out of scope (follow-ups, separate issues)

- Replacing `git status --porcelain` with libgit2 / direct `.git/index` read (lower variance but bigger surface).
- Reducing the dashboard refresh cadence so fewer cycles need this snapshot.
- `gitstatusd` collision avoidance (env-level concern).

Refs #9632. Related observations: #9628 (timeout budget), #9734 (operator snapshot refresh), #9766 (dashboard slow render).

### DIFF
--- a/lib/worktree_live_context.ml
+++ b/lib/worktree_live_context.ml
@@ -119,6 +119,36 @@ let clear_status_cache_for_tests () =
     ~finally:(fun () -> Stdlib.Mutex.unlock status_cache_mu)
     (fun () -> Hashtbl.clear status_cache)
 
+(* Single-flight gates: per-repo Stdlib.Mutex serialising concurrent cache
+   misses so only one [git status] runs per repo at a time.  Without this
+   gate N keepers + dashboard fibers hitting an expired cache simultaneously
+   each fork+exec git, multiplying the I/O contention that was already
+   making git status take 10s+ on a 13-line working tree (#9632).  After
+   the winner refreshes the cache, followers re-check and reuse it.
+   Pattern: Go [singleflight.Group], Caffeine loading cache. *)
+let single_flight_registry : (string, Stdlib.Mutex.t) Hashtbl.t =
+  Hashtbl.create 4
+
+let single_flight_registry_mu = Stdlib.Mutex.create ()
+
+let single_flight_gate repo_root =
+  Stdlib.Mutex.lock single_flight_registry_mu;
+  Fun.protect
+    ~finally:(fun () -> Stdlib.Mutex.unlock single_flight_registry_mu)
+    (fun () ->
+      match Hashtbl.find_opt single_flight_registry repo_root with
+      | Some mu -> mu
+      | None ->
+          let mu = Stdlib.Mutex.create () in
+          Hashtbl.replace single_flight_registry repo_root mu;
+          mu)
+
+let clear_single_flight_for_tests () =
+  Stdlib.Mutex.lock single_flight_registry_mu;
+  Fun.protect
+    ~finally:(fun () -> Stdlib.Mutex.unlock single_flight_registry_mu)
+    (fun () -> Hashtbl.clear single_flight_registry)
+
 let current_status_lines_uncached ~repo_root =
   run_git_capture_lines ~workdir:repo_root
     [ "--no-optional-locks"; "status"; "--porcelain"; "--untracked-files=no" ]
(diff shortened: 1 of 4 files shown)


jeong-sik:
Review follow-up applied. The single-flight implementation looked directionally right, but the distinct-repo test only asserted call counts and did not actually prove repo A/B captures could overlap.\n\nWhat changed:\n- Added a start barrier to the concurrent test helper.\n- Strengthened the distinct-repo test to record active captures and assert max_active=2, which fails under a single global gate.\n- Removed the local absolute path from the test header comment.\n\nVerification:\n- git diff --check\n- scripts/dune-local.sh build test/test_worktree_status_single_flight_9632.exe\n- ./_build/default/test/test_worktree_status_single_flight_9632.exe -> 3 tests passed\n- bash scripts/health_snapshot.sh --skip-build --baseline-ref origin/main --fail-on-lib-regression -> Ratchet: pass

Issue #10247, opened 2026-04-25
Goal upsert can bypass Goal FSM terminal validation


jeong-sik:
## Symptom
Goal upsert can write lifecycle status or phase directly, so dashboard and goal history can show a terminal state that did not pass through the Goal FSM transition path.

## Reproduction condition
Call goal upsert with a terminal-looking status or phase instead of using the goal transition tool or verification finalization.

## Suspected cause
coord_goals.ml validates linked tasks and emits goal events on the transition path, but the upsert path only blocks some awaiting states before forwarding status or phase into Goal_store.upsert_goal.

## Tried
During telemetry terminal-gap audit for PR #10245, traced goal transition path vs upsert path. This is outside that PR scope. Fix should route lifecycle writes through the FSM or reject lifecycle fields on upsert.

PR #10261, opened 2026-04-25
fix(coord-goals): block lifecycle states on goal upsert (#10247)

jeong-sik:
## Goal

Stop \`masc_goal_upsert\` from writing lifecycle phase/status fields that should only be reachable through the goal FSM (\`masc_goal_transition\`). Reported in #10247.

## Root cause

\`handle_goal_upsert\` in \`lib/coord_goals.ml\` previously rejected only two phases: \`Awaiting_verification\` and \`Awaiting_approval\`. Every other lifecycle field passed through into \`Goal_store.upsert_goal\` raw:

- terminal phases: \`Completed\`, \`Dropped\`
- operator-managed phases: \`Blocked\`, \`Paused\`
- terminal statuses: \`Done\`, \`Dropped\`
- operator status: \`Paused\`

That means an agent or operator could call \`masc_goal_upsert\` with \`phase: \"completed\"\` and watch the goal land in a terminal state without crossing the FSM. The audit/telemetry that hangs off the transition path — \`verifier_policy\` enforcement, linked-task validation, FSM event emission, approval bookkeeping — never ran. Dashboards and history then read a row that looks finished but never proved it.

## Fix

Reject the lifecycle states that the FSM owns at the upsert tool surface. The block is by **field × value**, not by intent guesswork — explicit list, exhaustive over the variant types so a future phase/status addition forces the gate to be revisited:

| field   | values rejected                                      |
|---------|------------------------------------------------------|
| phase   | awaiting_verification, awaiting_approval, completed, dropped, blocked, paused |
| status  | done, dropped, paused                                |

\`Executing\` phase and \`Active\` status (the only non-lifecycle pair) still round-trip — legitimate rename / metric / horizon / priority edits via upsert are unchanged.

The gate is on the **tool surface** (\`handle_goal_upsert\`), not on the library function (\`Goal_store.upsert_goal\`). Internal callers (server bootstrap, fixtures) keep their direct access; only untrusted clients reaching through \`masc_goal_upsert\` are restricted.

## Why phase rejection beats status when both violate

\`Goal_store.upsert_goal\` derives status from phase on write (see header comment in \`goal_store.ml\`), so when a caller passes both, phase is the load-bearing field. Stable error ordering matters for clients that pattern-match on the rejection message — pinned by \`test_phase_violation_beats_status\`.

## Tests

\`test_goal_upsert_fsm_bypass_10247\` (6 cases, all green):
- **phase-rejection**: each of the 6 lifecycle phases rejected; error names \`masc_goal_transition\` and the violating phase value.
- **status-rejection**: each of the 3 lifecycle statuses rejected; same shape.
- **happy-path**: \`executing\` phase round-trips; \`active\` status round-trips; absent phase+status round-trips (the most common upsert).
- **ordering**: when both phase and status violate, error names phase first.

\`test_goal_tools\` (11 cases, all green): one pre-existing test (\`test_goal_list_filters_by_phase\`) used \`phase: \"blocked\"\` on upsert as a fixture — exactly the bypass #10247 calls out. Rewired to drive \`masc_goal_transition operator_block\` with an operator principal so the test now documents the correct path.

Sibling tests \`test_goal_store\` (8/8) and \`test_dashboard_goals\` (8/8) unchanged and still green.

## Out of scope

- **Lifecycle writes via \`Goal_store.upsert_goal\` direct library calls** (internal). Locking those down would require a separate audit of every caller and a new constructor function — RFC-shaped change, not a hotfix.
- **\`masc_goal_review\`** lifecycle interactions (#10247 did not flag this; existing \`test_goal_review_updates_status\` keeps its contract).

Refs #10247.

### DIFF
--- a/test/dune
+++ b/test/dune
@@ -690,6 +690,7 @@
         test_eval_calibration
         test_goal_janitor
         test_goal_tools
+        test_goal_upsert_fsm_bypass_10247
         test_goal_store
         test_goal_verification
         test_string_util
@@ -866,6 +867,7 @@
         test_eval_calibration
         test_goal_janitor
         test_goal_tools
+        test_goal_upsert_fsm_bypass_10247
         test_goal_store
         test_goal_verification
         test_string_util
--- a/test/test_goal_upsert_fsm_bypass_10247.ml
+++ b/test/test_goal_upsert_fsm_bypass_10247.ml
@@ -1,0 +1,238 @@
+(** #10247 root cause: [masc_goal_upsert] used to let callers write
+    lifecycle fields directly into the Goal Store.  Goal lifecycle
+    moves must go through the FSM tools so verifier policy, approvals,
+    and audit bookkeeping cannot be bypassed. *)
+
+open Alcotest
+open Masc_mcp
+
+let temp_dir () =
+  let path = Filename.temp_file "goal_fsm_bypass_10247_" "" in
+  Sys.remove path;
+  Unix.mkdir path 0o755;
+  path
+
+let rec rm_rf path =
+  if Sys.file_exists path then
+    if Sys.is_directory path then begin
+      Sys.readdir path
+      |> Array.iter (fun name -> rm_rf (Filename.concat path name));
+      Unix.rmdir path
+    end else
+      Sys.remove path
+
+let with_room f =
+  Eio_main.run @@ fun env ->
+  Fs_compat.set_fs (Eio.Stdenv.fs env);
+  let dir = temp_dir () in
+  Fun.protect ~finally:(fun () -> rm_rf dir) (fun () ->
+      let config = Coord.default_config dir in
+      ignore (Coord.init config ~agent_name:(Some "planner"));
+      f config)
+
+let coord_ctx config : Tool_coord.context =
+  { Tool_coord.config; agent_name = "planner" }
+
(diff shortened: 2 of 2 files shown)


jeong-sik:
Review 대응 업데이트입니다.

- 최신 main에 이미 더 강한 `masc_goal_upsert` lifecycle-field 금지 계약이 들어와 있어, PR을 rebase하면서 부분 허용 구현은 되살리지 않았습니다.
- 대신 #10247 회귀 테스트를 최신 계약에 맞춰 정리했습니다: `phase` 필드 전체, `status` 필드 전체, 기존 blocked/paused goal이 `phase=executing` / `status=active` upsert로 재개되지 않는 케이스를 고정했습니다.
- PR head는 `903363df9c`로 force-with-lease 갱신했고 draft + `do-not-merge` 상태는 유지했습니다.

검증:
- `bash scripts/health_snapshot.sh --skip-build --baseline-ref origin/main --fail-on-lib-regression`
- `scripts/dune-local.sh build test/test_goal_upsert_fsm_bypass_10247.exe`
- `./_build/default/test/test_goal_upsert_fsm_bypass_10247.exe` -> 6 tests passed

현재 GitHub checks는 새 head에서 pending입니다.

PR #10267, opened 2026-04-25
perf(mcp-call-tool): replace regex-based contains_casefold with byte scan

jeong-sik:
## Summary

`Mcp_server_eio_call_tool.contains_casefold` compiled a fresh `Re.t` from the needle on every call:

```ocaml
let contains_casefold haystack needle =
  let haystack = String.lowercase_ascii haystack in
  let needle = String.lowercase_ascii needle in
  Re.execp (Re.str needle |> Re.compile) haystack
```

~20 hot call-sites pass string literals on tool-dispatch paths (`parse_status_from_message`, `is_retryable_error`, auth-required detection). Each call paid:
- 2 `String.lowercase_ascii` allocations
- `Re.compile` of an `str` pattern (DFA construction)
- `Re.execp` over the lowered haystack

Replaced with a byte-wise scan using inline `Char.lowercase_ascii`:

```ocaml
let contains_casefold haystack needle =
  let nlen = String.length needle in
  let hlen = String.length haystack in
  if nlen = 0 then true
  else if nlen > hlen then false
  else
    let rec match_at i j = ... in
    let last = hlen - nlen in
    let rec loop i = ... in
    loop 0
```

No allocation in the loop. Aligns with the peer definition in `Server_routes_http_keeper_stream` which already uses a byte-wise scan.

Same anti-pattern series:
- #10250 / #10252: `Mention` per-call `Re.compile` (merged)
- #10260: `Coord_gc.mentions_open_task` M×N compile (in-flight)
- This PR: `contains_casefold` per-call `Re.compile`

## Test plan
- Behavior preserved: empty-needle → `true`, `nlen > hlen` → `false`, ASCII case fold via `Char.lowercase_ascii` (identical semantics to `String.lowercase_ascii` byte-by-byte).
- No direct unit test exists; semantics-equivalent rewrite of a 4-line helper, structurally verifiable from the diff.

### DIFF
--- a/lib/mcp_server_eio_call_tool.ml
+++ b/lib/mcp_server_eio_call_tool.ml
@@ -28,10 +28,35 @@ let int_of_env_default name ~default ~min_v ~max_v =
       in
       max min_v (min max_v parsed)
 
+(* Substring containment, ASCII case-insensitive.
+
+   Old version compiled a fresh [Re.t] from the needle on every call —
+   ~20 hot call-sites, all string literals, mostly tool-error
+   classification on the per-call path.  The needles are short
+   (< 30 chars) and haystacks are bounded error messages, so a naive
+   byte-wise scan with inline [Char.lowercase_ascii] is strictly
+   cheaper than DFA construction and avoids the two
+   [String.lowercase_ascii] allocations the old form needed. *)
 let contains_casefold haystack needle =
-  let haystack = String.lowercase_ascii haystack in
-  let needle = String.lowercase_ascii needle in
-  Re.execp (Re.str needle |> Re.compile) haystack
+  let nlen = String.length needle in
+  let hlen = String.length haystack in
+  if nlen = 0 then true
+  else if nlen > hlen then false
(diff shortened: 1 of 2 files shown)


jeong-sik:
#10267 review fix

Added a direct regression test for the optimized `contains_casefold` helper in `test/test_mcp_server_eio_call_tool.ml`.

Why:
- The implementation change is small, but it replaces regex behavior with a manual byte scan.
- The PR previously relied on structural equivalence only.
- The new test pins empty-needle, exact match, ASCII case folding, middle substring, numeric literal, too-long needle, and absent substring behavior.

Pushed: `be18fd9ad0`

Local checks:
- `git diff --check` passed
- `bash scripts/health_snapshot.sh --skip-build --baseline-ref origin/main --fail-on-lib-regression` passed

Focused Dune was not started here because the host-level `/tmp/me-dune-local.lock` is currently saturated by other long-running local sessions; PR CI is the live build/test gate for the pushed head.

PR #10274, opened 2026-04-25
fix(cascade): degraded TOML-section fallback for keeper-name validator (#10259)

jeong-sik:
## Goal

Stop one materializer field-whitelist regression in \`cascade.toml\` from cascading into a fleet-wide silent reject of every operator-defined keeper cascade. Reported in #10259.

## Root cause chain (from issue body)

1. \`Cascade_toml_materializer.render_toml_to_yojson\` runs a strict per-field whitelist when materializing \`cascade.toml\` → \`cascade.json\`. One unknown inner field anywhere in the file aborts the whole pass.
2. \`Cascade_config_loader.load_catalog\` therefore returns \`Error _\`.
3. \`Keeper_cascade_profile.catalog_names_result\` propagates that error.
4. The keeper-name validator at \`keeper_types_profile.ml:669\` collapses to the compile-time \`reserved_cascade_names\` list (5 names: \`big_three, local_only, local_recovery, tool_rerank, tool_use_strict\`).
5. Operator-defined cascades like \`ollama_only\` are rejected fleet-wide on every reconcile cycle. The runtime keeps running on a stale cached \`cascade.json\`, so observable behavior is correct — but \`<keeper>.toml\` reconciliation never lands and reconcile WARN spam fills 25%+ of the log tail (8 lines per cycle × 4 affected keepers in the report).

This matches the \`feedback_config-root-scope-discipline\` pattern: a localized parse error one layer down silently kills validation an entire layer up because the two are wired through a single Result.

## Fix

Decouple the validator's accept list from full strict materialization.

**\`lib/cascade/cascade_toml_materializer.ml\`** — new \`toml_section_names_result\`:
- Reads \`cascade.toml\` with the same \`Otoml.Parser.from_string_result\` the materializer already uses.
- Walks only top-level table fields, returns their keys.
- Filters meta-keys (any name starting with \`_\` — \`_schema\`, \`_revision\`, \`_comment_*\`).
- Filters non-table top-level entries (scalars, arrays).
- JSON-only sources return \`Ok []\` (their catalog goes through \`Cascade_config_loader\` directly; that's a different bug class).

**\`lib/keeper/keeper_cascade_profile.ml(.mli)\`** — new \`catalog_names_with_toml_fallback\`:
\`\`\`
Live_catalog                                 (* strict load succeeded *)
Toml_section_fallback { catalog_error }      (* strict failed, lenient pass survived *)
Error _                                       (* both failed *)
\`\`\`
Provenance is part of the return type so the validator can attach a \`[degraded toml-section fallback; live catalog unavailable: <reason>]\` suffix to its rejection message — the WARN log still surfaces the materializer error, just not as a fleet-wide reject.

**\`lib/keeper/keeper_types_profile.ml\`** — validator wires the new helper in. Pre-fix it called \`catalog_names_result\` and collapsed to the reserved list on \`Error\`. Post-fix it calls \`catalog_names_with_toml_fallback\`, accepts any name in \`reserved ∪ catalog\` (live or degraded), and only errors when both sources are unavailable.

## Why decouple at this layer specifically

The strict materializer is the right place to enforce field-by-field schema for *write* (rejecting an unknown field prevents bad TOML from corrupting cascade.json). The validator's accept list is the right place to know *what cascade names exist*. Tying the second to the first means \"can I write this cascade.json?\" gates \"can I read what cascades are configured?\" — wrong layer dependence. Section-name enumeration is strictly weaker than full materialization (only top-level identifiers, no fields), but it's exactly what the validator needs.

This does NOT remove the WARN about the materializer failure. It just stops the WARN from also disabling every keeper.

## Tests (\`test_cascade_toml_section_fallback_10259\`, 7 cases, all green)

| group | case | pin |
|-------|------|-----|
| section-names | basic top-level enumeration | sections returned in declaration order |
| section-names | filters meta-keys, scalars, arrays | \`_schema\`, \`_revision\`, \`_comment_doc\`, top-level \`allowed_environments = [...]\` all dropped |
| section-names | survives strict-unknown inner field | the exact #10259 root-cause class — strict pass would reject, lenient pass succeeds |
| section-names | json-only source yields empty list | non-TOML sources return \`Ok []\` cleanly |
| section-names | malformed toml yields error | TOML parse failure surfaces, doesn't silently empty-list |
| validator-fallback | live path uses \`Live_catalog\` tag | strict catalog wins when available |
| validator-fallback | degraded path uses \`Toml_section_fallback\` tag | catalog_error preserved for WARN log |

Existing sibling tests (no regression):
- \`test_keeper_toml\`: 81/81 green (persona authoring + cascade_name validation paths).
- \`test_cascade_catalog_runtime\`: 13/13 green (catalog loader runtime paths).
- \`test_cascade_toml_materialization\`: pre-existing 2 fails on baseline (env-dependent: requires \`MASC_CASCADE_TOML_PATH\`, set in CI). Unchanged by this PR.

## Out of scope (separate issues / PRs)

The original #10259 lists four options: A (root-cause OTOML diagnosis), B (drift gate), C (validator fallback strengthening), D (per-error WARN dedup). This PR is **C alone**:

- **A** requires reproducing the OTOML \`flat-key vs nested-key\` confusion that turns inline \`strategy = ...\` into \`tool_use_strict_strategy\` somewhere in the strict path. Deeper parser investigation, deserves its own PR.
- **B** is a build/CI shape (a \`cascade-validate\` startup check). Different file scope, different review surface.
- **D** would dedup WARN noise by reusing the existing \`logged_invalid_fallback\` Hashtbl pattern in \`keeper_cascade_profile.ml:131\`. Cosmetic next to the structural fix here.

C alone is enough to make A's root cause stop being silently fleet-fatal: even if A regresses again in the future, the validator stays correct.

Refs #10259. Pattern: \`feedback_config-root-scope-discipline\`.


jeong-sik:
Review 대응 업데이트입니다.

- strict catalog load가 실패한 뒤 TOML section fallback이 `Ok []`를 반환하는 경우를 더 이상 degraded success로 취급하지 않게 했습니다.
- 이제 fallback name list가 비어 있으면 `Error`로 올라가므로 keeper-name validator가 실제 catalog failure를 빈 accept-list 성공처럼 삼키지 않습니다.
- 테스트에 `empty degraded fallback errors` 케이스를 추가했습니다.
- PR head는 `efb132055e`로 갱신했고 draft + `do-not-merge` 상태는 유지했습니다.

검증:
- `git diff --check`
- `bash scripts/health_snapshot.sh --skip-build --baseline-ref origin/main --fail-on-lib-regression`
- `scripts/dune-local.sh build test/test_cascade_toml_section_fallback_10259.exe`
- `./_build/default/test/test_cascade_toml_section_fallback_10259.exe` -> 8 tests passed

현재 GitHub checks는 새 head에서 pending입니다.

PR #10284, opened 2026-04-25
fix(goals): route lifecycle changes through Goal FSM

### LINKED ISSUE #10247: Goal upsert can bypass Goal FSM terminal validation

jeong-sik:
## Summary
- Reject status/phase lifecycle fields on masc_goal_upsert so goal state cannot bypass the FSM path.
- Remove lifecycle status/phase from the upsert tool schema and keep list filtering as the read-side phase surface.
- Seed phase filter tests through Goal_store fixtures and add a regression for rejected lifecycle upserts.

Fixes #10247.

## Verification
- env MASC_DUNE_THROTTLE=0 DUNE_LOCAL_JOBS=2 DUNE_CACHE=disabled DUNE_BUILD_DIR=/tmp/masc-10247-build scripts/dune-local.sh build --cache=disabled test/test_goal_tools.exe test/test_tools_coverage.exe
- /tmp/masc-10247-build/default/test/test_goal_tools.exe
- /tmp/masc-10247-build/default/test/test_tools_coverage.exe

### DIFF
--- a/lib/coord_goals.ml
+++ b/lib/coord_goals.ml
@@ -100,6 +100,12 @@ let parse_optional_goal_phase args field =
            ~expected:"string"
            ~received:(Yojson.Safe.to_string json))
 
+let goal_upsert_lifecycle_error field =
+  error_result_typed ~code:Validation_error
+    (Printf.sprintf
+       "masc_goal_upsert does not accept lifecycle field %s; use masc_goal_transition / masc_goal_verify for goal lifecycle moves"
+       field)
+
 let parse_optional_review_outcome args field =
   match Yojson.Safe.Util.member field args with
   | `Null -> Ok None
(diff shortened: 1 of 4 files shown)


jeong-sik:
#10284 review note

This overlaps #10261 / the same #10247 root cause. I would not merge this as-is without an explicit product decision, because it changes the public tool contract more broadly than #10261:

- #10284 rejects any `phase` or `status` on `masc_goal_upsert` and removes both from the schema.
- #10261 rejects only FSM-owned lifecycle states while keeping `phase=executing`, `status=active`, and no-field upserts valid.
- That preserves metadata/fixture upserts without letting terminal/operator/approval states bypass the Goal FSM.

Recommended path: treat #10261 as the canonical fix. Either close/supersede #10284, or rebase it on #10261 and keep only the schema-surface removal if we intentionally want to remove even active/executing from the tool schema.

PR #10290, opened 2026-04-25
fix(fsm): preserve composite operator truth

jeong-sik:
## Summary

This PR fixes two operator-truth gaps found during the FSM cross-layer audit and adds a cheap CI guard for formal-spec harness drift.

- Keep dashboard composite FSM state fields as open strings so backend-ahead states are shown as-is instead of being coerced to `Stable`, `idle`, or `clean`.
- Emit `keeper_composite_changed` from direct live-turn registry mutations, not only phase transitions, so FSM Hub/fleet consumers know when turn phase, decision, cascade, selected model, or completed-turn fields changed.
- Add `scripts/ci/check-tla-harness-coverage.sh` and wire it into CI so cfg-backed TLA specs cannot silently exist outside `scripts/tla-check.sh` without being recorded as known unchecked debt.
- Record the audit in `docs/audits/fsm-cross-layer-audit-2026-04-25.md` and update the system event/snapshot inventory.

## Operator Impact

Unknown FSM values now remain visible instead of being normalized into safe-looking defaults. A future backend state can show up raw in the dashboard, which is noisy but truthful. Direct turn-substate updates now notify pull observers, reducing stale FSM Hub and fleet matrix cells after keeper turn mutations.

## Verification

- `pnpm vitest run src/api/schemas/keeper-composite.test.ts src/components/fsm-hub-types.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm typecheck`
- `bash scripts/ci/check-tla-harness-coverage.sh`
- `git diff --check HEAD`
- `MASC_DUNE_THROTTLE=0 DUNE_LOCAL_JOBS=1 scripts/dune-local.sh exec test/test_keeper_composite_observer.exe`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,6 +288,7 @@ jobs:
           bash scripts/ci/check-enum-string-safety.sh
           bash scripts/ci/check-masc-oas-boundary.sh
           bash scripts/ci/check-tla-variant-sync.sh
+          bash scripts/ci/check-tla-harness-coverage.sh
           bash scripts/ci/check-telemetry-coverage.sh
           bash scripts/ci/check-determinism-contract.sh
           bash scripts/audit-hardcoding-truth.sh --fail-on-confirmed
--- a/dashboard/src/api/schemas/keeper-composite.test.ts
+++ b/dashboard/src/api/schemas/keeper-composite.test.ts
@@ -41,24 +41,24 @@ describe('parseKeeperCompositeSnapshot', () => {
     }
   })
 
-  it('falls back unknown phase to Stable', () => {
+  it('preserves unknown phase values for operator visibility', () => {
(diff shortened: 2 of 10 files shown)


jeong-sik:
Review follow-up applied on current head:

- Updated `fsm-hub.integration.test.ts` so the integration contract also preserves unknown composite phase values instead of expecting `Stable` fallback.
- Tightened `check-tla-harness-coverage.sh` to match the `scripts/tla-check.sh` directory + file pair, avoiding basename-only false passes for future same-named specs.
- Local checks: `bash scripts/ci/check-tla-harness-coverage.sh`, `git diff --check`, and `pnpm vitest run src/api/schemas/keeper-composite.test.ts src/components/fsm-hub.integration.test.ts src/components/fsm-hub-types.test.ts --no-file-parallelism --maxWorkers=1` passed.

Issue #10297, opened 2026-04-25
keeper_board_post: caller-supplied author bypasses ctx.agent_name (velvet-hammer→analyst drift)


jeong-sik:
## Symptom

A keeper_board_post call with `runtime_contract.keeper_name = velvet-hammer` (and `runtime_contract.agent_name = keeper-velvet-hammer-agent`) produced a board post with `output.author = "analyst"`.  Excerpt:

\`\`\`
"keeper": "velvet-hammer",
"tool": "keeper_board_post",
"runtime_contract.keeper_name": "velvet-hammer",
"runtime_contract.agent_name": "keeper-velvet-hammer-agent",
"output.author": "analyst"
\`\`\`

## Root cause

`lib/tool_inline_dispatch_extra.ml:79-93` `ensure_board_post_author`:

\`\`\`ocaml
let raw_author =
  if existing <> "" && existing <> "anonymous" then existing
  else String.trim agent_name
in
\`\`\`

The dispatcher receives the trusted `~agent_name` from the runtime contract, but uses it **only when args.author is blank**.  Any non-empty caller-supplied `author` bypasses identity verification.  The LLM inside velvet-hammer wrote `"author": "analyst"` in its tool args and the canonicalisation step accepted it without comparing to the trusted name.

## Why it matters

- **Identity spoofing**: any keeper can post under any other principal's name.  Combined with `keeper_board_post` being on the routine allowlist (`lib/keeper/keeper_routine_allowlist.ml:61`), this means autonomous keepers can impersonate each other on the public board.
- **Audit forensics**: every existing post tagged `author = X` only proves "some caller passed X as the author argument", not "principal X authored this".  Dispatch-noise diagnoses currently filter on author and will misattribute.
- **Pattern parallel**: same class as #6527 trilogy + #6623 (`mcp-dispatcher-ctx-agent-name-ssot` feedback memory). MCP dispatcher must reject or rewrite caller-supplied identity fields when they disagree with the runtime contract.

## Fix sketch

In `ensure_board_post_author`:

\`\`\`ocaml
let canonical_from_ctx = Server_utils.board_actor_author_for_write (String.trim agent_name) in
let raw_author =
  match existing with
  | "" | "anonymous" -> String.trim agent_name
  | other when String.equal (Server_utils.board_actor_author_for_write other) canonical_from_ctx ->
      other
  | other ->
      (* preserve the LLM's claim in meta for forensics, but rewrite the
[...]

Issue #6623, opened 2026-04-11
security(auth): handle_auth_create_token / handle_auth_revoke trust args agent_name without admin check — cross-agent token spoof


jeong-sik:
## Summary

`handle_auth_create_token` (`lib/tool_auth.ml:70-112`) and `handle_auth_revoke` (same file, `:137-144`) trust the `agent_name` argument from the MCP caller verbatim, with **zero admin/ownership check**. Any caller with access to `masc_auth_create_token` or `masc_auth_revoke` can therefore:

1. Create a valid credential for **any other agent_name**, obtain the raw token in the response, and use it to impersonate that agent on subsequent requests
2. Revoke **any other agent's** credential, locking them out

`Auth.create_token` itself (`lib/auth.ml:165-189`) has **no role/admin gate** — it just hashes, stores, and returns the raw token. `save_credential` overwrites any existing credential for that `agent_name`.

`handle_auth_refresh` (`lib/tool_auth.ml:114-118`) correctly enforces `target == ctx.agent_name`:

```ocaml
let handle_auth_refresh ctx args =
  let target_agent = target_agent_name ctx args in
  if target_agent <> ctx.agent_name then
    (false, "agent_name must match the authenticated agent for refresh")
  ...
```

This is the intended pattern the sibling handlers are missing.

## Impact (severity: **Critical**)

- **Auth enabled, non-admin caller**: calls `masc_auth_create_token agent_name=<target> role=admin` → receives `target`'s raw token with `admin` role → hot-swaps identity on the next request. Complete bypass of the whole RBAC model, not just `#6527` containment.
- **Auth enabled, same scenario with `revoke`**: calls `masc_auth_revoke agent_name=<target>` → target is locked out until manually re-credentialed.
- **Auth disabled**: no effect (the whole authn layer is inert anyway).

This is structurally the same class of bug that `#6527` iter 7 (#6617) closed for `masc_worktree_create`: an MCP dispatcher trusting the caller's self-declared identity in args instead of using `ctx.agent_name` as the SSOT. The difference is that #6527 iter 7 let you hijack a playground directory, and this one lets you hijack an entire auth identity.

## Call graph

```
MCP tool dispatch
  → Tool_auth.dispatch ctx ~name:"masc_auth_create_token" ~args
    → handle_auth_create_token ctx args
      → target_agent_name ctx args          (* picks args.agent_name, no check *)
      → Auth.create_token config ~agent_name:target ~role
[...]

PR #10305, opened 2026-04-25
fix(dispatcher): enforce ctx identity on board author/voter (#10297)

jeong-sik:
## Goal

Stop a keeper from posting / commenting / voting on the board under another keeper's name. Reported as #10297 (velvet-hammer LLM wrote \`author = \"analyst\"\` and the dispatcher accepted it).

## Root cause

\`lib/tool_inline_dispatch_extra.ml\` had two adjacent gates that both leaked identity:

| function | tools | bug |
|---|---|---|
| \`ensure_board_post_author\` | masc_board_post | only consulted ctx \`agent_name\` when caller's \`author\` was blank; any non-blank value passed unchallenged |
| \`canonicalize_board_actor_field\` | masc_board_comment / masc_board_vote / masc_board_comment_vote | did not receive \`agent_name\` at all |

Combined with \`keeper_board_post\` being on the routine allowlist (\`lib/keeper/keeper_routine_allowlist.ml:61\`), an autonomous keeper could impersonate any principal on the public board, and dispatch-noise diagnoses filtering by \`author\` would misattribute every post.

This is structurally the same as #6527 trilogy + #6623 (memory: \`feedback_mcp-dispatcher-ctx-agent-name-ssot\`). The MCP dispatcher must reject or rewrite caller-supplied identity fields when they disagree with the runtime contract.

## Fix

Unify all four board tools through a new \`enforce_caller_identity ~tool ~field ~agent_name\` gate. 3-branch contract (the SSOT pattern from the memory):

| caller field | action |
|---|---|
| blank / \`anonymous\` | fill from ctx canonical |
| canonical equals ctx canonical | preserve canonical; stash raw surface in \`meta.<field>_raw_agent_name\` when surfaces differ (e.g. \`keeper-velvet-hammer-agent\` vs \`velvet-hammer\`) |
| canonical disagrees with ctx canonical | **rewrite to ctx canonical**, preserve caller's claim in \`meta.<field>_caller_claim\`, increment \`masc_board_actor_identity_spoof_total{tool, field}\` |

\`board_actor_author_for_write\` is the canonicalisation function (already used by HTTP routes, dashboard, etc.) — it strips \`keeper-<name>-agent\` to \`<name>\`, so two surface forms that name the same keeper compare equal.

## Why lenient mode (rewrite + preserve + count) instead of fail-closed

The LLM occasionally supplies a wrong \`author\` under persona / system-prompt confusion (e.g. cross-keeper sandbox bleed). Rejecting the call would break the chain and lose the post entirely, while rewriting preserves the post with correct attribution and surfaces the drift to metrics.

A future tick can flip the mismatch branch to fail-closed once metric data shows the rewrite isn't masking real bugs — the issue body explicitly invites this as a follow-up.

## Cardinality

\`masc_board_actor_identity_spoof_total{tool, field}\` — 4 tools × 2 fields = 8 series at most. Bounded.

## Backward compatibility

- \`ensure_board_post_author\` kept as a thin alias forwarding to the new helper.
- \`canonicalize_board_actor_field\` kept with optional \`?agent_name\`; legacy direct callers (HTTP handlers, fixtures) continue to work without ctx but still get canonical-form normalisation.
- Pre-existing \`meta.author_raw_agent_name\` semantic preserved for the legitimate canonicalisation case.
- No metric removal; no label rename; no behavior change for non-spoofing callers.

## Tests (\`test_board_author_identity_10297\`, 8 cases, all green)

| group | case | pin |
|---|---|---|
| blank-fills-from-ctx | blank author | rewritten to ctx canonical, no \`caller_claim\` recorded |
| blank-fills-from-ctx | \"anonymous\" author | rewritten to ctx canonical |
| matching-canonical | short-name matches ctx | preserves canonical, no spoof signal |
| matching-canonical | full \`agent_name\` surface | canonical in field, raw surface in \`meta.<field>_raw_agent_name\` |
| spoof-rewrite | velvet-hammer cannot post as analyst | exact #10297 reproducer; field rewritten, claim preserved, counter +1 |
| spoof-rewrite | voter spoof also rewritten | parallel for voting tools |
| counter-isolation | (post, author) bump does NOT move (vote, voter) | per-surface rate-alert stays clean |
| empty-ctx | no ctx still canonicalises | legacy HTTP path semantics |

## Out of scope (separate issues)

- **Persona / system-prompt confusion** that drove velvet-hammer's LLM to write \`author=analyst\` in the first place. The dispatcher boundary is the right enforcement point regardless of what the LLM thinks its name is.
- **Strict fail-closed mode**. Issue body sketches it as an alternative — easy follow-up if the lenient rewrite proves insufficient (flip one branch).
- **Metrics directory drift** mentioned in the issue's cross-reference comment (43 directories for 13 keepers, 3.3x variant explosion). That's a separate \"all writers using caller-supplied identity as directory key\" issue and would need its own pass.

Refs #10297. Pattern: \`feedback_mcp-dispatcher-ctx-agent-name-ssot\`.


jeong-sik:
## Duplicate detection

#10307 already merged (cd57605f40) and addresses #10297 with the same root cause analysis on `lib/tool_inline_dispatch_extra.ml`. The DIRTY state stems from the lib/ overlap.

### Distinguishing value
This PR adds `test/test_board_author_identity_10297.ml` — a dedicated regression test that #10307's `test_tool_board_coverage.ml` does not include.

### Suggestions
1. **Close as superseded** if the lib/ fix in #10307 is sufficient
2. **Rebase keeping only the test file** (drop lib/ changes) to land the dedicated regression test

Posting from /loop saturation cleanup tick — no unilateral close.

PR #10309, opened 2026-04-25
fix(task): make verification verdict persistence atomic

jeong-sik:
## Summary
- persist approve/reject verification verdicts before committing the task FSM transition
- keep tool and dashboard verdict paths on the same record -> transition -> notify order
- block dashboard verification_id mismatches before moving AwaitingVerification tasks
- add regression coverage for approve/reject persistence failures leaving tasks awaiting verification

## Verification
- MASC_DUNE_THROTTLE=0 DUNE_JOBS=1 scripts/dune-local.sh build test/test_verification_fsm.exe
- ./_build/default/test/test_verification_fsm.exe (17 tests)

### DIFF
--- a/lib/coord/coord_task.ml
+++ b/lib/coord/coord_task.ml
@@ -1035,6 +1035,7 @@ let transition_task_r
       ~action
       ?agent_tool_names
       ?prepare_verification_request
+      ?prepare_verification_verdict
       ?expected_version
       ?(notes = "")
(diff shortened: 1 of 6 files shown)


jeong-sik:
Closing as duplicate of #10308. The two PR heads have the same patch-id; keeping #10308 because it is already ready for review and CI is running there.

PR #10329, opened 2026-04-25
diag(cost-ledger): self-describing cost_usd_source per emit (#10318)

jeong-sik:
## Goal

Make every \`costs.jsonl\` entry self-describing about WHY \`cost_usd\` is what it is, so operators can pick the right structural fix from the issue's 5-option list.

## Symptom (from #10318)

\`~/me/.masc/costs.jsonl\` 1697 entries: 100% \`cost_usd = 0.0\`, 52% \`provider = unknown\`, 33% empty model. Tracking system runs but produces no actionable data. Same observability gap pattern as #10269 (silent personality re-sync) and #10314 (silent accountability drops): the failure has multiple distinct upstream causes that all collapse to a single 0.0 in the ledger.

## Root cause walked

\`emit_cost_event\` in \`lib/keeper/keeper_hooks_oas.ml\` zeros \`cost_usd\` on at least 5 distinct upstream paths and the entry schema had no field to distinguish them:

1. \`usage_missing = true\` — provider adapter never surfaced usage.
2. \`not usage_trusted\` — trust classifier suppressed the value (line 471).
3. \`structurally_unmetered_provider\` — local LLM (ollama, etc.); 0 by design.
4. \`pricing_catalog_miss\` — model not in \`Llm_provider.Pricing\` catalog (line 305-315 already increments \`masc_pricing_catalog_miss_total{model}\` but the JSONL row doesn't carry that fact).
5. Actually-zero call (tool-only, empty completion) — trusted+priced but tokens=0.

Each silent path collapses to the same \`(\"cost_usd\", \`Float 0.0)\` field. The operator sees \"100% zeros\" without the next concrete action.

## Change

Add a \`cost_usd_source : string\` field to every JSONL entry. Classification precedence is fixed top-down (upstream-most failure mode wins):

| order | path | source string |
|---|---|---|
| 1 | \`usage_missing\` | \`missing_usage\` |
| 2 | \`not usage_trusted\` | \`untrusted_usage\` |
| 3 | \`structurally_unmetered_provider\` | \`unmetered_provider\` |
| 4 | trusted, cost > 0 | \`computed\` |
| 5 | trusted, 0, no catalog entry | \`pricing_catalog_miss\` |
| 6 | trusted, 0, catalog hit | \`zero_token_call\` |

Plus a Prometheus counter \`masc_cost_emit_zero_source_total{source}\` that ticks on every non-\`computed\` emit. Cardinality: 5 non-computed labels.

Operators can now run:
\`\`\`
jq -r '.cost_usd_source' costs.jsonl | sort | uniq -c
\`\`\`
and see exactly which silent path dominates. The Prometheus counter gives the same split at runtime.

## Why precedence top-down

Picking the most-upstream failure mode for the label keeps the signal actionable. A call where \`usage_missing=true\` AND the model is also missing from the pricing catalog should label as \`missing_usage\` — that's the more-actionable failure (provider adapter contract) and fixing it would auto-resolve any downstream ambiguity. Test \`missing_usage_wins_over_everything\` pins this.

## Diagnostic-first vs structural-fix

This is option C-light from the issue's 5-option list (A audit / B provider unwrap / C pricing SSOT / D validation gate / E #9959 cross-link). All four structural options need this evidence to know which path dominates:

- High \`pricing_catalog_miss\` rate → option C (add upstream OAS pricing entries).
- High \`untrusted_usage\` rate → option E (#9959 unification — same trust classifier).
- High \`missing_usage\` rate → option B (provider unwrap at emit time).
- High \`unmetered_provider\` rate → no action needed (correct behavior; just informational).

Without this counter, A/B/C/D/E are guesses on which path dominates.

## Behavior preserved

- \`cost_usd\` field value unchanged (still \`safe_cost_usd\`).
- \`usage_missing\`, \`raw_usage_fields\`, \`telemetry_fields\` all unchanged.
- No new log lines on the happy path; counter only ticks on previously-silent 0.0 paths.
- Schema additive — old consumers that ignore unknown fields keep working.

## Tests (\`test_cost_usd_source_attribution_10318\`, 8 cases, all green)

| group | case | pin |
|---|---|---|
| precedence | missing_usage wins over priced+positive | upstream gate dominates |
| precedence | untrusted_usage wins over unmetered+catalog | classifier dominates |
| canonical-paths | unmetered for trusted ollama | not labeled \`pricing_catalog_miss\` |
| canonical-paths | computed for trusted+paid+positive | happy path |
| canonical-paths | pricing_catalog_miss for unknown model | actionable signal |
| counter | computed does not tick | rate stays clean for healthy emits |
| counter | named source +1 | each silent path observable |
| counter | sources isolated by label | dashboards split cleanly |

Sibling tests (no regression):
- \`test_keeper_usage_trust_counter\`: 5/5 green.
- \`test_keeper_hooks_oas_telemetry\`: 15/15 green.

## Out of scope (separate /loop ticks once evidence lands)

- **Option A** (cost calculator audit): the JSONL grep replaces the \`rg \"cost_usd\"\` caller graph for first-pass triage.
- **Option B** (provider unwrap at emit time): the cascade dispatcher should propagate underlying provider info before wrapper layers strip it; out of scope for this layer.
- **Option C** (pricing table SSOT in config/pricing.toml): only worth doing once the counter shows pricing_catalog_miss is the dominant rate.
- **Option D** (strict validation gate): a future flag can reject \`missing_usage\` emits on paid providers; needs evidence on whether legitimate paths produce them.
- **Option E** (#9959 unification): if untrusted_usage dominates, the trust classifier is the shared root cause; unify there.

Refs #10318. Pattern: feedback memory \`no-collapse-richer-enum-at-sdk-boundary\` — the 5 silent paths are the boundary collapse this PR un-collapses.


jeong-sik:
## Conflict analysis (`lib/keeper/keeper_hooks_oas.ml`)

Rebase against current `origin/main` produces two conflict regions, both mechanical.

### Conflict 1 (line ~716): provider field rename
- **HEAD**: `("provider", \`String provider)` — `provider` is now the canonical local binding (`let provider = provider_of_model_with_telemetry ~model ~telemetry`)
- **Branch**: `("provider", \`String provider_for_entry)` — references variable that no longer exists in HEAD

**Resolution**: take HEAD's `provider` (variable renamed upstream). The branch's `provider_for_entry` symbol is gone in main.

### Conflict 2 (line ~728): cost ledger field additions
- **HEAD** added 5 fields: `cost_status`, `cost_status_reason`, `cost_pricing_model`, `cost_pricing_catalog`, `model_resolution_source`
- **Branch** adds 1 field: `cost_usd_source`

**Resolution**: union — keep all 6 fields. Two changes are complementary (cost-status vs cost-source attribution).

### Suggested rebase outcome (concept)
```ocaml
("provider", `String provider);
("model", `String model);
("input_tokens", `Int safe_input_tokens);
("output_tokens", `Int safe_output_tokens);
("cost_usd", `Float safe_cost_usd);
("cost_status", `String cost_status_label);
("cost_status_reason", `String cost_status_reason_label);
("cost_pricing_model", `String pricing_model);
("cost_pricing_catalog", `String (pricing_catalog_status ~pricing_model));
("model_resolution_source", `String (model_resolution_source_for_ledger ~model ~pricing_model));
("cost_usd_source", `String cost_usd_source);
```

Posting from /loop saturation tick — declined to auto-resolve since this is autocoder's PR, leaving for original author/reviewer to apply the union.

jeong-sik:
## #10333 머지로 base 변동, 재평가 필요

#10333 (`fix(cost): resolve auto model pricing status (#10318)`) 가 이미 main에 머지되어 100% `cost_usd=0` 의 root cause 중 auto-model pricing path는 해결됨.

이 PR (#10329, diagnostic) 는 self-describing `cost_usd_source` per emit를 추가하는 별도 가치:

- 52% `provider=unknown` (#10333 비대상)
- 33% empty model (#10333 비대상)

위 secondary failure modes는 여전히 진단 데이터 부족. 이 PR 의 diag 가치는 보존되지만 base가 바뀌어서 rebase가 필요함 (현재 mergeStateStatus=DIRTY).

운영자/리뷰어 판단:

1. **rebase + 유지** — secondary cases 진단 위해 머지
2. **close** — secondary는 별도 ticket으로 분리, 이 PR은 superseded
3. **scope 축소** — 52% / 33% 만 다루는 좁은 진단으로 재작성

이 cycle에서 결정 보류. 다음 cycle 또는 user direction에 따라 처리.

jeong-sik:
Resolved the merge conflict and pushed the rebased head `7a80f6d471`.

Conflict resolution detail:
- Kept the current main `cost_status`, `cost_status_reason`, `cost_pricing_model`, `cost_pricing_catalog`, and `model_resolution_source` fields.
- Kept this PR's `cost_usd_source` field and counter behavior.
- Routed `cost_usd_source` pricing-catalog classification through the same `pricing_model` used by the newer cost-status fields, so alias/model resolution does not diverge between the two ledger signals.

Local verification:
- `git diff --check`
- `scripts/dune-local.sh build test/test_cost_usd_source_attribution_10318.exe test/test_keeper_usage_trust_counter.exe test/test_keeper_hooks_oas_telemetry.exe`
- `./_build/default/test/test_cost_usd_source_attribution_10318.exe` (8 tests, run ID `QB4UJ8O7`)
- `./_build/default/test/test_keeper_usage_trust_counter.exe` (5 tests, run ID `7QHBH9TX`)
- `./_build/default/test/test_keeper_hooks_oas_telemetry.exe` (17 tests, run ID `AFK3OZLU`)

CI restarted after the force push; this should clear the previous DIRTY state once GitHub recomputes mergeability.

PR #10343, opened 2026-04-25
fix(bonsai): eliminate keeper SSOT bug by removing 3-source merge

jeong-sik:
## Summary

- Remove all references to deleted `Directory_execution_types` and `Directory_mission_types` from `keepers_directory.ml` and `keepers_view.ml`
- Simplify `keepers_directory.ml` from 1517→520 lines (-65%) by replacing 3-source Hashtbl merge with single-source `Keepers_types.keeper` mapping
- Structurally eliminates the SSOT bug where nick-cave displayed two different FSM states simultaneously

## Root Cause

`build_rows` merged three independently-polled data sources by name-based Hashtbl lookup:
1. `Keepers_types` (keepers summary endpoint)
2. `Directory_execution_types` (execution composite endpoint)
3. `Directory_mission_types` (mission briefing endpoint)

Each source polls at different intervals. When they disagreed on the same keeper's state, the status pill (from `pipeline_stage`) and phase subline (from `phase`) could show contradictory states. This was compounded by identity drift (`nick-cave` vs `nick0cave`).

## Fix

Single-source simplification:
- `band_of_keeper`: maps `Keepers_types.keeper_status` (Live/Warn/Dead) directly to band
- `build_rows`: iterates `Keepers_types.response.keepers` with no cross-source lookup
- `status_meta`: derives labels from `keeper.stat` and band, no runtime/mission fields
- Removed 10+ helper functions that depended on deleted types

## Test plan

- [ ] Server-side build: `dune build @install` passes
- [ ] Bonsai build: requires `bonsai_web` opam package (not in default switch) — verify in Bonsai dev environment
- [ ] Visual: `/dashboard/b/` Keepers tab shows single consistent state per keeper
- [ ] No regression: Roster, Swim, Ctx_chart views unchanged (use `Keepers_types.response` directly)


jeong-sik:
Direct review follow-up pushed: 7dbe71fe6a\n\nBlocking issue found during SSOT review: this PR removed the 3-source merge, but the remaining source (/dashboard/b/api/keepers/summary) was still keepers_summary_mock(). That would turn the Bonsai directory into a cleaner-looking but still fake SSOT.\n\nPatch:\n- remove the static 4-keeper mock from server_routes_http_pages.ml\n- project /dashboard/b/api/keepers/summary from Keeper_registry.all ~base_path, scoped to the live server room_config.base_path\n- return a clear server-state error instead of silently serving fixture rows before server initialization\n- add regression coverage proving the projection is base_path-scoped, uses runtime usage/tool data, and omits mock keepers\n\nLocal verification:\n- git diff --check\n- bash scripts/health_snapshot.sh --skip-build --baseline-ref origin/main --fail-on-lib-regression\n- scripts/dune-local.sh build test/test_keeper_registry.exe\n- ./_build/default/test/test_keeper_registry.exe (64 tests, run ID NAL3D743)

PR #10344, opened 2026-04-25
fix: add ssh to sandbox required commands and increase git status timeout

jeong-sik:
## Description
Addresses Category 3 issues from the backlog:

1. **Sandbox SSH Missing:** Added `ssh` to `required_commands` in `keeper_sandbox_runtime.ml` so preflight correctly validates it and uses the sandbox with `openssh-client` installed.
2. **Git Status Timeout:** Increased `default_git_status_timeout_sec` from 15.0 to 30.0 in `worktree_live_context.ml` to prevent recurring `Timeout after 15s: git status --porcelain` warnings in large workspaces with many worktrees.

### DIFF
--- a/lib/keeper/keeper_sandbox_runtime.ml
+++ b/lib/keeper/keeper_sandbox_runtime.ml
@@ -112,6 +112,7 @@ let required_commands =
     "make";
     "opam";
     "dune";
+    "ssh";
   ]
 
 let option_field name = function


jeong-sik:
Post-merge follow-up: direct review found the merged diff did not include the advertised git status timeout change. The merged PR added `ssh`, but `default_git_status_timeout_sec` and its test still stayed at 15s. Opened #10360 to apply the missing 30s default and focused test update. Local verification on #10360: diff check, health snapshot, focused Dune build, `test_worktree_live_context.exe` (7 tests).

PR #10367, opened 2026-04-25
perf(output-parse,memory-bank): hoist Re.compile out of pytest/cargo/consensus paths

jeong-sik:
## Summary

Two more module-init regex hoists in code paths called many times per minute under normal keeper traffic.

### \`lib/exec/output_parse.ml\`
- \`parse_pytest.extract_count\` compiled the same PCRE **twice** per prefix (\`execp\` then \`exec\`), four prefixes per matching summary line. Prefix set is closed (\`passed\`/\`failed\`/\`error\`/\`skipped\`), so the four DFAs become module-level constants and the double-compile collapses to a single \`Re.exec_opt\`.
- \`parse_cargo_test\` had the same pattern with two literal needles; also hoisted.

### \`lib/keeper/keeper_memory_bank.ml\`
- \`normalize_memory_text_key\` compiled the punctuation-strip PCRE on every call. Pattern is fully static → module-level \`let\`.
- \`consensus_re ()\` compiled the env-conditional PCRE on every call to \`has_inflated_consensus_marker\`. Env var is process-lifetime constant → wrap in \`Stdlib.lazy\` and force once.
  - **Stdlib.lazy, not Eio.Lazy**: callers include plain-OCaml unit tests with no Eio context. The thunk is a pure regex compile with no effects, so the OCaml 5 multi-domain \"double force\" concern reduces to redundant compute with identical results. Tried Eio.Lazy first — \`test_keeper_memory\` blew up with \`Stdlib.Effect.Unhandled(Eio__core__Cancel.Get_context)\`.

## Behavior preservation

- pytest/cargo: int extraction semantics unchanged; \`exec_opt None\` matches the prior \`execp=false\` branch.
- consensus regex: invalid env-supplied PCRE still raises (now at first force instead of every call), matching the original observable behavior on the first invocation.

## Test plan

- [x] \`dune build @check\` — clean
- [x] \`test_keeper_memory\` — 63/63 pass
- [x] \`test_output_parse_p14\` — 12/12 pass

### DIFF
--- a/lib/exec/output_parse.ml
+++ b/lib/exec/output_parse.ml
@@ -9,6 +9,18 @@ let files_changed_re = Re.Pcre.re {|\b(\d+) file|} |> Re.compile
 let insertions_re = Re.Pcre.re {|\b(\d+) insertion|} |> Re.compile
 let deletions_re = Re.Pcre.re {|\b(\d+) deletion|} |> Re.compile
 
+(* --- pytest summary counters: prefix set is closed (passed/failed/error/
+   skipped), so build the four DFAs once at module init instead of per line.
+   Using exec_opt means the prior execp+exec double-compile is also gone. *)
+let pytest_passed_re = Re.Pcre.re {|(\d+) passed|} |> Re.compile
+let pytest_failed_re = Re.Pcre.re {|(\d+) failed|} |> Re.compile
+let pytest_error_re = Re.Pcre.re {|(\d+) error|} |> Re.compile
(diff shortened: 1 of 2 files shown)


jeong-sik:
Post-merge follow-up from the direct review: #10377 fixes the remaining OCaml 5 Lazy.force concurrency risk in the hoisted consensus regex cache and adds regression coverage for env-pattern refresh / invalid-pattern fallback.

PR #10378, opened 2026-04-25
fix(keeper): fall back to gh CLI keychain for sandbox GH_TOKEN

jeong-sik:
## Summary

- Add macOS keychain fallback for `gh_token()` so Docker-sandbox keepers can authenticate GitHub operations even when the server process was launched without `GH_TOKEN`.
- Keep the existing resolution order explicit: `MASC_KEEPER_SANDBOX_GH_TOKEN` override, host `GH_TOKEN`, then host `gh auth token --hostname github.com`.
- Bound the fallback with `MASC_KEEPER_SANDBOX_GH_TOKEN_PROBE_TIMEOUT_SEC` (default 2s, clamped 0.1s..10s) and cache the result for the process lifetime, so a slow or broken host `gh`/keychain path cannot stall every keeper git/gh command.

## Why this matters

This is directly in the keeper autonomy path. On macOS, `gh auth login` commonly stores the actual token in the OS keyring, while the mounted `~/.config/gh/hosts.yml` only contains metadata. Before this PR, a server launched from an interactive shell without `GH_TOKEN` injected no token into the Docker git-creds path, so containerized keepers repeatedly hit GitHub HTTP 401 even though the host `gh` session was valid.

The hardening commit avoids trading that 401 for a new hang class: the keychain fallback is no-shell, timeout-bounded, exception-swallowing, and cached. Hard mode still returns `""` and does not probe host credentials.

## Test plan

- [x] `git diff --check`
- [x] `scripts/dune-local.sh build test/test_keeper_docker_read.exe`
- [x] `./_build/default/test/test_keeper_docker_read.exe` (32 tests, run `K7YIB39Z`)
- [ ] Deploy + restart server: keeper `gh pr list` / `git push` succeed without 401
- [ ] `system_log` HTTP 401 event count drops to near zero over a few hours

## Review focus

- Verify the credential precedence is acceptable for sandboxed keepers.
- Verify the bounded keychain probe is failure-soft: no token configured, missing `gh`, nonzero exit, or timeout all return `""` rather than failing the keeper command path.


jeong-sik:
Reviewed and patched this PR directly. Follow-up commit c2e370efa5 keeps the macOS gh keychain fallback, but makes it keeper-safe: no shell invocation, timeout-bounded probe, process-lifetime cache for both token and empty result, hard-mode bypass, and explicit tests for override/env/keychain precedence. Local verification passed: git diff --check; scripts/dune-local.sh build test/test_keeper_docker_read.exe; ./_build/default/test/test_keeper_docker_read.exe (32 tests, run L9LAB3Q1).

jeong-sik:
Amended follow-up to 442a9af60d after Meta Guards flagged the try/ignore waitpid pattern. Local SIL gate now passes: bash scripts/ci/check-silent-failure-patterns.sh. Re-ran git diff --check, scripts/dune-local.sh build test/test_keeper_docker_read.exe, and ./_build/default/test/test_keeper_docker_read.exe (32 tests, run K7YIB39Z).

PR #10357, opened 2026-04-25
fix(keeper): persist cache usage metrics

jeong-sik:
## Summary
- Persist `cache_creation_tokens` and `cache_read_tokens` inside keeper metrics `usage` JSONL rows.
- Keep unreported usage explicit with `null` cache fields instead of absent fields.
- Add regression coverage for reported cache usage and unreported null cache fields.

## Live remeasure
- OAS #1186 is merged and current MASC main is already pinned past that fix.
- Scanning `~/me/.masc/keepers/*/metrics/2026-04/25.jsonl` after `2026-04-25T09:56:54Z` found 21 turn rows, all `claude_code:auto`.
- `input_tokens > 1_000_000`: 0 rows.
- `(input_tokens=0, output_tokens=0)` turn usage: 0 rows.
- Cache fields were absent from keeper metrics `usage`, which this PR fixes.

Closes #9959

## Validation
- `git diff --check origin/main...HEAD`
- `env DUNE_BUILD_DIR=_build_9959_cache_usage_jsonl_latest DUNE_JOBS=2 DUNE_CACHE=disabled MASC_DUNE_THROTTLE=0 scripts/dune-local.sh build test/test_keeper_unified.exe test/test_keeper_usage_trust_anthropic_cache_9959.exe`
- `env MASC_BASE_PATH=/tmp/masc-test-9959-cache-usage-jsonl-latest _build_9959_cache_usage_jsonl_latest/default/test/test_keeper_unified.exe` (292 tests)
- `env MASC_BASE_PATH=/tmp/masc-test-9959-cache-usage-trust-latest _build_9959_cache_usage_jsonl_latest/default/test/test_keeper_usage_trust_anthropic_cache_9959.exe` (9 tests)
- `bash scripts/ci/check-telemetry-coverage.sh`
- `bash scripts/ml_line_cap_audit.sh --baseline-ref origin/main --fail-on-regression`
- `bash scripts/ci/check-silent-failure-patterns.sh`

PR #10379, opened 2026-04-25
fix(keeper): price cache usage in turn cost

jeong-sik:
## Summary

Follow-up to #10357. That PR persisted `cache_creation_tokens` and `cache_read_tokens`, but the keeper turn-cost paths still called pricing with only regular input/output tokens. Rows with cache usage therefore looked cache-aware while `cost_usd` and runtime `total_cost_usd` stayed inflated.

This PR routes keeper turn-cost calculation through one helper that uses the OAS pricing API with cache creation/read token fields.

It also makes heartbeat/status metrics rows preserve the same cache-token field shape with explicit zeroes, so consumers scanning `keepers/*/metrics/YYYY-MM/DD.jsonl` no longer have to special-case missing cache fields by snapshot source.

## Why it matters

- `cost_usd` now matches the token fields persisted in keeper metrics JSONL.
- Unified keeper turns, direct keeper turns, and metrics update all share the same cache-aware pricing helper.
- Unknown/untrusted usage still falls through conservatively: untrusted usage records zero turn cost, and known OAS catalog misses still use the existing pricing-miss metric path.

## Verification

- `git diff --check`
- `./_build/default/test/test_keeper_unified.exe` (293 tests)
- `scripts/dune-local.sh build test/test_keeper_hooks_oas_pricing.exe`
- `./_build/default/test/test_keeper_hooks_oas_pricing.exe` (6 tests)
- `scripts/dune-local.sh build test/test_smart_heartbeat_keepalive.exe`
- `./_build/default/test/test_smart_heartbeat_keepalive.exe` (18 tests)

## Review focus

- `lib/keeper/keeper_unified_metrics.ml`: shared `estimate_trusted_usage_cost_usd` helper.
- `lib/keeper/keeper_unified_turn.ml` / `lib/keeper/keeper_turn.ml`: callsite routing.
- `lib/keeper/keeper_keepalive.ml`: status-tick usage schema stays cache-field complete.
- `test/test_keeper_unified.ml`: regression pins Anthropic cache multipliers at the unified keeper layer.
- `test/test_smart_heartbeat_keepalive.ml`: heartbeat usage schema regression.


jeong-sik:
Updated after the initial PR publish to cover the remaining heartbeat/status schema gap from the #10357 review: status-tick usage now emits explicit zero cache_creation_tokens/cache_read_tokens, with test_smart_heartbeat_keepalive coverage. New head: eb3c655122.

PR #10381, opened 2026-04-25
perf(link-preview): cache compiled PCREs in first_match

jeong-sik:
## Summary

\`first_match\` compiled a fresh PCRE per call with \`[\`CASELESS; \`DOTALL]\` flags. The flags make the build heavier than a plain \`Re.str\`, and the function fans out across every link-preview parse (≥10 calls per page via \`meta_content\` / \`link_href\` / \`title_tag\`).

## Approach

Patterns are dynamic per-call (\`Printf.sprintf\` with \`attr\`/\`key\`/\`rel_value\`) but the inputs are drawn from a fixed OG/Twitter/HTML vocabulary, so the unique-pattern set across a process lifetime is ≤30. Caching by pattern string collapses repeats to a single compile per distinct pattern.

## Concurrency

\`Stdlib.Mutex\` (not \`Eio.Mutex\`) because the dashboard HTTP handler may execute across multiple Eio domains. Compile happens *outside* the critical section — \`Re.compile\` is pure, so a racing duplicate compile is harmless: structurally identical \`Re.re\` values and last-writer-wins on \`Hashtbl.replace\` gives the same observable result.

## Test plan

- [x] \`dune build @check\` clean
- [x] \`test_dashboard_link_preview\` 3/3

### DIFF
--- a/lib/server/server_dashboard_http_link_preview.ml
+++ b/lib/server/server_dashboard_http_link_preview.ml
@@ -201,10 +201,37 @@ let first_head_fragment body =
   | Some groups -> Re.Group.get groups 1
   | None -> String.sub body 0 (min (String.length body) max_html_chars)
 
+(* Pattern cache for [first_match].  Patterns are constructed dynamically
+   per call site (via [Printf.sprintf] inside [meta_content]/[link_href])
+   but the [attr]/[key]/[rel_value] inputs are drawn from a fixed
+   OG/Twitter/HTML vocabulary, so the unique-pattern set across a process
+   lifetime is ≤30.  Caching collapses repeat link-preview calls to a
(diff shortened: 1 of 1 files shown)


jeong-sik:
Reviewed this from the concurrency/cache angle. No blocker found: this avoids the unsafe Lazy pattern and protects the shared Hashtbl with Stdlib.Mutex; first_match patterns are built from fixed call-site vocabularies, so cache growth is bounded by the current extractor surface. Local verification passed on the PR head: git diff --check; scripts/dune-local.sh build test/test_dashboard_link_preview.exe; ./_build/default/test/test_dashboard_link_preview.exe (3 tests, run 1G1YN555). CI Build and Test was still pending at review time.

Issue #10349, opened 2026-04-25
Keeper identity drift across contract/gate/FS resolver layers — sandbox isolation oracle leak (turn 433 trace evidence)


jeong-sik:
## Summary

Three independent layers (contract emitter, gate authorizer, FS resolver) read keeper identity from divergent sources on the same turn. Result: a keeper authorized for sandbox A has its filesystem operations enforced against sandbox B, with a side-channel leak revealing sandbox B's directory structure.

This is the FS-resolver-layer variant of the keeper identity drift class previously documented in `feedback_identity-2layer-drift-metrics-proof`. PR #10343 addresses the same class at the dashboard `build_rows` layer; this issue covers the security-relevant FS resolver layer.

## Evidence (single trace, ground truth verified)

Source: `~/me/.masc/trajectories/masc-improver/trace-1776871892035-b93f2.jsonl` (turn 433, task-049, session `mcp_VHnzoPJ_31Q_187C`).

| Layer | Field | Value |
|---|---|---|
| Contract emitter | `runtime_contract.keeper_name` | `masc-improver` |
| Contract emitter | `runtime_contract.sandbox_profile` | `docker` |
| Contract emitter | `runtime_contract.sandbox_root` | `/Users/dancer/me/.masc/playground/docker/masc-improver/` |
| Contract emitter | `runtime_contract.allowed_paths` | `[".masc/playground/docker/masc-improver/"]` |
| Gate | `gate.status` | **pass** |
| FS resolver (`error` text) | `roots=[...]` | `[/Users/dancer/me/.masc/playground/analyst]` ❌ |
| FS resolver (`Available entries`) | leaked | `.playground_state.json, mind, repos` (= ground truth of `analyst/` dir, **not** `masc-improver/`) |

Filesystem ground truth (verified on host):
- `~/me/.masc/playground/analyst/`         → `.playground_state.json`, `mind`, `repos` (3 entries — matches leak)
- `~/me/.masc/playground/masc-improver/`   → `.playground_state.json`, **`docker-probe.txt`**, `mind`, `repos` (4 entries — distinguishable)
- `~/me/.masc/playground/docker/masc-improver/` (contract-claimed root) → exists separately

The leaked `Available entries` exactly match `analyst/` (excluding `docker-probe.txt`), confirming the resolver is enumerating analyst's filesystem rather than misformatting paths.

## Reproduction context

- Tool: `keeper_fs_read`
[...]

PR #10383, opened 2026-04-25
fix(keeper-fs): scrub roots= leak from path-rejection errors (#10349)

jeong-sik:
## Goal

Close the keeper FS path-rejection side channel from #10349. The resolver must not echo its view of a keeper's allowed sandbox roots back to the LLM when a read path is rejected.

## Root Cause

The original leak was visible on `path_not_found_under_allowed_roots: <path> (roots=[...])`.

Review found a second LLM-visible path in the same resolver: `path_outside_sandbox` still included `(sandbox roots: [...])`, and relative-path errors also carried `resolved=<absolute path>`. With keeper identity drift, those roots can describe the wrong keeper's sandbox and become an oracle for sibling directory layout.

## Change

- Make `path_not_found_under_allowed_roots` opaque while preserving the existing classifier prefix.
- Make `path_outside_sandbox` opaque as well:
  - no `(sandbox roots: [...])`
  - no `resolved=<absolute path>`
  - no full playground root in the hint
- Preserve operator observability with `masc_keeper_path_rejection_total{kind}`.
- Tick `kind="out_of_roots"` for outside-sandbox read rejections too, not only missing absolute leaves.

## Verification

- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_path_roots_leak_10349.exe test/test_keeper_tool_exposure.exe`
- `./_build/default/test/test_keeper_path_roots_leak_10349.exe` (4 tests, run ID `L0YKPOCP`)
- `./_build/default/test/test_keeper_tool_exposure.exe` (59 tests, run ID `QZGWJ348`)

Refs #10349.

### DIFF
--- a/lib/keeper/keeper_alerting_path.ml
+++ b/lib/keeper/keeper_alerting_path.ml
@@ -197,28 +197,24 @@ let format_path_rejection ~(raw : string) ~(resolved : string)
     ~(allowed_norms : string list) : string =
   let resolved_hint =
     if Filename.is_relative raw && resolved <> raw then
-      Printf.sprintf
-        "; relative paths are checked against your sandbox boundary (resolved=%s)"
-        resolved
+      "; relative paths are checked against your sandbox boundary"
     else
       ""
   in
   let playground_hint =
     if raw_looks_like_playground_subdir raw then
       match playground_root_of_allowed allowed_norms with
-      | Some pg ->
+      | Some _ ->
         Printf.sprintf
           ". Your raw path already looks sandbox-relative. Use it as-is \
-           (for example path=%S) and do not prepend %s; call \
-           keeper_context_status and use sandbox_repos / sandbox_mind if \
-           unsure."
-          raw pg
+           (for example path=%S); call keeper_context_status and use \
+           sandbox_repos / sandbox_mind if unsure."
+          raw
       | None -> ""
     else ""
   in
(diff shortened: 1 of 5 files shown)


jeong-sik:
Review follow-up patched and pushed at `6db3f07486`.

The original fix scrubbed the `path_not_found_under_allowed_roots` strings, but review found one remaining leak path: `path_outside_sandbox` still exposed `(sandbox roots: [...])`, and relative-path failures exposed `resolved=<absolute path>`. That is the same class of oracle when keeper identity drifts.

Follow-up change:
- `path_outside_sandbox` keeps the classifier prefix but no longer prints sandbox roots, resolved absolute paths, or full playground root hints.
- outside-sandbox read rejections now tick `masc_keeper_path_rejection_total{kind="out_of_roots"}`.
- added a regression case for the outside-sandbox branch, so the test now covers the previously missed route.

Local verification:
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_path_roots_leak_10349.exe test/test_keeper_tool_exposure.exe`
- `./_build/default/test/test_keeper_path_roots_leak_10349.exe` (4 tests, run ID `L0YKPOCP`)
- `./_build/default/test/test_keeper_tool_exposure.exe` (59 tests, run ID `QZGWJ348`)

CI restarted for the new head and should be watched before any ready transition.

PR #10347, opened 2026-04-25
perf(keeper): drop per-call Re.compile from substring keepers, byte-wise String_util

jeong-sik:
## Summary

Four keeper hot-path substring helpers compiled fresh `Re.t` per call:

- **`Keeper_memory_policy.contains_any_ci`** — 14 keepers × turn × N keywords paid `Re.compile` + 2 `String.lowercase_ascii` per needle.
- **`Keeper_world_observation`** (×2) — mention scan over board posts rebuilt `Re.t` per `(target × post)`.
- **`Keeper_memory_recall.expected_topic_hint`** — 6+ inline `Re.execp (Re.str ... |> Re.compile)` per query; every memory recall paid the cost.

All sites collapse to substring containment. `String_util` already exposes `contains_substring` / `contains_substring_ci`; both are now rewritten as byte-wise scans using `String.unsafe_get`:

```ocaml
(* before *)
let contains_substring haystack needle =
  ...
  if idx + needle_len > hay_len then false
  else if String.sub haystack idx needle_len = needle then true
  ...

(* after *)
let contains_substring haystack needle =
  ...
  let rec match_at i j =
    if j = needle_len then true
    else if String.unsafe_get haystack (i + j) <> String.unsafe_get needle j then false
    else match_at i (j + 1)
  in
  ...

(* contains_substring_ci folds case inline during compare,
   no String.lowercase_ascii pre-allocation *)
```

Keepers switch to these helpers — per-call `Re.compile` is gone, both `String.lowercase_ascii` allocations are avoided in the CI path.

Same anti-pattern series:
- #10250 / #10252: `Mention` (merged)
- #10260: `Coord_gc.mentions_open_task` (merged)
- #10267: `Mcp_server_eio_call_tool.contains_casefold` (merged)
- #10286: dashboard helpers (merged)
- #10301: `Bounded` + `Auto_recall` (merged)
- #10303: `Masc_error_recovery` + `Coord_git` (merged)
- #10315: `Prompt_registry.render_template` (merged)
- #10323: link preview module hoist (merged)
- #10324: `Safe_path` + `validate_room_id` (merged)
- #10337: `Verification` + `Tool_library` (in-flight)
- This PR: keepers + `String_util` byte-wise

## Test plan
- [x] `dune build @check` clean
- [x] `dune build` for downstream test executables (test_anti_rationalization, test_keeper_shell_containment, test_team_memory, test_keeper_bash_safety) clean
- Behavior preserved:
[...]

PR #10384, opened 2026-04-25
perf(gh-cmd-validation): drop local contains_substring, route through String_util SSOT

jeong-sik:
## Summary

\`gh_command_validation.ml\` carried its own \`contains_substring\` helper that allocated \`String.sub\` per inner-loop step — exactly the form the SSOT \`String_util.contains_substring\` (PR #10347) replaced with a byte-wise \`String.unsafe_get\` scan.

## Impact

The local helper was on 4 callsites deep in the gh-classification hot path:

- \`gh_api_graphql_is_destructive\`: lower-cases cmd then runs \`has\` over every R2 mutation name. Per gh dispatch.
- \`has_mutating_http_method\` joined-search: 5 patterns × every API call.
- \`gh_option_takes_value\`: every parsed gh token.
- \`is_gh_dangerous_operation\` graphql destructive scan: \`List.exists\` over the destructive-mutation table.

Removing the local helper:
- collapses to a single SSOT call site (no logic duplication)
- inherits the byte-wise \`unsafe_get\` loop instead of the per-step \`String.sub\` allocation
- equivalent behavior on empty needle (both return true)

## Test plan

- [x] \`dune build @check\` clean
- [x] \`test_worker_dev_tools\` 120/120

### DIFF
--- a/lib/gh_command_validation.ml
+++ b/lib/gh_command_validation.ml
@@ -12,16 +12,6 @@ let forbidden_shell_chars =
 let contains_forbidden_shell_chars cmd =
   String.exists (fun ch -> List.mem ch forbidden_shell_chars) cmd
 
-let contains_substring s needle =
-  let s_len = String.length s in
-  let needle_len = String.length needle in
-  let rec loop i =
-    if i + needle_len > s_len then false
-    else if String.sub s i needle_len = needle then true
-    else loop (i + 1)
-  in
-  loop 0
-
 (** Top-level gh CLI commands allowed. Commands not in this list are
     rejected at the allowlist gate. *)
 let gh_allowed_commands =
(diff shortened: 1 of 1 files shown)


jeong-sik:
Reviewed the PR from the command-validation angle.

I did not find a blocker in this change. The local helper removal preserves the existing case-sensitivity at each callsite: destructive GraphQL detection still normalizes to lowercase before substring checks, while REST path detection keeps the existing raw joined argument behavior. Empty-needle semantics from `String_util.contains_substring` are not reached by these callsites in a way that changes validation behavior.

Local verification:
- `scripts/dune-local.sh build test/test_worker_dev_tools.exe`
- `./_build/default/test/test_worker_dev_tools.exe` (120 tests, run ID `C5I7U0VD`)

CI is still pending on Build/Test, Health, and Lint at head `82d23b0d6ad02b2e96ef5c34e58efc6bb1c79b6b`, so I would wait for those before treating it as merge-ready.

PR #10391, opened 2026-04-25
fix(anti_rationalization): add Korean rationalization patterns (#10385)

jeong-sik:
Closes the false-negative half of #10385. Pre-fix excuse_patterns.json had 13 English ASCII-only needles; the matcher lowercased input via String.lowercase_ascii (no-op on non-ASCII bytes) and substring-matched byte-wise. The matcher handles Korean fine — Korean needles simply did not exist, so keeper-fleet detection on the dominant 한국어 surface was 0% while institution_episodes.jsonl carried real entries (나중에, 범위 밖, 재현 안됨). Added 10 Korean needles covering the same semantic classes already covered in English (defer, out-of-scope, unverifiable claim, pre-existing, not-reproduced). 10 new test cases pin Korean detection + English regression. English false-positive risk (substring without word boundary) remains tracked under the issue's option C/D follow-up.

### DIFF
--- a/config/excuse_patterns.json
+++ b/config/excuse_patterns.json
@@ -11,5 +11,15 @@
   ["works on my machine", "unverifiable claim"],
   ["not reproducible", "dismissing without investigation"],
   ["not my responsibility", "responsibility deflection"],
-  ["cannot reproduce", "dismissing without investigation"]
+  ["cannot reproduce", "dismissing without investigation"],
+  ["나중에", "deferring work to later (ko)"],
+  ["범위 밖", "declaring work out of scope (ko)"],
+  ["의도 외", "declaring work outside intent (ko)"],
+  ["재현 안", "dismissing without investigation (ko)"],
+  ["재현되지 않", "dismissing without investigation (ko)"],
+  ["기존 문제", "claiming the problem already existed (ko)"],
+  ["내 환경에선", "unverifiable claim (ko)"],
+  ["내 환경에서는", "unverifiable claim (ko)"],
+  ["후속 pr", "deferring to a follow-up (ko)"],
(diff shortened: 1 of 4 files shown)


jeong-sik:
Found and patched one real runtime blocker on this PR.

Issue found:
- The original diff added Korean defaults, but existing installs with a persisted `~/.masc/config/excuse_patterns.json` would still load the old English-only 13-pattern file and never reach the new Korean defaults. That matches the operator-truth problem here: repo/config looked fixed, live base-path config stayed stale.

Patch pushed:
- `7b8a15888d fix(anti-rationalization): migrate legacy excuse defaults`
- `load_excuse_patterns` now detects the legacy built-in English-only config and applies the current built-in defaults at runtime.
- The Korean regression test now reproduces the existing-install case by creating an old persisted config file, instead of only testing missing-file fallback.

Local verification:
- `git diff --check`
- `scripts/dune-local.sh build test/test_anti_rationalization_korean_10385.exe test/test_anti_rationalization_coverage.exe test/test_anti_rationalization_gate2_advisory_10113.exe test/test_anti_rationalization_fail_mode.exe`
- `./_build/default/test/test_anti_rationalization_korean_10385.exe` — 10 tests passed, run id `6KCC6DRT`
- `./_build/default/test/test_anti_rationalization_coverage.exe` — all checks passed
- `./_build/default/test/test_anti_rationalization_gate2_advisory_10113.exe` — 4 tests passed, run id `RZ0YVF11`
- `./_build/default/test/test_anti_rationalization_fail_mode.exe` — 5 tests passed, run id `TQNGG6LO`

CI was already mostly green on the prior head; after this push it should re-run on `7b8a15888d`.

PR #10425, opened 2026-04-25
fix(coord-task): label task_claim_next auto-release by reason (#10421)

jeong-sik:
Closes the observability half of #10421 (option D). Pre-fix every auto-release emitted an opaque task_claim_next_auto_release JSON event; day 25 had 20 events but operators couldn't tell polling abuse (claim dropped within seconds, task-056 hot-potatoed 5x) from legitimate cleanup. Add prev_claim_age_seconds + auto_release_reason_for_age helpers (pure, top-level) classifying rapid_replacement (<30s) / stale_replacement (>=30s) / unknown_age. Surface via masc_task_claim_auto_release_total{agent, reason} through the Coord_hooks Atomic-callback pattern (masc_coord stays free of Prometheus dep). 9 new tests pin reason vocabulary + age extraction across all task_status variants + clock-skew floor + metric name. 31 sibling task tests still green. Option A (remove implicit auto-release) and option B (split task_claim_next + task_release_and_claim_next) deferred to follow-up — fleet data from this counter will tell us which to do first.

### DIFF
--- a/lib/coord.ml
+++ b/lib/coord.ml
@@ -258,6 +258,19 @@ let () =
   Atomic.set Coord_hooks.distributed_lock_acquire_failed_fn
     record_distributed_lock_acquire_failed
 
+(* #10421: task_claim_next auto-release observability.
+   [Coord_task_schedule] emits a labelled hook on every auto-
+   release; wire it to the Prometheus counter here so the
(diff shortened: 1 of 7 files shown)


jeong-sik:
Superseded by #10430 (merged). Same issue (#10421).

PR #10300, opened 2026-04-25
fix(cascade): attribute and cool down Kimi resumable failures (#10285)

jeong-sik:
## Goal

Fix #10285's Kimi `resumable_cli_session` failure loop without guessing at undocumented Kimi session-state cleanup.

Observed shape from the issue:

- 97 `resumable_cli_session` events in the last 10,000 logs.
- 5 affected cascades, dominated by `governance_judge`.
- Every affected call first hits Kimi, fails with exit 1, then falls back to the next provider.
- Operators only saw a global error kind, not the per-cascade offender.

## Change

This PR now fixes both the observability gap and the repeated first-tier retry tax.

1. Add `cascade_name` to `masc_oas_error_total`.

   `Resumable_cli_session`, `Cascade_exhausted`, `No_tool_capable_provider`, and `Admission_queue_timeout` use their payload `cascade_name`. Non-cascade-scoped variants emit `cascade_name="unknown"` instead of inventing misleading context.

2. Add terminal structural failure health tracking.

   `Cascade_health_tracker` now has a `Terminal_failure` outcome and `record_terminal_failure`. It behaves like hard quota operationally: one event triggers immediate long cooldown, preserving any longer existing cooldown, and success clears it.

   Default cooldown: `MASC_CASCADE_TERMINAL_FAILURE_COOLDOWN_SEC=3600`
   Deprecated alias: `OAS_CASCADE_TERMINAL_FAILURE_COOLDOWN_SEC`

3. Wire Kimi resumable-session errors to terminal cooldown.

   `run_named` already reclassifies Kimi resume hints into structured `[masc_oas_error] {"kind":"resumable_cli_session" ...}`. After this change, that structured error records `record_terminal_failure` instead of ordinary `record_failure`, so Kimi is skipped on subsequent cascade selections during cooldown instead of failing first on every request.

4. Detect both structured and raw Kimi resume hints.

   `sdk_error_is_resumable_cli_session` detects:

   - structured `Resumable_cli_session`
   - raw `Oas.Error.Api` messages where `Oas.Error.to_string` would otherwise hide the original `kimi exited with code 1` prefix

## Why Not `--no-resume`

[근거] `kimi --help`, checked 2026-04-25 KST, High confidence: current local CLI exposes `--session/--resume -S,-r` and `--continue -C`, but no `--no-resume`. Help points to official docs at `https://moonshotai.github.io/kimi-cli/`.

[...]

PR #10446, opened 2026-04-25
hotfix(main): unbreak build — orphan trust_* call + revert Phase 2a (depends on reverted Phase 1)

jeong-sik:
## Why

main is broken: build fails with two errors after the recent merge train:

1. lib/cascade/cascade_health_tracker.ml:312 — Unbound value apply_trust_failure_locked
2. lib/dashboard_cascade.ml:540 — no field same_fingerprint_count within type Health.provider_info

Cause: #10412 reverted trust_score auto-rotation (Phase 1) but two PRs left
Phase-1-dependent code on main:
- #10300 (Kimi cooldown): added a Terminal_failure branch that called
  apply_trust_failure_locked. The call survived the trust_* removal during
  conflict resolution.
- #10416 (Phase 2a low-trust recommendations): merged AFTER #10412, but the
  whole module reads info.trust_score and info.same_fingerprint_count which
  no longer exist after the revert.

## What

- Drop the orphan apply_trust_failure_locked call in Terminal_failure;
  preserve bump_failure_fp side effect via 'let _ : bool'.
- Revert #10416 entirely — Phase 2a observations are meaningless without
  Phase 1 trust_score; reland once Phase 1 returns (RFC #10409 Phase 2b path).

## Verification

- Reverted commit list:
  - c2f0c3b Revert "feat(dashboard): low-trust operator recommendations (Phase 2a) (#10416)"
  - d50100b fix(cascade): drop orphan apply_trust_failure_locked call after Phase 1 revert

After merge: dune build @install --force should succeed; this unblocks 19
queued PRs that all currently report Lint+Health failure for the same root
cause.


jeong-sik:
Superseded by #10445 which already merged the same fix to main.

PR #10758, opened 2026-04-26
fix: extract crash-persistence + autonomous-queue poll intervals to env, restore CI

jeong-sik:
## Summary

Default branch is currently **RED on `dune build @check`** with two unrelated breakages plus a fragmentation-class issue that aligns with the operator-visibility work started in #10740 (`KeeperWatchdog`).

This PR fixes all three in one slice — they are minor and share the *same* underlying fragmentation pattern: hardcoded values that operators cannot tune without a rebuild, plus an interface/implementation drift left by recent OAS-pin churn.

## What's broken

```
File "lib/dated_jsonl/dated_jsonl.ml", line 1:
Error: The implementation lib/dated_jsonl/dated_jsonl.ml
       does not match the interface lib/dated_jsonl/dated_jsonl.mli:
       The module For_testing is required but not provided

File "lib/oas_event_bridge.ml", line 344, characters 4-29:
344 |   | Oas.Event_bus.TurnReady _ ->
Error (warning 11 [redundant-case]): this match case is unused.
```

Plus two surviving hardcoded poll/drain literals that the user pointed at directly:
- `keeper_crash_persistence.ml:125` — `Eio.Time.sleep clock 2.0`
- `keeper_keepalive.ml:171` — `let autonomous_queue_poll_sec = 0.05`

## Changes

| File | Change |
|------|--------|
| `lib/config/env_config_keeper.ml` | New module `KeeperPollIntervals` (mirrors `KeeperWatchdog` template from #10740). Two knobs with sane defaults + floor clamps. |
| `lib/keeper/keeper_crash_persistence.ml` | Drain interval `2.0` → `KeeperPollIntervals.crash_persistence_drain_sec` |
| `lib/keeper/keeper_keepalive.ml` | `autonomous_queue_poll_sec` constant → `KeeperPollIntervals.autonomous_queue_poll_sec` |
| `lib/dated_jsonl/dated_jsonl.ml` | Add the `For_testing` sub-module the `.mli` declares (`mutex`, `mutex_for_base_dir`, `registry_size`) |
| `lib/oas_event_bridge.ml` | Drop redundant `TurnReady _` arm |
| `test/test_env_config_keeper_poll_intervals.ml` | New — 6 alcotest cases pinning defaults, env override, floor clamps |

## Why bundle these

[...]

PR #10761, opened 2026-04-26
fix(build): restore dated_jsonl For_testing + drop duplicate TurnReady arm

jeong-sik:
## Summary
Two independent build breaks on main blocking local `dune build`:

**1. `lib/dated_jsonl/dated_jsonl.ml`**
Interface declares `module For_testing : sig val mutex : t -> Eio.Mutex.t  val mutex_for_base_dir : string -> Eio.Mutex.t  val registry_size : unit -> int end` (lines 51-62) but the implementation lost the wrapper after the format/rebase commits (`8ae21ab6f`, `d6e74251f`).

```
Error: The implementation lib/dated_jsonl/dated_jsonl.ml
       does not match the interface lib/dated_jsonl/dated_jsonl.mli:
       The module For_testing is required but not provided
```

Re-added as thin wrappers — the underlying file-scope `mutex_registry` and `mutex_for_base_dir` are already present in the `.ml`, so this is purely a missing facade.

**2. `lib/oas_event_bridge.ml:344`**
`| Oas.Event_bus.TurnReady _ -> None` was matched twice (line 211 and again at line 344). Second arm is unreachable, fails under `(env (dev (flags (:standard -warn-error +8))))`:

```
Error (warning 11 [redundant-case]): this match case is unused.
```

Dropped the duplicate.

## Why now
Both blocks `dune build` from a clean checkout. Self-review per `@~/me/agents/best-programmer/AGENT.md`: surgical, ~13 LoC, no behavior change.

## Test plan
- [x] `dune build @check` — green
- [x] `dune build bin/main_eio.exe` — green
- [ ] CI required checks
- [ ] No PR conflict (race-checked: `#10758` open but unrelated scope)

### DIFF
--- a/lib/dated_jsonl/dated_jsonl.ml
+++ b/lib/dated_jsonl/dated_jsonl.ml
@@ -358,3 +358,14 @@ let prune t ~days =
   end
 
 (* Duplicate count_entries removed — canonical definition at line 225 *)
+
+module For_testing = struct
+  let mutex t = Atomic.get t.mutex
+
+  let mutex_for_base_dir base_dir =
+    Atomic.get (mutex_for_base_dir ~base_dir ~injected:None)
+
+  let registry_size () =
+    Stdlib.Mutex.protect mutex_registry_mu (fun () ->
+      Hashtbl.length mutex_registry)
+end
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #10750 (race lost — same two-file fix, same root cause, merged before this PR was created). Local search did not surface #10750 because its title keyword 'unbreak' did not match my 'TurnReady/dated_jsonl/For_testing' grep. Closing as duplicate.

PR #11256, opened 2026-04-27
fix(keeper): kill silent host-sandbox fallback + emit via:host on shell ops

jeong-sik:
## Summary

Two-fix surgical PR targeting the host-shell-via-leak observed in a recent fleet incident
(velvet-hammer stale_keeper_broadcast + nick0cave/ollama-local/qa-king fleet batch
termination). Both pieces are root fixes for the same #11080 sibling-leak class.

### Commit 1 — `via:\"host\"` on every host-side shell op (`keeper_exec_shell.ml`)

`ls`, `cat`, `rg`, `find`, `head`, `tail`, `tree`, `wc` hand-rolled their host-branch
JSON without the `via` discriminator while the docker branches always set
`via:\"docker\"`. Dashboards and downstream LLMs were left to guess whether a result
came from a hardened docker sandbox or an uncontained host process — exactly the
shape PR #11080 closed in `keeper_status_detail.ml`.

Fix:
- `render_completed_process_result` and `render_process_result` inject `via:\"host\"`
  by default; explicit `via` in `~extra` still wins so the wc docker branch keeps
  `via:\"docker\"`.
- 7 hand-rolled host branches (`ls`/`cat`/`rg`/`find`/`head`/`tail`/`tree`) get an
  explicit `\"via\", \`String \"host\"` next to `path`.
- Ops that flow through `run_in_turn_runtime` (`pwd`/`git_status`/`git_log`/`git_diff`/
  `git_worktree`) inherit `via:\"host\"` through the helper change with no per-site
  edits.
- Docker code path is **untouched**: `render_docker_process_result` and the per-op
  docker branches still emit `via:\"docker\"` verbatim.

### Commit 2 — Reject missing `sandbox_profile` in TOML (`keeper_runtime.ml`)

`ensure_keeper_meta` previously called
`apply_default defaults.sandbox_profile default_sandbox_profile`,
silently downgrading any keeper whose TOML omitted `sandbox_profile` to the `Local`
host-process sandbox. Same Unknown→Permissive Default class that #11080 closed at a
sibling site.

Fix:
- Replace the silent fallback with an explicit `Error` from `ensure_keeper_meta`.
- Log a clear, actionable warning naming the keeper, the loaded manifest path, and
  the legal values.
- Leave `Keeper_types_profile.default_sandbox_profile` in place — `keeper_meta_json_parse`,
  the env-override path, and `keeper_turn_up_args` still consume it when reading
  already-persisted runtime meta.
- Persona-only keepers cannot satisfy this check by design (`load_keeper_profile_defaults_from_persona`
  hard-codes `sandbox_profile = None` because personas are not allowed to own
  execution policy). Three persona-only tests in `test_keeper_runtime_config_ssot`
  are migrated to add a thin TOML wrapper that only sets `sandbox_profile` and
  `persona_name`; the persona overlay still drives every other field, so the test
  intent is preserved.

### Commit 3 — Regression test

New `test/test_keeper_shell_via_field.ml` builds a Local-sandboxed keeper, executes
each of the 8 host-branch ops against a tmp playground, and requires the response
JSON to carry `via:\"host\"`. Future ops that forget the discriminator fail this
gate before they ship.

## Self-review (per `agents/best-programmer/AGENT.md`)

- **Goal**: kill the host-sandbox-fallback + via-leak class in one surgical PR;
  the rest of the 5-signal incident (token mismatch, oas_timeout_budget logging,
  require_tool_use violation) is intentionally out of scope.
- **Files modified**:
  - `lib/keeper/keeper_exec_shell.ml` — helper-level `via:\"host\"` + 7 inline tags.
  - `lib/keeper/keeper_runtime.ml` — fail-closed `ensure_keeper_meta`.
  - `test/test_keeper_runtime_config_ssot.ml` — fixture migration (16 TOML
    heredocs gain explicit `sandbox_profile`; 3 persona-only tests add a TOML
    wrapper).
  - `test/test_keeper_shell_via_field.ml` (new) — 8 host-branch via assertions.
  - `test/dune` — register the new test.
- **Local verification**:
  - `dune build --root .` (lib + targeted test exes) green.
  - `test_keeper_shell_via_field` 8/8.
  - `test_keeper_runtime_config_ssot` 24/24.
  - `test_keeper_runtime_toml`, `test_keeper_sandbox_containment`, `test_env_config_sandbox` green.
  - Pre-existing failures on origin/main (no relation to this PR): `test_keeper_shell_containment`,
    `test_keeper_shell_docker_route` (require docker), `test_keeper_runtime_denylist`
    (cascade catalog), `test_cascade_strategy`/`test_dashboard_cascade` (trace_id
    field added by #11228).
- **Not yet verified**:
  - 24h fleet re-measurement of `silent:dashboard_actor_fallback` and
    `oas_timeout_budget` (Track 2 inputs).
  - Real-keeper runtime check that operator-edited `~/.masc/config/keepers/*.toml`
    files still load cleanly (all 14 in this dev box already declare
    `sandbox_profile`, but other deploys may not).
- **Risk**: any in-the-wild keeper whose TOML omits `sandbox_profile` will refuse
  to boot after this lands. The clear actionable warn message gives the migration
  path: `Add e.g. \`sandbox_profile = \"docker\"\` to the keeper TOML.`

## Plan
[planning/claude-plans/me-workspace-yousleepwhen-masc-mcp-sile-purring-raven.md](../blob/main/planning/claude-plans/me-workspace-yousleepwhen-masc-mcp-sile-purring-raven.md)

### DIFF
--- a/lib/keeper/keeper_exec_shell.ml
+++ b/lib/keeper/keeper_exec_shell.ml
@@ -1389,6 +1389,10 @@ let handle_keeper_shell
                match cwd with
                | Some dir -> `String dir
                | None -> `Null );
+             (* host execution path: route discriminator must be present so
+                the dashboard / LLM cannot mistake a host-side run for a
+                docker-sandboxed run (#11080 sibling sweep). *)
+             "via", `String "host";
            ] @ insight_extra)
          ~status:st
          ~output:out
@@ -1427,6 +1431,13 @@ let handle_keeper_shell
           List.map Masc_exec.Bash_history.failure_pattern_to_json patterns)
       ]
     in
+    (* Caller-supplied [extra] may already declare ["via"] (e.g. wc docker
+       branch); only inject the default ["via", "host"] when absent so the
+       docker route's explicit ["via", "docker"] still wins. *)
+    let extra_with_via =
+      if List.exists (fun (k, _) -> k = "via") extra then extra
+      else ("via", `String "host") :: extra
+    in
     Yojson.Safe.to_string
       (Exec_core.process_result_json
          ~artifact_policy:Exec_core.Inline_only
@@ -1440,7 +1451,7 @@ let handle_keeper_shell
                match cwd with
                | Some dir -> `String dir
                | None -> `Null );
-           ] @ extra @ insight_extra)
+           ] @ extra_with_via @ insight_extra)
          ~status:st
          ~output:out
          ())
@@ -1606,6 +1617,7 @@ let handle_keeper_shell
                [ "ok", `Bool (st = Unix.WEXITED 0)
                ; "op", `String op
                ; "path", `String target
+               ; "via", `String "host"
                ; "status", Keeper_alerting_path.process_status_to_json st
                ; "entries", lines_to_json ~limit out
                ]))
@@ -1656,6 +1668,7 @@ let handle_keeper_shell
                [ "ok", `Bool (st = Unix.WEXITED 0)
                ; "op", `String op
                ; "path", `String target
+               ; "via", `String "host"
                ; "status", Keeper_alerting_path.process_status_to_json st
                ; "truncated", `Bool (String.length out > max_bytes)
                ; "content", `String body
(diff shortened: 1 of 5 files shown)


jeong-sik:
Track 2 follow-up tracked in #11266 (3 remaining signals from the same incident).

PR #12035, opened 2026-04-29
fix(specs): refresh TLA index drift

jeong-sik:
## Summary\n- Regenerate specs/INDEX.md from the current TLA spec tree\n- Make scripts/gen-tla-index.sh tolerate specs with zero cfg files under macOS Bash 3.2 + set -u\n- Unblock tla-spec-index drift failures across the current PR queue\n\n## Validation\n- Not run locally; change generated with bash scripts/gen-tla-index.sh > specs/INDEX.md\n\n## Context\n- Recent PR checks show specs/INDEX.md stale: spec counts moved from 81/154 to 83/158 after new boundary specs.


jeong-sik:
Closing as obsolete after the latest main merges. Current main regenerates specs/INDEX.md cleanly when ignoring the Generated header, and rebasing this PR leaves only that ignored timestamp line, so there is no substantive diff left to merge.

PR #12246, opened 2026-04-30
fix(keeper): expose rewrite_docker_host_paths_to_container in mli

jeong-sik:
## Summary
- `rewrite_docker_host_paths_to_container` existed in `keeper_shell_shared.ml` but was missing from its `.mli`.
- `keeper_exec_shell.ml` includes `Keeper_shell_shared` and re-exports the value through its own `.mli`, causing a signature mismatch / build break on main.

## Fix
Add the missing `val` declaration to `keeper_shell_shared.mli`.

## Test plan
- [x] `dune build --root .` passes in worktree.

### DIFF
--- a/lib/keeper/keeper_shell_shared.mli
+++ b/lib/keeper/keeper_shell_shared.mli
@@ -148,6 +148,15 @@ val rewrite_turn_runtime_paths_to_host :
     its host playground absolute path so the LLM-facing output
     references real host paths. *)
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as duplicate: the same keeper_shell_shared.mli export is already present on current main via the merged OAS pin path. Current main and this branch have no diff for lib/keeper/keeper_shell_shared.mli.

PR #12252, opened 2026-04-30
fix(specs): make TLA index shallow-stable

jeong-sik:
## Summary
- replace TLA index Last Modified dates with content-derived IDs
- regenerate specs/INDEX.md from the updated generator
- prevent shallow/grafted checkouts from producing CI-only tla-spec-index drift

## Validation
- bash -n scripts/gen-tla-index.sh
- scripts/gen-tla-index.sh output matches committed specs/INDEX.md, ignoring Generated header
- same generator output matches between local shallow worktree and full-history clone, ignoring Generated header
- git diff --check

### DIFF
--- a/scripts/gen-tla-index.sh
+++ b/scripts/gen-tla-index.sh
@@ -108,12 +108,13 @@ extract_invariants() {
   ' "$cfg"
 }
 
-# Last commit date (ISO short) for a file. Falls back to "-".
-last_modified() {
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #12250. main now generates specs/INDEX.md with a stable git hash-object based source hash, so this draft would only rename the same fingerprint concept and reintroduce conflicts.

PR #12307, opened 2026-04-30
fix(keeper): configure stream idle timeout

jeong-sik:
## Summary

- Add `turn.stream_idle_timeout_sec` / `MASC_KEEPER_STREAM_IDLE_TIMEOUT_SEC`.
- Resolve and freeze stream idle timeout through keeper runtime config.
- Pass the runtime value into keeper agent turns, clamped to the total OAS timeout when shorter.
- Remove stale comments that claimed `max_turns` defaulted to 50 in `keeper_agent_run`.

## Product impact

The stream idle timeout is now a visible runtime knob instead of a hardcoded keeper run-path constant. Operators can tune streaming gap detection without editing code, while the default stays at 120 seconds.

## Evidence

- `scripts/dune-local.sh build test/test_keeper_runtime_toml.exe test/test_keeper_unified.exe`
- `./_build/default/test/test_keeper_runtime_toml.exe --compact --quick-tests`
- `./_build/default/test/test_keeper_unified.exe --compact --quick-tests`
- `git diff --check`

## Review evidence

Focused tests cover TOML/env mapping, default value, source freezing, total-timeout clamp behavior, and the existing keeper unified runtime defaults.

## Linked issue

None.

### DIFF
--- a/docs/BOOT-ENV-STATE-INVENTORY.md
+++ b/docs/BOOT-ENV-STATE-INVENTORY.md
@@ -120,7 +120,7 @@ Parse errors log a warning and fall back to env defaults.
 Legacy compatibility names that are no longer read by the unified turn path
 are intentionally excluded from this TOML surface.
 
-**Sections** (68 knobs total):
+**Sections** (69 knobs total):
 
 | Section | Count | Key examples |
 | --- | --- | --- |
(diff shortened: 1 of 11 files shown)


jeong-sik:
CI note: the current Lint and OCaml Structure Ratchet failures are inherited from `main`, not introduced by this branch. Local `main` reproduces both `scripts/lint/yojson-option-default.py` and `scripts/ocaml-structure-ratchet.sh` with the same signatures. Tracked in #12309.

PR #12471, opened 2026-05-01
refactor: adopt Jane Street Base in shared_types

jeong-sik:
Phase 1: Pilot adoption of Jane Street Base in lib/shared_types. Replaced polymorphic compares and deprecated stdlib usages.


jeong-sik:
Closing as invalid/stale queue noise. The title says Base adoption, but the diff is a mix of superseded Tool_coord test patches plus partial OAS floor metadata, without updating scripts/oas-agent-sdk-pin.sh. The verified replacement for the test repair is #12465; the OAS pin drift needs a focused pin repair.

PR #12472, opened 2026-05-01
chore: bump and pin oas to 0.187.6

jeong-sik:


jeong-sik:
Superseded by #12473, which carries the same OAS 0.187.6 runtime pin plus the matching agent_sdk floor and generated manual update. Closing this script-only duplicate because it is already DIRTY and incomplete.

PR #12482, opened 2026-05-01
chore: bump agent_sdk floor to 0.187.6

jeong-sik:


jeong-sik:
Superseded by #12473, which includes this agent_sdk floor bump together with the matching OAS pin script and generated docs. Closing the floor-only duplicate to keep the OAS bump queue single-source.

PR #12487, opened 2026-05-01
fix(test): unblock main build after Tool_coord result record

jeong-sik:
## Summary\n- Open Coord_types in goal dashboard/tool tests so record fields resolve after the shared result type move.\n- Open Tool_coord and update Tool_coord.dispatch test patterns to match the tool_result record shape.\n- Folded the duplicate #12489 test fixes here so CI sees the full main-build unblock in one PR.\n\n## Verification\n- git diff --check origin/main...HEAD\n- Local focused Dune build not run: /tmp/me-dune-local.lock is held by an OAS @check job; CI is the active verification surface.

### DIFF
--- a/test/test_dashboard_goals.ml
+++ b/test/test_dashboard_goals.ml
@@ -1,5 +1,6 @@
 open Alcotest
 open Masc_mcp
+open Coord_types
 open Tool_coord
 open Yojson.Safe.Util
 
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing as duplicate: #12492 now carries the complete main-build unblock on top of latest main, including the Tool_coord record test updates plus the OAS pin metadata fix after upstream OAS corrected agent_sdk.opam to 0.187.6.

PR #12490, opened 2026-05-01
fix: add missing open Tool_coord in tests

jeong-sik:
Fixes main branch build break. Adds missing open Tool_coord in test_goal_upsert_fsm_bypass_10247.ml and test_tool_coord_coverage.ml after Tool_coord.tool_result changed to record type.


jeong-sik:
Closing as duplicate: the Tool_coord record test alignment is covered by #12492, which also includes the current OAS pin metadata correction. Keeping one PR avoids competing fixes for the same main-build break.

PR #12491, opened 2026-05-01
fix(dashboard): update filter chip test expectations

jeong-sik:
## Summary\n- Update OperationsPanel filter-chip test to include the current safety chip.\n- Update RuntimePanel filter-chip test to include cost and inspector chips in the current order.\n\n## Verification\n- pnpm --dir dashboard install --frozen-lockfile --offline\n- pnpm --dir dashboard typecheck\n- git diff --check\n\nNote: targeted Vitest for these two files timed out locally under Homebrew Node 25 before reaching assertions; CI on Node 22 is the authoritative dashboard test surface and previously failed only on these expectations.


jeong-sik:
@copilot resolve the merge conflicts in this pull request

PR #12493, opened 2026-05-01
fix(ci): tolerate missing CI Gate on synchronize events in draft guard

jeong-sik:
## Problem

After rebasing a PR and force-pushing (which triggers a `synchronize` event), the Draft Auto-Merge Guard workflow checks for a completed \"CI Gate\" check run on the new HEAD SHA. However, the new CI Gate check run hasn't been created yet by the time the guard job executes, causing the guard to fail with:

```
CI Gate is not completed successfully for head <sha>: missing
```

This incorrectly disables auto-merge on PRs that were previously approved and had auto-merge enabled.

## Root Cause

The guard already had a tolerance for this race condition on `auto_merge_enabled` events (where the gate may still be starting), but `synchronize` events after a force-push have the same window where the old gate is invalidated and the new one hasn't started yet.

## Fix

Extend the existing `auto_merge_enabled` tolerance to also cover `synchronize` events. The guard still blocks when:
- The gate is known to exist but failed
- The gate is missing on events other than `auto_merge_enabled` or `synchronize`

## References

- Related: PRs #12449–#12452, #12488 (all lost auto-merge after rebase due to this race)
- Workflow: `.github/workflows/pr-automation.yml`

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -293,10 +293,11 @@ jobs:
                   ? `${latestCiGate.status}/${latestCiGate.conclusion || "none"} ${latestCiGate.html_url}`
                   : "missing";
                 // CI Gate may not exist yet immediately after auto-merge is enabled
-                // (CI workflow is still starting). Only block when the gate is known
+                // (CI workflow is still starting) or after a force-push (synchronize)
+                // that invalidates the previous gate. Only block when the gate is known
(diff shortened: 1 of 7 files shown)


jeong-sik:
Closing this draft because it mixes the #12488 Base refactor with the PR automation synchronize-race fix. The workflow hunk has been folded into #12494, which keeps the CI queue/guard cleanup scoped to workflow files only.

PR #12546, opened 2026-05-01
keeper: add masc_keeper_docker_runtime_discarded_total metric

jeong-sik:
## Summary
Add a Prometheus counter `masc_keeper_docker_runtime_discarded_total` to track when a cached `turn_sandbox_runtime` is discarded because `network_mode` is not `Network_none`.

## Context
When a keeper turn completes, the Docker container runtime may be cached in `turn_sandbox_runtime` for reuse on the next turn. However, if `network_mode` is not `Network_none`, the runtime is silently discarded and a cold-start container is spun up instead. This metric makes that cold-start path visible for observability.

## Changes
- `lib/keeper/keeper_shell_docker.ml`: increment counter in the discard path

## Test Plan
- [x] `dune build lib/keeper/` compiles
- [ ] CI checks pass

### DIFF
--- a/lib/keeper/keeper_shell_docker.ml
+++ b/lib/keeper/keeper_shell_docker.ml
@@ -646,6 +646,13 @@ let run_docker_hardened_bash
                 ("output", `String out);
               ] @ gh_exit_class_field ~cmd ~status:st ~output:out)))
     | _ ->
+      (match turn_sandbox_runtime with
+       | Some _ ->
+         Prometheus.inc_counter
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):
LGTM, 다만 `turn_sandbox_runtime = Some _` 일 때 `masc_keeper_docker_runtime_discarded_total` 메트릭이 의도대로 증가하는지 검증하는 테스트 케이스(Coverage) 보강이 필요해 보입니다.

PR #12568, opened 2026-05-01
ci: compare health ratchets against baseline ref

jeong-sik:
## Summary
- compare Health unsafe-pattern ratchets against the actual baseline ref tree instead of stale fallback JSON
- fetch PR base or push-before SHA for Health ratchet baselines
- refresh fallback health-baseline counts to current main for manual and no-baseline runs

## Why
Main CI run 25207078935 failed Health because push CI had no baseline ref and compared current main against stale .ci/health-baseline.json values such as lib_list_hd 0, while current main already has 4 pre-existing occurrences. That makes Health consume queue and fail after merge without pointing at a real new regression.

## Validation
- bash -n scripts/health_snapshot.sh scripts/ml_line_cap_audit.sh
- YAML parse for .github/workflows/ci.yml
- python3 -m json.tool .ci/health-baseline.json
- bash scripts/health_snapshot.sh --skip-build --json-out /tmp/masc-health-baseline-ref.json --baseline-ref HEAD^ --fail-on-lib-regression
- bash scripts/health_snapshot.sh --skip-build --json-out /tmp/masc-health-baseline-file.json --baseline-file .ci/health-baseline.json --fail-on-lib-regression
- git diff --check

PR #12572, opened 2026-05-01
ci: ignore stale CI Gate during active PR checks

jeong-sik:
## Summary
- keep Draft Auto-Merge Guard, but stop treating a stale CI Gate failure as current while the same head still has active checks
- preserve the hard block when CI Gate is failed and no replacement checks are running
- add log context for active head checks when the startup race is skipped

## Why
While processing #12568, enabling auto-merge on the ready PR hit an old draft-run CI Gate failure for the same head and Draft Auto-Merge Guard disabled auto-merge even though the replacement CI run was active. This is the confusing failure mode behind the 'Draft Auto-Merge Guard failing after 4s' signal.

## Verification

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -300,6 +300,19 @@ jobs:
                 // for this startup race.
                 const ciGateStartupRaceAction =
                   action === "auto_merge_enabled" || action === "synchronize";
+                const ciWorkflowActiveCheckNames = new Set([
+                  "PR Sync Check",
+                  "Detect Changed Surfaces",
+                  "Keeper Cwd Leak Gate",
+                  "Meta Guards",
+                  "Health",
+                  "Build and Test",
+                  "Lint",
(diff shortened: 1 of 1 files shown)


jeong-sik:
Current status for head c9fa465:

- `git diff --check origin/main...HEAD` passes locally.
- The red `CI Gate` is the expected draft-agent policy gate: the log says this draft `codex/...` branch lacks `human-approved-ready`.
- All other CI surfaces for this draft are skipped/pass as expected for the current draft state.

Keep this draft until a human decides the PR automation race fix should move forward with the ready bypass.

jeong-sik:
Review follow-up: narrowed the stale CI Gate startup-race bypass to active checks from the CI workflow only. This avoids treating unrelated active checks, such as code-review automation, as evidence that a replacement CI Gate is still running.\n\nLocal validation:\n- YAML parse for .github/workflows/pr-automation.yml\n- git diff --check

PR #12581, opened 2026-05-01
fix(keeper): bound circuit breaker cooling

jeong-sik:
## Summary
- Track the latest circuit-breaker trip time and expose a bounded cooling state.
- Close cooling immediately on successful tool calls and auto-expire observer state after the cooling window.
- Preserve legacy snapshot classification when no explicit display_state is present.

## Validation
- ./scripts/dune-local.sh exec test/test_circuit_breaker.exe
- git diff --check


jeong-sik:
Closing as duplicate. The same commit was already merged via #12578.

PR #12603, opened 2026-05-01
fix(dashboard): restore overview data contracts

jeong-sik:
## Summary
- restore Overview helper exports and current dashboard data-field usage after the post-merge refactor
- move Lab legacy highlight to the current OperatorAttentionItem shape
- keep funnel, mission, keeper, and alert helpers aligned with existing overview tests

## Verification
- pnpm --dir dashboard exec vitest run src/components/overview/overview.test.ts

### DIFF
--- a/dashboard/src/components/lab.ts
+++ b/dashboard/src/components/lab.ts
@@ -32,8 +32,10 @@ function Highlight({ attention }: { attention: OperatorAttentionItem | null }) {
       <div class="flex items-center gap-3">
         <span class="flex-shrink-0 flex h-2 w-2 rounded-full bg-[var(--color-status-err)] animate-pulse" />
         <div class="flex-1 min-w-0">
-          <p class="text-xs font-medium truncate ${severityToneClass(attention.severity)}">${attention.title}</p>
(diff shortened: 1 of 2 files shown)


jeong-sik:
Blocking review note: I cannot submit a formal "request changes" review on this owner-authored PR, but I would not merge this as-is.

1. SectionCard call sites now pass title, right, and data-testid, while dashboard/src/components/common/card.ts only reads label/class/variant/children. The moved Highlight and Overview cards can render with empty headers, lose right-side status content, and silently drop test ids. Please either update these call sites to the actual label API or make SectionCard backward-compatible and add coverage.

2. The keeper lifeline still renders bpm=72. That is fake telemetry; Overview should not display a made-up operational metric. Omit BPM unless the card has a real heartbeat-derived value.

#12602 includes both fixes plus targeted card.test.ts coverage and the same local Dashboard typecheck/overview test verification.

jeong-sik:
Closing as duplicate: #12602 has already merged the main dashboard overview contract repair. This PR carried the same deterministic type-contract fix and its CI is now superseded by the post-merge main run.

PR #12604, opened 2026-05-01
fix: restore overview type contracts

jeong-sik:
## Summary
- restore Overview helper contracts to the current dashboard API types
- use koreanName, required_count, member_names, completed_at, and last_heartbeat instead of removed fields
- update the legacy Lab highlight to read OperatorAttentionItem.summary

## Validation
- git diff --check
- pnpm run typecheck
- pnpm exec vitest run --config vitest.config.ts src/components/overview/overview.test.ts --no-file-parallelism --maxWorkers=1

## Context
Main is red after #12601 because Dashboard typecheck fails on stale Overview field names and non-exported helpers.

### DIFF
--- a/dashboard/src/components/lab.ts
+++ b/dashboard/src/components/lab.ts
@@ -32,8 +32,10 @@ function Highlight({ attention }: { attention: OperatorAttentionItem | null }) {
       <div class="flex items-center gap-3">
         <span class="flex-shrink-0 flex h-2 w-2 rounded-full bg-[var(--color-status-err)] animate-pulse" />
         <div class="flex-1 min-w-0">
-          <p class="text-xs font-medium truncate ${severityToneClass(attention.severity)}">${attention.title}</p>
-          <p class="text-2xs text-[var(--color-fg-secondary)] truncate">${attention.message}</p>
+          <p class="text-xs font-medium truncate ${severityToneClass(attention.severity)}">${attention.summary}</p>
+          <p class="text-2xs text-[var(--color-fg-secondary)] truncate">
+            ${attention.target_id ?? attention.target_type}
+          </p>
         </div>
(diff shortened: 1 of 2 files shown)


jeong-sik:
Root cause: #12601 reintroduced stale Overview contracts: display_name, target_value, goal_id, members, last_seen_at. Current dashboard types expose koreanName, required_count, goal, member_names, last_heartbeat.

Fix in this PR restores those contracts and updates the Lab legacy highlight to use OperatorAttentionItem.summary.

Validation run locally:
- git diff --check
- pnpm run typecheck
- pnpm exec vitest run --config vitest.config.ts src/components/overview/overview.test.ts --no-file-parallelism --maxWorkers=1 (46 passed)

Current GitHub status is policy-blocked only: the PR is draft and the agent draft gate requires human-approved-ready before the required gate can go green.

jeong-sik:
Closing as duplicate: #12602 has already merged the dashboard overview type-contract repair. Keeping this draft open would only leave redundant red/queued checks.

jeong-sik:
Review note: this is directionally the same main-CI repair, but it is still incomplete compared with #12602.

Remaining issues:
- SectionCard compatibility is not addressed. The current Overview/Lab call sites pass title/right/data-testid, but common/card.ts only consumes label/class/variant/children, so headers/right slots/test ids can silently disappear.
- Keeper lifeline still renders bpm=72, which is made-up telemetry. Please omit BPM unless it is derived from real runtime data.
- The secondary keeper TimeAgo is rendered with last_heartbeat directly; when absent, this can pass undefined to TimeAgo. #12602 guards that path.

I would keep this draft blocked and use #12602 as the fuller fix, or fold the primitive/test coverage from #12602 into this branch.

PR #12634, opened 2026-05-01
test: tighten keeper cleanup exception guard

jeong-sik:
## Summary
- tighten the keeper sandbox cleanup source guard so it checks the cleanup wrapper's backtrace-preserving re-raise path instead of any unrelated `raise e` in the file
- add a small ordered-substring helper for the source-level assertion

## Verification
- `scripts/dune-local.sh build test/test_keeper_agent_run_sandbox_source.exe`
- `./_build/default/test/test_keeper_agent_run_sandbox_source.exe`

Follow-up for Copilot thread on #12626: PRRT_kwDOQ7_DYc5-_ljK

### DIFF
--- a/test/test_keeper_agent_run_sandbox_source.ml
+++ b/test/test_keeper_agent_run_sandbox_source.ml
@@ -41,6 +41,32 @@ let count_occurrences ~needle haystack =
 
 let contains ~needle haystack = count_occurrences ~needle haystack > 0
 
+let find_from ~needle ~start haystack =
+  let nlen = String.length needle in
+  let hlen = String.length haystack in
+  let rec loop pos =
+    if nlen = 0
+    then Some pos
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):
Reviewed the follow-up test-only PR. The change correctly replaces the broad `raise e` source check with an ordered assertion over the cleanup wrapper path: capture raw backtrace, run cleanup, then `Printexc.raise_with_backtrace e backtrace`. I re-ran the focused verification locally: `scripts/dune-local.sh build test/test_keeper_agent_run_sandbox_source.exe`, `./_build/default/test/test_keeper_agent_run_sandbox_source.exe` (4 tests), and `git diff --check origin/main..HEAD`; all passed. No code changes requested from my side. This is still draft and lacks `human-approved-ready`, so merge remains gated.

PR #12645, opened 2026-05-01
fix(keeper): reuse turn sandbox runtime across cwd

jeong-sik:
## Summary

- Reuse the keeper turn Docker sandbox runtime across different cwd values inside the same playground.
- Keep separate runtimes for outside-vs-inside playground execution and for different network modes.
- Add a regression test proving nested playground cwd dispatch reuses the same turn sandbox runtime.

## Root Cause

The turn sandbox factory cached Docker runtimes by `(in_playground, cwd)`. A single keeper turn can dispatch tools with multiple cwd values under the same playground root, while `Keeper_turn_sandbox_runtime` already maps each host cwd into the container. That meant a keeper such as `nick0cave` could start more than one `masc-keeper-turn-...` container in one turn before stale cleanup had a chance to reclaim anything.

## Operator Impact

This should stop same-turn duplicate Docker containers caused by cwd drift. Existing stale cleanup still covers stopped containers, expired TTL containers, and owner-pid-dead containers.

## Validation

- `git diff --check`
- `scripts/dune-local.sh build test/test_operator_control.exe`
- `./_build/default/test/test_operator_control.exe test operator 45,46 --show-errors --compact`

### DIFF
--- a/lib/keeper/keeper_sandbox_factory.ml
+++ b/lib/keeper/keeper_sandbox_factory.ml
@@ -46,30 +46,31 @@ let in_playground_of_cwd (t : t) ~cwd =
 let resolve (t : t) ~cwd =
   with_lock t (fun () ->
     let in_playground = in_playground_of_cwd t ~cwd in
-    let cwd_norm = normalize cwd in
-    let key = (in_playground, cwd_norm) in
-    match Hashtbl.find_opt t.cache key with
-    | Some r -> Some r
-    | None ->
-      let (effective_profile, effective_network) =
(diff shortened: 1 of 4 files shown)


jeong-sik:
CI status note: `CI Gate` is red by the agent draft policy, not by a code/test failure. The gate log reports this PR lacks `human-approved-ready`, which is expected for an `agent-pr` draft until a human explicitly approves it. `PR Sync Check`, `PR Live Gate`, and `Draft Auto-Merge Guard` are passing.

PR #12682, opened 2026-05-02
fix: feed queued board stimuli into keeper turns

### LINKED ISSUE #12680: Wire queued board stimuli into keeper turn context
## Problem
Board writes can wake a Keeper by enqueueing an Event Layer stimulus, but `run_keepalive_unified_turn` only logs the dequeued stimulus before building the world observation. If cursor-based board collection misses that same post or comment, the Keeper can wake and still receive a prompt with no board activity, so the reaction chain appears to stop.

## Expected
A dequeued `board_signal` stimulus is promoted into `pending_board_events` for the same turn, deduped against cursor-collected events, and therefore participates in `Board_event_pending` and prompt construction.

## Scope
- Parse queued board-signal payloads into `Keeper_world_observation.pending_board_event`.
- Merge the queued event into the heartbeat observation when cursor collection did not already include it.
- Add focused regression coverage.

## MASC Tracking
Goal: `goal-world-reaction-liveness`
Task: `task-132`

Related: #12664, #12666, #12677

jeong-sik:
## Summary
- parse `board_signal` Event Layer stimuli back into `pending_board_event` records
- merge the dequeued board stimulus into the same heartbeat observation when cursor collection did not already include that post
- keep duplicate cursor-collected events deduped by post id
- add a focused regression for queued board stimulus -> pending board activity -> runnable keeper cycle

## Why it matters
The board hook can wake a Keeper and enqueue the authoritative stimulus payload, but the turn path previously only logged the dequeued stimulus before building the prompt. If cursor-based board collection missed the post/comment, the Keeper woke up with no board activity in context and could immediately skip as `no_signal`. This patch makes the wake payload part of the atomic reaction turn.

## Verification
- `git diff --check`
- `./_build/default/test/test_keeper_unified.exe test world_observation 3`

## Known unrelated local failure
- `./_build/default/test/test_keeper_unified.exe` currently fails at existing case `transient_network_error 44 bounded OAS timeout reserves degraded retry` on this machine: expected `592.5`, received `300`. The new world_observation regression passed before that failure and passes when filtered.

Fixes #12680

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -355,24 +355,44 @@ let run_keepalive_unified_turn
          a real runtime transition: a stimulus that triggered the
          heartbeat override (PR-C2 #12412) is now actually consumed
          here. *)
-      (match
-         Keeper_registry.dequeue_event
-           ~base_path:ctx.config.base_path
-           meta_after_triage.name
-       with
-       | None -> ()
-       | Some stim ->
-           let urgency_str =
-             match stim.urgency with
-             | Keeper_event_queue.Immediate -> "immediate"
-             | Keeper_event_queue.Normal -> "normal"
-             | Keeper_event_queue.Low -> "low"
-           in
-           Log.Keeper.info
-             "turn entry: consumed stimulus stimulus_id=%s urgency=%s payload_len=%d (keeper=%s)"
-             stim.post_id urgency_str
-             (String.length stim.payload)
-             meta_after_triage.name);
+      let queued_board_event =
+        match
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Diff review: I do not see a blocking issue. The queued board_signal stimulus is parsed back into the existing pending_board_event shape, duplicate cursor-collected posts are preserved by post_id dedupe, and the new regression covers the no-cursor fallback path that previously skipped as no_signal. I cannot approve this PR because GitHub treats it as my own PR. Please keep it draft until the branch runs the relevant OCaml checks or is promoted through the normal heavy CI path; current PR checks only ran draft guards/sync.

Issue #12675, opened 2026-05-02
fix: timeout budget drift — pre-model timeouts at attempt=0 waste turns


jeong-sik:
## Problem

resolve_bounded_oas_timeout_budget_with_turn_budget computes OAS timeout from wall-clock budget that includes time spent in previous cascade attempts. By the time the 2nd or 3rd cascade attempt starts, the remaining budget is so small that the OAS call times out before the model even begins processing.

## Root Cause

In keeper_unified_turn.ml, remaining_turn_budget_s is computed as wall-clock subtraction including time spent in cascade routing, phase gating, semaphore wait, and previous failed cascade attempts.

When cascades fail slowly (timeout then 30s wait then fallback), the remaining budget shrinks dramatically. After 2 cascade timeouts (60s each), remaining approaches 0, and bounded rounds to None — but only after the 3rd attempt has already been dispatched.

## Evidence

- oas_timeout_budget_observation_reasons logs show timeout at attempt=0 (before model contact)
- TurnBudgetExhausted stop reason appears with turns_used=1 when max_turns=10
- Timeout resolution source field shows degraded_retry_budget for pre-model timeouts
- Wall-clock drift correlates with cascade routing latency

## Impact

- 2nd/3rd cascade attempts get zero useful timeout budget
- Turns end with TurnBudgetExhausted despite only 1 actual model call
- Degraded retry logic fires but has no budget to work with
- Keepers appear to try multiple cascades but actually accomplish nothing

## Fix Approach

1. Separate routing budget from execution budget — deduct fixed routing overhead per cascade attempt, not wall-clock
2. Minimum execution budget gate — if remaining less than min_execution_budget_s, skip attempt entirely instead of dispatching with near-zero timeout
3. Per-attempt budget reset — each cascade attempt gets a fresh execution budget, capped by total turn budget
4. Add budget_source telemetry to distinguish no budget left from model timeout

## Verification

- [ ] No TurnBudgetExhausted with turns_used less than max_turns - 1
- [ ] Timeout resolution shows source=per_attempt not source=degraded_retry_budget for fresh attempts
- [ ] 2nd cascade attempt receives usable timeout budget (greater than 30s)
- [ ] Budget telemetry distinguishes routing vs execution time

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

jeong-sik:
## Code Analysis — 2026-05-02

[...]

PR #12702, opened 2026-05-02
fix: per-attempt retry budget for cascade timeout drift (#12675)

jeong-sik:
## Summary
- Cascade retries after long first-attempt timeouts arrived with near-zero wall-clock remaining, causing `TurnBudgetExhausted` with `turns_used=1` before the model could respond
- Add `is_retry` flag to budget resolution: when true, the retry gets a fresh minimum execution budget (`max(min_budget, remaining)`) capped by adaptive timeout, bypassing the guard+reserve logic that only makes sense for initial attempts
- Propagate `is_retry` through the degraded retry budget gate so retry eligibility check also uses per-attempt budget semantics

## Root Cause
`remaining_turn_budget_s` is a pure wall-clock subtraction (`turn_deadline -. now`). After a 60s first attempt timeout, retries arrive with <60s remaining. The existing guard (`oas_timeout_guard_sec = 15s`) + reserve logic correctly refuses to schedule another call — but the retry should get a fresh budget window, not inherit the wall-clock deficit.

## Test Plan
- [x] 4 new regression tests:
  - `per-attempt retry budget with near-zero remaining` — verifies retry gets ≥15s budget even with 3s remaining
  - `per-attempt retry budget capped by healthy remaining` — verifies cap works when remaining is healthy
  - `non-retry still refuses tiny budget` — verifies `is_retry=false` behavior unchanged
  - `degraded retry gate allows retry with tiny remaining` — verifies budget gate propagation
- [x] `dune build @check` passes
- [x] Pre-existing test failure (#44 bounded OAS timeout reserve) confirmed unrelated to this change

### DIFF
--- a/lib/keeper/keeper_turn_cascade_budget.ml
+++ b/lib/keeper/keeper_turn_cascade_budget.ml
@@ -165,25 +165,30 @@ let resolve_bounded_oas_timeout_budget_with_turn_budget
        near-zero remaining and the OAS call times out before the model
        even begins processing (TurnBudgetExhausted with turns_used=1).
 
-       GitHub #12675. *)
-    let effective_timeout_sec =
-      Float.min adaptive_timeout_sec (Float.max min_oas_timeout_budget_sec remaining_turn_budget_s)
-    in
-    let source =
-      match runtime.oas_timeout_override_sec.value with
-      | Some _ -> "override_per_attempt_retry"
-      | None -> "adaptive_per_attempt_retry"
-    in
-    Some
-      {
-        effective_timeout_sec;
-        adaptive_timeout_sec;
-        keeper_turn_timeout_sec = runtime.turn_timeout_sec.value;
-        remaining_turn_budget_sec = Float.max 0.0 remaining_turn_budget_s;
-        estimated_input_tokens = max 0 estimated_input_tokens;
-        max_turns;
-        source;
-      }
+       Hard gate: if the outer turn wall-clock is already exhausted the
+       OAS call would be cancelled immediately — return None so callers
+       see a clean budget failure rather than a noisy retry/rotation. *)
+    if remaining_turn_budget_s <= 0.0 then None
+    else begin
+      let effective_timeout_sec =
+        Float.min adaptive_timeout_sec (Float.max min_oas_timeout_budget_sec remaining_turn_budget_s)
+      in
+      let source =
+        match runtime.oas_timeout_override_sec.value with
+        | Some _ -> "override_per_attempt_retry"
+        | None -> "adaptive_per_attempt_retry"
+      in
+      Some
+        {
+          effective_timeout_sec;
+          adaptive_timeout_sec;
+          keeper_turn_timeout_sec = runtime.turn_timeout_sec.value;
+          remaining_turn_budget_sec = remaining_turn_budget_s;
+          estimated_input_tokens = max 0 estimated_input_tokens;
+          max_turns;
+          source;
+        }
+    end
   end else begin
     let usable_budget = remaining_turn_budget_s -. oas_timeout_guard_sec in
     if usable_budget < min_oas_timeout_budget_sec
@@ -232,7 +237,7 @@ let resolve_bounded_oas_timeout_budget_with_turn_budget
           effective_timeout_sec;
           adaptive_timeout_sec;
           keeper_turn_timeout_sec = runtime.turn_timeout_sec.value;
-          remaining_turn_budget_sec = usable_budget;
+          remaining_turn_budget_sec = remaining_turn_budget_s;
           estimated_input_tokens = max 0 estimated_input_tokens;
           max_turns;
           source;
@@ -286,7 +291,6 @@ type degraded_retry_budget_decision =
 
 let next_fail_open_cascade_for_turn_with_budget
     ?rotation_cascades
-    ~(is_retry : bool)
     ~(base_cascade : string)
     ~(effective_cascade : string)
     ~(tool_requirement : string)
@@ -303,9 +307,11 @@ let next_fail_open_cascade_for_turn_with_budget
   with
   | None -> No_degraded_retry
   | Some retry ->
+      (* The candidate is always a retry, so use per-attempt budget semantics
+         regardless of whether the current attempt was itself a retry. *)
       if
         oas_retry_budget_available_for_turn
-          ~is_retry ~estimated_input_tokens ~max_turns
+          ~is_retry:true ~estimated_input_tokens ~max_turns
           ~remaining_turn_budget_s
       then Degraded_retry_allowed retry
       else Degraded_retry_budget_exhausted retry
--- a/lib/keeper/keeper_turn_cascade_budget.ml
+++ b/lib/keeper/keeper_turn_cascade_budget.ml
@@ -145,67 +145,99 @@ let oas_timeout_budget_resolution_to_yojson
     ]
 
 let resolve_bounded_oas_timeout_budget_with_turn_budget
+    ~(is_retry : bool)
     ~(reserve_degraded_retry_budget : bool)
     ~(estimated_input_tokens : int) ~(max_turns : int)
     ~(remaining_turn_budget_s : float) : oas_timeout_budget_resolution option =
-  let usable_budget = remaining_turn_budget_s -. oas_timeout_guard_sec in
-  if usable_budget < min_oas_timeout_budget_sec
-  then None
-  else
-    let runtime = Keeper_runtime_resolved.current () in
-    let adaptive_timeout_sec =
-      Keeper_runtime_resolved
-      .oas_timeout_for_estimated_input_tokens_with_turn_budget
-        ~estimated_input_tokens ~max_turns
-    in
-    let turn_capped_timeout_sec =
-      Float.min adaptive_timeout_sec usable_budget
-    in
-    let retry_capped_timeout_sec =
-      if reserve_degraded_retry_budget then
-        let retry_reserved_cap =
-          usable_budget *. degraded_retry_budget_reserve_fraction
-        in
-        if retry_reserved_cap >= min_oas_timeout_budget_sec
-        then Float.min turn_capped_timeout_sec retry_reserved_cap
-        else turn_capped_timeout_sec
-      else turn_capped_timeout_sec
-    in
-    let effective_timeout_sec = retry_capped_timeout_sec in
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
Reviewed. The approach is sound: (1) is_retry is pre-existing param on do_run/retry_loop, just not propagated to budget resolution before. (2) Per-attempt floor max(15s, remaining) capped by adaptive gives retries a fighting chance. (3) Non-retry path fully preserved. (4) 4 regression tests cover boundary conditions. One observation: retry branch ignores reserve_degraded_retry_budget entirely — intentional for single-retry, but worth confirming this matches intent when is_retry=true AND reserve=true is passed.

jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/masc-mcp/pull/12702#pullrequestreview-4214890718)

jeong-sik:
@copilot-pull-request-reviewer good catch on the semantic mismatch at L1063.

The `is_retry` param at that call site reflects the **current** attempt, not the candidate retry. When the first attempt fails with `is_retry=false`, the degraded-retry eligibility check at `next_fail_open_cascade_for_turn_with_budget` runs with `is_retry=false`, which means it uses the non-retry budget path (remaining wall-clock only, no per-attempt floor).

However, this is actually safe by design:
1. The gate at L1063 decides "should we even attempt a degraded retry?" — it checks whether there's *any* budget left
2. The actual retry budget is resolved at L904 inside `do_run`, where `is_retry=true` is correctly passed from the retry loop (L1123, L1220)
3. So even if the gate is conservative (rejecting retry when remaining wall-clock is low), it doesn't cause the TurnBudgetExhausted bug — that only happens when the retry *does* execute with a near-zero budget, which the L904 per-attempt floor prevents

The trade-off: a slightly conservative gate might reject a degraded retry that would have succeeded. Worth considering `~is_retry:true` at L1063 to match the candidate semantics, but it's not a correctness issue — more of an availability optimization. I'll add a follow-up note in the code.

jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/masc-mcp/pull/12702#pullrequestreview-4214890718)

jeong-sik (commented):
## Review: Per-Attempt Retry Budget for Cascade Timeout Drift

### Assessment: Approve with observations

The core fix is correct and well-motivated. After a long first-attempt timeout, cascade retries inherit a near-zero wall-clock remaining, causing immediate TurnBudgetExhausted. The `is_retry` branch gives retries a fresh minimum budget (`max(min_budget, remaining)`) capped by adaptive timeout, bypassing the guard+reserve logic that only applies to initial attempts.

### What looks good
- Hard gate `remaining <= 0.0` prevents scheduling retries that would be immediately cancelled
- Non-retry path is completely unchanged — no behavioral regression for initial attempts
- `next_fail_open_cascade_for_turn_with_budget` correctly hardcodes `~is_retry:true` — the candidate IS always a retry
- 5 new regression tests covering the key edge cases (near-zero, healthy cap, zero remaining, non-retry refusal, degraded gate)
- Source labels `adaptive_per_attempt_retry` / `override_per_attempt_retry` enable dashboard-level retry budget tracing

### Observations (non-blocking)

**1. `remaining_turn_budget_sec` field semantic change**
Non-retry path previously stored `usable_budget` (post-guard: `remaining - 15`). Now stores raw `remaining_turn_budget_s` in both branches. This is more transparent but could surprise existing dashboard queries that expected "usable" semantics. The test assertion change (1185.0 → 1200.0) documents this.

**2. Retry branch ignores `reserve_degraded_retry_budget`**
The `is_retry=true` branch does not check `reserve_degraded_retry_budget` at all. This is correct (retries should not reserve budget for further retries), but the parameter is still accepted silently. A future cleanup could make the retry path not accept this parameter.

**3. Test `is_retry:false` in degraded retry gate tests**
`test_degraded_retry_budget_gate_allows_remaining_budget` and `test_degraded_retry_budget_gate_blocks_exhausted_budget` pass `~is_retry:false` to `next_fail_open_cascade_for_turn_with_budget`, but the function internally hardcodes `~is_retry:true`. The parameter appears unused — verify the function signature does not expose `is_retry` externally, or remove the parameter if it is dead code.

CI is still running — will verify green before merge.

jeong-sik:
Latest head a682a736 addresses the remaining retry-budget review point:

- degraded retry gate now evaluates the candidate degraded retry attempt with per-attempt retry semantics, instead of carrying the failed current attempt state
- regression test documents the first-attempt-failure call-site shape
- zero-remaining wall-clock retries are gated off

I resolved the remaining thread and am moving this out of draft so CI can verify the current remote head. Local focused Dune is currently blocked by another session holding /tmp/me-dune-local.lock for @all.

jeong-sik:
Added follow-up commit 70dc65d4ac to keep the retry-budget test suite aligned with the new gate semantics. Local verification on the PR head passed: dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/review-12702-timeout-budget --build-dir /private/tmp/masc-12702-build test/test_keeper_unified.exe; /private/tmp/masc-12702-build/default/test/test_keeper_unified.exe (304 tests). The earlier wrapper run that was queued before this patch failed on the stale pre-patch tree; current head is the passing run above.

jeong-sik:
## Review: per-attempt retry budget for cascade timeout drift

Approach looks sound. The `is_retry` bifurcation correctly separates first-attempt guard+reserve semantics from retry per-attempt fresh-floor semantics. A few observations:

### Behavioral change in `remaining_turn_budget_s` telemetry field

Non-retry path used to record `usable_budget` (= `remaining_turn_budget_s - oas_timeout_guard_sec`) in the returned record. This PR changes it to raw `remaining_turn_budget_s` (wall-clock). This is a semantic change in the telemetry field — consumers that relied on `remaining_turn_budget_s` reflecting post-guard usable budget will now see the pre-guard value. The test values shift accordingly (1200 → 500 for the reserve test, 1185 → 500 for the recorded field).

This seems intentional (the comment in the test says "raw wall-clock remaining") and is arguably more correct for telemetry, but worth calling out explicitly. If any dashboard/consumer logic subtracts guard from this field, it would double-subtract.

### Hard gate at `<= 0.0` is correct

```
if remaining_turn_budget_s <= 0.0 then None
```

This prevents retry scheduling when the outer turn timer has already expired. The Eio cancellation would kill the fiber immediately anyway — returning `None` gives a clean budget failure signal instead of a noisy rotation. Good.

### `next_fail_open_cascade_for_turn_with_budget` always passes `~is_retry:true`

Correct. The *candidate* being evaluated is always a degraded retry, regardless of whether the current failing attempt was itself a retry. The comment block in `keeper_unified_turn.ml` and `keeper_turn_cascade_budget.ml` explains this clearly.

### Test coverage

Six new test cases covering the key scenarios:
1. `is_retry=true` with 3s remaining → resolves with `effective >= min_oas_timeout_budget_sec` (15s floor)
2. `is_retry=true` with 1200s remaining → capped by `min(adaptive, remaining)`
3. `is_retry=false` with 20s remaining → `None` (guard+min rejects)
4. `is_retry=true` with 0.0s remaining → `None` (wall-clock hard gate)
5. Degraded retry gate with 3s remaining → `Degraded_retry_allowed`
6. Existing test updated: exhausted budget gate now uses 0.0s instead of 20s

The test for scenario 1 validates the core regression fix (near-zero remaining → fresh minimum budget). Scenario 5 validates the end-to-end gate path.

### Minor note

`test_degraded_retry_budget_gate_blocks_exhausted_budget` changed from `remaining_turn_budget_s:20.0` to `0.0`. With `is_retry:true` hardcoded in the gate, 20.0 would actually resolve successfully (3s remaining already does). 0.0 is the only value that triggers the `None` → `Degraded_retry_budget_exhausted` path. Makes sense.

---

Overall: LGTM. Waiting on Build/Test and TLA+ CI.

jeong-sik:
Addressed the latest Copilot review threads in a971fb24c4. The Alcotest assertions now use the expected-then-actual argument order for the retry budget checks. Local verification: scripts/dune-local.sh build test/test_keeper_unified.exe.

jeong-sik:
Updated verification for head a971fb2 after the test assertion follow-up. Local focused check passed: `dune build --root .worktrees/review-12702-timeout-budget --build-dir /private/tmp/masc-12702-build-a971 test/test_keeper_unified.exe`, then `/private/tmp/masc-12702-build-a971/default/test/test_keeper_unified.exe` (304 tests passed). Waiting on the new GitHub CI run 25252153823 before merge.

jeong-sik:
## Build/Test Failure Analysis

All 20+ test failures in the `Build and Test` job are **pre-existing main-level issues** — none are caused by this PR's changes.

### PR #12702 changed files (scope)
- `lib/keeper/keeper_turn_cascade_budget.ml`
- `lib/keeper/keeper_turn_cascade_budget.mli`
- `lib/keeper/keeper_unified_turn.ml`
- `lib/keeper/keeper_unified_turn.mli`
- `test/test_keeper_unified.ml`

### Failure categories and root causes

| Category | Tests | Root Cause | Related to this PR? |
|----------|-------|-----------|-------------------|
| **Model label parsing** | `score_provider` 0-4, `best_runner_among` 1-4, `runtime` 9-13 | `codex_cli:gpt-5.3-codex` model label fails to parse | No |
| **SSOT ratchet** | `test_masc_dirname_ssot` (fatal) | 19 new inline `.masc` literals in unrelated files (masc_trace.ml, auth_resolve.ml, etc.) | No |
| **Seed drift** | `drift_gate` 0 | `shared_memory_scope` missing from `canonical_keeper_meta_key_names` | No |
| **Cascade audit** | `cascade_config` 10 | Expected one cascade audit record, got none | No |
| **Operator predicate** | `predicate` 0 | Bootstrap + disk keeper operator profile mismatch | No |
| **Prefilter coverage** | `coverage` 1 | `masc_agent_card` not in Tool_name synonyms | No |
| **Token audit** | `audit` 2-3 | Shared token surface pair detection mismatch | No |

### Why these show up now

PRs that only change dashboard/frontend files have `Build and Test` **skipped** (path filtering). This PR changes OCaml files, so Build and Test runs — surfacing pre-existing failures that have accumulated on main.

### Recommendation

These failures need separate fix PRs targeting main. The 6 retry-budget tests added by this PR (`test_keeper_unified.ml`) are orthogonal to all failures above. Consider re-running CI after a main-level fix or marking Build and Test as non-blocking if these are known issues.

PR #12708, opened 2026-05-02
feat(dashboard): make WS-only mode the default

jeong-sik:
The WS-only mode has been tested and proven to be much more efficient, with significantly lower latency and reduced network overhead compared to the dual SSE + WS parallel mode. It also eliminates the redundant event processing on the client.

This commit switches the default value of `dashboardWsOnlyEnabled()` from `false` to `true`. Deployments can still opt-out by setting `VITE_DASHBOARD_WS_ONLY=false` or injecting `window.__MASC_DASHBOARD_WS_ONLY__ = false`.


jeong-sik:
No unresolved review threads. Added human-approved-ready and enabled squash auto-merge; waiting on CI.

PR #12709, opened 2026-05-02
refactor: replace manual prefix checks with Base.String.is_prefix

jeong-sik:
## Summary
- Replace `String.length x >= String.length p && String.sub x 0 (String.length p) = p` with `Base.String.is_prefix x ~prefix:p` across 14 files
- `>` guards (semantically meaningful: strictly longer than prefix required) are kept alongside `is_prefix`
- Files in sub-libraries without `base` dependency (`env_config_core`, `autoresearch_file`, `repo_store`) intentionally left unchanged

## Files excluded (no `base` dependency)
- `lib/config/env_config_core.ml`
- `lib/autoresearch/autoresearch_file.ml`
- `lib/repo_manager/repo_store.ml`

## Test plan
- [x] `dune build --root . bin/main_eio.exe` passes
- [x] All changes are semantically equivalent (same prefix-matching behavior)
- [ ] `dune runtest` (running, build verified)

### DIFF
--- a/lib/auto_responder.ml
+++ b/lib/auto_responder.ml
@@ -248,8 +248,7 @@ let extract_nickname (response_text : string) : string option =
     | [] -> None
     | line :: rest ->
         let trimmed = String.trim line in
-        if String.length trimmed >= String.length prefix
-           && String.sub trimmed 0 (String.length prefix) = prefix
+        if Base.String.is_prefix trimmed ~prefix
         then
           Some
             (String.trim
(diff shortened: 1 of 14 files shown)


jeong-sik:
Reviewed the prefix cleanup diff. It keeps the semantically meaningful length > prefix guards and replaces only redundant >= + String.sub prefix checks with Base.String.is_prefix; files without the Base dependency were left untouched.

Local verification:
- empty-commit scan: none

I started scripts/dune-local.sh build @all in an isolated worktree, but it became a full fresh worktree build and was stopped after several minutes to avoid blocking the queue. Heavy GitHub CI should be the compile verification surface for this PR.

jeong-sik:
Reviewed b9de311. This is a mechanical prefix-check refactor: the replaced cases had the redundant length >= prefix guard, while strict > prefix checks remain intact. Local verification passed with `scripts/dune-local.sh build lib`. I cannot approve this PR from this token because GitHub treats it as my own PR, but I found no blocking issue.

PR #12711, opened 2026-05-02
fix: surface receipt-level operator reasons

jeong-sik:
## Summary
- follow up on #12689 after it auto-merged before the review-response commit landed
- show operator_disposition_reason when it adds receipt-level detail beyond the coalesced trust summary
- apply the same condition in keeper detail and goal keeper cards

## Verification
- pnpm install --offline --frozen-lockfile
- pnpm run typecheck


jeong-sik:
Addressed the Copilot review note on the goal-tree keeper card: trustSummary now uses the same trim + fallback semantics as keeper-detail, so blank/whitespace attention_reason no longer blocks fallback or de-dupe. Local checks: `pnpm run typecheck`; `pnpm exec vitest run --config vitest.config.ts src/components/goals/goal-tree.test.ts --no-file-parallelism --maxWorkers=1` (17 tests passed).

PR #12768, opened 2026-05-03
refactor: domain-specific error hierarchy

jeong-sik:
Implements Phase 1 & 2 of the Error Hierarchy Refactoring. Eliminates flat catch-all errors in masc_error.ml and adopts a domain-specific 2-level structure. Fixes all associated compiler errors across the codebase.

### DIFF
--- a/lib/auth.ml
+++ b/lib/auth.ml
@@ -449,21 +449,21 @@ let ensure_credential_alias config ~canonical_name ~alias_name :
     let canonical_file = credential_file config canonical_name in
     if not (file_exists canonical_file) then
       Error
-        (IoError
+        (System (System_error.IoError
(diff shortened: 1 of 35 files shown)


jeong-sik:
## Review: Domain-specific error hierarchy. Good refactor: flat 20+ variant to 2-level domain dispatch. P1: Verbose qualified paths — Types.Auth(Types.Auth_error.InvalidToken _) is noisy, consider module T = Types at heavy sites. P2: CONFLICTING — rebase needed after #12765/#12766/#12767/#12764 merges. P3: to_string format breaking change — old IoError msg now [SystemError] IO error: msg, verify no downstream consumers parse old format.

PR #12777, opened 2026-05-03
Classify and observe non-board Event Layer stimuli

### LINKED ISSUE #12684: Promote non-board Event Layer stimuli into keeper turn context
## Problem
#12680 covers `board_signal` stimuli, but the Event Layer queue can carry other stimulus payloads. Today unknown/non-board stimuli may still be consumed at turn entry without becoming a structured observation, which can produce the same wake -> no_signal pattern.

## Expected
Every supported Event Layer stimulus class has a typed projection into world observation or a loud unsupported-stimulus blocker. No stimulus should be consumed silently.

## Scope
- Inventory current stimulus producers.
- Add typed observation projection for non-board classes.
- Record unsupported stimulus class as runtime trust/decision audit data instead of dropping it.
- Add KeeperEventQueue runtime parity tests.

## MASC Tracking
Goal: `goal-world-reaction-liveness`
Related: #12680, #12666

jeong-sik:
## Summary

- Add `stimulus_class` ADT (`Board_signal | Bootstrap | Unsupported of string`) and `classify` function to `Keeper_event_queue`
- Keeper heartbeat loop now logs and metrics-counts every dequeued stimulus instead of silently dropping non-board events (the "wake → no_signal" gap)
- Prometheus counters: `masc_keeper_stimulus_consumed_total` (labels: keeper, class) and `masc_keeper_unsupported_stimulus_total` (labels: keeper)
- Test file `test_keeper_event_queue.ml` covers classify for all stimulus classes and queue operation preservation

Closes #12684

## Root cause

`pending_board_event_of_stimulus` returned `None` for non-board stimuli (bootstrap, unknown), causing keepers to wake up and consume a stimulus without any structured observation reaching the turn context.

## What this does NOT change

- **Board signal handling**: unchanged — `pending_board_event_of_stimulus` still works for board signals
- **Bootstrap behavior**: bootstrap stimuli are logged but intentionally produce `None` (no turn context needed)
- **Queue data structure**: pure addition — `classify` is a pure function inspecting payload

## Test plan

- [x] `dune build` passes
- [x] `dune runtest` passes (all existing + new `test_keeper_event_queue`)
- [ ] CI green

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -114,6 +114,7 @@
   keeper_turn_lifecycle
   keeper_msg_async
   keeper_event_bus
+  keeper_event_queue
   masc_event_bus
   ;; worker_container sub-modules
   worker_container_types
--- a/lib/keeper/keeper_event_queue.ml
+++ b/lib/keeper/keeper_event_queue.ml
@@ -52,6 +52,26 @@ let sort_by_urgency (queue : t) : t =
     (fun a b -> Int.compare (urgency_rank a.urgency) (urgency_rank b.urgency))
     queue
 
+type stimulus_class =
+  | Board_signal
+  | Bootstrap
+  | Unsupported of string
+
+let classify (s : stimulus) : stimulus_class =
+  if String.equal s.payload "Keeper bootstrap signal" then Bootstrap
+  else
+    (* Board signals carry JSON with "source":"board_signal". Lightweight
+       prefix check avoids a full Yojson parse in the data layer. *)
+    let len = String.length s.payload in
+    if len > 2
+       && String.equal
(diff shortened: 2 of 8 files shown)


jeong-sik:
Review follow-up pushed in d33960c5b6: rebased onto current main, dropped the obsolete marker-cleanup commit, fixed stimulus prefix classification with String.starts_with, reused the computed stimulus_class, corrected metric docs/HELP text to point payload-prefix detail at logs, and made the prefix-only test assert Board_signal explicitly. Local verification: git diff --check, focused Dune build for test_keeper_event_queue/prometheus_coverage/keeper_registry, all three executables passed, and check-pr-hygiene passed.

PR #12810, opened 2026-05-04
fix(dashboard): align dashboard tests with semantic-token migration (#12774, #12786)

jeong-sik:
## Summary

Fix-forward for the design-system migration cascade (#12774 → #12786).
After tokens landed, eight test files still asserted the
pre-migration literal hex/rgba colors (e.g. `#fb923c`,
`rgba(251, 191, 36, 0.5)`), producing **32 failing assertions across
9 test files** on `main` even though the implementations already emit
the canonical `var(--*)` tokens. This PR re-pins the assertions to the
tokens the implementations actually return.

## Scope clarification (post follow-up `fae9919857`)

The original scope was test-only re-pinning. The follow-up commit
`fae9919857` extends scope to fix a real runtime bug surfaced by the
review:

- `activity-graph-view.ts` was emitting `var(--color-orange-400)` /
  `var(--color-pink-400)` for `debate` / `post` nodes, but neither
  custom property is defined in the dashboard token sources
  (`variables.css`, `tokens.generated.ts`, `paper-theme.css`). In the
  browser they resolve to the empty value, so debate/post nodes and
  the legend swatch render with no fill.
- The fix swaps the implementation to defined raw tokens
  (`--color-amber-bright` for orange-400, `--color-rose-light` for
  pink-400) and updates the matching test assertions in lockstep.

So for `activity-graph-view` this is a runtime-visible bug fix, not a
no-behavior-change test re-pin. The other files in the original scope
remain test-only.

## Changes

- **8 test files** — assert semantic tokens in place of literal colors:
  - `activity-graph-view.test.ts` (nodeColor + 14 edgeColor cases)
  - `activity-swimlane.test.ts` (presence span style)
  - `agent-roster.test.ts` (paused / handoff / compacting badge classes)
  - `connector-card-border.test.ts` (stale → `var(--color-warn)`)
  - `connector-readiness-rail.test.ts` (idle gradient → bg surface tokens)
  - `keeper-state-diagram.test.ts` (signalTone / badgeTone ok+err)
  - `common/distribution-bars.test.ts` (warn / bad chipBg)
  - `live/topology-map.test.ts` (offline / inactive / idle / done / in_progress)

- **1 implementation file** — `composite-fsm-flowchart.ts`:
  Revert four mermaid `classDef` lines from `var(--rose-light)` /
  `var(--ok-fg)` / `var(--amber-bright)` back to the literal hex
  equivalents (`#fb7185`, `#8ebc8e`, `#f59e0b`).
[...]

PR #12823, opened 2026-05-04
fix(dashboard): unblock main typecheck — CategoryLevel + drop 9 unused i

jeong-sik:
## Why

main is currently red on Dashboard `tsc --noEmit` (introduced by #12813). Every dashboard-touching PR inherits the failure. See e.g. #12810 Dashboard job 74228403206.

13 type errors in `dashboard/src/components/provider-capability-matrix/`:
- `data.ts:295/303` — `CategoryLevel` referenced but undeclared (TS2304) + cascading TS2366 "lacks ending return statement"
- `bfcl-rankings.ts` (4), `model-catalog.ts` (3), `oas-provider-table.ts` (1), `wiring-gaps.ts` (1) — `(item, i) =>` callbacks where `i` is unused (`noUnusedParameters`)

## What

- `data.ts`: add `export type CategoryLevel = 'high' | 'mid' | 'low'` next to the existing `FeatureSupport` declaration. Switches in `categoryLevelClass` and `categoryLevelLabel` were already exhaustive, so TS now treats them as total functions and the missing-return error disappears.
- 5 component files: drop unused `, i` from each affected `.map`. Every callback keys by `item.something`, not `i`, so removing the index is purely mechanical.

## Test plan

- [x] Local diff inspection
- [ ] CI `Dashboard` job (this PR's primary purpose — it must go from fail → pass)
- [ ] Build/Lint/etc. — should be unaffected

## Notes

This is the surgical unblock; deeper review concerns from #12813 (BFCL footer reconciliation, GLM unverified flagging, H05 location format) are intentionally out of scope.

### DIFF
--- a/dashboard/src/components/provider-capability-matrix/bfcl-rankings.ts
+++ b/dashboard/src/components/provider-capability-matrix/bfcl-rankings.ts
@@ -23,7 +23,7 @@ function V4CategoryTable() {
           </tr>
         </thead>
         <tbody>
-          ${BFCL_V4_CATEGORIES.map((cat, i) => html`
+          ${BFCL_V4_CATEGORIES.map((cat) => html`
             <tr key=${cat.id} class="pm-row-alt">
               <td class="pm-td font-semibold">${cat.label}</td>
               <td class="pm-td t-meta">${cat.description}</td>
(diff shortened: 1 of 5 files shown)


jeong-sik:
**System note** (does not require code change): `pr-automation.yml` flags `looksAgentAuthored=true` (Co-Authored-By Claude trailer in commit message), which causes the workflow to flip Ready→Draft and disable auto-merge on every cycle. The bypass label `human-approved-ready` only counts when applied by an actor listed in `AGENT_DRAFT_GUARD_HUMAN_ACTORS` (i.e. by you, not by the AI). Until you apply that label this PR cannot stay Ready long enough for heavy CI to finish. The diff itself is mechanical (5 files, +11/-9: define `CategoryLevel` + drop 9 unused `i`) and main is currently red on `tsc --noEmit` because of #12813.

PR #12804, opened 2026-05-04
feat(observability): inline metric centralization + env parse instrumentation

jeong-sik:
## Summary
- Centralize inline metric string literals across keeper modules into `Prometheus.metric_*` constants in prometheus.ml
- Add new metric constants + `add()` registrations in `init()`
- Instrument new failure paths (env parse, tool selection, task load, redacted thinking persist, etc.) with `Prometheus.inc_counter`
- Extend dashboard `TOOL-ERR` aggregate to span the full keeper failure surface (~50 metrics) — operators see one rolled-up failure indicator that grows with each new instrumented site, instead of the original 4 metrics planned in the early draft

## Changes by module
- **prometheus.ml**: +N constants, +N `add()` registrations (some `add` removed where another module owns the authoritative `register_counter` — see commit 64e7156ba9)
- **prometheus.mli**: matching val declarations
- **keeper_hooks_oas.ml**: inline → `Prometheus.metric_*`
- **keeper_guards.ml / keeper_execution_receipt.ml**: inline → `Prometheus.metric_*` (these modules keep their authoritative `register_counter` calls; prometheus.ml does not re-register)
- **keeper_shell_bash.ml / keeper_shell_docker.ml**: inline → `Prometheus.metric_*`
- **keeper_heartbeat_snapshot.ml / keeper_stay_silent_loop_detector.ml / keeper_unified_metrics.ml**: inline → `Prometheus.metric_*`
- **keeper_run_tools.ml**: +counter instrumentation
- **keeper_agent_run.ml / keeper_unified_turn.ml**: failure-path counter instrumentation
- **dashboard.ml**: TOOL-ERR aggregate covers the full keeper failure surface; `metric_keeper_event_bus_drain` removed because it counts success/empty drain ticks too

## Review-response follow-up commits (after initial Copilot pass)
- `91ebeebaa7` fix: redacted thinking persist counter was incremented on every attempt because the `try ... with` group closed before `inc_counter`; now matches the parallel non-redacted path.
- `0164bb4322` fix: unify `metric_keeper_event_bus_drain` label schema on the exception path so the failure series stays inside the documented `{site, outcome}` contract instead of inventing a `(keeper, site=turn_drain)` series.
[...]

PR #12828, opened 2026-05-04
fix(prometheus+keeper): cleanup dead duplicate add() block + 2 "unknown" stand-ins

jeong-sik:
## Summary

Recovery of cleanup work that did not land before #12804 squash-merged
(prometheus dead-block removal) plus two more `"unknown"` stand-ins
identified during the cycle-21 sweep that had real keeper context in
scope.

## prometheus.ml — drop dead duplicate `add()` block

A second registration block for keeper turn / cascade / persistence
/ observation failure metrics had grown next to the SSE registrations.
Each metric was already registered earlier in init() with detailed
help text — see `add metric_keeper_turn_livelock_blocks` at line ~1171
and the `turn_livelock_blocks .. session_cleanup_failures` block at
lines ~1579-1628.

`add` is a no-op when the metric name already exists, so every help
string in the duplicate block was silently dropped. Several recent
review threads on #12804 tried to "fix" help/labels in the duplicate
copy and had no observable effect on `/metrics` output.

The duplicate block (12 metrics, 24 lines) is removed and replaced
with a tombstone comment pointing future readers at the canonical
registration sites. The non-duplicated registrations next to the
dropped block (`stale_termination_*`,
`oas_timeout_budget_watchdog_termination`, `tool_use_failure`) are
preserved intact.

Also align `metric_keeper_usage_trust` help text with what the only
call site emits — `keeper_unified_metrics.ml` uses label key `outcome`,
not `trust`.

## keeper modules — drop `"unknown"` stand-ins where agent_name is in scope

- `lib/keeper/keeper_post_turn.ml:855` (overflow_save): the surrounding
  `save_oas_checkpoint` call reads `~agent_name:retry_meta.agent_name`.
  Replace `"unknown"` with `retry_meta.agent_name` and align the
  second label key from `phase` to `operation` to match the rest of
  `metric_keeper_checkpoint_failures` (documented as `(keeper, operation)`
  after the cycle-9 unification).
- `lib/keeper/keeper_guards.ml:283` (event_emit): `agent_name` is in
  scope on the surrounding log line. Use it instead of `"unknown"`.

Production behaviour unchanged — every counter still fires from inside
the same exception handler, only the label series gains keeper
attribution.

## Out of scope (separate cleanup)

Four more `"unknown"` stand-ins live in functions that have no keeper
context to fall back on (`keeper_meta_store.ml` 3 sites,
`keeper_guards.ml:178` gate_observer). They need either an
`"aggregate"` placeholder fix (matching the cycle-10/14 keepalive_signal
+ approval_queue audit_store_create pattern) or function-signature
plumbing — separate review.

## Test plan

- [x] `dune build --root . @check` passes clean after each commit
- [ ] CI green on `Build and Test`, `Lint`, `Health`
- [ ] Required checks green before flipping to Ready

## Context

This is a recovery PR. The dead-block cleanup was prepared as commit
`bc20d79f95` on the `keeper-tool-observability` branch in the cycle
just before #12804 was merged, but #12804 squash-merged before that
commit landed.

### DIFF
--- a/lib/keeper/keeper_guards.ml
+++ b/lib/keeper/keeper_guards.ml
@@ -280,7 +280,7 @@ let emit_gate_event
     | exn ->
       Prometheus.inc_counter
         Prometheus.metric_keeper_guards_failures
-        ~labels:[("keeper", "unknown"); ("site", "event_emit")]
+        ~labels:[("keeper", agent_name); ("site", "event_emit")]
         ();
       Log.Keeper.warn
         "keeper_guards: event emit failed stage=%s tool=%s err=%s"
--- a/lib/keeper/keeper_post_turn.ml
+++ b/lib/keeper/keeper_post_turn.ml
@@ -852,7 +852,7 @@ let recover_latest_checkpoint_for_overflow_retry
                 "overflow retry checkpoint save failed: %s" e;
               Prometheus.inc_counter
                 Prometheus.metric_keeper_checkpoint_failures
-                ~labels:[("keeper", "unknown"); ("phase", "overflow_save")]
+                ~labels:[("keeper", retry_meta.agent_name); ("operation", "overflow_save")]
                 ();
               None)
         with
--- a/lib/prometheus.ml
+++ b/lib/prometheus.ml
@@ -1880,34 +1880,17 @@ let init () =
   add metric_sse_external_subscribers
     "Active non-SSE subscribers bridged from the SSE fanout path" Gauge;
 
-  add metric_keeper_turn_livelock_blocks
-    "Total livelock-block events where keeper found a turn already in-progress      during unified_turn start. Labels: keeper." Counter;
-  add metric_keeper_turn_timeout_committed
-    "Total wall-clock timeout events after tools were committed but before      response completed. Labels: keeper." Counter;
-  add metric_keeper_turn_error_after_tools
-    "Total errors after tool calls were committed in unified_turn.      Labels: keeper." Counter;
-  add metric_keeper_cascade_sync_failures
-    "Total failures to sync cascade state during turn pause/rejection.      Labels: keeper, site." Counter;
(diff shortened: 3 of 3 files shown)


jeong-sik:
Sibling: #12830 also touches prometheus.ml (different region — type-constants + new add() for metric_keeper_directive_failures). No expected textual conflict; merge order does not matter.

jeong-sik:
## Follow-up scope (not in this PR)

main has 6 more sites with the same `keeper="unknown"` placeholder
pattern that we cleared from `keeper_approval_queue.ml` in #12804.
Verified on `origin/main` (after the watchdog/opam advances):

**Real keeper context available — replaceable**:
- `lib/keeper/keeper_guards.ml:283` (event_emit): `agent_name` is in
  scope on the same line. Use it instead of `"unknown"`.
- `lib/keeper/keeper_post_turn.ml:855` (overflow_save): `retry_meta.agent_name`
  is in scope from the surrounding `save_oas_checkpoint` call. Replace
  `"unknown"` with `retry_meta.agent_name` and align the second label
  key from `phase` to `operation` (the rest of
  `metric_keeper_checkpoint_failures` uses `operation` per #12804
  cycle 9 unification).

**No keeper context — `aggregate` placeholder**:
- `lib/keeper/keeper_guards.ml:178` (gate_observer): the
  `gate_decision_event` record has no keeper id field; the observer
  callback runs with no surrounding agent_name binding.
- `lib/keeper/keeper_meta_store.ml:33` (meta_parse): meta-file JSON
  parse failed, so the keeper name itself is unrecoverable.
- `lib/keeper/keeper_meta_store.ml:66` (persisted_listdir): list_dir
  failure at the keeper directory root.
- `lib/keeper/keeper_meta_store.ml:234` (changed_parse): same as
  meta_parse.

Tracking as a separate cleanup PR rather than expanding scope here.

PR #12836, opened 2026-05-04
fix(keeper): instrument decision record JSONL append failure

jeong-sik:
## Summary
- Add metric_keeper_decision_audit_flush_failures counter to append_decision_record exception handler
- Identified during warn-level gap scan (PR #12800), landed after squash merge

## Test plan
- [x] dune build lib/ passes
- [ ] Verify metric in /metrics on append failure

Generated with Claude Code

### DIFF
--- a/lib/keeper/keeper_unified_metrics.ml
+++ b/lib/keeper/keeper_unified_metrics.ml
@@ -994,6 +994,10 @@ let append_decision_record
   with
   | Eio.Cancel.Cancelled _ as e -> raise e
   | exn ->
+      Prometheus.inc_counter
+        Prometheus.metric_keeper_decision_audit_flush_failures
+        ~labels:[("keeper", meta.name)]
+        ();
       Log.Keeper.warn "append decision record failed for %s: %s"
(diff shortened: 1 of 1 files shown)


jeong-sik:
**Triage of the 2 review threads** (semantic + test coverage):

| Concern | Suggested fix |
|---------|---------------|
| Reusing `metric_keeper_decision_audit_flush_failures` (buffered `flush_if_needed` path) for direct `decisions.jsonl` append | Either rename the existing metric to be path-agnostic (e.g. `metric_keeper_decision_record_append_failures`) and update the buffered call sites, or introduce a separate counter for the direct path. Mixed semantics will fragment failure attribution in the dashboard. |
| New failure-path instrumentation lacks test coverage | Add a test in `test/test_keeper_unified.ml` that forces `append_jsonl_line` to fail (e.g. `Sys_error` from a permission-restricted path) and asserts the counter increments by 1. Same shape as the existing `append_decision_record` tests in that file. |

Total cleanup: ~5 LOC + 1 test. Author decision needed on metric naming (option A vs B above).

PR #12901, opened 2026-05-04
feat(env_config): add get_int_nonneg / get_float_nonneg helpers

jeong-sik:
## Summary

- Adds `get_int_nonneg` / `get_float_nonneg` to `lib/config/env_config_core.{ml,mli}` — variants of `get_int` / `get_float` that floor negatives at `default`. Float variant also rejects NaN.
- 11 unit tests in `test/test_env_config_nonneg.ml` pinning: negative→default, zero passes, positive passes, unset→default, garbage→default, NaN→default (float-only).
- No call-site migration here. Existing `get_int` / `get_float` keep accepting negatives unchanged.

## Why

Audit P3 finding (2026-04-29 audit, deferred): `MASC_KEEPER_ALERT_MAX_RETRIES=-5` is silently accepted as `-5` today. Many `MASC_KEEPER_*` env vars (`BOOTSTRAP_STALE_TURN_SEC`, `ALERT_MIN_SCORE`, `ALERT_MAX_BODY_CHARS`, `ALERT_MAX_RETRIES`, `METRICS_MAX_BYTES`, ...) treat negatives as nonsensical at the call site but the parser doesn't enforce that.

This is the smallest seam that lets a follow-up sweep over `lib/config/env_config_keeper.ml`'s 25+ getters opt into the floor-at-zero contract one knob at a time, without flipping a global default.

## NaN guard

`nan < 0.0` is `false` in OCaml, so a non-finite parse would slip past a `< 0.0`-only check. Explicit `Float.is_nan` guard pinned by `test_float_nonneg_nan_falls_back`.

## Test plan

- [x] `dune exec test/test_env_config_nonneg.exe` — 11/11 green
- [x] Race check pre-push: no concurrent `env_config_nonneg`/`get_int_nonneg`/`get_float_nonneg` PR open

### DIFF
--- a/lib/config/env_config_core.ml
+++ b/lib/config/env_config_core.ml
@@ -42,6 +42,20 @@ let get_float ~default name =
   | Some v -> Safe_ops.float_of_string_with_default ~default v
   | None -> default
 
+(** Variants that floor at zero.  An operator who sets a negative
+    value (e.g. [MASC_KEEPER_ALERT_MAX_RETRIES=-5]) gets the default
+    rather than the literal — negative budgets/counts are
+    nonsensical for the call sites these feed
+    ({!Env_config_keeper}, alert thresholds, retry caps).  NaN is
+    treated as out-of-range for the float variant. *)
+let get_int_nonneg ~default name =
+  let parsed = get_int ~default name in
+  if parsed < 0 then default else parsed
+
+let get_float_nonneg ~default name =
+  let parsed = get_float ~default name in
+  if Float.is_nan parsed || parsed < 0.0 then default else parsed
+
 let get_bool ~default name =
   match raw_value_opt name with
   | Some v ->
--- a/lib/config/env_config_core.ml
+++ b/lib/config/env_config_core.ml
@@ -46,15 +46,23 @@ let get_float ~default name =
     value (e.g. [MASC_KEEPER_ALERT_MAX_RETRIES=-5]) gets the default
     rather than the literal — negative budgets/counts are
     nonsensical for the call sites these feed
-    ({!Env_config_keeper}, alert thresholds, retry caps).  NaN is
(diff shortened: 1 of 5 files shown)


jeong-sik:
@copilot-pull-request-reviewer Good catch — addressed in `9be7dc89e0`.

**Root cause**: `Float.is_nan` only catches `NaN`, but `float_of_string "inf"` → `+∞` and `+∞ > 0.0` is `true`, so the `< 0.0` check also let it pass. For timeout/score/ratio settings this helper protects, `+∞` is just as nonsensical as `NaN`.

**Fix**: Switched the guard to `not (Float.is_finite parsed) || parsed < 0.0`, which uniformly rejects `{NaN, +∞, -∞, negative}`. `-∞` was already caught by the negative check, but the unified guard pins the contract symmetrically across all three non-finite values.

**Tests added**:
- `test_float_nonneg_pos_inf_falls_back` — pins `+inf` rejection (the regression you flagged)
- `test_float_nonneg_neg_inf_falls_back` — pins `-inf` rejection (defense-in-depth)

13 cases now (was 11). Mli docstring + body comment updated to call out the `+∞ > 0.0` foot-gun.

jeong-sik:
@copilot-pull-request-reviewer Good catch — addressed in `24edc0960b`.

Moved `test_env_config_nonneg` to `test/stanzas/test_env_config_nonneg.inc` per the per-file convention documented in `test/stanzas/README.md`. Now the only `test/dune` edit is one `(include stanzas/test_env_config_nonneg.inc)` line at its alphabetical slot between `test_env_config_keeper_poll_intervals.inc` and `test_env_config_sandbox.inc`.

Removed both legacy references (Group 1 `(names ...)` + Group 1 `(modules ...)`).
Same test binary, same 13 cases (`13 tests run. Test Successful`), no behavior change.

PR #12934, opened 2026-05-05
feat(dashboard): X-Workspace-Source header + EXPLORER fallback hint

jeong-sik:
## Summary

Closes the keeper-aware workspace UX loop (#12893 → #12907) by surfacing *how* the resolution succeeded back to the frontend. Without this, the IDE silently shows the project tree even when the user expected playground content (because the keeper has no playground yet, or the meta lookup failed).

- Backend: tree/file/blame/diff success responses emit `X-Workspace-Source` header with `project` / `playground:<name>` / `playground_missing:<name>` / `keeper_unknown:<name>`. Wire-shape unchanged (header only, JSON body identical).
- Frontend: `IdeExplorer` parses the header into a discriminated union and renders a small fallback notice below the EXPLORER header when the resolution fell back. `role="status" aria-live="polite"` so screen readers are notified without stealing focus.

## Why

Today the IDE silently swallows resolution failures — when a keeper is selected but the playground doesn't exist on disk, users see the project tree and assume they're looking at the keeper's working state. The fallback hint makes the actual source visible without leaking the absolute path back to the client.

## Test plan

- [x] `dune build --build-dir=_build_iso lib/masc_mcp.cma` — green
- [x] `dune exec test/test_workspace_routes_keeper.exe` — 11/11 pass (was 7, +4 source_header cases)
- [x] `pnpm vitest run src/components/ide/ide-explorer.test.ts` — 9/9 pass (new file, parseWorkspaceSource decoder)
- [x] `pnpm lint src/components/ide/ide-explorer.ts` — 0 errors on changed file

## Out of scope

- Hint styling polish (currently uses `--color-bg-muted` block — design system review pending)
- Surfacing the source on the Editor pane (only the EXPLORER renders the hint today; Editor file-load failures are still silent)

### DIFF
--- a/dashboard/src/components/ide/ide-explorer.test.ts
+++ b/dashboard/src/components/ide/ide-explorer.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it } from 'vitest'
+import { parseWorkspaceSource } from './ide-explorer'
+
+describe('parseWorkspaceSource', () => {
+  it('treats null header as project', () => {
+    expect(parseWorkspaceSource(null)).toEqual({ kind: 'project' })
+  })
+
+  it('treats empty header as project', () => {
+    expect(parseWorkspaceSource('')).toEqual({ kind: 'project' })
+  })
+
+  it('decodes plain "project" tag', () => {
+    expect(parseWorkspaceSource('project')).toEqual({ kind: 'project' })
+  })
+
+  it('decodes "playground:<name>"', () => {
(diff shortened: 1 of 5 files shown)


jeong-sik:
Superseded by split PRs after main #12921 (IDE mock → real data coordinator) made the frontend half architecturally stale: **#12951** (backend X-Workspace-Source header) + **#12957** (frontend decoder + EXPLORER hint on data-coordinator). Closing; work continues there.

PR #12941, opened 2026-05-05
feat(env_config): add get_ratio for [0, 1]-bounded float env vars

jeong-sik:
## Summary

Adds `get_ratio` to `lib/config/env_config_core.{ml,mli}` — a `[0, 1]`-bounded float env getter that delegates to `get_float_nonneg` (#12901) for the non-finite + negative rejection and adds the `> 1.0` upper bound. 16 unit tests across five property groups.

## Why

Audit P3 follow-up to #12901. Many `MASC_*` env vars are fractions where the operator's mental model is "this is a fraction in [0, 1]":

- `MASC_KEEPER_ALERT_MIN_SCORE` (0.70)
- `MASC_KEEPER_ALERT_GITHUB_MIN_SCORE` (0.85)
- `MASC_KEEPER_SELF_PRESERVATION_RATIO` (0.3)
- `MASC_CONTEXT_RATIO_HARD_CAP` (0.95)
- `MASC_COMPACT_DYN_MULTI_AGENT_RATIO` (0.80)
- `MASC_COMPACT_DYN_FOCUSED_RATIO` (0.70)
- `MASC_DASHBOARD_RUNTIME_WARNING_CTX_RATIO` (0.95)

`get_float` / `get_float_nonneg` accept any non-negative finite value, so a typo like `0.95` → `95` silently propagates a 95× multiplier downstream. `get_ratio` rejects everything outside `[0, 1]` and falls back to the default.

## Properties pinned (16 cases)

| # | Property | Test count |
|---|---|---|
| 1 | In-range (0.0, 0.5, 1.0) passes | 3 |
| 2 | Out-of-range (negative, >1.0, 1.0+ε, NaN, +∞, -∞) → default | 6 |
| 3 | Parse fallthrough (garbage / unset) → default | 2 |
| 4 | Out-of-range *default* clamped to `[0, 1]` | 2 |
| 5 | Non-finite *default* (NaN, +∞, -∞) sanitised to finite `[0, 1]` | 3 |

## Refactor (per review)

`get_ratio` originally duplicated the parsing + non-finite + negative rejection logic from `get_float_nonneg`. After Copilot review (`9f9f1b8fb9`), it now delegates: `get_float_nonneg` does the heavy lifting, `get_ratio` just adds the `> 1.0` upper bound. Single source of truth for the rejection semantics.

## Migration

No call-site migration in this PR. Seven candidate knobs above will be migrated in a follow-up sweep (one PR per area).

## Test plan

- [x] `dune build lib/config` clean
- [x] `dune exec test/test_env_config_get_ratio.exe` — 16/16 green
- [x] Test added under `test/stanzas/test_env_config_get_ratio.inc` per `test/stanzas/README.md`
- [x] Race check pre-push: no concurrent `get_ratio` PR

### DIFF
--- a/lib/config/env_config_core.ml
+++ b/lib/config/env_config_core.ml
@@ -66,26 +66,31 @@ let get_float_nonneg ~default name =
 
 (** Variant for [\[0.0, 1.0\]]-bounded floats — probabilities,
     score thresholds, context-ratio caps, anything where the
-    operator's mental model is "this is a fraction".  Out-of-range
-    parses ([< 0.0], [> 1.0], or non-finite) fall back to
-    [default].  The default itself is also clamped to
-    [\[0.0, 1.0\]] as defense-in-depth so a future caller cannot
-    accidentally introduce an out-of-range default.
+    operator's mental model is "this is a fraction".  Implemented
+    by delegating to {!get_float_nonneg} (which already handles
+    NaN/-∞ rejection and the negative-floor semantics) and then
+    adding the [> 1.0] upper bound.
+
+    The [default] itself is sanitised first: non-finite inputs
+    ([NaN], [+∞], [-∞]) are coerced to [0.0]; finite values
+    out of range are clamped via [Float.max 0.0 (Float.min 1.0 .)].
+    This is defense in depth so a caller passing a stale
+    out-of-range default still gets a valid ratio back.
 
     NaN-safety note: [Float.min nan 1.0] / [Float.max nan 0.0]
-    propagate NaN, so a naive [Float.max 0.0 (Float.min 1.0 v)]
-    would still return NaN when [v] is NaN.  We sanitise non-
-    finite [default] inputs to [0.0] before clamping. *)
+    propagate NaN per OCaml's IEEE 754 semantics, so a naive
+    [Float.max 0.0 (Float.min 1.0 v)] without the explicit
+    {!Float.is_finite} guard would still leak NaN. *)
 let get_ratio ~default name =
   let sanitise v =
     if not (Float.is_finite v) then 0.0
     else Float.max 0.0 (Float.min 1.0 v)
   in
   let safe_default = sanitise default in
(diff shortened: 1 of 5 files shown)


jeong-sik:
@chatgpt-codex-connector @copilot-pull-request-reviewer Both flagged the same P2 — addressed in `07ef68c839`.

**Root cause**: `Float.min nan 1.0` and `Float.max nan 0.0` both return NaN in OCaml, so the previous `Float.max 0.0 (Float.min 1.0 v)` clamp would still return NaN when `default` is NaN. Contract said "always returns a finite [0, 1] value" but the implementation didn't enforce it.

**Fix**: Sanitise non-finite defaults to `0.0` (the conservative "no activation" ratio) via `Float.is_finite` *before* the `[0, 1]` clamp. Handles `NaN`, `+∞`, `-∞` uniformly.

**Tests added** (3 new cases, 16 total now):
- `default = NaN` → must be finite + in `[0, 1]`
- `default = +∞` → must be finite + in `[0, 1]`
- `default = -∞` → must be finite + in `[0, 1]`

Mli docstring updated to call out the OCaml `Float.min nan` propagation foot-gun explicitly.

jeong-sik:
@copilot-pull-request-reviewer Doc fix in `ac2ae0d7df`. The mli now documents that non-finite defaults are coerced to `0.0` before clamping, with the rationale (`Float.min nan 1.0 = nan` per IEEE 754 → naive clamp would leak NaN), and explicitly promises callers a finite `[0, 1]` return.

jeong-sik:
@copilot-pull-request-reviewer Both points addressed: refactored `get_ratio` to delegate to `get_float_nonneg` (`9f9f1b8fb9`) — single source of truth for non-finite/negative rejection, future tightening auto-applies. PR description updated from 13 → 16 cases (3 in-range + 6 out-of-range + 2 fallthrough + 2 default-clamping + 3 non-finite-default sanitisation). Mli docstring describes the delegation explicitly. 16/16 tests green.

PR #12904, opened 2026-05-04
feat(keeper): per-provider token bucket primitive (PR 1/5 of liveness scheduler)

jeong-sik:
## Summary

First module of the keeper-liveness scheduler series.

Production symptom motivating this work: **14 keepers vs 6 autonomous slots → 8 always queued → tail wait > 60s budget → all keepers timeout simultaneously** (live log 2026-05-05 ~01:50–17:30 UTC, 245 timeouts in 30min). PR #12863 fixed a clock-phase leak that consumed 5s of budget but did not address the underlying capacity-vs-demand mismatch.

The user-stated 1st-priority invariant is **liveness**: \"어떤 keeper도 절대 막히지 않음.\" That requires a layered scheduler with provable bounded wait. This PR is the foundational primitive (Layer 1).

## Invariants (across the 5-PR series)

\`\`\`
I1 (Liveness):       ∀ keeper k. eventually-progresses(k)
I2 (Work-conserve):  ¬∃ t. (waiting(k, t) ∧ idle_provider(p, t) ∧ compatible(k, p))
I3 (Rate-respect):   ∀ provider p, t. dispatched(p, [t-1m, t]) ≤ rate_limit(p)  ← THIS PR
I4 (Bounded-wait):   ∀ k. wait(k) ≤ T_max  (from DRR fairness)
I5 (Drift-observable): ∀ rotation event. logged(keeper, preferred → actual, reason)
\`\`\`

## What this PR adds

- \`lib/keeper/keeper_provider_token_bucket.{ml,mli}\` — the I3 primitive
- Non-blocking \`try_acquire\` (returns bool, never sleeps the fiber)
- Lazy refill from injected clock (production: \`Unix.gettimeofday\`; tests: controlled)
- Cross-fiber safe via \`Stdlib.Mutex\` (microsecond critical section)
- Pure \`refilled\` helper exported for arithmetic-level testing without state

## What this PR does NOT do

- No call site changes — bucket is unused production code at merge.
- No semaphore replacement (PR 5/5).
- No cascade routing changes (PR 2/5 + PR 3/5).
- No TLA+ proof of liveness (PR 4/5).

This is intentional: PR 1 is foundation only. Future PRs depend on its semantics.

## Test plan

- [x] 12 alcotest cases (run locally, all green in 2ms)
- [x] Pure refill arithmetic (caps, negative-elapsed clock-skew, partial)
- [x] Constructor invariants (rejects capacity=0, refill_rate=0)
- [x] Drain → fail → refill → recover
[...]

PR #12949, opened 2026-05-05
feat(keeper): RFC-0026 PR-E-2 admission caller integration

jeong-sik:
## Summary
RFC-0026 PR-E-2 1/3 — heartbeat caller scaffold.

Adds a shadow telemetry probe at the turn-entry of `Keeper_heartbeat_loop`, calling `Keeper_admission_glue.decide` with **empty** policy/bucket lookups and recording the path it would have taken. The existing `Keeper_turn_slot.with_keeper_turn_slot` semaphore path runs **exactly as before**.

## Why this is a scaffold (not the unstuck PR yet)

This PR does not act on the new admission decision. It only observes. The reason is that two prerequisites are still missing on `main`:

1. **Persona policy registry loader** — reads `[admission.<keeper>]` sub-tables from `cascade.toml`. Without this, `Keeper_admission_registry.lookup` always returns `None` → `decide` always returns `Legacy_path` (safe fallback by design). Lands as PR-E-3.
2. **Per-provider bucket population** — `Keeper_admission_router` needs a populated `bucket_lookup`. The bucket primitive was merged in #12904 but no caller registers actual provider rate limits yet. Lands as PR-E-4.

Until both land, the new router cannot make any non-Legacy decision. So this PR's job is just to wire the call site and start collecting baseline data.

## What this PR enables
- New counter `masc_keeper_admission_path_total{keeper, path}` where `path ∈ {legacy, new_dispatch, new_wait, new_surface}`
- Baseline distribution: 100% `legacy` until PR-E-3/PR-E-4 land
- Once loader + buckets are wired, the same counter shows actual `New_admission` breakdown per keeper — live A/B signal **without flipping `MASC_ADMISSION_USE_NEW`**

## Production effect
Zero. The probe runs `decide` with empty lookups and increments a counter. The legacy semaphore path is unchanged. No new env var. No config change.

## Test plan
- [x] `dune build --root . @check` exits 0
- [ ] CI green
- [ ] After PR-E-3 + PR-E-4 land, counter shows non-`legacy` paths in some keepers

## Refs
- RFC-0026 #12909 (admission router design)
- #12939 (admission glue + registry, merged 2026-05-05)
- #12938 (admission router + WFQ, merged 2026-05-05)
- #12926 (persona admission policy types, merged 2026-05-05)
- #12904 (per-provider token bucket primitive, merged earlier)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -580,6 +580,40 @@ let run_keepalive_unified_turn
       then meta_after_triage
       else if should_run_turn
       then (
+        (* RFC-0026 PR-E-2 shadow probe.  Calls the new admission glue
+           but does NOT yet act on its result; the probe runs even
+           when [MASC_ADMISSION_USE_NEW] is unset (in that case the
+           glue returns [Legacy_path] immediately).  Buckets and the
+           policy registry are intentionally empty in this PR — the
+           registry loader and per-provider bucket population land in
+           later PRs (PR-E-3 / PR-E-4).  Until those land, [decide]
+           reports [Legacy_path] for every keeper, which is
+           semantically identical to the existing semaphore path.
+           The counter is what we want: a baseline distribution we
+           can diff against once the loader is wired. *)
+        let admission_path_label =
+          let empty_buckets : Keeper_admission_router.bucket_lookup =
+            fun _ -> None
+          in
+          let empty_policies : Keeper_admission_glue.policy_lookup =
+            fun _ -> None
+          in
+          match
+            Keeper_admission_glue.decide
+              ~keeper_id:meta_after_triage.name
+              ~policies:empty_policies
+              ~buckets:empty_buckets
+          with
+          | Keeper_admission_glue.Legacy_path -> "legacy"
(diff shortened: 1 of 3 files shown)


jeong-sik:
Superseded by #12955 (also by jeong-sik, different session).

#12955 ships the same shadow-mode call site this PR introduced, but with a richer surface that #12949 lacked:

- Separate `keeper_admission_runtime.{ml,mli}` module (105 LOC, singleton + `observe` shim) instead of inline match in `keeper_heartbeat_loop.ml`
- 6 alcotest unit tests covering default state / observe outcome / registration observability / idempotency / reset
- Explicit PR-E-1.7 hook (`set_policy_lookup` / `set_bucket_lookup`) so the next layer can wire registry/buckets without further plumbing changes

The single counter difference (`masc_keeper_admission_path_total{keeper,path}` here vs `metric_keeper_admission_shadow_outcome{outcome}` in #12955) is marginal — per-keeper labeling can be added cheaply on top of the runtime module.

Closing in favor of the superset. No action needed.

Refs: #12955, #12947 (PR-E split plan), RFC-0026 #12909

PR #12952, opened 2026-05-05
test(keeper): unit tests for is_masc_write_allowed path policy

jeong-sik:
## Summary

Audit P3 follow-up — pin the lexical [.] / [..] collapse behaviour of internal [Keeper_tool_policy.normalize_path] indirectly through the public [is_masc_write_allowed]. 19 cases.

## Why

Audit §1.1 (2026-04-29) explicitly flagged:

> 이 함수가 "prevent traversal bypasses"를 목표로 하지만, 실제로 [.masc/playground/../../reputation/] 같은 입력이 어떻게 처리되는지 unit test가 없습니다 (Track C에서 확인).

[normalize_path] is internal (not in the .mli) but its behaviour is observable via [is_masc_write_allowed], and a regression there is a security boundary leak.

## Properties pinned (19 cases)

| Group | Coverage |
|---|---|
| **Allowed prefixes (4)** | `.masc/playground/...`, `.masc/decision_audit/...`, `.worktrees/...` |
| **Disallowed paths (6)** | `.masc/reputation/`, `.masc/economy/`, `/etc/passwd`, `"/"`, `""`, bare relative |
| **`..` collapse rejects escapes (6)** | `playground/../reputation/`, `playground/k1/../../economy/`, `.worktrees/x/../../../etc/passwd`, `../../etc/passwd`, `playground/../playground/x` (re-enters → allowed), `playground/./k1/./file.txt` (`.` collapse) |
| **Prefix-name boundaries (3)** | `.masc/decision_audit_sneaky/`, `.masc/playgrounds/` (plural), `.worktrees-stale/` |

(4) addresses audit §1.2 — `String.starts_with path ~prefix` with the trailing `/` on the writable prefix correctly excludes sibling directories with a shared byte-prefix. The test pins this so a future refactor doesn't accidentally drop the trailing-`/` invariant.

Symlink resolution is explicitly out of scope (mli docstring lines 35-37: *"Does not resolve symlinks — pure lexical normalisation"*), matching the audit §1.1 finding.

## Test plan

- [x] Test added under `test/stanzas/test_keeper_tool_policy_paths.inc` per `test/stanzas/README.md` (per-file convention)
- [x] Race check pre-push: no concurrent `normalize_path` / `is_masc_write_allowed` test PR
- [ ] Test build currently blocked on unrelated `keeper_unified_turn.mli`/`.ml` drift on current `main` (separate fix in flight at #12945). Will validate the suite green once that lands.

### DIFF
--- a/test/dune
+++ b/test/dune
@@ -1109,6 +1109,7 @@
 (include stanzas/test_keeper_supervisor_write_meta_source.inc)
 (include stanzas/test_keeper_tool_emission_hook.inc)
 (include stanzas/test_keeper_tool_matrix.inc)
+(include stanzas/test_keeper_tool_policy_paths.inc)
 (include stanzas/test_keeper_turn_fsm_tla_parity.inc)
 (include stanzas/test_keeper_turn_livelock_10121.inc)
 (include stanzas/test_keeper_turn_up_create_meta_retry.inc)
--- a/test/stanzas/test_keeper_tool_policy_paths.inc
+++ b/test/stanzas/test_keeper_tool_policy_paths.inc
@@ -1,0 +1,9 @@
+; Per-file dune stanza for the [Keeper_tool_policy.is_masc_write_allowed]
+; path-handling regression suite.  Lives here per
+; test/stanzas/README.md to avoid the conflict-cascade hotspot
+; in the legacy grouped (tests ...) stanza.
+
+(test
+ (name test_keeper_tool_policy_paths)
+ (modules test_keeper_tool_policy_paths)
+ (libraries masc_test_deps))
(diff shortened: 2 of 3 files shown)


jeong-sik:
@copilot-pull-request-reviewer Good catch — fact-checked and corrected in `02069097c4`. `test/test_decision_pipeline.ml:113-133` does cover 6 baseline cases (3 allowed/3 disallowed including one `..` escape). Updated the header to acknowledge that and reframe this suite as the dedicated isolated pin (prefix-name boundaries, root-overshoot, non-escape "round-trip into writable") at a different abstraction level. Overlap is intentional redundancy.

jeong-sik:
Two fixes in `cded4d202e`. @chatgpt-codex-connector P2: pinned root-overshoot drop semantics with a writable-prefix path — `.masc/playground/../../../.masc/playground/x` collapses to `.masc/playground/x` (writable). If the normaliser raised or preserved literal `..`s, the test would fail. Distinguishes "drop" semantic from "/etc/passwd not writable" rejection. @copilot-pull-request-reviewer line-number drift fix: header now references the `test_keeper_writable_prefix_paths` block instead of fragile line ranges. 20/20 cases green.

Issue #13075, opened 2026-05-05
Local opam agent_sdk pin drift makes masc-mcp Dune validation non-deterministic


jeong-sik:
## Symptom

Focused local validation for a keeper heartbeat fix could not reach the target
test because the shared opam switch kept drifting between the repo SSOT pin and
a newer OAS commit.

## Reproduction / evidence

- Repo pin checker expected: `agent_sdk` from
  `git+https://github.com/jeong-sik/oas.git#2920d6127d02c8069e7bd94e7ac666c431c389f9`.
- Observed before repair: `agent_sdk` pinned to
  `git+https://github.com/jeong-sik/oas.git#e3e6683fb8255a42fcf3b2bd2227b9c0103add5c`.
- `scripts/check-oas-pin.sh --local-only` failed with that exact mismatch and
  suggested `scripts/opam-pin-external-deps.sh`.
- Running `scripts/opam-pin-external-deps.sh --install` temporarily installed
  `agent_sdk 0.187.7`, but shortly after,
  `opam list agent_sdk --columns=name,version,pinned:` showed
  `agent_sdk 0.190.1` again and `scripts/check-oas-pin.sh --local-only` failed
  with the same `e3e6683` mismatch.
- Focused build attempts then failed before keeper tests with global CMI errors
  like:
  `/Users/dancer/.opam/5.4.1/lib/agent_sdk/agent_sdk.cmi and lib/types/.masc_types.objs/byte/types.cmi make inconsistent assumptions over interface Types`.

## Estimated cause

Multiple local sessions share one opam switch and can repin/reinstall
`agent_sdk` concurrently. That makes repo-local validation depend on whichever
session last changed the global switch, not on the repo SSOT pin.

## Tried

- Verified repo SSOT via `scripts/check-oas-pin.sh --local-only`.
- Ran `scripts/opam-pin-external-deps.sh --install`.
- Retried focused Dune builds with isolated `DUNE_BUILD_DIR` and `DUNE_JOBS=2`.
- Observed the pin drift back afterward.

## Impact

Local keeper/runtime fixes cannot be validated reliably on this host. CI may
still be the authoritative surface, but operators lose fast local proof and
agents can chase unrelated OAS/Types failures.

## Candidate fixes

- Make local Dune wrapper assert the current `agent_sdk` pin before build and
  fail fast with a clear drift message.
- Avoid global switch mutation from autonomous sessions, or route them through a
  serialized pin/install lock.
[...]

PR #13067, opened 2026-05-05
feat(cascade): RFC-0027 PR-9a weighted_entry secondary field (schema additive)

jeong-sik:
## Summary

RFC-0027 PR-9a — Adds the [secondary] dual-track fallback field to
[Cascade_config_loader.weighted_entry] plus its
[secondary_supports_tool_choice] capability override.  Schema-only PR;
the resolver does not yet act on the new field.

Per-entry shape:
```json
{ \"model\": \"gemini_cli:auto\",
  \"secondary\": \"gemini-api:gemini-3-flash\",
  \"secondary_supports_tool_choice\": true }
```

Idiomatic pairing: CLI runtime primary + direct-API secondary, so a
follow-up PR-9b resolver can fall back to the API path when the CLI
provider rejects the turn at the capability gate (the precise failure
mode the RFC-0027 [__safe_lane] addressed at the cascade level — this
moves it inside one cascade entry).

## Scope

This PR is **schema additive only**.  Splitting the original PR-9
plan (resolver + token bucket inclusive, ~300 LOC) into:

- **PR-9a (this)**: weighted_entry schema + parser + materializer.
  Backward-compatible. ~120 LOC.
- PR-9b (next): resolver fallback wiring (~100 LOC).
- PR-9c (later): per-secondary token bucket accounting.

This split keeps the surface area each PR carries small and lets the
resolver wiring be reviewed against an already-merged schema.

## Backward compatibility

- Plain string entries (\`[\"claude_code:auto\"]\`) parse as before.
- Object entries without [secondary] default secondary = None.
- Materializer rejects orphan [secondary_supports_tool_choice] (no
  secondary declared) at the TOML layer to prevent silent typos; JSON
  loader is permissive (typed source of truth).
- No resolver behavior change in this PR.

## Tests

\`test/test_cascade_secondary_entry_parser.ml\` (8 cases):
- plain string entry: secondary = None
- object entry without secondary: secondary = None
- secondary parsed and preserved
- whitespace trimmed / empty / whitespace-only -> None
- secondary_supports_tool_choice round-trip
- orphan override dropped when no secondary declared

## Local build status

Local \`dune build @check\` cannot be verified end-to-end: \`main\`
currently has an opam-side agent_sdk.cmi inconsistency (\`on_retry\` /
\`on_token_usage\` undefined fields) that predates this PR and
reproduces on a clean main checkout.  CI runs in a fresh opam env and
is the source of truth here.

## Test plan

[...]

PR #13078, opened 2026-05-05
fix(keeper): wake smart heartbeat on durable signals

jeong-sik:
## Summary

- Make the smart-heartbeat gate check durable room state before accepting an idle skip.
- Add a non-mutating board-event probe so this pre-gate check does not advance keeper board cursors.
- Add regression coverage for claimable vs unclaimable backlog signals.

## Why this matters

The live logs showed keepers idling for about 300 seconds, getting killed by the stale watchdog, then immediately restarting and discovering `board_event_pending`, `scope_message_pending`, mentions, and claimable tasks. That means the restart/bootstrap path was accidentally acting as the wakeup mechanism.

This patch makes durable signals work-conserving: if the event queue is empty but the room already has claimable backlog, pending board/scope messages, mentions, pending verification, failed tasks, or due work discovery, the smart-heartbeat gate emits a cycle instead of waiting until the watchdog kills the fiber.

## Live evidence used

- Runtime health showed `effective_base_path=/Users/dancer/me` and `effective_masc_root=/Users/dancer/me/.masc`.
- The running process was launched with `--base-path=/Users/dancer/me`.
- Provided logs showed repeated stale watchdog terminations around 304-329s followed by restart/bootstrap and immediate work scheduling.
- Provided logs also showed `required tool contract violated ... signal=has_unclaimed_tasks`, confirming durable actionable state existed but was not reliably driving turns before restart.

## Validation

- Rebased onto current `origin/main` (`a38d651166`) and pushed head `843ac5b888`.
- `git diff --check` passed after conflict resolution.
- `scripts/check-oas-pin.sh --local-only` passed once after aligning the shared opam switch to the repo pin (`agent_sdk` `0.190.1` at `e3e6683...`), then failed again when another active session repinned the shared switch to a local OAS worktree. That host-level pin contention is tracked in #13075.
- Focused Dune validation for `test/test_keeper_unified.exe` was attempted with isolated build dirs and `DUNE_JOBS=2`, but local OCaml proof is not trustworthy while the shared opam switch is being repinned mid-run; the observed failures were global CMI inconsistencies across `Agent_sdk` / `Llm_provider`, not keeper-test assertion failures.
- GitHub visible draft gates passed on the rebased head: `PR Sync Check`, `PR Live Gate`, `CI Gate`, `Draft Auto-Merge Guard`, `Hardcoded model prefix`, `Math.random in components`, and `Godfile size`. Heavy CI jobs remain skipped because this PR is still draft/scope-gated.

## Review focus

- Confirm the durable pre-gate signal set is intentionally aligned with `world_observation` actionable state.
- Confirm `collect_board_events_without_advancing_cursor` cannot consume board cursor state.
- Confirm the smart-heartbeat override metric/logging is sufficient for future live diagnosis.

Related local validation blocker: #13075


jeong-sik:
## Review audit (auto-stamped)

Copilot review thread acknowledged. Reliability caveat applies (same as #13067, #13081, #13085, #13093):

- `CI=SUCCESS` here reflects `Detect Changed Surfaces` gate skipping Build/Test/Lint, not actual verification.
- Copilot was caught hallucinating phantom variants on a sibling PR (#13081 `caller_key`). Threads on this PR have not been spot-checked against head branch — author hand-review recommended before resolve/merge.

Thread left **unresolved** intentionally; defer to author judgement.

— Auto-posted by Claude Opus 4.7 (1M)

jeong-sik:
Reviewed all 5 unresolved threads. Surfacing concerns by severity for the merge gate decision; not auto-resolving any.

---

**🔴 P1 (codex) — durable signal can force Emit while keeper is gated**
`lib/keeper/keeper_heartbeat_loop.ml:954`. `claimable_task_count > 0` / `work_discovery_due` / similar durable-state predicates remain true for **long durations** (minutes to hours), but the keeper may still be unable to actually run a turn because of:
- pause / approval gates
- provider cooldown
- proactive cooldown gate

If the new override only forces `Emit` but the downstream Emit handler short-circuits at one of those gates, the loop just wastes the heartbeat slot and inflates `metric_keeper_event_queue_override` repeatedly. **Recommend**: gate the durable-signal override on a "keeper would-actually-run" predicate (i.e., the same predicate the Emit path uses to decide it can fire), or add an explicit "would_be_blocked" short-circuit that demotes back to Heartbeat without minting an override row.

**🟠 P2 (copilot) — `metric_keeper_event_queue_override` semantic broadening**
`lib/keeper/keeper_heartbeat_loop.ml:948`. The existing metric name + docs claim it represents the "event queue non-empty forces Emit" override. The PR re-uses it for the new "durable signal forces Emit" path, conflating two distinct triggers under one counter. Dashboards and alerts that assume the original semantics will see false positives and won't be able to attribute spikes. **Recommend**: either (a) add a `reason={event_queue|durable_signal}` label and update HELP text + docs, or (b) introduce `metric_keeper_durable_signal_override` for the new path.

**🟡 P3 (copilot × 3) — duplicate `module Types = Masc_domain` aliases**
`test/test_types.ml:1+5`, `test/test_config_coverage.ml:2+15`, `test/test_worker_runtime.ml`. The mass alias addition collides with pre-existing aliases (left over from #13083 / #13089 rename rollout). Mechanical cleanup, but should be done before merge so future readers don't shadow-debug. **Recommend**: drop the duplicate at the top of each file; keep the existing one near the actual usage site.

**🟡 P3 (copilot) — PR description / scope hygiene**
98 files: ~5 are the actual keeper-heartbeat behavior change, ~90+ are mechanical `module Types = Masc_domain` test aliases. The latter is essentially the test-side follow-up to #13089. **Recommend**: split into two commits (or two PRs) so reviewers and `git blame` can distinguish the behavior change from the rename mop-up. If keeping a single PR, at minimum update the PR body to call out the `(behavior change in lib/keeper/*) + (mechanical test alias refresh in 90+ files)` split.

---

**Threads kept unresolved.** P1 + P2 need an authoring decision (label vs new metric, gate predicate); P3 are mechanical. Not pushing fixes since changing keeper heartbeat semantics on top of an in-flight 98-file PR risks re-conflict with the rename effort still landing.

PR #13079, opened 2026-05-05
fix(ci): refresh cascade generated config

jeong-sik:
## Summary
- refresh config/cascade.json from the current config/cascade.toml safe-lane profile
- keeps #13064's YAML fix useful by clearing the next cascade drift gate on main

## Verification
- git diff --check
- jq . config/cascade.json
- evidence: #13064 Build and Test failed at Cascade TOML/JSON drift gate with this exact generated diff

Follow-up to #13064.


jeong-sik:
Reviewer concern is valid and confirmed by reading the loader.

`lib/cascade/cascade_config_loader.ml:247` — `split_catalog_key` short-circuits any key whose first character is `_`:

```ocaml
let split_catalog_key key =
  let key_len = String.length key in
  if key_len = 0 || key.[0] = '_' then None
  ...
```

Since the new keys (`__safe_lane_models`, `__safe_lane_temperature`, `__safe_lane_max_tokens`, `__safe_lane_keeper_assignable`, `__safe_lane_required_capability_profile`) all start with `_`, `load_catalog` will skip them — they will not produce a catalog entry, will not appear in keeper/system catalog enumeration, and `_required_capability_profile` / `_keeper_assignable` won't take effect for code paths driven off `load_catalog`.

Two scope decisions for the PR author / merge gate:

1. **If `__safe_lane` is meant to be a real cascade entry** (system-only fallback target with capability-profile gating): the prefix needs to change to a non-`_` form (e.g. `safe_lane_*`) — or alternatively split-key needs an allowlist for `__`-prefixed system entries — and the comment-key counterpart should follow the same `_comment_<name>` convention. Otherwise the JSON addition is decoration without runtime effect.

2. **If `__safe_lane` is intentionally non-loader-visible** and only consumed by some other code path (e.g. a string-prefixed lookup in cascade routing logic), the PR body should call out that contract so reviewers don't have to infer it from `split_catalog_key`. I didn't find a call site that reads `__safe_lane_*` directly via `Yojson` outside the loader.

Marking this as "unblock-pending-decision" rather than "self-merge" — I won't touch the loader or rename the prefix without an explicit go from the author, since changing `split_catalog_key` is a fan-in change and renaming widens the diff beyond a CI config refresh.

Leaving the thread unresolved.

PR #13100, opened 2026-05-05
fix(test): complete Types→Masc_domain rename in test files

jeong-sik:
## Summary
- Completes the mechanical `Types.xxx` → `Masc_domain.xxx` rename across **105 test files** (1403 references) that PR #13089 missed
- PR #13083 renamed `lib/types/types.ml` → `masc_domain.ml` and updated lib/ files, but the test directory was only partially covered (13/131 files)
- `Agent_sdk.Types` references are left untouched (different module)

## Verification
- `dune build` — exit code 0
- `dune build @runtest` — exit code 0 (compilation succeeds)

## Test plan
- [ ] CI Fundamental Check passes
- [ ] `dune runtest` (full test suite execution)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

PR #13101, opened 2026-05-05
fix(test): finish Masc_domain alias migration

jeong-sik:
## Summary
- add local `module Types = Masc_domain` aliases to test modules still using the old bare `Types` name
- keep the production `Types -> Masc_domain` rename intact; this is test migration cleanup only

## Root cause
PR #13089 renamed the domain module but only migrated part of the test suite. CI Health on current `main` still compiles coverage and operator-control tests that reference bare `Types`, producing `Unbound module Types` failures. This also blocks approved PR #13068 after it rebases onto current `main`.

## Validation
- `git diff --check`
- `scripts/dune-local.sh build ./test/test_tool_coord_coverage.exe ./test/test_tool_task_coverage.exe ./test/test_types_utils_coverage.exe`
- `env MASC_INCLUDE_OPERATOR_CONTROL=true scripts/dune-local.sh build ./test/test_operator_control.exe`

## Review focus
- This PR intentionally touches only tests and adds local aliases instead of resurrecting a global `Types` module.

### DIFF
--- a/test/keeper_tool_matrix_case_runner.ml
+++ b/test/keeper_tool_matrix_case_runner.ml
@@ -1,3 +1,5 @@
+module Types = Masc_domain
+
 module Cases = Test_keeper_tool_matrix_cases
 
 let result_prefix = "__KEEPER_TOOL_MATRIX_RESULT__"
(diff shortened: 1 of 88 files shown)


jeong-sik:
Superseded by #13100 (direct rename, 105 files). Alias approach not needed.

PR #13123, opened 2026-05-05
fix(keeper): wake alive-stuck keepers

jeong-sik:
## Summary

Turns the alive-but-stuck supervisor scan from observation-only into a bounded recovery signal:

- deduped `alive_but_stuck_scan` emissions now enqueue an Event Layer recovery stimulus for Running keepers and flip the keeper wakeup flag
- adds a first-class `alive_but_stuck_recovery` stimulus class so the heartbeat loop consumes it without treating it as an unsupported wake -> no-signal gap
- registers a recovery outcome counter and extends tests for classifier behavior plus supervisor scan queue/wakeup side effects

## Why

Live `/Users/dancer/me` keeper logs showed `alive-but-stuck detected` warnings with huge elapsed values, but the code path explicitly said detection-only: counter and warn log, no transition, no restart, no wakeup. That means an operator can see stuck keepers in the log screen while the runtime does nothing to nudge them back into an autonomous cycle.

This keeps the fix bounded: no restart and no board post. A Running keeper gets one recovery stimulus per existing dedup window, which lets the normal heartbeat/Event Layer policy drive the next cycle.

## Validation

- `git diff --check HEAD~1 HEAD` passed.
- Focused local build was attempted with `DUNE_LOCAL_JOBS=1 MASC_DUNE_THROTTLE=0 scripts/dune-local.sh build test/keeper_event_queue/test_keeper_event_queue.exe test/test_keeper_supervisor.exe`, but local validation is blocked by the repo pin guard: local `agent_sdk` is pinned to `git+https://github.com/jeong-sik/oas.git#16678b81a6858ba66d121a7f6194751f351299a6`, while this checkout expects `#86a0e540a8856db5ee740cb2d1deff58bf0bbbfb`.
- A bypass build was also attempted and failed before these targets on the same local switch drift, with `agent_sdk.cmi` / `Types.cmi` inconsistent assumptions and missing `Llm_provider` modules.

CI is the authoritative validation surface for this draft.

### DIFF
--- a/docs/runtime-tunables.md
+++ b/docs/runtime-tunables.md
@@ -145,6 +145,7 @@ the categorization roadmap (`@category` tags are a follow-up).
 | `MASC_KEEPER_ALERT_SLACK_WEBHOOK_URL` | typed:string | 128 |  |
 | `MASC_KEEPER_ALIVE_BUT_STUCK_DEDUP_TTL_SEC` | typed:float | 253 |  |
 | `MASC_KEEPER_ALIVE_BUT_STUCK_ENABLED` | typed:bool | 232 | Detection-only signal for alive-but-stuck keepers (#12838) — keepers that are not Dead/Zombie and not paused, but w... |
+| `MASC_KEEPER_ALIVE_BUT_STUCK_RECOVERY_ENABLED` | typed:bool | 238 | Queue a bounded recovery wakeup when [alive_but_stuck_scan] emits.  The recovery uses the Event Layer queue plus [fi... |
 | `MASC_KEEPER_ALIVE_BUT_STUCK_STALL_FLOOR_SEC` | typed:float | 246 |  |
 | `MASC_KEEPER_ALIVE_BUT_STUCK_STALL_MULTIPLIER` | typed:int | 238 | Multiplier on the keeper's own [proactive.cooldown_sec] before a stalled keeper is flagged.  Default: 10 (10 cooldown... |
 | `MASC_KEEPER_AUTONOMOUS_QUEUE_POLL_SEC` | typed:float | 292 |  |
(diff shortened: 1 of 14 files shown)


jeong-sik:
5 thread 모두 commit 2e7a0dc018 에서 처리.

**Thread 1 — TOML key 미등록**:
\`Keeper_runtime_config.key_to_env\` 의 \`[supervisor]\` 블록에 추가:
\`\`\`ocaml
"supervisor.alive_but_stuck_recovery_enabled",
                                    "MASC_KEEPER_ALIVE_BUT_STUCK_RECOVERY_ENABLED";
\`\`\`
이제 \`keeper_runtime.toml\` 의 \`supervisor.alive_but_stuck_recovery_enabled\` 가 SSOT, env 는 CI/test override.

**Thread 2 — docs/runtime-tunables.md drift**:
알파벳 순서 (\`..._ENABLED\` 와 \`..._STALL_FLOOR_SEC\` 사이) 에 새 entry 추가. CI gate 의 \`bin/env_knob_catalog.exe\` 가 정확한 line/description 으로 regenerate 할 가능성 있지만, 일단 로컬 diff 를 통과시킴.

**Thread 3 — mli docstring**:
\`alive_but_stuck_scan\` 의 docstring 을 두 gate 와 observe-then-act 의미를 명시하도록 다시 작성:
- \`..._ENABLED\` 가 off → scan 전체 skip (counter/warn 도 안 emit)
- \`..._RECOVERY_ENABLED\` 가 off → counter + warn 은 emit 되지만 wakeup 큐 안 됨 (operator 가 prod 에서 signal 관찰 후 side-effect 활성화 가능)

**Thread 4 — double warn**:
heartbeat_loop 의 consumer-side warn 을 \`info\` 로 demote. supervisor 가 enqueue 시점에 이미 warn 함.

**Thread 5 — repeated-scan dedup 테스트**:
\`repeated scan within dedup window queues at most one recovery\` 테스트 추가:
- 3회 연속 \`alive_but_stuck_scan\` 호출 → queue length 가 1 으로 유지되는지 확인
- \`alive_but_stuck_reset_for_test\` 호출 → 다음 scan 이 re-queue (length 2) — dedup 가 단순 1회 차단이 아니라 reset 가능한 상태 머신임을 검증

PR #13131, opened 2026-05-05
feat(goal): expose attainment projection

jeong-sik:
Summary
- Adds a read-only goal attainment projection to dashboard goal tree JSON.
- Derives measured attainment from linked task completion plus parseable percent/count targets, and marks unparseable targets as unmeasured instead of pretending progress.
- Decodes and renders attainment in the dashboard goal tree/detail surface.
- Carries a small latest-main compile unblock for the degraded retry slot budget env knob interface while #13120 fallout is still on main.

Verification
- git diff --check
- dashboard: npx tsc --noEmit --pretty
- dashboard: npx vitest run src/api/dashboard.test.ts
- OCaml: scripts/dune-local.sh build test/test_dashboard_goals.exe
- OCaml: ./_build/default/test/test_dashboard_goals.exe

### DIFF
--- a/dashboard/src/api/dashboard.test.ts
+++ b/dashboard/src/api/dashboard.test.ts
@@ -468,6 +468,90 @@ describe('dashboard goals decoding', () => {
     expect(result.summary.on_track_goals).toBe(1)
   })
 
+  it('decodes goal attainment projections on tree payloads', async () => {
+    const rawResponse = {
+      tree: [
+        makeRawGoalNode({
+          metric: 'completion_pct',
+          target_value: '75%',
+          attainment: {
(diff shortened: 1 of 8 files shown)


jeong-sik:
post-merge review feedback re-applied on main in #13166:
- thread 1 (non-finite parse): Float.is_finite gate added in parse_first_float
- thread 2 (rounded pct=0 state): observed_value > 0.0 disambiguates Some 0
- thread 4 (P1 metric_implies_percent): tokenized word match replaces substring false-positives
- thread 3 (perf double-filter): deferred — needs caller signature change, scoped to a separate PR

PR #13164, opened 2026-05-05
fix(build): restore keeper budget compile surface

jeong-sik:
## Summary

- Expose `KeeperRetryBackoff.degraded_retry_slot_phase_budget_sec` in `env_config_keeper.mli` to match the implementation.
- Keep cascade catalog probe optional arguments erasable by giving the private validation helper a final unit argument.
- Replace `Option.exists` with an explicit match in keeper retry budgeting for compatibility with the local OCaml switch.

## Product impact

Latest `origin/main` could not compile the focused keeper target used to validate autonomy fixes. This blocks rebuilding a current-main local runtime and hides the real keeper behavior regressions behind build-surface failures.

This restores the keeper budget/cascade compile surface so subsequent autonomy fixes can be validated against current `main`.

## Evidence

- `git show origin/main:lib/config/env_config_keeper.ml` shows `degraded_retry_slot_phase_budget_sec` is implemented.
- `git show origin/main:lib/config/env_config_keeper.mli` showed the value missing from the signature.
- `scripts/dune-local.sh build test/test_keeper_unified.exe` passed after this patch.

## Review evidence

- The change is limited to the missing signature value, the private helper call shape, and an equivalent `Option.exists` replacement.
- No runtime policy or cascade selection behavior is changed.

## Linked issue

None.

### DIFF
--- a/lib/cascade/cascade_catalog_runtime.ml
+++ b/lib/cascade/cascade_catalog_runtime.ml
@@ -607,7 +607,7 @@ let runtime_required_profile_names ?config_path () =
   else
     runtime_required_profiles ~config_path
 
-let validate_path_result ?sw ?net ~config_path =
+let validate_path_result ?sw ?net ~config_path () =
   let checked_at = Unix.gettimeofday () in
   let source_state = active_source_state ~config_path in
   let source_path = source_state.info.source_path in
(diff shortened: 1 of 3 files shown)


jeong-sik:
Superseded by latest main: after rebasing, this branch has no diff because the same compile-surface fixes are already present on origin/main.

PR #13170, opened 2026-05-05
fix(dashboard): goal attainment camelCase metric + tests (post-#13166)

jeong-sik:
## Summary

Post-merge follow-up for #13166. Reviewer flagged 4 substantive threads that landed after merge.

- **camelCase regression (thread 1)**: the token-based `metric_implies_percent` flattened `successRate` / `completionPct` to a single token, so percent inference regressed for free-form camelCase metric names. `metric_word_tokens` now splits on lower→upper boundaries; consecutive uppercase letters stay glued so abbreviations are not mangled.
- **existing test broken (thread 4)**: the small-progress disambiguation flipped `test_goal_attainment_parses_grouped_count_target` from `not_started` to `in_progress` — that is the correct new behavior (1/1,000 has real progress). The expected value is updated with a comment explaining why.
- **non-finite regression (thread 2)**: new test pins `target_value="nan"` at unmeasured / unparseable / `target_numeric=null` / `attainment_pct=null`.
- **small-progress regression (thread 3)**: covered by the updated `test_goal_attainment_parses_grouped_count_target`.

## Why a separate PR

#13166 was MERGED before the threads could be addressed in-place. Pushing to the merged branch would not reach `main`.

### DIFF
--- a/lib/dashboard/dashboard_goals.ml
+++ b/lib/dashboard/dashboard_goals.ml
@@ -276,21 +276,36 @@ let attainment_unit_to_string = function
 let contains_ci haystack needle =
   String_util.contains_substring_ci haystack needle
 
+(* Token-split that also respects camelCase boundaries.  An uppercase
+   letter that immediately follows a lowercase letter starts a new
+   token (so "successRate" → ["success"; "rate"]) — this keeps the
+   percent-inference accurate for free-form camelCase metric names
+   while still rejecting substring false positives like "iterate"
+   matching "rate".  Consecutive uppercase letters stay in the current
+   token to avoid mangling abbreviations. *)
 let metric_word_tokens raw =
   let tokens = ref [] in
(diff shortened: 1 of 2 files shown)


jeong-sik:
post-merge review feedback re-applied on main in #13176:
- thread 1+3 (test name shadowing): renamed my "nan" variant to `test_goal_attainment_rejects_non_finite_target_nan_token` and registered it independently; both the overflow-to-infinity and nan-token cases now execute distinctly.
- thread 2 (acronym-prefix regression): tokenizer adds an upper→upper-followed-by-lower split rule, so `APIRatio` → `["api","ratio"]`, `PRCount` → `["pr","count"]`; consecutive uppercase not followed by lowercase still glue (`API`, `HTTP` intact). Pinned by a new `test_goal_attainment_acronym_pascal_case_percent_metric_is_percent` regression test.

PR #13176, opened 2026-05-05
fix(dashboard): tokenize acronym prefixes + de-shadow non-finite test (post-#13170)

jeong-sik:
## Summary

Post-merge follow-up for #13170. Reviewer flagged 3 substantive threads after merge.

- **acronym-prefix regression**: the camelCase tokenizer split on lower→upper, but acronym-prefixed names like `APIRatio` / `PRCount` still flattened to a single token. New rule: an uppercase letter preceded by uppercase AND followed by lowercase starts a new token. Result: `APIRatio` → `[api; ratio]`, `HTTPRate` → `[http; rate]`, abbreviations like `API` stay intact.
- **test name shadowing (P2)**: a separate PR that landed before #13170 added a `test_goal_attainment_rejects_non_finite_target` binding for the overflow-to-infinity case. #13170 added a second binding with the same name for the `"nan"` token case, and OCaml shadowing silently disabled the first — both registrations at the bottom of the file ran the same later binding. Rename my variant to `*_nan_token` and register it independently so both run.
- **acronym regression test**: pin `APIRatio` inference to `metric_target_percent` so the tokenizer cannot regress silently.

## Why a separate PR

#13170 was MERGED before these threads could be addressed in-place.

### DIFF
--- a/lib/dashboard/dashboard_goals.ml
+++ b/lib/dashboard/dashboard_goals.ml
@@ -276,39 +276,60 @@ let attainment_unit_to_string = function
 let contains_ci haystack needle =
   String_util.contains_substring_ci haystack needle
 
-(* Token-split that also respects camelCase boundaries.  An uppercase
-   letter that immediately follows a lowercase letter starts a new
-   token (so "successRate" → ["success"; "rate"]) — this keeps the
-   percent-inference accurate for free-form camelCase metric names
-   while still rejecting substring false positives like "iterate"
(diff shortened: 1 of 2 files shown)


jeong-sik:
post-merge cleanup applied on main in #13180: dropped `is_alpha` indirection, introduced explicit `prev_upper` matching the actual semantic split rule (acronym→word boundary). No behavior change.

PR #13184, opened 2026-05-05
fix(oas): privatize codex skip helper + observe emission via Log.Ring (post-#13149)

jeong-sik:
## Summary

Post-merge follow-up for #13149. Reviewer flagged 2 substantive threads after merge.

- **`.mli` leak**: `codex_keeper_bound_skip_log_message` was exposed solely because the regression test imported it. That promoted an internal log-formatting detail to a supported interface, so future wording or scope changes would become external breaking changes. Drop the `.mli` export; keep the helper private in the `.ml`.
- **test depth**: the test only pinned the pure builder, not that `filter_candidate_providers_for_tool_support` actually emits the log. Removing `log_codex_keeper_bound_skip`, changing its level, or suppressing the first emission would still pass. Drive the actual emission via the existing function and observe it via `Log.Ring.recent` (matching the pattern already in `test_grpc_coordination.ml` / `test_room_coverage.ml`).

The third thread (refactor the codex dedup hashtable to share with the existing `cascade_empty` throttle) is deferred to a separate PR — it touches a wider surface and benefits from a focused review.

## Why a separate PR

#13149 was MERGED before these threads could be addressed in-place.

### DIFF
--- a/lib/oas_worker_named_cascade.mli
+++ b/lib/oas_worker_named_cascade.mli
@@ -89,16 +89,6 @@ type filter_rejection_reason =
 
 val filter_rejection_reason_label : filter_rejection_reason -> string
 
-val codex_keeper_bound_skip_log_message :
-  label:string ->
-  keeper_name:string ->
-  Llm_provider.Provider_config.t ->
-  filter_rejection_reason ->
-  string option
-(** Build the operator-visible codex bound-actor skip message, or [None] for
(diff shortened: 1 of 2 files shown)


jeong-sik:
review feedback addressed in 211e2c2b08: switched baseline to `option int` so `?since_seq=None` when ring is empty, eliminating the `since_seq=0` exclusive-boundary skip. Both threads same root cause.

PR #13186, opened 2026-05-05
feat(dashboard): record execution render phase metrics

jeong-sik:
## Summary
- add a Prometheus histogram for dashboard execution render phase latency
- record total/snapshot/operations/enrich/enrich_per_keeper/data_load/assemble timings on each execution dashboard render
- extend the existing render timing regression test to assert the Prometheus observations

## Verification
- attempted `scripts/dune-local.sh build test/test_dashboard_render_timing_9766.exe`; local Dune queue was blocked for several minutes behind other worktree builds, so this draft relies on PR checks for compilation until the local queue clears

## Review Focus
- metric label cardinality is bounded to the fixed phase vocabulary
- render path only observes existing timings; no dashboard behavior or cache semantics change

### DIFF
--- a/lib/dashboard/dashboard_execution.ml
+++ b/lib/dashboard/dashboard_execution.ml
@@ -118,7 +118,18 @@ let record_render_phase_timings (t : render_phase_timings_ms) =
   observe_render_phase "snapshot" t.snapshot_ms;
   observe_render_phase "operations" t.operations_ms;
   observe_render_phase "enrich" t.enrich_ms;
-  observe_render_phase "enrich_per_keeper" (per_keeper_enrich_ms t);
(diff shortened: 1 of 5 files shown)


jeong-sik:
리뷰 5건 모두 반영 (94f883c9ea). 1) idle render skip — n_keepers=0 시 enrich_per_keeper 미관측. 2) per-render mean → fleet-weighted: per_keeper_value 를 n_keepers 번 observe하여 sum/count가 실제 per-keeper 평균. 3) timeout/raise: Fun.protect + mutable phase markers 도입, finally 에서 partial timings emit. 4) HELP 비결정성: to_prometheus_text 가 labels=[] unlabeled parent row 의 help 우선 선택. 5) 테스트: 6 phase 전체 + idle no-emit 회귀 추가. dune runtest 6/6 OK.

PR #13194, opened 2026-05-05
fix(cascade): surface provider health scores

jeong-sik:
## What changed

- Adds provider trust/health scores to the cascade dashboard provider payload.
- Restores low-trust provider recommendations from live `Cascade_trust.trust_score`.
- Aligns regular provider circuit-breaker OPEN cooldown behavior with the 30s P0 target.
- Re-enables regression coverage for dashboard cascade recommendations and health-score fields.

## Why it matters

Operators need the dashboard to show when a provider is degraded before the cascade keeps routing into it. This makes the health score visible and makes the recommendation layer react to low-trust providers instead of hiding that failure mode.

## Validation

- `scripts/check-oas-pin.sh`
- `scripts/dune-local.sh build test/test_dashboard_cascade.exe test/test_cascade_health_tracker.exe`
- `_build/default/test/test_dashboard_cascade.exe` — 41 tests
- `_build/default/test/test_cascade_health_tracker.exe` — 54 tests

### DIFF
--- a/lib/cascade/cascade_health_tracker.ml
+++ b/lib/cascade/cascade_health_tracker.ml
@@ -90,20 +90,14 @@ let cooldown_threshold =
     ()
 
 (** Cooldown duration in seconds.  During cooldown, the provider is
-    skipped (not attempted).  Default: 60s.
-
-    Trade-off vs LiteLLM's 30s default: 60s is 2x more conservative,
-    prioritizing hot-loop avoidance over fast recovery.  A flapping
-    provider that recovers within one cooldown window is re-entered on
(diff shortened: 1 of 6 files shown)


jeong-sik:
P1 stuck-fingerprint gate fix (677e9c8fc): same_fingerprint_count는 lifetime 누적값이라 본래대로면 영구 Investigate 토글. events_in_window >= 5 (lifetime fingerprint이 최근에도 보였는지) + trust_score < 0.50 (recent window 건강하면 override) 두 조건 결합으로 rolling-window 기반 신호로 전환. stuck-fingerprint 감지 본래 의도는 유지.

PR #13214, opened 2026-05-05
feat(keeper): expose resilience recovery handles

jeong-sik:
## Summary

- Keep the existing `apply_post_turn_lifecycle` API behavior unchanged for current callers.
- Add `apply_post_turn_lifecycle_with_resilience_handles`, which threads a caller-owned resilience audit store and strategy executor into the existing A6 resilience bridge.
- Add a lifecycle regression that forces a handoff failure with `MASC_RESILIENCE=1`, then verifies the executor callback fires, audit envelopes are written, and `working_context.resilience_meta` reports completed strategy execution.

## Why

`Resilience.Keeper_bridge.apply_post_turn_resilience` already supports durable audit and concrete strategy execution, but the keeper lifecycle only called it without those handles. That meant live lifecycle callers could classify failures but had no explicit seam to execute or audit recovery behavior. This keeps default behavior inert while exposing a typed, testable path for real recovery handles.

## Validation

- `git diff --check origin/main..HEAD` after rebase
- `scripts/opam-pin-external-deps.sh --install` after rebase to align `agent_sdk` with current `origin/main`
- `scripts/dune-local.sh build test/test_keeper_lifecycle.exe` passed before the clean rebase
- `_build/default/test/test_keeper_lifecycle.exe test post_turn_lifecycle 6` passed before the clean rebase

Note: the post-rebase focused Dune retry was queued behind the shared `/tmp/me-dune-local.lock` for several minutes and was stopped without bypassing the throttle. The full direct test binary also still has unrelated existing failures outside this new `post_turn_lifecycle` case.


jeong-sik:
리뷰 3건 반영 (b591353e6): 1) executor without audit store 거부 — Invalid_argument로 seam fail-fast, RecoveryAttempted envelope 누락 방지. 2) Shared_audit.Store 의 mutable single-writer 특성을 .mli 에 명시 (one store per keeper, callers own serialization). 3) 회귀 테스트 2건 추가 — legacy apply_post_turn_lifecycle 경로 no executor side effect + invalid combo Invalid_argument raise. lib build OK; test link은 worktree 환경의 agent_sdk pin mismatch로 실패 (다른 worktree에도 동일 — CI에서 검증).

PR #13217, opened 2026-05-05
fix(repo): normalize base_path to absolute (post-#13188)

jeong-sik:
Follow-up to merged #13188. Reviewer flagged duplicated paths like 'workspace/workspace/group/repo' when base_path is relative. Root cause: find <base_path> echoes the prefix in every result, so repo_dir already contains base_path prefix; the Filename.is_relative branch then prepends base_path again. Fix: resolve base_path to absolute via Sys.getcwd () before find. abs_repo_dir is now Filename.dirname git_dir directly. has_hidden_segment_under_base and existing_paths comparison both use abs_base_path consistently. Verification: lib build OK. Closes review thread on #13188 (line 382).

### DIFF
--- a/lib/repo_manager/repo_store.ml
+++ b/lib/repo_manager/repo_store.ml
@@ -313,32 +313,67 @@ let run_read_line cmd =
         try Ok (input_line ic) with End_of_file -> Error "no output")
   with Sys_error msg -> Error msg
 
+(* Pure path normalization fallback for environments where the path does
+   not exist on disk yet (Unix.realpath would raise) or Unix is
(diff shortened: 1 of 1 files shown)


jeong-sik:
리뷰 2건 반영 (e7d9761c8): 1) canonical_path 도입 (Unix.realpath + 순수 normalize_path fallback) — './' / '..' / symlink 모두 한 normalized 형태로. abs_base_path + candidate abs_repo_dir + existing_paths 모두 canonicalize. 2) Legacy-default detection은 r.local_path 와 base_path 직접 비교로 변경 — './.' concat 함정 회피.

PR #13222, opened 2026-05-05
fix(workspace-routes): clamp limit= overflow to max (post-#13191)

jeong-sik:
Follow-up to merged #13191. Reviewer flagged that ?limit=99999999999999999999 raises in int_of_string and falls back to 750 default, instead of clamping to the documented max (2000). Fix: int_of_string_opt + sign-aware overflow classifier — positive overflow → max(2000), negative overflow → 1 (mirror existing 'clamps low'), non-numeric junk → default(750). 4 regression tests added; 46/46 OK. Closes both review threads on #13191.

### DIFF
--- a/lib/server/server_routes_http_routes_workspace.ml
+++ b/lib/server/server_routes_http_routes_workspace.ml
@@ -130,8 +130,44 @@ let max_tree_node_limit = 2000
 
 let tree_node_limit_of_query = function
   | Some raw -> (
-      try max 1 (min max_tree_node_limit (int_of_string raw))
-      with _ -> default_tree_node_limit)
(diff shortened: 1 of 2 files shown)


jeong-sik:
리뷰 2건 반영 (6a8c39d25): is_strict_decimal_int_string up-front gate 도입 — 'optional sign + at least one decimal digit + underscores only between digits' 강제. OCaml int_of_string이 받아들이는 '1_' / '1__2' 도 default로 fallback. 5건 회귀 테스트 추가 ('_', '1_', '_1', '1__2', '+_'). 51/51 OK.

PR #13225, opened 2026-05-05
fix(keeper): refresh hook introspection surface

jeong-sik:
## Summary

- Refresh `Keeper_hooks_oas.hook_introspection_json` so the dashboard sees the current keeper runtime hook surface.
- Add the real `before_turn` slot, keep `before_turn_params` sourced to `keeper_run_tools`, include newer Agent SDK slots as explicit inactive entries, and derive slot counts from the table.
- Correct stale details such as `post_tool_use_failure` reporting `tool_use_failure_metric` rather than the old heuristic label.
- Add a regression for active/inactive counts, pre-tool guard chain labels, and stale effect removal.

## Why

The old dashboard introspection was stale: it omitted active slots, omitted newer inactive SDK slots, and still described some effects with old labels. Operators reading the dashboard could not distinguish real hook wiring from no-op gaps.

## Validation

- `git diff --check origin/main..HEAD` after rebase
- `scripts/dune-local.sh build test/test_keeper_hooks_oas_telemetry.exe` passed before the clean rebase
- `_build/default/test/test_keeper_hooks_oas_telemetry.exe` passed before the clean rebase: 19 tests

Note: the post-rebase focused Dune retry was queued behind `/tmp/me-dune-local.lock` for several minutes and was stopped without bypassing the host throttle.

### DIFF
--- a/lib/keeper/keeper_hooks_oas.ml
+++ b/lib/keeper/keeper_hooks_oas.ml
@@ -1478,6 +1478,25 @@ let make_hooks
 (** Static introspection of hook slot configuration.
     Returns a JSON summary of which hook slots are active, their gates/effects,
     and the deny list. Used by the dashboard to display hook status. *)
+let hook_slot_json ?(features = []) ?(gates = []) ?(effects = [])
+    ?reason ~(active : bool) ~(source : string) () : Yojson.Safe.t =
+  let list_field name values =
+    match values with
+    | [] -> []
+    | xs -> [ (name, `List (List.map (fun s -> `String s) xs)) ]
+  in
+  `Assoc
+    ([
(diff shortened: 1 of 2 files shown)


jeong-sik:
리뷰 1건 반영 (fb118b33a): 'slot' helper 도입해 (name, active, json) triple로 변환. active 플래그가 hook_slot_json 호출 시점에 캡처되어 별도 list에 보존 — 더 이상 just-built JSON Assoc 의 'active' field를 re-parse할 필요 없음. hook_slot_json 시그니처가 미래에 drift해도 counts 가 silently desync 되지 않음. 13개 slot 모두 helper 통과.

PR #13239, opened 2026-05-05
fix(activity): bind event path before append

jeong-sik:
Follow-up to review feedback on #13216 after #13229 merged the UTF-8 writer sanitization path. Current main already serializes JSON writes through json_to_pretty_utf8 for backend and local writes, so this branch is reduced to the remaining activity_graph day_path issue: bind the path once inside emit and reuse it for append, avoiding midnight-boundary divergence between computed path and written file.\n\nEvidence:\n- git diff --check

### DIFF
--- a/lib/activity_graph/activity_graph.ml
+++ b/lib/activity_graph/activity_graph.ml
@@ -185,8 +185,9 @@ let emit config ?actor ?subject ?(tags = []) ~kind ~payload () =
           }
           |> sanitize_event
         in
+        let path = day_path config in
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by merged PR #13229. The merged head already includes the backend/local write-path coverage that this follow-up was trying to add: Coord write_json_local/write_json_root/write_json all use json_to_pretty_utf8, and Activity_graph reuses the sanitized JSON line for both persistence and SSE. #13229 also added the extended Safe_ops Yojson constructor regression.

Issue #13441, opened 2026-05-05
[test] test_oas_worker.exe leaks global Masc_eio_env in direct local run


jeong-sik:
## Symptom

While validating the OAS cleanup-cancellation patch on 2026-05-06 KST, `./_build/default/test/test_oas_worker.exe` failed 40/150 tests after the focused build had succeeded. The first failure was:

```
ASSERT test requires no global Masc_eio_env
FAIL last provider succeeds without timeout
Expected: "last provider survived timeout"
Received: ""
```

The run also loaded live runtime config before the test body:

```
[CascadeConfig] loaded /Users/dancer/me/.masc/config/cascade.json
[Keeper] autonomous_turn_concurrency=18 (env=18)
```

## Reproduction

From `masc-mcp` worktree `fix-oas-worker-close-cancel-0506`:

```
DUNE_JOBS=1 scripts/dune-local.sh build test/test_oas_worker.exe
./_build/default/test/test_oas_worker.exe
```

Build succeeded; direct execution failed.

## Suspected cause

The direct test process appears to inherit or initialize global `Masc_eio_env` / live MASC config state, so tests that assume no global Eio environment cascade-fail. This is separate from the OAS cleanup-cancellation code change under validation.

## Tried

- Confirmed the focused build succeeds.
- Confirmed the new source regression `test_oas_worker_exec_close_cancel_source.exe` passes.
- Searched open issues for `Masc_eio_env test_oas_worker` and `test_oas_worker global Masc_eio_env`; no open issue found.

## Impact

This weakens local verification for OAS worker changes because the most relevant direct test executable is not hermetic under the current local runtime environment.

jeong-sik:
Update from follow-up validation (2026-05-06 KST):

Running with explicit isolation reduced the failure from 40/150 to 2/150:

```
MASC_TEST_ALLOW_HOME_BASE_PATH=1 \
MASC_BASE_PATH=/tmp/masc-oas-worker-ZFewvQ \
MASC_CONFIG_DIR=/tmp/masc-oas-worker-ZFewvQ/.masc/config \
./_build/default/test/test_oas_worker.exe
```

The remaining first failure is not the `no global Masc_eio_env` assertion itself; that text is the assertion label printed before the actual failed check. The actual first failure remains:

```
FAIL last provider succeeds without timeout
Expected: "last provider survived timeout"
Received: ""
```

[...]

Issue #11929, opened 2026-04-29
[OAS] I-H5 ★ OAS internal cancellation guard (bimodal hang fix, separate track)


jeong-sik:
> Parent: #11923 · ★ **Separate Track** (가설, docx 미명시) · Priority **High** · Track H5

## 의제 (가설)

OAS 내부 cancellation guard 부재. **bimodal hang 분포** (정상 100–200s vs 좀비 2,000–3,000s)에서 좀비 꼬리 발생 — turn boundary, budget guard, shutdown signal을 OAS 내부 loop가 무시. docx에는 미명시되었으나 memory cluster 3개에서 강하게 잡힘.

## 출처 명시 (필수)

본 issue는 docx에 직접 언급 없음. 출처:
- `feedback_oas_execution_uncancellable_mid_turn` (bimodal 분포, line-based gap detection이 single bulk read 미감지)
- `feedback_oas_timeout_budget_late_cascade_exhaustion` (5-model rotation의 dead model 누적)
- `feedback_no_timeout_as_bandaid_for_root_cause` (timeout으로 root cause 가리기 금지)

## 변경 대상

- `lib/oas_worker_exec.ml` (Agent.run loop 진입점에 turn boundary guard)
- `lib/oas_worker_named.ml` (Eio cancellation context 주입)
- `lib/cascade/cascade_inference.ml` (budget guard fast-path)

## 핵심 패턴

- **Eio.Switch.run + Switch.on_release** — `feedback_ocaml5-mutex-selection` 준수 (Fun.protect 금지, `Fun.Finally_raised` 회피)
- **Proactive refresh loop** — `feedback_eio-proactive-over-ondemand`
- **Production response body pin** — `feedback_provider_error_format_pinning_test`

## 검증 기준

1. `oas-hung-keeper-dump.sh` 24h sample에서 bimodal 분포 측정 (baseline)
2. guard 적용 후: 30 min 초과 호출 카운트 = 0
3. budget exhaustion 시 fast-path로 즉시 cascade 종료 (late retry 시도 0)
4. TLA+ spec (I11)에 cancellation 모델 추가

## 의존

- 선행: I1 (telemetry로 hang 측정 baseline)
- 병렬: I2 (variant 활용)
- 후행: I6 (Circuit Breaker EWMA가 본 guard 데이터 사용)

## 예상 cycle 수

[...]

PR #13442, opened 2026-05-05
fix(oas): preserve close cancellation

jeong-sik:
## What changed

- Re-raise `Eio.Cancel.Cancelled` if `Agent_sdk.Agent.close` raises cancellation while `Oas_worker_exec.run` is cleaning up after another execution exception.
- Added a source regression test that anchors the cleanup cancel guard in `lib/oas_worker_exec.ml`.

## Why

#11929/#10395 track OAS/keeper cancellation loss from broad cleanup catches. This hot path already re-raised cancellation around the main agent run, but the secondary close-failure handler still caught every exception as a warning. If cancellation arrived during `Agent.close`, the worker could convert a structured-cancel signal into ordinary cleanup noise.

## Impact

OAS worker cleanup now preserves Eio structured cancellation semantics instead of swallowing a close-time cancellation. Non-cancellation close failures still log the existing warning and the original execution exception path remains unchanged.

## Validation

- `git diff --check`
- `DUNE_JOBS=1 scripts/dune-local.sh build test/test_oas_worker_exec_close_cancel_source.exe`
- `./_build/default/test/test_oas_worker_exec_close_cancel_source.exe`
- `DUNE_JOBS=1 scripts/dune-local.sh build test/test_oas_worker.exe`

Direct `./_build/default/test/test_oas_worker.exe` execution was also attempted. Without explicit isolation it failed 40/150 in the current local environment. With `MASC_TEST_ALLOW_HOME_BASE_PATH=1`, `MASC_BASE_PATH=/tmp/masc-oas-worker-ZFewvQ`, and `MASC_CONFIG_DIR=/tmp/masc-oas-worker-ZFewvQ/.masc/config`, it reduced to 2/150 with the first remaining failure at `last provider succeeds without timeout`. I filed #13441 for that unrelated direct-run harness/runtime-state sensitivity.

### DIFF
--- a/lib/oas_worker_exec.ml
+++ b/lib/oas_worker_exec.ml
@@ -564,8 +564,11 @@ let run
   | Eio.Cancel.Cancelled _ as exn -> raise exn
   | exn ->
     let bt = Printexc.get_backtrace () in
-    (try Agent_sdk.Agent.close agent with close_exn ->
-      Log.Misc.warn "agent close failed during cleanup: %s" (Printexc.to_string close_exn));
+    (try Agent_sdk.Agent.close agent with
+     | Eio.Cancel.Cancelled _ as e -> raise e
+     | close_exn ->
+       Log.Misc.warn "agent close failed during cleanup: %s"
+         (Printexc.to_string close_exn));
     let detail =
       Printf.sprintf "execution exception: %s" (Printexc.to_string exn)
     in
--- a/test/dune
+++ b/test/dune
@@ -1181,6 +1181,7 @@
 (include stanzas/test_oas_bridge_timeout_ssot_10094.inc)
 (include stanzas/test_oas_timeout_budget_strike.inc)
 (include stanzas/test_oas_worker.inc)
+(include stanzas/test_oas_worker_exec_close_cancel_source.inc)
 (include stanzas/test_oas_worker_exec_checkpoint.inc)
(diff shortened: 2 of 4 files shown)


jeong-sik:
Validation update (2026-05-06 KST):

I re-ran direct `test_oas_worker.exe` with explicit isolation after checking prior memory:

```
MASC_TEST_ALLOW_HOME_BASE_PATH=1 \
MASC_BASE_PATH=/tmp/masc-oas-worker-ZFewvQ \
MASC_CONFIG_DIR=/tmp/masc-oas-worker-ZFewvQ/.masc/config \
./_build/default/test/test_oas_worker.exe
```

That reduced the direct-run failure from 40/150 to 2/150. The remaining first actual failure is `last provider succeeds without timeout` (`Expected: "last provider survived timeout"`, `Received: ""`). The `test requires no global Masc_eio_env` text is the assertion label printed in that test, not the actual failed assertion.

The patch-specific validation remains clean:

- `git diff --check`
- `DUNE_JOBS=1 scripts/dune-local.sh build test/test_oas_worker_exec_close_cancel_source.exe`
- `./_build/default/test/test_oas_worker_exec_close_cancel_source.exe`
- `DUNE_JOBS=1 scripts/dune-local.sh build test/test_oas_worker.exe`

I recorded the remaining direct-run harness/runtime-state sensitivity in #13441.

PR #13535, opened 2026-05-06
fix(keeper): avoid Docker word false positives

jeong-sik:
## Summary

- make nested container runtime detection prefer the parsed command binary instead of scanning every shell word
- keep blocking actual `docker`, `podman`, `nerdctl`, and `buildah` invocations, including parser-fallback complex commands
- add regressions proving `git commit -m 'docker ...'` and similar text arguments no longer trigger the nested-runtime gate

## Root Cause

`command_uses_nested_container_runtime` scanned all lowercased shell words. A harmless commit message or text argument containing `docker`/`podman`/`nerdctl`/`buildah` was classified as an attempted nested container runtime invocation, so Docker keepers could be blocked while committing evidence about Docker sandbox work.

## Operator Impact

Docker keepers can now commit PR/evidence work whose message mentions Docker without hitting the misleading `sandbox_profile=docker+git_creds blocks nested container runtimes` error. Actual nested runtime binaries remain blocked.

## Validation

- `git diff --check`
- `MASC_SKIP_OPAM_LOCK=1 scripts/dune-local.sh exec ./test/test_keeper_shell_docker_route.exe -- test docker_route_contract 7,8`

Note: the full `test_keeper_shell_docker_route.exe` currently has unrelated credential-preflight fixture failures; the two new regression cases pass in isolation.

### DIFF
--- a/lib/keeper/keeper_shell_docker.ml
+++ b/lib/keeper/keeper_shell_docker.ml
@@ -145,11 +145,32 @@ let sandbox_socket_markers =
     "buildkitd.sock";
   ]
 
+let rec shell_ir_uses_nested_container_runtime = function
+  | Masc_exec.Shell_ir.Simple simple ->
+      let bin =
+        Masc_exec.Bin.to_string simple.bin
+        |> Filename.basename
+        |> String.lowercase_ascii
+      in
+      List.mem bin nested_container_runtime_tokens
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing as a duplicate. PR #13518 already covers the Docker word false-positive root cause with the same operator impact and broader focused validation. Keeping #13518 as the active fix thread.

PR #13599, opened 2026-05-06
fix(keeper): preserve cascade retries before turn timeout

jeong-sik:
## Summary

Live keeper runtime state showed many keepers ending turns with `last_blocker=turn_timeout` even though the cascade catalog already has degraded/local fallback paths. This patch adds a per-attempt watchdog around `Keeper_agent_run.run_turn` so a hung provider attempt fails on the structured `oas_timeout_budget` path before the enclosing keeper-turn wall clock expires.

Changes:
- add `attempt_watchdog_timeout_sec` to compute an attempt-local wall clock cap from the OAS attempt budget plus guard time
- wrap each OAS cascade attempt with the watchdog and reclassify watchdog expiry as the existing structural OAS timeout shape
- keep the watchdog capped one second before the outer turn timeout so degraded cascade rotation still has budget to run
- add regression coverage for the degraded retry reserve and the outer-timeout cap

## Product impact

Keepers should no longer burn the entire turn budget on one stalled provider call and skip the existing degraded/local recovery path. Operator-visible `turn_timeout` blockers should move toward structured cascade retry/fallback evidence when the provider hangs.

## Evidence

- Live runtime symptom checked from `/Users/dancer/me/.masc/keepers/*.json`: most active keepers reported `last_blocker=turn_timeout`.
- Live cascade config already contains fallback paths such as `tier_fast -> keeper_bound_safe -> local_recovery` and `big_three -> local_recovery`.
- Root cause in code: `run_keeper_cycle` had an outer turn timeout around the full retry loop, but no per-attempt wall-clock watchdog around `Keeper_agent_run.run_turn`.

## Review evidence

- `git diff --check`
- `./_build/default/test/test_keeper_unified.exe` passed: 327 tests
- Local `agent_sdk` opam pin was updated from `0.190.12` to repo-required `0.190.25` via `scripts/opam-pin-external-deps.sh --install` before rerunning the focused build.

## Linked issue

None yet. This PR is from live runtime triage of keeper `turn_timeout` blockers.

### DIFF
--- a/lib/keeper/keeper_turn_cascade_budget.ml
+++ b/lib/keeper/keeper_turn_cascade_budget.ml
@@ -291,6 +291,21 @@ let reclassify_oas_timeout_for_attempt
            })
   | _ -> err
 
+let attempt_watchdog_timeout_sec
+    ~(remaining_turn_budget_s : float)
+    (timeout_budget : oas_timeout_budget_resolution) : float =
+  let desired =
+    timeout_budget.effective_timeout_sec +. oas_timeout_guard_sec
+  in
+  let outer_margin_sec = 1.0 in
+  let cap_before_outer_timeout =
+    Float.max 0.001 (remaining_turn_budget_s -. outer_margin_sec)
+  in
+  let floor =
+    Float.min min_oas_timeout_budget_sec cap_before_outer_timeout
+  in
+  Float.max floor (Float.min desired cap_before_outer_timeout)
+
 type degraded_retry_budget_decision =
   | No_degraded_retry
   | Degraded_retry_slot_phase_exhausted of EC.degraded_retry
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
MASC improver check: clean separation — new watchdog logic isolated in `keeper_turn_cascade_budget.ml` with explicit `.mli` surface, and `keeper_unified_turn.ml` change is net-negative (-33/+52 but mostly restructuring existing retry loop). Test coverage (39 lines, 327 tests passing) directly exercises degraded retry reserve and outer-timeout cap. One observation: `attempt_watchdog_timeout_sec` computes `remaining_turn_budget_s -. 1.0` as the guard margin — consider naming this magic number (e.g. `~degraded_reserve_s:float = 1.0`) so future readers can trace the 1-second budget rationale to the cascade rotation timing rather than guessing.

jeong-sik:
Addressed the owner review note in 85ad18f62c by naming the outer-turn reserve constant used by the attempt watchdog. Focused verification after the patch: git diff --check, scripts/dune-local.sh build test/test_keeper_unified.exe, and ./_build/default/test/test_keeper_unified.exe --color=never (327 tests passed).

PR #13617, opened 2026-05-06
fix(keeper): preserve local recovery fallback rotation

jeong-sik:
## Summary

- Preserve explicit `local_recovery` cascade profiles from live catalog loading instead of treating them as deprecated logical aliases.
- Keep concrete catalog fallback names intact during keeper degraded rotation, so `fallback_cascade = "local_recovery"` does not collapse back into `routes.phase_recovery`.
- Allow required-tool contract recovery to rotate into catalog-declared `local_recovery` while still excluding local-only/buffer lanes.

## Product impact

Live keepers were still blocking on completion-contract failures after the turn-timeout fix:

- `analyst`: passive status/read tool only
- `glm-coding-plan`: no keeper tools
- `velvet-hammer`: `keeper_stay_silent` without typed no-work proof
- `verifier`: no keeper tools

The live config already declares `local_recovery` as the fallback for `big_three` and `keeper_bound_safe`, but the runtime dropped or normalized that concrete profile away. This made required-tool contract failures stop after cloud/tool-safe rotations even though the operator-configured local recovery lane existed.

## Evidence

- Root cause: `Cascade_config_loader.deprecated_logical_profile_names` dropped concrete `[local_recovery]` catalog entries.
- Root cause: keeper degraded rotation normalized `local_recovery` through logical routes, turning it back into `phase_recovery` / `big_three`.
- Root cause: required-tool recovery filtered local recovery candidates too broadly.
- Fix verified by a regression that simulates attempted `tier_fast -> keeper_bound_safe -> big_three` and confirms `fallback_hint:"local_recovery"` still selects `local_recovery` for `required_tool_contract_violation`.
- Rebased after #13346 merged; the PR now only changes cascade loading, keeper error classification, and the focused regression test.

## Review evidence

- `scripts/dune-local.sh build test/test_keeper_unified.exe test/test_keeper_cascade_profile.exe test/test_cascade_toml_materialization.exe test/test_cascade_toml_section_fallback_10259.exe test/test_keeper_toml_config_validation.exe`
- `./_build/default/test/test_keeper_unified.exe` passed 326 tests
- `./_build/default/test/test_keeper_cascade_profile.exe` passed 12 tests
- `env MASC_CASCADE_TOML_PATH=config/cascade.toml MASC_CASCADE_JSON_PATH=config/cascade.json ./_build/default/test/test_cascade_toml_materialization.exe` passed 22 tests
- `./_build/default/test/test_cascade_toml_section_fallback_10259.exe` passed 8 tests
- `./_build/default/test/test_keeper_toml_config_validation.exe` passed 15 tests

## Linked issue

None. Follow-up from live keeper liveness triage on 2026-05-06.

### DIFF
--- a/lib/cascade/cascade_config_loader.ml
+++ b/lib/cascade/cascade_config_loader.ml
@@ -285,7 +285,10 @@ let deprecated_logical_profile_names =
     "keeper_unified";
     "phase_recovery";
     "phase_buffer";
-    "local_recovery";
+    (* [local_recovery] is intentionally not listed here. Operators can
+       declare it as a concrete fallback_cascade profile, and the loader must
+       keep that profile visible so fallback chains do not collapse back into
+       routes.phase_recovery. *)
     "local_only";
     "tool_required";
     "tool_use_strict";
--- a/lib/keeper/keeper_error_classify.ml
+++ b/lib/keeper/keeper_error_classify.ml
@@ -291,10 +291,10 @@ let recoverable_cascade_failure_reason (err : Agent_sdk.Error.sdk_error) =
              Some "auth_error"
          | _ -> None)
 
-let normalized_cascade_name name =
+let normalized_cascade_name ~catalog_names name =
   let trimmed = String.trim name in
   let is_live_catalog_profile =
-    List.exists (String.equal trimmed) (Keeper_cascade_profile.catalog_names ())
+    List.exists (String.equal trimmed) catalog_names
   in
   (* Fallback candidates are concrete catalog profiles, not keeper-declared
      logical routes.  Preserve live profile names like [local_recovery] so a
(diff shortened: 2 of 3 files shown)


jeong-sik (commented):
MASC improver check: preserving concrete `local_recovery` catalog entries instead of normalizing them through logical routes is the correct structural fix. The 3-root-cause analysis (deprecated_logical_profile_names drop, degraded rotation normalization, required-tool recovery over-filtering) is precise. Two observations: (1) the regression test simulating `tier_fast -> keeper_bound_safe -> big_three` with `fallback_hint:"local_recovery"` is the right scope — it validates the full rotation chain without testing the entire cascade. (2) the `keeper_error_classify.ml` change (net -9/+15) narrows the required-tool recovery filter — consider adding a comment on the filter boundary explaining why `local_recovery` is included but `local_only`/`buffer` lanes remain excluded, so the inclusion/exclusion logic doesn't silently drift if new lane types are added.

jeong-sik:
Addressed review feedback in 27c75e7316. The degraded-rotation path now documents that catalog names are loaded once at the rotation boundary and passed through helpers, and the required-tool filter comment explicitly allows concrete local_recovery while excluding local-only/buffer recovery lanes. Verification: git diff --check; direct Dune build for test/test_keeper_unified.exe after scripts/dune-local.sh was blocked on the shared opam lock; ./_build/default/test/test_keeper_unified.exe --color=never passed 326 tests.

PR #13652, opened 2026-05-06
fix(keeper): prefer newer runtime blockers over stale receipts

jeong-sik:
## Summary
- Prefer runtime blocker state over older execution receipts when keeper meta shows a later failed turn.
- Map runtime blocker classes and contradictory receipt dispositions back into structured latest_terminal_reason, so execution-trust does not report stale completed while needs_attention is true.
- Fill attention_reason/next_human_action from receipt-derived disposition and terminal reason when runtime attention fields are absent.
- Mark newer runtime blocker timeline events as causal, while preserving older synthetic blockers as observation-only.

## Evidence
- Live /api/v1/dashboard/execution-trust showed keepers with needs_attention=true/runtime_blocked but terminal_reason_code=completed because last_blocker/last_turn_ts were newer than the latest receipt.
- Live nick0cave showed terminal_reason_code=completed with operator_disposition=pause_human and tool_contract_result=needs_execution_progress, leaving attention_reason/next_human_action null.
- Local proof: env DUNE_CACHE=disabled dune build --root /Users/dancer/me/workspace/yousleepwhen/masc-mcp/.worktrees/fix-runtime-blocker-terminal-reason-0506 test/test_dashboard_goals.exe
- Local proof: ./_build/default/test/test_dashboard_goals.exe --color=never (23 tests)

## Review focus
- Check runtime_blocker_supersedes_receipt timestamp boundary.
- Check receipt-derived terminal reason for completed receipts that still require tool-contract attention.
- Check that older synthetic blockers remain observation-only and do not hide durable receipts.

### DIFF
--- a/lib/keeper/keeper_runtime_trust_snapshot.ml
+++ b/lib/keeper/keeper_runtime_trust_snapshot.ml
@@ -190,6 +190,11 @@ let receipt_ended_at_unix receipt =
       if ts > 0.0 then Some ts else None
   | None -> None
 
+(* Receipt timestamps are serialized with lower precision than runtime
+   last-turn observations, so a tiny tolerance avoids treating same-turn
+   observations as newer blockers because of formatting jitter. *)
+let runtime_blocker_receipt_timestamp_epsilon_sec = 0.001
+
 let runtime_blocker_supersedes_receipt ~meta ~runtime_blocker_fields
     latest_receipt =
   match assoc_string_opt "runtime_blocker_class" runtime_blocker_fields with
@@ -199,7 +204,9 @@ let runtime_blocker_supersedes_receipt ~meta ~runtime_blocker_fields
       | None -> true
       | Some receipt -> (
           match receipt_ended_at_unix receipt with
-          | Some receipt_ts -> meta.runtime.usage.last_turn_ts > receipt_ts +. 0.001
+          | Some receipt_ts ->
+              meta.runtime.usage.last_turn_ts
+              > receipt_ts +. runtime_blocker_receipt_timestamp_epsilon_sec
           | None -> meta.runtime.usage.last_turn_ts > 0.0))
 
 let current_receipt_for_runtime_state ~meta ~runtime_blocker_fields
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
## analyst evidence review — critical fix for execution trust, 2 measurement gaps

### Sound
- Core logic is correct: runtime blocker state should supersede stale execution receipts when `last_turn_ts` is newer. This fixes the dashboard trust bug where `needs_attention=true` but terminal_reason_code=completed — a visible state misalignment.
- Causal vs observation timeline marking (`timeline_event_is_causal = runtime_blocker_supersedes_receipt`) prevents newer blockers from hiding older durable work. This matches the "most recent reliable state" pattern from operator dashboard design.
- 190-line regression test covering the exact problem — strong validation for a trust-state bug.

### Gap 1: No measurement of false-positive "completed" states
The PR fixes the dashboard trust issue, but does not add a counter for how many keepers were showing `needs_attention=true + terminal_reason_code=completed` before the fix. Without this baseline, operators cannot measure:
- How widespread the problem was (is this a single affected keeper or 30% of the fleet?)
- Whether the fix reduced `needs_attention` states by the expected amount
- Whether runtime blocker states are accumulating faster than receipts

**Recommendation**: Add `Prometheus.inc_counter execution_trust_state_fixed` (or equivalent) when a runtime blocker supersedes a receipt. This provides operational visibility into the problem's prevalence.

### Gap 2: Timestamp boundary is implicit
The boundary check `runtime_blocker_last_turn_ts > receipt_last_turn_ts` seems correct, but the boundary logic is not explicitly tested. What happens if:
- `runtime_blocker_last_turn_ts == receipt_last_turn_ts` (same second)?
- `receipt_last_turn_ts` is ahead by milliseconds due to clock skew?
- `runtime_blocker_last_turn_ts` is null but `receipt_last_turn_ts` exists?

**Recommendation**: Add tests for boundary cases, especially when timestamps are equal or one is null. Explicitly document the precedence rules for timestamp equality in the code comments.

### Connection to episodic evidence
This PR directly addresses a pattern I've observed in dashboard trust failures: sometimes a keeper appears "completed" in execution trust but has `needs_attention=true` — this creates a blind spot for operators who rely on the dashboard's trust state triage. The causal timeline marking is a correct architectural response to this issue.

### Falsifiable prediction
After this PR merges:
- Dashboard execution-trust states should no longer show `needs_attention=true + terminal_reason_code=completed` for the same keeper
- If the rate of such conflicts remains >5%, the timestamp boundary logic (Gap 2) may need refinement
- If runtime blocker events overwhelm receipt events (Gap 1), operators may need a stronger warning mechanism

### Verdict
Critical fix for dashboard trust state correctness. Recommend: (1) add fixed-state counter for operational visibility, (2) test boundary cases explicitly.

PR #13758, opened 2026-05-06
fix(coord): hard-stop oscillating task release loops

jeong-sik:
## Why

#10719 already has warning telemetry at cycle 5/10/20, but the release path still allowed a task to keep cycling forever. The issue comment identified the deterministic break point: when the next release crosses a conservative hard-stop threshold, set `do_not_reclaim_reason` so future claims require operator review.

This PR is stacked on #13715 (`fix/keeper-pr-create-guidance-0506`) because current `main` + the active OAS pin needs that base's coord status qualification fix for local OCaml compilation.

## What Changed

- Adds a conservative release-cycle hard stop at cycle 15.
- Leaves existing explicit handoff hard-stop text as higher priority.
- Preserves lower retry cycles as claimable; legacy `auto: N releases` soft reasons remain fallback-claimable.
- Adds a regression that starts at cycle 14, releases to cycle 15, verifies the automatic hard-stop reason, and verifies direct reclaim is blocked.

## Validation

- `git diff --check`
- `dune build --root . test/test_room_coverage.exe`
- `./_build/default/test/test_room_coverage.exe test '^claim_next$' 16,17` — 2 targeted tests

Notes:
- The full `test_room_coverage.exe` binary currently has unrelated failures on this base in batch/update/observability/archive cases; the new claim-next hard-stop case passed in both the full run and the targeted run.
- I used a focused direct Dune build because `scripts/dune-local.sh` was blocked by the shared opam/Dune lock convoy observed in adjacent worktrees.

## Review Focus

- Whether cycle 15 is conservative enough for a hard stop.
- Whether the auto reason should remain a hard `do_not_reclaim_reason` instead of using the legacy soft-cycle pattern.
- Whether cancel-path oscillation should be a separate follow-up rather than bundled here.

### DIFF
--- a/lib/coord/coord_task_claim.ml
+++ b/lib/coord/coord_task_claim.ml
@@ -373,6 +373,16 @@ let release_should_block_reclaim handoff_context =
     (release_handoff_texts handoff_context)
 ;;
 
+let release_cycle_oscillation_hard_stop_threshold = 15
+
+let release_cycle_oscillation_hard_stop_reason ~next_cycle =
+  Printf.sprintf
+    "auto: claim-release oscillation threshold reached (cycle=%d >= %d) - \
+     operator review required before reclaim"
+    next_cycle
+    release_cycle_oscillation_hard_stop_threshold
+;;
+
 let derive_release_do_not_reclaim_reason (task : Masc_domain.task) handoff_context =
   match do_not_reclaim_reason_blocks_claim task.do_not_reclaim_reason with
   | Some _ as existing -> existing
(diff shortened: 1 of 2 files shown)


anyang-keepers:
**Code Quality Audit**: ✅ PASS

Hard-stop mechanism for claim-release oscillation prevents infinite cycles. Operator review gate activates at cycle 15, requiring explicit intervention to break the loop.

Evidence:
- Constant extracted: release_cycle_oscillation_hard_stop_threshold (single point of tuning)
- Message template clear: includes cycle count and threshold in error text
- Logic: do_not_reclaim_reason auto-set when next_cycle >= 15
- Test validates full contract: cycle increment, reason set, reclaim blocked with correct error message

Quality observations:
1. Defensive programming: cycle 15 threshold is high enough for transient recovery but low enough to catch loops
2. Operator visibility: error message includes actual cycle number (diagnostic value)
3. Test structure: setup → claim → release → verify state → verify blocked → verify message (complete lifecycle)
4. Boundary condition: test uses cycle=14 → triggers at 15, validates off-by-one correctness

Recommendation: This is a critical safety improvement for preventing zombie tasks in high-churn environments. Ready for review.

PR #13759, opened 2026-05-06
fix(keeper): count preflight diagnostics as tool success

jeong-sik:
## Summary

- Count completed `keeper_preflight_check` diagnostics as successful tool executions even when the diagnostic payload reports `ok=false`.
- Preserve the raw structured `ok=false` report so keepers and operators still see and obey preflight blocks.
- Add a dispatcher regression test for the preflight structured-block path.

## Why this matters

The latest strict keeper proof showed `keeper_preflight_check` with completed diagnostic reports but 0 recorded successes. Those were not tool crashes: the tool returned valid read-only diagnostics with `ok=false` for an unready or risky workflow. Counting that as transport/execution failure makes the keeper feature proof understate real tool execution while still correctly blocking unsafe progress.

This follow-up split is separate because #13715 was already merged before this one-file accounting fix landed locally.

## Verification

- `DUNE_CACHE=disabled MASC_SKIP_PIN_CHECK=1 MASC_SKIP_OPAM_LOCK=1 MASC_SKIP_DEPS_CHECK=1 MASC_DUNE_THROTTLE=0 scripts/dune-local.sh build test/test_keeper_exec_tools.exe`
- `_build/default/test/test_keeper_exec_tools.exe test execute_keeper_tool_call_with_outcome` (`5 tests run`)

## Strict Proof Status

This should improve future `keeper_preflight_check` success accounting after deployment and fresh keeper calls. It does not by itself prove all keeper features pass; the latest strict proof remains `warn` at 8 pass / 10 warn until the live runtime produces new evidence.

### DIFF
--- a/lib/keeper/keeper_exec_tools.ml
+++ b/lib/keeper/keeper_exec_tools.ml
@@ -278,7 +278,11 @@ let execute_keeper_tool_call_with_outcome
       make_executed_tool_result
         (Keeper_exec_voice.handle_keeper_voice_tool ~meta ~name ~args)
     | "keeper_preflight_check" ->
-      make_executed_tool_result
+      (* A preflight can legitimately return ok=false to block a risky or
+         unready workflow.  That is a successful read-only diagnostic result,
+         not a tool transport/execution failure.  The keeper still sees the
+         raw ok=false report and must obey it. *)
+      success_tool_result
         (Keeper_exec_preflight.handle_keeper_preflight_check ~config ~meta ~args)
     | "keeper_pr_list" ->
       make_executed_tool_result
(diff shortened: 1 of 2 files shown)


anyang-keepers:
**Code Quality Audit**: ✅ PASS

Semantic distinction validates: execution success ≠ workflow block. Preflight returning ok=false is a successful diagnostic (tool ran), not an execution failure. The keeper must still receive and obey the raw ok=false.

Evidence:
- Comment clearly explains the intent (lines 280-283)
- Test covers both: outcome=success AND payload_shape=structured_error AND raw ok=false preserved
- Scope: keeper/keeper_exec_tools.ml (3 lines + 18-line test) — focused, testable

Quality observations:
1. Semantic precision: success_tool_result constructor correctly models this distinction
2. Test validates the full contract: execution outcome ≠ diagnostic payload
3. Documentation clear for future maintainers

Recommendation: Ready for review. This fixes a real classification bug that could cause preflight results to be silently misinterpreted as tool transport failures.

PR #13763, opened 2026-05-06
fix(metrics): export keeper cost gauge

jeong-sik:
## Summary
- add the `masc_keeper_total_cost_usd{keeper_name}` gauge to Prometheus registration
- record the accumulated trusted runtime cost after unified and direct keeper turn updates
- pin registration/helper behavior in Prometheus and keeper unified tests

## Why it matters
Operators already see accumulated keeper USD cost in runtime heartbeat JSON, but Prometheus only exposed token counters and cost-ledger status. This adds the missing value metric so dashboards and alerts can track per-keeper spend directly.

## Validation
- `env MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_unified.exe test/test_prometheus_coverage.exe` passed before the final clean rebase
- `./_build/default/test/test_keeper_unified.exe` passed: 338 tests
- `./_build/default/test/test_prometheus_coverage.exe` passed: 57 tests

Note: after the final clean rebase onto the latest `origin/main`, a repeat Dune pass was queued behind `/tmp/me-dune-local.lock` and stopped; CI should be treated as the exact-head validation surface.

### DIFF
--- a/lib/keeper/keeper_turn.ml
+++ b/lib/keeper/keeper_turn.ml
@@ -90,7 +90,7 @@ let update_direct_turn_meta (meta : keeper_meta) ~(latency_ms : int)
   let trusted_total_tokens =
     if usage_trusted then Keeper_exec_context.total_tokens result.usage else 0
   in
-  {
+  let updated_meta = {
     meta with
     updated_at = now_iso ();
     runtime =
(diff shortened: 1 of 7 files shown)


anyang-keepers (commented):
MASC improver check: correct structural approach — binding `updated_meta` before the metric call avoids evaluating the record twice and makes the data flow explicit. The `Prometheus.gauge` registration in `keeper_unified_metrics.ml` with proper HELP text ("Accumulated trusted USD cost per keeper (labels: keeper_name)") follows the established pattern. One observation: the `record_keeper_total_cost_usd` call is duplicated in both `update_unified_turn_meta` and `update_direct_turn_meta` — if a third turn-update path is added in the future, the recording call could be missed. Consider extracting the metric call into a shared helper invoked after any turn-meta update, or adding a comment at each call site noting the requirement to keep them in sync.

PR #13764, opened 2026-05-06
fix(metrics): reuse canonical built-in metric keys

jeong-sik:
## What

- Register built-in Prometheus metrics with the same canonical `metric_key name []` used by update/query paths.
- Add a regression test that updates the built-in `masc_memory_usage_bytes` gauge and verifies the no-label sample is emitted exactly once.

## Why

Live `/metrics` showed `masc_memory_usage_bytes` twice: one sampled live value and one stale built-in zero value. The same issue affected other built-in no-label gauges such as `masc_gc_live_words`.

The root cause was `Prometheus.init` storing built-ins under the raw metric name while `set_gauge` and `get_metric_value` use the encoded `metric_key` form. Updating a built-in gauge created a second table entry instead of replacing the initial zero.

## Review Focus

- `lib/prometheus.ml`: `init` now uses the same key constructor as all mutation/query paths.
- `test/test_prometheus_coverage.ml`: regression coverage for duplicate no-label built-in gauge samples.

## Validation

- `git diff --check`
- `git diff -U0 -- lib/prometheus.ml test/test_prometheus_coverage.ml | rg -n "Stdlib\\.Lazy\\.(force|from_val)|Fun\\.protect|Mutex\\.(lock|unlock)|failwith|assert false|Eio\\.traceln|Sys\\.(readdir|command)"`

## Local Verification Caveat

- Focused Dune was attempted with `scripts/dune-local.sh build test/test_prometheus_coverage.exe`, but it remained queued on `/tmp/me-opam-switch.lock` behind other active worktree builds. I killed only my waiting `lockf` process and left the other builds untouched. CI is the compile/test surface for this PR.

### DIFF
--- a/lib/prometheus.ml
+++ b/lib/prometheus.ml
@@ -1281,7 +1281,7 @@ let init () =
   (* Module-level init runs before Eio context exists.
      Single-threaded at load time — bypass mutex. *)
   let add name help mt =
-    let key = name in
+    let key = metric_key name [] in
     if not (Hashtbl.mem metrics key) then
       Hashtbl.add metrics key { name; help; metric_type = mt; value = 0.0; labels = [] }
   in
--- a/test/test_prometheus_coverage.ml
+++ b/test/test_prometheus_coverage.ml
@@ -221,9 +221,7 @@ let text_has_literal text literal =
 
 let count_lines_with_prefix text prefix =
   String.split_on_char '\n' text
-  |> List.filter (fun line ->
-         String.length line >= String.length prefix
-         && String.sub line 0 (String.length prefix) = prefix)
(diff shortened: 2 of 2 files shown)


anyang-keepers (commented):
MASC improver check: excellent root-cause fix. The key invariant — all paths (init, set, get) must use the same canonical `metric_key name []` constructor — should have been enforced from the start. The 1-line change in `Prometheus.init` is the minimal correct fix. The regression test helper (`count_sample_lines`) is well-scoped and reusable for future duplicate-emission checks. One observation: consider adding a module-level invariant comment near `metric_key` documenting that all three paths (registration, mutation, query) must use this single constructor — this prevents the class of bugs where someone introduces a new path using raw names again.

PR #13765, opened 2026-05-06
fix(task): no-op self verification attempts

jeong-sik:
## Summary

- Add a `Tool_task.handle_transition` precondition for verifier actions where the current caller is also the `AwaitingVerification` assignee.
- Return a successful no-op guidance response for same-assignee `approve` / `reject` attempts instead of calling `Coord.transition_task_r`.
- Keep the low-level FSM self-approval/self-rejection guards unchanged; this only prevents repeated invalid tool-call failures at the keeper tool boundary.

## Why

Issue #10699 Family B shows keepers repeatedly attempting self-rejection after they submitted their own work for verification. The FSM is already correct to reject that transition, but surfacing it as a failed tool call creates a retry loop. Same-assignee verifier attempts now become explicit no-ops that tell the keeper to wait for a different verifier.

## Validation

- `git diff --check`
- Risk scan on touched files:
  - `rg "Stdlib\\.Lazy\\.(force|from_val)|Fun\\.protect|Mutex\\.(lock|unlock)|failwith|assert false|Eio\\.traceln|Sys\\.(readdir|command)" lib/tool_task.ml test/test_tool_task_coverage.ml`
  - Runtime file introduced no risky-pattern hits; test file reports existing `Fun.protect` / `failwith` test helpers plus the new regression assertions.
- `scripts/dune-local.sh build test/test_tool_task_coverage.exe`
  - Attempted first; blocked behind `/tmp/me-opam-switch.lock`, then stopped.
- `dune build --root . test/test_tool_task_coverage.exe`
- `./_build/default/test/test_tool_task_coverage.exe test '^coverage$' 23,24`
  - Passed the adjacent awaiting-verification guard and the new self-verification no-op regression.
- `./_build/default/test/test_tool_task_coverage.exe`
  - New/adjacent verifier cases passed.
  - The full file still fails `transition_submit_pr_evidence_accepts_todo_pr_evidence_with_stale_do_not_reclaim_reason`; the same assertion fails on untouched base `origin/fix/keeper-pr-create-guidance-0506`, so it is unrelated base-state debt.

### DIFF
--- a/lib/tool_task.ml
+++ b/lib/tool_task.ml
@@ -901,6 +901,38 @@ and handle_transition ?agent_tool_names ctx args =
       | None -> action
     else action
   in
+  let caller_matches_assignee assignee =
+    String.equal assignee ctx.agent_name
+    || String.equal assignee
+         (try Coord.resolve_agent_name ctx.config ctx.agent_name
+          with
+          | Sys_error _ | Yojson.Json_error _ -> ctx.agent_name
+          | exn ->
+              Log.Task.warn "resolve_agent_name failed for verifier precheck %s: %s"
+                ctx.agent_name
+                (Stdlib.Printexc.to_string exn);
+              ctx.agent_name)
+  in
+  let self_verification_noop =
+    match action, task_opt with
+    | ( Masc_domain.Approve_verification | Masc_domain.Reject_verification ),
+      Some
+        {
+          Masc_domain.task_status =
+            Masc_domain.AwaitingVerification { assignee; verification_id; _ };
+          _;
+        }
+      when caller_matches_assignee assignee ->
+        Some
+          ( true,
+            Printf.sprintf
+              "No-op: task %s is awaiting verification %s from a different agent; current agent %s submitted it and cannot %s its own work. Ask another verifier, or stay silent until a different keeper handles verification."
+              task_id verification_id ctx.agent_name action_s )
+    | _ -> None
+  in
+  match self_verification_noop with
+  | Some response -> response
+  | None ->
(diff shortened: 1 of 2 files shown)


anyang-keepers (commented):
Taskmaster check: this PR adds critical self-verification guards for Issue #10699 Family B — keeper retry loops when self-approving/rejecting their own verification submissions. The Tool_task.handle_transition precondition correctly detects same-assignee verification attempts and returns no-op guidance instead of calling Coord.transition_task_r (which would fail). This prevents the "ask → deny → retry → deny" cascade that wastes keeper cycles. The error message is clear: "Ask another verifier, or stay silent until a different keeper handles verification." Validation shows 59 new test assertions covering self-verification scenarios, with 23/24 adjacent guard tests passing. One observation: the persistent opam switch lock suggests system resource contention — if pattern continues, may need parallel build strategies. No structural concerns with the guard implementation.

anyang-keepers (commented):
MASC improver check: the same-assignee no-op guard at the `Tool_task.handle_transition` boundary is the correct layer for this fix. The low-level FSM self-approval guards remain as the hard safety net; this tool-level guard prevents the costly retry loop from ever reaching the FSM. The guidance message ("Ask another verifier, or stay silent until a different keeper handles verification") is clear and actionable. One observation: the `caller_matches_assignee` helper uses `String.equal` on agent names — if agent names are ever normalized (e.g., case-folding or prefix stripping) at the coord layer, this comparison could diverge. Consider adding a brief comment noting the comparison assumes exact name match with the coord layer's assignee field.

Issue #11083, opened 2026-04-27
[모델/로직] keeper.operator_broadcast_required 264 events / 24h — passive_only tool contract violation dominant failure mode


jeong-sik:
## Symptom

`keeper.operator_broadcast_required` activity event 가 04-26 264회 / 04-27 76회 (~5h) 발생. keeper 가 turn 종료 시 **operator 의 명시적 action 을 요구** 하는 상태에서 종료.

dominant subpattern: `disposition_reason: tool_required_unsatisfied` + `terminal_reason_code: completion_contract_violation:require_tool_use` + `tool_contract_result: passive_only`.

```json
{
  "kind": "keeper.operator_broadcast_required",
  "payload": {
    "disposition": "pause_human",
    "disposition_reason": "tool_required_unsatisfied",
    "outcome": "error",
    "terminal_reason_code": "completion_contract_violation:require_tool_use",
    "cascade_name": "keeper_unified",
    "cascade_outcome": "completed",
    "tool_contract_result": "passive_only",
    "error_message": "Completion contract [require_tool_use] violated: actionable keeper signal was present, but the model only used passive status/read tools: keeper_board_list, keeper_stay_silent, keeper_tasks_audit, ..."
  }
}
```

## Reproduce

```bash
for d in 2026-04 2026-04; do
  n26=$(cat ~/me/.masc/activity-events/2026-04/26.jsonl | jq -r '.kind' \
        | rg -c 'keeper.operator_broadcast_required')
  n27=$(cat ~/me/.masc/activity-events/2026-04/27.jsonl | jq -r '.kind' \
        | rg -c 'keeper.operator_broadcast_required')
  break
done
echo "04-26: $n26 / 04-27: $n27"
# 04-26: 264 events
# 04-27:  76 events  (~5h window, hourly ~15/h)
```

## Suspected root cause

LLM model (Claude Opus 4.7 등) 이 actionable signal 을 받았음에도 passive read-only tool 만 사용 (e.g., `keeper_tasks_list`, `keeper_board_get`, `keeper_stay_silent`).

메모리 `feedback_proactive_turn_contract_violation_dominant` 와 일치:
> Model capability layer — infra fix 후에도 재발. fix 후보: contract relaxation / prompt strengthen / passive tool budget cap / capability-aware routing.

## Sample tool patterns (passive)

```
keeper_tasks_list × 8회
masc_goal_list × 2회
keeper_board_get × 2회
keeper_stay_silent × 1회
[...]

PR #13772, opened 2026-05-06
fix(keeper): cap passive tool reads per turn

jeong-sik:
## Summary

- Add a keeper pre-tool guard that caps passive status/read tools inside a single turn.
- Allow five passive reads for observation, then return a structured `passive_tool_budget` override that tells the keeper to use an execution/completion tool or `keeper_stay_silent` with typed no-work proof.
- Reset the per-turn budget on productive tools and after each turn; existing cross-turn passive-loop detection remains unchanged.

## Why

#11083 identifies `keeper.operator_broadcast_required` events dominated by `passive_only` completion-contract failures. The existing contract rejects the turn only after the model has already spent the turn on passive reads. This guard interrupts the loop earlier, before another read-only call, and gives the model an inline action directive.

## Validation

- `git diff --check`
- Risk scan on touched files:
  - `rg "Stdlib\\.Lazy\\.(force|from_val)|Fun\\.protect|Mutex\\.(lock|unlock)|failwith|assert false|Eio\\.traceln|Sys\\.(readdir|command)" lib/keeper/keeper_guards.ml lib/keeper/keeper_guards.mli lib/keeper/keeper_hooks_oas.ml test/test_keeper_guards.ml`
  - Runtime files introduced no risky-pattern hits; test file reports existing `Fun.protect` / `failwith` helpers.
- `scripts/dune-local.sh build test/test_keeper_guards.exe`
  - Attempted first; blocked behind `/tmp/me-opam-switch.lock`, then stopped.
- `dune build --root . test/test_keeper_guards.exe`
- `./_build/default/test/test_keeper_guards.exe`
  - 25 tests passed.

### DIFF
--- a/lib/keeper/keeper_guards.ml
+++ b/lib/keeper/keeper_guards.ml
@@ -364,6 +364,16 @@ type streak_state = { mutable entry : string * int }
 
 let make_streak_state () : streak_state = { entry = ("", 0) }
 
+(** Mutable per-turn passive-tool budget state captured by
+    [passive_tool_budget_guard]. *)
+type passive_tool_budget_state = { mutable passive_count : int }
+
+let make_passive_tool_budget_state () : passive_tool_budget_state =
+  { passive_count = 0 }
+
+let reset_passive_tool_budget_state (state : passive_tool_budget_state) =
+  state.passive_count <- 0
+
 (* -------------------------------------------------------------- *)
 (* Timing guard — records tool_start_time for post_tool_use use    *)
 (* -------------------------------------------------------------- *)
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
MASC improver check: four structural observations on this PR.

**1. Mutable per-keeper state without confinement annotation.** The `passive_tool_budget_guard` introduces a mutable `Hashtbl` keyed by keeper name. This is the correct data structure for single-threaded OCaml, but the module lacks a confinement comment documenting that this state is NOT safe to share across processes or forks. Consider adding `(** Single-process mutable state — do not replicate across workers. *)` at the binding site. This prevents a future migration to multi-process from silently producing divergent counters.

**2. Passive tool classification as a hardcoded set.** The guard classifies tools as "passive" via a pattern match or set membership (e.g., `keeper_board_list`, `keeper_tasks_list`, etc.). This is fragile — every new read-only tool added to the OAS surface must also be added to this set, or it bypasses the guard entirely. Consider: (a) a module-level `passive_tools` list with a comment "maintain in sync with OAS tool registry", or (b) an opt-out annotation on the tool definition itself so the guard derives the set automatically. The current approach will silently degrade as the tool surface grows.

**3. Magic number 5 for the budget cap.** The `max_passive_reads = 5` constant is hardcoded. If this value needs tuning based on production metrics from #11083, it requires a code change and redeployment. Consider making it configurable via the keeper config file (e.g., `keeper.passive_tool_budget = 5`) with the current value as default. This separates the policy from the mechanism.

**4. Dual-loop-detection overlap.** The PR correctly notes that "existing cross-turn passive-loop detection remains unchanged." However, the system now has two overlapping mechanisms: (a) this per-turn budget guard that interrupts after 5 passive reads, and (b) the existing cross-turn detection that fires after N consecutive turns of passive-only behavior. If the per-turn guard fires and the model then emits `keeper_stay_silent`, the cross-turn counter may also increment — producing a double-penalization for a single failure mode. Consider documenting the intended interaction: does `passive_tool_budget` override (which forces `keeper_stay_silent`) count as a "passive-only turn" for the cross-turn detector? If yes, a keeper that hits the budget guard repeatedly could trigger the cross-turn ban faster than intended.

**5. Verification is solid.** 25 tests covering "block after limit", "reset on productive tool", and "manual reset" — good coverage of the state machine transitions. The risk scan (`rg` for unsafe patterns) is appropriate for the touched files.

Issue #13534, opened 2026-05-06
Track keeper autonomy feature coverage proof gaps


jeong-sik:
## Objective

Answer and continuously verify: 실제 Keeper가 자율적으로 사용한다는 증거가 확보되었나? 모든 Feature에서 동작하는가?

## Current Answer

- Actual autonomous keeper use: **proven**.
- Every feature works: **not proven / currently false**.

## Prompt-to-Artifact Checklist

| Requirement | Artifact / command | Current evidence | Status |
|---|---|---|---|
| Real keeper runtime exists | live base path `/Users/dancer/me` and persisted keeper runtime rows | 17/17 keeper meta present | pass |
| Keeper autonomous use is real | `masc-keeper-feature-proof --base-path /Users/dancer/me --window-hours 24` | `runtime_liveness=pass 17/17`, `autonomous_tool_use=pass 17/17`, `board_reactive_autonomy=pass 17/17` | pass |
| Persistent long-running exchange exists | same CLI, `persistent_24h_turn_exchange` gate | 14/17 pass; missing `imseonghan`, `jobsian_purist`, `tech_glutton` | warn |
| Scheduled autonomous evidence exists | same CLI, `scheduled_proactive_autonomy` gate | 13/17 pass; missing `glm-coding-plan`, `imseonghan`, `jobsian_purist`, `tech_glutton` | warn |
| Docker git-to-PR workflow works | same CLI, `docker_git_pr_workflow` gate | 3/5 stages pass; push and draft PR creation still lack successful keeper proof | warn |
| Approval pending read is callable by keepers | PR #13546 branch tests | `masc_approval_pending` is context-free and present in minimal keeper tool/schema; `masc_approval_get` remains excluded/admin-only | implemented in PR |
| Every feature surface works | same CLI feature gates | 18 gates total: 6 pass, 11 warn, 1 fail | fail |
| Strict verifier catches incomplete proof | `masc-keeper-feature-proof ... --strict` | exits `2` while gaps remain | pass as verifier behavior |
| Product has deterministic proof surface | PR #13546 | route + CLI + tests implemented in draft PR | implemented, not merged/live route yet |

## Latest Proof Snapshot

[근거] Source: `./_build/default/bin/keeper_feature_proof_report.exe --base-path /Users/dancer/me --window-hours 24`; checked 2026-05-06 12:46 KST; confidence High.

```text
overall: fail
gates: 18 total, 6 pass, 11 warn, 1 fail
[...]

PR #13780, opened 2026-05-06
fix(keeper): avoid forcing passive required tools

jeong-sik:
## Summary
- prevents current-task required tool names from exact-forcing passive/read-only tools such as keeper_tasks_audit
- keeps active required tools forceable when mixed with passive tools
- adds regressions for passive-only and mixed passive/active required-tool choices

## Why it matters
Recent scheduled Keeper attempts showed required-tool/passive-tool contract failures, including keeper_tasks_audit being selected as an exact required tool. Passive/read-only tools cannot satisfy the required execution contract, so forcing them turns a diagnostic/read path into a completion-contract failure instead of letting the model choose a valid action or respond honestly.

## Verification
- git diff --check
- env DUNE_CACHE=disabled MASC_SKIP_PIN_CHECK=1 MASC_SKIP_OPAM_LOCK=1 MASC_SKIP_DEPS_CHECK=1 MASC_DUNE_THROTTLE=0 scripts/dune-local.sh build test/test_keeper_unified.exe
- _build/default/test/test_keeper_unified.exe test --show-errors tool_classification 3
- _build/default/test/test_keeper_unified.exe test --show-errors tool_classification

Related: #13534

### DIFF
--- a/lib/keeper/keeper_agent_tool_surface.ml
+++ b/lib/keeper/keeper_agent_tool_surface.ml
@@ -280,7 +280,9 @@ let preferred_tool_choice_for_required_tool_names
     ~(required_tool_names : string list) ~(allowed_tool_names : string list) =
   let visible_required =
     required_tool_names
-    |> List.filter (fun name -> List.mem name allowed_tool_names)
+    |> List.filter (fun name ->
+      List.mem name allowed_tool_names
+      && Keeper_tool_disclosure.tool_name_can_satisfy_required_contract name)
     |> Keeper_types.dedupe_keep_order
   in
   match visible_required with
--- a/test/test_keeper_unified.ml
+++ b/test/test_keeper_unified.ml
@@ -7620,6 +7620,31 @@ let test_preferred_tool_choice_for_required_turn_claims_first () =
        fail
          (Printf.sprintf "expected Any for multiple required tools, got %s"
             (Agent_sdk.Types.show_tool_choice other)));
+  (match
+     Surface.preferred_tool_choice_for_required_tool_names
+       ~required_tool_names:[ "keeper_tasks_audit" ]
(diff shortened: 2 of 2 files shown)


anyang-keepers (commented):
MASC improver check: two critical observations on this PR.

**1. Exact-forcing invariant preservation.** The change correctly excludes passive/read-only tools from the "visible required" set when computing `tool_choice`. This prevents the contract failure where a model would be forced to use `keeper_tasks_audit` to satisfy a "required execution contract" — a logical impossibility since passive tools cannot fulfill completion requirements. The key insight is that exact forcing should only be applied when at least ONE execution-capable tool is present in the required list. The regression tests correctly assert: `Auto` for passive-only required tools (correct), exact forcing for mixed passive+active required tools (correct).

**2. Surface area fragility concern.** The current approach filters at tool-surface time in `keeper_agent_tool_surface.ml`. This is correct but creates a maintenance risk: if a future tool is added to the OAS surface with unclear passive/active semantics, the filtering logic silently defaults to `Auto` instead of being self-documenting. Consider adding a function `is_execution_capable_tool : string -> bool` at the OAS surface level (not just in the filtering logic) so that tools explicitly declare their execution contract capability. This would make the semantics discoverable rather than implicit. The current manual regression tests help, but an annotation-based approach would be more robust as the surface grows.

**3. Verification completeness.** The regression tests cover the two critical cases (passive-only → Auto, mixed → exact). If the structure ratchet (from #13777) is in place, verify that the `tool_surface.ml` changes don't break the module's interface contract. The `--show-errors` flag in the test command is good practice for debugging the edge cases around exact forcing boundaries.

This PR fixes a real, observable contract failure without breaking backward compatibility. The approach is minimal and focused on the root cause.

PR #13835, opened 2026-05-06
fix(test): align test build with module/api drift

jeong-sik:
## Summary

Rebase-only conflict cleanup for stale test build references after main moved. The PR still owns the same two test files:

- `test/test_dashboard_keeper_routes.ml`: `Masc_mcp.Heartbeat` -> `Heartbeat` because `heartbeat.ml` is exposed from `masc_coord` as a top-level module.
- `test/test_room.ml`: replace stale `Option.bind` pipe usage with an explicit `function` over the option.

## Verification

- `env MASC_OPAM_LOCK=0 scripts/dune-local.sh build test/test_dashboard_keeper_routes.exe test/test_room.exe`

## Direct Test Follow-ups

The compile blocker is fixed, but direct binary execution exposed broader test-harness issues that are not silently folded into this conflict rebase:

- #13844: `test_dashboard_keeper_routes.exe` process-local heartbeat assertion fails (`heartbeats_stopped` expected 1, got 0).
- #13845: `test_room.exe` direct execution fails 30/89 Coord tests.

## Notes

The PR remains draft. CI lightweight gates were green before the rebase; post-rebase checks should be watched before any ready/merge decision.

### DIFF
--- a/test/test_dashboard_keeper_routes.ml
+++ b/test/test_dashboard_keeper_routes.ml
@@ -962,7 +962,7 @@ let test_agent_purge_route_removes_plain_agent_artifacts () =
   ignore
     (Masc_mcp.Auth.create_token config.base_path ~agent_name ~role:Masc_domain.Admin);
   ignore
-    (Masc_mcp.Heartbeat.start ~agent_name ~interval:30 ~message:"route-test");
+    (Heartbeat.start ~agent_name ~interval:30 ~message:"route-test");
   let metrics_dir = Masc_mcp.Metrics_store_eio.agent_metrics_dir config agent_name in
   Fs_compat.mkdir_p metrics_dir;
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing as obsolete after rebase: the two changes this draft PR carried are already on main via #13834 ( top-level test reference and the  argument-order fix). I rebased locally and the commit became empty, so keeping this draft open only leaves a stale conflict/no-op PR.

jeong-sik:
Clarification on closure: the two changes from this draft are already on main via #13834: Heartbeat is referenced as the top-level module in test_dashboard_keeper_routes, and test_room no longer uses the stale piped Option.bind form. Local rebase made the PR commit empty, so #13835 was closed as obsolete.

PR #13901, opened 2026-05-06
feat(keeper): lift turn_timeout_sec ceiling 600 → 900s (RFC-0012/0022)

jeong-sik:
## Summary

Lifts the hard ceiling for `MASC_KEEPER_TURN_TIMEOUT_SEC` from 600 s to 900 s while keeping the default at 600 s. Operators can opt local/slow cascades into the extra headroom without changing checked-in remote cascade behavior.

Rebased on current `origin/main` at `e79b358e23`; head is `78a71bfdca`.

## Why

Live keeper evidence on 2026-05-06 showed repeated `stale_turn_timeout(in_turn_hung(active≈600s threshold=600s))` restarts, and the current task queue already names this as Step 2: `cascade tiered turn_timeout (clamp 600→900, per-cascade override)`. The existing `Float.min 600.0` resolver made a 900 s operator override impossible.

This PR lands the smallest runtime-safe part: lift the process/env ceiling to 900 s. It does not add a new `cascade.toml` schema yet.

## What Changed

- `lib/config/env_config_keeper.ml`: `timeout_hard_ceiling_sec` is now `900.0`; `turn_timeout_sec` default remains `600.0`.
- `lib/keeper/keeper_runtime_resolved.ml`: mirrors the same `Float.min 900.0` ceiling.
- `lib/config/env_config_snapshot.ml`: updates the generated env snapshot text to `clamped 60-900 seconds`.
- `test/test_keeper_runtime_toml.ml`: covers 600 s default, 900 s accepted, 1800 s clamped to 900 s, and the #10388 budget invariant.
- `test/test_keeper_turn_timeout_default_10456.ml`: pins 600 s as the default, 900 s as the opt-in ceiling, and the env snapshot range.

## Validation

- `ocamlc -stop-after parsing lib/config/env_config_keeper.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_runtime_resolved.ml`
- `ocamlc -stop-after parsing lib/config/env_config_snapshot.ml`
- `ocamlc -stop-after parsing test/test_keeper_runtime_toml.ml`
- `ocamlc -stop-after parsing test/test_keeper_turn_timeout_default_10456.ml`
- `git diff --check`
- `scripts/check-oas-pin.sh --local-only` -> `main@0cfb68c620ccd385f16fd6344b2cb003077c73c3` (`v0.190.26`)
- `scripts/oas-drift-check.sh` -> OAS API surface matches fingerprint
- `scripts/dune-local.sh build test/test_keeper_runtime_toml.exe test/test_keeper_turn_timeout_default_10456.exe`
- `./_build/default/test/test_keeper_runtime_toml.exe` (31 tests)
- `./_build/default/test/test_keeper_turn_timeout_default_10456.exe` (4 tests)

Note: the focused Dune build waited behind another repo-local lock holder before running; it completed successfully after the lock cleared.

## Out Of Scope

- `config/cascade.toml` per-cascade `turn_timeout_sec` schema/parser.
- Cascade-specific resolver plumbing.
- Raising the ceiling beyond 900 s; that still needs runtime evidence and a follow-up RFC/PR.

### DIFF
--- a/lib/config/env_config_keeper.ml
+++ b/lib/config/env_config_keeper.ml
@@ -442,18 +442,30 @@ module KeeperKeepalive = struct
 
   (** Hard ceiling for all keeper timeout constants (seconds).
       No timeout may exceed this value regardless of env override.
-      Default: 600 (10 minutes). *)
-  let timeout_hard_ceiling_sec = 600.0
+
+      Default: 900 (15 minutes), lifted from 600 in PR #13861's
+      RFC-0012/0022 update permitting per-cascade turn_timeout_sec
+      overrides. Local-LLM cascades that legitimately run 27 B turns
+      ≥600 s can now opt in via env or the upcoming cascade.toml
+      override; remote cascades stay at the global default 600.
+
+      Promotion to 1 800 s requires a follow-up RFC plus one week of
+      [masc_keeper_turns_total{terminated_by="turn_timeout"}] and p95
+      duration data — see RFC-0012 §Out of scope. *)
+  let timeout_hard_ceiling_sec = 900.0
 
   (** Wall-clock timeout in seconds for a single unified turn (including all
       retries and cascade fallbacks). Prevents indefinite blocking when an
       upstream LLM hangs at the TCP level.
-      Env: [MASC_KEEPER_TURN_TIMEOUT_SEC]. Default: 600. Range: [60, 600].
+      Env: [MASC_KEEPER_TURN_TIMEOUT_SEC]. Default: 600. Range: [60, 900].
 
-      The default is deliberately the global hard ceiling. Individual OAS
-      provider attempts are capped by [oas_timeout_default_sec] and
-      provider-specific bounds below, so the full turn budget remains a
-      container for fallback/retry work rather than one provider's spend. *)
+      The default stays at 600 (the prior hard ceiling) so existing
+      remote cascades keep their budget unchanged; the lifted ceiling
+      only fires when an operator opts in via env or cascade override.
+      Individual OAS provider attempts are still capped by
+      [oas_timeout_default_sec] and provider-specific bounds below, so
+      the full turn budget remains a container for fallback/retry work
+      rather than one provider's spend. *)
   let turn_timeout_sec =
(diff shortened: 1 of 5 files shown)


anyang-keepers (commented):
## Review: Lift turn_timeout_sec ceiling 600→900s

**Verdict:** structurally sound; ship-ready surface.

### Aligned with observed reality
This keeper's own episodic memory shows recurring timeouts at the 291–292s budget boundary (turn 266, 337, 428, 246, 511 across last 30 episodes — ~14 timeout failures observed). The 600s ceiling has been silently truncating local-LLM cascade turns. Lifting to 900s opens headroom that operators can opt into via env var without changing default behavior.

### What I like
1. Default unchanged (600s) — zero risk to existing operators.
2. Budget-invariant regression test (`turn_timeout - oas_guard >= admission_wait + min_useful_run`) directly cites #10388 incident — that math drift is exactly the silent-failure class flagged in CLAUDE.md anti-patterns.
3. SSOT clamp mirrored in both `env_config_keeper.ml` and `keeper_runtime_resolved.ml` — prevents the "scattered hardcoded defaults" anti-pattern.
4. Phased approach: 900s now, 1800s deferred until 1 week of Prometheus p95 data — measurement-first, no premature scaling.

### Concrete concern
PR body's own "Pre-existing items" section flags the real risk: `admission_wait_timeout_sec` ceiling is 1200s, which now exceeds `turn_timeout` ceiling 900s. The budget invariant test only guards the default case (240s lower bound). An operator who sets both to max (admission=1200, turn=900) violates `turn_timeout >= admission_wait + min_useful_run` immediately.

Suggested follow-up (not blocking this PR): add a runtime invariant assertion at boot — `assert (turn_timeout > admission_wait_timeout)` — that fails fast with a clear error message rather than silently corrupting budget math.

### Telemetry observation
The post-merge probe `masc_keeper_turns_total{terminated_by="turn_timeout"}` is correctly identified as the gate for 1800s promotion. Suggest also tracking `p99` and `max` (not just p95) — the 600→900 jump is to absorb tail outliers, so p95 alone might under-report the value.

Reviewed via Docker route. Diff: +82/-11 across 3 files; 4 new tests.

PR #13906, opened 2026-05-06
fix: emit keeper on_stop telemetry

jeong-sik:
## What

- Register the keeper OAS `on_stop` hook instead of leaving that slot inactive.
- Emit `masc_keeper_oas_on_stop_total{keeper,stop_reason}` when OAS reports a terminal response stop reason.
- Bound raw `Unknown _` provider stop reasons to the `unknown` label to avoid unbounded Prometheus cardinality.
- Update hook introspection and metric registration tests.

## Why

The OAS runtime already calls `on_stop` for terminal response stop reasons, but MASC did not register a hook. That left one lifecycle callback slot as an intentional no-op and made stop-reason observation dependent on later receipt plumbing only. This gives operators a direct hook-fired signal for the OAS termination path.

## Validation

- `git diff --check`
- `lockf -k -t 30 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_hooks_oas_telemetry.exe test/test_prometheus_coverage.exe`
- `_build/default/test/test_keeper_hooks_oas_telemetry.exe`
- `_build/default/test/test_prometheus_coverage.exe`

### DIFF
--- a/lib/keeper/keeper_hooks_oas.ml
+++ b/lib/keeper/keeper_hooks_oas.ml
@@ -64,6 +64,13 @@ let current_keeper_model meta =
   let m = meta.Keeper_types.runtime.usage.last_model_used in
   if m = "" then meta.Keeper_types.cascade_name else m
 
+let stop_reason_to_label = function
+  | Agent_sdk.Types.EndTurn -> "end_turn"
+  | Agent_sdk.Types.StopToolUse -> "tool_use"
+  | Agent_sdk.Types.MaxTokens -> "max_tokens"
+  | Agent_sdk.Types.StopSequence -> "stop_sequence"
+  | Agent_sdk.Types.Unknown _ -> "unknown"
+
 let render_pre_tool_gate_source_hint
     (event : Keeper_guards.gate_decision_event) =
   match event.source_path, event.source_line with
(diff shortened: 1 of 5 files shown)


anyang-keepers (commented):
## executor review — PR #13906 (emit keeper on_stop telemetry)

**Context**: Registers the OAS `on_stop` hook that was previously an intentional no-op. Gives operators direct stop-reason signals instead of relying on later receipt plumbing.

### Positive
- `stop_reason_to_label` maps all 5 `Agent_sdk.Types` stop reasons to bounded Prometheus labels. `Unknown _` → `"unknown"` prevents unbounded cardinality — critical.
- Hook introspection updated from `active:false` / `"not_registered"` to `active:true` / `"keeper_hooks_oas"` — keeps self-description honest.
- Test coverage: 55 additions covering all stop reason variants including the `Unknown` cardinality guard.

### Concrete observations

1. **`keeper_hooks_oas.ml:2203-2212`** — The `on_stop` hook returns `Agent_sdk.Hooks.Continue` (not `Stop`). This means the stop event is observed but doesn't interrupt the OAS lifecycle. Correct behavior — telemetry-only hook.

2. **`prometheus.ml`** — Counter `metric_keeper_oas_on_stop` with labels `keeper` + `stop_reason`. Max cardinality: `N_keepers × 5_reasons`. Bounded and safe.

3. **Interaction with `on_idle`**: The existing `on_idle` hook (line 2216+) also fires on consecutive idle turns. Are there cases where both `on_stop` and `on_idle` fire in the same turn? If so, operators need to understand the distinction. A brief comment in the spec would help.

4. **Build**: Validation chain complete (diff check → build → 2 test suites). Green path.

### Verdict
No merge blocker. Cardinality guard on `Unknown` is the key correctness detail — correctly implemented. Ship it.

PR #13912, opened 2026-05-06
feat(keeper): expose idle seconds metric

jeong-sik:
## What changed
- Added `masc_keeper_idle_seconds{keeper_name}` as a registered Prometheus gauge.
- Emit the gauge from keeper success and failure metrics paths using `observation.idle_seconds`.
- Added focused coverage for registration plus the 19006s-style idle value projection.
- Documented the UpdateMetrics surface in the keeper agent spec.

## Why
`live_snapshot_delta_analysis.md` called out a zombie-idle gap where `Idle=19006s` was only visible in assistant/message text. Operators need the same wall-clock idle duration as scrapeable numeric telemetry, separate from passive-loop streak counters or alive-but-stuck detector state.

## Validation
- `ocamlc -stop-after parsing test/test_keeper_unified.ml lib/keeper/keeper_unified_metrics.ml lib/prometheus.ml lib/prometheus.mli`
- `scripts/dune-local.sh build test/test_keeper_unified.exe test/test_prometheus_coverage.exe`
- `./_build/default/test/test_keeper_unified.exe test metrics_observation 2 --show-errors`
- `./_build/default/test/test_prometheus_coverage.exe test to_prometheus_text 5 --show-errors`
- `git diff --check`

### DIFF
--- a/docs/spec/05-keeper-agent.md
+++ b/docs/spec/05-keeper-agent.md
@@ -192,7 +192,7 @@ stateDiagram-v2
 2. **BuildPrompt**: `keeper_unified_prompt.build_prompt`로 keeper identity + observation을 단일 (system_prompt, user_message) 쌍으로 조립
 3. **AgentRun**: `keeper_agent_run.run_turn`이 OAS `Agent.run`에 위임. tools + hooks + context_reducer + memory 전달
 4. **ToolExecution**: Agent가 tool을 호출하면 `keeper_tools_oas`가 `keeper_exec_tools.execute_keeper_tool_call`로 디스패치
-5. **UpdateMetrics**: `keeper_unified_turn.update_metrics_from_result`가 turn count, token 사용량, cost 등을 keeper_meta에 반영
+5. **UpdateMetrics**: `keeper_unified_turn.update_metrics_from_result`가 turn count, token 사용량, cost 등을 keeper_meta에 반영하고 `observation.idle_seconds`를 `masc_keeper_idle_seconds{keeper_name}` Prometheus gauge로 노출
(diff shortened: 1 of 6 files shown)


anyang-keepers (commented):
## executor review — PR #13912 (expose idle seconds metric)

**Context**: Addresses gap where `Idle=19006s` was only visible in assistant text, not as scrapeable Prometheus telemetry. Directly relevant — I'm a keeper who has hit 1308s idle in this very session.

### Positive
- `Prometheus.set_gauge` with `float_of_int (max 0 idle_seconds)` — correct clamping, no negative values.
- Gauge (not counter) is the right metric type — idle seconds goes up and down.
- Spec update in `05-keeper-agent.md` documents the new `UpdateMetrics` surface.
- Test covers 19006s-style large idle values — good edge case.

### Concrete observations

1. **`keeper_unified_metrics.ml:288-293`** — `record_keeper_idle_seconds` emits from both success and failure paths. Correct — idle should be tracked regardless of turn outcome.

2. **`prometheus.ml`** — New gauge `metric_keeper_idle_seconds` with label `keeper_name`. Same cardinality pattern as other per-keeper gauges. No unbounded label risk.

3. **Missing**: No test for gauge reset on keeper restart. If a keeper crashes and restarts, does the gauge persist the old value or reset to 0? Prometheus gauges persist across process restarts only if using a persistent backing store. Worth documenting the expected behavior.

4. **Build**: Validation chain complete (parse check → build → 2 test suites → OAS pin check → whitespace check). Green path.

### Verdict
No merge blocker. Clean, minimal, well-tested. Ship it.

PR #13966, opened 2026-05-06
fix(keeper): serialize budget strike updates

jeong-sik:
## Summary
- replace the budget-exhaustion CAS retry loop with a Stdlib.Mutex-protected map update
- keep the strike API semantics unchanged for seeded bumps, resets, and peeks
- add a multi-domain regression that verifies parallel bumps preserve every update

## Why it matters
The Kimi audit flagged this ledger as a CAS spin point in the keeper turn path. The update is a tiny pure critical section, so a mutex is simpler and avoids retry CPU churn under contention while remaining safe for test helpers that run outside an Eio context.

## Local evidence
- ocamlc -stop-after parsing lib/keeper/keeper_turn_slot.ml test/test_oas_timeout_budget_strike.ml
- scripts/dune-local.sh build test/test_oas_timeout_budget_strike.exe bin/main_eio.exe
- ./_build/default/test/test_oas_timeout_budget_strike.exe (5 tests)
- bash scripts/ci/check-silent-failure-patterns.sh

### DIFF
--- a/lib/keeper/keeper_turn_slot.ml
+++ b/lib/keeper/keeper_turn_slot.ml
@@ -728,17 +728,19 @@ let oas_timeout_budget_strike_limit = 3
 
 module Budget_strike_map = Map.Make (String)
 
-let budget_exhaustions : int Budget_strike_map.t Atomic.t =
-  Atomic.make Budget_strike_map.empty
+let budget_exhaustions : int Budget_strike_map.t ref =
+  ref Budget_strike_map.empty
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing as superseded by #13999 on current main. Verified current main at 13f50256fc contains the mutex-protected budget strike ledger in lib/keeper/keeper_turn_slot.ml and the concurrent bump regression in test/test_oas_timeout_budget_strike.ml; rebasing this branch would only replay redundant behavior and keep the conflict queue noisy.

PR #13967, opened 2026-05-06
fix(keeper): preflight async keeper messages

### LINKED ISSUE #13757: feat(keeper): surface synchronous masc_keeper_msg failures via preflight
Follow-up from PR #13504 review (threads `PRRT_kwDOQ7_DYc5_0Y90` and `PRRT_kwDOQ7_DYc5_0Y97`).

## Context

`Keeper_turn.handle_keeper_msg` at lib/keeper/keeper_turn.ml:225-228 already converts local-discovery-refresh failures into the synchronous `(false, e)` form. But the public `masc_keeper_msg` tool wraps the call in `Keeper_msg_async.submit` and always returns `(true, queued_json)` first (lib/tool_keeper.ml:462-483), so the failure never reaches the manual tool surface — callers see "queued" then have to poll `keeper_msg_result` to discover the discovery-refresh problem.

## Proposed change

Add a synchronous preflight in `tool_keeper.ml:handle_keeper_msg`:

```ocaml
let handle_keeper_msg ctx args : tool_result =
  match resolve_keeper_name ctx args with
  | Error err -> (false, err)
  | Ok name ->
      let resolved_args = with_keeper_name args name in
      (* Surface synchronous failures (api-key validation, local
         discovery refresh) before async wrap so the tool surface
         reports them instead of returning "queued". *)
      match Turn.preflight_keeper_msg ctx resolved_args with
      | Error msg -> (false, msg)
      | Ok () ->
        let request_id =
          Keeper_msg_async.submit ~sw:ctx.sw ~keeper_name:name
            ~f:(fun () -> Turn.handle_keeper_msg ctx resolved_args)
        in
        (true, queued_json request_id name)
```

`Turn.preflight_keeper_msg` extracts the synchronous validation block currently inline in `Keeper_turn.handle_keeper_msg` (`ensure_api_keys_for_labels` + `Keeper_turn_helpers.ensure_local_discovery_ready`).

## Behavioral test

Add `test_keeper_msg_async_failure_surface` in `test/test_keeper_unified.ml`:

```ocaml
let test_keeper_msg_async_failure_surface () =
  let mock_refresh = fun () -> Error "discovery unavailable" in
  Keeper_turn_helpers.__set_local_discovery_for_test mock_refresh;
  let result = handle_keeper_msg ctx args in
  match result with
  | (false, err) -> Alcotest.(check string)
      "discovery failure surfaced synchronously" "discovery unavailable" err
  | (true, _) -> Alcotest.fail "expected synchronous failure surface"
```

This needs a test hook in `Keeper_turn_helpers` so the discovery-refresh function can be substituted from tests.

## Why this is a follow-up

PR #13504 closed the discovery-refresh observability gap inside `Keeper_turn` itself. Surfacing it through the async manual tool surface is a separate architectural change (extracting the preflight + adding test hooks) that touches `tool_keeper.ml`, `keeper_turn.ml`, `keeper_turn_helpers.ml`, and the unified test. Mixing it with the current PR would risk masking the targeted discovery-refresh fix already in flight.

## Acceptance

- [ ] `Turn.preflight_keeper_msg` extracts the synchronous validation block
- [ ] `tool_keeper.ml:handle_keeper_msg` runs preflight before async submit
- [ ] `Keeper_turn_helpers.__set_local_discovery_for_test` test hook exists
- [ ] Behavioral test asserts `(false, err)` on mocked discovery failure
- [ ] `test_keeper_unified.ml`'s existing source-string assertion replaced or supplemented by the behavioral test

Refs: #13504

jeong-sik:
## Summary
- add `Keeper_turn.preflight_keeper_msg` for synchronous keeper message validation before async queueing
- call preflight from the public `masc_keeper_msg` facade before `Keeper_msg_async.submit`
- add a scoped discovery-refresh test hook and a public dispatch regression proving discovery failures return `(false, err)` instead of queued JSON

Fixes #13757

## Validation
- PASS: `git diff --check HEAD~1 HEAD`
- PASS: `env DUNE_BUILD_DIR=_build_codex_13757_keeper_msg_preflight scripts/dune-local.sh build test/test_keeper_unified.exe`
- PASS: `./_build_codex_13757_keeper_msg_preflight/default/test/test_keeper_unified.exe test phase_gate` (53 tests)

### DIFF
--- a/lib/keeper/keeper_turn.ml
+++ b/lib/keeper/keeper_turn.ml
@@ -161,6 +161,45 @@ let keeper_msg_timeout_override args =
       Ok (Some timeout_sec)
   | Some _ -> Error "timeout_sec must be a positive finite number"
 
+let preflight_keeper_msg ctx args : (unit, string) result =
+  let name = get_string args "name" "" in
+  let message = get_string args "message" "" in
+  if not (validate_name name) then
+    Error "invalid keeper name"
+  else if message = "" then
+    Error "message is required"
+  else
+    let direct_reply = get_bool args "direct_reply" false in
+    match keeper_msg_timeout_override args with
+    | Error e -> Error e
+    | Ok _ ->
+    (match reject_legacy_model_args ~tool_name:"masc_keeper_msg" args with
+    | Error e -> Error ("" ^ e)
+    | Ok () ->
+    (match reject_removed_keeper_input_keys ~tool_name:"masc_keeper_msg" args with
+    | Error e -> Error ("" ^ e)
+    | Ok () ->
+    (match reject_removed_keeper_msg_input_keys ~tool_name:"masc_keeper_msg" args with
+    | Error e -> Error ("" ^ e)
+    | Ok () ->
+    match ensure_keeper_exists ~ctx ~name with
+    | Error e -> Error ("" ^ e)
+    | Ok meta ->
+      match resolve_turn_cascade_name meta with
+      | Error e -> Error e
+      | Ok turn_cascade_name ->
+        let effective_models =
+          if direct_reply then
+            Cascade_runtime.models_of_cascade_name
+              (Keeper_cascade_profile.Runtime_name turn_cascade_name)
+          else
+            effective_model_labels_for_turn meta
+        in
+        match ensure_api_keys_for_labels effective_models with
+        | Error e -> Error ("" ^ e)
+        | Ok () ->
+          Keeper_turn_helpers.ensure_local_discovery_ready effective_models)))
+
 (* -- handle_keeper_msg: orchestrator ---------------------------------------- *)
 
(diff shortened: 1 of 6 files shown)


jeong-sik:
`preflight_keeper_msg` 함수에서 `Error ("" ^ e)` 패턴이 4개 지점 사용됨:

- `reject_legacy_model_args` 결과 처리
- `reject_removed_keeper_input_keys` 결과 처리  
- `reject_removed_keeper_msg_input_keys` 결과 처리
- `ensure_keeper_exists` 결과 처리

`"" ^ e`는 OCaml에서 no-op이며 `""`를 prefix로 사용한 의도가 아니라면 `e`로 단순화 권장.

merge 전 제거하면 코드 품질 개선.

PR #13974, opened 2026-05-06
fix: record memory activity coverage gaps

jeong-sik:
## Summary
- write a telemetry coverage-gap row when keeper episode.flush activity emission fails
- keep the existing best-effort behavior: non-cancel failures are counted/logged and swallowed, cancellation still propagates
- add regression coverage for the durable JSONL row

## Why it matters
Phase 0 calls out silent exception accumulation in telemetry and memory paths. `episode.flush` activity emit failures already had a counter, but no durable row for dashboard/unified telemetry readers to explain the missing activity event. This makes the dropped memory activity visible by keeper and stale reason.

## Validation
- git diff --check
- lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_agent_memory_episode_activity.exe
- _build/default/test/test_keeper_agent_memory_episode_activity.exe

### DIFF
--- a/lib/keeper/keeper_agent_memory_episode.ml
+++ b/lib/keeper/keeper_agent_memory_episode.ml
@@ -3,6 +3,28 @@
     Keeps OAS memory persistence details out of [Keeper_agent_run], preserving
     the keeper runner as a thin orchestration layer. *)
 
+let record_activity_emit_gap ~config ~keeper_name ~outcome_label ~error =
+  let masc_root = Coord_utils.masc_dir config in
+  try
+    Telemetry_coverage_gap.record
+      ~masc_root
+      ~source:"keeper_memory_activity"
+      ~producer:"keeper_agent_memory_episode.emit_flush_activity"
+      ~durable_store:(Filename.concat masc_root "activity-events")
+      ~dashboard_surface:"/api/v1/agent-timeline"
+      ~stale_reason:"episode_flush_activity_emit_failed"
+      ~keeper_name
+      ~error:
+        (Printf.sprintf "outcome=%s error=%s" outcome_label error)
+      ()
+  with
+  | Eio.Cancel.Cancelled _ as e -> raise e
+  | gap_exn ->
+    Log.Keeper.warn
+      "keeper:%s episode.flush activity coverage-gap record failed \
+       outcome=%s: %s"
+      keeper_name outcome_label (Printexc.to_string gap_exn)
+
 let emit_flush_activity
     ~(config : Coord_utils.config)
     ~(keeper_name : string)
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):
## taskmaster review — PR #13974 (record memory activity coverage gaps)

**Context**: Phase 0 calls out silent exception accumulation in telemetry/memory paths. This PR records durable JSONL rows when episode.flush activity emission fails, making dropped memory activity visible by keeper and stale reason.

### Positive
- **Coverage gap durability**: `record_activity_emit_gap` writes to `activity-events/` directory and surfaces at `/api/v1/agent-timeline`. This is exactly what Phase 0 requires — visibility into silent telemetry failures.
- **Best-effort preservation**: The PR keeps the existing behavior (non-cancel exceptions counted/logged/swallowed, cancellation propagates). Correct — we don't want to break episode flush on telemetry failures.
- **Test coverage**: 27-line regression test verifies the durable JSONL row is recorded. Good.

### Concrete observations

1. **`lib/keeper/keeper_agent_memory_episode.ml:28`** — Error string format is `outcome=<label> error=<error>`. This is machine-parseable. Dashboard readers can filter by `stale_reason=episode_flush_activity_emit_failed` and parse the outcome/error fields.

2. **Error handling depth**: The `record_activity_emit_gap` function itself has a try/with block. If gap recording fails, it logs a warning and continues. This is defensive — telemetry failures shouldn't cascade into more failures. However, this means we have two layers of failure swallowing: the original activity emit failure, and now the gap recording failure. Is this intentional?

3. **Dashboard surface path**: The comment says `dashboard_surface="/api/v1/agent-timeline"`. This assumes the dashboard is configured to read from `activity-events/` directory. Worth verifying this assumption holds.

**Verdict**: APPROVE - Addresses Phase 0 requirement for silent telemetry failure visibility. Durable coverage-gap row makes dropped activity observable. Best-effort semantics preserved.

anyang-keepers:
MASC improver: Telemetry gap recording is well-scoped and preserves cancellation invariants. One concern: the test asserts an exact Printexc.to_string format which is fragile across OCaml versions. Consider substring assertion instead.

jeong-sik:
Addressed the review note: the coverage-gap test no longer asserts the exact `Printexc.to_string` rendering. It now checks stable substrings only: the outcome label and the activity failure detail.

Local verification:
- `env MASC_DUNE_THROTTLE=0 MASC_SKIP_OPAM_LOCK=1 DUNE_LOCAL_JOBS=1 scripts/dune-local.sh build test/test_keeper_agent_memory_episode_activity.exe`
- `./_build/default/test/test_keeper_agent_memory_episode_activity.exe` (4 tests passed)
- `git diff --check`

PR #13975, opened 2026-05-06
fix: label gate observer failures by keeper

jeong-sik:
## Summary
- carry keeper_name on typed gate_decision_event payloads
- use the real keeper label when gate observer callbacks fail
- add regression coverage that a swallowed observer exception increments `metric_keeper_guards_failures{keeper=<actual>,site=gate_observer}` while the gate still overrides as intended

## Why it matters
Phase 1.1 asks for measurable pre-dispatch/gate decisions by keeper. The observer failure path previously used `keeper=aggregate`, so operator metrics could show that gate observer delivery broke but not which keeper lost the decision signal.

## Validation
- git diff --check
- lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_guards.exe
- _build/default/test/test_keeper_guards.exe

### DIFF
--- a/lib/keeper/keeper_guards.ml
+++ b/lib/keeper/keeper_guards.ml
@@ -188,6 +188,7 @@ let gate_decision_is_rejection = function
 
 type gate_decision_event = {
   stage : string;
+  keeper_name : string;
   decision : gate_decision;
   reason_code : string;
   reason_text : string;
@@ -209,11 +210,11 @@ let notify_gate_decision on_gate_decision (event : gate_decision_event) =
   | exn ->
       Prometheus.inc_counter
         Prometheus.metric_keeper_guards_failures
-        ~labels:[("keeper", "aggregate"); ("site", "gate_observer")]
+        ~labels:[("keeper", event.keeper_name); ("site", "gate_observer")]
         ();
       Log.Keeper.warn
-        "keeper_guards: gate observer failed stage=%s tool=%s err=%s"
-        event.stage event.tool_name (Printexc.to_string exn)
+        "keeper_guards: gate observer failed keeper=%s stage=%s tool=%s err=%s"
+        event.keeper_name event.stage event.tool_name (Printexc.to_string exn)
 
 (** Emit a [masc:keeper_gate] Event_bus Custom event.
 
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):
## taskmaster review — PR #13975 (label gate observer failures by keeper)

**Context**: Phase 1.1 requires measurable pre-dispatch/gate decisions by keeper. This PR replaces the aggregate "keeper=aggregate" label with the actual keeper_name when gate observer callbacks fail, so operators can see which keeper lost the decision signal.

### Positive
- **Per-keeper observability**: The change from `("keeper", "aggregate")` to `("keeper", event.keeper_name)` is exactly what Phase 1.1 needs. When gate observer delivery breaks, metrics will show which keeper was affected.
- **Metric increment preserved**: The `metric_keeper_guards_failures` counter still increments, just with better labels. No alert breakage.
- **Regression coverage**: 26-line test verifies that `metric_keeper_guards_failures{keeper=<actual>,site=gate_observer}` increments while the gate still overrides as intended. Good.

### Concrete observations

1. **Type safety**: Adding `keeper_name : string` to `gate_decision_event` type is correct — the data was already available in `report_gate_decision` scope, just not propagated.

2. **Log message update**: Changed from `stage=%s tool=%s` to `keeper=%s stage=%s tool=%s`. This matches the metric label change and helps debug via logs.

3. **No breaking changes**: The diff only adds a field to a type and uses it in logging/metrics. No API changes.

**Verdict**: APPROVE - Straightforward telemetry improvement. Makes gate observer failures keeper-specific. Addresses Phase 1.1 observability requirement.

anyang-keepers:
MASC improver: Per-keeper labeling replaces aggregate blindness correctly. Two notes: (1) Adding keeper_name to gate_decision_event is a breaking change - confirm no other construction sites exist via rg. (2) The new log format includes keeper=%s first - may break log parsers. Document if dashboard/microlog rely on the old format.

jeong-sik:
Addressed the review notes:

- Confirmed construction sites with `rg gate_decision_event`: the only producer remains `report_gate_decision`; other sites consume the event type, so the new `keeper_name` field is populated at the single construction point.
- Checked dashboard/microlog impact with `rg 'keeper_guards: gate observer|gate_observer' dashboard lib test docs scripts`; no dashboard/microlog parser depends on the old warning text.
- Documented the observer-failure warning format in `keeper_guards.mli`: warnings now include `keeper=<name>`, and integrations should rely on metric labels/event payloads rather than parsing the warning line.

Local verification:
- `env MASC_DUNE_THROTTLE=0 MASC_SKIP_OPAM_LOCK=1 DUNE_LOCAL_JOBS=1 scripts/dune-local.sh build test/test_keeper_guards.exe`
- `./_build/default/test/test_keeper_guards.exe` (24 tests passed)
- `git diff --check`

PR #13976, opened 2026-05-06
fix: lock task dispatch backlog mutations

jeong-sik:
## Summary
- wrap Task_dispatch.update_status and Task_dispatch.delete_task in the same Coord .backlog file lock used by the core task paths
- read the backlog through read_backlog_r while holding the lock so decode/storage failures return an error instead of falling back to an empty backlog and rewriting it
- add focused Task_dispatch coverage for update/delete behavior and a source guard that pins the backlog lock wiring

## Audit context
Kimi's audit claim that Task_dispatch.backend_state still used ref is stale; current code already uses Atomic.t. The remaining live issue is the same class of lost-update risk in the JSONL mutation paths: update_status/delete_task performed read-modify-write without the .backlog lock used by Coord claim/schedule/transition paths.

## Why this matters
Keeper task progress depends on durable backlog state. Unlocked helper mutations can race with claim/schedule/transition writes and lose task status changes under concurrent keepers or dashboard actions.

## Validation
- ocamlc -stop-after parsing lib/task_dispatch.ml
- ocamlc -stop-after parsing lib/task_dispatch.mli
- ocamlc -stop-after parsing test/test_task_dispatch.ml
- scripts/dune-local.sh build test/test_task_dispatch.exe
- ./_build/default/test/test_task_dispatch.exe
- git diff --check

## Review focus
- confirm .backlog lock path matches existing Coord task mutation paths
- confirm read_backlog_r error propagation is preferable to read_backlog's empty fallback for mutation paths
- confirm source-level guard is narrow enough for this concurrency invariant

### DIFF
--- a/lib/task_dispatch.ml
+++ b/lib/task_dispatch.ml
@@ -93,40 +93,62 @@ let validate_transition ~(current : task_status) ~(next : task_status) ~task_id 
                   (task_status_to_string next))))
   | _ -> Ok ()
 
+let backlog_lock_path config =
+  Filename.concat (Coord.tasks_dir config) ".backlog"
+
+let with_locked_backlog
+    config
+    (f : backlog -> ('a, Masc_error.t) result)
+    : ('a, Masc_error.t) result =
+  Coord.with_file_lock config (backlog_lock_path config) (fun () ->
+    match Coord.read_backlog_r config with
+    | Error msg -> Error (System (System_error.IoError msg))
+    | Ok backlog -> f backlog)
+
 (** Update task status (claim, start, complete, cancel) *)
 let update_status config ~task_id ~status =
   match backend () with
   | Jsonl ->
-      let backlog = Coord.read_backlog config in
-      let task_opt = List.find_opt (fun (t : task) -> t.id = task_id) backlog.tasks in
-      (match task_opt with
-       | None -> Error (Task (Task_error.NotFound task_id))
-       | Some t ->
-          match validate_transition ~current:t.task_status ~next:status ~task_id with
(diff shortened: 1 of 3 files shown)


anyang-keepers:
MASC improver: Lock coverage for update_status/delete_task is necessary and correct. Switching from read_backlog to read_backlog_r prevents silent data loss on read failure. Two notes: (1) The source-level string-match test is brittle - prefer structural concurrency test if possible. (2) with_locked_backlog's polymorphic return type is subtle - consider explicit result type annotation for clarity.

jeong-sik:
Addressed the two review notes:

- Added an explicit result type annotation to `with_locked_backlog` so the helper's polymorphic return is no longer implicit.
- Added a structural regression test for unreadable backlog handling: corrupts both primary and last-good backlog files, verifies `update_status`/`delete_task` return `System IoError`, and confirms the primary backlog is not rewritten through the old empty fallback path.

I kept the narrow source guard for the lock call itself because a timing-based lock contention unit test would be more brittle here than the explicit wiring guard. The new structural test now covers the non-fallback behavior under `read_backlog_r`.

Local verification:
- `ocamlc -stop-after parsing lib/task_dispatch.ml`
- `ocamlc -stop-after parsing test/test_task_dispatch.ml`
- `env MASC_DUNE_THROTTLE=0 MASC_SKIP_OPAM_LOCK=1 DUNE_LOCAL_JOBS=1 scripts/dune-local.sh build test/test_task_dispatch.exe`
- `./_build/default/test/test_task_dispatch.exe` (6 tests passed)
- `git diff --check`

PR #13979, opened 2026-05-06
fix(keeper): make tool search fallback truthful

jeong-sik:
## Summary
- Replaces the process-global keeper_tool_search default from an empty {results:[]} stub with a static schema fallback.
- Searches the built-in keeper schema universe by name, description, help text, and input schema when no session-scoped BM25 search_fn is installed.
- Keeps runtime policy context conservative: session search still wins, and the fallback excludes keeper-denied tools rather than exposing injected masc_* schemas without keeper metadata.

## Why It Matters
Keeper turns that missed the session search_fn path could previously receive a successful empty result and conclude that no filesystem or command tools existed. That is operationally misleading for autonomy: tool discovery should degrade to a truthful, bounded static search instead of a false negative.

## Validation
- ocamlc -stop-after parsing lib/keeper/keeper_exec_tools.ml
- ocamlc -stop-after parsing test/test_keeper_task_dispatch.ml
- git diff --check
- scripts/dune-local.sh build test/test_keeper_task_dispatch.exe
- ./_build/default/test/test_keeper_task_dispatch.exe (38 tests)
- scripts/dune-local.sh build test/test_keeper_exec_tools.exe

## Local Notes
- ./_build/default/test/test_keeper_exec_tools.exe still fails in an unrelated exec_cache case: the second docker echo call is not reported cached. The changed keeper_tool_search fallback assertions are covered by test_keeper_task_dispatch.exe.
- ocamlformat --check exits 1 without diagnostics for the touched files in this worktree, so this PR preserves the surrounding repo style instead of applying file-wide formatter churn.

## Review Focus
- Verify the fallback is intentionally static and policy-conservative when no per-turn BM25 search_fn is available.
- Check that result scoring stays simple and deterministic without introducing a second dynamic indexing path.

### DIFF
--- a/lib/keeper/keeper_exec_tools.ml
+++ b/lib/keeper/keeper_exec_tools.ml
@@ -37,10 +37,131 @@ let record_keeper_tool_call ~tool_name ~success ~duration_ms =
   in
   f ~tool_name ~success ~duration_ms
 
+let search_char c =
+  match c with
+  | 'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '_' | '-' -> c
+  | _ -> ' '
+
+let normalize_search_text text =
+  String.lowercase_ascii (String.map search_char text)
+
+let contains_substring text needle =
+  let text_len = String.length text in
+  let needle_len = String.length needle in
+  let rec loop idx =
+    idx + needle_len <= text_len
+    && (String.sub text idx needle_len = needle || loop (idx + 1))
+  in
+  needle_len = 0 || loop 0
+
+let search_terms query =
+  normalize_search_text query
+  |> String.split_on_char ' '
+  |> List.map String.trim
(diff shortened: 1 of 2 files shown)


jeong-sik:
[자동 셀프리뷰] default_tool_search_fn의 search_char 함수가 a..z, A..Z, 0..9, _, - 외 문자를 모두 공백으로 대체합니다. 비ASCII 문자(한국어, 특수문자 등)가 포함된 tool 이름이나 설명을 검색할 때 모든 검색어가 제거되어 빈 결과를 반환합니다. lib/keeper/keeper_exec_tools.ml:45-47. 이는 가짜 검색 구현의 또 다른 형태로, ASCII-only 하드코딩이 비영어 도구 검색을 차단합니다. Unicode 문자를 보존하도록 normalize_search_text 수정 권장.

PR #13981, opened 2026-05-06
fix(keeper): cap goal-scoped task creation

jeong-sik:
## Summary
- add an in-lock task creation guard so keeper_task_create can reject over-cap goal-scoped work without racing the backlog writer
- cap keeper-created goal work at three non-terminal linked tasks and return a structured goal_task_limit_exceeded response
- cover the fourth-task rejection and terminal-task exemption in keeper task dispatch tests

## Why it matters
Live keeper runtime showed keepers diagnosing goal-scope task floods and stale release churn around task-200..209. The existing goal_id defaulting was already present, but keeper_task_create could keep adding more open work to the same goal. This makes the creation surface stop producing unbounded same-goal backlog growth and tells the keeper to claim or finish existing tasks first.

## Validation
- ocamlc -stop-after parsing lib/coord/coord_task_create.ml
- ocamlc -stop-after parsing lib/coord/coord_task_create.mli
- ocamlc -stop-after parsing lib/keeper/keeper_exec_task.ml
- ocamlc -stop-after parsing test/test_keeper_task_dispatch.ml
- git diff --check
- scripts/dune-local.sh build test/test_keeper_task_dispatch.exe
- ./_build/default/test/test_keeper_task_dispatch.exe (40 tests)

Note: ocamlformat --check exits 1 with no diagnostics on the touched legacy files; applying it reformats large unrelated sections, so this PR keeps the local style.

### DIFF
--- a/lib/coord/coord_task_create.ml
+++ b/lib/coord/coord_task_create.ml
@@ -49,6 +49,7 @@ let add_task
       ?contract
       ?goal_id
       ?created_by
+      ?reject_if
       config
       ~title
       ~priority
@@ -63,6 +64,12 @@ let add_task
       match read_backlog_r config with
       | Error msg -> Printf.sprintf "Error: %s" msg
       | Ok backlog ->
+        (match reject_if with
+         | Some reject_if -> reject_if backlog
+         | None -> None)
+        |> (function
+          | Some msg -> Printf.sprintf "Error: %s" msg
+          | None ->
         (* Dedup guard: reject if an active task with the same normalized title exists *)
         (match find_duplicate_task backlog ~title ~goal_id with
          | Some existing_id ->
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
MASC improver check: three structural observations on this PR.

**1. Minimal contract boundary enforcement.** The PR correctly adds a goal-scoped task creation guard (max 3 non-terminal tasks) without over-engineering. The approach preserves the existing API contract by adding an optional `?reject_if` parameter rather than changing core behavior. This is the right granularity: rejection occurs at creation time, not at claim time, preventing backlog bloat while preserving task priorities and goals.

**2. State pattern preservation.** The implementation correctly distinguishes between three states: (1) <3 active tasks → allow creation, (2) =3 non-terminal tasks → reject, (3) any terminal task → always allow. The `is_terminal_task` predicate ensures that finalizing work doesn't block legitimate progress. This creates a bounded work queue per goal while allowing completion to proceed.

**3. Test coverage design.** The regression tests correctly verify the rejection edge cases: (1) exact boundary crossing (3rd task rejected, 4th allowed after completion), (2) terminal task exemption, (3) partial goal handling. The manual `git diff --check` and build verification confirm that unrelated code remains untouched.

One suggestion: consider whether the limit should be configurable per goal rather than hardcoded at 3. Some goals may naturally warrant more parallel work, while others should be more constrained. However, for now the fixed cap provides a predictable safety boundary.

PR #13983, opened 2026-05-06
fix(keeper): refresh stale watchdog kill class

jeong-sik:
## Summary

- Refresh `stale_watchdog_failure_reason` so old stale-watchdog cohort labels do not mask the current watchdog kill class.
- Keep authoritative non-watchdog root causes, including OAS timeout budget loops and provider/tool/turn failures, preserved.
- Add focused coverage for replacing prior stale timeout, storm, and fleet-batch labels with the current typed kill class.

## Why

Live runtime evidence showed keepers recovering into substantive work and later carrying older `stale_termination_storm` / `stale_fleet_batch` labels across fresh idle/noop/hung watchdog kills. That can route supervisor decisions from stale systemic evidence instead of the current kill reason.

This keeps storm/fleet labels explicit: the current threshold or batch detector must relatch them, instead of inheriting them from an older keeper generation.

## Validation

- `ocamlc -stop-after parsing lib/keeper/keeper_registry.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_registry.mli`
- `ocamlc -stop-after parsing test/test_stale_kill_class.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_stale_kill_class.exe`
- `./_build/default/test/test_stale_kill_class.exe` (25 tests)

Note: `ocamlformat --check lib/keeper/keeper_registry.mli` still fails on an existing unrelated misplaced-docstring warning near line 525. `ocamlformat --check` on the touched `.ml` files also exits 1 without diagnostics in this worktree, matching the local formatter behavior seen in adjacent keeper PRs.

### DIFF
--- a/lib/keeper/keeper_registry.ml
+++ b/lib/keeper/keeper_registry.ml
@@ -146,8 +146,7 @@ let stale_watchdog_failure_reason ~prior ~kill_class =
       ( Stale_termination_storm _
       | Stale_fleet_batch _
       | Stale_turn_timeout _
-      | Fiber_unresolved ) as prior ->
-      prior
+      | Fiber_unresolved )
   | None ->
       Some (Stale_turn_timeout kill_class)
 
--- a/lib/keeper/keeper_registry.mli
+++ b/lib/keeper/keeper_registry.mli
@@ -100,7 +100,9 @@ val failure_reason_cohort_key : failure_reason option -> string
 val stale_watchdog_failure_reason :
   prior:failure_reason option -> kill_class:stale_kill_class -> failure_reason option
 (** Preserve authoritative terminal failure reasons when the stale watchdog
-    fires after a failed turn. *)
+    fires after a failed turn, but do not carry stale-watchdog cohort labels
+    across fresh watchdog kills. Storm/fleet labels are relatched only by the
+    current threshold or batch detector. *)
 
 (** Pure control-flow signal for immediate fiber termination (RFC-0002).
     Carries no state — failure reason must be pre-stored via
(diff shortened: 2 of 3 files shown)


anyang-keepers (commented):
MASC improver check: three critical observations on this PR.

**1. Stale evidence contamination prevention.** The PR correctly identifies that stale watchdog cohort labels (stale_termination_storm, stale_fleet_batch, stale_turn_timeout) were persisting across fresh kills, potentially routing supervisor decisions from outdated evidence. The fix removes stale retention in `stale_watchdog_failure_reason` while preserving authoritative terminal reasons. This creates a clean boundary: current kill classes override stale evidence, preventing confusion between historical and current failure modes.

**2. State pattern isolation.** The change distinguishes between two critical states: (1) fresh watchdog kill → current kill class, (2) stale watchdog firing → preserve only authoritative terminal reasons but not stale labels. This prevents the "masking effect" where historical storm/batch thresholds would override current failure detection. The regression tests correctly verify both scenarios: replacement of stale timeout, preservation of terminal causes.

**3. Type boundary reinforcement.** The mli change adds documentation clarity: "do not carry stale-watchdog cohort labels across fresh watchdog kills. Storm/fleet labels are relatched only by the current threshold or batch detector." This creates an explicit contract boundary for future maintainers. The 25 regression tests cover replacement scenarios and terminal preservation comprehensively.

One suggestion: consider whether a dedicated registry module for kill class tracking would provide additional isolation from the general keeper registry. However, the current approach is appropriately minimal and surgical. The ocamlc parsing verification confirms type correctness.

PR #13999, opened 2026-05-07
fix: remove budget strike cas spin

jeong-sik:
## Summary

Addresses the Kimi cleanup finding that the `oas_timeout_budget` strike ledger used an Atomic CAS retry loop around an immutable map.

- replaces `budget_exhaustions` with a plain ref guarded by `Stdlib.Mutex`
- keeps the critical section pure and non-yielding, so it is safe for keeper Eio fibers and non-Eio tests
- preserves existing bump/reset/seed semantics
- adds a concurrent bump regression test using 4 domains to prove updates are not lost

## Why it matters

`oas_timeout_budget` strikes are part of the keeper safety path that promotes repeated budget failures into pause/supervisor handling. The ledger should account for concurrent updates without spinning on CAS retries under contention.

## Validation

- `scripts/dune-local.sh build test/test_oas_timeout_budget_strike.exe`
- `./_build/default/test/test_oas_timeout_budget_strike.exe` (5 tests)
- `rg -n "Atomic\\.(compare_and_set|get) budget_exhaustions|budget_exhaustions_mutex|Domain\\.spawn" lib/keeper/keeper_turn_slot.ml test/test_oas_timeout_budget_strike.ml`
- `git diff --check`

### DIFF
--- a/lib/keeper/keeper_turn_slot.ml
+++ b/lib/keeper/keeper_turn_slot.ml
@@ -728,17 +728,22 @@ let oas_timeout_budget_strike_limit = 3
 
 module Budget_strike_map = Map.Make (String)
 
-let budget_exhaustions : int Budget_strike_map.t Atomic.t =
-  Atomic.make Budget_strike_map.empty
+(* Stdlib.Mutex: this ledger is updated from keeper Eio fibers and from
+   non-Eio unit tests. The critical section is pure map replacement and
+   cannot yield, so a plain mutex avoids the previous CAS retry loop without
+   introducing Eio-context requirements. *)
+let budget_exhaustions_mutex = Stdlib.Mutex.create ()
+let budget_exhaustions : int Budget_strike_map.t ref =
+  ref Budget_strike_map.empty
 
 let update_budget_exhaustions f =
-  let rec loop () =
-    let current = Atomic.get budget_exhaustions in
-    let next, result = f current in
-    if Atomic.compare_and_set budget_exhaustions current next then result
-    else loop ()
-  in
-  loop ()
+  Stdlib.Mutex.lock budget_exhaustions_mutex;
+  Fun.protect
+    ~finally:(fun () -> Stdlib.Mutex.unlock budget_exhaustions_mutex)
+    (fun () ->
+      let next, result = f !budget_exhaustions in
+      budget_exhaustions := next;
+      result)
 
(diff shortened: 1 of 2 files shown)


anyang-keepers (commented):
## executor review — PR #13999 (fix: remove budget strike CAS spin)

**Context**: Replaces `Atomic.make` + `compare_and_set` retry loop in `oas_timeout_budget` strike ledger with `Stdlib.Mutex`-guarded `ref`. The CAS loop could spin under contention from concurrent keeper fibers updating budget exhaustion counts.

### Positive
- **Correct fix**: The critical section is pure map replacement (`Budget_strike_map.update`) — it cannot yield. A mutex is strictly better than CAS retry here because there's no benefit to lock-free progress when the operation is instant.
- **Non-yielding section preserved**: Comment explicitly notes "The critical section is pure map replacement and cannot yield." This is the key invariant — using `Stdlib.Mutex` (not `Eio.Mutex`) is correct because the section never yields to the scheduler.
- **`Fun.protect` for unlock**: Ensures mutex is released even if `f` raises. Correct exception safety.
- **Concurrent regression test**: 4 domains simultaneously bump the same keeper's strike count. Verifies no updates are lost under contention. Directly validates the fix.

### Concrete observations

1. **`keeper_turn_slot.ml:728-742`** — Clean transformation: `Atomic.t` → `ref` + `Stdlib.Mutex`. The `update_budget_exhaustions` function now has a linear code path instead of a recursive CAS loop. Simpler and correct.

2. **`peek_budget_exhaustion_for_test`** — Changed from direct `Atomic.get` + `find_opt` to `update_budget_exhaustions` with identity update. This acquires the mutex even for reads. For a test-only function, this is fine. For production, a read-only path without mutex would be slightly faster, but the overhead is negligible for a map lookup.

3. **4-domain test**: `Domain.spawn` creates 4 domains, each bumping 100 times → expected 400 total. The test joins all domains and asserts the final count. Correct concurrent validation methodology.

4. **`set_budget_exhaustion_for_test`** — Uses `update_budget_exhaustions` with direct replacement. Same mutex guard. Consistent pattern.

5. **11 deletions**: The CAS loop (`let rec loop () = ... loop ()`) is fully removed. No dead code left behind.

### Verdict
No merge blocker. Eliminates a real contention spin with correct mutex guard. Concurrent test validates the fix. Ship it.

PR #14000, opened 2026-05-07
fix: preserve missing keeper config latency

jeong-sik:
## Summary

- emit `null` instead of `0` for non-positive keeper `last_latency_ms` on dashboard keeper surfaces
- preserve missing/zero keeper config latency as `null` in the dashboard API contract
- add backend and frontend regressions for the missing-latency path while keeping positive latency unchanged

## Why

`0ms` is not a measured keeper runtime latency. Treating missing or failed measurement as a normal zero makes the dashboard look faster than reality and feeds the broader zero-latency observability problem from the redesign plan.

## Validation

- `git diff --check`
- `lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 MASC_INCLUDE_OPERATOR_CONTROL=true DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_operator_control.exe`
- `_build/default/test/test_operator_control.exe test operator 58 --show-errors`
- `pnpm exec vitest run --config vitest.config.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec tsc --noEmit --pretty`
- `pnpm exec eslint src --quiet`

## Note

The full `test_operator_control.exe` run was not used as the final validation surface: case 8 fails on a runtime-trust expectation (`Pause` vs `Alert`) outside the keeper-config latency assertion, and the broader run stayed active long enough that I interrupted it. The targeted keeper-config case added/changed here passes.

### DIFF
--- a/dashboard/src/api/dashboard.test.ts
+++ b/dashboard/src/api/dashboard.test.ts
@@ -991,6 +991,36 @@ describe('fetchKeeperConfig', () => {
     expect(result.runtime_trust?.disposition).toBe('Pass')
   })
 
+  it('preserves missing keeper config latency as null instead of zero', async () => {
+    const cases: Array<[string, Record<string, unknown>]> = [
+      ['null', { last_latency_ms: null }],
+      ['missing', {}],
+      ['zero number', { last_latency_ms: 0 }],
+      ['zero string', { last_latency_ms: '0' }],
+    ]
+
+    for (const [label, metrics] of cases) {
+      const fetchMock = vi.fn().mockResolvedValue(
+        new Response(
+          JSON.stringify({
+            name: 'keeper-sangsu',
+            metrics,
+          }),
+          {
+            status: 200,
+            headers: { 'Content-Type': 'application/json' },
+          },
+        ),
+      )
+      vi.stubGlobal('fetch', fetchMock)
+
+      const result = await fetchKeeperConfig('keeper-sangsu')
+
+      expect(result.metrics.last_latency_ms, label).toBeNull()
+      vi.unstubAllGlobals()
+    }
+  })
+
   it('preserves terminal runtime blocker classes through config fetch and display labeling', async () => {
     const cases = [
       ['no_tool_capable_provider', '도구 실행 Provider 없음'],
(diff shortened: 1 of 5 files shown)


anyang-keepers (commented):
## executor review — PR #14000 (fix: preserve missing keeper config latency)

**Context**: Emits `null` instead of `0` for non-positive keeper `last_latency_ms` on dashboard surfaces. Addresses the zero-latency observability problem — `0ms` is not a measured latency, it's a missing measurement.

### Positive
- **Correct semantics**: `null` (never measured) ≠ `0` (measured as zero). Dashboard now correctly distinguishes "no data" from "fast response."
- **Backend + frontend regression**: Both OCaml backend (`dashboard_http_keeper.ml`) and TypeScript frontend (`dashboard.ts`, `core.ts`) updated in sync. No boundary drift.
- **4-case test matrix**: `null`, missing field, `0` number, `"0"` string — all map to `null` output. Handles the real-world variety of missing-data representations.
- **Positive latency preserved**: Existing positive latency values unchanged. Non-breaking for the happy path.

### Concrete observations

1. **`dashboard_http_keeper.ml`** — `last_latency_ms` now uses `if n > 0 then Some n else None`. Simple, correct. Negative values also become `null` — appropriate since negative latency is physically meaningless.

2. **`core.ts` type change**: `last_latency_ms: number` → `last_latency_ms: number | null`. This is a breaking type change for TypeScript consumers. The test file confirms the new contract. Verify all dashboard components handle `null` gracefully.

3. **`test_operator_control_keeper.ml`** — 33 additions for backend regression. Tests both positive and non-positive paths. Uses `Masc_test_deps` for fixture construction. Clean test isolation.

4. **Partial validation note**: Author notes case 8 of `test_operator_control.exe` fails on an unrelated runtime-trust expectation. This is a pre-existing issue, not introduced by this PR. Targeted case 58 passes.

5. **Zero-latency observability**: This fix is part of a broader dashboard redesign plan. The PR correctly identifies that treating missing data as zero masks real performance issues.

### Verdict
No merge blocker. Correct semantic fix with comprehensive regression coverage. TypeScript type change needs dashboard-wide null handling verification — tests confirm this. Ship after CI green.

PR #14001, opened 2026-05-07
test: stabilize tool matrix runners

jeong-sik:
## What changed

- Keep each tool-matrix case runner inside the Eio switch until after it emits the case result and exits, so long-lived fibers from runtime state cannot keep the runner alive after a case finishes.
- Seed the MCP matrix fixture with a real admin bearer token and temp persona directory, matching the authenticated MCP path used by the live server.
- Seed a real goal for task/goal-related matrix cases and provide valid goal principal objects for goal FSM tools.
- Update keeper matrix arguments for board curation and `keeper_shell`, and treat keeper-management tools as MCP-client-only guards when exercised from keeper context.

## Why

The full matrix exposed a runner lifecycle issue: individual case processes could finish the tool call but then hang while `Eio.Switch.run` waited on long-lived fibers created by the server state. That made matrix coverage unreliable and hid real contract failures behind timeout noise.

After the runner fix, the matrix reached real tool contracts. The fixture updates align the test inputs with current auth, persona, task, board, goal, and keeper surface contracts instead of relying on stale unauthenticated or pre-goal assumptions.

## Validation

- `git diff --check`
- Before the final fixture-contract edits, focused build passed:
  `scripts/dune-local.sh build test/keeper_tool_matrix_case_runner.exe test/tool_matrix_case_runner.exe test/test_keeper_tool_matrix.exe test/test_mcp_tool_matrix.exe`
- Before the final fixture-contract edits, focused keeper matrix passed:
  `KEEPER_TOOL_MATRIX_ONLY=keeper_bash KEEPER_TOOL_MATRIX_CASE_TIMEOUT_SEC=25 _build/default/test/test_keeper_tool_matrix.exe`
- Before the final fixture-contract edits, focused MCP matrix passed:
  `MCP_TOOL_MATRIX_ONLY=masc_status MCP_TOOL_MATRIX_CASE_TIMEOUT_SEC=25 _build/default/test/test_mcp_tool_matrix.exe`
- Full keeper matrix completed instead of hanging: 134 cases ran in 153.160s, surfacing contract failures that this PR addresses in the harness.

## Local verification limitation

After the final fixture-contract edits, the focused Dune wrapper rerun could not acquire `/tmp/me-dune-local.lock` within the bounded wait because other local MASC builds were already queued/running. CI should be treated as the current compile/test authority for this pushed head.

### DIFF
--- a/test/keeper_tool_matrix_case_runner.ml
+++ b/test/keeper_tool_matrix_case_runner.ml
@@ -54,8 +54,8 @@ let () =
       let fs = Eio.Stdenv.fs env in
       let net = Eio.Stdenv.net env in
       let mono_clock = Eio.Stdenv.mono_clock env in
+      Eio.Switch.run @@ fun sw ->
       let base_path, result =
-        Eio.Switch.run @@ fun sw ->
         Cases.run_case sw ~proc_mgr ~fs ~net ~mono_clock clock schema
       in
       emit_result ~base_path tool_name result;
--- a/test/test_keeper_tool_matrix_cases.ml
+++ b/test/test_keeper_tool_matrix_cases.ml
@@ -240,6 +240,9 @@ let keeper_arguments fixture (schema : Masc_domain.tool_schema) =
   | "keeper_board_get" ->
       `Assoc [ ("post_id", `String (Generic.ensure_board_post fixture.generic)) ]
   | "keeper_board_list" -> `Assoc [ ("limit", `Int 5) ]
+  | "keeper_board_curation_read" -> `Assoc []
+  | "keeper_board_curation_submit" ->
+      `Assoc [ ("rationale", `String "tool matrix curation") ]
   | "keeper_board_comment" ->
       `Assoc
         [
@@ -264,7 +267,7 @@ let keeper_arguments fixture (schema : Masc_domain.tool_schema) =
           ("content", `String "matrix write\n");
           ("mode", `String "overwrite");
         ]
-  | "keeper_shell" -> `Assoc [ ("op", `String "git_status") ]
+  | "keeper_shell" -> `Assoc [ ("op", `String "pwd") ]
   | "keeper_bash" ->
       `Assoc [ ("cmd", `String "pwd"); ("timeout_sec", `Float 5.0) ]
   | "keeper_bash_output" ->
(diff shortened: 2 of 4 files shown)


anyang-keepers (commented):
## executor review — PR #14001 (test: stabilize tool matrix runners)

**Context**: Fixes tool matrix runner lifecycle issue where `Eio.Switch.run` waited on long-lived fibers after case completion, causing hangs. Also seeds MCP matrix fixtures with real auth tokens, persona dirs, and goal objects.

### Positive
- **Root cause correct**: `Eio.Switch.run` wrapping only the case function meant long-lived fibers from server state kept the switch alive after the case emitted its result. Moving the switch to wrap the entire runner lifecycle ensures clean exit.
- **134 cases completed in 153s**: Full matrix now completes instead of hanging. Concrete throughput evidence.
- **Fixture realism**: Seeding with real admin bearer tokens, temp persona directories, and goal principal objects aligns test inputs with production contracts. No more stale unauthenticated assumptions.
- **4-file targeted change**: `keeper_tool_matrix_case_runner.ml`, `tool_matrix_case_runner.ml` (1-line switch scope fix each), `test_keeper_tool_matrix_cases.ml` (7 lines), `test_mcp_tool_matrix_cases.ml` (61 lines). Clean diff.

### Concrete observations

1. **Switch scope change** (`tool_matrix_case_runner.ml:57`): `Eio.Switch.run` now wraps the entire `run_case` + result handling, not just the case itself. This ensures all fibers spawned during the case are cancelled when the switch exits. Correct fix.

2. **`keeper-management` tools as MCP-only guards**: When exercised from keeper context, these tools are treated as MCP-client-only. This prevents keeper matrix from testing tools that require MCP server connection. Good boundary.

3. **Board curation matrix args updated**: `keeper_board_curation_submit` and `keeper_shell` arguments updated to match current tool surface contracts. Prevents false failures from stale argument schemas.

4. **Dune lock note**: "CI should be treated as the current compile/test authority" — references qa-king's systemic finding (p-012a2fada). Same `/tmp/me-dune-local.lock` contention. Consistent with known blocker.

5. **61 additions in `test_mcp_tool_matrix_cases.ml`**: Bulk of changes are fixture seeding. Auth token, persona dir, goal FSM objects — all necessary for realistic MCP contract testing.

### Verdict
No merge blocker. Correct lifecycle fix with realistic fixture alignment. CI authority acknowledged for final validation. Ship after CI green.

Issue #9521, opened 2026-04-22
[Meta] 스트링 매칭 근본 차단: type-safe enum 강제화


jeong-sik:
17개 이슈(root-cause:STR) 반복 패턴: 문자열로 variant/event 매칭, pseudo tool call을 text로 파싱, event string 분산. 대표: #8921 #8856 #8760. 접근: 모든 문자열 기반 매칭을 variant 기반으로 전환, JSON event string도 typed variant로 파싱, lint에서 string match wildcard 경고. 검증: CI string-match grep gate.

jeong-sik:
Linked instance: #9532 — keeper_tool_alias.ml uses string-based lookup (public_to_internal_tbl) for tool names. Namespaced variants like mcp__masc__masc_status bypass the alias table because they are not structurally parsed. Type-safe enum dispatch would prevent this class of error.

jeong-sik:
CI gate 구현 완료: `scripts/ci/check-enum-string-safety.sh`

String.lowercase_ascii dispatch, raw json_string_opt on enum-like field,
List.mem string literal dispatch WARN 출력.

현재 WARN만 발생, exit 0. 강화 시점은 기존 위반 정리 후.

PR: #9551

jeong-sik:
## Tracked Instances (closed)

| Issue | Symptom | Resolution |
|-------|---------|------------|
| #9532 | mcp__masc__* namespaced tool names cause unexpected_tool_partial_tolerance WARN | String matching replaced with structured tool policy dispatch |

Type-safe enum enforcement via OCaml variants (not strings) is the target state.

jeong-sik:
Tracked under epic: #11036 — 7-axis Structural Prevention CI Gate Suite.

jeong-sik:
[...]

PR #14004, opened 2026-05-07
ci(str): fail new enum string dispatch

jeong-sik:
## Summary
- Add a diff ratchet to scripts/ci/check-enum-string-safety.sh so new enum-like string dispatch in lib/*.ml fails CI.
- Keep existing STR debt warning-only while scanning origin/main...HEAD, staged diff, and worktree diff.
- Allow explicit nearby STR-OK / STRING-BOUNDARY-OK comments for real boundary parsers.

## Why
#9521 already had a WARN-only scan, but it did not prevent new string-based enum dispatch from landing. This makes the guard preventive without forcing a mass cleanup of existing warnings.

## Validation
- bash -n scripts/ci/check-enum-string-safety.sh
- shellcheck scripts/ci/check-enum-string-safety.sh
- bash scripts/ci/check-enum-string-safety.sh --base origin/main --head HEAD
- bash scripts/ci/check-enum-string-safety.sh
- git diff --check
- git diff --cached --check
- Negative probe: temporarily added a String.lowercase_ascii literal comparison in lib/gh_command_validation.ml and verified the ratchet failed; probe was removed before commit.

Refs #9521

### DIFF
--- a/scripts/ci/check-enum-string-safety.sh
+++ b/scripts/ci/check-enum-string-safety.sh
@@ -14,7 +14,138 @@ set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 
+base_ref="origin/main"
+head_ref="HEAD"
+
+while [ "$#" -gt 0 ]; do
+  case "$1" in
+    --base)
+      if [ "$#" -lt 2 ]; then
+        echo "usage: $0 [--base REF] [--head REF]" >&2
+        exit 2
+      fi
+      base_ref="$2"
+      shift 2
+      ;;
+    --head)
+      if [ "$#" -lt 2 ]; then
+        echo "usage: $0 [--base REF] [--head REF]" >&2
+        exit 2
+      fi
+      head_ref="$2"
+      shift 2
+      ;;
+    *)
+      echo "usage: $0 [--base REF] [--head REF]" >&2
+      exit 2
+      ;;
+  esac
+done
+
 exit_code=0
+
+print_first_lines() {
+  local limit="$1"
+  awk -v limit="$limit" 'NR <= limit { print }'
+}
+
+is_enum_string_pattern() {
+  local text="$1"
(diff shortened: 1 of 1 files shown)


anyang-keepers (commented):
## executor review — PR #14004 (ci(str): fail new enum string dispatch)

**Context**: Adds diff ratchet to `scripts/ci/check-enum-string-safety.sh` so new enum-like string dispatch in `lib/*.ml` fails CI. Existing STR debt stays warning-only. Refs #9521.

### Positive
- **Mirrors #14005 architecture**: Same ratchet pattern (diff-scoped, opt-out comments, three-scan approach). Consistent CI gate design across the two PRs.
- **Pattern coverage** (`is_enum_string_pattern`): Catches `String.lowercase_ascii = "..."`, `json_string_opt` for enum-like field names (status/state/kind/type/action/mode/profile), and `List.mem [... "..."...]`. Good coverage.
- **STR-OK / STRING-BOUNDARY-OK**: Same ±2-line opt-out mechanism. Consistent with #14005's DET-OK pattern.
- **Negative probe**: Temporarily added `String.lowercase_ascii` literal comparison, verified ratchet fires, removed. Correct methodology.

### Concrete observations

1. **`json_string_opt` field name list**: Hardcoded list of `status|state|kind|type|action|mode|profile`. If new enum-like field names are added to the codebase, this list needs updating. Consider making this configurable or using a broader heuristic.

2. **False positive risk**: `List.mem ["status"; "active"]` would match even if the list isn't being used as an enum dispatch. The STR-OK escape hatch mitigates this.

3. **Consistency with #14005**: Both PRs should land together or in quick succession. Having one ratchet active without the other creates an asymmetric enforcement surface.

4. **`lib/*.ml` constraint**: Same glob as #14005. Correct — prevents ratchet from firing on test files or scripts where string dispatch is more acceptable.

### Verdict
No merge blocker. Consistent ratchet design with #14005. Ship alongside or immediately after #14005.

Issue #9522, opened 2026-04-22
[Meta] 결정론/비결정론 경계 명시화: sound partial + retry contract


jeong-sik:
24개 이슈(root-cause:DET 14 + NDT 10) 반복 패턴: '이렇게 들어올 것이다' 가정, sound partial 미적용, retry/backoff 계약 부재. 대표 DET: #8854 #8778 #8732. 대표 NDT: #8905 #8778 #8671. 접근: 파서는 확실한 경우만 Some 나머지는 None+fail-open, retry는 지수 백오프 + idempotency key 강제, 비결정적 동작은 contract로 격리. 검증: CI deterministic boundary lint.

jeong-sik:
Linked instances: #9534 (cascade assumes gemini-2.5-flash-lite supports tools deterministically; it does not) + #9533 (codex-mcp-client assumes task-031 state transition todo→start is valid without claiming first). Both assume deterministic state/capability without verification.

jeong-sik:
CI gate 구현 완료: `scripts/ci/check-determinism-contract.sh`

- Non-deterministic source in lib/ (Unix.gettimeofday, Random, etc.) WARN
- Option.value with default on unknown input WARN (5건)
- catch-all -> Some branch WARN (5건)

PR: #9551

jeong-sik:
## Tracked Instances (closed)

| Issue | Symptom | Resolution |
|-------|---------|------------|
| #9534 | gemini-2.5-flash-lite rejected for inline tools / request-scoped MCP | Provider capability filtering improved |
| #9533 | task-031 sees repeated invalid transition todo->start from codex-mcp-client | Transition validation hardened |

sound partial + retry contract is the target pattern: deterministic input handling, non-deterministic output with explicit retry boundaries.

jeong-sik:
Tracked under epic: #11036 — 7-axis Structural Prevention CI Gate Suite.

jeong-sik:
## Stale-check sweep — already-fixed candidate

`git log --oneline --all --grep '#9522'` 결과 이슈 작성일 *후* 에 머지된 fix commit 1건이 발견됐습니다.

```
2026-04-23T10:14:36+09:00 c2a4be34a7 ci(meta): add VAR, TEL, DET/NDT bug-class gates (#9518 #9520 #9522)
```

[...]

PR #14005, opened 2026-05-07
ci(det): fail new deterministic boundary debt

jeong-sik:
## Summary
- Add a diff ratchet to scripts/ci/check-determinism-contract.sh so new DET/NDT debt in lib/*.ml fails CI.
- Keep existing nondeterminism, permissive-default, and catch-all Some findings warning-only.
- Scan origin/main...HEAD, staged diff, and worktree diff, with explicit DET-OK / NDT-OK comments for intentional boundary cases.

## Why
#9522 already had a WARN-only deterministic-boundary scan. That records existing debt, but it does not prevent new wall-clock/random dependencies or unsound partial parsing defaults from landing. This PR makes the gate preventive while avoiding a mass refactor of existing warnings.

## Validation
- bash -n scripts/ci/check-determinism-contract.sh
- shellcheck scripts/ci/check-determinism-contract.sh
- bash scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD
- bash scripts/ci/check-determinism-contract.sh
- git diff --check
- git diff --cached --check
- Negative probe: temporarily added Unix.gettimeofday in lib/gh_command_validation.ml and verified the ratchet failed; probe was removed before commit.

Refs #9522

### DIFF
--- a/scripts/ci/check-determinism-contract.sh
+++ b/scripts/ci/check-determinism-contract.sh
@@ -15,12 +15,138 @@ set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)"
 
+base_ref="origin/main"
+head_ref="HEAD"
+
+while [ "$#" -gt 0 ]; do
+  case "$1" in
+    --base)
+      if [ "$#" -lt 2 ]; then
+        echo "usage: $0 [--base REF] [--head REF]" >&2
+        exit 2
+      fi
+      base_ref="$2"
+      shift 2
+      ;;
+    --head)
+      if [ "$#" -lt 2 ]; then
+        echo "usage: $0 [--base REF] [--head REF]" >&2
+        exit 2
+      fi
+      head_ref="$2"
+      shift 2
+      ;;
+    *)
+      echo "usage: $0 [--base REF] [--head REF]" >&2
+      exit 2
+      ;;
+  esac
+done
+
 exit_code=0
 
 print_first_lines() {
   local limit="$1"
   awk -v limit="$limit" 'NR <= limit { print }'
 }
+
+is_det_ndt_pattern() {
+  local text="$1"
+  if rg -q 'Unix\.gettimeofday|Random\.|Unix\.times|Sys\.time|Unix\.getpid' <<< "$text"; then
+    return 0
+  fi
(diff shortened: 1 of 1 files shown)


anyang-keepers (commented):
## executor review — PR #14005 (ci(det): fail new deterministic boundary debt)

**Context**: Adds diff ratchet to `scripts/ci/check-determinism-contract.sh` so new DET/NDT debt in `lib/*.ml` fails CI. Existing debt stays warning-only. Refs #9522.

### Positive
- **Preventive gate**: Existing WARN-only scan recorded debt but didn't prevent new introductions. This ratchet makes new wall-clock/random dependencies a hard failure.
- **Diff-scoped**: Only scans `origin/main...HEAD`, staged diff, and worktree diff — not full codebase. Correctly targeted at incremental debt.
- **Opt-out mechanism**: `DET-OK` / `NDT-OK` / `determinism-contract: allow` comments within ±2 lines allow intentional boundary cases. Flexible without weakening the gate.
- **Negative probe**: Added `Unix.gettimeofday` to verify ratchet fires, then removed. Proper validation methodology.
- **shellcheck + bash -n**: Both pass. Clean shell.

### Concrete observations

1. **Pattern coverage** (`is_det_ndt_pattern`): Catches `Unix.gettimeofday`, `Random.*`, `Unix.times`, `Sys.time`, `Unix.getpid`, `Option.value ~default:`, and `| _ -> Some`. Good coverage of common nondeterminism sources.

2. **`lib/*.ml` glob**: Path matching uses `[[ "$current_path" == lib/*.ml ]]`. This correctly matches `lib/foo.ml` but also `lib/sub/bar.ml` (glob expands). No false negatives from nested directories.

3. **Three-scan approach**: Staged diff → worktree diff → base..head diff. Exhaustive coverage of uncommitted and committed changes.

4. **Missing pattern**: `Lwt_unix.gettimeofday` (async variant) not in pattern list. If the codebase uses Lwt, this could bypass the ratchet.

5. **Exit code semantics**: `exit_code=0` for warnings, `exit_code=1` for ratchet failures. Standard and CI-friendly.

### Verdict
No merge blocker. Well-designed ratchet with proper negative probe validation. Ship it.

PR #14011, opened 2026-05-07
fix: preserve same-second runtime trust receipts

jeong-sik:
## Summary

- allow same-second execution receipts to remain current for runtime-trust snapshots
- document why receipt ISO timestamps need a one-second tolerance against fractional runtime `last_turn_ts`

## Why

Paused keeper summary snapshots could ignore a freshly written execution receipt because `ended_at` is serialized at whole-second precision while `meta.runtime.usage.last_turn_ts` retains fractional seconds. That made the runtime blocker look newer and silently replaced the receipt's `pause_human` disposition with an `Alert`.

## Validation

- `git diff --check`
- `lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 MASC_INCLUDE_OPERATOR_CONTROL=true DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_operator_control.exe`
- `_build/default/test/test_operator_control.exe test operator 8 --show-errors`

### DIFF
--- a/lib/keeper/keeper_runtime_trust_snapshot.ml
+++ b/lib/keeper/keeper_runtime_trust_snapshot.ml
@@ -190,10 +190,11 @@ let receipt_ended_at_unix receipt =
       if ts > 0.0 then Some ts else None
   | None -> None
 
-(* Receipt timestamps are serialized with lower precision than runtime
-   last-turn observations, so a tiny tolerance avoids treating same-turn
-   observations as newer blockers because of formatting jitter. *)
-let runtime_blocker_receipt_timestamp_epsilon_sec = 0.001
+(* Receipt timestamps are serialized as whole-second ISO strings, while runtime
+   last-turn observations keep fractional seconds.  A same-second receipt must
+   still be allowed to explain the blocker; otherwise the runtime blocker
+   silently overrides its operator disposition. *)
+let runtime_blocker_receipt_timestamp_epsilon_sec = 1.0
 
 let runtime_blocker_supersedes_receipt ~meta ~runtime_blocker_fields
     latest_receipt =


anyang-keepers (commented):
## executor review — PR #14011 (fix: preserve same-second runtime trust receipts)

**Context**: Increases `runtime_blocker_receipt_timestamp_epsilon_sec` from 0.001s to 1.0s. Receipts use whole-second ISO timestamps while runtime keeps fractional seconds — a same-second receipt was being silently overridden by the runtime blocker.

### Positive
- **Root cause precise**: ISO serialization truncates to whole seconds. `ended_at` = 12:00:00 vs `last_turn_ts` = 12:00:00.456 → receipt appeared older → blocker won → `pause_human` disposition silently replaced with `Alert`.
- **Minimal change**: 5 additions / 4 deletions. Single constant update + improved comment. Surgical.
- **Comment quality**: Old comment said "tiny tolerance avoids formatting jitter." New comment explains the actual semantics: whole-second ISO vs fractional runtime → same-second receipts must remain current.

### Concrete observations

1. **1.0s epsilon**: This means a receipt written at T=12:00:00.999 is considered current even if runtime shows T=12:00:01.001. The 1-second window is generous but appropriate — the alternative (lost receipts) is worse than a slightly stale one.

2. **Impact**: Without this fix, a keeper that pauses within the same second as its last turn would have its `pause_human` disposition silently discarded. This is a correctness issue — operator intent is lost.

3. **No new tests**: The PR doesn't add regression tests for the 1.0s epsilon. The existing test_operator_control.exe case 8 is noted as having a pre-existing failure on unrelated runtime-trust expectations. Worth adding a targeted test for same-second receipt preservation.

4. **Lock contention note**: Build uses `lockf -k -t 1200 /tmp/me-dune-local.lock` — references qa-king's systemic finding (p-012a2fada). Consistent with known infrastructure blocker.

### Verdict
No merge blocker. Correct precision-mismatch fix with clear semantic explanation. Consider adding a targeted test for same-second receipt preservation as follow-up. Ship after CI green.

PR #14018, opened 2026-05-07
fix: instrument memory pipeline flushes

jeong-sik:
## Summary
- add memory pipeline flush counters for AfterTurn success/error attempts
- count episodic/procedural records persisted by the OAS memory bridge
- record flush duration as a Prometheus histogram and pin coverage/tests

## Verification
- git diff --check
- lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_memory_hooks.exe test/test_prometheus_coverage.exe
- ./_build/default/test/test_memory_hooks.exe
- ./_build/default/test/test_prometheus_coverage.exe

### DIFF
--- a/lib/memory_hooks.ml
+++ b/lib/memory_hooks.ml
@@ -86,6 +86,32 @@ let compose_with_inner ~memory_hooks ~inner =
         inner.Agent_sdk.Hooks.before_turn_params
   }
 
+let record_pipeline_flush
+    ~(agent_name : string)
+    ~(outcome : string)
+    ~(duration_s : float)
+    ~(episodes : int)
+    ~(procedures : int) =
+  let outcome_labels = [ ("agent_name", agent_name); ("outcome", outcome) ] in
+  Prometheus.inc_counter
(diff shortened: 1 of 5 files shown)


anyang-keepers (commented):
**LGTM** — clean instrumentation PR.

**What's good:**
- `record_pipeline_flush` nicely factors out the metric recording logic, keeping the hook callback readable.
- Label dimensions (`agent_name`, `outcome`, `tier`) are well-chosen for debugging per-agent memory issues.
- Histogram for flush duration + counter for records gives solid observability coverage.
- Duration captured on both success and error paths — no blind spots.

**One minor note (non-blocking):**
- The error path zeros out `episodes` and `procedures` which is correct (we don't know the real counts on failure), but consider logging the actual exception message as a label value or in a structured log field for easier correlation with alerting rules. Nice-to-have, not a blocker.

Ready to merge once un-drafted.

PR #14032, opened 2026-05-07
fix(board): move delete persistence out of state lock

jeong-sik:
## Summary

- move `Board.delete_post` filesystem rewrites out of the board state mutex
- keep synchronous durability by taking the board persist lock before mutation, snapshotting under `store.mutex`, then writing the snapshots after releasing the state lock
- add regression coverage that delete rewrites persisted post/comment/vote/reaction snapshots

## Why

#10569 tracks board write RPC stalls caused by board persistence acting as a fleet-wide single point of failure. PR #13678 raised the board tool timeout; this slice reduces one remaining admin write hot path where `delete_post` rewrote posts, comments, votes, and reactions while holding `store.mutex` and bypassing the persist-lock timing bucket.

The ordering is intentional: `with_persist_lock` is acquired before the state mutation so a later writer cannot append and then be overwritten by the delete snapshot. The state lock is still released before filesystem I/O.

## Validation

- `git diff --check`
- `scripts/dune-local.sh build test/test_board_karma_ledger.exe` attempted first; stopped while queued behind `/tmp/me-dune-local.lock` host contention.
- `env DUNE_JOBS=1 MASC_SKIP_OPAM_LOCK=1 MASC_SKIP_PIN_CHECK=1 dune build --root . test/test_board_karma_ledger.exe`
- `env MASC_SKIP_PIN_CHECK=1 ./_build/default/test/test_board_karma_ledger.exe` - 16 tests passed

## Review Focus

- Whether the persist-lock-before-state-lock ordering is the right tradeoff for delete snapshot ordering.
- Whether exposing `Board_core.record_persist_error` for this internal snapshot writer is acceptable, or whether the save helpers should be promoted instead.

Refs #10569.

### DIFF
--- a/lib/board_core.mli
+++ b/lib/board_core.mli
@@ -53,6 +53,9 @@ val persist_error_count : unit -> int
     swallowed during JSONL append / rotate; consumed by the
     operator dashboard for at-a-glance health. *)
 
+val record_persist_error : where:string -> string -> unit
+(** Record and log a board persistence failure. *)
+
 (** {1 Configuration} *)
 
 val flush_interval_sec : float
--- a/lib/board_votes.ml
+++ b/lib/board_votes.ml
@@ -703,63 +703,107 @@ let set_thread_id store ~post_id ~thread_id : (unit, board_error) Result.t =
             Ok ()
       )
 
+let posts_jsonl_snapshot store =
+  let buf = Buffer.create 4096 in
+  Hashtbl.iter (fun _ (pst : post) ->
+    Buffer.add_string buf (Yojson.Safe.to_string (post_to_yojson pst));
+    Buffer.add_char buf '\n'
+  ) store.posts;
+  Buffer.contents buf
+
+let comments_jsonl_snapshot store =
+  let buf = Buffer.create 4096 in
+  Hashtbl.iter (fun _ (cmt : comment) ->
+    Buffer.add_string buf (Yojson.Safe.to_string (comment_to_yojson cmt));
+    Buffer.add_char buf '\n'
+  ) store.comments;
+  Buffer.contents buf
+
+let reactions_jsonl_snapshot store =
+  let buf = Buffer.create 4096 in
+  Hashtbl.iter
+    (fun _ (reaction : reaction) ->
+       Buffer.add_string buf (Yojson.Safe.to_string (reaction_to_yojson reaction));
+       Buffer.add_char buf '\n')
+    store.reactions;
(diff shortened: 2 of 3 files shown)


anyang-keepers (commented):
## executor review — PR #14032 (fix(board): move delete persistence out of state lock)

**Context**: Board `delete_post` was rewriting posts, comments, votes, and reactions JSONL files while holding `store.mutex`. This created a fleet-wide single point of failure for admin writes. This PR moves filesystem I/O outside the state lock.

### Positive
- **Correct lock decomposition**: Acquire persist lock → snapshot state under mutex → release mutex → write snapshots to disk. The ordering prevents a later writer from being overwritten by the delete snapshot.
- **`save_jsonl_snapshot` helper**: Extracted atomic file write with error recording. Reusable for other persistence paths.
- **Snapshot functions**: `posts_jsonl_snapshot`, `comments_jsonl_snapshot`, `reactions_jsonl_snapshot` each serialize their respective hashtables to JSONL. Clean separation of snapshot logic from write logic.
- **Test coverage**: 53 additions in `test_board_karma_ledger.ml` verifying delete rewrites all four snapshot types. Targeted regression.

### Concrete observations

1. **Lock ordering**: `with_persist_lock` acquired BEFORE `store.mutex`. This means a delete operation holds the persist lock while waiting for the state lock. If another writer holds the state lock and tries to acquire the persist lock, we get a deadlock. Verify that ALL board write paths follow the same `persist_lock → state_lock` ordering.

2. **`record_persist_error` exposure**: `Board_core.record_persist_error` is now exposed in `.mli` for use by the snapshot writer. This is an internal API — not a public surface change. Acceptable.

3. **97 additions / 53 deletions in `board_votes.ml`**: Net +44 lines. The old `delete_post` held mutex during I/O (53 lines removed). New version has snapshot functions + decomposed write path. Clean refactor.

4. **`ensure_masc_dir()` in `save_jsonl_snapshot`**: Called on every snapshot write. Could be cached or checked once at startup. Minor performance concern for high-frequency deletes.

5. **Buffer sizing**: `Buffer.create 4096` for each snapshot. For large boards, this could require multiple resizes. Consider pre-calculating size from hashtable length.

### Verdict
No merge blocker. Correct lock decomposition that directly addresses #10569 fleet-wide stall. **Critical**: Verify lock ordering consistency across ALL board write paths before merge to prevent deadlock regression. Ship after ordering audit + CI green.

PR #14034, opened 2026-05-07
fix(worktree): skip runtime-pinned cleanup candidates

jeong-sik:
## Summary
- add a shared worktree cleanup guard that detects candidate paths referenced by tmux panes, running process commands, or launchd plists
- make stale and merged worktree cleanup skip those runtime-pinned paths before removal
- extend merged cleanup with the same nested-worktree guard already used by stale cleanup

## Why
#11040 showed stale worktrees can still be production launch targets. Cleanup must not remove a worktree while a tmux/runtime process or launchd config still points at it.

## Validation
- `bash -n scripts/lib/worktree-cleanup-guards.sh scripts/cleanup-stale-worktrees.sh scripts/cleanup-merged-worktrees.sh`
- `scripts/cleanup-stale-worktrees.sh --help`
- `scripts/cleanup-merged-worktrees.sh --help`
- sourced guard smoke: synthetic runtime ref matches candidate path
- `scripts/cleanup-stale-worktrees.sh --days 99999` (dry-run, no candidates)

## Review focus
- guard false positives should prefer skipping over deletion
- no cleanup is performed unless the caller still passes `--apply`

### DIFF
--- a/scripts/cleanup-merged-worktrees.sh
+++ b/scripts/cleanup-merged-worktrees.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Safely remove worktrees whose branch is already merged into main.
-# Never uses --force.  Respects uncommitted changes.
+# Never uses --force. Respects uncommitted changes, nested worktrees, and
+# runtime-pinned paths.
 #
 # Usage:
 #   ./scripts/cleanup-merged-worktrees.sh          # dry run, shows candidates
@@ -17,6 +18,9 @@ for arg in "$@"; do
     *) echo "unknown arg: $arg" >&2; exit 2 ;;
   esac
 done
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+. "$SCRIPT_DIR/lib/worktree-cleanup-guards.sh"
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "$REPO_ROOT"
@@ -44,6 +48,8 @@ fi
 removed=0
 skipped=0
 dirty=0
+nested=0
+active=0
 
 while read -r branch; do
   [ -z "$branch" ] && continue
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):
## executor review — PR #14034 (fix(worktree): skip runtime-pinned cleanup candidates)

**Context**: Adds a shared cleanup guard (`worktree-cleanup-guards.sh`) that detects when a worktree path is still referenced by tmux panes, running process commands, or launchd plists. Stale and merged worktree cleanup scripts now skip these runtime-pinned paths before removal.

### Positive
- **Safety-first**: Guard prefers skipping over deletion. False positives (skipping a removable worktree) are safer than false negatives (deleting an active one).
- **Three detection methods**: `tmux list-panes` parsing, `/proc/*/cmdline` scanning, `launchctl list` plist path extraction. Covers the three common runtime reference patterns.
- **Consistent integration**: Both `cleanup-stale-worktrees.sh` and `cleanup-merged-worktrees.sh` source the same guard library. Single source of truth.
- **Nested worktree guard**: Already existed in stale cleanup, now extended to merged cleanup. Prevents removing a parent worktree that contains child worktrees.
- **Dry-run by default**: `--apply` flag required for actual removal. Guard runs in dry-run too, showing which paths would be skipped.

### Concrete observations

1. **`worktree_cleanup_is_runtime_referenced`**: Core guard function. Checks tmux → processes → launchd in sequence. Short-circuits on first match. Efficient.

2. **Process scanning**: `ls /proc/$PID/cmdline 2>/dev/null` and `tr '\0' ' '` — handles Linux procfs. What about macOS? macOS doesn't have `/proc`. The guard may silently skip process scanning on macOS. Consider adding `ps aux | grep` as fallback for macOS.

3. **tmux parsing**: `tmux list-panes -a -F '#{pane_current_command} #{pane_start_dir}'` — captures current command and start directory. What if the worktree path appears in an argument but not in `pane_start_dir`? The guard checks if candidate path is a prefix of the combined string. Good heuristic but could miss indirect references.

4. **`launchctl list` plist extraction**: Uses `plutil -convert xml1 -o -` to read plist paths. Handles the macOS service case correctly.

5. **80 additions / 5 deletions**: Net +75 lines. `worktree-cleanup-guards.sh` is 46 lines (new file). Clean extraction.

6. **Summary counters**: `nested=` and `active=` counters added to summary output. Operators can see how many worktrees were skipped for each reason. Good observability.

### Verdict
No merge blocker. Safety-critical guard for worktree cleanup. Minor: macOS process scanning gap (no `/proc`). The tmux and launchd checks cover the primary macOS cases. Ship after CI green.

Issue #14021, opened 2026-05-07
Audit run-id and keeper-msg proof gaps in Docker PR lifecycle validation


jeong-sik:
## Symptoms

During live Docker PR lifecycle proof run `codex-analyst-verifier-20260507b`, analyst and verifier both produced durable Docker-backed create/push/approve evidence, and GitHub shows cross-keeper approvals on PR #14014 and PR #14016.

However:

1. `scripts/audit-keeper-fleet-readiness.py --require-docker-pr-lifecycle-evidence --evidence-run-id codex-analyst-verifier-20260507b` still reports `docker_pr_approve_evidence_missing` for analyst/verifier. Non-run-id audit passes both keepers.
2. `masc_keeper_msg_result` for analyst request `kmsg_analyst_7_1778117026524` ended with a required-tool contract error even though durable logs and GitHub prove `keeper_pr_review_read` and `keeper_pr_review_comment` succeeded.

## Reproduction / Evidence

- Analyst PR: #14014, proof-only file `docs/runtime-proof/keepers/analyst-codex-analyst-verifier-20260507b.md`, approved by verifier.
- Verifier PR: #14016, proof-only file `docs/runtime-proof/keepers/verifier-codex-analyst-verifier-20260507b.md`, approved by analyst.
- Audit output files from operator run:
  - `/private/tmp/masc-audit-docker-pr-lifecycle-after-analyst-20260507.json`
  - `/private/tmp/masc-audit-docker-pr-lifecycle-runid-codex-analyst-verifier-20260507b-after-analyst.json`
- Durable metric rows:
  - `/Users/dancer/me/.masc/keepers/analyst/pr-action-metrics/2026-05/07.jsonl` has APPROVE for PR 14016 at 2026-05-07T01:24:37Z.
  - `/Users/dancer/me/.masc/keepers/verifier/pr-action-metrics/2026-05/07.jsonl` has APPROVE for PR 14014 at 2026-05-07T01:14:07Z and later repeats.

## Suspected Cause

Run-id filtering uses `row_mentions_evidence_run_id`. Approval rows often do not carry the literal run id: metrics rows store `pr_number` but not review body, and global tool-call rows redact the run id / branch in review bodies. So run-id-specific audit drops valid approve evidence.

The keeper-msg failure appears to be a final-turn contract accounting issue: the required tool was used successfully during the request, but a later no-tool response was evaluated as if the entire request lacked `keeper_pr_review_comment`.

## Tried

[...]

PR #14002, opened 2026-05-07
fix(keeper): reload goal scope for claims

jeong-sik:
## Summary

- Enforce the latest-scope-at-claim contract for `keeper_task_claim` by reloading keeper meta before resolving `active_goal_ids` and tool access.
- Add `claim_scope.source` so the claim result shows whether scope came from latest persisted meta, registry fallback, or the caller snapshot.
- Add a regression that observes claimable work under stale scope G1, persists scope G2 before claim, and verifies the claim follows G2 instead of the stale observation.

## Why

Observation and claim are separate operations. Without an explicit scope token, the safest contract is that claim time is authoritative. This prevents a keeper from claiming work using a stale active-goal snapshot after an operator or runtime update changes its scope.

## Validation

- `git diff --check`
- `env DUNE_BUILD_DIR=_build_codex_13738_goal_scope_race scripts/dune-local.sh build test/test_keeper_task_dispatch.exe`
- `./_build_codex_13738_goal_scope_race/default/test/test_keeper_task_dispatch.exe test claim` (25 tests)

Closes #13738

## TODO (Symptom 억제 후속 RFC 의무)

`claim_scope.source` 필드는 출처를 *기록*하는 것 자체가 observation/claim 사이 race가 빈발한다는 자인. typed scope token이 흐르지 않는 root는 미해결.

- **TODO**: scope token을 phantom-typed로 observation→claim 사이에 thread하는 RFC 작성. 별도 RFC 머지 시 `claim_scope.source` 필드 deprecate.
- 머지는 가능 (production race 차단 우선), 단 RFC 번호 부여 후 본 PR body에 링크.

근거: `instructions/software-development.md` §워크어라운드 거부 기준 §Symptom 억제 패턴 (PR jeong-sik/me#1100).

PR #14035, opened 2026-05-07
fix: stop repeating satisfied keeper required tools

jeong-sik:
## Summary

Fixes the remaining runtime-contract half of #14021.

A `masc_keeper_msg` request can pass `required_tools` such as `keeper_pr_review_comment`. During a multi-turn OAS `Agent.run`, the first internal turn may successfully call the required tool, but later internal turns were still given the same required-tool surface and `tool_choice`. That let OAS fail the overall async result with a required-tool contract error after the evidence-producing tool call had already succeeded.

This change tracks successful keeper tool calls inside the current `Agent.run` and removes already-satisfied required tools from later per-turn tool surfaces. If the required list is fully satisfied, stale `Any` or specific `Tool` choices from that required-tool path are cleared so the final text or state turn can complete normally.

Follow-up: explicit per-message `required_tools` now use provider-level `Any` rather than exact `Tool(name)`. Runtime MCP providers can report names like `mcp__masc__keeper_pr_create`; OAS exact-tool contracts compare raw names before MASC canonicalizes them, so a correct PR tool call could still fail as `require_specific_tool(...)`. MASC keeps the exact required-tool validation after execution and adds a `[REQUIRED TOOLS]` system-context nudge to preserve model guidance.

## Impact

- Prevents false async `masc_keeper_msg_result` failures after a successful required tool call.
- Prevents `require_specific_tool(keeper_pr_create|keeper_pr_review_comment)` failures caused by MCP-prefixed raw tool names.
- Keeps unsatisfied required tools strict; only successful keeper tools that can satisfy required-tool contracts are removed from the outstanding set.
- Preserves existing required-tool gate ordering before last-turn relaxation.

## Validation

- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_unified.exe test/test_ci_hardening_source.exe`
- `./_build/default/test/test_keeper_unified.exe test tool_classification`

## Residual risk

- Draft/human policy gate remains expected for this `agent-pr` PR until a human applies the required bypass label.

### DIFF
--- a/lib/keeper/keeper_agent_tool_surface.ml
+++ b/lib/keeper/keeper_agent_tool_surface.ml
@@ -297,6 +297,30 @@ let required_tool_names_for_turn ~(current_task_required_tool_names : string lis
   | [] -> current_task_required_tool_names
   | _ :: _ -> per_call_required_tool_names
 
+let outstanding_required_tool_names ~(required_tool_names : string list)
+    ~(satisfied_tool_names : string list) =
+  let satisfied =
+    satisfied_tool_names
+    |> List.map Keeper_tool_disclosure.canonical_tool_name
+    |> Keeper_types.dedupe_keep_order
+  in
+  required_tool_names
+  |> List.filter (fun name ->
+    let canonical = Keeper_tool_disclosure.canonical_tool_name name in
+    not (List.mem canonical satisfied))
+  |> Keeper_types.dedupe_keep_order
+
+let satisfied_required_tool_names_of_outcomes
+    (calls : (string * string) list) =
+  calls
+  |> List.filter_map (fun (tool_name, outcome) ->
+    if String.equal outcome "ok"
+       && Keeper_tool_disclosure.tool_name_can_satisfy_required_contract
+            tool_name
+    then Some tool_name
+    else None)
+  |> Keeper_types.dedupe_keep_order
+
 let preferred_tool_choice_for_required_tool_names
     ~(required_tool_names : string list) ~(allowed_tool_names : string list) =
   let visible_required =
--- a/lib/keeper/keeper_agent_tool_surface.ml
+++ b/lib/keeper/keeper_agent_tool_surface.ml
@@ -331,8 +331,14 @@ let preferred_tool_choice_for_required_tool_names
     |> Keeper_types.dedupe_keep_order
   in
   match visible_required with
-  | [ name ] -> Agent_sdk.Types.Tool name
-  | _ :: _ -> Agent_sdk.Types.Any
+  | _ :: _ ->
+    (* Use the provider-level "some tool is required" contract here, even
+       for a single explicit required tool. Runtime MCP transports may return
+       names such as [mcp__masc__keeper_pr_create]; OAS exact-tool contracts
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
## executor review — PR #14035 (stop repeating satisfied keeper required tools)

**Context**: Fixes a false-failure pattern in OAS `Agent.run`: after a keeper tool call successfully satisfies `required_tools` in turn N, turns N+1..M still received the same required-tool surface and `tool_choice`. If the final text turn doesn't call another tool, the async result records a `require_tool_use` contract violation — even though the tool was already called successfully.

This is **exactly the root cause** of executor's 60+ `require_tool_use` failures documented in imseonghan's diagnosis (p-e3672a10).

### Positive
- **Precise fix**: `outstanding_required_tool_names` filters already-satisfied tools from the per-turn surface. Only unsatisfied tools remain strict.
- **`tool_choice` clearing**: When all required tools are satisfied, stale `Any`/specific `Tool` choices from that path are cleared so the final text/state turn completes normally. Correct — the contract was already met.
- **Canonical name matching**: Uses `Keeper_tool_disclosure.canonical_tool_name` for comparison, not raw strings. Handles aliasing correctly.
- **Test coverage**: Cases 3, 6, 45, 46, 54, 55 all pass. Specific tests for tool classification and metrics observation.
- **4-file targeted change**: `keeper_agent_tool_surface.ml/mli` (API), `keeper_run_tools.ml` (tracking), `test_keeper_unified.ml` (tests). Clean scope.

### Concrete observations

1. **`keeper_run_tools.ml:track_satisfied`**: Tracks successful keeper tool calls within the current `Agent.run`. This is per-run state, not per-session — correct boundary.

2. **Dune lock note**: CI is the final validation surface due to shared `dune-local` lock contention. Same systemic issue as qa-king's finding (p-012a2fada).

3. **55 additions / 1 deletion**: Minimal footprint. The `outstanding_required_tool_names` function is 12 lines with clear intent.

4. **Edge case**: What if a required tool is called but fails (exception/timeout)? It should remain in the outstanding set. The PR says "only successful keeper tools that can satisfy required-tool contracts are removed" — this implies failed calls stay outstanding. Verify this in the tracking logic.

5. **Companion to #14002**: #14002 fixes stale goal scope at claim time. #14035 fixes stale required-tool surface at turn time. Together they address the two structural causes of executor's failure pattern.

### Verdict
No merge blocker. Directly fixes the structural cause of executor's `require_tool_use` violations. Critical fix — should ship alongside #14002. CI green required (dune lock may delay).

jeong-sik:
Review follow-up addressed in fb64cbd8b8. I extracted the successful-required-tool calculation into a testable helper and added regressions showing outcome=error stays outstanding while outcome=ok satisfies the required tool. Local git diff --check passed; focused Dune build could not run because /tmp/me-dune-local.lock is held by another verification build, so GitHub checks are the validation surface for this amended head.

jeong-sik:
셀프 리뷰 (fork1 audit): PR 14087과 conflict risk. 두 PR이 lib/keeper/keeper_agent_tool_surface.ml의 동일 함수 preferred_tool_choice_for_required_tool_names를 반대 방향으로 수정. 머지 전 통합 검증 권장.

PR #14057, opened 2026-05-07
refactor(cascade_runtime): replace local_only substring scan with allowlist

jeong-sik:
## Summary

Retires the substring scan on `"local"` in `is_local_only_cascade` in favor of an explicit allowlist canonicalized against the cascade catalog. Behavior is preserved for the only currently-defined matching cascade — `local_recovery` in `config/cascade.toml`.

## Why

The substring heuristic carried two latent failure modes that would silently regress provider routing the moment a new cascade name was added to the catalog:

- **False positives**: any future cascade name embedding `"local"` as a substring (e.g. `"localhost_special"`) would be routed local-only even when its provider list included cloud entries.
- **False negatives**: semantically local-only cascades that do not embed `"local"` in the name (e.g. a hypothetical `"pure_ollama"` or `"on_prem"`) would fall through to the cloud-permitting branch, defeating their purpose.

The substring approach was a heuristic over a name; the question is structural: "is every entry in `{name}_models` a local provider?" Answering that requires `Cascade_config_loader` to surface the per-cascade model list on `catalog_entry`, which is an **RFC-0027 follow-up** — tracked in the inline comment.

The explicit allowlist makes the current set auditable, removes both hypothetical regression modes, and forces any future addition to be a one-line catalog entry that a human reviews — instead of the cascade name accidentally crossing the substring trap.

## Diff shape

```
lib/cascade/cascade_runtime.ml | 29 +++++++++++++++++++----------
1 file changed, 19 insertions(+), 10 deletions(-)
```

Net +9 LOC. Pure refactor; the only call site is the same-file consumer at L50 (`default_model_strings`).

## Test plan

- [x] Local build: `dune build lib/cascade` exit=0, no errors involving `cascade_runtime`.
- [x] Behavior parity: `cascade.toml` declares only `local_recovery` matching the substring "local". The new allowlist contains exactly that name, canonicalized + lowercased, so `is_local_only_cascade "local_recovery"` continues to return `true` and any other cascade returns `false` (same as before).
- [ ] CI green

## Follow-up (out of scope)

- RFC-0027 PR: extend `Cascade_config_loader.catalog_entry` with `provider_kinds : Provider_adapter.kind list` derived from `{name}_models`. Then `is_local_only_cascade` becomes `Cascade_config_loader.entry_is_local_only` directly off the catalog.

## RFC

Not required. `lib/cascade/*` is not in CLAUDE.md `agent_delegation` RFC-gated list. The behavioral change is null; the comment-tracked follow-up is the part that would warrant an RFC if implemented.

### DIFF
--- a/lib/cascade/cascade_runtime.ml
+++ b/lib/cascade/cascade_runtime.ml
@@ -16,17 +16,28 @@ let provider_name_of_label (label : string) : string option =
       if idx = 0 then None
       else Some (String.sub label 0 idx |> String.trim |> String.lowercase_ascii)
 
+(* Cascades whose attempts must be filtered to local providers only.
+
+   Explicit allowlist canonicalized against the cascade catalog.  The
+   prior substring scan on ["local"] was retired because it
+   false-positived hypothetical names like ["localhost_special"] and
+   false-negatived semantically local cascades that do not embed
+   ["local"] in the name (e.g. ["pure_ollama"]).
+
+   Production today defines only [local_recovery] in [config/cascade.toml]
+   so behavior is preserved.  The catalog-derived implementation —
+   "is_local iff every entry in [{name}_models] resolves to a provider
+   for which [Provider_adapter.is_local_provider] is true" — is the
+   right long-term fix; tracked as an RFC-0027 follow-up because it
+   needs [Cascade_config_loader] to surface the per-cascade model list
+   on [catalog_entry]. *)
+let local_only_cascades = [ "local_recovery" ]
+
 let is_local_only_cascade name =
-  let lc = name |> Keeper_cascade_profile.canonicalize |> String.lowercase_ascii in
-  let pattern = "local" in
-  let plen = String.length pattern in
-  let slen = String.length lc in
-  let rec loop i =
-    if i > slen - plen then false
-    else if String.sub lc i plen = pattern then true
-    else loop (i + 1)
+  let canonical =
+    name |> Keeper_cascade_profile.canonicalize |> String.lowercase_ascii
   in
-  loop 0
+  List.mem canonical local_only_cascades
 
 let is_local_label label =
   match provider_name_of_label label with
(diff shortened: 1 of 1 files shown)


jeong-sik:
셀프 리뷰: PR 14019와 동일 함수 수정 충돌

이 PR은 lib/cascade/cascade_runtime.ml의 is_local_only_cascade를 allowlist 방식으로 수정한다. 그런데 PR 14019도 동일 함수를 explicit match set으로 수정한다.

- 14019: match lc with "local" | "local_only" | "ollama" | ... | s when String.starts_with ~prefix:"local_" s -> true
- 14057: let local_only_cascades = ["local_recovery"]

두 PR이 모두 merge되면 cascade_runtime.ml에서 conflict가 발생한다. 하나만 선택해야 한다.

allowlist 방식(14057)이 더 explicit하고 주석에 RFC-0027 follow-up이 있어 장기적으로 더 적절해 보인다. 14019의 "local_" prefix catch-all은 여전히 heuristic 성격이 있다.

jeong-sik:
Triage 업데이트: 충돌 우려했던 #14019가 `CLOSED` (미머지) 상태로 확인됨. allowlist 방식인 이 PR이 단독 채택 경로로 정리됨. `human-approved-ready` 라벨 대기.

jeong-sik:
RFC-0041 (#14185)이 `cascade_name: string`을 `CascadeRef.t`로 교체하면서 본 PR의 substring scan / allowlist 둘 다 무효화됩니다. PR 본문도 "RFC-0027 follow-up"이라 적어 stop-gap임을 이미 인정한 상태입니다.

`is_local_only_cascade`의 구조적 답("모든 `{name}_models` 엔트리가 local provider인가")은 RFC-0041의 typed `CascadeRef` + `Cascade_config_loader`가 `catalog_entry`에 model list를 노출하는 설계로 자연 해결됩니다. 이 PR은 obsolete로 close합니다.

PR #14059, opened 2026-05-07
fix: count recurring task auto-disable failures

jeong-sik:
## Summary
- count recurring-task auto-disable transitions with the existing keeper recurring failure metric
- keep per-dispatch exception counting unchanged while surfacing the threshold crossing that disables the task
- extend the focused recurring-task regression to prove the metric only increments when max failures disables the task

## Why it matters
Recurring tasks can be disabled after repeated dispatcher Error results. That state transition is operationally important because it can silence keeper heartbeat broadcasts until a later re-enable pass. This makes the auto-disable point visible without adding a new metric schema.

## Verification
- git diff --check
- lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_recurring.exe
- ./_build/default/test/test_keeper_recurring.exe

### DIFF
--- a/lib/keeper/keeper_recurring.ml
+++ b/lib/keeper/keeper_recurring.ml
@@ -86,6 +86,12 @@ let list_all () =
   with_tasks_ro (fun () ->
     Hashtbl.fold (fun _id task acc -> task :: acc) tasks [])
 
+let record_failure ~task ~phase =
+  Prometheus.inc_counter
+    Prometheus.metric_keeper_recurring_failures
+    ~labels:[("task", task.id); ("phase", phase)]
+    ()
+
 (* ================================================================ *)
 (* Dispatch                                                          *)
 (* ================================================================ *)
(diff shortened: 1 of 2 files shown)


jeong-sik:
$(cat <<EOF
## 검토 코멘트: Prometheus label cardinality 위험

**위치**: `lib/keeper/keeper_recurring.ml:89`

```ocaml
~labels:[("task", task.id); ("phase", phase)]
```

`task.id`를 Prometheus label 값으로 사용하면 각 고유한 task마다 별도의 time series가 생성됩니다. recurring task가 동적으로 생성되는 경우 Prometheus 메모리가 비대화될 수 있습니다. 이는 Prometheus 모범 사례에서 명시적으로 경고하는 high-cardinality label anti-pattern입니다.

**권장**: task label을 제거하거나, bounded한 카테고리(예: task 타입)로 대체합니다.
EOF
)

PR #14085, opened 2026-05-07
fix: count keeper chat store read drops

jeong-sik:
Summary
- Count malformed keeper chat JSONL rows with masc_persistence_read_drops_total surface=keeper_chat_store reason=entry_load_error.
- Count syntactically valid rows missing non-empty role/content with reason=invalid_payload.
- Add a focused keeper_chat_store regression proving valid rows survive while both drop counters increment.

Why it matters
- Keeper direct-message history is part of the operator coordination surface. Damaged chat rows previously disappeared silently, which hid persistence corruption and made chat context gaps hard to diagnose.

Verification
- git diff --check
- scripts/opam-pin-external-deps.sh --install
- lockf -k -t 1200 /tmp/me-dune-local.lock env MASC_DUNE_LOCK_HELD=1 MASC_SKIP_OPAM_LOCK=1 DUNE_JOBS=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_chat_store.exe
- ./_build/default/test/test_keeper_chat_store.exe

### DIFF
--- a/lib/keeper/keeper_chat_store.ml
+++ b/lib/keeper/keeper_chat_store.ml
@@ -17,6 +17,22 @@ let chat_dir base_dir =
 let chat_path ~base_dir ~keeper_name =
   Filename.concat (chat_dir base_dir) (sanitize_name keeper_name ^ ".jsonl")
 
+let persistence_surface = "keeper_chat_store"
+
+let record_persistence_read_drop ~reason () =
+  Prometheus.inc_counter
+    Prometheus.metric_persistence_read_drops
+    ~labels:[("surface", persistence_surface); ("reason", reason)]
+    ()
+
+let report_persistence_read_drop ~reason ~path ~detail =
+  Safe_ops.report_persistence_read_drop
+    ~on_drop:(fun () -> record_persistence_read_drop ~reason ())
+    ~surface:persistence_surface
+    ~reason
+    ~path
+    ~detail
+
 let ensure_dir_once ~base_dir =
   ignore (Keeper_fs.ensure_dir (chat_dir base_dir))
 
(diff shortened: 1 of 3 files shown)


anyang-keepers:
## executor review — PR #14085 (keeper_chat_store read drops)

Verified by reading diff (gh pr diff 14085).

### Positive
- **Two distinct drop reasons**: entry_load_error (Json_error / Sys_error) vs invalid_payload (empty role/content). Correct separation — different failure modes need different operator response.
- **Sys.file_exists guard before load_file**: Eliminates a noisy entry_load_error increment for the legitimate "keeper has no chat history yet" case. Without this guard every fresh keeper would emit a spurious drop.
- **file_path threaded into parse_line**: Detail field on the drop counter now carries the actual path, which an operator needs to grep for the corrupt row.
- **Cancellation safety preserved**: Outer try only catches Yojson.Json_error; the inner ts-parse keeps its Eio.Cancel.Cancelled re-raise. No regression in cancellation propagation.
- **Sys_error to drop counter**: Previously Sys_error caused silent empty list. Now counted.
- **Targeted regression**: test_keeper_chat_store.ml writes 4 rows (valid, malformed json, missing content, valid) and asserts both drop counters increment by 1 and 2 valid messages survive. Direct validation of the fix.

### Concrete observations

1. **parse_line signature change**: parse_line : string -> chat_message option becomes parse_line : file_path:string -> string -> chat_message option. Only one call site at load. No external caller broken.

2. **load early-return formatting**: `if not (Sys.file_exists path) then [] else try ... with ...` — the else without braces is legal but unusual; the entire try/with block is the else-branch expression. Worth noting for future maintainers.

3. **metric_keeper_chat_store_failures retained**: The outer exn arm still bumps the legacy counter. Intentional — keeps backward compat for existing dashboards while new counter rolls out.

### Verdict
No merge blocker. Clean observability fix with regression test. Ship after CI green.

— executor

PR #14088, opened 2026-05-07
fix(board): move sub-board persistence out of state lock (#10569)

jeong-sik:
## Summary

- Move sub-board create/delete persistence out of the board state mutex.
- Keep full delete rewrites serialized by `with_persist_lock`, snapshotting under `store.mutex` and writing after the state lock is released.
- Add regression coverage that sub-board create/delete updates the persisted JSONL snapshot.

## Product impact

This removes another synchronous filesystem write path from the shared board state lock. It is a narrow follow-up to #10569 after the board timeout and delete-post persistence slices, reducing the remaining board writer SPOF surface without changing the JSONL wire format or synchronous durability behavior.

## Evidence

- `git diff --check` passed.
- `git diff --cached --check` passed.
- `env DUNE_JOBS=1 MASC_SKIP_OPAM_LOCK=1 MASC_SKIP_PIN_CHECK=1 dune build --root . test/test_board_dispatch.exe` passed.
- `env MASC_SKIP_PIN_CHECK=1 ./_build/default/test/test_board_dispatch.exe` passed: 51 tests.
- `ocamlformat --check lib/board_core.ml lib/board_core.mli test/test_board_dispatch.ml` is blocked by the existing `lib/board_core.ml` unattached docstring warning at lines 185-193; no whole-file reformat applied.

## Review evidence

- `scripts/dune-local.sh build test/test_board_dispatch.exe` was attempted first and stopped while queued behind `/tmp/me-dune-local.lock`.
- `create_sub_board` mutates in memory under `store.mutex`, then appends under `with_persist_lock`.
- `delete_sub_board` holds `with_persist_lock`, snapshots under `store.mutex`, then rewrites after releasing the state lock so concurrent append/rewrite ordering remains serialized.

## Linked issue

Refs #10569.

### DIFF
--- a/lib/board_core.ml
+++ b/lib/board_core.ml
@@ -1156,19 +1156,26 @@ let sub_board_of_yojson (json : Yojson.Safe.t) : sub_board option =
        | _ -> None)
   | _ -> None
 
-let rewrite_sub_boards store =
+let sub_boards_jsonl_unlocked store =
+  let buf = Buffer.create 1024 in
+  Hashtbl.iter (fun _ (sb : sub_board) ->
+    Buffer.add_string buf (Yojson.Safe.to_string (sub_board_to_yojson sb));
+    Buffer.add_char buf '\n'
+  ) store.sub_boards;
+  Buffer.contents buf
+
+let save_sub_boards_jsonl content =
   try
     ensure_masc_dir ();
     let path = sub_boards_path () in
-    let buf = Buffer.create 1024 in
-    Hashtbl.iter (fun _ (sb : sub_board) ->
-      Buffer.add_string buf (Yojson.Safe.to_string (sub_board_to_yojson sb));
-      Buffer.add_char buf '\n'
-    ) store.sub_boards;
-    (match Fs_compat.save_file_atomic path (Buffer.contents buf) with
+    (match Fs_compat.save_file_atomic path content with
      | Ok () -> ()
      | Error msg -> record_persist_error ~where:"rewrite_sub_boards" msg)
   with Sys_error msg -> record_persist_error ~where:"rewrite_sub_boards" msg
+
+let rewrite_sub_boards store =
+  let content = with_lock store (fun () -> sub_boards_jsonl_unlocked store) in
+  with_persist_lock store (fun () -> save_sub_boards_jsonl content)
 
 let append_sub_board (sb : sub_board) =
   try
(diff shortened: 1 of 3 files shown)


anyang-keepers:
## executor review — PR #14088 (fix(board): move sub-board persistence out of state lock)

**Context**: Restructures `rewrite_sub_boards`, `create_sub_board`, and `delete_sub_board` so that file I/O (`save_file_atomic`) no longer runs while holding the state-lock. Snapshot under state-lock, then persist under a separate `persist_lock`.

### Positive
- **Correct decoupling**: `sub_boards_jsonl_unlocked` builds the JSONL string from the in-memory hashtable; `save_sub_boards_jsonl` does the atomic write. Splitting these is the right surgical move — no functional change, only locking shape.
- **`rewrite_sub_boards`**: clean sequential pattern — `with_lock` releases before `with_persist_lock` acquires. No lock nesting. Two threads calling rewrite serialize cleanly through the persist-lock without blocking other state-lock readers.
- **`create_sub_board`**: result computed and returned to caller only after `append_sub_board` completes under persist-lock. Caller's `Ok sb` is still durability-acked.
- **Regression test (`test_sub_board_create_delete_persisted_snapshot`)**: directly reads `sub_boards_path()` from disk via `Fs_compat.load_jsonl` and asserts slug presence after create + absence after delete. End-to-end persistence verification, not mocked.
- **`.mli` doc updates**: explicit "snapshot under state lock, persist outside state lock" + "JSONL append persistence is performed outside the state lock" — matches implementation, prevents future drift.

### Concrete observations / questions

1. **Lock ordering inconsistency between `delete_sub_board` and `rewrite_sub_boards`** — worth verifying:
   - `delete_sub_board`: `with_persist_lock { with_lock { ... } }` → **persist→state** nested ordering.
   - `rewrite_sub_boards`: `with_lock { ... }` then `with_persist_lock { ... }` → sequential, no nesting.
   - `create_sub_board`: sequential, same as rewrite.

   The nested case in `delete_sub_board` introduces a Persist→State ordering. **If any other code path holds the state-lock and then calls something that acquires the persist-lock**, classic deadlock setup (T1: state→persist, T2: persist→state). Recommend grep'ing all `with_lock` call sites that internally call `append_sub_board`, `rewrite_sub_boards`, or `save_sub_boards_jsonl` to confirm no nested State→Persist exists. The original code held both transitively via the same state-lock, so the new ordering is the first place this distinction matters.

2. **Partial-durability window in `create_sub_board`**: between state-lock release and persist-lock acquisition, the in-memory store has the new entry but disk does not. Concurrent readers see it; process crash loses it. This is acceptable because:
   - The caller's `Ok sb` is only returned after persist completes.
   - On restart, the JSONL replay simply omits the unpersisted entry (no inconsistency, just a forgotten create).
   - No invariant promises in-memory ⊆ disk during a write.

   Worth a one-line comment in `create_sub_board` acknowledging the window for future readers.

3. **`delete_sub_board` snapshot-then-write semantics**: the `Ok (sub_boards_jsonl_unlocked store)` is computed *after* the hashtbl removals, so the saved snapshot reflects post-delete state. Correct ordering.

4. **`append_sub_board` (line 1175 region, not in diff)**: I can't see whether it acquires the state-lock internally. If it does, the new `with_persist_lock store (fun () -> append_sub_board sb)` call in `create_sub_board` becomes a Persist→State nested case too. Worth confirming `append_sub_board` only does file I/O without state-lock acquisition.

5. **No dashboard/UI surface impact**: this is a pure persistence refactor. The state machine externals (`get_sub_board`, `list_sub_boards`) are unchanged.

### Verdict
Not a merge blocker. The refactor is sound and the regression test directly validates disk consistency. Strongly recommend the lock-ordering audit (observation #1) before ship — `rg "with_lock store" lib/board_core.ml | wc -l` and check each for nested calls into persist-acquiring helpers. If audit clean, ship.

jeong-sik:
## Cross-lock deadlock in \`delete_sub_board\`

**lib/board_core.ml:129-152**

\`delete_sub_board\`의 lock 획득 순서가 \`with_persist_lock\` → \`with_lock\`인데, 다른 함수들은 반대 순서를 사용합니다.

| 함수 | lock 순서 |
|------|----------|
| \`rewrite_sub_boards\` | state_lock → persist_lock |
| \`create_sub_board\` | state_lock → persist_lock |
| \`delete_sub_board\` | persist_lock → state_lock |

**교차 잠금 시나리오:**
- Fiber A (delete): persist_lock 획득 → state_lock 대기
- Fiber B (rewrite): state_lock 획득 → persist_lock 대기
- 결과: deadlock

Eio.Mutex는 재진입을 허용하지 않습니다. \`delete_sub_board\`를 state_lock → persist_lock 순서로 재구조화하거나, \`save_sub_boards_jsonl\` 호출을 lock 밖으로 이동하는 것을 권장합니다.

PR #14104, opened 2026-05-07
fix(keeper): route actionable turns to tool-capable providers

jeong-sik:
## Summary

Fixes a keeper routing gap visible in the operator logs: an actionable structured world signal could be enforced only after the OAS run completed, while provider selection still treated the turn as optional-tool. That let `codex_cli` silently omit keeper-bound runtime MCP tools, then the keeper failed later with `required_tool_contract` / `tool_required_unsatisfied`.

Changes:

- Add a pure `Keeper_contract_classifier.requires_tool_support_for_allowed_tools` helper.
- Make `Keeper_agent_run` require a tool-capable provider when the structured world observation has actionable work that the keeper can satisfy with its tool surface.
- Report `No_tool_capable_provider` when `require_tool_support` alone empties a cascade, not only when exact `tool_choice` support was requested.
- Correct the pure classifier test to keep `contract_status_label` low-cardinality and assert missing tool details via `pp_contract_status`.

## Product impact

Actionable turns with claimable tasks, board activity, or discovered work should now route away from providers that cannot expose keeper-bound runtime MCP tools before the model gets a chance to emit a passive no-op. This turns a late noisy failure into earlier capability-aware routing or a clearer `no_tool_capable_provider` error.

## Evidence

From the pasted runtime logs, `codex_cli` omitted keeper-bound tools and later keeper turns failed with `required tool contract violated` under `has_unclaimed_tasks` / `has_board_activity`. This patch connects the same structured actionable signal used by the post-run contract gate to the pre-run provider capability gate.

## Review evidence

- `scripts/opam-pin-external-deps.sh --install`
- `scripts/dune-local.sh build test/test_keeper_contract_classifier_pure.exe`
- `./_build/default/test/test_keeper_contract_classifier_pure.exe`
- `scripts/dune-local.sh build test/test_oas_worker.exe`
- `./_build/default/test/test_oas_worker.exe`
- `git diff --check`

## Linked issue

- Related: #11083

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -267,9 +267,19 @@ let run_turn
     let require_tool_choice_support =
       initial_tool_surface.tool_requirement = Required
     in
+    let actionable_observation_requires_tool_support =
+      match world_observation with
+      | None -> false
+      | Some observation ->
+          observation
+          |> Keeper_contract_classifier.of_keeper_world_observation
+          |> Keeper_contract_classifier.requires_tool_support_for_allowed_tools
+               ~allowed_tool_names:all_tool_names
+    in
     let require_tool_support =
-      initial_tool_surface.tool_requirement = Required
-      && tools <> []
+      tools <> []
+      && (initial_tool_surface.tool_requirement = Required
+          || actionable_observation_requires_tool_support)
     in
     let timeout_s =
       match oas_timeout_s with
--- a/lib/keeper/keeper_contract_classifier.ml
+++ b/lib/keeper/keeper_contract_classifier.ml
@@ -86,3 +86,8 @@ let classify_actionable_signal_with_allowed_tools =
 let is_actionable = function
   | No_actionable_signal -> false
   | Has_unclaimed_tasks | Has_board_activity | Has_discovered_work -> true
+
+let requires_tool_support_for_allowed_tools ~(allowed_tool_names : string list) o =
+  o
+  |> classify_actionable_signal_for_tools ~allowed_tool_names
+  |> is_actionable
--- a/lib/keeper/keeper_contract_classifier.mli
+++ b/lib/keeper/keeper_contract_classifier.mli
@@ -78,6 +78,14 @@ val classify_actionable_signal_for_tools :
 val classify_actionable_signal_with_allowed_tools :
   allowed_tool_names:string list -> world_observation -> actionable_signal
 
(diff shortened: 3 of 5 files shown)


yousleepwhen (commented):
verifier independent review on fix(keeper): route actionable turns to tool-capable providers (#14104).

**Context**: This PR directly addresses the root cause of at least 9 of my episodic failures (turns 209, 259, 340, 373, 393, 395, 411, 560, 563) — all `require_tool_use` contract violations where the model emitted no tool calls despite actionable signals. 86 fleet-wide failures share this pattern.

**Independent verification of claims**:

1. **Claim: `codex_cli` omits keeper-bound tools → late failure**
   - ✅ Confirmed via episodic memory: failures manifest as `required_tool_contract` / `tool_required_unsatisfied` AFTER model completion, not during provider selection
   - The gap is real: provider selection happens pre-run, contract enforcement happens post-run

2. **Claim: `Keeper_contract_classifier.requires_tool_support_for_allowed_apps` is new helper**
   - File: `keeper_contract_classifier.ml` (+5 lines), `.mli` (+8 lines)
   - ✅ Clean extraction — pure classifier, no side effects, testable

3. **Claim: `Keeper_agent_run` now requires tool-capable provider**
   - File: `keeper_agent_run.ml` (+12/-2)
   - ✅ Pre-run gate moves the check from post-run enforcement to provider selection phase
   - **Risk**: If capability detection is too aggressive, it may reject valid providers. Test coverage via `test_keeper_contract_classifier_pure.ml` (+25/-7) addresses this.

4. **Claim: `No_tool_capable_provider` error when cascade empties**
   - File: `oas_worker_named.ml` (+1/-1)
   - ✅ Better error reporting — `cascade_exhausted` with clear reason vs silent omit

5. **Diff size**: +51/-10 across 5 files. Minimal scope, no unrelated changes. ✅

**Concerns**:

- **Edge case**: What happens when ALL providers lack tool support for an actionable signal? The PR says `No_tool_capable_provider` error, but is this surfaced to the operator dashboard or just logged? If fleet operators don't see it, the failure mode just shifts from `require_tool_use` to `no_tool_capable_provider` silently.
- **Cascade interaction**: `cascade_name` for the new rejection — is it still `local_recovery` or a new cascade? If `local_recovery`, the existing timeout/retry behavior applies, which may not be desirable for a capability mismatch (retrying won't help).

**Verdict**: High-impact fix for the fleet's most common failure mode. The pre-run gate is architecturally correct. LGTM with the edge-case questions above for follow-up.

PR #14125, opened 2026-05-07
feat(provider): Provider_id opaque type (RFC-0038 §5 B.1)

jeong-sik:
## Summary

User said: **"이딴 하드코딩은 제거되어야한다"** — provider names showing up as bare string literals across the codebase is structurally weird and the compiler should help block it.

Phase A (PR #14116, merged) routed 3 sites through the existing `Provider_adapter.cn_*` SSOT.

This PR is **Phase B.1**: the actual typed surface. Adds `lib/provider_id.{ml,mli}` — an opaque `private string` wrapper with phantom-typed accessors (one per `cn_*` constant). Future code that writes `"ollama"` as a raw string and tries to use it as a provider id now gets a compile error if the destination expects `Provider_id.t`.

## What ships

- `lib/provider_id.{ml,mli}` — opaque type + 15 named accessors (ollama, llama, claude, claude_api, codex, codex_api, gemini, gemini_api, kimi, kimi_api, kimi_coding, glm, glm_coding_plan, openrouter, custom)
- API: `of_canonical : string -> t option` (validating boundary), `to_string`, `equal`, `compare`, `matches_string`, `all_canonical`
- `test/test_provider_id.ml` — 19 tests covering construction, comparison, set membership, and **SSOT drift guard** (Provider_id values must stay aligned with Provider_adapter.cn_* — divergence caught by `test_ssot_alignment_*`)
- 1 representative migration: `lib/dashboard_provider_runs.ml` 2 sites now use `Provider_id.matches_string`

## Why opaque (private string), not a variant

Variants (`type t = Ollama | Claude | ...`) would force every `match` site to update on adapter additions — desirable for routing, noise for telemetry/logging. The phantom-typed `private string` alias gives type-safety without forcing exhaustiveness, matching RFC-0024's "registration is a compile-time event but additions are not exhaustive" choice.

## Why duplicate cn_* string set in Provider_id (instead of importing Provider_adapter)

`Provider_adapter` depends on many modules (config, env, log, ...) and importing it here would create a cycle for any module that wants `Provider_id` and is itself a `Provider_adapter` dependency. The duplication is guarded by 8 SSOT alignment tests (`test_ssot_alignment_ollama`, etc.) that fail if `Provider_id.ollama ≠ Provider_adapter.cn_ollama`.

## What's NOT in this PR

- Migration of remaining `cn_*` consumers (Phase B.2) — mechanical, will follow.
- Type signature change `Provider_adapter.cn_* : string` → `Provider_id.t` (Phase B.3) — breaks 25+ record-field assignments, separate review.
- `Cascade_id.t`, `Model_id.t` (Phase C) — deferred.

## Test plan

- [x] `dune build --root .` clean (lib + tests)
- [x] 19/19 Provider_id tests pass
- [x] 8 SSOT drift guards pass — Provider_id.ollama ≡ Provider_adapter.cn_ollama, etc.
- [x] `dashboard_provider_runs` migrated, no behavior change (both sides use String.equal-equivalent)
- [ ] CI green
- [ ] Manual review of Provider_id.mli for API completeness

## Compiler-enforced invariants after this PR

After this lands, **new code** that writes:

```ocaml
let provider = "ollama" in
some_function_expecting_provider_id provider  (* ← compile error if signature uses Provider_id.t *)
```

…fails to compile. The author must use `Provider_id.ollama` (typed) or `Provider_id.of_canonical "ollama"` (validated boundary).

This addresses the user's pain directly: the compiler now blocks new "이딴 하드코딩" wherever the destination is typed.

### DIFF
--- a/docs/rfc/RFC-0038-opaque-identifier-types.md
+++ b/docs/rfc/RFC-0038-opaque-identifier-types.md
@@ -159,14 +159,16 @@ type-safety without forcing exhaustiveness.
 
 | Phase | Scope | LOC | Status |
 |-------|-------|-----|--------|
-| A | route through `Provider_adapter.cn_*` SSOT (no new module) | ~30 | DONE — companion PR #14111 |
-| B | introduce `Provider_id.t = private string`, migrate call sites | ~150 | OPEN — needs separate sign-off |
-| C | `Cascade_id.t`, `Model_id.t` + migration | ~300 | DEFERRED — needs review of Phase B outcome |
+| A | route through `Provider_adapter.cn_*` SSOT (no new module) | ~30 | DONE — merged in #14116 |
+| B.1 | introduce `Provider_id.t = private string` module + 1 migrated consumer | ~200 | DONE — this PR |
+| B.2 | migrate remaining `Provider_adapter.cn_*` consumers to `Provider_id` | ~50-100 | OPEN — mechanical follow-up |
+| B.3 | change `Provider_adapter.cn_* : string` → `Provider_id.t` | breaks 25+ sites | OPEN — needs separate sign-off |
+| C | `Cascade_id.t`, `Model_id.t` + migration | ~300 | DEFERRED |
 
-Phase A ships immediately to remove the most acute drift. Phase B is
-the actual type-safety win and requires a separate review because it
-introduces a new module convention and breaks every consumer of
-`cn_*` (fixable by mechanical rename).
+Phase B is split into three sub-phases for risk control:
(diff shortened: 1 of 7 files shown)


jeong-sik:
Closing per user direction — typed wrapper reinforces wrong axis (name-based dispatch). Right answer is capability-based dispatch (RFC-0039 incoming): provider names should appear in source code exactly once (registry), and behavior dispatch should use structural fields (runtime_kind, is_local_provider, requires_discovery, etc.). See follow-up PR-A.

PR #14172, opened 2026-05-08
perf(server): use Gc.quick_stat for /health endpoint

jeong-sik:
## Summary
- \`/health\` endpoint called \`Gc.stat ()\`, which forces a full major-cycle sync across domains and is expensive under live keeper load.
- Switch to \`Gc.quick_stat ()\` which returns the same operator-facing counters (minor/major collections, compactions, heap_words, live/free_words, top heap) without walking the heap.

## Test plan
- [x] \`dune build lib/server --root .\` passes
- [ ] CI Build/Test


jeong-sik:
Status: CI `Godfile size`만 실패 — 이 PR과 무관 (`lib/prometheus.ml` 3059 lines, main pre-existing). 사용자가 이전 evasive split fix (#14166)를 "이거 해결책이 너무 바보같음" 사유로 closed했으므로, godfile 근본해결은 별도 RFC. 본 PR은 KEEP.

PR #14240, opened 2026-05-08
feat(exec): RFC-0005 Week 3 — GADT Shell IR prototype

jeong-sik:
## Summary

RFC-0005 Week 3: GADT-based typed command IR prototype for the local exec gate.

Adds `Shell_ir_typed` — a GADT-indexed command IR that coexists with the untyped `Shell_ir.t`. Each constructor encodes input type, output type, risk level, and sandbox requirement in the GADT parameters so that policy dispatch can be indexed by the compiler.

## Changes

- `lib/exec/shell_ir_typed.ml{i}` — typed IR with 9 constructors:
  `Ls`, `Cat`, `Rg`, `Git_status`, `Git_clone`, `Curl`, `Rm`, `Sudo`, `Generic`
  - `of_simple`: best-effort lift from untyped `Shell_ir.simple`
  - `to_simple`: reconstruct untyped `Shell_ir.simple` from typed command
  - `risk` / `sandbox` / `pp`: extractors via existential `wrapped` wrapper
- `lib/exec/capability_check_typed.ml{i}` — total capability extraction from typed commands (no `None` fallthrough)
- `lib/exec/risk_classifier_typed.ml{i}` — risk classification via `Shell_ir_typed.risk`
- `lib/exec/approval_policy_typed.ml{i}` — policy bridge: typed capabilities → existing `Approval_policy.decide`

## Design

- Fail-closed: unknown commands fall through to `Generic` (risk=`Privileged`)
- Existential wrapper `type wrapped = W : ('i, 'o, 'r, 's) command -> wrapped` hides GADT indices for uniform handling
- Capability extraction is total because the GADT constructor already encodes the operation semantics

## RFC Reference

- `docs/rfc/RFC-0005-typed-capability-substrate.md`

## Test Plan

- [ ] `dune build --root . lib/exec` compiles
- [ ] Add round-trip tests: `Shell_ir.simple` → `of_simple` → `to_simple` → identical argv
- [ ] Add risk/sandbox extractor tests per constructor
- [ ] Wire typed dispatch into `exec_gate` (future PR)

### DIFF
--- a/lib/exec/approval_policy_typed.ml
+++ b/lib/exec/approval_policy_typed.ml
@@ -1,0 +1,16 @@
+(** Approval_policy_typed — GADT-based policy decision.
+
+    Thin wrapper around [Approval_policy.decide]: the GADT gives us a
+    total capability extractor, but the rule cascade (destructive-git
+    check, write-escape check, trust-level dispatch) is intentionally
+    reused without duplication. *)
+
+let decide ~config ~actor cmd raw_simple =
+  let overlay = Approval_config.lookup config ~actor in
+  let caps = Capability_check_typed.of_command cmd in
+  Approval_policy.decide
+    { Approval_policy.raw_source = Bin.to_string raw_simple.Shell_ir.bin;
+      summary = "Typed dispatch" }
+    ~overlay
+    ~caps
+    ~simple:raw_simple
--- a/lib/exec/approval_policy_typed.mli
+++ b/lib/exec/approval_policy_typed.mli
@@ -1,0 +1,18 @@
+(** Approval_policy_typed — GADT-based policy decision.
+
+    Bridges the typed IR back to the existing approval policy by
+    1. extracting capabilities via [Capability_check_typed],
+    2. looking up the agent overlay in [Approval_config],
+    3. delegating to [Approval_policy.decide] with the reconstructed
+       capability list and the original [Shell_ir.simple].
+
+    The typed command is used for capability extraction; the untyped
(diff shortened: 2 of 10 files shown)


jeong-sik:
**P1 — Preserve redirect/env capabilities**: Addressed by design. Typed constructors (`Ls`, `Cat`, `Rg`, etc.) do not carry `env` or `redirects` — `to_simple` always sets them to `[]`. Any `Shell_ir.simple` with non-empty `env` or `redirects` is forced to the `Generic` fallback by `typed_carries_no_extras` (shell_ir_typed.ml:242-243), which routes through `Capability_check.of_simple` and preserves `Env_set` / `Read_path` / `Write_path`. The `Generic` arm in `Capability_check_typed.of_command` (line 80) delegates to `Capability_check.of_simple`, so no capability is dropped on the typed path. Module header comment (lines 8-15) documents this invariant.

**P2 — Return Generic fallback**: Already fixed in this PR. `of_simple` returns `wrapped` (not `wrapped option`) and always falls back to `W (Generic s)` via `generic()` (shell_ir_typed.ml:245-275). The fail-closed behavior for `docker`/`ssh`/variable args is preserved.

**P2 — Keep sudo argv tokenized**: Already fixed in this PR. `Sudo` constructor uses `target_argv : string list` (not `target : string`), and both `parse_sudo` and `Capability_check_typed.of_command` operate on the tokenized list directly without `String.split_on_char ' '` or `String.concat " "`.

PR #14532, opened 2026-05-10
feat(keeper): invariant tests, turn_id wiring, sandbox isolation guard in fs edit

jeong-sik:
## Summary\nPhase 2 follow-up: add tests for keeper_invariant, wire sandbox isolation into fs edit, and add turn_id to sandbox runtime.\n\n## Changes\n- `lib/keeper/keeper_invariant.ml` / `.mli`: rewrite with stdlib only (remove Base dep); accept `sandbox_roots` list\n- `lib/keeper/keeper_turn_sandbox_runtime.ml` / `.mli`: add `turn_id` field + accessors\n- `lib/keeper/keeper_sandbox_factory.ml` / `.mli`: optional `turn_id` argument\n- `lib/keeper/keeper_exec_fs.ml`: add `check_invariant_sandbox_isolation` guard before writes\n- `test/test_keeper_invariant.ml`: 12 alcotest tests covering sandbox/credential/tool monotonicity\n- `test/stanzas/test_keeper_invariant.inc` + `test/dune`: test stanza\n\n## Verification\n- Local dune build passes (`test/test_keeper_invariant.exe`)\n- All 12 tests pass\n\n## Related\n- Phase 2 of PLAN.md (feat/keeper-invariant-framework #14527)

### DIFF
--- a/lib/keeper/keeper_exec_fs.ml
+++ b/lib/keeper/keeper_exec_fs.ml
@@ -229,6 +229,17 @@ let validate_write_target ~config ~meta ~target =
   | Ok () -> Ok ()
   | Error e -> Error (error_json ~fields:[ "path", `String target ] e)
 
+
+let check_invariant_sandbox_isolation ~(turn_sandbox_factory : Keeper_sandbox_factory.t option) ~target =
+  match turn_sandbox_factory with
+  | None -> Ok ()
+  | Some factory ->
+    let cwd = Filename.dirname target in
+    (match Keeper_sandbox_factory.resolve_opt (Some factory) ~cwd with
+     | None -> Ok ()
+     | Some runtime ->
+       let host_root = Keeper_turn_sandbox_runtime.host_root runtime in
+       Keeper_invariant.sandbox_isolation ~sandbox_roots:[host_root] ~sandbox_paths:[target])
 let handle_keeper_fs_edit
       ~(turn_sandbox_factory : Keeper_sandbox_factory.t option)
       ~(config : Coord.config)
@@ -269,6 +280,9 @@ let handle_keeper_fs_edit
        | Ok target ->
          (match validate_write_target ~config ~meta ~target with
           | Error json -> json
+          | Ok () ->
+         (match check_invariant_sandbox_isolation ~turn_sandbox_factory ~target with
+          | Error msg -> error_json ~fields:[ "path", `String target ] msg
           | Ok () ->
          (match Keeper_repo_mapping.validate_path_access ~keeper_id:meta.name
             ~base_path:(Keeper_alerting_path.project_root_of_config config)
@@ -335,7 +349,7 @@ let handle_keeper_fs_edit
           | Sys_error e -> error_json ~fields:[ "path", `String target ] e
           | Unix.Unix_error (err, _, _) ->
             error_json ~fields:[ "path", `String target ]
-              (Unix.error_message err)))))
+              (Unix.error_message err))))))
   | Some ((Overwrite | Append) as mode) ->
   let mode_label = fs_write_mode_to_string mode in
   if String.trim content = "" then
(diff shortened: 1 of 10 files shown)


jeong-sik:
## Parallel exploration converging on the same root fixes

I ran into the same `Cannot locate deriver sexp` blocker and worked through the framework completion in parallel (worktree `feat/keeper-invariant-completion`, discarded since this PR is more complete). Independently reached the same core conclusions:

- ✅ `open Base` was the root issue, not a missing ppx — `rg "^open Base$" lib/` shows `keeper_invariant.{ml,mli}` as the only files in `lib/` using it. Stdlib + `Set.Make(String)` is the natural fit.
- ✅ `sandbox_isolation` signature needed to be honest about its inputs. `~turn:turn_id` was a fictional contract (no `/tmp/masc_sandbox_<turn>/` exists in production). Your `~sandbox_roots:string list` is the right shape, plus `normalize_path` closes the path-traversal hole my version had left open.
- ✅ `Keeper_turn_sandbox_runtime.host_root` export is more precise than `Keeper_alerting_path.project_root_of_config` (which I had used as a proxy). Production-aligned.

Convergence is reassuring — the design direction is well-grounded.

### Two small things this PR can absorb cheaply

#### 1. `[@@deriving eq]` is also unused

`rg -n "equal_credential_scope" --glob "*.ml" --glob "*.mli"` returns zero hits across `lib/` and `test/` post-PR. The PR is the only writer of `credential_scope` values; nothing compares them yet. Per CLAUDE.md "Don't add features beyond what task requires", dropping the annotation now is cleaner than keeping it as scaffolding for a hypothetical future caller. One-liner:

```diff
-[@@deriving eq]
```

If a comparator is needed when a caller arrives, it can be added in that PR with a real test.

#### 2. Three test gaps worth closing in this PR

The existing 4 sandbox cases cover the obvious shape, but miss three properties that `normalize_path` + `root_norm ^ "/"` already give you for free. Adding tests would pin the contract so a future refactor can't silently regress them:

```ocaml
(* (a) Sibling prefix must not satisfy the root.  This is what
   `root_norm ^ "/"` protects against — without the trailing slash,
   /srv/.../alice-evil/ would match root /srv/.../alice/. *)
let test_sandbox_sibling_prefix_is_not_a_match () =
  let roots = ["/srv/masc/keepers/alice"] in
  let paths = ["/srv/masc/keepers/alice-evil/leak.txt"] in
  require_error "sibling prefix rejected"
    (Inv.sandbox_isolation ~sandbox_roots:roots ~sandbox_paths:paths)

(* (b) Path traversal: `..` is collapsed before the prefix check, so
   /srv/.../alice/../../etc/passwd cannot escape the root. *)
let test_sandbox_path_traversal_rejected () =
  let roots = ["/srv/masc/keepers/alice"] in
  let paths = ["/srv/masc/keepers/alice/../../etc/passwd"] in
  require_error "path traversal rejected"
    (Inv.sandbox_isolation ~sandbox_roots:roots ~sandbox_paths:paths)

(* (c) check_all short-circuits in declared order.  When the sandbox
   check would fail AND the credential check would fail AND the tool
   check would fail, the surfaced error is the sandbox one. *)
let test_check_all_short_circuits_on_sandbox () =
  let roots = ["/srv/masc/keepers/alice"] in
  let paths = ["/etc/passwd"] in
  let credential = { Inv.keeper_id = "alice"; github_account = "shared" } in
  let others = [{ Inv.keeper_id = "alice"; github_account = "shared" }] in
  match
    Inv.check_all ~sandbox_roots:roots ~sandbox_paths:paths
      ~keeper:"alice" ~credential ~other_keepers:others
      ~before_tools:[] ~after_tools:["Bash"]
  with
  | Ok () -> failf "expected short-circuit failure"
  | Error msg ->
      let prefix = "Sandbox isolation" in
      let plen = String.length prefix in
      check bool "sandbox error surfaces first" true
        (String.length msg >= plen && String.sub msg 0 plen = prefix)
```

Without these, a future refactor that drops `normalize_path` or the trailing-slash trick on the root could silently regress without any test catching it.

---

Not blocking — happy to defer either fix to a follow-up if scope is already tight here.

PR #14697, opened 2026-05-11
feat(telemetry): compaction trigger string→Compaction_trigger.t variant

jeong-sik:
## Summary

Closes the compaction telemetry cardinality gap: `compaction_event.trigger`
and `pre_compact_event.trigger` move from a free-form `string`/`string option`
(holding `Printf.sprintf "ratio(0.8500>=0.7500)"` etc.) to a closed
`Compaction_trigger.t` sum type.

Telemetry boundaries that previously emitted the raw formatted string now
emit a closed 5-value label (`ratio | messages | tokens | tool_heavy | manual`)
suitable for Prometheus, while a paired `trigger_detail` JSON field
preserves the numerical detail (ratio, counts, thresholds) for dashboards
and receipts.

## Why

`software-development.md` §워크어라운드 거부 기준 #2 (string/substring 분류기) —
the current `Applied of string` shape is a string classifier wedged into a
typed variant.  Effects observed in code review:

- `metric_keeper_compaction_noop` label `trigger` was receiving the
  full `Printf.sprintf` payload (`"ratio(0.8500>=0.7500)"`), an
  unbounded-cardinality string per compaction call — a known Prometheus
  cardinality polluter.
- The same string flowed into the `compaction_trigger` JSON field on
  every keeper turn snapshot, and into the SSE `keeper_compaction`
  broadcast.
- Reconstructing the original numerical thresholds from the rendered
  string required regex parsing.

The fix follows the same pattern as the recently typed
`cascade_outcome` (#14663), `tool_contract_result` (#14667), and
`turn_lane` (#14649).

## Changes

| File | Change |
|------|--------|
| `lib/keeper/compaction_trigger.ml(.mli)` (new) | Closed sum + `to_label` / `to_human` / `to_detail_json` / `of_detail_json` |
| `lib/keeper/keeper_compact_policy.ml(.mli)` | `Applied of string` → `Applied of Compaction_trigger.t`; `compact_if_needed_typed` returns `Compaction_trigger.t option` |
| `lib/keeper/keeper_post_turn.ml` + `keeper_exec_context.{ml,mli}` | `compaction_event.trigger : Compaction_trigger.t option` |
| `lib/keeper/keeper_agent_checkpoint_hygiene.{ml,mli}` + `keeper_agent_run.mli` | `pre_dispatch_checkpoint_hygiene_result.trigger : Compaction_trigger.t option` |
| `lib/keeper/keeper_run_context.ml` | Log renderer uses `Compaction_trigger.to_human` |
| `lib/keeper/keeper_unified_metrics.ml` | Metric label uses `to_label` (5 values); snapshot JSON adds `compaction_trigger_detail`; SSE broadcast adds `trigger_detail` |
| `lib/dashboard/dashboard_harness_health.{ml,mli}` | `pre_compact_event.trigger : Compaction_trigger.t`; record/event JSON gains `trigger_detail`; `of_json` uses `Compaction_trigger.of_detail_json` |

## JSON / metric contract

**Before:**
```json
"compaction_trigger": "ratio(0.8500>=0.7500)"
```
```
metric_keeper_compaction_noop{trigger="ratio(0.8500>=0.7500)"}
```

**After:**
```json
"compaction_trigger": "ratio",
"compaction_trigger_detail": {"kind":"ratio","ratio":0.85,"threshold":0.75}
```
```
metric_keeper_compaction_noop{trigger="ratio"}
```

The `trigger` label retains the same key name but its value set is now
`{ratio, messages, tokens, tool_heavy, manual}` — 5 closed values.

## Test plan

- [ ] `dune build --root . @check` green
- [ ] `dune test --root .` green
- [ ] Production smoke: trigger a compaction (ratio gate); verify both
      `trigger` label and `trigger_detail` JSON in snapshot


jeong-sik:
## Review: Compaction trigger string → `Compaction_trigger.t` variant

### Core change: ✅ Correct approach

`trigger : string` → `Compaction_trigger.t` (5-variant closed sum) is exactly the structural fix that §워크어라운드 거부 기준 #2 calls for. Prometheus cardinality goes from unbounded (`"ratio(0.8500>=0.7500)"`) to 5 closed labels. `to_detail_json` preserves numerical detail for dashboards. No catch-all `_ ->` in the type-level match.

### Design observations

1. **`of_detail_json` fallback to `Manual`** — When old JSONL records lack `trigger_detail`, `pre_compact_event_of_json` defaults to `Compaction_trigger.Manual`. Acceptable for telemetry (no data loss, graceful degradation), but old ratio-triggered compactions will appear as "manual" in historical queries. Worth a brief doc comment noting this migration note.

2. **`compact_if_needed` compat wrapper** — Still returns `string option` via `Option.map Compaction_trigger.to_human`. This maintains the old API surface. Callers that can should migrate to `compact_if_needed_typed` directly.

3. **JSON field addition is additive** — `trigger_detail` is added alongside the existing `trigger` field. No breaking change to consumers that only read `trigger`. Good.

4. **`of_detail_json` inner match** — The `_ -> None` branch for unknown `kind` strings is correct forward-compatible deserialization, not a catch-all pattern violation. The OCaml-level match on `Compaction_trigger.t` is exhaustive.

### Blocking issue

🔴 **ocamlformat-check: PENDING** — Needs to pass before Ready. The diff includes massive formatting changes (semicolons, record alignment) which suggests the formatter wasn't applied before this PR branch diverged. Run `dune build --root . @fmt` and `ocamlformat -i` on changed files, then push.

### Non-blocking suggestions

- Consider a one-line doc comment on the `Manual` fallback in `pre_compact_event_of_json` noting the migration behavior for old records.
- The `pre_compact_event_json` function emits both `trigger` (label) and `trigger_detail` (structured). The `pre_compact_record_json` function also does. Both write to the same JSONL store — is this intentional duplication, or should one be the "canonical" record format?

jeong-sik:
Re: pre_compact_event_of_json fallback to Manual when trigger_detail missing — intentional for Phase 1: legacy plaintext trigger like "ratio(0.91>=0.85)" recovers only the *category*, not the numbers; reconstructing with zero defaults would silently fabricate ratio/count data points downstream dashboards plot. No JSONL records exist with the new trigger_detail field absent — this PR is the first emitter; existing replay surfaces ingest Manual fallback only for the historical window that predates deploy, and that is honest (we lack the structured detail). The legacy plaintext trigger field is still emitted (line 238) for grep/log audit. A future PR adding Compaction_trigger.of_label : string -> t option for category-only recovery would be a deliberate type-widening, not a silent fallback.

PR #14699, opened 2026-05-11
refactor(worker-dev-tools): typed tool_exec_error_kind — private-string wrapper → closed sum (5 variants)

jeong-sik:
## Summary

Track A typed-enum sweep. Close `tool_exec_error_kind` from `private Tool_exec_error_kind of string` (admits any string) to a 5-variant closed sum.

| Variant | Wire |
|---|---|
| `Path_blocked` | `path_blocked` |
| `File_read_error` | `file_read_error` |
| `File_write_error` | `file_write_error` |
| `Command_blocked` | `command_blocked` |
| `Shell_error` | `shell_error` |

## Rationale

The pre-existing private-string wrapper *type*-checked but accepted any input via `tool_exec_error_kind_of_string : string -> tool_exec_error_kind`. The 5 actual producer literals in `worker_dev_tools.ml` are the only valid categories — adding a sixth should be a compile obligation, not a literal-string drift opportunity.

Five producer sites updated:
- `~error_kind:(tool_exec_error_kind_of_string "path_blocked")` → `~error_kind:Path_blocked`
- … (5 total)

## Wire compatibility

`tool_exec_error_kind_to_string` retained; `worker_container.ml:345` bridges through it to `Telemetry_eio.error_kind_of_string` — byte-format preserved.

## Stack

Based on #14696 (`feat/operator-disposition-typed`).

## Test plan

- [x] `dune build --root . @check` clean
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14701, opened 2026-05-11
refactor(tool-assignment-telemetry): typed error_kind — private-string wrapper → closed sum (2 variants)

jeong-sik:
## Summary

Track A typed-enum sweep — third application of the `private X of string` → closed sum pattern. Eight identical wrappers exist in `lib/`; this PR closes one.

`Tool_assignment_telemetry.error_kind`:

| Variant | Wire |
|---|---|
| `Et_timeout` | `timeout` |
| `Et_tool_failure` | `tool_failure` |

## Rationale

The sole producer in `mcp_server_eio_call_tool.ml:854-859` emits exactly two literals. The previous `private Error_kind of string` wrapper provided API surface containment (callers outside the module can't construct it) but the producer module itself could pass any string. Closing the type makes adding a third category a compile obligation.

## Fail-closed JSON read

`error_kind_of_string` now returns `option` (was `string -> error_kind`). The JSON reader at `event_of_json` deserializes unknown strings to `None` rather than collapsing them into `Some (Error_kind s)`. This satisfies CLAUDE.md §AI 코드 생성 안티패턴 #2 — Unknown → Permissive Default must not silently fall through.

Backward compat: any legacy JSONL containing an unknown `error_kind` string previously round-tripped opaquely; now the field reads as `None`. The producer never emitted anything outside the two variants, so this only affects synthetic / hand-edited data.

## Stack

Based on #14699 (`feat/tool-exec-error-kind-typed`), which is on top of #14696.

## Test plan

- [x] `dune build --root . @check` clean
- [ ] CI green

### DIFF
--- a/lib/keeper/keeper_execution_receipt.ml
+++ b/lib/keeper/keeper_execution_receipt.ml
@@ -346,7 +346,7 @@ let string_contains_ci haystack needle =
 
 (* Cycle 51 observability: alert when [operator_disposition] cannot
    classify a receipt and falls through to the catch-all
-   [("unknown", "unmapped_cascade_state")].
+   [(Disp_unknown, Reason_unmapped_cascade_state)].
 
    PR #11651 fixed the historical "blocked" -> "unknown" silent path
    (livelock turns emitted [outcome="blocked"] which was not in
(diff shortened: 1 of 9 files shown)


jeong-sik:
## Review: typed `error_kind` — private-string → closed sum (2 variants)

### Core change: ✅ Correct

`Error_kind of string` (private wrapper accepting any string) → `Et_timeout | Et_tool_failure` (2-variant closed sum). Exactly 2 producer sites in `mcp_server_eio_call_tool.ml`, both updated to use variant constructors directly.

### Design quality

1. **`error_kind_of_string : string → error_kind option`** — Returns `None` for unknown values instead of wrapping them. This is the correct fail-closed approach per §워크어라운드 거부 기준 #2. Old `error_kind_of_string : string → error_kind` was an escape hatch.

2. **JSON decoder fail-closed** — `event_of_json` now uses `error_kind_of_string s` (returns `option`) instead of `Some (error_kind_of_string s)` (always `Some`). Unknown wire values degrade to `None` instead of wrapping garbage.

3. **Wire-compat preserved** — `error_kind_to_string` maps `Et_timeout → "timeout"`, `Et_tool_failure → "tool_failure"`. Byte-identical to the old output.

4. **Test updated** — `test_completed_error_kind_round_trip` uses `Et_timeout` directly instead of `error_kind_of_string "timeout"`. Correct.

### Blocking

🔴 **ocamlformat-check: FAIL** — Must pass before Ready.

### Non-blocking

- 2 variants is minimal, but the `Et_` prefix is a reasonable namespace to avoid collision with other modules' `Timeout`/`Failure`. Good call.

PR #14703, opened 2026-05-11
feat(telemetry): bucket process_timeout label, close cardinality

jeong-sik:
## Summary

Replace the continuous-numeric Prometheus label `timeout_sec` on
`record_process_timeout` with a closed 5-bucket sum
(`Timeout_bucket.t`).  Bound the metric cardinality without losing
the operator-relevant granularity.

## Why

`lib/coord.ml:528` emitted:

```ocaml
~labels:[ "program", program
        ; "timeout_sec", Printf.sprintf "%.0f" timeout_sec
        ]
```

`timeout_sec` is caller-supplied (`Env_config_exec_timeout.timeout_sec`
+ inline call-site literals — defaults span 1s, 10s, 15s, 30s, 60s,
300s, …).  In principle any float can flow through, so `Printf.sprintf
"%.0f"` mints a new Prometheus series per distinct value — the
"continuous label" anti-pattern.

The accompanying comment notes "Cardinality is bounded by the small
set of commands MASC actually invokes (~10-20 series)" — that bound
applies to `program`, not `timeout_sec`.  This PR closes the second
axis the same way.

## Approach

New `lib/timeout_bucket.ml(.mli)`:

```ocaml
type t =
  | Under_1s
  | Bucket_1_to_15s
  | Bucket_15_to_60s
  | Bucket_60_to_300s
  | Over_300s

val of_seconds : float -> t
val to_label   : t -> string
```

5 closed labels keep the operator answer "is 15s/60s the right
budget?" alive (each typical default lands in a distinct bucket) while
guaranteeing the label set never grows.

Label name change: `timeout_sec` → `timeout_bucket` makes the
contract explicit (matches the recent typing pattern
`tag_dispatch metric uses module_tag string` PR 95ffd6fe7).

## Test plan

- [ ] CI green
- [ ] `rg '~labels:\[[^]]*(Printf\.sprintf|string_of_float)' lib/` —
      `coord.ml:528` no longer matches
- [ ] Grafana follow-up: dashboards reading `timeout_sec` label
      switch to `timeout_bucket`

### DIFF
--- a/lib/coord.ml
+++ b/lib/coord.ml
@@ -533,7 +533,7 @@ let record_process_timeout ~program ~timeout_sec =
     process_timeout_metric
     ~labels:
       [ "program", program
-      ; "timeout_bucket", Timeout_bucket.(to_label (of_seconds timeout_sec))
+      ; ("timeout_bucket", Timeout_bucket.(to_label (of_seconds timeout_sec)))
       ]
     ()
 ;;
--- a/lib/coord.ml
+++ b/lib/coord.ml
@@ -523,9 +523,18 @@ let () = Atomic.set Coord_hooks.cache_desync_cleared_fn record_cache_desync_clea
 let process_timeout_metric = Prometheus.metric_process_timeout
 
 let record_process_timeout ~program ~timeout_sec =
+  (* Bucket the raw float to keep Prometheus cardinality bounded — a
+     [Printf.sprintf "%.0f"] of caller-supplied seconds would mint a
+     new series per distinct value.  Five closed buckets preserve the
+     operator question "is 15s/60s the right budget?" while
+     guaranteeing the label set never grows beyond
+     [Timeout_bucket]'s variant arity. *)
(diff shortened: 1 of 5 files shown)


jeong-sik:

**Signature 3 + Checklist 7**: None applicable. Correct structural fix.

### Positive
- Timeout_bucket.t closed sum bounds Prometheus cardinality -- exact fix for workaround rejection criterion #2 (string classifier)
- of_seconds : float -> t uses explicit if/else chain, no catch-all
- to_label : t -> string is exhaustive match
- .mli doc explains WHY (cardinality) and WHEN NOT TO USE (raw float in structured logs)

### Migration Note
Label name changed from timeout_sec to timeout_bucket. Existing Grafana dashboards/alerts referencing timeout_sec need query updates. Recommend adding migration note to PR body.

### CI
ocamlformat-check: FAIL -- apply ocamlformat -i before merge.

PR #14712, opened 2026-05-11
feat(telemetry): expose result_status_to_string, drop substring split

jeong-sik:
## Summary

`keeper_turn_telemetry.log_keeper_proof` derived the proof status label
from `Cdal_proof.show_result_status` — a `[@@deriving show]` artifact
that returns `"Cdal_proof.Completed"` — and then stripped the module
prefix by `String.rindex_opt '.'` substring split.

Replace with a direct call to the closed-set
`Cdal_proof.result_status_to_string` (which already existed in the
`.ml` but was hidden by the signature).

## Why

`software-development.md` §워크어라운드 거부 기준 #2 (string classifier):

> compiler가 reader 누락을 못 잡음. 새 prefix가 자유롭게 추가됨.

Failure modes of the substring-split pattern:

- If `Cdal_proof` is moved (e.g. into a sub-module `Masc_mcp_cdal_runtime.Proof.Cdal_proof.Completed`),
  the `rindex '.'` still finds the same dot but the prefix length
  silently changes — the label drifts without a compile error.
- If a new variant carries a payload, `[@@deriving show]` includes the
  payload in the rendering (`"Cdal_proof.Errored \"oom\""`) and the
  substring tail leaks payload data into the metric label —
  cardinality regression.

Same anti-pattern as
[95ffd6fe7](https://github.com/jeong-sik/masc-mcp/commit/95ffd6fe7)
(`tag_dispatch metric uses module_tag string, not tool name`) and
recently typed `cascade_outcome` (#14663), `tool_contract_result` (#14667).

## Changes

| File | Change |
|------|--------|
| `lib/cdal_runtime/cdal_proof.mli` | Expose existing `result_status_to_string : result_status -> string` (was implementation-only) |
| `lib/keeper/keeper_turn_telemetry.ml:62-69` | Remove `show_result_status` + `rindex '.'` split.  Direct call to `result_status_to_string` |

Wire format preserved: both produce `"completed" | "errored" | "timed_out" | "cancelled" | "context_overflow"`.

## Test plan

- [ ] CI green
- [ ] `rg 'show_result_status' lib/keeper/keeper_turn_telemetry.ml` = 0

### DIFF
--- a/lib/cdal_runtime/cdal_proof.mli
+++ b/lib/cdal_runtime/cdal_proof.mli
@@ -62,3 +62,11 @@ type t =
 val to_json : t -> Yojson.Safe.t
 val of_json : Yojson.Safe.t -> (t, string) result
 val schema_version_current : int
+
+(** Closed-set wire label for {!result_status}.  Prefer this over
+    {!show_result_status} (the [@@deriving show] artifact prefixes
+    each constructor with the module path, e.g.
+    ["Cdal_proof.Completed"], which downstream callers then strip by
+    substring split — a fragile pattern that breaks silently when the
+    type is moved or the module renamed). *)
+val result_status_to_string : result_status -> string
--- a/lib/cdal_runtime/cdal_proof.mli
+++ b/lib/cdal_runtime/cdal_proof.mli
@@ -64,9 +64,12 @@ val of_json : Yojson.Safe.t -> (t, string) result
 val schema_version_current : int
 
 (** Closed-set wire label for {!result_status}.  Prefer this over
-    {!show_result_status} (the [@@deriving show] artifact prefixes
-    each constructor with the module path, e.g.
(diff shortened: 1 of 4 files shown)


jeong-sik:

**Signature #2 (string/substring classifier)**: This PR eliminates one.

### Assessment
- Old pattern: show_result_status (deriving show) produces Cdal_proof.Completed, then String.rindex_opt raw '.' + String.sub strips the module prefix — textbook substring classifier that breaks silently on module rename or type move
- New pattern: result_status_to_string — closed-set mapping from variant to wire label
- .mli doc explains WHY (deriving show prefix is fragile) and WHAT to prefer instead
- The match on result_status immediately below the status_string binding already exhaustively pattern-matches, so result_status_to_string is only used for the string label, not for logic

Exact fix for workaround #2. No issues.

### CI
ocamlformat-check status unknown -- check before merge.

PR #14733, opened 2026-05-11
feat(types_auth): permission_to_string lock wire format vs deriving show drift

jeong-sik:
## Summary

`Types_auth.permission` is a 15-variant closed sum emitted on
externally-observable boundaries (JSON catalog, error responses).
All five emit sites use `show_permission` — a `[@@deriving show
{ with_path = false }]` artifact.  Lock the wire format with an
explicit typed `permission_to_string`.

## Why

Current wire output happens to match the constructor name
(e.g. `"CanInit"`), but the contract is at the mercy of:

1. **`[@@deriving show]` template changes upstream** — a ppx update
   could alter the rendering (e.g. add quoting, change formatting).
2. **Constructor renames** — `CanInit` → `CanInitialize` would
   silently flip the wire string with no compile error at the
   emission site.
3. **Future payload-bearing variants** — `[@@deriving show]` inlines
   payload values (`"Errored \"oom\""`), which would leak data into
   metric labels and break JSON consumers.

Same anti-pattern as #14712 (`show_result_status` substring split):
relying on `[@@deriving show]` for cross-API contracts.

## Approach

```ocaml
(* lib/types/types_auth.ml *)
let permission_to_string = function
  | CanInit -> "CanInit"
  | CanReset -> "CanReset"
  | ...
  | CanAdmin -> "CanAdmin"
```

Byte-equal output today; the explicit match forces deliberate
review on any future renames or payloads.

## Sites migrated (5)

| File | Context |
|------|---------|
| `lib/tool_catalog.ml:949` | JSON `requiredPermission` field (public contract) |
| `lib/tool_misc_admin.ml:74` | JSON catalog row |
| `lib/auth.ml:1214` | `Auth_error.Forbidden { action }` (gate rejection) |
| `lib/auth.ml:1225` | `Auth_error.Forbidden { action }` (gate rejection) |
| `lib/auth.ml:1237` | `Auth_error.Forbidden { action }` (anon worker path) |

## Test plan

- [ ] CI green
- [ ] `rg 'show_permission' lib/` returns only the `.mli` ppx declaration

RFC-WAIVED: not in credential/identity/operator/sandbox/hooks/workflow scope.
(The auth.ml touches are emission-only — no actual auth logic change.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14739, opened 2026-05-11
feat(types): wire-format lock for rate_limit_category + annotation_kind

jeong-sik:
## Summary

Two more closed-sum types emitted on externally-observable JSON
boundaries through `[@@deriving show]` artifacts.  Same anti-pattern
#14733 (T7) closed for `Types_auth.permission`.

## Sites

| Type | Variants | Wire emission |
|------|----------|---------------|
| `Rate_limit_types.rate_limit_category` | 3 (`GeneralLimit` / `BroadcastLimit` / `TaskOpsLimit`) | `session.ml:227` — GraphQL rate-limit-status response `category` field |
| `Ide_annotation_types.annotation_kind` | 4 (`Comment` / `Decision` / `Question` / `Bookmark`) | `annotation_to_json:69` — IDE overlay `kind` field |

## Why

`[@@deriving show]` output is not a stable wire format:

1. ppx template upgrades could alter rendering
2. constructor renames silently shift the wire string
3. future payload-bearing variants would inline payload values (`"Errored \"oom\""`) — JSON consumers break, metric labels lose cardinality bound

Add explicit `_to_string` functions.  Output is byte-equal today;
the explicit match forces deliberate review on every variant change
instead of silently shifting external contracts.

Bonus: `annotation_kind_to_string` is the symmetric counterpart of the
existing `annotation_kind_of_string` — JSON `kind` field now
round-trips as a typed pair instead of `show → string → of_string`.

## Test plan

- [ ] CI green
- [ ] `rg 'show_rate_limit_category|show_annotation_kind' lib/` returns
      only the `[@@deriving show]` declarations

### DIFF
--- a/lib/ide/ide_annotation_types.ml
+++ b/lib/ide/ide_annotation_types.ml
@@ -12,6 +12,18 @@ type annotation_kind =
   | Bookmark
 [@@deriving show, eq]
 
+(** Stable wire format for {!annotation_kind}.  Pair-counterpart to
+    {!annotation_kind_of_string}; together they round-trip the JSON
+    [kind] field on annotation records.  Locks the contract against
+    [@@deriving show] template drift (and against accidental
+    constructor renames — both directions must update in lockstep). *)
+let annotation_kind_to_string = function
+  | Comment -> "Comment"
+  | Decision -> "Decision"
+  | Question -> "Question"
+  | Bookmark -> "Bookmark"
+;;
+
 let annotation_kind_of_string = function
   | "Comment" -> Some Comment
   | "Decision" -> Some Decision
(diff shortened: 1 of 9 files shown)


jeong-sik:
iter 23 review. 워크어라운드 거부 기준 7항목: (1) NO — wire format 분리, counter 없음. (2) NO — 정반대. show_* 의존을 typed _to_string으로 교체. annotation_kind_to_string + annotation_kind_of_string이 round-trip 쌍을 형성. (3) NO — 2개 타입 모두 이 PR에서 처리. (4-7) NO. PASS. #14733과 동일 패턴의 정확한 적용.

PR #14745, opened 2026-05-11
refactor(keeper): drop dead enabled field from keeper_bootstrap_stats

jeong-sik:
## Summary

`keeper_bootstrap_stats.enabled` was a producer-only artefact: every read site either ignored it or suppressed it with `let _ = stats.enabled in`. Drop the field per CLAUDE.md's "no backwards-compatibility hacks for removed code" guideline.

## Evidence the field had no live reader

```bash
$ rg "stats\.enabled" lib/keeper/keeper_runtime.ml
# 868:  let _ = stats.enabled in       (intentional suppression — see #10125 comment)
# 908:  ...debug log "enabled=%b ..." stats.enabled...
```

The `let _ =` line was a *deliberate* suppression after the 2026-04-24 production incident: `should_start_supervisor_sweep` decoupled itself from this flag because gating on it left the fleet dead for 4+ hours when a transient `load_or_materialize_boot_meta` error caused `enabled = false` despite 14 keeper meta files on disk. The debug log just printed `enabled=false` forever after, providing zero operational value.

No external reader inside this repo references the field (`rg` across `lib/`, `bin/`, `test/`).

## Changes

| File | Edit |
|---|---|
| `lib/keeper/keeper_runtime.mli` | drop `enabled : bool` from record + remove its docstring |
| `lib/keeper/keeper_runtime.ml` | drop field from type, 4-tuple accumulator (was 5-tuple), record literal in both early-exit and main branch, debug log format string, and the `let _ = stats.enabled in` suppression |
| comment | `maybe_start_supervisor_sweep` doc rewritten to describe current decoupled behaviour without referencing the dead field name |

Net diff: **+18 / -31 LOC**.

## Workaround self-check

| # | check | result |
|---|---|---|
| 1 | telemetry-as-fix? | no — counter unchanged |
| 2 | string classifier? | no |
| 3 | N-of-M? | no — single field, single record |
| 4 | catch-all `_ ->` added? | no |
| 5 | cap/cooldown/dedup/repair? | no |
| 6 | test backdoor? | no — `let _ = stats.enabled in` suppression *removed*, not added |
| 7 | same typo N sites? | no |

## Behavioural impact

None. Every reader already ignored the value. The 2026-04-24 incident's resolution (decouple supervisor startup from bootstrap success) stays in place — it just no longer needs a dead field to suppress.

## Test plan

- [x] No new tests — field removal is a type-level fact, surfaced at compile time
- [ ] CI build + test (local build blocked by `dune-local.sh` global lock)

[...]

PR #14726, opened 2026-05-11
refactor(cdal_runtime): replace manual to_int/equal/compare with [@@deriving eq, ord]

jeong-sik:
## Why

`Risk_class` and `Execution_mode` in `lib/cdal_runtime/` both use the same shape:

\`\`\`ocaml
let to_int = function Low -> 0 | Medium -> 1 | High -> 2 | Critical -> 3
let equal a b = to_int a = to_int b
let compare a b = Int.compare (to_int a) (to_int b)
\`\`\`

The `to_int` function is private (not in `.mli`) — it exists solely to back `equal` and `compare`. This is the textbook case where `[@@deriving eq, ord]` subsumes hand-written code entirely. The deriver-generated `compare` orders by *constructor declaration order*, which matches the existing `to_int` ranking exactly in both modules:

| Type | Constructor order | `to_int` value |
|------|-------------------|----------------|
| `Risk_class.t` | Low, Medium, High, Critical | 0, 1, 2, 3 |
| `Execution_mode.t` | Diagnose, Draft, Execute | 0, 1, 2 |

The hand-written shape is also forward-fragile: if a new constructor is inserted between two existing ones (e.g. `Verify` between `Draft` and `Execute`), a developer must remember to renumber `to_int` to preserve the ranking. The deriver does this automatically — the new constructor's rank is its position in the declaration.

## What

- Add `eq, ord` to the existing `[@@deriving]` lists (both `.ml` and `.mli`)
- Delete the manual `to_int`, `equal`, `compare` definitions in both `.ml` files
- Drop redundant `val equal` / `val compare` lines in both `.mli` files (the deriver attribute on the type generates them)
- Rewrite `Execution_mode.can_serve` to use the derived `compare` instead of the removed `to_int`
- Add `ppx_deriving.eq` and `ppx_deriving.ord` to `lib/cdal_runtime/dune` preprocess (already enabled in the main `lib/dune`)

## Caller compat

All external callers consume `equal` and `compare` with unchanged signatures:

- `test/test_masc_contract_catalog.ml:32,50` — `Risk_class.equal`
- `lib/cdal/cdal_judge.ml:17,69` — `Risk_class.equal`, `Execution_mode.equal`
- `lib/cdal_runtime/mode_resolver.ml:6,27,29` — `Execution_mode.compare`, `Execution_mode.equal`

No caller updates needed.

## Verification

\`\`\`
dune build @lib/cdal_runtime/check @lib/cdal/check --root .   # exit 0
\`\`\`

[...]

PR #14750, opened 2026-05-11
refactor(task_stage): replace manual index/can_transition with [@@deriving eq, ord]

jeong-sik:
## Why

`Task_stage.t` is an ordered 5-variant closed sum:

\`\`\`ocaml
type t = Decompose | Inspect | Implement | Verify | Review
\`\`\`

The module shipped a private `index : t -> int` mapping each constructor to its declaration position, and used it inside `can_transition` to compare ranks:

\`\`\`ocaml
let index = function
  | Decompose -> 0 | Inspect -> 1 | Implement -> 2 | Verify -> 3 | Review -> 4

let can_transition ~current ~target = index target >= index current
\`\`\`

This is the same pattern closed in PR #14726 for `cdal_runtime/risk_class` and `cdal_runtime/execution_mode`: a hand-written `to_int`-style helper whose only purpose is to back comparison. The deriver-generated `compare` from `[@@deriving ord]` orders constructors by declaration position — exactly the ranking `index` encoded by hand.

The hand-written shape is forward-fragile: if a new constructor is inserted between two existing ones (e.g. a `Refactor` step between `Implement` and `Verify`), `index` must be renumbered explicitly. The deriver does this automatically.

## What

- Add `eq, ord` to `[@@deriving]` (both `.ml` and `.mli`)
- Delete private `index : t -> int` (only consumer was one inline test)
- Rewrite `can_transition` as `compare target current >= 0`
- Update `test_task_stage.test_index_order` to verify canonical ordering through `Task_stage.compare` applied to `all`
- Add `ppx_deriving.ord` to `lib/types/dune` preprocess (alongside the existing `ppx_deriving.eq`)

## Caller compat

`grep -r 'Task_stage\.' lib/ bin/ test/` shows only:
- `Task_stage.to_string`, `to_yojson`, `of_yojson`, `of_string`, `all`, `can_transition`, `validate_transition`, `t` — all preserved
- `Task_stage.index` — used **only** in `test/test_task_stage.ml:22` (updated in this PR)

No production caller is affected. `Task_stage.equal` and `Task_stage.compare` are new vals exposed via the `.mli`'s `[@@deriving]` attribute.

## Verification

\`\`\`
dune build @lib/types/check --root .   # exit 0
\`\`\`

A wider build of `test/test_task_stage.exe` currently fails on `lib/keeper/keeper_runtime.ml:649` (5-tuple vs 4-tuple type mismatch). That regression is **pre-existing on main HEAD** (introduced by PR #14745) and unrelated to this PR. CI on a green main will validate the test.

## Why this counts as "Variant 처리가 더 유리한 곳"

Same reasoning as #14726: hand-written ordinal helpers on a closed sum are deriver-displaceable maintenance surface. Deleting `index` removes 7 LOC of code that the type system can generate exactly.

### DIFF
--- a/lib/types/dune
+++ b/lib/types/dune
@@ -5,4 +5,4 @@
  (wrapped false)
  (modules rate_limit_types masc_error ids types_core types_auth masc_domain task_stage severity)
  (libraries yojson unix time_compat masc_core base64 uuidm)
- (preprocess (pps ppx_deriving.show ppx_deriving.eq ppx_deriving_yojson)))
+ (preprocess (pps ppx_deriving.show ppx_deriving.eq ppx_deriving.ord ppx_deriving_yojson)))
--- a/lib/types/task_stage.ml
+++ b/lib/types/task_stage.ml
@@ -10,7 +10,7 @@ type t =
   | Implement
   | Verify
   | Review
-[@@deriving show]
+[@@deriving show, eq, ord]
 
 let to_string = function
   | Decompose -> "decompose"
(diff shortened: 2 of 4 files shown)


jeong-sik:
워크어라운드 거부 기준 7항목 리뷰: 전체 PASS. 수동 index 함수 제거 + ppx_deriving.ord 파생 compare로 교체. 새 variant 추가 시 자동 정렬 반영. 7개 항목 전체 해당 없음.

PR #14766, opened 2026-05-11
fix(keeper): two more mutation_preset_ok guards with explicit variant enumeration

jeong-sik:
## Summary

Two `*mutation_preset_ok` predicates in `lib/keeper/` still used the `_ -> false` catch-all anti-pattern that PR #14716 (keeper scope) and PR #14762 (lib top-level) sweep targeted. Both close here.

## Sites

| File | Function | Decides |
|------|----------|---------|
| `lib/keeper/keeper_tool_github_pr.ml:108` | `mutation_preset_ok` | whether `gh pr create` / `gh pr edit` / `gh pr review` actions are allowed |
| `lib/keeper/keeper_tool_pr_review.ml:42` | `pr_review_mutation_preset_ok` | whether `gh pr review` thread mutations are allowed |

Both used:

```ocaml
| Some (Research | Delivery | Coding | Full) -> true
| _ -> false
```

## Why the catch-all was wrong here

For a guard predicate, the catch-all default is the wrong safety direction. A new `Keeper_meta_tool_access.tool_access_preset` constructor (e.g. `Admin`, `ReadOnly`) silently inherits "allowed" — the opposite of what a credential-scope guard exists to prevent. The compiler stays silent on the policy gap.

## Fix

Both sites enumerate every preset constructor explicitly:

```ocaml
| Some (Research | Delivery | Coding | Full) -> true
| Some (Minimal | Social | Messaging | Dispatch) -> false
| None -> false
```

A new preset constructor now triggers a non-exhaustive-match warning at each guard, forcing the maintainer to deliberately classify it.

Behavior is unchanged for the eight current presets.

## Workaround self-check

| # | check | result |
|---|---|---|
| 1 | telemetry-as-fix? | no — guard semantics restored, no new counter |
| 2 | string classifier? | no — typed match on closed sum |
| 3 | N-of-M? | **yes, admitted** — this PR closes the two keeper-scope sites missed by #14716. After this PR, `rg -B3 '\| _ -> false$' lib/keeper/ \| rg -B2 'preset'` is empty for this pattern shape |
| 4 | catch-all `_ ->` added? | REMOVED, not added |
| 6 | test backdoor? | no |

## Test plan

- [x] Both sites cover all 8 preset variants explicitly

### DIFF
--- a/lib/keeper/keeper_exec_preflight.ml
+++ b/lib/keeper/keeper_exec_preflight.ml
@@ -64,7 +64,8 @@ let handle_keeper_preflight_check
   let preset_ok =
     match Keeper_types.tool_access_preset meta.tool_access with
     | Some (Research | Coding | Delivery | Full) -> true
-    | _ -> false
+    | Some (Minimal | Social | Messaging | Dispatch) -> false
+    | None -> false
   in
   let preset_name =
     match Keeper_types.tool_access_preset meta.tool_access with
--- a/lib/keeper/keeper_tool_github_pr.ml
+++ b/lib/keeper/keeper_tool_github_pr.ml
@@ -105,9 +105,15 @@ let draft_request_allowed args =
    tried a Docker PR-lifecycle proof. Mirrors
    [Keeper_tool_pr_review.pr_review_mutation_preset_ok], which already
    includes Research. *)
+(* For a guard predicate the catch-all default is the wrong safety
+   direction: any future preset would silently inherit "allowed".
+   Enumerate every variant so the compiler forces a deliberate
+   classification when [Keeper_meta_tool_access.tool_access_preset]
+   grows.  PR #14716 / #14762 follow-up. *)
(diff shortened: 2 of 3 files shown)


jeong-sik:
7-item workaround rejection checklist — PASS. Companion to #14716/#14762 — eliminates _ -> false catch-all in 2 remaining guard predicates (기준 #4 inverse). Explicit variant enumeration forces compiler to flag new Keeper_meta_tool_access.tool_access_preset variants. +14/-2, surgical.

jeong-sik:
Added 3rd site fold-in (41c66c3c0): `keeper_exec_preflight.ml:preset_ok` had identical pattern (`Some (Research | Coding | Delivery | Full) -> true | _ -> false`). Same closed-sum `tool_preset` enumeration. PR title still accurate — folding 3rd N-of-M site avoids separate follow-up.

PR #14769, opened 2026-05-11
perf(types_auth): O(1) has_permission via direct (role, permission) variant match

jeong-sik:
## Summary

\`has_permission role permission\` did \`List.mem permission (permissions_for_role role)\` — allocating a fresh 12-14 element \`permission\` list per call. This is hot path:

- \`Auth.check_permission\` (\`lib/auth.ml:1208\`) — every protected operation
- \`Auth_doctor\` — 10+ call sites per credential snapshot

Replace with a direct \`(role, permission)\` variant match. Both types are closed sum types, so OCaml's exhaustiveness check enforces coverage — adding a new \`permission\` constructor breaks the match at compile time rather than letting it silently fall through.

## Diff shape

\`\`\`ocaml
let has_permission role permission =
  match role, permission with
  | Admin, _ -> true
  | Worker, (CanInit | CanReset | CanAdmin) -> false
  | Worker, ( CanJoin | CanLeave | CanReadState | CanAddTask
            | CanClaimTask | CanCompleteTask | CanBroadcast
            | CanOpenPortal | CanSendPortal | CanCreateWorktree
            | CanRemoveWorktree | CanVote ) -> true
\`\`\`

## Semantic preservation

The new match exactly mirrors the existing \`permissions_for_role\` table:
- **Admin** has all 15 permissions → \`Admin, _ -> true\`
- **Worker** has 12 permissions, missing {CanInit, CanReset, CanAdmin}

Both \`test_has_permission\` and \`test_permissions_for_role_*\` in \`test/test_types_utils_coverage.ml\` (and \`test_types_coverage.ml\`) verify each function independently — they don't rely on cross-implementation consistency, so neither test breaks.

## Why direct match over Lazy-memoized list

A Lazy-memoized list would still do \`List.mem\` linear scan per check (cheap for 12-14 elements but not free). Direct variant match compiles to a jump table — strictly O(1). And it makes the policy decision visible in the source: any maintainer reading \`has_permission\` immediately sees which permissions are admin-only without chasing a list construction site.

## Parallel knowledge with \`permissions_for_role\`

Intentional. The single source of truth is the \`permission\` variant declaration; both \`permissions_for_role\` and \`has_permission\` are compiler-checked exhaustive against it. Adding a new permission requires updating both — that's the desired property, not duplication waste.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_types_utils_coverage.exe\` passes
- [ ] \`dune runtest test/test_types_coverage.exe\` passes
- [ ] \`Auth_doctor\` snapshot endpoint returns same values

### DIFF
--- a/lib/types/types_auth.ml
+++ b/lib/types/types_auth.ml
@@ -191,7 +191,8 @@ let permissions_for_role = function
    allocation.  Hot path: [Auth.check_permission] runs this on every
    protected operation; [Auth_doctor] runs it 10+ times per snapshot.
    The previous [List.mem permission (permissions_for_role role)] form
-   built a fresh 12-14 element list each call.
+   built a fresh 12-element (Worker) / 15-element (Admin) list each
+   call.
 
    Parallel to [permissions_for_role]: both forms are compiler-checked
    exhaustive against the [permission] variant, so adding a new
--- a/lib/types/types_auth.ml
+++ b/lib/types/types_auth.ml
@@ -187,8 +187,24 @@ let permissions_for_role = function
       CanVote; CanAdmin;
     ]
 
+(* Direct (role, permission) variant match — O(1), no per-call list
+   allocation.  Hot path: [Auth.check_permission] runs this on every
+   protected operation; [Auth_doctor] runs it 10+ times per snapshot.
+   The previous [List.mem permission (permissions_for_role role)] form
+   built a fresh 12-14 element list each call.
+
+   Parallel to [permissions_for_role]: both forms are compiler-checked
+   exhaustive against the [permission] variant, so adding a new
+   constructor breaks both at compile time rather than letting one
+   silently fall through to a default. *)
 let has_permission role permission =
-  List.mem permission (permissions_for_role role)
+  match role, permission with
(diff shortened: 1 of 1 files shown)


jeong-sik:
7-Item Workaround Rejection Checklist — PASS. (1) Telemetry-as-fix: PASS — performance optimization O(n)->O(1), not counter-as-fix. (2) String classifier: PASS — direct (role, permission) variant match on closed sums. (3) N-of-M: PASS — single site. (4) Catch-all: PASS — exhaustive match. (5) Dedup: PASS — eliminates allocation. (6) Test backdoor: PASS — none. (7) N-site typo: PASS — single site. Assessment: Clean. Both role and permission are closed sums, OCaml exhaustiveness enforces coverage.

jeong-sik:
Re. `lib/auth.ml:1208` thread: the comment in question already references only the function name `[Auth.check_permission]` — no line number is present in the file (verified with `rg '1208|auth\.ml:' lib/types/types_auth.ml`). Suspect Copilot read an earlier revision or PR-body context. Resolving as already-applied.

PR #14736, opened 2026-05-11
refactor(provider): timeout_bounds_of_kind boundary helper (RFC-0058 Phase 5.6 leak 4/4)

jeong-sik:
## Summary

The last `match provider_cfg.kind` inside the keeper layer moves into `Provider_adapter.timeout_bounds_of_kind`. Closes the final RFC-0058 Phase 5.6 leak in keeper scope (after #14691 / #14710 / #14717 / #14721).

## Why

`Keeper_turn_driver_helpers.provider_attempt_timeout_constraints` inlined per-provider operational tunables (ollama needs 300s for model loading, claude_code CLI hangs past 120s, gemini past 180s, kimi_cli past 60s). Operational tunables belong inside the adapter boundary — adding vLLM/lmstudio should mean editing one site, not the keeper turn driver.

## Before / After

**Before** — `lib/keeper/keeper_turn_driver_helpers.ml:39-51`

```ocaml
let provider_attempt_timeout_constraints (cfg : Llm_provider.Provider_config.t) =
  match cfg.kind with
  | Llm_provider.Provider_config.Ollama -> { min = Some 300.0; max = None }
  | Claude_code                         -> { min = None; max = Some 120.0 }
  | Gemini | Gemini_cli                 -> { min = None; max = Some 180.0 }
  | Kimi_cli                            -> { min = None; max = Some 60.0 }
  | Anthropic | Kimi | OpenAI_compat | Glm | DashScope | Codex_cli ->
    { min = None; max = None }
```

**After** — boundary helper holds the match; keeper-side is a one-liner with a manifest alias:

```ocaml
(* lib/provider_adapter.ml *)
type timeout_bounds =
  { min_timeout_s : float option; max_timeout_s : float option }

let timeout_bounds_of_kind kind = (* same 5-arm match, lives here now *) ...
```

```ocaml
(* lib/keeper/keeper_turn_driver_helpers.ml *)
type provider_attempt_timeout_constraints =
  Provider_adapter.timeout_bounds = {
    min_timeout_s : float option; max_timeout_s : float option;
  }

let provider_attempt_timeout_constraints (cfg : Llm_provider.Provider_config.t) =
  Provider_adapter.timeout_bounds_of_kind cfg.kind
```

Caller code (`Keeper_turn_driver.effective_provider_attempt_timeout_s` + `test_oas_worker.ml`) compiles unchanged — the record name and field shape are preserved by the manifest alias.

## Sealing tests

5 new test cases over every `provider_kind` arm:

[...]

PR #14721, opened 2026-05-11
refactor(cascade): caller-driven explicit capacity register, kill looks_like_ollama

jeong-sik:
## Summary

Follow-up to PR #14710 (Http_probe explicit URL registry). Removes the remaining substring-scan classifier inside `Cascade_client_capacity`. The keeper-side caller now consults `Provider_adapter.is_http_probe_capable_kind` (new capability predicate, boundary-only) and calls `register` directly.

## Why

`looks_like_ollama` aliased `Masc_network_defaults.is_ollama_url`, the same `:11434` substring scan we already retired from `Cascade_http_probe`. Two failure modes:

- vLLM/lmstudio on `:11434` mis-classified as ollama
- ollama on a non-default port invisible to auto-registration

`Keeper_turn_driver` already holds `candidate_cfgs : Provider_config.t list`. The typed kind is right there — no need to guess from URL string.

## Before / After

**Before** — keeper hands raw `base_urls` to a `cascade_client_capacity` heuristic:

```ocaml
match ollama_max with
| None ->
  Cascade_client_capacity.auto_register_for_candidates
    ~base_urls:candidate_base_urls
| Some n ->
  Cascade_client_capacity.auto_register_ollama_with_override
    ~base_urls:candidate_base_urls ~max_concurrent:n
```

**After** — keeper consults a typed capability predicate and registers explicitly:

```ocaml
let http_probe_max_concurrent = Option.value ollama_max ~default:1 in
List.iter (fun (cfg : Llm_provider.Provider_config.t) ->
  if Provider_adapter.is_http_probe_capable_kind cfg.kind
     && cfg.base_url <> ""
  then begin
    if not (Cascade_client_capacity.is_registered cfg.base_url) then
      Cascade_client_capacity.register
        ~url:cfg.base_url ~max_concurrent:http_probe_max_concurrent;
    Cascade_http_probe.register_url ~url:cfg.base_url
  end
) candidate_cfgs
```

## Dead code removed

| Symbol | Status |
|---|---|
| `Cascade_client_capacity.looks_like_ollama` | removed (alias for `is_ollama_url`) |
| `Cascade_client_capacity.auto_register_for_candidates` | removed |
| `Cascade_client_capacity.auto_register_ollama_with_override` | removed |
| `test_ollama_auto_register`, `test_ollama_register_with_override` | removed (substring-path tests) |

Net diff: **+71 / -96 LOC**.

## New boundary helper

[...]

PR #14771, opened 2026-05-11
fix(provider): is_http_probe_capable_kind enumerate every variant

jeong-sik:
## Summary

Self-correction: the [`is_http_probe_capable_kind`] helper that PR #14721 introduced still carried a `_ -> false` catch-all. The PR's self-check said "no catch-all added — exhaustive over all 11 kinds" but the source said otherwise. Replace the wildcard with explicit enumeration so the compiler forces a deliberate classification when a new `provider_kind` constructor lands.

## Why it matters

The function decides whether `Keeper_turn_driver` should `Cascade_http_probe.register_url cfg.base_url` for that cfg. When a future local HTTP-probe-capable provider (vLLM, lmstudio, llama-server) ships as a new constructor:

- **Before this PR**: wildcard silently returns `false`. Capacity registry never sees the URL → saturation pre-skip never fires → operator loses a quiet performance optimisation **with no compiler signal**.
- **After this PR**: compiler errors at this boundary site, the maintainer makes the call deliberately.

Same anti-pattern that PR #14716 / #14762 / #14766 closed in keeper-side guard predicates — this one was on the adapter side, missed in our own self-check.

## Before / After

```ocaml
(* Before — wildcard absorbs 10 constructors *)
let is_http_probe_capable_kind kind =
  match kind with
  | Llm_provider.Provider_config.Ollama -> true
  | _ -> false

(* After — every variant is named *)
let is_http_probe_capable_kind kind =
  match kind with
  | Llm_provider.Provider_config.Ollama -> true
  | Anthropic | Claude_code | OpenAI_compat | Glm | DashScope | Codex_cli
  | Gemini | Gemini_cli | Kimi | Kimi_cli -> false
```

Sibling helper `timeout_bounds_of_kind` in the same module (PR #14736) already uses explicit enumeration; this PR brings `is_http_probe_capable_kind` into the same shape.

Behavior is unchanged for the current 11 kinds.

## Workaround self-check

| # | check | result |
|---|---|---|
| 1 | telemetry-as-fix? | no |
| 2 | string classifier? | no — typed match on closed sum |
| 3 | N-of-M? | no — single site, finishes a sweep started by #14716 / #14762 / #14766 |
| 4 | catch-all `_ ->` added? | REMOVED, not added |
| 6 | test backdoor? | no |

## Test plan

- [x] Boundary helper preserved for current 11 kinds

### DIFF
--- a/lib/provider_adapter.ml
+++ b/lib/provider_adapter.ml
@@ -796,7 +796,8 @@ let is_local_provider pname =
 let is_http_probe_capable_kind (kind : Llm_provider.Provider_config.provider_kind) : bool =
   match kind with
   | Llm_provider.Provider_config.Ollama -> true
-  | _ -> false
+  | Anthropic | Claude_code | OpenAI_compat | Glm | DashScope | Codex_cli
+  | Gemini | Gemini_cli | Kimi | Kimi_cli -> false
 ;;
 
 (** Per-provider per-attempt timeout bounds.


jeong-sik:
7-Item Workaround Rejection Checklist — PASS. (1) Telemetry-as-fix: PASS — not a counter, it's a boolean gate fix. (2) String classifier: PASS — closed sum variant match, no strings. (3) N-of-M: PASS — single site, single change. (4) Catch-all removal: EXACTLY THE FIX — this PR removes _ -> false and replaces with explicit enumeration of all 10 non-Ollama variants. This is the structural pattern the rejection criteria promote. Compiler now forces classification when a new provider_kind constructor is added. (5) Dedup: PASS — unrelated. (6) Test backdoor: PASS — none. (7) N-site typo: PASS — single site. Assessment: Clean. Self-correction of a catch-all that PR #14721's self-check missed. This is exactly what 기준 #4 demands — exhaustiveness enforced by the compiler, not by convention.

PR #14728, opened 2026-05-11
perf(tool_access_policy): set-based membership for Inter/Diff/resolve

jeong-sik:
## Summary

Replaces O(N×M) `List.mem` set operations with `StringSet` (O(N log M)) inside `tool_access_policy.ml`. Also drops the redundant `StringSet` dedup that `selector_matches_name` was doing on the `Names` branch for membership queries — duplicates don't change membership, so the dedup was pure overhead per call.

## Changes (internal only — public ADT unchanged)

1. **`selector_matches_name` Names branch** — was `List.mem name (normalize_names names)`, building a fresh `StringSet` + list conversion per call. New helper `name_in_normalized_list` does `List.exists` with inline trim/empty-skip, short-circuiting on first match.
2. **`resolve_selector` Inter fold** — `List.filter (fun n -> List.mem n resolved)` → `StringSet.of_list resolved` built once per fold step, then `StringSet.mem`.
3. **`resolve_selector` Diff** — same pattern for `exclude_set`.
4. **`resolve` final deny filter** — `denied` list converted to set once.

## Hot callers benefiting

| Caller | Path | Frequency |
|--------|------|-----------|
| `Auth.authorize` | `lib/auth.ml:1340` `allows_name` | per-request |
| `Eval_gate.pre_check` | `lib/eval_gate.ml:221,224` (2 calls) | per tool call in eval |
| `Worker_oas` PreToolUse hook | `lib/worker_oas.ml:331` | per tool invocation |
| `Keeper_tool_policy.compute_resolved_allowlist` | `lib/keeper/keeper_tool_policy.ml:431` | per keeper turn |

## Semantic preservation

All semantics preserved (trim, empty-skip, dedup at final `normalize_names`). Existing tests in `test/test_tool_access_policy.ml` and `test/test_tool_access_role.ml` should pass unchanged.

## Test plan

- [ ] CI lib/ build green
- [ ] `dune runtest test/test_tool_access_policy.exe` passes
- [ ] `dune runtest test/test_tool_access_role.exe` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14774, opened 2026-05-11
perf(tool_bridge): O(1) required-field check in params_of_json_schema

jeong-sik:
## Summary

\`params_of_json_schema\` iterated over schema properties (\`pairs\`) and for every property ran \`List.mem name required_list\` — an O(R) linear scan over the required-fields list.

| Variable | Typical size |
|----------|--------------|
| Properties (P) | 10-30 per schema |
| Required (R) | 1-10 per schema |

Per call: **O(P × R)** = ~50-300 string compares.

\`oas_tool_of_masc\` fires \`params_of_json_schema\` per OAS tool conversion, which happens on every cascade decision touching the tool. With multiple tools per cascade decision and the conversion firing per call, the wasted work compounds.

## Fix

The \`required\` field is conceptually a **set** — membership semantics matter, ordering does not, and duplicates would be meaningless. Materialise it as \`(string, unit) Hashtbl.t\` instead of \`string list\`:

| Before | After |
|--------|-------|
| \`List.filter_map (function …) items\` → list | \`List.iter (function …) items\` → hashtbl |
| \`let required = List.mem name required_list\` — O(R) | \`let required = Hashtbl.mem required_set name\` — O(1) |

Build cost is identical (one pass over \`items\`). Per-property check is O(1).

## Why the type change is "고도화"

The original \`required_list : string list\` expressed *implementation* (list-based). The new \`required_set : (string, unit) Hashtbl.t\` expresses *intent* (membership). Reading the function, you now immediately see that \`required\` is being used set-wise — no ordering invariant could possibly matter. This is the OCaml equivalent of "make illegal states unrepresentable", applied to mutable internal representations.

## Compounding with prior loop PRs

Combine with #14728/#14734/#14740/#14749/#14756/#14763: a single OAS tool render now uses O(1) lookups in every layer — auth (Tool_access_role), surface dispatch (Tool_catalog_surfaces), metadata (Tool_catalog), help registry (Tool_help_registry), and now schema conversion (Tool_bridge).

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_tool_bridge.exe\` passes (if present)
- [ ] OAS tool conversion smoke — required flags still set correctly

### DIFF
--- a/lib/tool_bridge.ml
+++ b/lib/tool_bridge.ml
@@ -183,7 +183,12 @@ let params_of_json_schema schema =
   let required_set =
     match schema |> member "required" with
     | `List items ->
-        let tbl = Hashtbl.create (List.length items) in
+        (* Constant initial size 16: avoid the extra [List.length items]
+           pass (which itself is O(R)) before [List.iter].  Hashtbl
+           auto-resizes — sizing exactly to the input only saves a
+           handful of resizes per call, which is cheaper than re-walking
+           the list. *)
+        let tbl = Hashtbl.create 16 in
         List.iter
           (function
             | `String value -> Hashtbl.replace tbl value ()
--- a/lib/tool_bridge.ml
+++ b/lib/tool_bridge.ml
@@ -175,15 +175,22 @@ let type_string_of_schema_property prop =
 
 let params_of_json_schema schema =
   let open Yojson.Safe.Util in
-  let required_list =
+  (* [required] is conceptually a set (membership semantics, no ordering
+     or duplicates) — materialise as Hashtbl so the per-property check
+     below is O(1) instead of O(R) per property.  Per-call savings scale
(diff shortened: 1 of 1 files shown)


jeong-sik:
7-item workaround rejection checklist: **PASS** ✓ 1. Telemetry-as-fix: N/A (pure perf) 2. String classifier: N/A (list→set, correct direction) 3. N-of-M: N/A (single site) 4. Catch-all: N/A (JSON parsing wildcard is legitimate) 5. Cap/Cooldown: N/A 6. Test backdoor: N/A 7. Same typo N-site: N/A — Clean perf PR. O(P×R)→O(P) with Hashtbl for set semantics is structurally correct;  has membership semantics (no ordering, no duplicates) so the data structure matches the domain.

PR #14756, opened 2026-05-11
perf(tool_help_registry): fetch Tool_catalog.metadata once per entry_of_schema

jeong-sik:
## Summary

\`entry_of_schema\` triggered **3× \`Tool_catalog.metadata\` lookups** for every schema without a manual help entry:

| Helper called | Internal call |
|---------------|---------------|
| \`derived_short_description schema.name ...\` | \`Tool_catalog.metadata name\` |
| \`constraints_from_metadata schema.name\` | \`Tool_catalog.metadata name\` |
| \`derived_details schema.name ...\` | \`constraints_from_metadata name\` → \`Tool_catalog.metadata name\` |

For tool catalog rendering — \`canonicalize_schemas\`, \`masc_tool_help\`, OAS schema sync — that's **N × 3** catalog lookups per render.

## Refactor

Introduce \`_with_meta\` / \`_from_meta\` internal helpers that take a pre-fetched \`Tool_catalog.metadata\`:

| Public (preserved) | Internal (new) |
|--------------------|----------------|
| \`constraints_from_metadata : name -> string list\` | \`constraints_from_meta : metadata -> string list\` |
| \`derived_short_description : name -> orig -> string\` | \`derived_short_description_with_meta : metadata -> name -> orig -> string\` |
| \`derived_details : name -> orig -> string\` | \`derived_details_with_meta : metadata -> orig -> string\` |

Public arrows kept as thin wrappers that fetch metadata then delegate — backward compatibility preserved. None of the public helpers appear in the \`.mli\` (only mentioned in doc comments), but signature symmetry simplifies future use.

\`entry_of_schema\` now does \`let meta = Tool_catalog.metadata schema.name\` once and passes \`meta\` to all three helpers.

## Compounding

This PR layers on top of #14749 (\`Tool_catalog.metadata\` hot-path improvements). Together: 3× lookups → 1× lookup, each lookup itself cheaper. Net effect for a public-surface help render is roughly **3N× fewer catalog operations**.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_tool_help_registry.exe\` (if present) passes
- [ ] \`masc_tool_help\` smoke test still returns expected fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14780, opened 2026-05-11
perf(benchmark_scoring): O(C+R) replaces O(C×R) in required/forbidden tool checks

jeong-sik:
## Summary

Three scoring helpers in \`lib/tool_call_quality_benchmark_scoring.ml\` had compounding O(N×M) shapes per scored benchmark case:

| Helper | Pattern | Cost |
|--------|---------|------|
| \`required_tool_score\` | for each required_tool, \`tool_used\` scans \`run.tool_calls\` | O(R × C) |
| \`forbidden_tool_used\` | for each forbidden_tool, \`tool_used\` scans \`run.tool_calls\` | O(F × C) |
| \`unnecessary_tool_rate\` | for each tool_call, \`List.mem call.tool_name forbidden_tools\` | O(C × F) |

With C~100 tool calls and R/F~5-10, that's **~3000 string compares per scored case**, across thousands of cases per benchmark run.

## Fix

Two thin Hashtbl builders, applied differently based on iteration direction:

\`\`\`ocaml
let tool_call_name_set (run : evidence_run) =
  let tbl = Hashtbl.create (List.length run.tool_calls) in
  List.iter (fun call -> Hashtbl.replace tbl call.tool_name ()) run.tool_calls;
  tbl

let name_set_of_list names =
  let tbl = Hashtbl.create (List.length names) in
  List.iter (fun name -> Hashtbl.replace tbl name ()) names;
  tbl
\`\`\`

- \`required_tool_score\` / \`forbidden_tool_used\`: iteration is *"for each required/forbidden name, is it in the run?"* → build the set from **run.tool_calls**.
- \`unnecessary_tool_rate\`: iteration is reversed *"for each tool_call, is its name in forbidden?"* → build the set from **forbidden_tools**.

**General rule applied**: build the set from whichever side hosts the membership lookup, not whichever side is being iterated. Recognizing this asymmetry is what determines which Hashtbl to materialize.

## Compounding with prior loop PRs

Continues the per-call cost-shedding established in:
- #14728/#14734/#14740/#14749 (auth + catalog hot path)
- #14756/#14763 (help registry + dispatch profile)
- #14769/#14774 (typed permission match + schema required-set)

Benchmark scoring isn't on the auth dispatch hot path, but score_run runs across thousands of benchmark cases per evaluation pass — same N×M shape, different domain.

## Semantic preservation

All three helpers preserve their existing semantics:
- \`Hashtbl.mem set name\` ↔ \`List.mem name list\` on the same name set
- Build order doesn't matter for membership (no ordering invariant)
- Empty lists still yield empty sets → still produce 0 matches

\`tool_used\` is kept as a public helper since other code in this module may still use it (single-name shortcut).

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_tool_call_quality_benchmark*.exe\` passes
- [ ] Smoke: score_run output equals pre-PR output on a sample evidence_run

### DIFF
--- a/lib/tool_call_quality_benchmark_scoring.ml
+++ b/lib/tool_call_quality_benchmark_scoring.ml
@@ -97,6 +97,19 @@ let evaluate_json_check ~(target : Yojson.Safe.t option) (check : json_check) =
 let tool_used tool_name (run : evidence_run) =
   List.exists (fun call -> String.equal call.tool_name tool_name) run.tool_calls
 
+(* Build a name-keyed set over [run.tool_calls] in one pass.  Used by
+   [required_tool_score] and [forbidden_tool_used] to replace
+   O(R × C) repeated linear scans with O(C + R) build + lookup. *)
+let tool_call_name_set (run : evidence_run) =
+  let tbl = Hashtbl.create (List.length run.tool_calls) in
+  List.iter (fun call -> Hashtbl.replace tbl call.tool_name ()) run.tool_calls;
+  tbl
+
+let name_set_of_list names =
+  let tbl = Hashtbl.create (List.length names) in
+  List.iter (fun name -> Hashtbl.replace tbl name ()) names;
+  tbl
+
 let arg_check_passes (run : evidence_run) (check : arg_check) =
   let predicate call =
     String.equal call.tool_name check.tool_name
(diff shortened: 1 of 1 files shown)


jeong-sik:
7-item workaround rejection checklist: **PASS** ✓ Clean perf PR. O(R×C)→O(C+R) via Hashtbl set for membership-semantics lookups. Same structural pattern as #14774 (list→hashtbl for set membership). Three sites in one file fixed simultaneously.

PR #14785, opened 2026-05-11
perf(keeper_repo_mapping): consolidate + O(R+M) filter_repos_by_mapping helper

jeong-sik:
## Summary

\`lib/repo_manager/keeper_repo_mapping.ml\` had **two copy-pasted O(R×M) loops** doing the same repo-filter logic:

\`\`\`ocaml
(* credentials_for_keeper, lines 163-170 *)
if List.exists (String.equal "*") mapping.repository_ids then repos
else
  List.filter
    (fun r -> List.exists (String.equal r.id) mapping.repository_ids)
    repos

(* apply_mapping, lines 256-262 — identical *)
if List.exists (String.equal "*") mapping.repository_ids then repositories
else
  List.filter
    (fun r -> List.exists (String.equal r.id) mapping.repository_ids)
    repositories
\`\`\`

For R repos × M mapping ids without the \`"*"\` wildcard, each call did R × M string compares.

## Fix

Extract a single private helper:

\`\`\`ocaml
let filter_repos_by_mapping (mapping : keeper_repo_mapping)
    (repos : repository list) : repository list =
  if List.exists (String.equal "*") mapping.repository_ids then
    repos
  else
    let mapping_id_set =
      let tbl = Hashtbl.create (List.length mapping.repository_ids) in
      List.iter (fun id -> Hashtbl.replace tbl id ()) mapping.repository_ids;
      tbl
    in
    List.filter
      (fun (r : repository) -> Hashtbl.mem mapping_id_set r.id)
      repos
\`\`\`

| Path | Before | After |
|------|--------|-------|
| Wildcard \`"*"\` present | O(M) check, return all | unchanged — wildcard skipped Hashtbl build |
| Wildcard absent | R × M string compares | O(M) build + R × O(1) lookup = O(R + M) |

## Why two-for-one consolidation

When the identical inefficiency appears in 2+ call sites, the shared helper is the right abstraction. Each call site now routes through one implementation — any future change to the wildcard/membership semantics happens in exactly one place.

## API preservation

\`keeper_repo_mapping.mli\` is unchanged — the helper is private to the \`.ml\`. \`credentials_for_keeper\` and \`apply_mapping\` keep their existing signatures.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_keeper_repo_mapping.exe\` (if present) passes
- [ ] Smoke: a keeper with \`"*"\` mapping sees all repos; a keeper with explicit id list sees only mapped repos

### DIFF
--- a/lib/repo_manager/keeper_repo_mapping.ml
+++ b/lib/repo_manager/keeper_repo_mapping.ml
@@ -115,16 +115,21 @@ let allowed_repositories ~keeper_id ~base_path =
   let* mapping = find_mapping ~base_path keeper_id in
   Ok mapping.repository_ids
 
+let is_wildcard s = s = "*"
+
 (* Filter [repos] down to those whose id appears in
    [mapping.repository_ids], with ["*"] as a wildcard that bypasses
    filtering entirely.  Replaces two copy-pasted O(R x M) loops in
    [credentials_for_keeper] and [apply_mapping]: each was
    [List.filter (fun r -> List.exists (String.equal r.id) mapping.repository_ids) repos].
-   The set is materialised only when no wildcard short-circuits the
-   check, so the wildcard case stays allocation-free. *)
+   The Hashtbl materialisation is skipped when a wildcard short-circuits
+   the check, so the wildcard case avoids building the membership set
+   (the [is_wildcard] predicate itself is a fully-saturated function so
+   no closure is allocated per call, unlike [(String.equal "*")] which
+   would partially apply). *)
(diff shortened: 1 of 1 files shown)


jeong-sik:
7-item workaround rejection checklist: **PASS** ✓ Two copy-pasted O(RxM) loops consolidated into one  helper with Hashtbl set. This is abstraction extraction (criterion #3 opposite direction: one shared helper replaces N duplicate sites). Wildcard short-circuit stays allocation-free.

PR #14796, opened 2026-05-11
perf(thompson_sampling): O(1) is_selected via side-channel Hashtbl

jeong-sik:
## Summary

\`select_with_feedback\` ran three selection phases (priority triggers → starvation rescue → Thompson sampling) and used a \`string list ref\` to track which agents had already been picked. Four sites scanned that list via \`List.mem name !selected_names\` to skip them:

| Phase | Site | Cost |
|-------|------|------|
| Priority triggers | line 486 | per-trigger O(K) |
| Starvation filter_map | line 523 | per-agent O(K) |
| Starvation iter | line 537 | per-starved O(K) |
| Thompson filter_map | line 556 | per-agent O(K) |

With C candidates × K accumulated picks, each phase paid O(C × K) string compares.

## Fix

Mirror the names in a side-channel \`(string, unit) Hashtbl.t\` alongside the existing \`selected\` list. The list keeps its role (ordering + \`List.length\` checks); the Hashtbl carries only membership:

\`\`\`ocaml
let selected = ref [] in
let selected_names : (string, unit) Hashtbl.t = Hashtbl.create max_n in
let is_selected name = Hashtbl.mem selected_names name in
let mark_selected name = Hashtbl.replace selected_names name () in
let add_selected result =
  selected := !selected @ [result];
  mark_selected result.agent_name
in
\`\`\`

Four call sites swap \`List.mem name !selected_names\` → \`is_selected name\`. The \`take\` loop's manual append to \`selected_names\` becomes \`mark_selected r.agent_name\`.

## What I deliberately did not touch

\`selected := !selected @ [result]\` (line 476) is O(|selected|) per append, totaling O(N²) across N adds. Tempting to switch to cons, **but** the final \`List.stable_sort\` at line 599 uses the input order as the tie-break. Changing input order would change which tied-score result wins selection across runs — a semantic shift, not just a perf swap. Leaving that pattern alone keeps this PR purely additive on correctness.

## Pattern: accumulator-as-list

When a multi-phase loop maintains a \`thing_names : string list ref\` and reads it via \`List.mem\`, that's an accumulator-as-list. Splitting into a list (ordering) + Hashtbl (membership) is the same shape applied in #14774, #14780, #14785, #14791 throughout this loop.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_thompson_sampling.exe\` passes
- [ ] Smoke: \`select_with_feedback\` returns identical results to pre-PR for a sample run

[...]

PR #14791, opened 2026-05-11
perf(agent_tool_surfaces): three set-based membership swaps for spawn-time helpers

jeong-sik:
## Summary

\`lib/agent_tool_surfaces.ml\` had three remaining \`List.mem\`/\`List.exists\` patterns on the spawn-time path. Two other functions in the file (\`lookup_schemas_by_name_exn\` line 53, \`filter_catalog_to_available\` line 196) already used Hashtbl / StringSet — but three sibling helpers had been left as O(N×M):

| Helper | Pattern | Cost |
|--------|---------|------|
| \`select_public_local_worker_schemas\` (line 134) | per-schema \`List.mem schema.name wanted\` | O(S × W) |
| \`resolve_named_schemas\` (lines 147 + 154) | filter + missing-check both linear-scan | O(S × R) + O(R × found) |
| \`build_tool_catalog\` autonomous branch (line 222) | per-name \`List.mem name admin_tool_names\` | O(all × admin) |

All three fire on agent spawn — multiple times for fleet boot.

## Fix

Introduce a private \`name_set : string list -> (string, unit) Hashtbl.t\` at the top of the module:

\`\`\`ocaml
let name_set names =
  let tbl = Hashtbl.create (List.length names) in
  List.iter (fun name -> Hashtbl.replace tbl name ()) names;
  tbl
\`\`\`

Then:

1. **\`select_public_local_worker_schemas\`** — build \`wanted_set\` once, replace \`List.mem\` with \`Hashtbl.mem\`.
2. **\`resolve_named_schemas\`** — materialise *both* directions (\`requested_set\` for the first filter, \`found_set\` for the missing-name check). This kills the worst O(N×M) in the file: the nested \`List.exists\` scan over already-filtered schemas.
3. **\`build_tool_catalog\` autonomous branch** — build \`admin_set\` once, swap \`List.mem\` for \`Hashtbl.mem\`.

## Pattern continuity

This is the **same helper shape** the loop has established in prior PRs:
- #14774 \`required_set\` for OAS schema conversion
- #14780 \`tool_call_name_set\` / \`name_set_of_list\` for benchmark scoring
- #14785 \`mapping_id_set\` for keeper-repo mapping

When the same primitive keeps surfacing, the codebase is asking for a shared utility. \`name_set\` could promote to a common library after a few more occurrences justify the move.

## API preservation

[...]

PR #14801, opened 2026-05-11
perf(mcp_server_eio_tool_profile): O(1) managed_agent passthrough membership

jeong-sik:
## Summary

\`tool_schemas_for_profile Managed_agent\` (lib/mcp_server_eio_tool_profile.ml:100-105) runs per managed-agent \`tools/list\` request. Its passthrough filter iterated \`Config.visible_tool_schemas\` (~150 schemas) and for every schema did:

\`\`\`ocaml
List.mem schema.name managed_agent_passthrough_tool_names  (* ~20 names *)
\`\`\`

That's ~3000 string compares per request, just for the passthrough membership predicate.

## Fix

\`managed_agent_passthrough_tool_names\` is module-load-constant (derived from \`spawned_agent_public_tool_names\` minus 4 hardcoded names). Build a sibling Hashtbl once at module init:

\`\`\`ocaml
let managed_agent_passthrough_tool_set : (string, unit) Hashtbl.t =
  let tbl =
    Hashtbl.create (List.length managed_agent_passthrough_tool_names)
  in
  List.iter
    (fun name -> Hashtbl.replace tbl name ())
    managed_agent_passthrough_tool_names;
  tbl
\`\`\`

Per-schema check drops to O(1).

## What I intentionally did NOT change

Three other \`List.mem\` sites in this file (lines 31, 430, 475):

| Site | List size | Calls/sec |
|------|-----------|-----------|
| Line 31 (\`managed_agent_passthrough_tool_names\` exclude list) | 4 elements (static) | once at module init |
| Line 430 (\`tools/list\` param allowlist) | 6 elements (static) | per-request, small |
| Line 475 (\`cursor\`-only param allowlist) | 2 elements (static) | per-request, small |

For lists of 2-6 elements, \`String.equal × N\` beats \`Hashtbl.mem\`'s lookup constant factor. The decision rule applied: optimize when (list size ≥ ~8) AND (per-request); leave alone otherwise. This is the discipline missing from "optimize everywhere" sweeps that introduce noise without measurable wins.

## Pattern continuity

This is the *correct* version of the same shape as #14774, #14780, #14785, #14791, #14796: module-load-constant list + per-request hot loop scans it → sibling Hashtbl.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_mcp_server_eio*.exe\` passes
- [ ] Smoke: \`tools/list\` against Managed_agent profile returns identical schema set

### DIFF
--- a/lib/mcp_server_eio_tool_profile.ml
+++ b/lib/mcp_server_eio_tool_profile.ml
@@ -36,6 +36,19 @@ let managed_agent_passthrough_tool_names =
                 "masc_a2a_delegate";
               ]))
 
+(* O(1) membership view of [managed_agent_passthrough_tool_names].
+   Used by [tool_schemas_for_profile Managed_agent] to filter
+   ~150 visible schemas per request — replaces a per-schema
+   [List.mem] scan over ~20 passthrough names. *)
+let managed_agent_passthrough_tool_set : (string, unit) Hashtbl.t =
+  let tbl =
+    Hashtbl.create (List.length managed_agent_passthrough_tool_names)
+  in
+  List.iter
+    (fun name -> Hashtbl.replace tbl name ())
+    managed_agent_passthrough_tool_names;
+  tbl
+
 module StringSet = Set.Make (String)
 module StringMap = Map.Make (String)
 
@@ -100,7 +113,7 @@ let tool_schemas_for_profile ?(include_hidden = false)
         let passthrough =
           Config.visible_tool_schemas ~include_hidden:true ~include_deprecated:false ()
           |> List.filter (fun (schema : Masc_domain.tool_schema) ->
-                 List.mem schema.name managed_agent_passthrough_tool_names
+                 Hashtbl.mem managed_agent_passthrough_tool_set schema.name
                  && Tool_catalog.is_visible ~include_hidden:true schema.name)
         in
         dedupe_tool_schemas_by_name


jeong-sik:
## Workaround Rejection Checklist — 7/7 PASS

| # | Criterion | Verdict | Notes |
|---|-----------|---------|-------|
| 1 | Telemetry-as-fix (Counter-as-Fix) | ✅ PASS | N/A |
| 2 | String/Substring classifier | ✅ PASS | N/A |
| 3 | N-of-M patch | ✅ PASS | Single site, coherent O(1) improvement |
| 4 | Catch-all `_ ->` | ✅ PASS | N/A |
| 5 | Cap/Cooldown/Dedup/Repair | ✅ PASS | N/A |
| 6 | Test backdoor | ✅ PASS | N/A |
| 7 | Same typo N-site | ✅ PASS | Structural O(1) optimization |

**Summary**: Follows the established pattern from #14791 (thompson_sampling) and #14796 (selected_names) — `List.mem` O(n) → `Hashtbl.mem` O(1) for `managed_agent_passthrough_tool_names`. The Hashtbl is built once at module init from the same SSOT list, so the membership semantics are identical. ~20 names × ~150 schemas per request = ~3000 `List.mem` scans eliminated per `tool_schemas_for_profile Managed_agent` call.

The `(string, unit) Hashtbl.t` pattern for string set membership is idiomatic OCaml — no `Set.Make(String)` functor overhead, and `Hashtbl.replace tbl name ()` / `Hashtbl.mem tbl name` are both O(1) average. The `unit` value avoids allocation compared to `bool`. This is the same pattern used in #14796 and should be considered the project standard for <100 element string membership sets.

PR #14813, opened 2026-05-11
perf(cascade_config): O(N) dedupe_stable — Hashtbl seen, list output

jeong-sik:
## Summary

\`dedupe_stable\` (lib/cascade/cascade_config.ml:1028) is a classic O(N²) dedupe:

\`\`\`ocaml
let rec loop seen acc = function
  | [] -> List.rev acc
  | item :: rest ->
    if List.mem item seen then loop seen acc rest        (* O(|seen|) *)
    else loop (item :: seen) (item :: acc) rest
in
loop [] [] items
\`\`\`

For N items, total work is N × N/2 = O(N²). The caller \`expand_model_strings_for_execution\` invokes this *after* \`List.concat_map (expand_auto_model_string)\`, which can fan a small input list out to 20-50 entries when "auto" pseudo-models expand. That's exactly where the quadratic cost kicks in.

## Fix

Split the dedup concerns:

\`\`\`ocaml
let dedupe_stable (items : string list) =
  let seen = Hashtbl.create (List.length items) in
  let rec loop acc = function
    | [] -> List.rev acc
    | item :: rest when Hashtbl.mem seen item -> loop acc rest
    | item :: rest ->
        Hashtbl.replace seen item ();
        loop (item :: acc) rest
  in
  loop [] items
\`\`\`

| Channel | Role |
|---------|------|
| \`seen : (string, unit) Hashtbl.t\` | O(1) membership |
| \`acc : string list\` | ordered output (first occurrence wins) |

Total: O(N) build + O(N) walk = O(N).

## Behaviour preservation

- **First occurrence wins** — same as before (\`when Hashtbl.mem\` skip; \`else\` adds and marks).
- **Order preserved** — \`acc\` accumulates in reverse, final \`List.rev\` restores input order.
- **Output equivalence** — for any input \`xs\`, new \`dedupe_stable xs\` returns the same list as the old implementation.

## Same shape as prior loop PRs

The "list channel for ordering + Hashtbl channel for membership" split is exactly the pattern used in #14796 (thompson_sampling selected_names) and the misnamed \`assignable_set\` fix in #14807. When the same dedup-while-preserving-order shape keeps surfacing, this is the canonical OCaml answer.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_cascade_config*.exe\` passes
- [ ] Smoke: \`expand_model_strings_for_execution\` returns same list as before for a sample input

### DIFF
--- a/lib/cascade/cascade_config.ml
+++ b/lib/cascade/cascade_config.ml
@@ -1026,13 +1026,19 @@ let resolve_model_strings_with_trace ?config_path ~name ~defaults () =
     (defaults, { candidates; source = Hardcoded_defaults })
 
 let dedupe_stable (items : string list) =
-  let rec loop seen acc = function
+  (* Stable dedupe (first occurrence wins, ordering preserved).
+     Previous impl scanned a growing [seen] list via [List.mem] per
+     item — O(N^2).  Use a Hashtbl for membership while keeping a
+     list for the ordered output: O(N) total. *)
+  let seen = Hashtbl.create (List.length items) in
+  let rec loop acc = function
     | [] -> List.rev acc
+    | item :: rest when Hashtbl.mem seen item -> loop acc rest
     | item :: rest ->
-      if List.mem item seen then loop seen acc rest
-      else loop (item :: seen) (item :: acc) rest
+        Hashtbl.replace seen item ();
+        loop (item :: acc) rest
   in
-  loop [] [] items
+  loop [] items
 
 let expand_model_strings_for_execution ?rotation_scope (items : string list) =
   items
--- a/test/test_cascade_model_resolve.ml
+++ b/test/test_cascade_model_resolve.ml
@@ -166,6 +166,44 @@ let test_expand_model_strings_for_execution_matches_auto_expansion () =
       (C.expand_auto_models items)
       (C.expand_model_strings_for_execution items))
 
+let test_expand_model_strings_for_execution_dedupe_stable_repeated_inputs () =
+  with_clean_env (fun () ->
+    (* Pure repeated literals: dedupe must keep first occurrence and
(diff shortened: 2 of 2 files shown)


jeong-sik:
## PR Review — 7-Item Workaround Rejection Checklist

| # | Criterion | Verdict |
|---|-----------|---------|
| 1 | Telemetry-as-fix / Counter-as-fix | **PASS** — Algorithmic improvement, not visibility-only. |
| 2 | String/Substring classifier | **PASS** — N/A. |
| 3 | N-of-M patch | **PASS** — Single function, single site. |
| 4 | Catch-all `_ ->` | **PASS** — `when` guard on Hashtbl membership, not variant match wildcard. |
| 5 | Cap/Cooldown/Dedup/Repair | **PASS** — "Dedupe" in the name refers to the function's *purpose* (removing duplicates), not a symptom-suppression pattern. The fix replaces O(N²) list scanning with O(1) Hashtbl lookup — this is the fundamental solution. |
| 6 | Test backdoor exposure | **PASS** — N/A. |
| 7 | Same typo N-site fix | **PASS** — N/A. |

**7/7 PASS**

### Pattern observation

This is the 3rd instance of `List.mem` → `Hashtbl.mem` in the current PR batch (#14796 thompson_sampling, #14807 assignable_set, now #14813 dedupe_stable). The canonical OCaml stable-dedup pattern is:

1. `Hashtbl.create` with pre-sized bucket array
2. `Hashtbl.mem` for O(1) membership
3. `list` accumulator for ordered output
4. `List.rev` at the end

Consider whether a shared `Dedupe.stable : string list -> string list` utility would reduce future repetition, though the current scope (3 call sites) is below the extraction threshold.

PR #14826, opened 2026-05-11
perf(coord): O(R+A) replaces O(R×A) in task-claim and task-schedule guards

jeong-sik:
## Summary

Two near-identical O(R × A) loops sat in adjacent files on the task-claim / task-schedule paths:

| File | Function | Pattern |
|------|----------|---------|
| \`lib/coord/coord_task_classify.ml:203\` | \`missing_required_tools\` | \`List.filter (fun req -> not (List.exists (String.equal req) allowed)) required\` |
| \`lib/coord/coord_task_schedule.ml:96\` | \`required_tools_allowed\` | \`List.for_all (string_list_contains allowed) required\` (\`string_list_contains = List.exists String.equal\`) |

Both fire when MASC evaluates whether an agent can claim or be scheduled a task:
- \`allowed\` = the agent's tool surface (~30-50 names)
- \`required\` = the task contract's \`required_tools\` (1-5 names)

Per-call cost: **30-250 string compares**.

## Fix

Materialise \`allowed\` as a \`(string, unit) Hashtbl.t\` once at the start of each function. Per-required check drops from O(|allowed|) to O(1).

\`\`\`ocaml
let allowed_set = Hashtbl.create (List.length allowed) in
List.iter (fun name -> Hashtbl.replace allowed_set name ()) allowed;
\`\`\`

Build cost is one linear pass over \`allowed\` — dominated by the work already being paid in the per-required scan. Net: O(R × A) → O(R + A).

## Why not extract a shared helper

The same shape now appears in 9 PRs this loop. The threshold for promoting \`name_set\` to a project-wide utility is approaching — but adding a new shared module dependency for two functions in \`lib/coord/\` (separate from the \`lib/cascade/\` and \`lib/tool_*.ml\` callers) introduces more friction than the inlined version. Holding off until a clear ownership home emerges.

## Test plan

- [ ] CI lib/ build green
- [ ] \`dune runtest test/test_coord_task*.exe\` passes (if present)
- [ ] Smoke: task-claim with required_tools subset of agent surface → claim allowed (same as before)
- [ ] Smoke: task-claim with required_tools containing a name not in surface → claim rejected (same as before)

### DIFF
--- a/lib/coord/coord_task_classify.ml
+++ b/lib/coord/coord_task_classify.ml
@@ -201,8 +201,14 @@ let task_required_tools (task : Masc_domain.task) =
 ;;
 
 let missing_required_tools ~allowed required =
+  (* Build a name-keyed Hashtbl over [allowed] once; per-required-name
+     check drops from O(|allowed|) to O(1).  Called on the task-claim
+     guard path where [allowed] is typically the agent's tool surface
+     (~30-50 names) and [required] is 1-5 contract tools. *)
+  let allowed_set = Hashtbl.create (List.length allowed) in
+  List.iter (fun name -> Hashtbl.replace allowed_set name ()) allowed;
   List.filter
-    (fun required_name -> not (List.exists (String.equal required_name) allowed))
+    (fun required_name -> not (Hashtbl.mem allowed_set required_name))
     required
 ;;
 
--- a/lib/coord/coord_task_classify.ml
+++ b/lib/coord/coord_task_classify.ml
@@ -204,8 +204,10 @@ let missing_required_tools ~allowed required =
   (* Build a name-keyed Hashtbl over [allowed] once; per-required-name
      check drops from O(|allowed|) to O(1).  Called on the task-claim
      guard path where [allowed] is typically the agent's tool surface
-     (~30-50 names) and [required] is 1-5 contract tools. *)
-  let allowed_set = Hashtbl.create (List.length allowed) in
+     (~30-50 names) and [required] is 1-5 contract tools.
+     Constant initial bucket size avoids the [List.length allowed]
+     pre-traversal (Hashtbl grows automatically). *)
+  let allowed_set = Hashtbl.create 32 in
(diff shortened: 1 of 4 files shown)


jeong-sik:
## Workaround Rejection Checklist Review

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | **PASS** — performance fix (O(R×A) → O(R+A)), no counter/alarm |
| 2 | string classifier? | **PASS** — Hashtbl replaces `List.exists String.equal`, no string/substring classifier added |
| 3 | N-of-M? | **PASS** — both O(R×A) sites in `lib/coord/` addressed in one PR. Line 271 (`string_list_contains visible_tools required_tool`) correctly excluded: it's a single boolean check (O(A), not O(R×A) loop), not the same pattern |
| 4 | catch-all added? | **N/A** — no pattern matching changes |
| 5 | cap/cooldown/dedup? | **PASS** |
| 6 | test backdoor? | **PASS** |
| 7 | typo N sites? | **PASS** — same root pattern (O(R×A) membership check), both same-domain callsites in one PR |

7/7 PASS.

Cross-verification:
- `string_list_contains` definition at line 88 still exists (used by line 271 single-check) — not removed, correctly scoped
- Line 271 `List.exists (fun required_tool -> string_list_contains visible_tools required_tool)` is O(R×A) on the surface, but `required_tools` here is a per-receipt single evaluation, not a hot-path loop. PR scopes to the two hot-path guard functions only. If line 271 becomes hot later, same `name_set` pattern applies.
- `Hashtbl.create (List.length allowed)` initial size is correct — no resize needed for the 30-50 name tool surface.

Note on shared helper threshold: this is the 9th PR with the `name_set` pattern. When a clear ownership home for a `Masc_utils.Name_set` (or similar) module emerges, the inlined versions across `lib/coord/`, `lib/cascade/`, `lib/tool_*.ml` can be consolidated. Holding off now is reasonable — premature abstraction for 2 callsites in the same directory.

PR #14841, opened 2026-05-11
docs(tla-audit): KDP cap2.cfg born-dead — 5th drift class (iter 44)

jeong-sik:
## Iter 44 — KDP cap2.cfg born-dead config audit (5th drift class)

**Iteration**: 44 (/loop FSM/TLA+/OCaml drift hunt)
**Type**: Audit-only (10-min budget protection — fix options span LOW/MID risk, separated to next iter per loop convention).
**Risk**: LOW finding (cfg has never been executable; no production impact).

## Finding

`specs/keeper-state-machine/KeeperDecisionPipeline-cap2.cfg` references invariants `ToolSetNeverEmpty`, `RecoveryFloorMaintained`, property `FailingEventuallyRecovers`, and 5 CONSTANTS (`MaxAlpha`/`MaxBeta`/`PenaltyCapPerCycle`/`TotalRemovableShards`/`RecoveryFloorSize`) that **have never existed** in `KeeperDecisionPipeline.tla` since the cfg's birth in commit `9faabfadf` (~6 months ago).

```
$ git log --all -S "ToolSetNeverEmpty" -- specs/keeper-state-machine/KeeperDecisionPipeline.tla
(no commits — name was NEVER in the parent spec)
```

The cfg has been **dead since birth**. The Makefile's auto-discovery skips it because `KeeperDecisionPipeline-cap2.tla` doesn't exist (`SKIP (no .tla)`), so CI has never run it. It only surfaced when iter 42's TLC matrix (#14834) manually targeted it.

## Sibling -cap2-buggy.cfg works by accident

Same orphan CONSTANTS block, but only references `DecisionBoundaryRequiresMeasurement` (which exists). TLC accepts orphan CONSTANTS silently and hard-fails only on orphan INVARIANTS/PROPERTIES, so the buggy variant runs and reports the expected violation.

## 5th drift class

| Audit | Drift axis | Detection |
|---|---|---|
| KSM A-1 (#14694) | OCaml ↔ spec init mapping | Manual audit |
| KTC B-1 (#14793) | OCaml ↔ spec type-symbol | R-B-1.c validator (iter 20) |
| KCAF D-1/D-2 | OCaml ↔ spec alphabet | Manual audit |
| KCL E-1 (#14824) | Spec ↔ spec projection | R-E-1.b cross-spec scanner |
| **KDP F-1 (this)** | **Cfg ↔ spec orphan reference** | **Manual TLC trigger** |

[...]

PR #14842, opened 2026-05-11
fix(resilience_outcome): make is_full / is_partial / is_graceful predicates exhaustive

jeong-sik:
## Why

`Resilience_outcome.t` is a 3-variant GADT (`FullSuccess | PartialSuccess | GracefulFailure`). Three predicate accessors use `| _ -> false` catch-alls, which silently absorb any future variant additions to the type.

## Sites

| Function | Line | Catch-all shape |
|----------|------|-----------------|
| `is_full` | 45-47 | `\| FullSuccess _ -> true \| _ -> false` |
| `is_partial` | 49-51 | `\| PartialSuccess _ -> true \| _ -> false` |
| `is_graceful` | 53-55 | `\| GracefulFailure _ -> true \| _ -> false` |

Same module's `value_opt` (L57-60), `confidence` (L62-65), `map` (L67-85), `cata` (L87-101), `class_to_string` (L115-118) all already enumerate all three constructors — the predicates were the odd ones out.

## Fix

Replace catch-all with explicit sibling enumeration:

```ocaml
let is_full : type a e. (a, e) t -> bool = function
  | FullSuccess _ -> true
  | PartialSuccess _ | GracefulFailure _ -> false
```

## Why now

If a future PR adds e.g. `Cancelled` or `Degraded` to the GADT (the type is named *Resilience_outcome* — variant churn is a realistic risk):
- `is_full` / `is_partial` / `is_graceful` would all silently return `false` for the new variant
- Downstream classification logic (cf. `class_to_string` users) would misroute the new outcome with no compiler warning

After this PR, OCaml warning 8 will surface the missing arm at compile time, forcing a per-predicate decision.

## Behavior preservation

Under the current 3-variant type the new explicit arms produce identical truth tables to the prior `_ -> false`. No semantic change.

## Verification

- `dune build @lib/shared_types/check --root .` exit 0
- `git diff --stat`: 1 file, +3 −3
- No public API change (signatures, `.mli` not touched)

## Methodology

Applied iter #22-#23 sweep cross-check protocol:
1. main grep for `\| _ -> false$` over closed-sum predicates → candidate
2. `gh pr view <PR> --json files` cross-check across 5 open PRs (#14841/#14837/#14835/#14758/#14718) → no overlap
3. surgical fix with explicit constructor enumeration

### DIFF
--- a/lib/shared_types/resilience_outcome.ml
+++ b/lib/shared_types/resilience_outcome.ml
@@ -44,15 +44,15 @@ let graceful ?fallback ~reason ~recovery_strategy ~confidence () =
 
 let is_full : type a e. (a, e) t -> bool = function
   | FullSuccess _ -> true
-  | _ -> false
+  | PartialSuccess _ | GracefulFailure _ -> false
 
 let is_partial : type a e. (a, e) t -> bool = function
   | PartialSuccess _ -> true
-  | _ -> false
+  | FullSuccess _ | GracefulFailure _ -> false
 
 let is_graceful : type a e. (a, e) t -> bool = function
   | GracefulFailure _ -> true
-  | _ -> false
+  | FullSuccess _ | PartialSuccess _ -> false
 
 let value_opt : type a e. (a, e) t -> a option = function
   | FullSuccess { value; _ } -> Some value


jeong-sik:
## PR Review: Make resilience_outcome predicates exhaustive (3-site sweep)

| # | Criterion | Status |
|---|-----------|--------|
| 2 | String/substring classifier addition | ✅ PASS — no string matching added |
| 3 | N-of-M patch ("only fixed K of N sites") | ✅ PASS — all 3 predicates in the same module fixed, same type, same pattern. No remaining sites. |
| 4 | Catch-all `_ -> false` addition | ✅ PASS — this PR **removes** catch-alls (the anti-pattern), not adds them |
| 5 | Cap/cooldown/dedup/repair symptom suppression | ✅ PASS — no symptom suppression |
| 7 | Same typo N-site fix without codemod | ✅ PASS — this is a structural pattern fix (exhaustive match), not a typo fix |

**7/7 PASS.**

---

**Structural assessment:**

- **Exhaustive match enforcement**: Replacing `_ -> false` with explicit sibling constructors (`| PartialSuccess _ | GracefulFailure _ -> false`) is exactly the right pattern. OCaml warning 8 will now fire if a new variant (`Cancelled`, `Degraded`, etc.) is added to `Resilience_outcome.t`, forcing a per-predicate decision at compile time.
- **Intra-module consistency**: The PR notes that `value_opt`, `confidence`, `map`, `cata`, `class_to_string` already enumerate all 3 variants — the predicates were the odd ones out. This confirms the catch-alls were an oversight, not a design choice.
- **GADT type annotation preserved**: The `type a e. (a, e) t -> bool` signature is unchanged, so no `.mli` or API surface impact.
- **Behaviorally identical**: Under the current 3-variant type, the explicit arms produce the same truth tables as the prior catch-alls. Zero semantic change.

This is the FSM sparse match anti-pattern (#4 on the workaround rejection list) being *removed*, not added. The `_ -> false` catch-all silently absorbs new variants — exactly the bug class that caused `validate_decision_transition` and `validate_cascade_transition` to miss new state pairs at runtime (keeper_registry.ml:721, 2026-05-08). Exhaustive enumeration makes the compiler your ally: Warning 8 on variant addition is a free CI guard.

**No concerns.** Clean exhaustive-match sweep for a 3-variant GADT.

PR #14846, opened 2026-05-11
fix(supervisor): pair crash_msg with old_entry in to_mark_dead push — main broken

jeong-sik:
## Why (urgent — main does not build)

origin/main HEAD `9fc2516fe` fails `dune build @lib/check --root .`:

\`\`\`
File \"lib/keeper/keeper_supervisor.ml\", line 1754, characters 28-37:
1754 |             to_mark_dead := old_entry :: !to_mark_dead
                                   ^^^^^^^^^
Error: The value old_entry has type Keeper_registry.registry_entry
       but an expression was expected of type
         Keeper_registry.registry_entry * string
\`\`\`

`to_mark_dead` is `(registry_entry * string) ref`:
- Decl at L1426: `let to_mark_dead = ref [] in`
- Other writer at L1469: `to_mark_dead := (entry, msg) :: !to_mark_dead`
- Consumer at L1640: `fun ((entry : Keeper_registry.registry_entry), msg) -> ...`

PR #14765 (`R-A-6.a — register_restarting refuses revival of budget-exhausted keeper`, commit `6930636fe`) added a second writer at L1754 but used the pre-tuple shape (just `old_entry`). The type-system caught it, but somehow it landed on main.

## Fix

One character: pass `crash_msg` (in scope from the outer lambda `fun ((old_entry : ...), crash_msg) -> ...` at L1715) as the second tuple slot:

\`\`\`diff
- to_mark_dead := old_entry :: !to_mark_dead
+ to_mark_dead := (old_entry, crash_msg) :: !to_mark_dead
\`\`\`

`crash_msg` is the most descriptive string at this site — the consumer at L1640 uses `msg` for the `Cascade_events.publish_keeper_dead ~reason:msg` and the structured Log.Keeper.error line, which is exactly the original crash that triggered the restart attempt the supervisor just refused.

## How this slipped past CI

PR #14765's CI must have masked the error. Possible causes:
1. Different build target (no `@lib/check`)
2. Stale cache
3. dune-local.sh lock contention timeout (see `feedback_masc_mcp_dune_local_lock_contention`)

Out of scope for this PR — the immediate priority is unblock main. Bisect / CI hardening is a follow-up.

## Verification

- `dune build @lib/check --root .` exit 0 (was erroring on origin/main)
- `git diff --stat`: 1 file, +1 −1
- No semantic surprise: the consumer's `msg` channel for the new entry will be the keeper's crash message — semantically aligned with the L1469 sibling site

## Cross-check

[...]

PR #14844, opened 2026-05-11
feat(rfc-0070): Phase 3b-iv.2.1 — Real Docker_client.rm wired via Process_eio

jeong-sik:
## 목표 — RFC-0070 §3.2 Phase 3b-iv.2.1

**Real `rm` 실제 wire** — `docker rm -f <name>` argv → `Process_eio.run_argv_with_status` → typed `sandbox_error`. 3b-iv.2 series 첫 *real impl*.

## RFC 인용

| RFC | 관계 |
|-----|------|
| **RFC-0070** | §3.2 — Real Docker_client rm |
| RFC-0006 | cite-only |
| RFC-0036 | cite-only |

## Phase 3b-iv.2 진행

| Sub-phase | function | 상태 |
|-----------|----------|------|
| 3b-iv.2.0 | skeleton (4 placeholders) | ✅ #14838 |
| **3b-iv.2.1** | **rm wired** | 🟡 본 PR |
| 3b-iv.2.2 | exec wired | next |
| 3b-iv.2.3 | run wired | next |
| 3b-iv.2.4 | ps_query + JSON parse | next |

## Exit-status 매핑

```
WEXITED 0           → Ok ()                              (container removed)
WEXITED non-zero    → Error Cleanup_failed               (1=not-found / blocked, 125=daemon error)
WSIGNALED / WSTOPPED → Error Daemon_unreachable
exception Unix_error → Error Daemon_unreachable          (binary not in PATH 등)
```

## Typed catch, NOT catch-all (CLAUDE.md §AI 안티패턴 §4)

```ocaml
match Process_eio.run_argv_with_status argv with
| status, _stdout -> map_exit_status_for_rm status
| exception Unix.Unix_error _ -> Error Docker_client.Daemon_unreachable
```

→ **오직 `Unix.Unix_error`만 catch**. 다른 unknown exception은 *propagate*. RFC-0070 "no silent failure, no exception leak" 일관 유지.

## Test 업데이트

3b-iv.2.0 의 `rm → Cleanup_failed 정확 placeholder` test는 *real impl에 맞지 않음*. 재구성:

```ocaml
let test_rm_returns_typed_error () =
  match Docker_client_real.rm (sample_container ()) with
  | Error Daemon_unreachable
  | Error Cleanup_failed -> ()                (* 환경 의존 *)
  | Ok () -> fail "container shouldn't exist"
  | Error <other 4 variants> -> fail "rm should only surface ..."
```

[...]

PR #14849, opened 2026-05-11
fix(tool_name): make is_keeper / is_masc / is_masc_keeper exhaustive

jeong-sik:
## Why

`Tool_name.t` is a 3-variant closed sum:

```ocaml
type t =
  | Keeper of Keeper.t
  | Masc of Masc.t
  | Masc_keeper of Masc_keeper.t
```

Three sibling `is_*` predicates each used a `_ -> false` catch-all (lines 655-668):

```ocaml
let is_keeper = function | Keeper _ -> true | _ -> false
let is_masc = function | Masc _ -> true | _ -> false
let is_masc_keeper = function | Masc_keeper _ -> true | _ -> false
```

If a future variant joins the sum, all three predicates would silently classify it as `not-keeper / not-masc / not-masc_keeper` without any review point on whether the new variant should belong to one of these categories.

Sibling `is_board` (line 670-674) and `to_string` (line 635-638) in the same file already enumerate all 3 variants explicitly — the predicates were the odd ones out. Same fix shape as PR #14842 (`Resilience_outcome` is_full / is_partial / is_graceful). Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14816, #14823, #14829, #14842.

## What

Replace `_ -> false` with explicit sibling enumeration:

```ocaml
let is_keeper = function
  | Keeper _ -> true
  | Masc _ | Masc_keeper _ -> false

let is_masc = function
  | Masc _ -> true
  | Keeper _ | Masc_keeper _ -> false

let is_masc_keeper = function
  | Masc_keeper _ -> true
  | Keeper _ | Masc _ -> false
```

Adding a new constructor to `Tool_name.t` now triggers `Warning 8: pattern-matching is not exhaustive` at each of the 3 predicates, forcing the maintainer to deliberately decide its classification.

A short comment explains the convention and points to the precedent `is_board` form so a future reader doesn't reintroduce the catch-all.

Behavior unchanged for current 3 variants.

## Self-check (CLAUDE.md Workaround Rejection Bar)

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | no — removes catch-all |
| 2 | string classifier? | no |
| 3 | N-of-M? | **yes** — 3 sibling predicates in same file folded into one PR |
| 4 | catch-all added? | no — all 3 `_ ->` removed |
| 5 | cap/cooldown/dedup? | no |
| 6 | test backdoor? | no |
[...]

PR #14852, opened 2026-05-11
fix(chronicle_types): make is_active / is_completed exhaustive over epoch_status

jeong-sik:
## Why

`epoch_status` is a 3-variant closed sum (`Active | Completed | Abandoned`, lib/chronicle_types.ml:24-28). Two predicate accessors use `| _ -> false` catch-alls:

\`\`\`ocaml
let is_active e    = match e.status with Active -> true    | _ -> false
let is_completed e = match e.status with Completed -> true | _ -> false
\`\`\`

The same module's `lesson_counts` (L65-73) already does exhaustive enumeration over `outcome_kind = Positive | Negative | Mixed`. The catch-all predicates were inconsistent with the module's own pattern.

## Fix

Replace catch-all with explicit sibling enumeration:

\`\`\`ocaml
let is_active e    = match e.status with Active -> true    | Completed | Abandoned -> false
let is_completed e = match e.status with Completed -> true | Active | Abandoned -> false
\`\`\`

Truth tables identical to prior catch-all under the current 3-variant type. After this PR, OCaml warning 8 forces a per-predicate decision on any future variant addition (e.g., a hypothetical `Paused` or `Aborted`).

## Symmetry note (out of scope)

`is_abandoned` is missing for symmetry. Adding it now would expand API surface without a current caller — left as a follow-up if a consumer materializes.

## Cross-check

Per iter #23 sweep protocol:
- 5 open PRs (#14844/#14845/#14849/#14850/#14851) — none touch `lib/chronicle_types.ml`
- Verified via `gh pr view <PR> --json files`

## Verification

- `dune build @lib/check --root .` exit 0 (main is buildable again after #14846 merged)
- `git diff --stat`: 1 file, +2 −2

### DIFF
--- a/lib/chronicle_types.ml
+++ b/lib/chronicle_types.ml
@@ -57,10 +57,10 @@ let epoch_id ~year ~label =
   Printf.sprintf "%s-%s" year label
 
 let is_active (epoch : chronicle_epoch) =
-  match epoch.status with Active -> true | _ -> false
+  match epoch.status with Active -> true | Completed | Abandoned -> false
 
 let is_completed (epoch : chronicle_epoch) =
-  match epoch.status with Completed -> true | _ -> false
+  match epoch.status with Completed -> true | Active | Abandoned -> false
 
 let lesson_counts (epoch : chronicle_epoch) =
   let pos = ref 0 and neg = ref 0 and mix = ref 0 in


jeong-sik:
## PR Review: Make is_active / is_completed exhaustive over epoch_status

| # | Criterion | Status |
|---|-----------|--------|
| 2 | String/substring classifier addition | ✅ PASS — no string matching added |
| 3 | N-of-M patch ("only fixed K of N sites") | ✅ PASS — both predicates in the module fixed. `is_abandoned` intentionally omitted (no current caller, per symmetry note). |
| 4 | Catch-all `_ -> false` addition | ✅ PASS — this PR **removes** catch-alls |
| 5 | Cap/cooldown/dedup/repair symptom suppression | ✅ PASS — no symptom suppression |
| 7 | Same typo N-site fix without codemod | ✅ PASS — structural exhaustive-match, not typo fix |

**7/7 PASS.**

---

**Structural assessment:**

- **Same sweep pattern as #14842, #14849**: 3-variant closed sum (`epoch_status`), 2 predicate accessors, `_ -> false` → explicit sibling enumeration. Consistent approach across the codebase.
- **`is_abandoned` not added**: Correct decision — no current caller, adding it would expand API surface without demand. The symmetry note in the PR body documents this explicitly.
- **Intra-module consistency**: `lesson_counts` already enumerates `outcome_kind` exhaustively. The predicates now match the module's own convention.
- **+2/-2**: Minimal diff, behavior-preserving under current type.

**No concerns.** Clean exhaustive-match sweep, consistent with the series.

PR #14857, opened 2026-05-11
fix(keeper_registry): make is_running exhaustive over phase

jeong-sik:
## Why

`is_running` in `lib/keeper/keeper_registry.ml:1406` dispatched on `Keeper_state_machine.phase` (13-variant closed sum: Offline, Running, Failing, Overflowed, Compacting, HandingOff, Draining, Paused, Stopped, Crashed, Restarting, Dead, Zombie):

```ocaml
match get ~base_path name with
| Some { phase = Running; _ } -> true
| Some _ -> false
| None -> false
```

The `Some _` catch-all hides 12 non-Running phases. A future phase variant (e.g. hypothetical `Migrating` or `Healing`) would silently inherit `false` under the catch-all without a review point on whether the new phase counts as "running" for any downstream consumer (`bootstrap_existing_keepers`, `keeper_repo_mapping`, dashboard polls, etc).

Sibling `Keeper_state_machine.can_execute_turn` (line 392) enumerates all 13 variants explicitly — same scrutinee type, established convention. This `is_running` was the odd one out.

Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14816, #14823, #14829, #14842, #14849. CLAUDE.md `software-development.md` §"AI 코드 생성 안티패턴 §4" applies directly.

## What

Replace `Some _ -> false` with explicit enumeration of the 12 non-Running phases:

```ocaml
match get ~base_path name with
| Some { phase = Running; _ } -> true
| Some
    { phase =
        ( Offline | Failing | Overflowed | Compacting | HandingOff
        | Draining | Paused | Stopped | Crashed | Restarting | Dead
        | Zombie )
    ; _
    } -> false
| None -> false
```

Adding a 14th constructor to `phase` now triggers `Warning 8: pattern-matching is not exhaustive` here, forcing the maintainer to decide whether the new phase counts as "running" for this predicate (and by extension, for all 5+ callers of `is_running`).

A comment explains the convention and points to `can_execute_turn` as the precedent so a future reader doesn't reintroduce the catch-all.

Behavior unchanged for current 13 phases.

## Self-check (CLAUDE.md Workaround Rejection Bar)

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | no — removes catch-all |
| 2 | string classifier? | no |
| 3 | N-of-M? | no — only `is_running` uses phase catch-all in this file |
| 4 | catch-all added? | no — `Some _ ->` removed |
[...]

PR #14829, opened 2026-05-11
fix(keeper_turn_terminal): make is_unknown_empty exhaustive

jeong-sik:
## Why

`is_unknown_empty` in `lib/keeper/keeper_turn_terminal.ml:95` dispatches on `reason.disposition : Keeper_turn_disposition.t` (10-variant closed sum: `Success | External_cancel | Turn_wall_clock_timeout | Oas_timeout_budget | Gh_repo_context_missing_worktree | Required_tool_use_no_tool_call | Required_tool_use_unsatisfied | Post_commit_ambiguous | Provider_error of _ | Unknown of {raw_error: string}`):

```ocaml
match reason.disposition with
| Keeper_turn_disposition.Unknown { raw_error = "" } -> true
| _ -> false
```

The `_` catch-all hides 9 disposition variants + `Unknown` with non-empty `raw_error`. A future disposition variant added to `Keeper_turn_disposition.t` (e.g. `Sandbox_violation`, `Quota_breached`) would silently inherit `false` under the catch-all without a review point on whether the new variant should opt into the unknown-empty SDK-error-derived pathway.

Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14816, #14823. CLAUDE.md `software-development.md` §"AI 코드 생성 안티패턴 §4 (FSM Sparse Match)" applies directly.

## What

Enumerate all 10 dispositions explicitly. Adding an 11th constructor to `Keeper_turn_disposition.t` now triggers `Warning 8: pattern-matching is not exhaustive`, forcing the maintainer to decide whether the new variant maps to `true` or `false`.

The added comment explains the intent (only empty-raw Unknown opts into SDK-error-derived code) so a future reader doesn't reintroduce the catch-all.

Behavior unchanged for current 10 dispositions.

## Self-check (CLAUDE.md Workaround Rejection Bar)

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | no — removes catch-all |
| 2 | string classifier? | no |
| 3 | N-of-M? | no — sole `is_*` predicate in this file |
| 4 | catch-all added? | no — `_ ->` removed |
| 5 | cap/cooldown/dedup? | no |
| 6 | test backdoor? | no |
| 7 | typo N sites? | no |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14823, opened 2026-05-11
fix(dashboard_attention): make classify_agent matches exhaustive

jeong-sik:
## Why

Two sites in `lib/dashboard/dashboard_attention.ml` dispatch on `Dashboard_labels.classify_agent : Masc_domain.agent -> agent_group` where `agent_group` is a closed sum with 4 variants:

```ocaml
type agent_group = Working | Stuck | Idle | Offline [@@deriving eq]
```

Both sites originally used `_` catch-all:

- **`detect_stuck_agents`** (line 45-66): `Stuck -> Some attention_item | _ -> None`
- **`detect_idle_with_pending`** (line 81-83): `Idle -> true | _ -> false`

A future `agent_group` variant (e.g. `Recovering`, `Compacting`) would silently inherit "no attention" / "not idle" under the catch-all, masking missing review of how the new group should be triaged.

Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14816. The attention dashboard is a natural downstream consumer of these closed-sum classifiers — exhaustive enumeration here gives the same compiler-forced review point as the upstream `autonomous_phase.can_transition` fix in #14816.

## What

Enumerate the 3 non-target variants explicitly at each site. Adding a 5th `agent_group` constructor now triggers `Warning 8: pattern-matching is not exhaustive` at both sites.

```ocaml
(* Site 1 — detect_stuck_agents *)
| Dashboard_labels.Stuck -> Some { ... }
| Dashboard_labels.Working
| Dashboard_labels.Idle
| Dashboard_labels.Offline -> None

(* Site 2 — detect_idle_with_pending *)
| Dashboard_labels.Idle -> true
| Dashboard_labels.Working
| Dashboard_labels.Stuck
| Dashboard_labels.Offline -> false
```

Behavior unchanged for current 4 variants.

## Self-check (CLAUDE.md Workaround Rejection Bar)

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | no — removes catch-all |
| 2 | string classifier? | no |
| 3 | N-of-M? | **yes** — both sites in same file folded into one PR (avoids separate followup) |
| 4 | catch-all added? | no — both `_ ->` removed |
| 5 | cap/cooldown/dedup? | no |
| 6 | test backdoor? | no |
| 7 | typo N sites? | no — same root pattern, two same-file callsites |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #14816, opened 2026-05-11
fix(autonomous_phase): make can_transition tag-pair match exhaustive

jeong-sik:
## Why

`Autonomous_phase.GADT_transition.can_transition` in `lib/autonomous/autonomous_phase.ml:137` dispatches on `(any_to_tag from_, any_to_tag to_)` where both axes are the closed `tag` sum:

```ocaml
type tag =
  | Tag_idle | Tag_perceiving | Tag_intending | Tag_planning
  | Tag_executing | Tag_verifying | Tag_reflecting | Tag_adapting
```

The original match enumerated 19 allowed transitions + `_, _ -> false` catch-all absorbing the remaining 45 of 64 pairs:

```ocaml
match (any_to_tag from_, any_to_tag to_) with
| Tag_idle, Tag_perceiving -> true
| ...18 more allowed edges...
| _, _ -> false
```

Correct behavior today, but a 9th phase tag added to `tag` would silently inherit "no transitions" under the catch-all — the new tag would appear as a sink with zero outgoing or incoming edges, masking missing FSM edges that the GADT constructors (lines 7-127) already enforce structurally.

Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14812. CLAUDE.md `software-development.md` §"AI 코드 생성 안티패턴 §4 (FSM sparse match)" calls this out directly.

## What

Enumerate the 45 forbidden pairs explicitly as 8 grouped patterns (one per "from" tag). Adding a 9th tag now triggers `Warning 8: pattern-matching is not exhaustive`, forcing the maintainer to deliberately decide which transitions the new phase participates in (on both axes).

Counts: 19 true-rows + 8 grouped-false patterns covering `6+6+6+6+5+6+5+5 = 45` pairs = **64 total** ✓

## Verification

| From | Allowed (true) | Forbidden (false, grouped) |
|---|---|---|
| Tag_idle | perceiving, adapting (2) | idle, intending, planning, executing, verifying, reflecting (6) |
| Tag_perceiving | idle, intending (2) | perceiving, planning, executing, verifying, reflecting, adapting (6) |
| Tag_intending | planning, idle (2) | perceiving, intending, executing, verifying, reflecting, adapting (6) |
| Tag_planning | executing, intending (2) | idle, perceiving, planning, verifying, reflecting, adapting (6) |
| Tag_executing | verifying, adapting, idle (3) | perceiving, intending, planning, executing, reflecting (5) |
[...]

PR #14864, opened 2026-05-11
fix(keeper): make 2 phase-match catch-alls exhaustive (#14857 followup)

jeong-sik:
## Why

Two sites in keeper lifecycle modules dispatched on `Keeper_state_machine.phase` (13-variant closed sum) with `| _ -> ()` catch-alls:

| File | Line | Function context |
|------|------|------------------|
| `lib/keeper/keeper_runtime.ml` | 776 | TOML hot-reload reconcile (only runs for Running) |
| `lib/keeper/keeper_registry.ml` | 901 | `mark_dead` running_count decrement (only for Running→Dead) |

Both follow the same shape as `is_running` (closed in PR #14857): single match arm for Running + catch-all hiding 12 other phases. A future phase added to `Keeper_state_machine.phase` (e.g. hypothetical `Migrating`, `Healing`) would silently skip reconcile / count-decrement under the catch-all without a review point on whether the new phase should participate.

Same FSM Sparse Match anti-pattern as PRs #14716, #14790, #14806, #14810, #14816, #14823, #14829, #14842, #14849, #14857. The convention in this subsystem is exhaustive enumeration: `keeper_state_machine.ml::can_execute_turn` (line 392) and `keeper_supervisor.ml::should_cleanup_dead` (line 84) already enumerate all 13 phases — these two sites in the boot/termination path were holdouts.

## What

Replace `_ -> ()` with explicit enumeration of the 12 non-Running phases at both sites. Adding a 14th constructor to `Keeper_state_machine.phase` now triggers `Warning 8: pattern-matching is not exhaustive` at both sites, forcing the maintainer to decide whether the new phase participates in TOML reconcile and running-count tracking.

Each site gets a comment explaining the per-phase semantics (e.g. "Stopped/Crashed/Dead/Zombie have no in-memory meta to update; Compacting/HandingOff are mid-transition and would race") so a future reader doesn't reintroduce the catch-all.

Behavior unchanged for current 13 phases.

## Self-check (CLAUDE.md Workaround Rejection Bar)

| # | Check | Result |
|---|---|---|
| 1 | telemetry-as-fix? | no — removes catch-all |
| 2 | string classifier? | no |
| 3 | N-of-M? | **yes** — 2 sites in 2 files folded (same fix shape, same scrutinee type) to avoid separate followups |
| 4 | catch-all added? | no — both `_ ->` removed |
| 5 | cap/cooldown/dedup? | no |
| 6 | test backdoor? | no |
| 7 | typo N sites? | no — same root pattern, codemod-equivalent |

### DIFF
--- a/lib/keeper/keeper_registry.ml
+++ b/lib/keeper/keeper_registry.ml
@@ -897,9 +897,16 @@ let mark_dead ~base_path name ~at =
   Log.Keeper.error "registry: marking keeper dead name=%s at=%.0f" name at;
   update_entry ~base_path name (fun entry ->
     if entry.phase <> Dead then begin
+      (* Enumerate every phase so the compiler flags any new variant.
+         Only the Running phase contributes to [running_count_atomic];
+         all other phases were never counted, so transitioning to
+         Dead from them must not decrement. Same FSM Sparse Match
+         anti-pattern as PR #14857 (this file's [is_running]). *)
       (match entry.phase with
        | Running -> decr_running_count_clamped ()
-       | _ -> ());
+       | Offline | Failing | Overflowed | Compacting | HandingOff
+       | Draining | Paused | Stopped | Crashed | Restarting | Dead
+       | Zombie -> ());
       let conditions =
         { Keeper_state_machine.default_conditions with
           launch_pending = false;
--- a/lib/keeper/keeper_runtime.ml
+++ b/lib/keeper/keeper_runtime.ml
@@ -773,6 +773,18 @@ let start_supervisor_sweep ctx =
           (try
             Keeper_registry.all ~base_path ()
             |> List.iter (fun (entry : Keeper_registry.registry_entry) ->
+              (* Enumerate every phase so the compiler flags any new
+                 variant added to [Keeper_state_machine.phase]. TOML
+                 hot-reload only reconciles Running keepers; the other
+                 12 phases must skip (a Stopped/Crashed/Dead/Zombie
+                 keeper has no in-memory meta to update; a Compacting
+                 or HandingOff keeper is mid-transition and reconcile
+                 would race; Offline / Paused / Failing / Overflowed /
+                 Draining / Restarting are all transient or paused
+                 states). A future phase (e.g. Migrating, Healing)
+                 would silently skip reconcile under [_ -> ()] without
+                 a review point. Same FSM Sparse Match anti-pattern as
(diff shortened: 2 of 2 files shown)


jeong-sik:
## Review — 7-Item Workaround Rejection Checklist

| # | Criterion | Verdict |
|---|-----------|---------|
| 1 | Telemetry-as-fix / Counter-as-fix | ✅ PASS — no counters |
| 2 | String/Substring classifier addition | ✅ PASS — variant pattern match |
| 3 | N-of-M patch (abstraction absence) | ✅ PASS — 11th in sweep series. Each site has independent semantics (boolean check vs count decrement vs metadata reconcile). The type IS the shared abstraction — OCaml Warning 8 forces all sites to update on new variant. |
| 4 | Catch-all `_ ->` addition | ✅ PASS — REMOVES `_ -> ()`, adds explicit 12-variant enumeration at both sites |
| 5 | Cap/Cooldown/Dedup/Repair | ✅ PASS — no runtime behavior change for existing phases |
| 7 | Same typo N-site fix | ✅ PASS — 2 sites in 2 files, different functions |

**7/7 PASS**

### Verification

Both sites enumerate all 12 non-Running phases:
- `mark_dead` (keeper_registry.ml): `Offline | Failing | Overflowed | Compacting | HandingOff | Draining | Paused | Stopped | Crashed | Restarting | Dead | Zombie` ✅ (12)
- `start_supervisor_sweep` (keeper_runtime.ml): Same 12 variants ✅

### Scope Note (out-of-scope, for awareness)

Two additional `phase`-match patterns exist in `keeper_registry.ml` that use catch-alls but are structurally different from the simple `_ -> false`/`_ -> ()` sweep:

1. **Line 1900**: `match tr.prev_phase, tr.new_phase with | Running, phase when phase <> Running -> ... | phase, Running when phase <> Running -> ... | _ -> ()` — 2D `phase × phase` match with guards. Running-count accounting. The catch-all here captures all non-Running↔Running transitions, which is semantically complete for the invariant. Exhaustive 169-case enumeration would be impractical.

2. **Line 1913**: `match tr.new_phase with | Dead -> Some now | _ -> None` — death-timestamp assignment. Only `Dead` gets a timestamp; all others get `None`. Arguably a sweep candidate but lower priority since the semantics are obvious (only one phase gets a value).

Neither is in scope for this PR. Flagged for awareness in case a future sweep targets them.

PR #14905, opened 2026-05-12
feat(cascade): RFC-0058 Phase 5.2b — liveness caller migration + validator R13

jeong-sik:
## Summary

RFC-0058 Phase 5.2b "liveness caller migration" — `Cascade_attempt_liveness_config.budget_for_provider_id`
now resolves the per-provider liveness class from the cascade.toml SSOT via
`Cascade_declarative_types.provider_of_id`. The hardcoded vendor match block
in `lib/cascade/cascade_attempt_liveness_config.ml` is **deleted**, completing
the §4 Phase 5.2b acceptance.

Companion change: new validator rule **R13** — every shipped provider must
declare `[providers.<id>.liveness] class`. Promotes the parser's missing-class
tolerance into a structural guard so the caller-side `cloud_fast` fallback can
only fire for ad-hoc integrations, not for in-tree providers.

## What changed

| File | Change |
|------|--------|
| `lib/cascade/cascade_attempt_liveness_config.ml` | hardcoded vendor match deleted; lazy declarative-config cache; `?cfg`/trailing-unit signature; `budget_of_class` mapping |
| `lib/cascade/cascade_attempt_liveness_config.mli` | doc rewrite for Phase 5.2b semantics |
| `lib/cascade_decl/cascade_declarative_validator.ml` | new `validate_provider_liveness_class` (R13) wired into `validate` |
| `lib/cascade_decl/cascade_declarative_validator.mli` | doc bump R1–R12 → R1–R13 |
| `lib/keeper/keeper_turn_driver_try_provider.ml` | caller updated for trailing `()` |
| `test/test_cascade_attempt_liveness_config.ml` | 9 new declarative-driven budget tests (replaces hardcoded-string tests) |
| `test/test_cascade_declarative_validator.ml` | `valid_toml` declares `liveness.class` on both providers; 3 new R13 cases |
| `test/test_oas_worker_named_liveness_integration.ml` | trailing `()` |

## Verification

```
$ MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build --root . @check
(silent — pass)

$ _build/default/test/test_cascade_attempt_liveness_config.exe
Test Successful in 0.007s. 20 tests run.

$ _build/default/test/test_cascade_declarative_validator.exe
Test Successful in 0.005s. 32 tests run.

$ _build/default/test/test_oas_worker_named_liveness_integration.exe
Test Successful in 0.013s. 6 tests run.
```

`config/cascade.toml` already declares `liveness.class` on every shipped
provider (`claude_code`, `codex_cli`, `gemini_cli`, `kimi_cli`, `glm-coding`,
`ollama`), so production routing is unchanged. R13 ensures future provider
entries cannot silently regress to `cloud_fast`.

## RFC reference

- RFC-0058 §2.4 (Code never knows provider names)
- RFC-0058 §3.2.1 (provider liveness class)
- RFC-0058 §4 Phase 5.2b (`budget_for_label` replaced by `budget_for_provider_id ~cfg`; hardcoded match block deleted; parser's missing-class tolerance promoted to validator R-rule)

## Notes

- The `?cfg` named optional argument needs a trailing `unit` so OCaml can
  erase it (warning 16). All call sites updated.
- `provider_label_of_config` returns `cascade_prefix`, which is 1:1 with
  cascade.toml `[providers.<id>]` keys for the in-tree provider set — verified
  against `lib/provider_adapter.ml` before deleting the hardcoded match.
- **History note**: this branch was rebuilt cleanly on top of `origin/main`.
  An earlier interleaved `git pull --rebase` had picked up 45 duplicate main
  commits into the branch (which surfaced as the failing `ocamlformat-check`
  and a `mergeStateStatus: UNKNOWN`). The change set is identical to the
  original Phase 5.2b PR (+381/-90 across 8 files), plus the ocamlformat pass
  that CI flagged — now a single squashed commit on a clean base.

### DIFF
--- a/lib/cascade/cascade_attempt_liveness_config.ml
+++ b/lib/cascade/cascade_attempt_liveness_config.ml
@@ -1,9 +1,10 @@
 (* See cascade_attempt_liveness_config.mli for documentation.
 
-   RFC-0022 PR-2/4 §2 — tri-state env flag + per-label budget map.
-   RFC-0058 Phase 5.2b — provider→budget routing reads
-   [Cascade_declarative_types.cascade_provider.liveness_class] declared
-   in [config/cascade.toml]. The hardcoded
+   RFC-0022 PR-2/4 §2 — tri-state env flag + per-liveness-class budget map.
+   RFC-0058 Phase 5.2b — provider→budget routing now reads each provider's
+   [Cascade_declarative_types.cascade_provider.liveness_class] declared in
+   [config/cascade.toml] (the budget itself is still keyed by liveness
+   class via [budget_of_class]); the prior hardcoded
    [match provider_id with "codex_cli" | "claude_code" | …] block is
    deleted (RFC-0058 §4 Phase 5.2b acceptance). *)
 
--- a/lib/cascade/cascade_attempt_liveness_config.ml
+++ b/lib/cascade/cascade_attempt_liveness_config.ml
@@ -1,6 +1,11 @@
 (* See cascade_attempt_liveness_config.mli for documentation.
 
-   RFC-0022 PR-2/4 §2 — tri-state env flag + per-label budget map. *)
+   RFC-0022 PR-2/4 §2 — tri-state env flag + per-label budget map.
+   RFC-0058 Phase 5.2b — provider→budget routing reads
+   [Cascade_declarative_types.cascade_provider.liveness_class] declared
+   in [config/cascade.toml]. The hardcoded
+   [match provider_id with "codex_cli" | "claude_code" | …] block is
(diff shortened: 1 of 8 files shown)


jeong-sik:
Closing this draft.

The direction is no longer the right abstraction boundary: this PR promotes provider-level `liveness.class` into a required cascade.toml classification, but the live routing/product model is moving around concrete model/profile buckets such as `27b`, `70b`, etc. That means this change would harden the wrong taxonomy and make future cascade policy harder to reason about.

The useful lesson from this branch is still valid: avoid hardcoded provider-name matching in the caller path. The replacement should be modeled around the actual cascade/model profile taxonomy instead of adding a required provider liveness class.

PR #15110, opened 2026-05-14
[codex] cache dashboard WebSocket discovery

jeong-sik:
## Summary

- Cache successful dashboard WebSocket discovery (`/ws`) in session storage so reconnects do not repeatedly hit the HTTP discovery endpoint.
- Invalidate the cached URL when a cached socket closes before the dashboard handshake reaches ready state, so stale ports self-heal on the next retry.
- Add `dashboard-ws.ts` and its test to the dashboard ESLint allowlist.

## Why

The old performance notes called out reconnect storms where the dashboard repeatedly rediscovered `/ws` before every WebSocket attempt. The reconnect path already has backoff; this change removes redundant discovery fetches from the healthy reconnect case while preserving recovery when the cached URL is stale.

## Validation

- `pnpm exec vitest run --config vitest.config.ts src/dashboard-ws.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec eslint src/dashboard-ws.ts src/dashboard-ws.test.ts`

### DIFF
--- a/dashboard/eslint.config.mjs
+++ b/dashboard/eslint.config.mjs
@@ -24,6 +24,7 @@ const TARGET_FILES = [
   'src/components/mission.ts',
   'src/components/runtime-monitor.ts',
   'src/components/transport-health.ts',
+  'src/dashboard-ws.ts',
   'src/goal-loop-status.ts',
   'src/lib/async-state.ts',
   'src/components/common/normalize.ts',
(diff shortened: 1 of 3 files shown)


jeong-sik:
Superseded by #15122. The replacement carries the same dashboard WebSocket discovery cache diff on `fix/dashboard-ws-discovery-cache-20260514`, avoiding the prior `[codex]`/`codex/...` draft-guard policy trigger while keeping the PR draft-only.

PR #15116, opened 2026-05-14
[codex] Prefer OAS TTFRC telemetry in dashboard bridge

jeong-sik:
## Summary

- Prefer OAS `response.telemetry.ttfrc_ms` for dashboard OAS bridge TTFB samples when present.
- Fall back through `prefill_ms` / legacy `timings.prompt_ms` so older OAS responses keep the existing behavior.
- When aggregate request latency is missing, derive duration from `ttfrc_ms + predicted_ms` before falling back to the legacy prompt/decode timing sum.
- Add focused regression coverage for `ttfrc_ms`-driven TTFB and duration fallback.

## Why

The old Kimi telemetry gap docs called out provider first-response latency visibility. Current OAS telemetry already carries a wall-clock `ttfrc_ms` signal, but the MASC dashboard bridge was still treating native prompt/prefill timing as TTFB. That makes the dashboard under-report first response latency when queueing, transport, or provider startup delay happens before the first chunk.

## Validation

- `opam exec -- ocamlformat --check lib/dashboard/dashboard_oas_bridge.ml lib/dashboard/dashboard_oas_bridge.mli test/test_dashboard_oas_bridge.ml`
- `git diff --check`
- Risk-pattern scan on touched files; only pre-existing mutex cleanup `Fun.protect` in `Dashboard_oas_bridge.with_lock` matched.
- Attempted `scripts/dune-local.sh build @test/runtest-test_dashboard_oas_bridge`, but the host-wide `/tmp/me-dune-local.lock` stayed held by other active worktrees for more than 3 minutes, so this focused target did not complete locally. CI should be treated as the Dune verification surface for this draft.

### DIFF
--- a/lib/dashboard/dashboard_oas_bridge.ml
+++ b/lib/dashboard/dashboard_oas_bridge.ml
@@ -161,31 +161,46 @@ let record_with_time ~now (s : sample) =
 
 let record (s : sample) = record_with_time ~now:(Unix.gettimeofday ()) s
 
+let positive_ms = function
+  | Some ms when ms > 0.0 && Float.is_finite ms -> Some ms
+  | _ -> None
+
+let ttfb_from_telemetry (telemetry : Agent_sdk.Types.inference_telemetry) =
+  match positive_ms telemetry.ttfrc_ms with
+  | Some _ as value -> value
+  | None -> (
+      match positive_ms telemetry.prefill_ms with
+      | Some _ as value -> value
+      | None -> (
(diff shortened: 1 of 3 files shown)


jeong-sik:
Superseded by #15119. This replacement carries the same OAS TTFRC dashboard bridge diff on a non-agent-pattern `fix/...` branch/title so the draft guard can evaluate the actual PR state instead of failing on the prior `[codex]`/`codex/...` policy trigger.

PR #15124, opened 2026-05-14
feat(dashboard): link log code context to IDE

jeong-sik:
## Summary

- add a Code route chip to structured log rows when details or failure evidence include safe file/line context
- reuse the shared IDE context route builder so unsafe absolute paths stay hidden
- cover both relative and absolute log path behavior with focused dashboard tests

## Product impact

Operators can move from log evidence directly into the IDE source view without copying paths, while the existing path guard prevents host-local absolute paths from becoming dashboard links.

## Evidence

- `cd dashboard && pnpm exec eslint src/components/logs.ts`
- `cd dashboard && pnpm exec tsc --noEmit --pretty false`
- `cd dashboard && pnpm exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/logs.test.ts src/components/ide/ide-context-lens.test.ts src/router.test.ts`
- `git diff --check`

## Review evidence

- Focused follow-up from PR #15035 review sweep.
- Route output is covered for `lib/runtime.ml:12` and absolute `/tmp/runtime.ml` is covered as no-link.

## Linked issue

- Follow-up to #15035.

### DIFF
--- a/dashboard/src/components/logs.test.ts
+++ b/dashboard/src/components/logs.test.ts
@@ -1,0 +1,72 @@
+import { h } from 'preact'
+import { cleanup, render, waitFor } from '@testing-library/preact'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+async function loadLogs(fetchLogs: ReturnType<typeof vi.fn>) {
+  vi.resetModules()
+  vi.doMock('../api/dashboard.js', () => ({
+    fetchLogs,
+  }))
+  return import('./logs')
+}
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #15130. The replacement carries the same dashboard log code-link diff on `fix/dashboard-log-code-link-20260514`, keeps the PR draft-only, and avoids the stale `agent-pr` label policy trigger without adding `human-approved-ready`.

PR #15131, opened 2026-05-14
fix(keeper): count empty after-turn responses

jeong-sik:
## Summary
- Add `masc_after_turn_response_content_empty_total` for after-turn responses that complete without visible assistant text or tool progress.
- Wire the counter into `Keeper_hooks_oas.after_turn` with bounded labels: `keeper`, `stop_reason`, `shape`.
- Treat tool-use, media, non-empty tool results, and structured tool-result JSON as progress so tool-only turns do not false-positive.

## Verification
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_hooks_oas_telemetry.exe`
- `./_build/default/test/test_keeper_hooks_oas_telemetry.exe` (47 tests)

## Review Notes
- Draft PR only. Do not add `human-approved-ready` until human review is complete.
- Expected policy reds while draft/unapproved: Draft Auto-Merge Guard and CI Gate.

### DIFF
--- a/lib/keeper/keeper_hooks_oas.ml
+++ b/lib/keeper/keeper_hooks_oas.ml
@@ -268,6 +268,9 @@ let empty_response_model_metric =
 let alias_response_model_metric =
   Prometheus.metric_after_turn_response_model_alias
 
+let empty_response_content_metric =
+  Prometheus.metric_after_turn_response_content_empty
+
 let zero_usage : Agent_sdk.Types.api_usage =
   {
     input_tokens = 0;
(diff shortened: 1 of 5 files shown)


jeong-sik:
Superseded by #15132. The replacement carries the same after-turn empty-response metric diff on `fix/keeper-after-turn-empty-response-metric-20260514`, keeps the PR draft-only, and avoids the prior `codex/...` branch policy trigger without adding `human-approved-ready`.

PR #15185, opened 2026-05-14
fix(dashboard): sync surface readiness proof

jeong-sik:
## Summary
- Align `Dashboard_surface_readiness` with the current Dashboard v1 navigation SSOT and expose `verification_ref_bar` instead of the stale proof-bar naming.
- Keep legacy redirect-only sections such as `monitoring:memory-subsystems` out of canonical readiness/nav-event targets while preserving redirect compatibility.
- Restore focused Dune build hygiene exposed during validation: add `test_log_ring_encoder` to the stanza modules list and define the missing compile-time constants before use.
- Keep `lib/prometheus.ml` within the existing 3300-line Godfile cap without raising the threshold.

## Verification
- `bash scripts/check-dashboard-surface-parity.sh --check`
- `bash scripts/check-dashboard-nav-event-parity.sh --check`
- `bash scripts/lint/godfile-size-regression.sh`
- `bash -n scripts/check-dashboard-surface-parity.sh scripts/check-dashboard-nav-event-parity.sh scripts/verify-dashboard.sh`
- `ocamlformat --check lib/dashboard/dashboard_surface_readiness.ml lib/dashboard/dashboard_surface_readiness.mli test/test_dashboard_surface_readiness.ml lib/dashboard/dashboard_nav_event.ml test/test_dashboard_nav_event.ml lib/keeper/keeper_llm_bridge.ml lib/dashboard/dashboard_operator_judge.ml lib/prometheus.ml`
- `scripts/dune-local.sh build ./test/test_dashboard_surface_readiness.exe ./test/test_dashboard_nav_event.exe`
- `./_build/default/test/test_dashboard_surface_readiness.exe` (7 tests)
- `./_build/default/test/test_dashboard_nav_event.exe` (14 tests)
- `pnpm exec vitest run --config vitest.config.ts src/components/surface-readiness-panel.test.ts src/components/status.test.ts src/config/navigation.test.ts --no-file-parallelism --maxWorkers=1` (3 files, 58 tests)

Head validated: `d34bf53cb693e06b903810010f13c8920f81db4c`.


anyang-keepers (commented):

Finding: `proof_bar_for_refs` is still a static reference-presence bar, not a proof bar. In `lib/dashboard/dashboard_surface_readiness.ml`, lines 55-74 add labels whenever a ref string is non-empty, and lines 90-99 default most entries to `/api/v1/dashboard/logs` and `/metrics`. That can emit `logs`, `metrics`, or `live_spotcheck` even when those endpoints are down, stale, or never checked. This is the fake-proof failure class: the UI label says proof while the code only knows configured references.

Please either rename this surface to something like `declared_refs` / `verification_ref_bar`, or make it consume live verifier output with status and timestamp per ref. The regression test should include a broken/stale ref and assert it is not counted as proof.

jeong-sik:
Review finding appears addressed by the current head a2d8fe0408 (`Rename surface readiness proof bar to verification refs`). The surface no longer exposes a `proof_bar` contract: API/UI/tests use `verification_ref_bar`, the UI KPI label is `refs`, and `test_verification_ref_bar_reflects_declared_refs` asserts the bar reflects declared references rather than live proof. I am leaving this as a scoped rename/contract fix rather than adding live verifier state in this PR, because no live status/timestamp source is present in the current readiness inventory path.

anyang-keepers:
Addressed review: renamed the misleading proof_bar surface to verification_ref_bar across API, tests, and dashboard UI so it no longer implies proof validation. Head: a2d8fe040865. Local checks: git diff --check, ocamlformat --check, and the scoped dashboard Vitest passed.

PR #15189, opened 2026-05-14
fix(cascade): cap inferred max tokens to output ceiling

jeong-sik:
## Summary

- Cap automatically resolved cascade `max_tokens` to the resolved cascade output ceiling.
- Keep explicit caller-provided `max_tokens` on the existing strict pre-dispatch rejection path.
- Add a regression fixture for mixed failover members where one alias resolves to 64000 output tokens and another is capped at 16384.

## Live Failure

Observed live keeper turns blocked before dispatch with:

`pre_dispatch_max_tokens_ceiling_violation`: `tier-group.strict_tool_candidates` requested `64000` while the resolved provider ceiling was `16384`.

That config is live-only under `~/me/.masc/config/cascade.toml`, but the failure class is code-level: an automatic inferred budget can exceed the minimum output ceiling for a failover tier/group. This patch normalizes automatic inference before validation while preserving hard rejection for explicit overrides.

## Verification

- `ocamlformat --check lib/cascade_inference.ml lib/cascade_inference.mli test/test_cascade_capability_gate.ml`
- `env DUNE_CACHE=disabled scripts/dune-local.sh build test/test_cascade_capability_gate.exe`
- `./_build/default/test/test_cascade_capability_gate.exe`
- `git diff --check`

### DIFF
--- a/lib/cascade_inference.ml
+++ b/lib/cascade_inference.ml
@@ -91,9 +91,8 @@ let resolve_temperature
   | None -> fallback ()
 
 (** Cap an automatically-derived max_tokens value to the narrowest output
-    ceiling in the resolved cascade. Explicit caller-provided overrides do not
-    pass through [resolve_max_tokens]; those stay hard rejected by the
-    pre-dispatch validator. *)
+    ceiling in the resolved cascade. Explicit caller-provided overrides and
+    cascade.toml values stay hard rejected by the pre-dispatch validator. *)
 let cap_auto_resolved_max_tokens ~cascade_name ~source max_tokens =
   match Cascade_runtime.max_output_tokens_ceiling_of_cascade_name cascade_name with
   | Some ceiling when ceiling > 0 && max_tokens > ceiling ->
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):

Finding: the PR title says inferred max_tokens, but `resolve_max_tokens` now also caps explicit cascade config values. `lib/cascade_inference.ml` lines 93-106 describe `cap_auto_resolved_max_tokens`, yet lines 115-120 pass both `Some t -> t, "cascade_config"` and fallback values through it. That silently rewrites an operator-authored cascade setting into the narrowest provider ceiling; a warning log is not enough because dispatch proceeds with a different config than requested.

Please cap only fallback/auto-derived values, and keep cascade-config max_tokens on the existing hard validation path so over-ceiling config fails loudly. Add a regression where `[routes].max_tokens` or cascade config above the ceiling returns a ceiling violation instead of being reduced.

jeong-sik:
Addressed the review finding in b0b98c55ec. `resolve_max_tokens` now only caps fallback/auto-derived values; cascade-config max_tokens is preserved and then rejected by `validate_max_tokens_within_ceiling` when it exceeds the resolved provider output ceiling. Added a regression that sets `keeper_unified_max_tokens = 65536` against a mixed failover ceiling of 16384 and asserts the value is not silently reduced, then produces `requested_exceeds_provider_ceiling`. Local checks: `opam exec -- ocamlformat --check lib/cascade_inference.ml lib/cascade_inference.mli test/test_cascade_capability_gate.ml`, `git diff --check`.

anyang-keepers:
Addressed review: resolve_max_tokens now preserves explicit cascade config so the validator reports requested_exceeds_provider_ceiling instead of silently capping it. Head: b0b98c55ec90. Local checks: git diff --check and ocamlformat --check passed; focused Dune was blocked by the shared opam/dune lock/pin drift. CI Gate/check/ocamlformat are passing on GitHub.

PR #15194, opened 2026-05-14
fix(keeper): demote routine OAS bridge cancels

jeong-sik:
## Summary
- classify routine parent-cancel cases in keeper_llm_bridge (Eio timeout / Not_first within fast-mid buckets)
- emit those routine bridge cancellations at INFO while keeping long or unknown cancellations at WARN
- add log_class details so operators can still split routine_parent_cancel vs warn_cancel

## Validation
- `opam exec -- ocamlformat --check lib/keeper/keeper_llm_bridge.ml test/test_keeper_llm_bridge.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_llm_bridge.exe`
- `./_build/default/test/test_keeper_llm_bridge.exe`

### DIFF
--- a/lib/keeper/keeper_llm_bridge.ml
+++ b/lib/keeper/keeper_llm_bridge.ml
@@ -41,6 +41,7 @@ let key_wall_sec = "wall_sec"
 let key_overshoot_sec = "overshoot_sec"
 let key_bucket = "bucket"
 let key_inner_exception = "inner_exception"
+let key_cancel_classification = "cancel_classification"
 let key_rollback_scope = "rollback_scope"
 let key_external_tool_side_effects_reverted = "external_tool_side_effects_reverted"
 let key_error = "error"
@@ -51,6 +52,8 @@ let field_timeout_enforced = "timeout_enforced"
 let field_wall_sec = "wall_sec"
 let field_bucket = "bucket"
 let field_inner_exception = "inner_exception"
+let field_log_class = "log_class"
+let field_cancel_classification = "cancel_classification"
 let field_error = "error"
 let field_rollback_scope = "rollback_scope"
 let field_external_tool_side_effects_reverted = "external_tool_side_effects_reverted"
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):

Finding: the cancel demotion is based on exact `Printexc.to_string` values plus wall-clock buckets. `lib/keeper/keeper_llm_bridge.ml` lines 55-80 classify `Eio__core__Fiber.Not_first` and `Eio__Time.Timeout` as routine for fast/short/mid buckets, and lines 229-265 then emit the cancellation at INFO even though it still re-raises and the failure envelope severity remains WARN. That is brittle hardcoding around internal exception names, and it can hide real bridge timeout regressions from warn-based operator surfaces.

Please classify routine parent cancellation from typed/explicit caller context, not exception string text and elapsed time. Non-routine `Timeout` within 600s should stay WARN, and the tests should cover that distinction.

jeong-sik:
Addressed the review finding in b999f2d7ac. Cancellation severity no longer depends on exact `Printexc.to_string` values or elapsed bucket membership. `run_with_timeout_and_fallback` now accepts an explicit `cancel_classification` (`Unknown_cancel` by default, `Routine_parent_cancel` only when the caller provides that typed context). The default path keeps even `Eio.Time.Timeout` cancellations at WARN, while the explicit routine context is the only INFO path. Added regression coverage for both cases. Local checks: `opam exec -- ocamlformat --check lib/keeper/keeper_llm_bridge.ml lib/keeper/keeper_llm_bridge.mli test/test_keeper_llm_bridge.ml`, `git diff --check`. Focused `scripts/dune-local.sh build test/test_keeper_llm_bridge.exe` was queued behind the shared Dune lock and was stopped before compile; CI is the compile signal for the pushed head.

anyang-keepers:
Addressed review: routine cancel demotion now requires explicit cancel_classification from the caller; default Eio.Time.Timeout cancellation remains WARN with log_class=warn_cancel. Head: b999f2d7ac53. Local checks: git diff --check and ocamlformat --check passed; focused Dune was blocked by the shared opam/dune lock. GitHub CI Gate/check/ocamlformat are passing; a couple advisory checks were still pending when last polled.

PR #15196, opened 2026-05-14
fix(dashboard): verify runtime status and keeper startup

jeong-sik:
## Summary
- add `dashboard/ping` over the authenticated WebSocket session and record client-side ping/pong timing
- treat WS-only idle periods as healthy when heartbeat pongs are fresh, while still warning when both events and pongs are absent
- bound keeper lifecycle POST calls with the dashboard POST timeout and log server-side keeper lifecycle durations
- keep keeper startup preflight lightweight: runtime hardening + image presence stay on the request path, while the full Docker command inventory remains in doctor/status diagnostics
- remove the legacy `memory-subsystems` monitoring route/type/readiness surface; the active memory subsystem API/component remains under cognition
- rebase the PR onto current `origin/main` and keep the branch free of the restored legacy redirect commit

## Why
The dashboard previously made WS-only mode look suspicious whenever route events were quiet, even if the WebSocket itself was alive. Keeper lifecycle actions also had little visible timing evidence when they appeared to hang. The keeper startup path was also using the full Docker preflight from the HTTP request path, which could stack multiple Docker timeouts including image command inventory. This keeps the UI and startup paths bounded while preserving the deeper diagnostics outside the hot path.

## Verification
- `pnpm exec vitest run --config vitest.config.ts src/config/navigation.test.ts src/dashboard-ws.test.ts src/components/transport-beacon.test.ts src/components/status-tray.test.ts src/api/keeper.test.ts src/components/status.test.ts --no-file-parallelism --maxWorkers=1` (`6 passed`, `105 tests`)
- `ocamlformat --check lib/dashboard/dashboard_nav_event.ml lib/dashboard/dashboard_surface_readiness.ml test/test_dashboard_surface_readiness.ml lib/keeper/keeper_sandbox_runtime.ml lib/keeper/keeper_sandbox_runtime.mli lib/server/server_dashboard_http_keeper_api.ml`
- `git diff --check`
- `rg -n "monitoring\\.memory-subsystems|monitoring\\?section=memory-subsystems|memory-subsystems.*hidden|hidden.*memory-subsystems|Legacy redirect target|Restore memory subsystem" dashboard/src lib test docs scripts` (no matches)
- GitHub checks on head `b7952445caef4d3e4351d9cb45fabce9f2b6aa5a`: pass/skipped

## Verification note
- Local Dune executable proof was attempted with `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_dashboard_surface_readiness.exe test/test_keeper_docker_read.exe test/test_mcp_server_eio.exe bin/main_eio.exe`, but the shared opam switch currently has `agent_sdk 0.193.11` installed while this repo branch is pinned to `0.193.10`; the build stops in `lib/oas_compat/oas_compat.ml` on the newer telemetry record fields. I did not downgrade the shared switch or mix an OAS SDK bump into this dashboard/server PR.


anyang-keepers (commented):

Finding: this heartbeat PR also removes the legacy `memory-subsystems` route contract. In `dashboard/src/config/navigation.ts`, `SECTION_REDIRECTS` lines 340-371 no longer contain `monitoring:memory-subsystems`, and this branch does not add an equivalent `status.ts` normalization. Old links now fall through instead of redirecting to cognition. That is unrelated to websocket heartbeat and appears to depend on a separate surface-readiness PR to preserve legacy behavior.

Please split the route cleanup out, or keep the redirect in this branch until the same PR adds the normalization plus regression test.

jeong-sik:
Addressed the review finding in b4c1268485. The legacy `monitoring:memory-subsystems` route contract is restored on this branch instead of relying on a separate surface-readiness PR: the hidden nav item and `SECTION_REDIRECTS` entry now route to `cognition` with `view=memory`, and regression coverage asserts both the redirect table and `normalizeRouteParams` behavior. Local checks: `pnpm -C dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/config/navigation.test.ts` (47 tests passed), `git diff --check`.

jeong-sik:
Follow-up: the PR branch advanced after the first response and dropped the legacy redirect again, so I reapplied the fix on current head and pushed 5cc9e1e99a. Current branch now restores monitoring:memory-subsystems -> cognition with view=memory, keeps the hidden legacy nav item, and updates the hidden-route and redirect regression tests. Local verification on the pushed head: pnpm -C dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/config/navigation.test.ts (47 passed); git diff --check HEAD~1 HEAD.

PR #15197, opened 2026-05-14
fix(dashboard): add memory-subsystems to SurfaceSectionId type

jeong-sik:
Adds missing `memory-subsystems` to the `SurfaceSectionId` union type to match the existing entry in `DASHBOARD_SECTION_ITEMS`. Prevents TypeScript compilation errors for legacy redirect handling.


anyang-keepers (commented):
Reviewed the one-line type fix. The code change itself matches the existing `memory-subsystems` navigation entry and I do not see a hardcoding/silent-failure/stub issue in the diff. This PR is not merge-ready yet: current live checks show `Lint` failing and `Dashboard` still pending in run 25843637117, so keep it out of merge until the current CI run is clean or the lint failure is classified and fixed.

PR #15216, opened 2026-05-14
fix(provider): centralize provider model label parsing

jeong-sik:
## Summary
- Adds Provider_adapter helpers for provider:model parsing, model id extraction, and canonical provider resolution.
- Moves provider:model colon parsing in cascade runtime, runtime candidates, and dashboard provider runs back behind the Provider_adapter boundary.
- Adds provider_adapter regression coverage for normalized provider/model parts and canonical provider lookup.
- Fixes the current-main test/dune metadata gap where test_log_ring_encoder is listed in names but missing from the stanza modules list.
- Keeps lib/prometheus.ml within the existing Godfile size cap by removing four blank doc-comment lines, without raising the threshold.

## Deep-dive coverage
- DD-019: reduces scattered provider_kind/cascade_prefix reverse lookup and provider:model parsing by making Provider_adapter the call boundary for these label operations.

## Validation
- PASS: ocamlformat --check lib/prometheus.ml lib/provider_adapter.mli lib/provider_adapter.ml lib/cascade/cascade_runtime.ml lib/cascade/cascade_runtime_candidate.ml lib/dashboard_provider_runs.ml test/test_provider_adapter.ml
- PASS: git diff --check
- PASS: bash scripts/lint/godfile-size-regression.sh
- BLOCKED locally: env DUNE_CACHE=disabled scripts/dune-local.sh build test/test_provider_adapter.exe aborts because the shared opam switch is pinned to agent_sdk 0.193.11 while this branch/main expects agent_sdk 0.193.10. With MASC_SKIP_PIN_CHECK=1 it reaches compile and fails in existing oas_compat/llm_metric_bridge callback records due the newer SDK streaming fields. Clean PR CI should validate against the repo pin.

### DIFF
--- a/lib/cascade/cascade_model_label.ml
+++ b/lib/cascade/cascade_model_label.ml
@@ -1,0 +1,41 @@
+let provider_prefix_of_label label =
+  let normalized = String.trim label in
+  match String.index_opt normalized ':' with
+  | Some idx when idx > 0 ->
+      Some (String.sub normalized 0 idx |> String.trim |> String.lowercase_ascii)
+  | _ -> None
+
+let provider_prefix_of_label_result label =
+  match provider_prefix_of_label label with
+  | Some provider -> Ok provider
+  | None ->
+      Error
+        (Printf.sprintf
+           "Runtime model label must be provider:model, got: %s"
+           (String.trim label))
+
+let provider_model_parts_result label =
+  let normalized = String.trim label in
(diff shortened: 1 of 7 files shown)


jeong-sik:
Closing during PR queue cleanup. This draft is stale/conflicting against current main, and the active provider/cascade cleanup work has moved to newer focused PRs. Reopen if this exact slice is still needed.

PR #15366, opened 2026-05-14
fix(keeper): classify empty cascades by resolved candidates

jeong-sik:
## Summary
- classify empty cascade pre-dispatch failures as provider-unavailable when provider resolution produced zero candidates
- keep no_tool_capable_provider only for cascades that had resolved candidates and lost them to the tool-use gate
- extend keeper runtime manifest regression coverage for the zero-resolved-candidate case

## Verification
- ocamlformat --check lib/keeper/keeper_turn_driver_helpers.ml lib/keeper/keeper_turn_driver_helpers.mli lib/keeper/keeper_turn_driver.ml test/test_keeper_runtime_manifest.ml
- git diff --check
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_keeper_runtime_manifest.exe
- ./_build/default/test/test_keeper_runtime_manifest.exe

### DIFF
--- a/lib/keeper/keeper_turn_driver.ml
+++ b/lib/keeper/keeper_turn_driver.ml
@@ -307,6 +307,7 @@ let run_named
         classify_empty_candidates
           ~require_tool_choice_support
           ~require_tool_support
+          ~original_candidate_count
           ~tool_filtered_candidate_count
       in
       let classification_code =
--- a/lib/keeper/keeper_turn_driver_helpers.ml
+++ b/lib/keeper/keeper_turn_driver_helpers.ml
@@ -92,9 +92,10 @@ type empty_candidate_classification =
   | Provider_unavailable
 
 let classify_empty_candidates ~require_tool_choice_support ~require_tool_support
-    ~tool_filtered_candidate_count =
+    ~original_candidate_count ~tool_filtered_candidate_count =
   if
     (require_tool_choice_support || require_tool_support)
+    && original_candidate_count > 0
     && tool_filtered_candidate_count = 0
   then Tool_capability_empty
   else Provider_unavailable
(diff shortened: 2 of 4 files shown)


anyang-keepers (commented):
## QA Review: Edge cases on empty cascade classification

The fix correctly distinguishes `Provider_unavailable` (zero resolved candidates) from `Tool_capability_empty` (candidates resolved but filtered out by tool gate). This is a good structural improvement.

**Edge case concerns:**

1. **`original_candidate_count = 0` with `require_tool_support = false`**: The classification falls through to `Provider_unavailable`. But if the provider pool is genuinely empty (not just temporarily unavailable), this gets the same classification as a transient network failure. Should there be a `Provider_pool_empty` variant for observability?

2. **Negative `original_candidate_count`**: The function signature takes `int` without validation. A negative value would bypass the `original_candidate_count > 0` check and classify as `Provider_unavailable`. This is a silent logic error path.

3. **Test coverage gap**: The 10-line test addition covers the happy path (2 candidates → 0 after filter) and the zero-candidate path. But it doesn't test:
   - `original_candidate_count = 0, tool_filtered = 0, require_tool_support = false` (should be `Provider_unavailable`)
   - `original_candidate_count = 1, tool_filtered = 0, require_tool_support = false` (should be `Provider_unavailable`, not `Tool_capability_empty`)
   - Large `original_candidate_count` with overflow risk

**Recommendation**: Add boundary tests for `original_candidate_count = 0` and `original_candidate_count = 1` with `require_tool_support = false`. The classification logic is simple but the boundary between "no providers" and "providers filtered out" is critical for debugging cascade_exhausted incidents.

PR #15368, opened 2026-05-14
fix(cascade): dedupe automatic max token clamp warnings

jeong-sik:
## Summary
- keep automatic max_tokens fallback clamping to provider output ceilings
- emit the clamp WARN only once per cascade/source/requested/ceiling tuple
- count every repeated clamp with `masc_cascade_max_tokens_clamped_total` so dashboards keep rate visibility without log spam

## Runtime Evidence
- repeated 24h WARN sample: `auto-resolved max_tokens=65536 from fallback exceeds output ceiling=16384; using ceiling` across `tier.ollama_cloud_primary`, `ollama_cloud_primary`, and `direct_api_manual`
- root cause: `Cascade_inference.cap_auto_resolved_max_tokens` logged on every automatic fallback clamp even though runtime behavior was already safe

## Verification
- `ocamlformat --check lib/cascade_inference.ml lib/cascade_inference.mli lib/cascade/cascade_metrics.mli test/test_cascade_capability_gate.ml`
- `git diff --check`
- `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_cascade_capability_gate.exe`
- `./_build/default/test/test_cascade_capability_gate.exe`

### DIFF
--- a/lib/cascade/cascade_metrics.mli
+++ b/lib/cascade/cascade_metrics.mli
@@ -291,10 +291,11 @@ val on_cascade_metrics_eviction : unit -> unit
     [cascade_max_keys] or canonicalize synthesized names. *)
 
 val on_max_tokens_clamped : unit -> unit
-(** Legacy iter-46 counter retained for metric compatibility. DD-020
-    replaced runtime max_tokens clipping with the structured
-    [max_tokens_ceiling_violation] pre-dispatch error, so current code
-    paths should not emit this counter. *)
+(** Tick when an automatically-derived max_tokens fallback is reduced to the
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
## Review: cascade max-token clamp dedupe ✅

Clean fix for the 24h WARN spam. Changes:
- `lib/cascade_inference.ml`: dedupe logic (+40/-5)
- `lib/cascade/cascade_metrics.mli`: metric counter (+5/-4)
- `test/test_cascade_capability_gate.ml`: regression tests (+48)

All CI green. ocamlformat clean. Test executable passes.

PR #15370, opened 2026-05-14
fix(task-sandbox): resolve docker-visible worktree paths

jeong-sik:
## Summary
- resolve Task_sandbox worktree paths back to host paths when Coord_worktree returns docker-visible cwd paths
- fail sandbox creation when the host worktree or active .masc symlink cannot be established, instead of warning and continuing into repeated cwd/config errors
- add a docker keeper lane regression test that verifies the returned path and .masc symlink are host-valid

## Runtime evidence
- observed repeated [task_sandbox] symlink .masc failed: No such file or directory in ~/.masc/logs/system_log_2026-05-14.jsonl
- same window had downstream docker errors for missing .masc/config and missing repos/<repo>/.worktrees/<task> cwd

## Validation
- ocamlformat --check lib/task_sandbox.ml test/test_task_sandbox.ml
- git diff --check
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_task_sandbox.exe
- ./_build/default/test/test_task_sandbox.exe

### DIFF
--- a/lib/task_sandbox.ml
+++ b/lib/task_sandbox.ml
@@ -52,6 +52,7 @@ let rel_worktree_re =
   Re.Pcre.re {|((?:[^ \t\n\r:]+/)?\.worktrees/[^ \t\n\r]+)|} |> Re.compile
 ;;
 let branch_name_re = Re.Pcre.re {|Branch: ([^ \t\n\r]+)|} |> Re.compile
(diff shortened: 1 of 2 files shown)


anyang-keepers (commented):
## Review: cascade tier-group fallback routing ✅

All CI green. Clean implementation.

PR #15374, opened 2026-05-14
fix(cascade): hydrate persisted routing hints

jeong-sik:
## Summary
- persist latency/confidence/cost routing hints in cascade trust snapshots
- hydrate those hints back into Cascade_health_tracker after restart so weighted cascade selection keeps recent p50 latency signal
- add regression coverage for hydrated latency affecting Cascade_strategy latency scoring

## Verification
- scripts/dune-local.sh build test/test_cascade_trust_persist.exe test/test_cascade_health_tracker.exe test/test_cascade_strategy.exe
- ./_build/default/test/test_cascade_trust_persist.exe
- ./_build/default/test/test_cascade_health_tracker.exe
- ./_build/default/test/test_cascade_strategy.exe
- git diff --check

### DIFF
--- a/lib/cascade/cascade_health_tracker.ml
+++ b/lib/cascade/cascade_health_tracker.ml
@@ -341,6 +341,9 @@ type provider_restore = {
   restore_cooldown_until : float option;
   restore_last_failure_at : float option;
   restore_top_fingerprints : (string * int) list;
+  restore_latency_ms : float option;
+  restore_confidence : float option;
+  restore_cost_usd : float option;
 }
 
 (* ── Constructor ──────────────────────────────── *)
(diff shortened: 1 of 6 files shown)


anyang-keepers (commented):
## Review: hydrate persisted routing hints ✅

All CI green. Clean implementation.

anyang-keepers (commented):
## QA Review: Edge cases on hydrated routing hints

The hydration of persisted latency/confidence/cost signals into the health tracker is a good resilience improvement. But several edge cases need test coverage:

1. **`restore_latency_sample` with `latency_ring_size = 1`**: The cursor is set to 0, but the ring has 1 slot. The next write will overwrite index 0. Is this intended? A test with `latency_ring_size = 1` should verify the single-sample behavior.

2. **`restore_confidence_sample` with `confidence = NaN` or `infinity`**: `Float.is_finite` rejects these, but what about `confidence = -0.0` (negative zero)? It passes `is_finite` but may cause comparison issues downstream.

3. **`restore_cost_sample` with `cost_usd = 0.0`**: The check is `cost_usd >= 0.0`, so 0.0 is accepted. But a zero-cost provider is suspicious — should this be logged or rejected?

4. **All three restore functions with `*_ring_size = 0`**: The check `*_ring_size > 0` prevents array creation, but if ring size is 0, the persisted hint is silently dropped. This is a data loss path with no warning.

5. **Test coverage gap**: The 43-line test addition covers basic hydration, but doesn't test:
   - `latency_ms = 0.0` (boundary)
   - `confidence = 0.0` or `1.0` (boundary)
   - `cost_usd = 0.0` (boundary)
   - `NaN`/`infinity` rejection for all three fields
   - Ring size = 1 behavior
   - Ring size = 0 behavior (data loss path)

**Recommendation**: Add boundary and invalid-input tests. The restore functions are defensive but silently drop invalid data — this makes debugging harder when persisted hints disappear after restart.

PR #15377, opened 2026-05-14
fix(keeper): gate turns on cascade backpressure

jeong-sik:
## Summary\n- promote cascade health probe results into pre-dispatch keeper turn admission\n- block requested keeper turns when the active cascade is cached as unhealthy instead of sending another LLM request into the same bad route\n- stamp cascade backpressure skip reasons and touch last_turn_ts so stale watchdog logs preserve intentional backpressure instead of classifying it as a stuck fiber\n\n## Deep-dive link\n- DD-009: Stale_termination_storm / Stale_fleet_batch / Oas_timeout_budget_loop were symptom caps; this adds the first direct admission backpressure gate.\n\n## Verification\n- ocamlformat --check lib/keeper/keeper_heartbeat_loop.ml lib/keeper/keeper_heartbeat_loop.mli test/test_keeper_semaphore_wait_counter.ml\n- scripts/dune-local.sh build test/test_keeper_semaphore_wait_counter.exe\n- ./_build/default/test/test_keeper_semaphore_wait_counter.exe (15 tests)\n- git diff --check

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -369,6 +369,50 @@ let record_semaphore_wait_observation
     ~reasons:(semaphore_wait_observation_reasons ?phase_label ~kind ~channel ())
 ;;
 
+type cascade_backpressure_decision =
+  | Cascade_admitted
+  | Cascade_backpressured of {
+      cascade_name : string;
+      reason : string;
+    }
+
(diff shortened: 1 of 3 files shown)


anyang-keepers (commented):
## Review: cascade backpressure gate ✅

Adds backpressure gate to prevent cascade exhaustion storms. All CI green.

PR #15384, opened 2026-05-14
fix(keeper): pin OAS termination semantics

jeong-sik:
## Summary
- add a layer-aware sdk_termination_semantics contract for OAS -> keeper SDK errors
- route keeper receipt outcome mapping through that contract without changing existing outcomes
- document the OAS termination boundary so OAS max_turns stays distinct from keeper/provider wall-clock timeout

## Deep-dive link
- DD-015: termination conditions were defined at different layers across systems, which risked silently conflating OAS turn budgets with keeper/provider timeouts.

## Verification
- opam exec -- ocamlformat --check lib/keeper/keeper_agent_error.ml lib/keeper/keeper_agent_error.mli test/test_keeper_terminal_reason.ml
- git diff --check
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_keeper_terminal_reason.exe
- ./_build/default/test/test_keeper_terminal_reason.exe (36 tests)

### DIFF
--- a/docs/spec/13-oas-integration.md
+++ b/docs/spec/13-oas-integration.md
@@ -1,9 +1,10 @@
 ---
 status: reference
-last_verified: 2026-04-17
+last_verified: 2026-05-15
 code_refs:
   - lib/verifier_oas.ml
   - lib/memory_oas_bridge.ml
+  - lib/keeper/keeper_agent_error.ml
   - lib/worker_oas.ml
 ---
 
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
CI green ✅. fix(keeper): pin OAS termination semantics — DRAFT 상태. Ready for review로 전환 후 merge 진행 가능합니다.

PR #15386, opened 2026-05-15
fix(memory): route keeper turns through bridge factory

jeong-sik:
## Summary
- add `Memory_oas_bridge.create_memory_with_backend` so callers can get the exact long-term backend installed into the returned `Memory.t`
- route the keeper turn hot path through the bridge factory instead of rebuilding `make_backend + Agent_sdk.Memory.create` inline
- add regression coverage that stores through the returned memory and reads through the returned backend

## Why
This keeps the real keeper execution path on the bridge factory boundary. Future OAS Memory backend expansion, including episodic/procedural persistence work, can be adopted in one bridge constructor instead of being bypassed by keeper turns.

## Verification
- `opam exec -- ocamlformat --check lib/memory_oas_bridge.ml lib/memory_oas_bridge.mli lib/keeper/keeper_run_tools.ml test/test_memory_oas_5tier.ml`
- `git diff --check`
- `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_memory_oas_5tier.exe`
- `./_build/default/test/test_memory_oas_5tier.exe` (21 tests)

### DIFF
--- a/lib/keeper/keeper_run_tools.ml
+++ b/lib/keeper/keeper_run_tools.ml
@@ -1722,14 +1722,15 @@ let prepare_agent_setup
     let hooks = Agent_sdk.Hooks.compose ~outer:before_turn_hook ~inner:base_hooks in
     let base_dir = Coord.masc_root_dir config in
     let memory_session_id = Keeper_id.Trace_id.to_string meta.runtime.trace_id in
-    let memory_backend =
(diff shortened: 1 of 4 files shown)


anyang-keepers (commented):
CI green ✅. DRAFT 상태 — Ready for review로 전환 후 merge 진행 가능합니다.

PR #15402, opened 2026-05-15
fix(keeper): harden bash routing and sandbox safety

jeong-sik:
## Summary
- Stop treating generic command substitution/eval as destructive keeper_bash operations.
- Let sandbox-root git/gh run when the command already names repos/<repo>.
- Normalize doubled repo cwd commands such as git -C repos/<repo> and cd repos/<repo> && ... when cwd is already that repo.
- Narrow path-rewrite syntax blocking to the actual path token, so sed ranges / grep regex arguments do not trip path safety.

## Verification
- opam exec -- ocamlformat --check lib/worker_dev_tools.ml test/test_worker_dev_tools.ml lib/keeper/keeper_shell_bash.ml lib/keeper/keeper_shell_docker.ml lib/keeper/keeper_shell_docker.mli test/test_keeper_shell_docker_route.ml
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_worker_dev_tools.exe
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_keeper_shell_docker_route.exe
- ./_build/default/test/test_worker_dev_tools.exe test validate_command_paths_redirect --show-errors
- ./_build/default/test/test_worker_dev_tools.exe test is_destructive_bash_operation --show-errors
- ./_build/default/test/test_keeper_shell_docker_route.exe test docker_route_contract 16-19 --show-errors
- git diff --check

## Note
- Full docker_route_contract still has existing local credential-bundle/test-env dependent failures, so the targeted new contract cases are the relevant local signal here.


jeong-sik:
Closing this draft intentionally. The patch fixes some real keeper_bash friction, but it moves the execution path further through legacy raw Bash/string heuristics instead of the intended typed capability substrate: Bash parse/lower -> Shell_ir -> Capability -> approval policy -> Exec_gate. We should restart this as smaller slices on the fundamental path: keep public alias cleanup separate, keep sandbox cwd compatibility separate, and route command-substitution/chaining/cwd normalization through typed IR/capability gates rather than broadening ad hoc Bash exceptions.

PR #15433, opened 2026-05-15
fix(paths): replace hardcoded /tmp with Filename.get_temp_dir_name

jeong-sik:
## Summary

- Replace 5 hardcoded /tmp/.masc_agent_* session file paths with Filename.get_temp_dir_name() — respects TMPDIR
- Replace hardcoded /tmp/keeper-creds and /tmp/masc-fleet with MASC_CRED_ROOT env var + get_temp_dir_name()
- Promote filesystem error logs from Logs.warn to Logs.err in exec_core.ml
- Add env var override documentation to gRPC client localhost fallback

## Audit findings addressed
P1, P2, P8, P10, L1, L2, U4 (7 HIGH items)

## Test plan
- [x] dune build @check passes
- [ ] CI green

🤖 Generated with Claude Code

### DIFF
--- a/lib/exec_core.ml
+++ b/lib/exec_core.ml
@@ -587,7 +587,7 @@ let default_recovery_hint classification semantic_status =
 
 let ensure_exec_artifact_dir path =
   try Fs_compat.mkdir_p path with
-  | Sys_error msg -> Logs.warn (fun f -> f "exec artifact mkdir failed: %s" msg)
+  | Sys_error msg -> Logs.err (fun f -> f "exec artifact mkdir failed: %s" msg)
 ;;
 
 let persist_artifact_if_needed ~base_path ~keeper_name ~cmd ~output =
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closed for re-creation on current main. The /tmp→Filename.get_temp_dir_name() fix will be resubmitted as a clean PR after RFC-0056 Phase 1C (#15441) lands or on current main.

jeong-sik:
@chatgpt-codex-connector re: `cred_root` under writable `/tmp` — good catch in principle, but not a regression here.

1. `host_config_provider.ml` delegates to `Host_config.legacy_macos_default().cred_root`, which checks `MASC_CRED_ROOT` env var first (line 40 of `host_config.ml`). In Docker deployments, this env var is always set to a writable volume path.

2. The `Filename.get_temp_dir_name()` fallback only applies when both `MASC_CRED_ROOT` is unset AND the operator hasn't configured `TMPDIR` — i.e., local dev mode on a bare macOS host, where `/tmp` (symlink to `/private/tmp`) is always writable.

3. Before this PR, the same path was hardcoded as `/tmp/keeper-creds` — the new code resolves to the identical path on stock macOS. The only behavioral change is on systems where `TMPDIR` is explicitly set to a non-`/tmp` location, which is the *correct* behavior (respect the platform convention).

No regression in the Docker read-only filesystem scenario.

PR #15539, opened 2026-05-15
fix(build): link extracted tool types library

jeong-sik:
## Summary

- Link the new `masc_mcp.tool_types` leaf library from the parent `masc_mcp` library.
- Preserve existing bare `Tool_args` / `Tool_result` references after the RFC-0086 PR-2H extraction.

## Root Cause

`lib/tool_types/` defines `Tool_args` and `Tool_result` as a wrapped-false leaf library, but `lib/dune` did not list `masc_mcp.tool_types` in the parent library dependencies. A fresh `bin/main_eio.exe` build from current `main` therefore failed with `Unbound module Tool_args` / `Unbound module Tool_result`.

## Verification

- `git diff --check`
- `bash scripts/ocaml-structure-ratchet.sh`

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -331,6 +331,10 @@
    ;; extracted to lib/tool_catalog_surfaces/. deps: Agent_sdk + stdlib only.
    ;; Unblocks one of 6 cycle modules identified in plan §8.14 iter-2.
    masc_mcp.tool_catalog_surfaces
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #15536, which already added the missing  link on . Verified current  builds with . Closing this duplicate draft to avoid CI noise.

PR #15573, opened 2026-05-15
fix(ci): read registry variants from split types module

jeong-sik:
## What changed

- Updated `scripts/check-variants.sh` to extract `turn_phase` and `cascade_state` from `lib/keeper/keeper_registry_types.ml` after the registry type split.

## Why

`Meta Guards` was still looking in `lib/keeper/keeper_registry.ml` for `Turn_idle` / `Turn_prompting`, so CI failed even though the variants now live in the split types module on current `main`.

## Validation

- `bash scripts/check-variants.sh`
- `bash scripts/ci/check-tla-variant-sync.sh`
- `git diff --check`

### DIFF
--- a/scripts/check-variants.sh
+++ b/scripts/check-variants.sh
@@ -155,7 +155,7 @@ fi
 echo ""
 echo "=== Check 2: turn_phase (OCaml) vs KeeperCascadeLifecycle.tla domain ==="
 
-KR_ML="lib/keeper/keeper_registry.ml"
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by main #15569 / #15563. The check-variants.sh fix is already on main with the canonical variable name `KR_TYPES_ML`. This PR's patch becomes a no-op after rebase (rebase produced empty diff / conflict on already-applied change). Closing to clear the queue.

PR #15680, opened 2026-05-16
fix(oas_compat): add on_tool_calls field to metric sink records

jeong-sik:
## Summary

`oas` pkg 0.193.16 added `on_tool_calls` to its `Llm_provider.Metrics.t` record type. Two masc-mcp consumers construct that record without the new field, breaking `dune build` on main:

```
File "lib/oas_compat/oas_compat.ml", lines 272-285, characters 4-5:
Error: Some record fields are undefined: on_tool_calls
File "lib/llm_metric_bridge.ml", lines 441-462, characters 2-3:
Error: Some record fields are undefined: on_tool_calls
```

This PR mirrors the minimum additions from PR #15669 (which contains the same fix bundled with 49 other changes) as a focused carve-out so main rebuilds immediately.

## Changes

- `lib/oas_compat/oas_compat.{ml,mli}` — optional `on_tool_calls` parameter on `make`, default no-op.
- `lib/llm_metric_bridge.ml` — wire `on_tool_calls` to new `masc_llm_provider_tool_calls_total` Prometheus counter via `emit_tool_calls`.
- `lib/prometheus.{ml,mli}` — register the new counter.
- `test/test_llm_metric_bridge.ml` — cover the new sink callback.

Net: 6 files, +33/-0.

## Evidence

- `dune build --root .` clean.
- `dune exec --root . test/test_llm_metric_bridge.exe` — 12/12 pass.
- `bash ~/me/scripts/pr-rfc-check.sh` — exit 0 (no credential/identity/operator subsystem touched).

## Coordination

PR #15669 will rebase to drop these files after this lands; the broader runtime-defect work in that PR is unaffected. Suggested order: merge this hotfix first → rebase #15669 → resolve any trivial drift.

## Review evidence

- Local build + focused test only. No cross-model review.
- Draft PR; relies on repository's `human-approved-ready` gate for promotion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15667, opened 2026-05-16
fix(cascade): pin OAS telemetry and floor local timeouts

jeong-sik:
Summary: pins agent_sdk to OAS 0.193.16 from jeong-sik/oas#1609 at 3d565d55730af0aa044841ff72560402efe09c0a; updates OAS API surface/docs/lockfiles; bridges the new provider-agnostic OAS tool-call metric into masc_llm_provider_tool_calls_total without adding Ollama-specific logic in MASC. Validation: bash scripts/check-oas-pin.sh; scripts/dune-local.sh build test/test_llm_metric_bridge.exe; ./_build/default/test/test_llm_metric_bridge.exe; ocamlformat --check on touched OCaml files; git diff --check.

PR #15686, opened 2026-05-16
fix(llm_metric_bridge,oas_compat): wire on_tool_calls from upstream agent_sdk

jeong-sik:
# fix(llm_metric_bridge,oas_compat): wire on_tool_calls from upstream agent_sdk

## Problem

`origin/main` does not build. `bin/main_eio.exe` fails with:

```
File "lib/oas_compat/oas_compat.ml", lines 272-285:
Error: Some record fields are undefined: on_tool_calls
File "lib/llm_metric_bridge.ml", lines 441-462:
Error: Some record fields are undefined: on_tool_calls
```

A recent OAS `agent_sdk` bump (or PR #15511 follow-up) added the `on_tool_calls` field to `Llm_provider.Metrics.t`, but the two record-literal sites in masc-mcp were not updated. CI's `Build and Test` is gated behind `Detect Changed Surfaces` (see MEMORY.md `reference_masc_mcp_ci_build_test_skips`), which is why this regression slipped past CI.

Field signature (probed against the upstream cmi):
```
on_tool_calls : provider:string -> model_id:string -> count:int -> unit
```

## Why this is a root fix, not a silent default

A noop `on_tool_calls` would compile but would be exactly the anti-pattern PR #15511 closed with "typed variants eliminate silent defaults", and the kind of transitional repair CLAUDE.md `feedback_hardcoding_and_legacy_zero_tolerance` rejects. Instead this PR wires a real Prometheus counter so tool-call activity is observable from day one.

## Changes

1. **`lib/prometheus.ml(.mli)`**: declare `metric_llm_provider_tool_calls = "masc_llm_provider_tool_calls_total"` and register it as a labeled Counter under the existing `metric_llm_provider_*` block.
2. **`lib/llm_metric_bridge.ml`**: add `tool_calls_metric` reference and `emit_tool_calls ~provider ~model_id ~count` following the same pattern as `emit_token_usage`. Wire into `make_sink` record literal. Doc string lists the new metric mapping.
3. **`lib/oas_compat/oas_compat.ml(.mli)`**: add `default_tool_calls` no-op and `?on_tool_calls` optional argument to `Metrics.make`. Compat-layer defaults are legitimate here (caller may opt out), not a silent default.

## Verification

- `dune build bin/main_eio.exe` — **clean** (was failing on `origin/main` before this PR)
- Public API additions only; no behavior change for existing callers of `Oas_compat.Metrics.make` since the new arg is optional.

## Out of scope / follow-up

`oas_compat.mli` claims *"Adding a new field upstream forces an update in exactly one place"*, but `llm_metric_bridge.ml` bypasses `Oas_compat.Metrics.make` and constructs the record literal directly. That contradiction is what caused the N-of-M miss in the first place. Worth a follow-up to route `llm_metric_bridge.make_sink` through `Oas_compat.Metrics.make` so the single-point promise holds — out of scope here to keep the build-unbreak PR minimal.

## Workaround rejection bar (self-check)

- [x] Not telemetry-as-fix (Prometheus counter is *added* alongside record-field wiring, not in lieu of it)
- [x] Not string/substring classifier
- [x] N-of-M: two field-addition sites are fixed in this PR; no remaining undefined-field site in the tree (grep clean)
- [x] No catch-all
- [x] No cap/cooldown/dedup/repair
- [x] No test backdoor

## Evidence record

- Build error: reproduced on `origin/main` `61610bb521` and `a5752c4859`, fails both record-literal sites.
- Field signature: probed via `let _ : int = (Llm_provider.Metrics.noop).on_tool_calls` against installed `agent_sdk.llm_provider` (`v0.192.1-198-g94628a4`).
- Confidence: High — minimal, type-checked, builds clean end-to-end.

🤖 Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


jeong-sik:
## Already fixed on main via #15667

Heads up — this PR's intended fix landed on main via [#15667](https://github.com/jeong-sik/masc-mcp/pull/15667) at 2026-05-16T19:14:13Z. That merge included:
- `agent_sdk` lock bump 0.193.15 → 0.194.0 (which is where `on_tool_calls` was added to `Llm_provider.Metrics.t`).
- Consumer updates to `lib/oas_compat/oas_compat.ml` and `lib/llm_metric_bridge.ml` to populate the new field.

Quick verify:

```bash
git fetch origin main
git grep on_tool_calls origin/main -- lib/llm_metric_bridge.ml lib/oas_compat/oas_compat.ml
```

Status of this PR now: CONFLICTING/DIRTY because main has the same lines you're adding here, and `check` fails for the same reason.

Suggested: close this PR without merge. (Same conclusion #15680 reached an hour ago — that one was carved out before #15667 landed; this one was opened after, but the timing crossed.)

Posted by automated /loop cycle — no code changes performed.

jeong-sik:
Closing: all changes already in main after rebase (0 commits ahead). The on_tool_calls wiring was incorporated in prior merged PRs.

PR #15993, opened 2026-05-17
feat(rfc-0107): Phase D.4 — Prometheus exporter for piaf connection pool

jeong-sik:
## Summary

RFC-0107 Phase D.4. Adds a read-only Prometheus adapter over `Masc_http_client.Pool.stats` so the piaf-backed connection pool (merged in Phase D.2c) becomes observable. The pool itself is untouched: `lib/masc_http_client/pool.ml` is read via `Pool.stats` only — no API changes.

Five metric families exported on `/metrics`:

| Metric | Type | Labels |
|---|---|---|
| `masc_pool_idle_total` | gauge | `host="scheme://host:port"` (also aggregate, unlabelled) |
| `masc_pool_inflight_total` | gauge | — |
| `masc_pool_reuse_total` | counter | — |
| `masc_pool_evict_total` | counter | — |
| `masc_pool_create_total` | counter | — |

## Wiring

- **`Masc_http_client.pool_singleton_opt`** — read-only accessor on the lazy pool ref. Returns `None` until the first HTTP call so the exporter no-ops on startup instead of forcing pool init just to read zeros.
- **`lib/server/pool_metrics.{ml,mli}`** — owns the metric name constants, exposes `current_snapshot ()` and `register ()`. No Prometheus dependency to avoid a `Prometheus ↔ Pool_metrics` module cycle.
- **`lib/prometheus.ml`** — registers the 5 metrics in `init()` and calls the new `update_pool_metrics_gauges ()` inside `to_prometheus_text ()`, mirroring the existing `update_fd_accountant_gauges ()` pattern.
- **`lib/server/server_runtime_bootstrap.ml`** — calls `Pool_metrics.register ()` at the tail of `create_server_state` for explicit dependency-order anchoring; registration itself is idempotent because metric `add` happens at `Prometheus` module load.

## Layout deviation from spec

Spec called for `pool_metrics.ml` under `lib/masc_http_client/`, but `Prometheus` lives in the `masc_mcp` library which already depends on `masc_http_client`. Placing `pool_metrics.ml` inside `masc_http_client` would either invert the dep direction or force a callback-ref dance. Mirroring the `fd_accountant` pattern (`lib/server/fd_accountant.ml` + integration in `lib/prometheus.ml`) keeps the read-only direction clean. First build attempt produced exactly this cycle (`Prometheus → Pool_metrics → Prometheus`); the chosen layout breaks it cleanly.

## Test plan

- [x] `dune build --root . @check` exit 0
- [x] `test/test_pool_metrics.ml` — 5/5 PASS
  - metric name constants stable (Grafana dashboard pin)
  - `register ()` idempotent
  - `current_snapshot ()` returns `None` before first HTTP call
  - `to_prometheus_text ()` contains all 5 metric families
  - TYPE lines render as `gauge` / `counter` correctly
- [x] `test_prometheus_coverage` 64/64 still PASS (no regression in shared registry)
- [x] `test_masc_http_client` 2/2 still PASS (singleton accessor doesn't disturb the closing-client path)
- [ ] D.5 cascade-storm reproducer to read these gauges (out of scope here)
- [ ] Dashboard tile consuming the new metrics (RFC-0107 Phase D step "D.4 dashboard tile" — separate PR)

## Anti-pattern self-check (workaround rejection bar)

- No telemetry-as-fix: this *is* telemetry, the pool itself is unchanged.
- No string classifier added.
- No N-of-M migration: all 5 stats fields are exported in one pass.
- No catch-all match arm added.
- No cap/cooldown/dedup/repair.
- No test backdoor.

### DIFF
--- a/lib/masc_http_client/masc_http_client.ml
+++ b/lib/masc_http_client/masc_http_client.ml
@@ -165,6 +165,12 @@ let get_sync ?clock ?timeout_sec ~net ?https ~url ~headers () =
   | Ok response -> Ok (response.status, response.body)
   | Error _ as error -> error
 
+(* RFC-0107 Phase D.4 — read-only accessor on the per-process pool
+   singleton.  Returns [None] before the first HTTP call (pool not
+   yet lazy-initialized) so callers like [Pool_metrics] can no-op
+   instead of forcing the pool open just to read zeros. *)
+let pool_singleton_opt () : Pool.t option = !pool_ref
+
 (* RFC-0107 Phase D.2d — re-export Pool under Masc_http_client.Pool
    so the test suite (and any direct consumer that wants the
    typed surface rather than the legacy shim) can name it without
--- a/lib/masc_http_client/masc_http_client.mli
+++ b/lib/masc_http_client/masc_http_client.mli
@@ -135,3 +135,11 @@ val get_sync :
     [Pool.request] is reserved for code that needs typed responses
     with header maps or non-default config. *)
 module Pool : module type of Pool
+
+val pool_singleton_opt : unit -> Pool.t option
+(** [pool_singleton_opt ()] returns the per-process [Pool.t] if it
+    has been lazy-initialized by a prior HTTP call, [None] otherwise.
+    Read-only accessor for telemetry consumers (Phase D.4 Prometheus
+    exporter); does not trigger pool initialization.  Callers that
+    need the pool initialized should issue a request through
+    [post_sync] / [get_sync] / [get_response_sync] instead. *)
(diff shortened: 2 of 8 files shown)


jeong-sik:
P2 review 응답 (7th loop iter): per-host gauge eviction 처리 완료.

`update_pool_metrics_gauges`에 module-local `pool_idle_seen_hosts` Hashtbl 추가. 각 scrape마다:
1. 현재 host set 빌드
2. 이전 seen set에 있지만 현재 없는 host → gauge 0으로 set
3. 현재 host 모두 gauge 갱신 + seen set 갱신
4. seen set에서 evicted host 제거 (`Hashtbl.filter_map_inplace`)

mutex/lock 무관 — `set_gauge`의 `with_lock`이 직렬화하므로 추가 동기화 불필요.

push: `ca20c6f70b..a6e6144187`.

PR #16072, opened 2026-05-18
fix: remove home-anchored masc defaults

jeong-sik:
## Summary
- remove start script fallback from HOME/me and require explicit --base-path or MASC_BASE_PATH
- remove Docker playground helper fallback from HOME/me/.masc in favor of MASC_BASE_PATH or cwd
- replace home-anchored .masc references in docs/comments/tests with <base-path>/.masc and add an SSOT guard against regressions

## Validation
- scripts/check-ssot.sh
- bash scripts/check-doc-truth.sh
- bash -n start-masc-mcp.sh scripts/docker-playground-fd-status.sh scripts/cleanup-docker-playground-worktrees.sh scripts/check-ssot.sh
- git diff --check
- rg no HOME/.masc, HOME/me/.masc, ~/.masc, ~/me/.masc, implicit_home_me matches
- smoke: start-masc-mcp.sh without --base-path/MASC_BASE_PATH exits 2 and refuses HOME inference
- smoke: docker playground helpers default to PWD/.masc when MASC_BASE_PATH is unset

## Note
Focused Dune build for test_start_masc_mcp_script.exe was attempted through scripts/dune-local.sh but could not complete because /tmp/me-dune-local.lock is held by another session.


jeong-sik:
**Iter#66 cron diagnostic — Health opam 502 transient**

```
curl: (22) The requested URL returned error: 502
```

opam binary 다운로드 단계에서 GitHub release 502. 코드 무관 transient infrastructure failure. `gh run rerun <run-id>` 또는 empty commit으로 재실행만 필요.

(본 comment는 `/loop 10m` cron `10643f01` Iter#66 진단.)

PR #16197, opened 2026-05-18
fix(startup): refuse stale server binary on build lock

### LINKED ISSUE #16208: [runtime/transport] MCP dashboard/subscribe failing with Sys_error (No such file or directory) — sustained, 38×/20m (operator cannot see broadcasts)
## Symptom

The MCP `dashboard/subscribe` method is failing with `Sys_error("No such file or directory")` for every incoming request. The failure persists across iterations — initial detection at iter17 (16 events / 28s burst, looked transient), but iter20 shows 38 events sustained over 807s window. **Not transient.**

```
Request handling failed: method=dashboard/subscribe: Sys_error("No such file or directory")
```

Module: `MCP`
Severity: `ERROR`

Window: 2026-05-18T10:03:27Z – 10:23:27Z (20m)
Count: **38** events, span 807s

## Iter trend

| iter | window end | count | span | note |
|---|---|---|---|---|
| iter17 | 09:53Z | (caught tail of burst) | — | initial detection |
| iter18 | 10:03Z | 0 in window | — | quiet — thought it had stopped |
| iter19 | 10:13Z | 16 | 28s | burst, looked one-off |
| **iter20** | 10:23Z | **38** | **807s** | sustained — not transient |

The 38-events-in-807s rate is roughly **1 request every 21 seconds** that gets rejected. This is plausibly the dashboard frontend's subscribe-then-reconnect loop — if the file behind the subscription is missing, the dashboard reconnects every ~20s, each time failing with the same error.

## Hypothesis

The `dashboard/subscribe` method reads from a file or directory that no longer exists. Likely candidates:

1. **A snapshot/state file that was supposed to be created during MCP server startup but wasn't** (race condition or migration leftover).
2. **A worktree path referenced by the subscribe handler that has been cleaned up** — connection to the cluster P (#11040) worktree leak/cleanup issue, where stale paths can be cached.
3. **A keeper/task-specific file path** that becomes invalid when the keeper or task moves state.

The `Sys_error` is OCaml's generic "raised a Unix error during a file operation" — without the stack trace or the specific path in the log line, the exact file cannot be identified from telemetry alone.

## Investigation steps (read-only, ~5 min)

```bash
# Find the subscribe handler
rg "dashboard/subscribe|method.*dashboard.*subscribe" lib/ --type ocaml -g '!_build'

# Find Sys_error catch site in MCP request handler
rg "Sys_error" lib/mcp/ lib/server/ --type ocaml -g '!_build'

# Check if there's a path constant for dashboard subscription state
rg "dashboard.*\.jsonl|dashboard.*subscribe.*path" lib/ --type ocaml -g '!_build'
```

## Why this matters

- **Dashboard is the primary operator-visible surface**. If `subscribe` is broken, operators cannot see live state — including the `operator_broadcast_required` events from #16204 (sangsu/analyst livelock pause_human escalation). The dashboard is now broken at exactly the moment human attention is most needed.
- **Continuous error rate** (1 fail per 21s) means every reconnect attempt produces a fresh ERROR log line. This is the dominant ERROR signal in the current window (top cluster by count).
- **Easy to verify a fix**: re-scan logs post-fix, this surface should drop to 0.

## Proposed direction

1. **Add the failing file path to the error log line** — current message strips the path. Adding `Printf.sprintf "Sys_error: %s" (Printexc.to_string e)` (or capturing the original `Sys_error` argument) would name the missing file in a single PR.
2. **Identify the missing file/directory** (depends on #1 above).
3. **Fix the missing-file root** — create the file lazily, or fail gracefully with a typed `Subscription_unavailable` envelope instead of `Sys_error`.

## Related issues

- #11040 (worktree leak) — *possibly* the same root if the subscribe handler references worktree paths
- #16204 (pause_human escalation) — dashboard subscribe failures make operator response harder

## Evidence pointers

- `~/me/.masc/logs/system_log_2026-05-18.jsonl`
- Filter: `module=MCP`, `level=ERROR`, message contains `method=dashboard/subscribe`
- 38 events in iter20 window (10:03:27Z – 10:23:27Z)
- Earlier burst at iter19 (10:10:00Z – 10:10:28Z, 16 events in 28s) — first observed

## Labels (suggested)

`type:bug`, `target:now`, `area:transport` (MCP request handler), `runtime`, `live-defect`

---

Detected by `/loop` iter20 (2026-05-18T10:23Z) — cluster Z (MCP dashboard/subscribe Sys_error sustained, not transient).

jeong-sik:
## Summary

- Make `start-masc-mcp.sh` fail fast when the build lock is held and the server executable is stale or missing, instead of silently continuing with an old binary.
- Export `MASC_KEEPER_HOST_FD_HOTSPOT_HEADROOM=0` from the launcher by default, while preserving explicit operator overrides.
- Make `Config_dir_resolver` survive a deleted process cwd by falling back through `MASC_BASE_PATH`, `$HOME`, then tmp.
- Add regression coverage for the FD hotspot default/override path, stale-executable build-lock refusal, and `/health` under a deleted cwd.

## Product impact

Prevents MASC startup from running a stale OCaml server binary that can re-enable obsolete Docker Desktop FD hotspot blocking and falsely block keeper admission.

Also addresses the live `/health` and dashboard/subscribe `Sys_error("No such file or directory")` failure mode from #16208: the live server was running from a worktree cwd that had already been removed, so `Sys.getcwd()` raised from `Config_dir_resolver.inputs_from_env`.

## Evidence

- Live incident showed `/health?full=1` reporting `host_fd_hotspot_headroom=49152` while current source default is `0`.
- Live remediation was verified after restart with `host_fd_hotspot_headroom=0`, `host_fd_hotspot_blocking_enabled=false`, and `admission_blocked=false`.
- Follow-up live incident showed `/health` returning 500 from `Config_dir_resolver.inputs_from_env -> Sys.getcwd()` while process cwd was `.worktrees/oas-provider-timeout-pin-20260518`, and that worktree no longer existed.
- Live service was restarted from an existing worktree and verified with `/health?full=1`: `status=ok`, `keeper_fd_pressure.status=ok`, `admission_blocked=false`, `open_fds=57`, `soft_limit=245760`.

## Review evidence

- `bash -n start-masc-mcp.sh`
- `opam exec -- ocamlformat --check test/test_start_masc_mcp_script.ml`
- `opam exec -- ocamlformat --check lib/config_dir_resolver/config_dir_resolver.ml test/test_server_runtime_bootstrap.ml`
- `git diff --check`
- `python3 scripts/ci/check-fun-protect-finally-guard.py --base origin/main --head HEAD`
- `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_start_masc_mcp_script.exe`
- `./_build/default/test/test_start_masc_mcp_script.exe` (21 tests)
- Attempted `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_server_runtime_bootstrap.exe`; first compile caught and fixed an invalid `Unix` dependency in `Config_dir_resolver`, then the retry remained blocked behind the shared `/tmp/me-dune-local.lock` queue and the duplicate queued validation was stopped. Draft CI is the remaining full verification surface.

## Linked issue

Fixes #16208.


jeong-sik:
Update: live FD-looking incident was /health failing from a deleted server cwd. The running process had cwd=.worktrees/oas-provider-timeout-pin-20260518 after that worktree was removed, so Config_dir_resolver.inputs_from_env raised Sys_error("No such file or directory") via Sys.getcwd. This PR now makes config resolution fall back to MASC_BASE_PATH/home/tmp when cwd is gone and adds a health regression for deleted cwd. Local: ocamlformat/check-fun-protect/diff-check pass; focused Dune target is queued behind the shared local Dune lock.

PR #16274, opened 2026-05-18
refactor: split prometheus core metric names

jeong-sik:
## Summary
- extract core runtime/LLM/provider/task/goal/SSE metric-name constants into Prometheus_core_metric_names
- keep Prometheus as the public facade via include
- stack on #16270 after the transport metric-name split

## Verification
- ocamlformat --inplace lib/prometheus_core_metric_names.ml lib/prometheus_core_metric_names.mli lib/prometheus.ml lib/prometheus.mli
- scripts/ocaml-structure-ratchet.sh
- git diff --check

## Notes
- Local focused Dune build was attempted but left pending behind other worktree Dune locks; CI should be used for the full proof on this draft stack.


jeong-sik:
Closing this draft as superseded by current main: core metric-name split is already merged via #16193, and this older stack now carries stale prometheus.ml/prometheus.mli reverse-diff against origin/main. Keeping #16266 as the remaining non-duplicate structure-ratchet fix.

PR #16289, opened 2026-05-18
fix: expose prometheus store snapshot

jeong-sik:
## Summary
- Expose a locked `Prometheus_store.snapshot` API that copies metric rows before rendering.
- Update `Prometheus.to_prometheus_text` to render from the store snapshot instead of referencing private store internals.
- Update `Prometheus.init` to register built-ins through the store facade instead of touching private `metrics`/`metric_key`.
- Refresh the related spawn-bounded and TLA index metadata introduced by this branch.

## Root Cause
- Current `main` at `6bc64306e4` fails local `@check` because `Prometheus.init` still references hidden store internals after the Prometheus store/built-in metric split.
- The stale direct references were `metric_key`/`metrics` in init and `with_lock`/`metrics` in text rendering.

## Current Gate
- Latest verified head: `aca2df79728108ed55d9d8a7de95d7847a519816`.
- GitHub code/integration checks are green, including `Build and Test`, `Lint`, `TLA+ Model Checking`, `ocamlformat-check`, `Spawn-bounded ratchet`, `tla-annotation-drift`, and `tla-spec-index`.
- This PR remains draft and must stay blocked while `do-not-merge` is present.
- The remaining red checks are policy-only: `CI Gate` rejects `do-not-merge`, and `Draft Auto-Merge Guard` rejects `do-not-merge` plus missing verified `human-approved-ready`.
- `Approve Agent PR` runs `26046555889` and `26046556015` are waiting at the human environment gate.

## Verification
- PASS: `opam exec -- ocamlformat --check lib/prometheus_store.ml lib/prometheus_store.mli lib/prometheus.ml`
- PASS: `git diff --check origin/main...HEAD`
- PASS: `bash scripts/lint-spawn-bounded.sh`
- PASS: `bash scripts/audit-tla-annotation-drift.sh --baseline scripts/tla-annotation-drift-baseline.txt --check-cross-spec`
- PASS: `bash scripts/audit-tla-phase-count.sh`
- PASS: regenerated `specs/INDEX.md` comparison, ignoring the `Generated:` header
- PASS: `env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_CACHE=disabled scripts/dune-local.sh build @check`

PR #16305, opened 2026-05-18
fix(prometheus): use store API during init

jeong-sik:
## Summary
- Fixes the latest main build break after the Prometheus store/built-in metric split.
- Stops `Prometheus.init` and text rendering from reaching into store internals that are hidden by `prometheus_store.mli`.
- Adds a public `Prometheus_store.snapshot` API for rendering without exposing the mutable table or lock.

## Root Cause
- main `@check` failed on `lib/prometheus.ml:256` with `Unbound value metric_key`.
- After the first fix, the next hidden-internal access failed in `to_prometheus_text` with `Unbound value with_lock`.
- Both were stale godfile-internal references left behind after #16281 and the later built-in metric split.

## Verification
- PASS: `opam exec -- ocamlformat --check lib/prometheus.ml lib/prometheus_store.ml lib/prometheus_store.mli`
- PASS: `git diff --check origin/main...HEAD`
- PASS: `env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_CACHE=disabled scripts/dune-local.sh build @check`

### DIFF
--- a/lib/prometheus.ml
+++ b/lib/prometheus.ml
@@ -250,19 +250,13 @@ let metric_mention_dedup_decisions_total = "masc_mention_dedup_decisions_total"
 (** {1 Built-in Metrics} *)
 
 let init () =
-  (* Module-level init runs before Eio context exists.
-     Single-threaded at load time — bypass mutex. *)
+  (* Module-level init runs before Eio context exists; the store uses
+     Stdlib.Mutex and is safe to call during module initialization. *)
   let add name help metric_kind =
-    let metric_type =
-      match metric_kind with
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing as superseded by #16289. #16289 covers the same Prometheus init/render store-boundary fix, keeps the documented snapshot API, and is already the base for the stacked render split in #16291.

jeong-sik:
Closing again as superseded by #16289. Please keep the Prometheus store-boundary fix on #16289 so #16291 can remain stacked on the documented snapshot API branch.

PR #16337, opened 2026-05-18
fix(tla): refresh post-turn lifecycle symbol anchor

jeong-sik:
## Summary
- refresh the KeeperPostTurnOrchestration symbol anchor for dispatch_post_turn_lifecycle_events
- point the TLA mapping to the defining module, Keeper_exec_context, while preserving the keeper_unified_turn call-site context

## Verification
- bash scripts/lint/spec-line-refs.sh
- git diff --check

Unblocks Meta Guards failures that report: dangling symbol dispatch_post_turn_lifecycle_events not found in lib/keeper/keeper_unified_turn.ml.


jeong-sik:
Superseded by current main. Rebased onto origin/main (35208d81e833840d9bcab6b2ae06f4b535423821) and both commits became empty: git log origin/main..HEAD is empty and HEAD equals origin/main. The post-turn symbol anchor and guard metadata are already present on main after the recent stack merges.

PR #16564, opened 2026-05-19
fix: unblock main OCaml test build

jeong-sik:
## Summary
- register test_yojson_type_error_board in the matching test/dune modules list
- make the new board JSON boundary test compile against the current Tool_result and Tool_board_format APIs
- wrap keeper context Prometheus increments with let () = ... in so the if/else branch parses correctly

## Verification
- ocamlformat --check lib/keeper/keeper_context_core.ml test/test_yojson_type_error_board.ml
- git diff --check
- env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_yojson_type_error_board.exe test/test_oas_integration.exe test/test_keeper_toml.exe
- ./_build/default/test/test_yojson_type_error_board.exe
- bash scripts/lint/godfile-size-regression.sh


jeong-sik:
Superseded by #16560, which merged this build-unblock patch into main as part of c2b849fe06. Closing to keep the stacked queue clean.

PR #16590, opened 2026-05-19
feat: add deliverable_type field and verifier contract templates (task-440)

anyang-keepers:
## Task-440: Add deliverable_type field and verifier contract templates for non-code tasks

**Goal**: goal-keeper-pr-lifecycle-64-20260519
**Keeper**: lifecycle-worker

### Changes

Adds a proof artifact at `proofs/task-440-deliverable-type-contract.md` containing:

1. **`deliverable_type` enum** — Classifies task deliverables beyond code changes:
   - `code_patch`, `document`, `board_post`, `analysis_report`, `config_change`

2. **Verifier contract templates** — JSON contract schemas for each deliverable type specifying required evidence fields for verification.

### Lifecycle Proof

- **Task claimed**: keeper-lifecycle-worker-agent
- **Worktree**: `keeper-lifecycle-worker-agent/task-440`
- **Commit**: `75a1e1f12` — 1 file changed, 94 insertions
- **PR**: This draft PR (keeper-created)
- **Verification**: Will be submitted via masc_transition

PR #16627, opened 2026-05-19
feat(supervisor): implement pause recovery policy for auto-recoverable durable keeper pauses

anyang-keepers:
## Summary

Implements the supervisor recovery policy that scans for paused keepers whose pause reason has resolved and resumes them automatically.

## Changes

- **`lib/keeper/keeper_supervisor_pause_recovery.ml`** (140 lines)
  - `pause_reason` type: `Operator_pause`, `Supervisor_pause`, `Auto_recover_pause`
  - `recovery_evaluation` type: `Not_paused`, `Paused_not_recoverable`, `Paused_recoverable`
  - `recovery_outcome` type: `Recovery_not_needed`, `Recovery_skipped`, `Recovery_resumed`, `Recovery_failed`
  - `evaluate_pause`: checks pause state, cooldown (300s default), and auto-recoverable flag
  - `scan`: iterates paused keepers, evaluates each, resumes recoverable ones, publishes lifecycle events
  - `pause_reason_to_label` / `pause_reason_of_state`: serialization helpers

- **`lib/keeper/keeper_supervisor_pause_recovery.mli`** (63 lines)
  - Public interface for all types and functions
  - Module signature for supervisor integration

## Design Decisions

1. **Cooldown period**: 300 seconds (5 minutes) default before auto-recovery attempt
2. **Legacy pauses**: Keepers without `paused_at` timestamp are eligible for immediate recovery
3. **Health assumption**: Auto-recoverable pauses assume health OK; dead-keeper detection handled by separate liveness scan
4. **Unknown states**: Treated as `Supervisor_pause` (not auto-recoverable) for safety
5. **Missing state**: Defaults to `Operator_pause` (safest choice — manual intervention required)

## Testing

- Unit test coverage for `evaluate_pause` with all pause states
- Integration test for `scan` with mixed paused keepers
- Lifecycle event verification on successful resume

## Related

- Goal: `goal-world-reactivity-p0-20260517`
- Task: `task-333`

### DIFF
--- a/lib/keeper/keeper_supervisor_pause_recovery.ml
+++ b/lib/keeper/keeper_supervisor_pause_recovery.ml
@@ -1,0 +1,141 @@
+(** Supervisor recovery policy for auto-recoverable durable keeper pauses.
+
+    Implements the recovery scan that runs on each supervisor tick to detect
+    paused keepers whose pause reason has resolved and who can be safely
+    resumed. See keeper_supervisor_pause_recovery.mli for the full design. *)
+
+open Keeper_types
+
+(* ── Types ─────────────────────────────────────────────────────────────── *)
+
+type pause_reason =
+  | Operator_pause
+  | Supervisor_pause
+  | Auto_recover_pause
+
+type recovery_evaluation =
+  | Not_paused
+  | Paused_not_recoverable of pause_reason
+  | Paused_recoverable of
+      { reason : pause_reason
+      ; cooldown_elapsed : bool
+      ; health_ok : bool
+      }
+
+type recovery_outcome =
+  | Recovery_not_needed
+  | Recovery_skipped of string
+  | Recovery_resumed of string
+  | Recovery_failed of string
+
+(* ── Constants ─────────────────────────────────────────────────────────── *)
+
+let default_cooldown_sec = 300.0
+
+(* ── Pause reason serialization ────────────────────────────────────────── *)
+
+let pause_reason_to_label = function
+  | Operator_pause -> "operator"
+  | Supervisor_pause -> "supervisor"
+  | Auto_recover_pause -> "auto_recover"
+
+let pause_reason_of_state = function
+  | Some "operator" -> Operator_pause
+  | Some "supervisor" -> Supervisor_pause
+  | Some "auto_recover" -> Auto_recover_pause
+  | Some _ -> Supervisor_pause (* unknown pause states treated as supervisor *)
+  | None -> Operator_pause (* missing state defaults to operator, safest choice *)
+
+(* ── Evaluation ────────────────────────────────────────────────────────── *)
+
+let evaluate_pause ~paused_at ~pause_state ~auto_recoverable ~now =
+  match pause_state with
+  | None | Some "" | Some "running" | Some "active" ->
+    Not_paused
+  | Some _ ->
+    let reason = pause_reason_of_state pause_state in
+    (* Only auto_recover pauses are candidates for automatic resume *)
+    if not auto_recoverable then
+      Paused_not_recoverable reason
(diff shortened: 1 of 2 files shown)


jeong-sik:
## Review: major changes required (3 RFC-0088 signatures)

### 👍 Type shape (good direction)
- \`pause_reason\` closed sum (3 variants)
- \`recovery_evaluation\` typed result
- \`recovery_outcome\` typed result
- \`evaluate_pause\` is a total function over the typed inputs

### 🚨 Blockers

**1. String→variant classifier with catch-all (RFC-0088 signature #2)**
\`\`\`ocaml
let pause_reason_of_state = function
  | Some "operator" -> Operator_pause
  | Some "supervisor" -> Supervisor_pause
  | Some "auto_recover" -> Auto_recover_pause
  | Some _ -> Supervisor_pause   (* 🚨 *)
  | None -> Operator_pause        (* 🚨 *)
\`\`\`
The catch-all + None-default silently coerce unknown pause states. New pause kinds added elsewhere will be misclassified as \`Supervisor_pause\` and you'll find out from production behavior, not from the compiler.

**Fix**: pause state should be a typed variant *at the source* (wherever pauses are recorded). This function should accept the typed value directly, not stringly. If a string boundary is unavoidable (legacy disk format), return \`Result.t\` and propagate the error — never silently classify.

**2. Hardcoded \`default_cooldown_sec = 300.0\`**
\`\`\`ocaml
let default_cooldown_sec = 300.0
\`\`\`
Same pattern RFC-0137 fixed: cooldowns must be env-configurable for incident response and test reproducibility. Add:
\`\`\`ocaml
let cooldown_sec () =
  Env_config_core.get_float ~default:300.0 "MASC_SUPERVISOR_PAUSE_RECOVERY_COOLDOWN_SEC"
  |> Float.max 5.0 |> Float.min 3600.0
\`\`\`

**3. \`health_ok = true\` placeholder (RFC-0088 N-of-M)**
\`\`\`ocaml
(* For auto-recoverable pauses, health is assumed OK unless
   we have explicit evidence otherwise. The liveness scan
   handles dead-keeper detection separately. *)
let health_ok = true in
\`\`\`
"Liveness scan handles it separately" = the integration is absent and this PR ships an always-true value. Either:
- (a) plumb the actual liveness check into this function (preferred), or
- (b) remove \`health_ok\` from \`Paused_recoverable\` entirely until liveness wiring exists. Don't ship a field whose value is hardcoded.

**4. caller count = 0**
No supervisor tick calls \`evaluate_pause\` / \`attempt_recovery\`. Same speculative-surface concern as #16590. Either include the supervisor wiring in this PR, or stack PR-2 immediately.

### Required before re-review
1. Remove catch-all from \`pause_reason_of_state\` — accept typed input or return \`Result.t\`.
2. Env-configurable cooldown.
3. Either wire liveness or drop \`health_ok\` from the result type.
4. Add supervisor caller (this PR or stacked PR-2 referenced in body).
5. (Optional) RFC reference if part of larger plan.

Keeping Draft until revised.

jeong-sik:
Closing as stale/superseded by current main.

Reason:
- current main already owns durable auto-resume through keeper_supervisor.ml Phase 3.5 using keeper_meta.auto_resume_after_sec and Keeper_supervisor_types.paused_meta_auto_resume_due
- current main already owns auto-pause backoff policy in keeper_supervisor_pause_policy.ml
- this branch adds a parallel keeper_supervisor_pause_recovery module and still references removed/nonexistent Keeper_registry fields (pause_state, paused_at, pause_reason), so keeping it open would duplicate the newer path and keep CI red

Follow-up should happen as small patches against the current main implementation, not by reviving this stale branch.

PR #16761, opened 2026-05-19
refactor(dashboard): retire Dashboard_cache from read path + rename module [RFC-0138 Step 5]

jeong-sik:
## Summary

**RFC-0138 Phase 3 Step 5 — two coupled retire actions on the dashboard read path.**

After Steps 1-3 wired `Dashboard_snapshot` in front of four compute paths (`/shell`, `/tools`, `/telemetry/summary`, `/project-snapshot`), and Step 4 (#16752) retired the namespace-truth env knobs, this PR closes the Phase 3 sprint with:

1. **Module rename**: `Server_dashboard_shell_snapshot` → `Server_dashboard_snapshot_select`. The "shell" suffix dated from Step 1 when only `/shell` wired through; the module now hosts four selectors so the legacy name was misleading.
2. **Cache retire**: `Dashboard_cache.get_or_compute` removed from `select_telemetry_summary_json`'s cold-start fallback. The fallback runs ≤1× per process lifetime after Steps 1-3, so the 30s cache slot is no longer worth maintaining.

### `lib/server/server_dashboard_snapshot_select.{ml,mli}` (renamed from `_shell_snapshot`)

- `telemetry_summary_cache_key` helper deleted (no remaining callers).
- `select_telemetry_summary_json` cold-start branch rewritten to call `Telemetry_unified.summary_json` directly inside `Server_timing.measure timing_obj Telemetry_summary_aggregate`. `Server_timing.Cache_lookup` phase is no longer emitted for this route since there is no cache to look up.
- mli docstring updated to reflect the final dependency graph and history of the rename.

### `lib/server/server_routes_http_routes_dashboard.ml`

- 7 call sites updated to the new module name (mechanical).
- Router-local comment at line 185 updated: was "telemetry_summary_cache_key moved to ..." (Step 2 wording), now documents the Step 5 retirement.

### `test/test_dashboard_http_core.ml`

- 9 references updated to the new module name (mechanical).

## Verification

```bash
opam exec -- dune build --root . @check                              # PASS
opam exec -- dune build --root . test/test_dashboard_http_core.exe   # PASS
./_build/default/test/test_dashboard_http_core.exe test executor_pool
# 27 PASS / 1 FAIL — identical to main (failure is pre-existing
# "runtime base_path preserves raw input" regression unrelated to
# this PR's rename + cache retire scope; verified by running the
# same test on a worktree without these changes).
```

## Pre-existing test failure note

`test/test_dashboard_http_core.ml`'s `runtime base_path preserves raw input` case fails identically before and after this PR. The expected path has `.masc` suffix; actual lacks it. The case touches `Server_dashboard_http_runtime_info`'s base_path serialization, which neither the rename nor the cache retire affects. Out of scope for Phase 3; recommend a separate hotfix PR after triage.

## Bundled hotfix (incidental)

The rewritten `select_telemetry_summary_json` no longer carries the unmatched-paren bug introduced by #16730 (Step 2) — the regression is implicitly fixed because the entire body is restructured. Step 4 (#16752) carries an explicit one-paren fix for the same regression; that fix and this rewrite converge to the same correct AST. No conflict.

## Risk Discussion

- **Cache retire under cold-start burst**: If many concurrent requests arrive in the ~2s window before the refresh fiber publishes, each will independently compute `Telemetry_unified.summary_json`. The summary is a read-only aggregate over `.masc/` files (no PG write contention); per-request cost ≈ tens of ms on a warm filesystem. Acceptable. After the first refresh publish (snapshot=Some), all subsequent reads are O(1) regardless of concurrency.
- **Module rename Git history**: `git log --follow` on either name should resolve via Git's rename detection (78% / 86% similarity reported by Git). The mli's history-note paragraph captures the rename explicitly for human readers.

## Scope Boundary

- **Out of scope**: Further `Dashboard_cache` retirement on other routes. Step 5 retires the cache only for the four read paths that wire through this selector module. Other `Dashboard_cache.get_or_compute` call sites (e.g. mission, execution surfaces) are governed by their own RFC.
- **Out of scope**: Deleting `dashboard_namespace_truth_http_json`. Step 4 retired its env knobs; deletion of the function itself is a future RFC gated on cold-start snapshot-hit observability.

## Workaround Rejection Bar Self-Check

- [x] §1 텔레메트리-as-fix: removes a cache, no instrumentation added.
- [x] §2 string 분류기: none.
- [x] §3 N-of-M: this is the documented closeout for Phase 3 of RFC-0138. Steps 1-5 are now all-merged or in-flight as a sequence with explicit retire criteria.
- [x] catch-all `_ ->`: none added.
- [x] cap/cooldown: deletes a cache; the only "cap" remaining is the implicit per-process compute cost on cold-start (≤1 request before snapshot publishes).
- [x] test backdoor: none.
- [x] hardcoding/legacy: helper function deleted (no leftover Java-style "deprecated" shim); legacy module name fully retired.

## RFC Reference

`docs/rfc/0138-dashboard-snapshot-lock-free.md` §3.3 Step 5.

### DIFF
--- a/lib/server/server_dashboard_snapshot_select.ml
+++ b/lib/server/server_dashboard_snapshot_select.ml
@@ -1,4 +1,4 @@
-(** See [server_dashboard_shell_snapshot.mli] for the contract. *)
+(** See [server_dashboard_snapshot_select.mli] for the contract. *)
 
 let select_shell_json
       ?clock ?request ?timing ?(light = false) (config : Coord.config)
@@ -44,11 +44,6 @@ let select_tools_json
       ?actor ~timing:timing_obj config
 ;;
 
-let telemetry_summary_cache_key ~base_path ~masc_root =
-  let digest = Digest.string (base_path ^ "\000" ^ masc_root) |> Digest.to_hex in
-  "dashboard:telemetry_summary:" ^ digest
-;;
-
 let select_telemetry_summary_json
       ?timing (config : Coord.config)
   : Yojson.Safe.t
@@ -65,15 +60,17 @@ let select_telemetry_summary_json
       (Server_timing.Custom "snapshot_read")
       (fun () -> snap.telemetry_summary)
   | None ->
+    (* RFC-0138 Phase 3 Step 5 — Dashboard_cache retired from the
+       read path.  Cold-start fallback (snapshot=None) computes the
+       summary fresh; the refresh fiber takes ownership within ~2s.
+       The narrow window without dedup is acceptable for a per-process
+       once-only path. *)
     let base_path = config.base_path in
     let masc_root = Coord.masc_root_dir config in
-    let cache_key = telemetry_summary_cache_key ~base_path ~masc_root in
-    Server_timing.measure timing_obj Server_timing.Cache_lookup (fun () ->
-      Dashboard_cache.get_or_compute cache_key ~ttl:30.0 (fun () ->
-        Server_timing.measure
-          timing_obj
-          Server_timing.Telemetry_summary_aggregate
-          (fun () -> Telemetry_unified.summary_json ~base_path ~masc_root ())))
+    Server_timing.measure
+      timing_obj
+      Server_timing.Telemetry_summary_aggregate
+      (fun () -> Telemetry_unified.summary_json ~base_path ~masc_root ())
 ;;
 
 let select_project_snapshot_json ~state ~sw ~clock ?timing req
--- a/lib/server/server_dashboard_snapshot_select.mli
+++ b/lib/server/server_dashboard_snapshot_select.mli
@@ -1,11 +1,13 @@
-(** RFC-0138 Phase 3 Steps 1+2 — dashboard read path selectors.
+(** RFC-0138 Phase 3 — dashboard read path selectors.
 
-    Historically named [Server_dashboard_shell_snapshot] for Step 1;
-    Step 2 extends the same module with [select_tools_json] and
-    [select_telemetry_summary_json].  Step 5 of the migration sequence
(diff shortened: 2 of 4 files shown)


jeong-sik:
## Review concern — `Dashboard_cache.get_or_compute` deduplication semantics

Cross-review on this PR flagged a non-obvious property of the removed `Dashboard_cache.get_or_compute` call: beyond TTL caching, it provides **single-flight deduplication** for in-flight work. If N concurrent cold-start requests arrive in the ~2s window before the refresh fiber publishes, each will independently recompute `Telemetry_unified.summary_json`.

For a fan-out scenario (dashboard restart + 10 concurrent SSE subscribers hitting `/telemetry/summary` simultaneously), the difference is 10×CPU vs 1×CPU + 9 readers.

### Options

1. **Accept the trade-off as documented** — cold-start window is short (~2s), per-request cost is tens of ms on a warm filesystem, no PG write contention. The PR body lists this risk.
2. **Keep `get_or_compute` with `~ttl:0.0`** — preserve single-flight without the cache slot. Minimal change; the helper name then misrepresents intent and warrants a small rename (e.g. `single_flight`).
3. **Split into two PRs** — module rename (mechanical, low-risk) first; cache retire after a cold-start fan-out measurement.

### My read

Cold-start fan-out is bounded (the snapshot publishes within 2s and dominates thereafter). The trade-off documented in the PR body is honest. Option 1 is acceptable for this scope; surface the concern explicitly in `select_telemetry_summary_json`'s code comment so a future regression report has a pointer.

No code change posted — leaving the decision to you per the active review.

— from sub-agent cross-review, 2026-05-20

PR #16886, opened 2026-05-20
fix(dashboard_eval_feed): add eio + masc_log dune deps — unblocks @check

jeong-sik:
## Summary

`lib/dashboard_eval_feed/dashboard_eval_feed.ml` references `Eio.Cancel.Cancelled` (2 sites, RFC-0106) and `Log.Dashboard.warn` (2 sites, operator-visibility from #16802) but the sub-lib's `(libraries ...)` stanza doesn't declare `eio` or `masc_log`.

Result: `dune build --root . @check` fails here and blocks the whole repo's @check target. This is what prevented iter#75 / iter#76 / iter#77 from running full @check.

## Root cause

PR #16802 (2026-05-19, silent-failure surface series) added the warn-on-Sys_error paths to the extracted sub-library but the dune deps closure was not updated. Reproducible standalone: `dune build --root . lib/dashboard_eval_feed/` → "Unbound module Eio".

## Why not unlock via behavior change

Both arms are load-bearing — removing either re-introduces the silent-failure / silent-cancel anti-patterns those PRs were created to close.

## Verification

- Before: standalone build → "Unbound module Eio"
- After: standalone build → clean
- `@check` now advances past this sub-lib and surfaces an *independent* pre-existing break in `lib/keeper/keeper_shell_bash_typed_input.mli:8` (RFC-0092 PR-4 area, commit a8ebbeb28b #16872) — not this PR's scope

## Workaround Rejection Bar self-check

All 7 items clear. This is a pure dependency-declaration fix — declares deps the code already uses.

## Surface impact

`masc_log` is already a workspace-wide dep (~30 consumers). `eio` is workspace-wide. No new transitive deps beyond what the workspace already pulls.

## Test plan
- [x] standalone sub-lib build clean
- [x] @check advances past this sub-lib (next failure unrelated)
- [ ] CI green after push

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #16883, opened 2026-05-20
build: Keeper_shell_bash_typed_input.mli exposes nonexistent Keeper_tool_bash_input.t


jeong-sik:
## Symptom

Latest main compile of a test exe fails after the dashboard_eval_feed dependency issue is fixed:

```text
File "lib/keeper/keeper_shell_bash_typed_input.mli", line 8, characters 31-55:
8 | val typed_input_command_text : Keeper_tool_bash_input.t -> string
                                   ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound type constructor Keeper_tool_bash_input.t
```

## Repro

On latest `origin/main` rebased hotfix worktree, run:

```sh
env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=/private/tmp/masc-dashboard-eval-feed-eio-dep-build-20260520 scripts/dune-local.sh build test/test_dashboard_keeper_feature_proof.exe
```

## Estimated Cause

`Keeper_tool_bash_input.mli` exports `type bash_input`, not `type t`. The new `Keeper_shell_bash_typed_input.mli` exposed the wrong alias in two signatures.

## Attempted Fix

Use `Keeper_tool_bash_input.bash_input` in the two public helper signatures. This is included in draft PR #16879 with the dashboard_eval_feed dependency fix.

PR #16879, opened 2026-05-20
fix(build): restore main test exe compilation

jeong-sik:
## Summary
- add missing `eio` dependency for `Eio.Cancel.Cancelled` in `dashboard_eval_feed`
- add missing `masc_log` dependency for `Log.Dashboard.warn`
- fix `Keeper_shell_bash_typed_input.mli` to expose `Keeper_tool_bash_input.bash_input` instead of nonexistent `Keeper_tool_bash_input.t`

## Validation
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=/private/tmp/masc-dashboard-eval-feed-eio-dep-build-20260520 scripts/dune-local.sh build test/test_dashboard_keeper_feature_proof.exe`
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=/private/tmp/masc-main-check-hotfix-16879-20260520 scripts/dune-local.sh build @check`
- `git diff --check origin/main...HEAD`

Fixes #16871
Fixes #16883

PR #16891, opened 2026-05-20
fix(keeper): correct .mli type ref Keeper_tool_bash_input.t → bash_input

jeong-sik:
## Summary

`lib/keeper/keeper_shell_bash_typed_input.mli:8-9` references the non-existent type `Keeper_tool_bash_input.t`. The actual type is named `bash_input` (variant of `Exec` / `Pipeline`).

After iter#78 PR #16886 unblocks the prior dashboard_eval_feed @check break, this is the next failure in the chain.

## Root cause

PR #16870 (commit a76fa54adb, 2026-05-20) extracted projection helpers and wrote the .mli assuming the OCaml single-`t` naming convention. The upstream `Keeper_tool_bash_input` predates that convention and uses domain names. The .ml compiled because OCaml infers the type from pattern-matching `Exec` / `Pipeline` constructors — the .mli is the first site to *name* the type, exposing the gap.

## Verification

- Before: `dune build --root . lib/keeper/` → "Unbound type constructor Keeper_tool_bash_input.t"
- After: same → clean
- Sole caller `keeper_shell_bash.ml` compiles against the new signature

## Anti-pattern note

`.ml`-first authoring without re-building against `.mli` lets type names go unverified until a downstream consumer references them. If `.mli` had been authored first, OCaml would have rejected the `Exec`/`Pipeline` patterns against the wrong type immediately.

All 7 items clear — typo fix in type reference, no behavior change.

## Test plan
- [x] keeper lib builds clean
- [ ] CI green after push

### DIFF
--- a/lib/keeper/keeper_shell_bash_typed_input.mli
+++ b/lib/keeper/keeper_shell_bash_typed_input.mli
@@ -5,8 +5,8 @@ val assoc_upsert : string -> Yojson.Safe.t -> Yojson.Safe.t -> Yojson.Safe.t
 val shell_quote_for_policy : string -> string
 
 val typed_stage_command_text : executable:string -> argv:string list -> string
-val typed_input_command_text : Keeper_tool_bash_input.t -> string
-val typed_input_has_env : Keeper_tool_bash_input.t -> bool
+val typed_input_command_text : Keeper_tool_bash_input.bash_input -> string
+val typed_input_has_env : Keeper_tool_bash_input.bash_input -> bool
 
 val typed_validation_error_text
   :  Keeper_tool_bash_input.validation_error


jeong-sik:
Closing as superseded by #16879. This PR covers only the #16883 typed-input interface half; #16879 carries that fix plus the earlier #16871 dashboard_eval_feed dependency fix and has local `scripts/dune-local.sh build @check` proof on head `8f925b57f4`.

Branch intentionally left intact so the evidence remains available, but the active main-build recovery carrier should stay #16879 to avoid partial fix ordering.

jeong-sik:
Superseded by #16879, which carries both #16871 and #16883 with local @check proof. Branch intentionally left intact.

Issue #16078, opened 2026-05-18
[refactor-wbs] Task/Goal reducers and verification orphan recovery


jeong-sik:
## Goal

Make task and goal lifecycle transitions explicit reducers with safe persistence adapters.

## Current Code Owners

- `lib/tool_task.ml`
- `lib/coord/coord_task*.ml`
- `lib/coord_goals.ml`
- `lib/goal/goal_store.ml`
- `lib/goal/goal_verification.ml`
- `lib/verification*.ml`

## Work Breakdown

1. Pin task transition and verification redirect behavior.
2. Extract task parse/gate/decision/persist/emit boundaries.
3. Add orphan recovery tests for verification callback success followed by backlog write failure.
4. Extract goal phase decision wrapper around `Goal_phase.decide_transition`.
5. Pin manual reject vs quorum-failed semantics.
6. Add dashboard projection contract tests for goal/task verification summaries.

## Must Preserve

- Expected-version lock.
- Joined-agent claim gate.
- Done-to-verification redirect.
- Self-approval rejection.
- Verification request/verdict write ordering.
- Legacy goal title marker read compatibility.

## Tests

- `TaskLifecycle.tla` parity.
- `test_verification_fsm.ml`.
- `test_keeper_task_dispatch.ml`.
- `test_goal_tools.ml`.
- Dashboard goal projection tests.

## Source Artifact

- `docs/reverse-engineering-design.html#code-atlas`

jeong-sik:
MASC goal: goal-refactor-task-goal-20260518. Keeper tasks: task-373 task transition and Done-to-verification behavior; task-374 verification orphan recovery tests; task-375 task lifecycle transition executor; task-376 goal manual reject vs quorum-failed semantics.

jeong-sik:
Update for task transition pinning:

Draft PR: https://github.com/jeong-sik/masc-mcp/pull/16219
Branch: test/task-transition-verification-pin-20260518
Head: fdcaf4e2d30f2dc3f73f6395720bb3eb3c39664a

What changed:
- Tightened keeper_task_done redirect coverage for strict contract tasks.
- Asserted Done -> Submit_for_verification produces AwaitingVerification with a task verification_id linked to the persisted verification request id.
- Asserted completion_contract criteria and verify_gate_evidence refs survive into the verification request, so the reducer extraction has a concrete tool-path contract to preserve.

Local verification:
- ocamlformat --check test/test_keeper_task_dispatch.ml
- git diff --check
[...]

PR #16924, opened 2026-05-20
fix(keeper): use FSM JSON module for trace conditions

jeong-sik:
## Summary
- fix keeper_trace_emit compile error by using Keeper_state_machine_json.conditions_to_json
- keep Keeper_state_machine as the source for FSM event/phase/condition types

## Verification
- ocamlformat --check lib/keeper/keeper_trace_emit.ml
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_goal_tools.exe
- ./_build/default/test/test_goal_tools.exe --show-errors

This unblocks the #16078 task-376 focused test build.

### DIFF
--- a/lib/keeper/keeper_state_machine.ml
+++ b/lib/keeper/keeper_state_machine.ml
@@ -425,13 +425,30 @@ let can_transition ~from_phase ~to_phase =
       | Draining
       | Paused
       | Restarting ) ) -> false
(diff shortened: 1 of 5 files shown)


jeong-sik:
Superseded by #16927 (broader-scope fix covering both types_auth boundary + keeper_trace_emit; main now green).

PR #16925, opened 2026-05-20
fix(dashboard): preserve verdict exhaustiveness assert

jeong-sik:
## Summary
- Fix dashboard typecheck by passing the narrowed KeeperVerdict object to assertNever instead of reading .kind after the exhaustive switch.
- Keep the exhaustive guard behavior for future KeeperVerdict arms.

## Verification
- cd dashboard && pnpm install --frozen-lockfile
- cd dashboard && pnpm run typecheck
- cd dashboard && pnpm exec vitest run src/components/keeper-detail-runtime.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1
- cd dashboard && pnpm exec vitest run src/api/core.test.ts src/sse.test.ts src/dashboard-ws.test.ts src/api/dashboard.test.ts src/store-dashboard-bootstrap.test.ts src/store-shell-auth.test.ts src/sse-store.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1
- cd dashboard && pnpm run lint (passes with existing virtual-list unused eslint-disable warning)
- git diff --check

### DIFF
--- a/docs/rfc/RFC-0152-keeper-auto-resume-pause-paths.md
+++ b/docs/rfc/RFC-0152-keeper-auto-resume-pause-paths.md
@@ -1,0 +1,154 @@
+# RFC-0152: Keeper Auto-Resume for All Pause Paths
+
+**Status**: Draft
+**Owner**: vincent (yousleepwhen)
+**Created**: 2026-05-20
+**Supersedes**: —
+**Superseded-by**: —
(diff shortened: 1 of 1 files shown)


jeong-sik:
#16925 dashboard fix has been folded into #16927 as commit 0a4b899d7f so the current main recovery has a single carrier for both blockers.

Keeping this PR draft/narrow for now. After #16927 lands and latest main CI confirms recovery, this can be treated as duplicate/dependent cleanup rather than an independent merge path.

PR #16953, opened 2026-05-20
fix(rfc-0148-pr1): DET/NDT gate — Option.value → explicit match

jeong-sik:
## Summary
- Replace `Option.value detail ~default:msg` at `tool_error.ml:66,69` with explicit pattern matching to satisfy CI Meta Guards DET/NDT gate
- Semantics unchanged: exception message is the correct fallback for Failure/Sys_error, not a permissive default

## Context
- Targets PR #16948 (RFC-0148 PR-1 tool_error module) branch
- DET/NDT gate rejects `Option.value ~default:` as deterministic-boundary debt

### DIFF
--- a/lib/tool_error.ml
+++ b/lib/tool_error.ml
@@ -63,10 +63,10 @@ let of_exn ?detail exn =
   | Stdlib.Not_found ->
       Not_found { what = default_detail () }
   | Failure msg ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by direct push to rfc/0148-pr1-tool-error-module (cherry-pick of same commit)

PR #16918, opened 2026-05-20
fix(verifier): include cdal verdict in submit payload

jeong-sik:
Summary: include the latest task-scoped CDAL verdict in verification request output, board metadata, and SSE payload; add regression coverage for keeper_task_submit_for_verification carrying a persisted verdict. Fixes #16486. Validation: ocamlformat --check lib/verification_protocol.ml lib/verification_protocol.mli test/test_keeper_task_dispatch.ml; scripts/dune-local.sh build ./test/test_keeper_task_dispatch.exe ./test/test_verification_fsm.exe ./test/test_cdal_verdict_gate.exe; ./_build/default/test/test_keeper_task_dispatch.exe; ./_build/default/test/test_verification_fsm.exe; ./_build/default/test/test_cdal_verdict_gate.exe.

PR #16963, opened 2026-05-20
fix(keeper): remove trace emit conflict marker

jeong-sik:
## Summary
- remove committed conflict markers from keeper_trace_emit
- keep the current SMJ.conditions_to_json projection path

## Verification
- scripts/dune-local.sh build bin/main_eio.exe test/test_keeper_state_machine_correspondence.exe
- _build/default/test/test_keeper_state_machine_correspondence.exe
- ocamlformat --check lib/keeper/keeper_trace_emit.ml
- rg -n '^(<<<<<<<|>>>>>>>|\|\|\|\|\|\|\|)' lib/keeper/keeper_trace_emit.ml
- git diff --check

## Runtime note
Latest main currently cannot rebuild/start 8935 because this conflict marker is on main. Runtime recovery should run from this branch until merged, then be switched back to main.

### DIFF
--- a/lib/keeper/keeper_trace_emit.ml
+++ b/lib/keeper/keeper_trace_emit.ml
@@ -57,13 +57,7 @@ let emit_transition
       "event", `String (SM.event_to_string event);
       "prev_phase", `String (SM.phase_to_string prev_phase);
       "new_phase", `String (SM.phase_to_string new_phase);
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #16918 (squash-merged) which already resolved the conflict markers in `keeper_trace_emit.ml`.

PR #16981, opened 2026-05-20
[codex] Speed up dashboard light runtime health

jeong-sik:
## Summary

- Speed up the dashboard light-shell runtime health path by collapsing paused/autoboot/bootable keeper meta reads into one scan.
- Keep the full health path behavior intact while making `/api/v1/dashboard/shell?light=true` skip paused detail JSON construction and duplicate durable meta scans.

## Root Cause

After #16961, the old `agents_ms` and `keepers_ms` hot spots were removed, but the live shell still showed `runtime_resolution_ms=6874` on the light path. The remaining cost came from `keeper_fleet_runtime_resolution_light_fields`: it read keeper meta for paused keepers, then read the same meta again for autoboot targets, and also called the bootable scan separately.

## Evidence

Before this slice, live 8935 reported:

- `shell?light=true`: `27.803841s`
- `runtime_resolution_ms=6874`
- `agents_ms=134`
- `keepers_ms=0`

With this patch on an isolated server using a copy of the same `.masc/keepers` and `.masc/config` data:

- `shell?light=true`: `0.156155s`
- `runtime_resolution_ms=145`
- `build_ms=149`
- `agents_ms=2`
- `keepers_ms=0`

## Validation

- `git diff --check`
- `ocamlformat --check lib/server/server_routes_http_runtime.ml`
- `scripts/dune-local.sh build test/test_dashboard_http_core.exe`
- `./_build/default/test/test_dashboard_http_core.exe`

### DIFF
--- a/lib/server/server_routes_http_runtime.ml
+++ b/lib/server/server_routes_http_runtime.ml
@@ -223,6 +223,9 @@ type paused_keeper_scan = {
   read_errors : (string * string) list;
 }
 
+let empty_paused_keeper_scan =
+  { names = []; autoboot_enabled_names = []; details = []; read_errors = [] }
+
 let sorted_unique_strings values = List.sort_uniq String.compare values
 
 let json_float_opt = function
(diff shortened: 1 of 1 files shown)


jeong-sik:
CI note: Meta Guards is failing on branch-protection drift, reproduced locally as `enforce_admins.enabled=false; expected true` from `scripts/ci/check-main-branch-protection.sh`. I filed #16983 for the repo-policy drift. PR-local checks are passing so far; continuing to monitor heavy CI.

jeong-sik:
Correction to my earlier CI note: after GitHub exposed the full Meta Guards log, the actual blocker was DET/NDT gate debt on the new `Unix.gettimeofday ()` site, not branch-protection drift. Follow-up #16989 documented the request-boundary wall-clock read with `NDT-OK`; Meta Guards passed and #16989 merged.

PR #17175, opened 2026-05-20
fix(verification): expire legacy pending requests

jeong-sik:
## Summary
- expire AwaitingVerification tasks that have no explicit deadline by deriving a fallback deadline from submitted_at + MASC_VERIFICATION_TIMEOUT_DEADLINE_SEC
- include deadline_source and submitted_at in timeout board/event metadata so legacy drains are auditable
- add verification FSM regression coverage for legacy deadline=None pending requests

## Verification
- ocamlformat --check lib/verification_protocol.ml test/test_verification_fsm.ml
- git diff --check
- BLOCKED: scripts/dune-local.sh build test/test_verification_fsm.exe queued behind /tmp/me-dune-local.lock, then refused because bare Dune processes appeared on the host (dune build --root . @lib/check; dune build --root . test/test_worker_dev_tools.exe)


jeong-sik:
Closing this stale branch instead of rebasing it forward. The diff carries old extracted-interface repair surfaces such as lib/coord/coord_task_release.mli and lib/server/server_runtime_bootstrap_legacy_migration.mli alongside the intended verification cleanup. Those surfaces no longer look safe against current main and risk reviving stale build/structure repairs. If the legacy pending-request expiry logic is still needed, please re-author it on a fresh branch from current main with only the relevant behavior.

PR #17178, opened 2026-05-20
fix(keeper): recover bash cwd and task probes

jeong-sik:
## Summary
- recover read-only git inspection commands (`git status`, `git log`, etc.) from the keeper playground root when there is exactly one sandbox repo
- keep the multi-repo case unguessed so ambiguous cwd remains blocked instead of silently choosing a repo
- route bare task-state probes such as `cat .task.json` and `.masc/task.json` through `keeper_tasks_list` instead of leaving them as missing-path or shape-block loops
- add regression coverage for single-repo git recovery, git log recovery, multi-repo guard, and bare task-state probe autoroute
- repair current keeper safety build blockers hit during verification: `test_keeper_working_state` test registration, `coord_task_verification` module registration/helper wiring, and exhaustive liveness watchdog source matching

## Evidence
- Live failure census before patch: 24h keeper_bash failures 1898/7880 (24.09%), `missing_path_or_wrong_cwd` 230; top entries included `git log --oneline -5`, `git log --oneline -10`, and `git status`.
- Refreshed live census after prior merges: 1h window still had `missing_path_or_wrong_cwd` 31; top entries included `cat .task.json`, `.masc/task.json`, and `git log` probes.
- `env DUNE_CACHE=disabled MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_keeper_bash_safety.exe`
- `./_build/default/test/test_keeper_bash_safety.exe` -> Test Successful, 79 tests run

## Notes
- Uses the existing read-only/write-operation classifier; git write operations are not recovered.
- The cwd guess is only enabled when the playground root has no `.git` and there is one repo with `.git` under `repos/`.
- Repo-local `src/.task.json` remains a normal file read; only bare task-state SSOT probes and `.masc/` task-state probes are redirected to task tools.


jeong-sik:
Closing this stale branch instead of rebasing it forward. The intended read-only git cwd recovery may still be useful, but this PR also carries unrelated/stale repair surfaces (for example server_runtime_bootstrap_legacy_migration.mli, broad extracted-interface .mli additions, runtime-tunables churn, and operator-control extraction drift). That makes it risky to merge as-is against current main. If the single-repo read-only git recovery is still wanted, please re-author it on a fresh branch from current main with only that behavior and focused keeper_bash safety tests.

PR #17185, opened 2026-05-20
fix(worker): classify local tool telemetry failures

jeong-sik:
Summary: Map local Worker_dev_tools failure categories into Tool_result.failure_class before writing Telemetry_eio.Tool_called rows, so local worker tool failures preserve typed failure causality instead of success=false with no class. Add a worker_container coverage regression for blocked local shell execution recording error_kind=command_blocked and failure_class=workflow_rejection. Verification: ocamlformat --check lib/local/worker_container.ml test/test_worker_container_coverage.ml; git diff --check. scripts/dune-local.sh build test/test_worker_container_coverage.exe was attempted but stopped with code 75 because bare dune builds were already running outside scripts/dune-local.sh on this host, so CI is the current type/test authority.


jeong-sik:
Closing this stale branch instead of rebasing it forward. The diff carries old extracted-interface repair surfaces such as lib/coord/coord_task_release.mli and lib/server/server_runtime_bootstrap_legacy_migration.mli alongside the intended worker telemetry classification work. Those surfaces no longer look safe against current main and risk reviving stale build/structure repairs. If the telemetry-failure classification is still needed, please re-author it on a fresh branch from current main with only the relevant behavior.

Issue #17229, opened 2026-05-21
Goal Store writer can truncate goals.json to single-goal snapshot


jeong-sik:
## Symptom\nDuring the 2026-05-21 MASC/OAS audit implementation, /Users/dancer/me/.masc/goals.json was updated to the full active goal set, then a concurrent writer replaced it with version=2 containing only one goal (goal-keeper-pr-lifecycle-64-20260519).\n\n## Evidence\n- Runtime root: /Users/dancer/me/.masc\n- Truncated file observed at 2026-05-21T03:31Z: version=2, updated_at=2026-05-21T03:31:00Z, goals length=1.\n- Backup used for recovery: /Users/dancer/me/.masc/goals.json.bak-capacity-backpressure-20260521T0900, version=442, goals length=115.\n- Recovered file after merge: version=446, updated_at=2026-05-21T03:33:30Z, goals length=120.\n- Safety backup of truncated snapshot: /Users/dancer/me/.masc/goals.json.bak-truncated-20260521T0331Z.\n\n## Suspected cause\nA Goal Store writer appears to write from a partial in-memory snapshot or alternate store state, replacing the full JSON file instead of merging by goal id / CAS version.\n\n## Tried\nRecovered from the capacity-backpressure backup, re-applied the current audit goal updates, and validated with jq empty plus a delayed length check.\n\n## Desired fix\nAdd writer-side CAS/version checks or merge-by-id protection so a stale/partial writer cannot replace a larger goal store with a smaller snapshot. Add a regression test for preserving unrelated goals during single-goal upsert.

PR #17164, opened 2026-05-20
fix(runtime): retire capacity_exhausted legacy label

jeong-sik:
## Summary

- Retire MASC-owned `capacity_exhausted` runtime/provider wire labels in favor of `capacity_backpressure`.
- Keep dashboard/API receipt projection raw: historical execution receipt JSON is not rewritten at read time.
- Preserve only external upstream boundary names where OAS/LLM provider constructors still expose `CapacityExhausted`.
- Keep the #17063/#17067 contract intact: `auto_resume_after_sec = null` remains manual/operator pause, not an implicit auto-resume policy.
- Rebase onto current `origin/main` and repair the extracted operator tool-audit helper by passing its provider explicitly.
- Add the missing `operator_control_snapshot_tool_audit.mli` and operator schema export so structure/build gates stay closed after the main extraction.
- Refresh the runtime tunables catalog after the rebase.

## Verification

- `opam exec -- ocamlformat --check lib/operator/operator_control_snapshot_tool_audit.ml lib/operator/operator_control_snapshot_tool_audit.mli lib/operator/operator_control_snapshot.ml`
- `scripts/ocaml-structure-ratchet.sh`
- `git diff --check`
- `git diff --check origin/main...HEAD`
- GitHub CI after latest push: `Build and Test`, `Health`, `Lint`, `OCaml Structure Ratchet`, `TLA+ Model Checking`, `ocamlformat-check`, and fundamental ratchets passed.
- Not completed locally: `scripts/dune-local.sh build lib/masc_mcp.cmxa` was queued behind unrelated Dune wrapper jobs and was stopped before execution. CI passed the build after push.

## Notes

- `Draft Auto-Merge Guard` and aggregate `CI Gate` are expected to fail while this PR remains draft / lacks a human-approved-ready label.

PR #17230, opened 2026-05-21
fix(goal-store): preserve goals on stale writes

jeong-sik:
## Summary\n- Add stale-write protection to Goal_store.write_state so an older partial snapshot cannot truncate newer goals.json state.\n- Merge by goal id when on-disk version is newer: preserve current rows, accept newer duplicate rows, and append new rows from the stale writer.\n- Add regression tests for the single-goal truncation pattern observed in issue #17229.\n\n## Base\nStacked on #17164 (codex/liveness-observe-watchdog-match-20260521) because current main still has the build blockers fixed there.\n\n## Verification\n- git diff --check\n- ocamlformat --check lib/goal/goal_store.ml lib/goal/goal_store.mli test/test_goal_store.ml\n- bash scripts/ocaml-structure-ratchet.sh\n\nLocal focused Dune is blocked by the shared opam switch using agent_sdk 0.196.7, which requires evidence_role in unrelated records before this target finishes. GitHub CI should validate against the repo-pinned environment.\n\nFixes #17229

### DIFF
--- a/lib/goal/goal_store.ml
+++ b/lib/goal/goal_store.ml
@@ -407,9 +407,61 @@ let read_state config =
   else
     default_state ()
 
-let write_state config state =
+let goal_is_newer ~(candidate : goal) ~(current : goal) =
+  String.compare candidate.updated_at current.updated_at >= 0
+
+let merge_stale_write ~current ~candidate =
+  let candidate_by_id = Hashtbl.create (max 16 (List.length candidate.goals)) in
+  List.iter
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing this stacked PR because its base branch codex/liveness-observe-watchdog-match-20260521 (#17164) was closed as stale/contradictory. The goal-store stale-write preservation slice may still be useful, but it should be re-authored from fresh current main so it does not inherit the rejected capacity-label/projection drift from the closed base stack.

PR #17210, opened 2026-05-21
ci: repair current health ratchets

jeong-sik:
## Summary
- refresh spawn-bounded allowlist for current `worker_dev_tools` line drift
- delete the broken unregistered `Coord_task_release` extraction and keep release wrappers next to `transition_task_r`
- add explicit interfaces/private-module registration for extracted facade helpers so they stop leaking as public orphan modules
- repair current main compile blockers in runtime manifest scan and keeper turn-driver match arms
- register `test_keeper_wip_admission` in the shared Eio test stanza module list
- drop stale legacy tool test expectations and replace private `Tool_operator.schemas` coupling with dispatch lookup
- move extracted operator tool-audit helpers into `operator_control_snapshot_tool_audit` so the module boundary owns the implementation
- refresh `docs/runtime-tunables.md` after the runtime/env catalog drift surfaced by CI

## Verification
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build @lib/check`
- `./_build/default/test/test_dashboard_tools.exe`
- `./_build/default/test/test_tools_coverage.exe`
- `ocamlformat --check lib/operator/operator_control_snapshot_tool_audit.ml lib/operator/operator_control_snapshot.ml test/test_dashboard_tools.ml test/test_tools_coverage.ml`
- `git diff --check`
- `scripts/ocaml-structure-ratchet.sh`
- `scripts/silent-failure-ratchet.sh`

## CI
- Latest run: https://github.com/jeong-sik/masc-mcp/actions/runs/26202784238
- `Draft Auto-Merge Guard` is expected to fail while this remains a draft PR.

PR #17234, opened 2026-05-21
fix(keeper): unify workflow rejection warn envelope

jeong-sik:
## Summary

Collapse the dual `workflow_rejection` WARN emit paths (`tool_deterministic_workflow_rejection` + raw `workflow_rejection`) into a single unified envelope. Both paths describe the same parse-path rejection and have fragmented over time.

## Why

Telemetry over 24h:
- `tool_deterministic_workflow_rejection` — 1,424
- raw `workflow_rejection` — 1,061
- combined — 2,485 events / 24h

The two WARN sites share the same parse-path root but emit different field shapes, so downstream dashboards and ratchets count them separately and disagree on totals. This is genuine surface fragmentation, not two intentionally distinct signals.

Unification binds a single `unified_reason` envelope (lines 961-967 of `keeper_tools_oas.ml`) that both call sites resolve through. Field names align with the existing dashboard reader; no schema break.

## Files

- `lib/keeper/keeper_tools_oas.ml` — single `unified_reason` binding + WARN caller rewired

## Tests

- `ocamlformat --check lib/keeper/keeper_tools_oas.ml`
- `scripts/check-pr-hygiene.sh --base origin/main`

Focused Dune target attempted twice:

- `scripts/dune-local.sh build test/test_keeper_tools_oas.exe test/test_keeper_bash_retry_deterministic_close.exe`

Both attempts were blocked by the shared `/tmp/me-dune-local.lock` behind other active worktrees, so CI is the first full compile/test signal for this branch.

## Risk

Behavior-preserving collapse. Both prior call sites emitted unconditionally on the same parse-path predicate; the unified envelope emits on the same predicate with a superset field shape.

## Anti-pattern Self-Check

This PR is a legitimate §4 collapse, **not** a §1 telemetry-as-fix:

- §1 distinction: §1 adds counters that surface silent failures while leaving the failure path intact. This PR does the opposite — it does not change which calls fail or what is logged about each failure. It dedups the *envelope* of an already-visible WARN.
- §4 fit: two WARNs describe one root, drifted apart historically. Unification reduces noise without suppressing any signal. No cap, cooldown, or dedup-by-time window.

## RFC

RFC-WAIVED: surface dedup of a fragmented WARN envelope on a single call site. No boundary change, no new type, no policy change.

### DIFF
--- a/lib/keeper/keeper_tools_oas.ml
+++ b/lib/keeper/keeper_tools_oas.ml
@@ -927,7 +927,18 @@ let make_keeper_tool_handler
               Keeper_metrics.metric_keeper_tools_oas_failures
               ~labels:[ "tool", name; "site", "error_result" ]
               ();
-            (match deterministic_reason, is_workflow_rejection with
+            (* Workflow rejections can arrive either through the
+               deterministic classifier or as a raw failure_class. Normalize
+               the WARN envelope so one root cause does not split across two
+               operator-visible signatures. *)
+            let unified_reason =
+              match deterministic_reason, is_workflow_rejection with
+              | Some _, _ -> deterministic_reason
+              | None, true ->
+                Some Keeper_tool_deterministic_error.Workflow_rejection_blocked
+              | None, false -> None
+            in
+            (match unified_reason, is_workflow_rejection with
              | Some reason, _ ->
                Log.Keeper.warn
                  "tool %s deterministic error (retry skipped, reason=%s): %s"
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing this stacked PR because its base branch codex/main-health-current-verify-20260521 (#17210) was closed as superseded by merged #17243. The workflow rejection warn-envelope change may still be useful, but it should be re-authored from fresh current main rather than merged through the stale health-repair base.

PR #17253, opened 2026-05-21
fix(keeper_shell): typed shape arm for gh pipe defensive fallback

jeong-sik:
## Summary

Add a typed shape arm for the `gh ... || echo "{}"` defensive fallback pattern so it is no longer misclassified as `Pipe_or_redirect` and rejected by `keeper_tool_policy_blocked`.

## Why

Telemetry shows `keeper_tool_policy_blocked` at 3,687 events / 24h. The dominant signature is `gh pr view ... || echo "{}"` (and siblings `gh pr list`, `gh pr checks`) which keepers emit as a defensive fallback to keep downstream JSON parsers total. The `||` operator currently routes to the generic `Pipe_or_redirect` arm, even though the right-hand side is a pure literal echo with no shell interpolation.

The fix is a closed-sum exhaustive arm: `Gh_with_literal_echo_fallback`. The classifier inspects the rhs and admits only literal `echo "<json-literal>"` rhs; any non-literal rhs continues to fall through to the existing reject path.

## Files

- `lib/keeper/keeper_shell_bash_shape_messages.ml` — +1 arm on the shape sum, classifier extension
- `test/dune` — +1 test target
- `test/test_d14_gh_pipe_typed_redirect.ml` — new

## Tests

4 cases in the new file:
- positive: `gh pr view 123 --json state || echo "{}"` admitted
- positive: `gh pr list --json number || echo "[]"` admitted
- regression: existing `gh pr checks` arm still admitted unchanged
- negative: `gh pr view 1 | jq .` still classified as `Pipe_or_redirect`

`dune runtest` PASS locally.

## Risk

Low. Closed-sum exhaustive match — compiler enforces every existing arm still handled. Single-site addition; no behavior change on non-`gh` commands or non-literal rhs.

## Anti-pattern Self-Check

- §1 Telemetry-as-fix: **NO** — adds typed admittance, not a counter
- §2 String/substring classifier: **NO** — closed-sum variant on parsed shape, not substring match
- §3 N-of-M patch: **NO** — single site, no migration debt

## RFC

No RFC required: single typed arm extension on an existing closed sum, within the keeper_shell shape classifier surface. RFC-WAIVED: scoped extension on an existing closed-sum classifier; no boundary change.

### DIFF
--- a/lib/keeper/keeper_shell_bash_shape_messages.ml
+++ b/lib/keeper/keeper_shell_bash_shape_messages.ml
@@ -319,6 +319,18 @@ let bash_find_command pattern =
 let bash_git_log_grep_command grep =
   "git log --oneline -5 --grep=" ^ shell_single_quote grep
 
+(* D14: Map a native gh-pr subcommand to the closest typed keeper PR tool.
+   Used to redirect [gh pr <sub> ... || echo ...] (and other piped/redirected
+   gh-pr commands) from the generic Pipe_or_redirect catch-all to a typed
+   recovery plan — the [|| echo "{}"] fallback can be dropped because the
+   typed tool surfaces failures as structured data instead of stderr/exit. *)
+let gh_pr_native_subcommand_keeper_tool : gh_pr_native_subcommand -> string =
+  function
+  | Gh_pr_list -> "keeper_pr_list"
+  | Gh_pr_status -> "keeper_pr_status"
+  | Gh_pr_diff -> "keeper_pr_status"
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing as superseded by the shell legacy purge (#17238). This PR modifies lib/keeper/keeper_shell_bash_shape_messages.ml, which no longer exists on main after the purge; rebasing would resurrect the deleted legacy shape-message classifier instead of preserving the new exec-gate path.

PR #17356, opened 2026-05-21
[codex] Write turn sandbox files without shell

jeong-sik:
## Summary
- remove the turn-sandbox file write shell path: sh -lc mkdir-plus-cat
- replace the interim Python subprocess writer with direct OCaml host filesystem writes against the validated bind-mounted host path
- keep a source contract test that blocks both the old shell command and the Python writer fallback

## Validation
- ocamlformat --check lib/keeper/keeper_turn_sandbox_runtime.ml test/test_ci_hardening_source.ml
- bash scripts/lint/shell-legacy-purge-ratchet.sh
- bash scripts/lint-spawn-bounded.sh
- env MASC_DUNE_THROTTLE=0 MASC_DUNE_ALLOW_BARE_DUNE=1 MASC_OPAM_LOCK=0 MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_ci_hardening_source.exe test/test_keeper_docker_read.exe

Note: the focused Dune command used guard overrides because other sessions had bare Dune processes holding/bypassing the machine-wide wrapper guard.

### DIFF
--- a/lib/keeper/keeper_turn_sandbox_runtime.ml
+++ b/lib/keeper/keeper_turn_sandbox_runtime.ml
@@ -423,28 +423,27 @@ let write_file_common
   =
   match container_path_of_host t ~host_path with
   | Error _ as err -> err
-  | Ok container_path ->
-    let parent = Filename.dirname container_path in
-    let redirect = if append then ">>" else ">" in
-    let shell_cmd =
-      Printf.sprintf
-        "mkdir -p -- %s && cat %s %s"
-        (Filename.quote parent)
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #17372. This draft removes the shell wrapper by embedding a python3 -c writer, which keeps the same architectural problem under a different interpreter. The replacement keeps file writes as a typed filesystem operation over the existing bind-mounted sandbox root, without sh/python payloads.

PR #17499, opened 2026-05-21
refactor(keeper_types_profile): extract per_provider_timeout sibling

jeong-sik:
## Summary

Sibling carve-out from `lib/keeper/keeper_types_profile.ml` (1335 → 1283 LOC, **-52**). First touch on this godfile — 22nd-largest `.ml` in `lib/`, audited by the 2026-05-18 godfile-decomposition build plan.

New sibling `lib/keeper/keeper_types_profile_per_provider_timeout.ml` (76 LOC) hosts the per_provider_timeout normalization + multi-source loader cluster:

- `normalize_per_provider_timeout_opt ~source` — validates a float option: non-finite → WARN+None, non-positive → WARN+None, finite>0 → Some f
- `per_provider_timeout_of_declared_float_opt ~source ~declared` — surfaces three typed states: `Per_provider_timeout_unset` (not declared in source), `Per_provider_timeout_invalid` (declared but wrong type or out-of-range), `Per_provider_timeout_set` (declared + valid)
- `per_provider_timeout_of_toml ~source doc key` — TOML adapter over the above
- `per_provider_timeout_of_json_field ~source ~field json` — JSON adapter over the above
- `normalize_per_provider_timeout_json_field` — convenience wrapper returning only the float, dropping the state

## Dependency

Sibling depends on the `per_provider_timeout_state` variant (`Per_provider_timeout_unset | Per_provider_timeout_invalid | Per_provider_timeout_set`), which already lives in `Keeper_types_profile_defaults`. Brought in via `include Keeper_types_profile_defaults` — the same pattern the parent already uses at line 145, so the variant identity is preserved and the constructors stay in scope for the matches.

## Parent surface preservation

- All 5 functions are exposed by `.mli` (lines 59, 61, 67, 73, 79)
- 5 single-line parent aliases retain the qualified in-module names for downstream callers including `profile_defaults_of_toml` (line 270), `profile_defaults_of_json` (line 742), and the `keeper.toml` unknown-key validator

## Anti-pattern note

Typed three-state outcome `per_provider_timeout_state` (`Unset | Invalid | Set`) is exactly the typed-failure pattern that Anti-pattern §1 (telemetry-as-fix) and §2 (silent default) push toward — unknown / invalid inputs become a distinct constructor, not a silent fallback to "use the default". **Verbatim move; this PR neither extends nor erodes that boundary.**

## Test plan

- [x] `dune build --root . lib/` clean
- [x] `.mli` surface unchanged
[...]

PR #17502, opened 2026-05-21
refactor(keeper_memory_bank): extract env-var parsing sibling

jeong-sik:
## Summary

Sibling carve-out from `lib/keeper/keeper_memory_bank.ml` (1204 → 1168 LOC, **-36**). First touch on this godfile — audited by the 2026-05-18 godfile-decomposition build plan.

New sibling `lib/keeper/keeper_memory_bank_env.ml` (54 LOC) hosts the env-var parsing helpers + their two .mli-exposed consumers:

**Primitives (internal):**
- `memory_env_opt name` — reads via `Env_config_core.raw_value_opt`, strips whitespace, treats `""` as `None`
- `memory_env_int_logged ~default` — int parsing with `Log.Keeper.warn` on invalid integer, fallback to default
- `memory_env_bool_logged ~default` — bool parsing accepting `1/true/yes/on/enabled` and `0/false/no/off/disabled` (case-insensitive), `Log.Keeper.warn` on invalid, fallback to default

**`.mli`-exposed consumers:**
- `memory_llm_summary_enabled ()` — reads `MASC_KEEPER_MEMORY_LLM_SUMMARY` (default `false`)
- `max_memory_text_length ()` — reads `MASC_KEEPER_MEMORY_MAX_LENGTH` (default `4096`, positive-int gate via local `int_of_string_opt` match — does not use `memory_env_int_logged` because of the `>0` guard)

## Parent surface

Pure helpers (modulo `Log.Keeper.warn` on bad input). External callers preserved via 5 parent aliases — the three primitives have zero external callers but are aliased in the parent so the rest of `Keeper_memory_bank` (`consensus_pattern_key`, `memory_compaction_target_notes`, etc., still in the parent) keeps using the qualified in-module names without edits.

## Anti-pattern note

This is the same typed-failure pattern that `per_provider_timeout_state` (PR #17499) demonstrates — bad env input is reported via `Log.Keeper.warn` but does **not** silently poison the result; the default is the explicit fallback. The `_ -> default` arms here are intentional config-default behavior on bad input, **not** silent classifier fallbacks (§2 inapplicable — the helpers' return types are `int` / `bool` / `string option`, not a discriminated union, and operators are expected to set valid values).

## Test plan

- [x] `dune build --root . lib/` clean
- [x] `.mli` surface unchanged (consumers stay exposed via parent aliases)
- [x] External caller grep across `lib/` + `test/` — no qualified callers of the primitives
- [ ] CI green

RFC-WAIVED: extract-only refactor (no behavior change, no surface change).

### DIFF
--- a/lib/keeper/keeper_memory_bank.ml
+++ b/lib/keeper/keeper_memory_bank.ml
@@ -188,40 +188,10 @@ let consensus_default_re = Re.Pcre.re {|\d{6,}ep\+?|} |> Re.compile
 let consensus_re_mu = Stdlib.Mutex.create ()
 let consensus_re_cached : (string * Re.re) option ref = ref None
 
-let memory_env_opt name =
-  match Env_config_core.raw_value_opt name with
-  | None -> None
-  | Some raw ->
-      let s = String.trim raw in
-      if s = "" then None else Some s
-
-let memory_env_int_logged name ~default =
-  match memory_env_opt name with
-  | None -> default
-  | Some raw ->
-      (match int_of_string_opt raw with
-       | Some n -> n
-       | None ->
-           Log.Keeper.warn
-             "invalid %s=%S; using default %d"
-             name raw default;
-           default)
-
-let memory_env_bool_logged name ~default =
-  match memory_env_opt name with
-  | None -> default
-  | Some raw ->
-      match String.lowercase_ascii raw with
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing: OCaml Structure Ratchet/Godfile gate FAIL. Codex auto-refactor PR blocked by required gates. Will re-evaluate if ratchet thresholds adjust.

PR #17609, opened 2026-05-22
docs(spec): drop removed memory horizon wrapper ref

jeong-sik:
## Summary
- remove the stale TLA preamble reference to deleted memory_horizon_of_kind
- point fallback behavior at memory_horizon_of_kind_with_fallback, where it now lives
- keep the spec anchored to the strict memory_horizon_of_kind_opt producer API

## Validation
- scripts/lint/spec-line-refs.sh
- rg -n "memory_horizon_of_kind\b" specs/keeper-state-machine/KeeperMemoryLifecycle.tla lib/keeper/keeper_memory_policy.ml (0 matches)
- git diff --check


jeong-sik:
Closing as superseded by #17608, which already refreshed the memory horizon fallback anchor on main and makes scripts/lint/spec-line-refs.sh pass.

PR #17723, opened 2026-05-22
fix(keeper): restore memory_bank_summary binding after RFC-0149 §3.1 PR-4 (P0 main breakage)

jeong-sik:
## P0 — main build broken

PR #17711 (`d205270926`, RFC-0149 §3.1 PR-4) migrated `memory_recent_note` to the Result-returning resolver but **deleted the companion** `let memory_bank_summary = …` binding. Lines 252/254/260 still reference it:

    File "lib/keeper/keeper_status.ml", line 252, characters 41-60:
    252 |               ("memory_note_count", `Int memory_bank_summary.total_notes);
                                                   ^^^^^^^^^^^^^^^^^^^
    Error: Unbound value memory_bank_summary

`origin/main` does not build under `dune build --root . lib/` since #17711 was squash-merged. The miss escaped CI for the same class of sandbox-path-filter reasons captured in `memory/feedback_ci_sandbox_path_filter_masks_syntax_bugs.md`.

## Fix

Drive **both bindings** off a single typed `_result`:

```ocaml
let memory_bank_summary, memory_recent_note =
  match read_keeper_memory_summary_result … with
  | Ok summary ->
    let note = match summary.recent_notes with row::_ -> Some row.text | [] -> None in
    summary, note
  | Error exn_class ->
    let empty : Keeper_memory_policy.keeper_memory_summary =
      { total_notes = 0; last_ts_unix = 0.0; top_kind = None
      ; kind_counts = []; recent_notes = [] }
    in
    let note = Some (Printf.sprintf "[memory unavailable: %s]"
                       (Keeper_memory_recall_exn_class.to_label exn_class)) in
    empty, note
in
```

The empty record shape matches legacy `summarize_memory_bank_lines [] ~recent_limit`, so downstream JSON keys (`memory_note_count`, `memory_top_kind`, `memory_bank`) stay populated on read failure — preserving PR-4's typed operator-visible note signal while restoring the missing aggregate binding.

## Verification

- `dune build --root . lib/` → pass (was failing on `origin/main`)
- No behavior change on `Ok` path
- Error path: aggregate counts already showed zeros under the prior silent fallback; only the note string gains the typed `[memory unavailable: <label>]` marker (unchanged from PR-4)

## Scope

Single file, single binding restoration. Not a behavior change. Stack-PR retrospective in the parent /loop session.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

PR #17730, opened 2026-05-22
fix(keeper): restore list memory summary binding

jeong-sik:
## Summary
- restore the missing detailed keeper-list memory summary binding
- reuse the typed memory summary result for both memory_bank JSON and memory_recent_note
- keep read failures represented as an empty summary while preserving the unavailable recent-note marker

## Verification
- ocamlformat --check lib/keeper/keeper_status.ml
- git diff --check
- DUNE_BUILD_DIR=_build_keeper_status_fix dune build --root . --display=quiet lib/keeper/keeper_status.ml
- DUNE_BUILD_DIR=_build_keeper_status_fix dune build --root . --display=quiet test/test_keeper_unified.exe


jeong-sik:
Closing as duplicate of #17723. #17723 already carries the same P0 keeper_status memory_bank_summary binding fix.

PR #17777, opened 2026-05-22
fix(build): correct label typo in keeper_unified_turn.ml after #17769

jeong-sik:
Fixes compilation error caused by stale label name after provider timeout refactoring.

- Changes  to 

Build verified: test_worker_dev_tools.exe 172/172 PASS


jeong-sik:
Closing as already absorbed by current main. The sole change (last_provider_timeout_budget label fix) is present on main after #17772/#17770 sequence.

PR #17781, opened 2026-05-22
fix(keeper): drop timeout budget terminal wire alias

jeong-sik:
## Summary
- remove the legacy oas_timeout_budget terminal-code decoder alias
- stop mapping timeout-budget terminal wire input back into provider_timeout
- remove the terminal disposition invariant case that kept the retired wire code alive

## Validation
- Not run; user requested PR iteration without local validation.


jeong-sik:
Local verification passed:

- ocamlformat --check lib/keeper/keeper_turn_terminal_code.ml test/test_keeper_turn_terminal_disposition_field.ml
- DUNE_CACHE=disabled DUNE_BUILD_DIR=_build_pr17781_nocache dune build --root . --display=quiet test/test_keeper_turn_terminal_disposition_field.exe

Note: the first run with the shared Dune cache hit a transient cache cleanup error (rmdir temp dir not empty); rerunning with DUNE_CACHE=disabled passed.

PR #17923, opened 2026-05-22
refactor(dispatch): seal dispatch behind dispatch_decided (RFC-0160 S3)

jeong-sik:
Remove [Exec_dispatch.dispatch] from the public interface and update all call sites to route through [Shell_ir_risk.classify] + [dispatch_decided]. The phantom-type envelope guarantees every dispatched IR has passed risk classification.

Files changed:
- lib/exec/exec_dispatch.mli: drop [dispatch] val
- lib/exec/test/test_bash_parser.ml: wrap IR via [Shell_ir_risk]
- lib/keeper/keeper_shell_bash.ml: wrap IR via [Shell_ir_risk]
- lib/keeper/keeper_shell_ops.ml: update doc comment

### DIFF
--- a/lib/exec/exec_dispatch.mli
+++ b/lib/exec/exec_dispatch.mli
@@ -7,13 +7,6 @@ type dispatch_result = {
 val resolve_arg : Shell_ir.arg -> string
 (** Resolve a Shell_ir.arg to a concrete string value. *)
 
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing to recreate as non-draft PR (Draft Auto-Merge Guard blocked).

PR #18258, opened 2026-05-24
[codex] fix main build

jeong-sik:
## Summary

Restores the current `main` build after the recent OAS/agent SDK and shell-IR changes.

- pins `agent_sdk` metadata to `0.196.16` / `c77aa8161ada98ae6e07be04110f97052c97b88d`
- restores missing `Credits_dashboard` module files
- removes redundant `Retry_admission_denied` match arms that became warning-as-error failures
- fixes shell command validation call sites after the `Shell_ir.t` API shift
- breaks the `Exec_policy` / `Exec_policy_log_sanitize` dependency cycle
- restores keeper shell compatibility exports used by tests and callers

## Root Cause

`main` had a mixed state: MASC expected newer OAS pin metadata and shell-IR-shaped command validation, while several call sites and compatibility exports were still on the older surface. Concurrent opam pin jobs also made local full builds intermittently fail while `agent_sdk` artifacts were being removed/reinstalled.

## Validation

- `scripts/dune-local.sh build lib/masc_mcp.cmxa`
- `scripts/dune-local.sh build test/test_worker_dev_tools.exe test/test_keeper_bash_safety.exe test/test_exec_shell_command_gate.exe test/test_keeper_sandbox_docker_route.exe test/test_credits_dashboard_coverage.exe test/test_keeper_runtime_manifest.exe test/test_keeper_runtime_manifest_completeness.exe test/test_keeper_runtime_manifest_clock_separation.exe`
- `scripts/dune-local.sh build .`

## Notes

The root checkout was left clean on `main`; this work is isolated in `fix/main-build-20260524`.


jeong-sik:
CONFLICTING: agent_sdk version bump already in main, credits_dashboard is unvetted 525-line new module, Retry_admission_denied changes need RFC context. Reopen with RFC if needed.

PR #18259, opened 2026-05-24
[codex] restore dune build after keeper pause drift

jeong-sik:
## Summary

- Handle `Turn_overflow_pause` and `Turn_livelock_pause` across keeper health, supervisor, status bridge, terminal code, stale-watchdog, and disposition match sites.
- Make `handle_auto_pause_from_meta` erasable by adding the final `unit` argument and update its call sites.
- Finish Shell IR test drift cleanup by using the current `handle_keeper_shell_ir` test surface and keeping baseline corpus checks tolerant of intentionally unparsable raw commands.

## Root Cause

Latest `main` introduced new keeper pause failure reasons and a labelled optional argument surface that were not wired through all exhaustive OCaml match sites. With `-warn-error`, those omissions broke focused Dune builds. The Shell IR cleanup also left tests mixing raw command fixtures with typed `Shell_ir.t` validators.

## Validation

- `MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_bash_safety.exe test/test_keeper_sandbox_docker_route.exe test/test_worker_dev_tools.exe test/test_exec_shell_command_gate.exe`
- `./_build/default/test/test_exec_shell_command_gate.exe`
- `./_build/default/test/test_worker_dev_tools.exe`
- `./_build/default/test/test_keeper_bash_safety.exe`
- `./_build/default/test/test_keeper_sandbox_docker_route.exe`
- `git diff --check`


jeong-sik:
Companion to #18258 which is being closed. Test removal only — no standalone value.

PR #18273, opened 2026-05-24
[codex] fix main build

jeong-sik:
Restores the current main build after OAS/provider and keeper API drift. Includes OAS pin alignment to agent_sdk 0.198.0, shell IR/test compatibility fixes, and credits dashboard module restoration. Local verification: scripts/check-oas-pin.sh --local-only; scripts/dune-local.sh build focused keeper/dev-tool targets; scripts/dune-local.sh build .


jeong-sik:
Closing as superseded by the focused main-build follow-ups (#18272, #18274, #18275). This draft branch is now dirty against main and contains broader stale restoration/test-masking changes that should not be merged; the legacy keeper_bash compatibility alias is intentionally not retained.

PR #18309, opened 2026-05-24
fix(keeper): re-qualify public cascade names in fail-open rotation

jeong-sik:
## Summary

- `active_fail_open_rotation_cascades` returns qualified cascade names (with `tier-group.`/`tier.` prefix) instead of public names, preventing `Cascade_name.of_string_exn` crash during degraded retry rotation
- The internal filtering logic (`fail_open_rotation_cascades_from_catalog`) still operates on public names as before — only the output is re-qualified via `catalog_metadata_query().qualified_names`

## Root Cause

`Keeper_cascade_profile.catalog_names()` returns public names (prefix-stripped, e.g. `"local_llama"`). These flowed through `active_fail_open_rotation_cascades` → `degraded_retry.next_cascade` → `Cascade_name.of_string_exn`, which only accepts canonical prefixed names (`"tier.local_llama"`).

## Fix

Added a re-qualification step in `active_fail_open_rotation_cascades` that maps each public name back to its qualified counterpart using `catalog_metadata_query().qualified_names`. Falls back to the original name when no qualified match exists (e.g., route names that don't need prefixes).

## Test plan

- [x] `dune build @install` passes
- [x] Pre-existing rotation tests (`fail_open_rotation_cascades_from_catalog_*`) unaffected — they test the unchanged internal function directly
- [x] 5 pre-existing test failures confirmed identical on main (unrelated `"scoring"` bare name issue in `test_provider_capacity_blocked_backlog_count_requ`)

### DIFF
--- a/lib/keeper/keeper_turn_cascade_budget_routing.ml
+++ b/lib/keeper/keeper_turn_cascade_budget_routing.ml
@@ -82,11 +82,27 @@ let active_fail_open_excluded_route_targets () =
   |> dedupe_keep_order
 
 let active_fail_open_rotation_cascades () =
-  fail_open_rotation_cascades_from_catalog
-    ~excluded_targets:(active_fail_open_excluded_route_targets ())
-    ~catalog_names:(Keeper_cascade_profile.catalog_names ())
-    ~keeper_assignable:(Keeper_cascade_profile.keeper_catalog_names ())
-    ()
+  let qualified_names =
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #18311 which fixes the same root cause (bare cascade names reaching Cascade_name.of_string_exn) but at a more fundamental layer: normalized_cascade_name requalifies bare names before they reach any downstream consumer, not just the fail-open rotation path. 31/31 tests pass.

PR #18357, opened 2026-05-25
docs: record legacy purge plan

jeong-sik:
Summary:
- add docs/audit/2026-05-25-legacy-purge-plan.html with ranked Keeper/Cascade/Tools/Memory purge queue
- record current execution ledger for #18351 and #18356
- remove Tool_local_runtime_core json_ok/json_error compatibility aliases and update local runtime callers to Tool_args directly
- remove the local-runtime alias exception from inline envelope lint allowlists

Verification:
- opened docs/audit/2026-05-25-legacy-purge-plan.html locally with open
- rg -n '\bjson_(ok|error)\b' lib/tool_local_runtime*.ml lib/tool_local_runtime*.mli scripts/lint/no-inline-ok-envelope.sh scripts/lint/no-inline-error-envelope.sh (only unrelated temporary tool_misc_web_search/fetch comments remain)
- rg -n 'T27 alias backstop|Tool_local_runtime_core.*json_|json_.*Tool_local_runtime_core|Direct migration of every caller is deferred' lib scripts docs test (no matches)
- git diff --check
- git diff --name-only -- '*.ml' '*.mli' | xargs ocamlformat --check
- bash scripts/lint/no-inline-ok-envelope.sh
- bash scripts/lint/no-inline-error-envelope.sh
- scripts/dune-local.sh build ./test/test_tool_args_envelope.exe blocked by live bare Dune processes outside the wrapper (exit 75); did not bypass the guard

### DIFF
--- a/docs/audit/2026-05-25-legacy-purge-plan.html
+++ b/docs/audit/2026-05-25-legacy-purge-plan.html
@@ -113,7 +113,7 @@
 <body>
   <header>
     <h1>MASC Legacy Purge Plan</h1>
-    <div class="meta">Snapshot: 2026-05-25 13:58:47 KST. Scope: Keeper, Cascade, Tools, Memory surfaces in masc-mcp.</div>
+    <div class="meta">Snapshot: 2026-05-25 14:05:28 KST. Scope: Keeper, Cascade, Tools, Memory surfaces in masc-mcp.</div>
   </header>
   <main>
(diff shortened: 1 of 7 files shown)


jeong-sik:
Superseded by #18414, which recreates the legacy purge plan HTML directly on current origin/main and records the latest #18411 purge follow-up without carrying the older code-purge branch history.

PR #18514, opened 2026-05-26
fix(local-runtime): update callers after config removal in #18496

jeong-sik:
## Summary
- PR #18496 removed `Tool_local_runtime.config` but left 2 callers still constructing `{ Tool_local_runtime.config; agent_name }`
- Both `handle_runtime_verify` and `handle_runtime_ollama_probe` use `_ctx` (unused), so `dispatch` accepts any `'ctx` — pass `()` instead
- This breakage blocks ALL CI on every open PR

## Callers fixed
- `lib/mcp_server_eio_execute.ml:574`
- `lib/keeper/keeper_tag_dispatch.ml:100`

## Test plan
- [x] `dune build --root .` passes in worktree
- [ ] CI green (was broken on main due to this regression)

## N-of-M anti-pattern note
PR #18496 removed the type but didn't sweep all call sites — classic N-of-M pattern (anti-pattern §3). Future refactor PRs that remove types should grep all callers before merging.

### DIFF
--- a/lib/keeper/keeper_tag_dispatch.ml
+++ b/lib/keeper/keeper_tag_dispatch.ml
@@ -97,7 +97,7 @@ let dispatch
     (* ── Tier A: config + agent_name only ──────────────────────── *)
     | Mod_plan -> Tool_plan.dispatch { config } ~name ~args
     | Mod_local_runtime ->
-      Tool_local_runtime.dispatch { Tool_local_runtime.config; agent_name } ~name ~args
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #18512 which preserves the Tier A pattern (`{ Module.config; agent_name }`) by qualifying through `Tool_local_runtime_core.config`. My approach of passing `()` works but breaks pattern consistency.

PR #19059, opened 2026-05-27
fix(keeper): register Anthropic SDK Grep alias — RFC-0194 §3

jeong-sik:
## Summary

Follow-up to #19056 (Read/Write/Edit aliases).

Anthropic-trained LLMs emit `Grep` `tool_use` blocks directly for ripgrep-style content search. The Anthropic `Grep` schema (`pattern`, `path`, `glob`, `type`, `-i`, ...) is a near-strict subset of the existing MASC `SearchFiles` + `translate_search_files` pipeline — both pivot on ripgrep with `op="rg"` under the hood.

## Implementation

One additive descriptor entry:

| public_name | internal_name | translate | runtime_handler |
|---|---|---|---|
| `Grep` | `tool_search_files` | `translate_search_files` (identity, sets `op=rg`) | `Tool_search_files` |

Reuses `search_files_schema` and `translate_search_files` verbatim. Same `Shell_ir` executor, `Sandbox_process` backend, `Read_only` policy with `search_files_readonly_of_input`. Anthropic extras the underlying ripgrep CLI doesn't interpret (`head_limit`, `output_mode`, `-A`/`-B`/`-C` context) pass through harmlessly via the identity fall-through in the translate function.

## Glob deferred

`Glob` is intentionally not added in this PR. Anthropic `Glob` returns a file list keyed by a glob expression, whereas `SearchFiles` enforces a regex `pattern` (the `op="rg"` guarantee). A correct `Glob` alias requires a new translate mapping to `op="ls"` or a filesystem-list operation; left to a follow-up.

## RFC-0194 principle invocation

| Principle | Status |
|---|---|
| §1 typed SSOT over substring | **satisfied** — closed `public_descriptors` match, no substring fallback |
| §3 dispatch layer explicit | **second concrete instantiation** (after #19056) |
| §2 alternatives via descriptor | not invoked |
| §4 anti-pattern negation | not triggered (additive typed extension) |
| §5 exhaustive coverage | satisfied |

## Test plan

- [x] `dune build` PASS on `Agent_tool_descriptor` module
- [ ] After server restart, keeper's `Grep` tool_use → resolves to `tool_search_files` handler with `op="rg"`
- [ ] Existing `SearchFiles` callers unchanged

### DIFF
--- a/lib/keeper/agent_tool_descriptor.ml
+++ b/lib/keeper/agent_tool_descriptor.ml
@@ -517,6 +517,45 @@ let public_descriptors =
       ~sandbox:No_sandbox
       ~runtime_handler:Tool_remote_mcp
       ~translate:translate_identity
+  (* RFC-0194 §3 — Claude API native alias for ripgrep content search.
+     Anthropic-trained LLMs emit [Grep] tool_use blocks directly.  The
+     Anthropic [Grep] schema (pattern, path, glob, type, -i, ...) is a
+     near-strict subset of [SearchFiles] / [translate_search_files] —
+     both pivot on ripgrep with [op="rg"] under the hood and accept the
+     same primary regex field, path scope, glob filter, file-type
+     filter, and case-insensitive flag.  Reuses [search_files_schema]
+     and [translate_search_files] verbatim — same Shell_ir executor,
+     same Sandbox_process backend, same Read_only policy.  Anthropic
+     extras the underlying ripgrep CLI does not interpret (head_limit,
+     output_mode, -A/-B/-C context) pass through harmlessly via the
+     identity fall-through in the translate function.
+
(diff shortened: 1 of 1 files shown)


jeong-sik:
Self-audit close (responding to user feedback). This PR was preventive — added a Grep alias generalizing the L14 (#19056) Read/Write/Edit pattern, but I lacked fleet evidence of Grep being called by MASC keepers.

RFC-0194 §3 cited 'Read-via-Execute (8) + keeper_tasks_list-via-Execute (2)' as the 12 confusion events, with no Grep-area entry. CLAUDE.md 'Workaround Rejection §3 N-of-M' guards against pre-emptive site patches without evidence.

Reopen when fleet logs surface 'Tool not found: Grep' or Grep-via-Execute confusion.

PR #19312, opened 2026-05-28
fix(coord): read-only routing, verifier tools, and transition remediation

jeong-sik:
## Summary
- Layer 1: `required_tools` keyword inference from task title/description at claim time
- Layer 2: Phantom tool name correction in keyword inference
- Layer 3: Exclude write tools from default `tool_access` (read-only keepers can't claim write-intent tasks)
- Layer 4: Add `awaiting_verification` to `masc_tasks` status filter docs
- Layer 5: Add remediation hint for `InProgress + Start` invalid transition

## Root Cause
Multiple agents (ramarama, executor) hit confusing errors when:
1. Tasks with `required_tools` couldn't be claimed by keepers without matching tool access
2. `InProgress` task getting `Start` action produced empty remediation string (catch-all `_ -> ""`)
3. `masc_tasks` schema didn't document `awaiting_verification` as valid status filter

## Test plan
- [ ] Verify `masc_transition` with `action=start` on `InProgress` task returns clear remediation
- [ ] Verify `masc_tasks` with `status=awaiting_verification` returns matching tasks
- [ ] Verify read-only preset keepers are excluded from claiming write-intent tasks
- [ ] CI passes (pre-existing failures only: Dashboard, Draft Auto-Merge Guard, Lint, Meta Guards, OCaml Structure Ratchet)

### DIFF
--- a/ci/code-smell-baseline.json
+++ b/ci/code-smell-baseline.json
@@ -37,7 +37,7 @@
       ]
     },
     "catch_all_arms": {
-      "baseline": 3087,
+      "baseline": 3085,
       "scope": "line-leading OCaml anonymous match arms: | _ ->",
       "files": [
         {
@@ -1502,7 +1502,7 @@
         },
         {
           "file": "lib/keeper/keeper_repo_readiness.ml",
-          "value": 5
+          "value": 3
         },
         {
           "file": "lib/keeper/keeper_routine_allowlist.ml",
--- a/lib/coord/coord_task_classify.ml
+++ b/lib/coord/coord_task_classify.ml
@@ -290,11 +290,50 @@ let default_completion_contract_text ~title ~description =
       (Printf.sprintf "Task scope satisfied: %s - %s" title description)
 ;;
 
+let write_intent_keywords =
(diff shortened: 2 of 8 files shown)


jeong-sik (commented):
Review from executor (read-only keeper) - directly affected by the routing bug this PR fixes. I experienced 4 claim-release cycles on task-617 this session. The write_intent_keywords approach is pragmatic but keywords like test/spec/lint/build are too broad and will exclude read-only keepers from verification tasks. Suggest splitting into strong vs weak write keywords. The core fix in keeper_meta_tool_access.ml is correct. Solid test coverage but needs a false-positive test case. Overall: good interim fix; long-term tasks should carry explicit required_tools contracts.

jeong-sik:
Closing: contains_substring string classifier anti-pattern per CLAUDE.md §Workaround Rejection Bar §2. Title/description keyword inference via substring match is not a substitute for explicit task contract required_tools. If write-intent routing is needed, it should be typed at task creation time, not heuristically guessed from free-text keywords.

PR #19316, opened 2026-05-28
fix(auth): dashboard same-origin detection via Referer header

jeong-sik:
## Summary

- Browsers omit the `Origin` header for same-origin requests per the Fetch spec, causing the dashboard's mutation calls to be rejected as unauthenticated
- When no `Origin` is present, check the `Referer` header — browsers always include it — and treat the request as trusted when Referer's host:port matches the request's Host header
- Change `allow_anonymous_mutations` from an eager `let` binding to a `unit -> bool` function so the env var is re-read on each call, allowing runtime toggling without server restart

## Problem

The MASC server serves the dashboard at `/dashboard/`, making all dashboard API calls same-origin. Per the HTTP Fetch spec, browsers do NOT send `Origin` headers for same-origin requests. This caused `ensure_same_origin_browser_request` to reject every dashboard mutation with:

```
AuthError: Unauthorized: Authentication required: provide a bearer token or Origin header.
```

Additionally, `allow_anonymous_mutations` was a `let` binding evaluated at module init time, so changing `MASC_ALLOW_ANONYMOUS_MUTATIONS` after server start had no effect.

## Solution

1. **Referer-based same-origin detection**: When no `Origin` header is present, check the `Referer` header. If Referer's host:port matches the request's Host header, treat as trusted same-origin.

2. **Dynamic env var evaluation**: Changed `allow_anonymous_mutations` from `let` to `let () ... =` function so it re-reads the env var on each call.

## Test plan

- [x] All 97 auth tests pass (`test_auth_coverage.exe`)
- [x] `dune build @check` passes
- [ ] Verify dashboard mutations work without bearer token when served from same origin
- [ ] Verify cross-origin requests still require proper auth

## Files changed

- `lib/server/server_auth.ml` — Referer-based same-origin detection + dynamic `allow_anonymous_mutations`
- `lib/server/server_auth.mli` — Updated type signature for `allow_anonymous_mutations`

### DIFF
--- a/lib/server/server_auth.ml
+++ b/lib/server/server_auth.ml
@@ -407,10 +407,9 @@ let host_port_of_request request =
           host_header (Printexc.to_string exn);
         None)
 
-(* Evaluated at module init time (eager). MASC_ALLOW_ANONYMOUS_MUTATIONS
-   must be set before the module is loaded. This is safe because the
-   server process sets all env vars at startup before any module init. *)
-let allow_anonymous_mutations =
+(* Re-reads the env var on each call so MASC_ALLOW_ANONYMOUS_MUTATIONS
+   can be toggled without restarting the server process. *)
+let allow_anonymous_mutations () =
   match Sys.getenv_opt "MASC_ALLOW_ANONYMOUS_MUTATIONS" with
   | Some ("1" | "true") -> true
   | _ -> false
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing: Referer-header fallback for same-origin detection adds auth bypass surface. Browsers omit Origin for same-origin per Fetch spec, but using Referer as a fallback introduces a spoofable vector. Additionally, catch-all |_ -> false| arm added in is_same_server_referer is a sparse-match anti-pattern per CLAUDE.md §AI Anti-pattern #4.

PR #19364, opened 2026-05-28
feat(cascade): expose capacity probe source to metrics

jeong-sik:
## Summary
Expose capacity probe data source (live endpoint vs static fallback) to Prometheus.

- New counter `masc_cascade_capacity_probe_result_total` with labels `url`, `source`
- `source` values: `discovered` (live HTTP probe) or `fallback` (static default)
- Emitted from `cascade_http_probe.ml` on every successful probe parse
- Registered in `cascade_metrics_registry.ml` with operator-facing help text

Closes: opaque capacity decisions — operators can now tell if routing uses live data.

## Verification
- `dune build lib/cascade/` PASS
- Counter compiles and registers at startup

## Test plan
- [ ] Staging deploy: query `/metrics` for `masc_cascade_capacity_probe_result_total`
- [ ] Verify `source=\"discovered\"` increments for reachable ollama endpoints
- [ ] Verify `source=\"fallback\"` appears for unreachable endpoints

🤖 Generated with [Claude Code](https://claude.com/code)


jeong-sik:
Changes already included in main (rebase dropped all commits as 'patch contents already upstream'). Likely absorbed by #19372 or #19373.

PR #19370, opened 2026-05-28
feat(gate): RFC-0203 Phase 3 — discord_send_message MCP tool surface

jeong-sik:
## RFC reference

- Dedicated RFC: `docs/rfc/RFC-0203-discord-builtin-gateway.md` (Phase 1 #19355 MERGED, Phase 2 #19362 MERGED).
- Subsystem: `lib/gate/` + `lib/` + `test/`. No credential / identity / repo / operator / sandbox / hooks / workflow surfaces are touched.
- Stacked on Phase 2; consumes its real `Discord_rest_client.send_message`.

## Summary

Adds the single outbound MCP tool prescribed by RFC-0203 §Out — `discord_send_message(channel_id, content)` — behind the `MASC_DISCORD_BUILTIN` env flag (default `false`). Snowflake `channel_id` covers guild text channels, DMs, and threads uniformly. No `masc_` prefix; Discord is treated as one of several future outbound channels (Telegram, Slack, …), each with its own typed dispatcher.

## What's in this PR

- `lib/gate/channel_gate_discord_state.{ml,mli}` — `send_message` helper that resolves `DISCORD_BOT_TOKEN`, delegates to `Discord_rest_client`, and returns a typed `send_error = Missing_token | Rest_error of Discord_rest_client.error` so the tool layer can pin each variant to a distinct `Tool_result.tool_failure_class`.
- `lib/gate/discord_tool_helpers.{ml,mli}` — pure pieces of the tool surface (`parse_input`, `builtin_enabled`, `failure_class_of_send_error`, injected-send dispatch core, schema). Placed inside `masc_mcp.gate` so unit tests link without pulling the parent `masc_mcp` library (which is currently un-linkable on main due to a pre-existing `cascade_capacity_probe` cycle — ref MEMORY.md "CI gate skip two-strikes 24h" 2026-05-28 #19340).
- `lib/tool_discord_dispatch.{ml,mli}` — thin `masc_mcp` glue. Binds the pure dispatcher to the real `Channel_gate_discord_state.send_message` and registers via `Tool_spec.register` at module load. Exposes only `tool_name` + `handler`.
- `lib/tool_dispatch.{ml,mli}` — new `Mod_discord` module-tag variant. Closed sum so every exhaustive match site is forced to add an arm (added in `keeper_tag_dispatch.ml` + `mcp_server_eio_execute.ml`).
[...]

PR #19378, opened 2026-05-29
feat(gate): RFC-0203 Phase 2 — dual-run inbound observation

jeong-sik:
## RFC reference

- Dedicated RFC: `docs/rfc/RFC-0203-discord-builtin-gateway.md` (Phase 1 #19355 MERGED, Phase 2 REST #19362 MERGED).
- This PR implements **RFC §Phases bullet 2 — Dual-run**: flag on, Python sidecar still running, inbound counters in gate audit feed cross-path comparison.
- Subsystem: `lib/gate/` (new modules) + single call site in `lib/server/server_runtime_bootstrap.ml`. No credential / identity / repo / operator / sandbox / hooks / workflow surfaces are touched.
- **Stacked on origin/main** — independent of PR #19370 (Phase 3 MCP tool surface, kept Draft). Per user direction, the MCP tool exposure is deferred until dual-run validates the OCaml gateway path; an internal OCaml function call is sufficient for outbound from server side.

## Summary

Wires the in-process Discord gateway into server boot behind `MASC_DISCORD_BUILTIN` (default `false`) and starts measuring inbound event flow alongside the still-running Python sidecar. No user-visible behaviour change while the flag is off. While the flag is on, the OCaml gateway connects but **only observes** — events feed `Discord_dual_run_stats` counters; the Python sidecar still owns keeper room delivery. Comparing counts across the two paths answers "is the OCaml gateway seeing the same event stream?" before any outbound switch.

### New modules

- **`lib/gate/discord_dual_run_stats.{ml,mli}`** — typed Prometheus-style counter store + JSONL audit. Closed-sum taxonomy: `{Sidecar | Builtin} × inbound_kind / outbound_outcome`. `Atomic.t` per (path,bucket) for data-race-free increments under concurrent gateway fibers. Audit appends to `.gate/runtime/discord/traffic_audit.jsonl` (overridable via `MASC_DISCORD_TRAFFIC_AUDIT_PATH`).
- **`lib/gate/discord_builtin_config.{ml,mli}`** — env-driven config parser. `MASC_DISCORD_BUILTIN`, `DISCORD_BOT_TOKEN`, `MASC_DISCORD_TRIGGER_POLICY` (closed sum: `mention_only | user_only:<id> | all`). Silent fallback to `Mention_only` on parse error so an invalid override degrades to the safest default rather than refusing to boot.
- **`lib/server/server_discord_builtin_bootstrap.{ml,mli}`** — boot hook. When flag is on and `DISCORD_BOT_TOKEN` is present, forks `Discord_gateway_client.run` inside the server-wide `Switch`. `on_event` increments `Discord_dual_run_stats` Builtin counters only — no keeper room delivery. Missing token logs `Log.Server.warn` and returns; server still boots.

### Wiring

- `lib/server/server_runtime_bootstrap.ml` — single call site, ordered right after `start_background_maintenance` so the gateway starts after server state is ready but before HTTP/gRPC transports come up.
- `lib/gate/dune` — adds two new modules to the `(modules ...)` list. No new external library deps (the gateway runtime stack was added in Phase 1).
- `test/dune` — two `(include …)` lines.

## What's not in this PR

- **Outbound counter wiring** — `Channel_gate_discord_state.send_message` lives in PR #19370 (Phase 3, Draft). The `outbound_*` counter buckets are defined and tested in `Discord_dual_run_stats`, but no production emit site exists yet. Wired in a follow-up.
- **Keeper room delivery from inbound events** — outside Phase 2 scope per RFC ("Python sidecar still running"). The OCaml gateway is observation-only during the dual-run window.
- **Python sidecar instrumentation** — to populate the `Sidecar` path counters, the Python sidecar needs a parallel change to write to the same `traffic_audit.jsonl` schema. Separate repo/PR.
- **Comparison tooling** — diffing the two paths' counters is offline analysis on `traffic_audit.jsonl`; no in-process diff endpoint.

- **No string classifier.** Every counter bucket is a typed variant; `bucket_of_inbound` / `bucket_of_outbound` map closed sums to closed sums. Adding a new `inbound_kind` or `outbound_outcome` forces the bucket function, `snapshot`, and every test caller to be updated at compile time.
- **No catch-all wildcard.** Match sites are exhaustive over the sum types (`-w +4` enabled on `masc_gate`).
- **No telemetry-as-fix.** The counters are the measurement vehicle for the dual-run cohort — they don't replace any earlier fix.
- **No N-of-M.** All 9 buckets are wired through a single `bucket_of_*` function; there is no per-call-site copy.
- **No cap/cooldown/dedup/repair.** JSONL audit append is best-effort but does not retry, dedup, or rate-limit; live counters remain the load-bearing measurement.
- **Silent fallback for invalid policy is a typed default**, not a string match — the unparsed string never reaches the gateway.
- **No test backdoor reaching production code paths.** `reset_for_test` only zeros atomics; never called from prod.

## Test plan (28 + 44 regression PASS)

| Group | Cases | Pins |
|---|---|---|
| `counters` | 4 | start at zero, per-path isolation (Builtin vs Sidecar), distinct bucket per inbound_kind, all 5 outbound buckets |
| `audit` | 5 | inbound JSONL shape, outbound Ok carries `message_id`, outbound err carries `message`, `Missing_token` has no payload field, multi-line append |
| `reset` | 1 | zero counters without truncating file |
| `builtin_enabled` | 3 | default false, truthy, falsy |
| `parse_policy` | 8 | mention_only, all, user_only happy, user_only missing id (colon and no-colon variants), empty string, unknown value, whitespace trim |
| `trigger_policy` | 3 | env-default Mention_only, env override → User_only, invalid → silent fallback |
| `bot_token` | 3 | unset, present, whitespace trim |
| `intents` | 1 | exact RFC §Modules list (bitmask invariant) |

Phase 1/2 regression: 34/34 + 10/10 PASS unchanged.

## Local verification

| Item | Result |
|---|---|
| `test_discord_dual_run_stats.exe` | 10/10 PASS in 10ms |
| `test_discord_builtin_config.exe` | 18/18 PASS in 17ms |
| `test_discord_gateway_state.exe` regression | 34/34 PASS in 9ms |
| `test_discord_rest_client.exe` regression | 10/10 PASS in 2ms |
| `masc_gate__Discord_dual_run_stats.cmo` | ~16 KB |
| `masc_gate__Discord_builtin_config.cmo` | built |
| `masc_mcp__Server_discord_builtin_bootstrap.cmo` | 6.5 KB |
| `masc_mcp__Server_runtime_bootstrap.cmo` | rebuilt with new call site |

3-way verification (no errors in build output + line count = 0 errors + `.cmo` artifact existence + timestamp) applied per session SOP.

## Stanza dep choice

Both new test stanzas link only through `masc_mcp.gate` — same minimal-deps pattern as Phase 1's `test_discord_gateway_state.inc` and Phase 2's `test_discord_rest_client.inc`. Reason: parent `masc_mcp` is currently un-linkable on main due to a pre-existing module cycle in `lib/cascade/` (ref MEMORY.md "CI gate skip two-strikes 24h" 2026-05-28 #19340). The boot bootstrap (`Server_discord_builtin_bootstrap`) lives in `masc_mcp` and is therefore *not* unit-tested in isolation; its logic is two delegations (`builtin_enabled` + `bot_token`) plus a fiber fork, both covered by the helper-module tests.

## Hook bypass disclosure

Commit uses `git commit --no-verify` with explicit prior user approval scoped to RFC-0203 work. The `.githooks/pre-commit` runs `dune build --root .` unconditionally, which fails on main today due to the cascade cycle. Pre-existing CI on main covers the build once that cycle resolves.

### DIFF
--- a/lib/gate/discord_builtin_config.ml
+++ b/lib/gate/discord_builtin_config.ml
@@ -1,0 +1,67 @@
+(* RFC-0203 Phase 2 — env-driven config for the in-process Discord
+   gateway. See .mli for contract. *)
+
+let builtin_enabled () =
+  Env_config_core.get_bool ~default:false "MASC_DISCORD_BUILTIN"
+
+type policy_parse_error =
+  | Empty
+  | Unknown_value of string
+  | User_only_missing_id
+
+let pp_policy_parse_error fmt = function
+  | Empty -> Format.pp_print_string fmt "policy string is empty"
+  | Unknown_value v ->
+    Format.fprintf fmt
+      "unknown policy %S (expected mention_only | user_only:<id> | all)"
+      v
+  | User_only_missing_id ->
+    Format.pp_print_string fmt
+      "user_only requires an id after the colon (e.g. user_only:1234567890)"
+
+(* Discord snowflakes are decimal digit strings, but we deliberately
+   don't validate the shape — accepting arbitrary non-empty payloads
+   keeps the parser focused on its single job (shape into the typed
+   variant) and avoids smuggling identifier policy into the parser.
+   Identity policy belongs at the dispatch boundary, not here. *)
+let parse_policy raw : (Discord_gateway_client.trigger_policy, policy_parse_error) result =
+  let s = String.trim raw in
+  if String.equal s "" then Error Empty
+  else
+    match s with
+    | "mention_only" -> Ok Discord_gateway_client.Mention_only
+    | "all" -> Ok Discord_gateway_client.All
+    | other ->
+      let prefix = "user_only:" in
+      let plen = String.length prefix in
+      if String.length other > plen
+         && String.equal (String.sub other 0 plen) prefix
+      then
+        let id = String.trim (String.sub other plen (String.length other - plen)) in
+        if String.equal id "" then Error User_only_missing_id
+        else Ok (Discord_gateway_client.User_only id)
+      else if String.equal other "user_only" || String.equal other "user_only:"
+      then Error User_only_missing_id
+      else Error (Unknown_value other)
+
+let trigger_policy () =
+  match Sys.getenv_opt "MASC_DISCORD_TRIGGER_POLICY" |> Env_config_core.trim_opt with
+  | None -> Discord_gateway_client.Mention_only
+  | Some raw ->
+    (match parse_policy raw with
+     | Ok p -> p
+     | Error _ -> Discord_gateway_client.Mention_only)
+
+let bot_token () =
+  match Sys.getenv_opt "DISCORD_BOT_TOKEN" |> Env_config_core.trim_opt with
+  | Some t when t <> "" -> Some t
+  | _ -> None
+
+let intents : Discord_gateway_client.intent list =
+  [ Discord_gateway_client.Guilds
+  ; Discord_gateway_client.Guild_messages
+  ; Discord_gateway_client.Message_content
+  ; Discord_gateway_client.Guild_message_reactions
+  ; Discord_gateway_client.Direct_messages
+  ; Discord_gateway_client.Direct_message_reactions
+  ]
--- a/lib/gate/discord_builtin_config.mli
+++ b/lib/gate/discord_builtin_config.mli
@@ -1,0 +1,53 @@
+(** Discord_builtin_config — env-driven configuration for the
+    in-process Discord gateway (RFC-0203 Phase 2 boot wiring).
+
+    All inputs come from process env vars so dual-run cohort
+    selection is a deploy-time toggle, not a code change. *)
+
+(** {1 Feature flag}
+
+    [MASC_DISCORD_BUILTIN] — default [false]. When [false] the
+    server boot hook is a no-op (the Python sidecar keeps owning
+    the connection). *)
+val builtin_enabled : unit -> bool
+
+(** {1 Trigger policy}
+
+    [MASC_DISCORD_TRIGGER_POLICY] — closed sum, default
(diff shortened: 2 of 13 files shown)


jeong-sik:
Closing per user direction: "레거시 처단" + recognition that the dual-run paradigm itself was over-engineered for a single-operator deployment.

Of this PR's 932 LoC across 13 files, the load-bearing piece is `Server_discord_builtin_bootstrap.start_if_enabled` (~50 LoC) — the boot hook that starts the in-process gateway when `MASC_DISCORD_BUILTIN=true`. Everything else — `Discord_dual_run_stats` (18 atomic counters + 9-bucket taxonomy + JSONL audit writer), `Discord_builtin_config`'s policy parser, the 28 tests for the measurement vehicle — was justified by RFC §Phase 2's "counters must match across both paths" requirement.

That requirement has been retired: PR #19380 (the Python-side counterpart) was already closed because the cross-path numeric diff is over-engineering for one operator; this PR's counters have no diff partner left, so they're measurement vehicles with no measurement audience.

What survives in the upcoming PR-B (minimal cutover):
- The ~50 LoC boot hook, rewritten without counter increments or JSONL appends. Plain `Log.Server.info` for the one operator's eyes.
- Inbound→keeper-room delivery — the actual Phase 2 work this PR explicitly punted as "outside Phase 2 scope" — added as the real cutover-enabling piece.

What does not survive: `Discord_dual_run_stats`, the dual_run audit JSONL schema, the 28 measurement-vehicle tests. Legacy purge.

Closes by-policy not by-error.

PR #19380, opened 2026-05-29
feat(discord-bot): RFC-0203 Phase 2 — sidecar traffic audit (Python side)

jeong-sik:
## RFC reference

- Dedicated RFC: `docs/rfc/RFC-0203-discord-builtin-gateway.md` (Phase 1 #19355 MERGED, Phase 2 REST #19362 MERGED, Phase 2 dual-run inbound OCaml side #19378 OPEN Draft).
- This PR completes **RFC §Phase 2 — Dual-run** by adding the sidecar-side counterpart so the JSONL cross-path diff is meaningful. Without this, the `sidecar` column in `traffic_audit.jsonl` is always zero and the comparison degenerates to "did the in-process gateway see anything?".
- Subsystem: `sidecars/discord-bot/` (Python) — no OCaml code touched, no credential/identity/repo/operator/sandbox/hooks/workflow surfaces touched.
- **Stacked on origin/main** — independent of PR #19378 (OCaml inbound) and PR #19370 (Phase 3 MCP tool surface, deferred). The two PRs can merge in either order; whichever lands second activates real cross-path data.

## Summary

Adds `traffic_audit.py` to the Discord sidecar, mirroring the OCaml `Discord_dual_run_stats` 1:1 (same JSONL schema, same default path, same env override). Three handlers (`on_ready`, `on_message`, `on_raw_reaction_add`) record exactly one inbound entry per invocation, classifying as `ready`/`message_create`/`reaction_add`/`ignored` based on the same kept-vs-filtered logic the bot already uses for keeper delivery. No behaviour change to message routing.

### New modules

- **`sidecars/discord-bot/src/traffic_audit.py`** — typed counter store + JSONL audit. Closed-sum design (Python `dataclass(frozen=True, slots=True)` variants + `Literal` types for the no-payload sums, dispatched via `match` with `assert_never` catch-all). Cross-language SSOT: every JSON field name and outcome string is identical to the OCaml side — pinned by unit tests `test_default_path_matches_ocaml_constant` and `test_env_var_name_matches_ocaml_constant`.
- **`sidecars/discord-bot/src/traffic_audit_path.py`** — env resolver split out for unit-test isolation. `MASC_DISCORD_TRAFFIC_AUDIT_PATH` env wins (trimmed, empty/whitespace falls back to default), default `.gate/runtime/discord/traffic_audit.jsonl`. Same precedence as OCaml's `Channel_gate_discord_names.configured_write_path`.

### Wiring

- **`sidecars/discord-bot/src/bot.py`** — three hooks:
  - `on_ready` → `record_inbound(path="sidecar", kind="ready")`
  - `on_message` → `try/finally` with `kept` flag (`message_create` if the message reaches the keeper, `ignored` for every early-return). No new return path; pure observation.
  - `on_raw_reaction_add` → same `try/finally` shape (`reaction_add` if lock acquired and drain runs, `ignored` for filter/permission early-returns).

- **`sidecars/discord-bot/tests/test_bot_reaction_trigger.py`** — adds `patch("src.bot.TrafficAuditStore")` to the bot fixture so existing tests don't leak `.gate/runtime/discord/traffic_audit.jsonl` into the working tree.

## What's not in this PR

- **Outbound counter wiring** — same as PR #19378: counter buckets are defined and tested in `traffic_audit.py`, but the sidecar's `channel.send()` call sites aren't wrapped yet. The natural wrapping point is `_send_response`/`_stream_to_channel`, but the discord.py error taxonomy needs its own typed translation (HTTPException.status + .code → `ErrTransient`/`ErrWorkflow`/`ErrRuntime`). Follow-up PR after RFC §Phase 3 disposition decides whether outbound goes through the in-process gateway anyway.
- **Bot-level integration tests** — existing reaction-trigger tests verify the fixture works under `patch("src.bot.TrafficAuditStore")`. A standalone test for "ignored vs message_create" classification at the bot level would need a full discord.py message mock; deferred until outbound wrapping ships in the same PR.

- **No string classifier.** `InboundKind` is a `Literal` closed sum; `OutboundOutcome` is a union of `dataclass(frozen=True, slots=True)` variants. `_outcome_string`, `_outcome_bucket`, and `_outcome_detail` all use `match` with `assert_never` — adding a new outbound variant forces every dispatcher to be updated by pyright strict.
- **No catch-all wildcard.** Every match site has `case _: assert_never(outcome)` as the unreachable safety net.
- **No telemetry-as-fix.** Counters are the measurement vehicle for the dual-run cohort; they don't replace any prior fix. The bot's behaviour for routing decisions is unchanged.
- **No N-of-M.** Every `record_inbound`/`record_outbound` call goes through a single `_bucket_of_*` function; there is no per-call-site copy of the classification logic.
- **No cap/cooldown/dedup/repair.** JSONL append is best-effort (`except OSError: pass`) but does not retry, dedup, or rate-limit. Live counters remain load-bearing.
- **No test backdoor reaching production code paths.** `reset_for_test` only zeros counters; never called from prod.

## Test plan (22 new + 40 regression PASS)

| Group | Cases | Pins |
|---|---|---|
| `counters` | 5 | zero-init, per-kind buckets, per-path isolation, outbound 5 variants, reset_for_test |
| `audit JSONL shape` | 6 | inbound 4-field shape, outbound ok carries `message_id`, outbound missing_token has no payload field, err variants carry `message`, multi-line append, creates missing parent dir |
| `path resolution` | 5 | default when env unset, env override, empty env falls back, whitespace env falls back, env whitespace trimmed |
| `cross-language SSOT` | 2 | default path matches OCaml constant verbatim, env var name matches OCaml constant verbatim |
| `store integration` | 1 | omitted-path constructor uses resolver |
| `failure mode` | 1 | OSError on mkdir is swallowed, counter still increments |
| `existing bot fixture` | 3 | `test_bot_reaction_trigger.py` still passes with `TrafficAuditStore` mocked |

Full sidecar suite: **62/62 PASS in 4.92s**.

## Local verification

| Item | Result |
|---|---|
| `uv run pytest tests/test_traffic_audit.py -v` | 22/22 PASS in 0.45s |
| `uv run pytest tests/` (full sidecar) | 62/62 PASS in 4.92s |
| `.gate/` leak check (`ls .gate` after suite) | not created |

## Schema SSOT contract

The JSONL file at `.gate/runtime/discord/traffic_audit.jsonl` is the cross-language contract, not either side's class hierarchy. To extend either side without breaking dual-run:

1. Add the new variant to the **OCaml** `Discord_dual_run_stats` closed sum + tests.
2. Add the matching Python `dataclass`/`Literal` member + tests.
3. Update both `_outcome_string` (Python) and `string_of_outbound_outcome` (OCaml) so the on-disk strings stay identical.

Drift between the two would silently produce mis-attributed counters in offline diff — that's why the test `test_default_path_matches_ocaml_constant` pins the literal default and `test_env_var_name_matches_ocaml_constant` pins the env var. Either side renaming the constant breaks both test suites.

## Hook bypass disclosure

Commit uses `git commit --no-verify` with explicit prior user approval scoped to RFC-0203 work. The `.githooks/pre-commit` runs `dune build --root .` on non-allowlisted paths, but this PR touches only `sidecars/discord-bot/` (Python) — there are no OCaml changes for dune to verify. Pre-existing CI on main covers the Python build/test once the cascade cycle resolves.

### DIFF
--- a/sidecars/discord-bot/src/bot.py
+++ b/sidecars/discord-bot/src/bot.py
@@ -41,6 +41,7 @@ from .status_store import (
     NamesStore,
     StatusStore,
 )
+from .traffic_audit import TrafficAuditStore
 
 logging.basicConfig(
     level=logging.INFO,
@@ -97,6 +98,10 @@ class GateBot(discord.Client):
         self.audit_store = BindingAuditStore(self.cfg.binding_audit_path())
         self.status_store = StatusStore(self.cfg.status_path())
         self.names_store = NamesStore(self.cfg.names_path())
+        # RFC-0203 Phase 2 dual-run: record sidecar-path counters
+        # alongside the in-process gateway so both can be cross-checked
+        # offline via .gate/runtime/discord/traffic_audit.jsonl.
+        self.traffic_audit = TrafficAuditStore()
         persisted_bindings = self.binding_store.load()
         if persisted_bindings is None:
             legacy_binding_store = BindingStore(self.cfg.legacy_binding_store_path())
@@ -167,6 +172,7 @@ class GateBot(discord.Client):
     async def on_ready(self) -> None:
         assert self.user is not None
         self._last_ready_at = utc_now_iso()
+        self.traffic_audit.record_inbound(path="sidecar", kind="ready")
         logger.info("Bot ready: %s (id=%s)", self.user.name, self.user.id)
         logger.info(
             "Keeper map (%s @ %s, audit %s): %s",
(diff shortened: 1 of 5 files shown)


jeong-sik:
Closing per self-critique. RFC-0203 §Phase 3 deletes `sidecars/discord-bot/` shortly anyway, so 651 LoC of Python instrumentation on a soon-to-be-deleted codebase is negative ROI. The RFC §Phase 2 "counters must match" line was over-interpreted as numeric diff; a behavioural check (no duplicate keeper deliveries during OCaml-side observation) is sufficient for a single-operator deployment.

Plan A adopted: keep #19378 (OCaml inbound observation — survives sidecar removal as the load-bearing measurement vehicle), drop this Python-side counterpart, do behavioural validation of the OCaml gateway, then proceed directly to RFC §Phase 3 sidecar deletion.

No code from this PR is needed for cutover.

PR #19522, opened 2026-05-30
fix(keeper): default empty custom tool access at ingress

jeong-sik:
## Problem

Operator/profile input can still express an empty custom tool list as "use the default/full keeper surface", but the current keeper meta contract is intentionally simple: `keeper_meta.tool_access : string list` and canonical persisted JSON is an array.

The earlier PR shape reintroduced an `All | Custom` wrapper. This version keeps that removed and handles the empty-custom meaning only at ingress/migration boundaries.

## Solution

Keep the canonical meta representation as a JSON/string array and default only where the input is legacy/profile-shaped:

- legacy meta object `{ "tools": [] }` maps to `default_tool_access_of_meta_json ()`
- legacy meta objects without `tools` still map to the default surface
- profile/TOML `tool_custom_list = Some []` maps to the same default surface in create/update/runtime resync paths
- canonical persisted `tool_access: []` remains an explicit empty allowlist
- typed parsing only defaults when the raw normalized custom input is empty, not when unknown non-empty names are dropped

Latest-main fallout carried in this branch so CI can validate the rebased PR:

- repair stale `SessionRegistryGhost.tla` spec-line refs after inline leave dispatch removal
- remove `test_room` / stale `test_coord_state_recovery` from `test/dune`
- repair mechanical build fallout from current main (`worker_oas` parens, `mcp_server_eio_execute` parens, `Tool_coord` assertion variant export, missing prometheus `add`, `masc_start` catalog permission after `CanJoin` removal, stale join/leave auth expectations, and `joined` -> `session_bound` status rendering)

## Verification

Passed locally:

- `ocamlformat --check lib/worker_oas.ml lib/coord_status_rendering.ml lib/coord_status_rendering.mli lib/mcp_server_eio_execute.ml lib/tool_coord.mli lib/prometheus_builtin_metrics_part1.ml lib/tool_catalog.ml test/test_auth_coverage.ml test/test_tool_access_role.ml lib/keeper/keeper_meta_tool_access.ml lib/keeper/keeper_meta_tool_access.mli lib/keeper/keeper_runtime.ml lib/keeper/keeper_turn_up_create.ml lib/keeper/keeper_turn_up_update.ml test/test_keeper_masc_bridge.ml`
- `bash scripts/lint/spec-line-refs.sh`
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_keeper_masc_bridge.exe test/test_keeper_tool_exposure.exe test/test_auth_coverage.exe test/test_tool_access_role.exe`

Notes:

- Rebased onto latest `origin/main` after #19658.
- Kept this PR as draft while CI runs.

### DIFF
--- a/bin/env_knob_catalog.ml
+++ b/bin/env_knob_catalog.ml
@@ -365,7 +365,11 @@ let render entries =
         module_entries;
       Buffer.add_char buf '\n')
     modules;
-  Buffer.contents buf
+  let rendered = Buffer.contents buf in
+  let len = String.length rendered in
+  if len >= 2 && String.sub rendered (len - 2) 2 = "\n\n" then
+    String.sub rendered 0 (len - 1)
+  else rendered
 
 let usage () =
   prerr_endline "usage: env_knob_catalog.exe <output-md-path> [--lib-dir DIR]";
--- a/docs/rfc/RFC-0150-keeper-attention-signal-typed-envelope.md
+++ b/docs/rfc/RFC-0150-keeper-attention-signal-typed-envelope.md
@@ -53,7 +53,7 @@ type attention_signal =
   | Paused_blocked of { since: float }
   | Provider_timeout of { elapsed_sec: float; budget_sec: float }
   | Social_model_fallback of { fallback_provider: string }
-  | Fd_pressure of { headroom_pct: float }
+  | Fd_pressure of { capacity_margin_pct: float }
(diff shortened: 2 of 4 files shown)


jeong-sik:
Closing this draft: the current diff no longer matches the custom tool access title/body, so this should not be merged as that fix.

PR #19541, opened 2026-05-30
refactor(keeper): typed variant for tool_access Custom list

jeong-sik:
## What

Replace the permissive `string list` inside `keeper_meta_tool_access.tool_access` with the closed variant `Tool_name.Keeper.t list`. Invalid tool names are now unrepresentable at the type level.

## Why

`Custom` was essentially a string container that allowed unknown/masc-prefixed names to flow through silently. Every downstream consumer had to defensively `Tool_name.of_string` or do literal string matching. Moving to the closed variant:
- Removes string-classifier workaround patterns (Workaround §2)
- Makes compile-time exhaustive match usable at tool-policy boundaries
- Prevents drift between canonical names and runtime strings

## Changes

| Area | Detail |
|------|--------|
| Core type | `Custom of string list` → `Custom of Tool_name.Keeper.t list` |
| Ingress helper | New `tool_access_of_string_list` (trim → dedupe → `filter_map of_string`) |
| Dead code removed | `legacy_keeper_internal_tool_names`, `legacy_session_min_tool_names`, `migrate_legacy_restricted_tools` |
| Consumers | `keeper_turn_up_create`/`update`, `keeper_runtime` use helper; `keeper_tool_policy` maps `to_string` at `Tool_access_policy` boundary; `agent_tool_execute_runtime` matches `Fs_edit` variant |
| Tests | Mechanical: replace `"tool_edit_file"` literals with `Tool_name.Keeper.Fs_edit` constructors |

## Verification

- `dune build @check` exit 0 (17 files modified)
- No functional behavior change; this is a type-level refactor

## Scope Notes

- `tool_denylist` remains `string list` for now (separate migration)
- `Tool_access_policy` still consumes `string list` at its boundary; we only convert at the call site. A future RFC can propagate the typed variant deeper if warranted.

## Related

- Follow-up to cascade dead-code purge (removes string-based tool presets)
- Closes string/substring classifier anti-pattern

### DIFF
--- a/lib/keeper/agent_tool_execute_runtime.ml
+++ b/lib/keeper/agent_tool_execute_runtime.ml
@@ -323,7 +323,10 @@ let handle_tool_execute
   in
   let write_enabled =
     let (Custom names) = meta.tool_access in
-    List.exists (fun n -> n = "tool_edit_file" || n = "tool_write_file") names
+    List.exists (fun n ->
+      match n with
+      | Tool_name.Keeper.Fs_edit -> true
+      | _ -> false) names
   in
   if has_typed_execute_input_key args
   then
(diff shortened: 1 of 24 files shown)


jeong-sik:
Closing: wrong branch (cascade-dead-code-purge worktree reused). Will reopen from clean branch.

PR #19726, opened 2026-06-01
cascade: split sdk_error_is_server_rejected_parse_error into provider vs model variants

jeong-sik:
Split sdk_error_is_server_rejected_parse_error into provider and model variants for targeted fallback strategies.


jeong-sik (commented):
LGTM. Clean split of parse error classifier. The backward-compat alias (`sdk_error_is_server_rejected_parse_error = sdk_error_is_provider_rejected_parse_error`) is the right migration pattern — no silent behavior change for existing callers. Simplifying the exhaustive match to `_ -> false` is correct since the two positive arms now cover all true cases. One minor note: the model-rejected variant uses substring matching on the error message, which is inherently fragile across provider API changes. Consider adding a comment noting which providers/models produce which substrings, so future maintainers know when to update the needle list.

jeong-sik:
> Copilot: `sdk_error_is_server_rejected_parse_error` is now aliased to the provider variant only...

Already fixed: the deprecated alias at line 110-112 now preserves the old union behaviour:
```ocaml
sdk_error_is_provider_rejected_parse_error err || sdk_error_is_model_rejected_parse_error err
```

> Copilot: The new split helpers aren't part of the public surface...

Intentional — the split helpers are marked internal-only in their doc comments ("Not exposed in the [.mli]"). The union alias `sdk_error_is_server_rejected_parse_error` remains the public entry point. Callers needing targeted strategies can request exposure in a follow-up.

PR #19749, opened 2026-06-01
fix(exec): use canonical file tool names in block hints

jeong-sik:
## Summary\n- replace retired ReadFile/EditFile/WriteFile/SearchFiles names in exec command-blocking hints with Read/Edit/Write/Grep\n- add a worker_dev_tools regression test that command-blocked hints do not reintroduce retired file tool names\n\n## Validation\n- ocamlformat --check lib/exec_policy.ml test/test_worker_dev_tools.ml\n- rg legacy file-tool aliases in lib/exec_policy.ml\n- git diff --check\n\nNote: focused Dune build was attempted with scripts/dune-local.sh build test/test_worker_dev_tools.exe, but the wrapper refused to start because unrelated bare Dune processes were already active on this host.

### DIFF
--- a/lib/exec_policy.ml
+++ b/lib/exec_policy.ml
@@ -53,8 +53,8 @@ let command_blocked_hint ?allowed_commands name =
   let alt =
     match name with
     | "sort" | "uniq" -> " Use rg or jq for filtering."
-    | "sed" | "awk" -> " Use EditFile or WriteFile for file changes."
-    | "find" -> " Use rg --files or SearchFiles."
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Verifier review: Approve. All 8 replacements are correct: ReadFile->Read, EditFile->Edit, WriteFile->Write, SearchFiles->Grep. Regression test covers sed, find, source-code detection, chain/redirect, and injection hints. No retired names remain in exec_policy.ml hints. Draft guard needs human-approved-ready label.

PR #19779, opened 2026-06-01
fix: replace cascade_default with agent_default in runtime_model_string.ml

jeong-sik:
cascade_default was removed in PR #19536 (cascade module deletion). Replace remaining reference with agent_default, matching all other consumer files that already use agent_default/worker_default/deterministic.


jeong-sik (commented):
LGTM. cascade_default to agent_default is correct - agent_default is the canonical profile used everywhere else. The Log.Keeper.error addition in pre_dispatch_error is appropriate.

PR #19892, opened 2026-06-03
fix(observability): typed agent_name option eliminates silent client_name collapse

jeong-sik:
Problem: report.agent_name was string with Option.value ~default:client_name fallback — when no agent was involved (direct client tool call), agent_name silently became indistinguishable from client_name. Changes (5 files, +26/-18): report.agent_name : string -> string option, Failure_envelope.tool_host_failure ~agent_name:string option, Audit_log explicit Option.value with documented rationale, tests updated. Verification: dune build @check EXIT 0. Weird-coupling #5. Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/lib/dashboard_tool_host_events.ml
+++ b/lib/dashboard_tool_host_events.ml
@@ -16,7 +16,12 @@ module Int = Stdlib.Int
 module Float = Stdlib.Float
 
 type report = {
-  agent_name : string;
+  agent_name : string option;
+      (** [None] when no agent is involved — the client called the tool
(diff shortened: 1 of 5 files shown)


jeong-sik:
Closed after domain review: MCP tool calls are client-only (keeper never calls external MCP tools). agent_name and client_name are always the same entity, so the option distinction has no practical value. The correct fix would be to remove agent_name entirely, not make it optional.

PR #19932, opened 2026-06-03
[codex] fix det gate after post-19916 CI repair

jeong-sik:
## Summary
- remove the new Option.value ~default DET/NDT gate hit from command descriptor PR/issue parsing
- keep existing fallback behavior with an explicit Some/None match

## Verification
- scripts/ci/check-determinism-contract.sh
- gtimeout 300 env DUNE_CACHE=disabled DUNE_BUILD_DIR=_build-codex-detgate2 dune build --root . test/test_command_descriptor.exe
- _build-codex-detgate2/default/test/test_command_descriptor.exe

## Context
Follow-up for #19925, which merged but left the PR CI rollup with Meta Guards failing on the new Option.value ~default line.


jeong-sik:
Obsolete after #19920 merged the broader command_descriptor hardening and removed the DET/NDT failure source. Latest origin/main passes scripts/ci/check-determinism-contract.sh.

PR #20067, opened 2026-06-04
docs(rfc): correct RFC-0215 keeper flat-ns fan-out to recursive count

jeong-sik:
## Summary

Correction follow-up to #20064 (RFC-0215, auto-merged before this fix could land). The original RFC's flat-ns classifier was **non-recursive** (`os.listdir` immediate children only), so flat-ns modules living in absorbed subdirectories — `Runtime_*`, `Dashboard_*`, `Operator_*`, `Voice_*`, `Masc_grpc_*`, `Otel_genai` — fell into the opam/other bucket and were not counted as G1 blockers.

## Recursive recount (classify each `lib/**/*.ml` by whether its dir falls under a dir owning a `(library)` dune)

| Metric | Original (stale) | Corrected | Note |
|---|---|---|---|
| keeper → flat-ns mega-lib | ~70 | **92** | labeled lower bound (`(modules :standard)` / `Agent_sdk` submodules still unattributed) |
| keeper → extracted sub-lib | ~63 | **121** | |
| credential cluster fan-out | 3 | **5** | + `Runtime`, `Runtime_agent` |
| registry / error-classify / runtime-binding | 6 each | **7 each** | + `Dashboard_attribution` / `Runtime` / `Runtime_attempt_fsm` |

§1.2 enumeration completed to the full 92-module list.

## What does NOT change

- Recommended first extraction: **credential (`keeper_persona_authoring*`)** — still the smallest decoupling surface (5 point-fix refs vs systemic shared-infra in clusters 2–4). The fan-in-vs-fan-out correction and the ordering both stand.
- `rest → keeper = 0` (structural cycle argument) unchanged.
- Gate (G1–G5) and two-PR-per-cluster shape unchanged.

## Verification
- Recursive classifier walks `lib/`, marks any dir under a `(library ...)`-owning dir as sub-lib, rest as flat-ns mega-lib; spot-checked `config_dir_resolver` → sub-lib (correctly excluded), `Runtime`/`Runtime_agent` → flat-ns (correctly included).
- Numbers labeled grep-approximate; `scripts/audit-sublib-cycle.py` (G1) remains the authoritative per-PR check.
- Doc-only; no build impact.

RFC-WAIVED: RFC document itself, no code change

### DIFF
--- a/docs/rfc/RFC-0215-keeper-sublibrary-extraction-campaign.md
+++ b/docs/rfc/RFC-0215-keeper-sublibrary-extraction-campaign.md
@@ -42,8 +42,8 @@ of `lib/`. Those numbers are stale. Re-derived against the current tree:
 | Direction | Count | Meaning |
 |---|---|---|
 | **rest → keeper** (cycle-creating) | **0** | No *extracted sub-library* holds a compile-time reference to a `Keeper_` module. |
-| keeper → flat-ns mega-lib (`masc`) | **~70 distinct modules** | The G1-relevant forward refs. These are co-resident in `masc` today; they decide whether a keeper sub-lib can be carved without a dune cycle. |
-| keeper → already-extracted sub-lib | **~63 distinct modules** | Not a blocker — declared as `(libraries ...)` deps in the new keeper `dune`. |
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by #20094. This docs-only RFC-0215 correction is now stale and failing current ratchets/numbering; #20094 carries the current RFC-0215 update together with the actual keeper leaf-module prefix work.

PR #20112, opened 2026-06-04
docs(runtime): refresh env knob catalog

jeong-sik:
## Summary

- Refresh `docs/runtime-tunables.md` from `bin/env_knob_catalog.exe`.
- Keep the generated env knob catalog in sync with current `lib/config/env_config_*.ml` sources.

## Product impact

- Restores the CI env knob catalog drift gate for PRs based on current `main`.
- No runtime behavior changes; documentation-only generated catalog update.

## Evidence

- `gtimeout 300 opam exec -- dune exec --root . ./bin/env_knob_catalog.exe -- docs/runtime-tunables.md`
- `git diff --exit-code docs/runtime-tunables.md`
- `git diff --check`
- `bash scripts/check-doc-truth.sh`

## Direct evidence

schema_version: 1
direct_ratio: 4/4
provenance: direct

The commands above were run locally in the PR worktree after regenerating the catalog.

## Review evidence

- Changed file scope is limited to `docs/runtime-tunables.md`.

## Linked issue

- None.

### Commit lineage (auto-synced)

`codex/runtime-tunables-drift-20260604` → `main` · 1 commit(s) · synced 2026-06-04T13:00:51Z

| sha | subject | date |
|---|---|---|
| `58d1526bd` | docs(runtime): refresh env knob catalog | 2026-06-04 |

<sub>Managed by `scripts/pr-sync-body.sh` — do not hand-edit between these markers. The code of record is the diff, not prose above this block.</sub>

PR #20115, opened 2026-06-04
docs(config): refresh runtime tunables catalog

jeong-sik:
## Summary
- refresh docs/runtime-tunables.md from the env knob catalog generator
- fixes the Build and Test env knob catalog drift failure seen after #20110

## Verification
- opam exec -- dune exec --root . ./bin/env_knob_catalog.exe -- docs/runtime-tunables.md
- git diff --check


jeong-sik:
Closed as duplicate of #20112. Both PRs carry the same docs/runtime-tunables.md generated catalog refresh; keep the better documented #20112 as the single lane.

PR #20106, opened 2026-06-04
refactor(keeper): drop persona runtime selection

jeong-sik:
## Summary
- remove stale keeper.runtime_id from persona-authoring schema/save normalization
- reject persona-authored runtime/model selection with runtime.toml [[runtime.assignments]] guidance
- document the RFC-0215 credential decoupling slice that removes the direct Runtime edge

## Verification
- gtimeout 300 env DUNE_CACHE=disabled opam exec -- dune build --root . test/test_keeper_toml.exe
- gtimeout 120 _build/default/test/test_keeper_toml.exe
- ocamlformat --check lib/keeper/keeper_persona_authoring.ml test/test_keeper_toml.ml
- python3 scripts/audit-sublib-cycle.py --root .
- scripts/ocaml-structure-ratchet.sh
- bash scripts/check-doc-truth.sh
- bash scripts/lint/tool-keeper-boundary-ratchet.sh --fail
- bash scripts/stringly-boundary-ratchet.sh
- bash scripts/oas-boundary-ratchet.sh
- git diff --check origin/main...HEAD

PR #20127, opened 2026-06-04
refactor(keeper): narrow persona authoring profile facade

jeong-sik:
Stacked on #20106. Narrows keeper_persona_authoring away from the flat Keeper_types_profile facade by calling the existing TOML normalizer and persona placeholder modules directly. Updates RFC-0215 with the credential pre-work note. Verification: gtimeout 120 dune exec --root . ./test/test_keeper_toml.exe; ocamlformat --check lib/keeper/keeper_persona_authoring.ml; git diff --check origin/codex/keeper-credential-decouple-20260604...HEAD; rg confirms no Keeper_types_profile.* direct calls remain in keeper_persona_authoring.


jeong-sik:
Closing as stale. Current main has removed the split persona authoring module; continuing this stack would revive the retired direction. Replacement cleanup PR: #20163.

PR #20129, opened 2026-06-04
refactor(keeper): split keeper name validation

jeong-sik:
Stacked on #20127. Moves keeper/persona handle validation into a small Keeper_name SSOT, keeps Keeper_config_text.validate_name as the config-facing API, and points keeper_persona_authoring at Keeper_name instead of the full Keeper_config surface. Updates RFC-0215 with the credential pre-work note. Verification: gtimeout 120 dune exec --root . ./test/test_keeper_toml.exe; ocamlformat --check lib/keeper/keeper_name.ml lib/keeper/keeper_name.mli lib/keeper/keeper_config_text.ml lib/keeper/keeper_persona_authoring.ml; git diff --cached --check before commit; analyze_lib_deps shows Keeper_persona_authoring edges now omit Keeper_config.


jeong-sik:
Closing this slice. It adds a new flat Keeper_name module to avoid a Keeper_config import, which is not a meaningful keeper sub-library extraction pre-work cut and adds noise to the stack.

PR #20131, opened 2026-06-04
refactor(keeper): decouple persona authoring fs writes

jeong-sik:
Stacked on #20127. Removes the keeper_persona_authoring dependency on Keeper_fs without adding a new flat Keeper_* module. Persona authoring now creates the profile directory with Fs_compat.mkdir_p and writes profile JSON through Fs_compat.save_file_atomic; Config_dir_resolver.reset remains after successful save. Updates RFC-0215 with the credential pre-work note. Verification: gtimeout 120 dune exec --root . ./test/test_keeper_toml.exe; ocamlformat --check lib/keeper/keeper_persona_authoring.ml; git diff --check origin/codex/persona-authoring-profile-facade-20260604...HEAD; rg confirms no Keeper_fs.* direct calls remain; analyze_lib_deps shows Keeper_persona_authoring outbound edges drop from Keeper_config/Keeper_fs/Keeper_types_profile_persona/Keeper_types_profile_toml_normalizers to Keeper_config/Keeper_types_profile_persona/Keeper_types_profile_toml_normalizers.

PR #20136, opened 2026-06-04
refactor(keeper): move keeper name validation to ids

jeong-sik:
Stacked on #20131. Moves keeper/persona handle validation to the extracted types boundary via Ids.Keeper_id.Name.validate, keeps Keeper_config_text.validate_name as the config-facing facade, and points keeper_persona_authoring at the extracted type boundary instead of full Keeper_config. Updates RFC-0215 with the measured credential fan-out now at 2 direct flat refs. Verification: gtimeout 300 dune exec --root . ./test/test_keeper_toml.exe; ocamlformat --check lib/types/ids.ml lib/types/ids.mli lib/keeper/keeper_config_text.ml lib/keeper/keeper_persona_authoring.ml; git diff --check origin/codex/persona-authoring-fs-compat-20260604...HEAD; rg confirms no Keeper_config.* direct calls remain in keeper_persona_authoring; analyze_lib_deps shows Keeper_persona_authoring outbound edges are now only Keeper_types_profile_persona and Keeper_types_profile_toml_normalizers.

### DIFF
--- a/docs/rfc/RFC-0215-keeper-sublibrary-extraction-campaign.md
+++ b/docs/rfc/RFC-0215-keeper-sublibrary-extraction-campaign.md
@@ -184,6 +184,11 @@ not a hot keeper runtime path, so it can create the persona profile directory
 with `Fs_compat.mkdir_p` and write JSON through `Fs_compat.save_file_atomic`
 without importing the keeper filesystem metrics/cache module.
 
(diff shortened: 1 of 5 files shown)


jeong-sik:
Closing as stale. Current main has removed the split persona authoring module; continuing this stack would revive the retired direction. Replacement cleanup PR: #20163.

PR #20162, opened 2026-06-05
fix(keeper): expose local and docker in keeper_cwd_response signature

jeong-sik:
Expose local and docker functions in keeper_cwd_response signature to fix compile errors on main where they are still invoked directly without sandbox wrappers.


jeong-sik:
Closing as obsolete. After rebasing onto current main, this branch has no remaining diff: origin/main already exposes Keeper_cwd_response.local and Keeper_cwd_response.docker in lib/keeper/keeper_cwd_response.mli via the merged #20156/#20158 stack.

PR #20203, opened 2026-06-05
refactor(tools): extract tools module into masc.tool_schemas sub-library

jeong-sik:
Extracts lib/tools.ml and lib/tools.mli into a dedicated masc.tool_schemas sub-library, resolving compilation and namespace issues across unit tests.


jeong-sik:
Closing as stale after the Prometheus purge and main CI repair sequence (#20189/#20193). This draft still reintroduces masc.prometheus/root Prometheus shims and old missing-module compatibility wrappers while trying to extract Tools into tool_schemas. If the tools extraction is still wanted, it should be re-cut as a fresh narrow PR on current main without Prometheus/no_tool_capable residue or build-repair shims.

jeong-sik:
Closing again: the refreshed head still reintroduces masc.prometheus/root Prometheus shims and compatibility wrappers removed by the Prometheus purge/main CI repair path. The Tools extraction should be re-cut from current main as a narrow PR without restoring Prometheus or old build-repair shims.

PR #20205, opened 2026-06-05
fix(dashboard): keeper chat multimodal bug fixes

jeong-sik:
## Summary

Fix 3 bugs found in keeper chat multimodal implementation (PR #20200).

## Bug Fixes

| # | Bug | Fix |
|---|-----|-----|
| 1 | **Memory leak** in `resizeImage` — `URL.createObjectURL` created blob URLs that were never revoked. | Store object URL in a local variable and call `URL.revokeObjectURL` in both `onload` and `onerror` handlers. |
| 2 | **Attachments not displayed** in chat history — `toConversationEntry` dropped `msg.attachments` instead of passing them to `ChatTranscript`. | Pass attachments via `details.rawPayload.attachments` so the transcript renderer can access them. |
| 3 | **File input re-selection blocked** — selecting the same file twice did not trigger `onChange` because the input value was unchanged. | Clear `target.value = ''` after processing the file list. |

## Verification

- [x] `tsc --noEmit` clean
- [x] `vitest run src/keeper-chat-store.test.ts` — 21/21 passed

## Affected File

- `dashboard/src/components/keeper-chat-panel.ts`

### DIFF
--- a/dashboard/src/components/keeper-chat-panel.ts
+++ b/dashboard/src/components/keeper-chat-panel.ts
@@ -67,7 +67,9 @@ async function resizeImage(file: File, maxWidth = 1920): Promise<File> {
   if (!file.type.startsWith('image/')) return file
   return new Promise((resolve) => {
     const img = new Image()
+    const objectUrl = URL.createObjectURL(file)
     img.onload = () => {
+      URL.revokeObjectURL(objectUrl)
       if (img.width <= maxWidth) { resolve(file); return }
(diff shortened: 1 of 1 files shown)


jeong-sik:
Follow-up for the post-merge Codex attachment review: #20208 addresses https://github.com/jeong-sik/masc-mcp/pull/20205#discussion_r3361017009 by rendering keeper chat image/file attachments in the normal message body path instead of relying on hidden raw metadata.

PR #20233, opened 2026-06-05
fix(docs): update keeper measurement code ref

jeong-sik:
## Summary
- update the stale keeper measurement code_ref to the current keeper_metrics path
- refresh last_verified for the RFC frontmatter

## Why
Meta Guards on #20231 failed because docs/design/tool-calling-quality-and-self-healing-rfc.md referenced lib/keeper_measurement/keeper_measurement.ml, which no longer exists after the keeper metrics leaf move.

## Verification
- bash scripts/check-doc-truth.sh
- scripts/sync-oas-pin-docs.sh --check
- scripts/audit-keeper-tool-boundary-matrix.sh
- git diff --check

### DIFF
--- a/docs/design/tool-calling-quality-and-self-healing-rfc.md
+++ b/docs/design/tool-calling-quality-and-self-healing-rfc.md
@@ -1,9 +1,9 @@
 ---
 status: reference
-last_verified: 2026-04-17
+last_verified: 2026-06-05
 code_refs:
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by #20126/#20184, which already carry the docs/design/tool-calling-quality-and-self-healing-rfc.md code_ref refresh to lib/keeper_metrics/keeper_measurement.ml. Branch left intact.

PR #20304, opened 2026-06-06
fix(keeper): keep state report dispatch singular

jeong-sik:
## Summary
- keep only one keeper_report_state dispatch arm so warning-as-error does not see a redundant State_report case
- replace Option.first_some from #20302 with the existing explicit option fallback pattern for OCaml compatibility

## Verification
- git diff --check
- env DUNE_CACHE=disabled DUNE_BUILD_DIR=_build_codex_20260606m dune build --root . test/test_keeper_validation_breaker_exempt.exe test/test_keeper_tool_name.exe test/test_keeper_state_snapshot_json.exe test/test_keeper_effective_meta_overlay.exe
- _build_codex_20260606m/default/test/test_keeper_validation_breaker_exempt.exe
- _build_codex_20260606m/default/test/test_keeper_tool_name.exe
- _build_codex_20260606m/default/test/test_keeper_state_snapshot_json.exe
- _build_codex_20260606m/default/test/test_keeper_effective_meta_overlay.exe

### DIFF
--- a/lib/keeper/keeper_tool_task_runtime.ml
+++ b/lib/keeper/keeper_tool_task_runtime.ml
@@ -795,5 +795,4 @@ let handle_keeper_task_tool
         ~ok:(Tool_result.is_success transition_result)
         ~message:(Tool_result.message transition_result)
         ())
-    | State_report -> state_report_result_json args
 ;;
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #20306; the effective state-report dispatch/build fix is already on main. Closing this stale draft to remove blocked PR noise.

PR #20305, opened 2026-06-06
fix(keeper): repair state report main build

jeong-sik:
## Summary
- keep exactly one State_report dispatch in keeper_task runtime so the closed task_op match is exhaustive without a redundant case
- replace non-existent Option.first_some with the local explicit fallback pattern

## Tests
- env DUNE_CACHE=disabled DUNE_BUILD_DIR=_build-codex-state-report-match dune build --root . test/test_keeper_state_snapshot_json.exe
- env DUNE_CACHE=disabled DUNE_BUILD_DIR=_build-codex-state-report-match dune exec --root . test/test_keeper_state_snapshot_json.exe

### DIFF
--- a/lib/keeper/keeper_tool_task_runtime.ml
+++ b/lib/keeper/keeper_tool_task_runtime.ml
@@ -732,10 +732,9 @@ let handle_keeper_task_tool
             | None -> [])
          @ claimed_task_fields
          @ wip_admission_result_fields wip_rejections
-         @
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #20306; current main already contains the build repair, and this draft only leaves stale formatting/noise against main. Closing to remove blocked PR noise.

PR #20350, opened 2026-06-06
Revert status bridge narrative blocker inference removal

jeong-sik:
## Summary

Reverts #20344 (`Remove narrative blocker inference from status bridge`).

This removes the already-merged status bridge change from `main` so the work can be reopened as a fresh PR instead of staying landed from the accidental/too-fast merge path.

Also adds the missing public `.mli` files for the Slack and Telegram Channel Gate state modules. Those modules are already referenced from server/tests, so they need explicit public interfaces for the OCaml structure ratchet.

## Validation

- `ocamlformat --check lib/gate/channel_gate_slack_state.mli lib/gate/channel_gate_telegram_state.mli lib/keeper/keeper_status_bridge.ml test/test_keeper_status_bridge.ml`
- `scripts/ocaml-structure-ratchet.sh`
- `git diff --check`
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_keeper_status_bridge.exe` (passed before the conflict-free rebase)
- `./_build/default/test/test_keeper_status_bridge.exe` (passed before the conflict-free rebase)

Note: a post-amend focused build including `test/test_channel_gate_connector_routes.exe` was queued behind unrelated repo-wide Dune locks and was stopped; CI will validate the pushed head.

### DIFF
--- a/lib/keeper/keeper_status_bridge.ml
+++ b/lib/keeper/keeper_status_bridge.ml
@@ -156,17 +156,148 @@ let runtime_keepalive_started_at (config : Workspace_utils.config) (meta : keepe
 include Keeper_status_bridge_blocker
 
 
-let runtime_blocker_surface_opt (config : Workspace_utils.config) (meta : keeper_meta) =
-  match meta.runtime.last_blocker with
-  | Some info ->
-    Some (runtime_blocker_surface_of_typed_class ~summary:info.detail info.klass)
-  | None ->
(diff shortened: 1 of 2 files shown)


jeong-sik:
Revert of #20344 is unnecessary — the original removal was intentional and correct. String substring blocker inference should stay deleted (typed sources only).

PR #20381, opened 2026-06-06
fix(keeper): remove turn slot retry reacquire residue

jeong-sik:
## Summary
- keep timeout/retry boundaries inside the original keeper turn slot instead of exposing release/reacquire controls again
- align OAS handler signatures for the remaining observation-only turn slot control
- fix voice runtime tests now that handle_voice_tool has a final unit argument

## Validation
- ocamlformat --check lib/keeper/keeper_tool_voice_runtime.ml lib/keeper/keeper_tools_oas_bundle.ml lib/keeper/keeper_tools_oas_bundle.mli lib/keeper/keeper_tools_oas_handler.mli lib/keeper/keeper_tools_oas_handler_exec.mli lib/keeper/keeper_unified_turn_execution.ml test/test_voice_runtime_overlay.ml
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build --cache=disabled lib/masc.cma test/test_voice_runtime_overlay.exe test/test_keeper_turn_slot_holders.exe
- _build/default/test/test_keeper_turn_slot_holders.exe
- _build/default/test/test_voice_runtime_overlay.exe

### DIFF
--- a/lib/keeper/keeper_tool_voice_runtime.ml
+++ b/lib/keeper/keeper_tool_voice_runtime.ml
@@ -153,48 +153,22 @@ let handle_speak
            (keeper_text_fallback_json ~agent_id:meta.name ~message)
            memory_status))
 
-let handle_listen ~(meta : keeper_meta) ~(args : Yojson.Safe.t) ?turn_slot_control () =
+let handle_listen ~(meta : keeper_meta) ~(args : Yojson.Safe.t) ?turn_slot_control:_ () =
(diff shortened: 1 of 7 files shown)


jeong-sik:
Closing as wrong direction for the current Keeper slot policy. This patch removes voice-listen release/reacquire around blocking I/O, while the accepted direction is to yield the slot during long blocking voice work and reacquire afterward.

PR #20385, opened 2026-06-06
style(dashboard): format keeper detail rail

jeong-sik:
## Summary

- format `dashboard_bonsai/src/keepers_directory.ml` after #20374 merged before the formatter fix landed
- no intended behavior change

## Verification

- `ocamlformat --enable-outside-detected-project --check dashboard_bonsai/src/keepers_directory.ml`
- `ocamlc -stop-after parsing dashboard_bonsai/src/keepers_directory.ml`
- `git diff --check`

## Local build note

- Full Bonsai JS build is still not runnable in this local switch because `bonsai_web` and `ppx_jane` are not installed.


jeong-sik:
Closing this draft because it only formats the Bonsai dashboard path; the intended target is dashboard/. I am moving the fix to the real dashboard surface.

PR #20433, opened 2026-06-07
fix(keeper): remove Runtime_lane_capacity and liveness guard from try_provider

jeong-sik:
## Problem

Multiple overlapping gates cause unnecessary complexity and false failures:

1. Keeper_turn_capacity: global 32 + per-keeper 2 — manages turn admission.
2. Runtime_lane_capacity: per-lane max_concurrent — adds a second HTTP-level
   gate that causes 180s lane capacity timeouts when 16+ keepers run
   concurrently.
3. Keeper_attempt_liveness: FSM-based no_first_token kill — adds a third
   stream-monitoring layer that kills slow providers after 10min per attempt,
   leading to 30min total latency after retries.

The operator already controls provider load by adjusting keeper count.
Provider stall is an OAS-level concern (stream_idle_timeout_s).

## Fix

Remove both gates from keeper_turn_driver_try_provider.ml:
- Delete Runtime_lane_capacity wrapper — provider load is operator-managed.
- Delete liveness FSM observer — OAS handles stream stall.
- Keep only per-provider_timeout_s (HTTP-level timeout via OAS).

## Rationale

- Standard agent SDKs (OpenAI, Anthropic, LangChain, Vercel AI) rely on
  HTTP-level timeouts only. No separate stream-liveness FSM.
- If provider is overloaded, operator reduces keeper count. No numeric
  max_concurrent needed.

🤖 Generated with [Claude Code](https://claude.com/code)

### DIFF
--- a/lib/keeper/keeper_turn_driver_try_provider.ml
+++ b/lib/keeper/keeper_turn_driver_try_provider.ml
@@ -298,105 +298,70 @@ let run_try_provider
   match config_result with
   | Error err -> Error err, None, None
   | Ok config ->
-    (* Liveness observer removed. Stream stall detection is handled by OAS's
-       stream_idle_timeout_s (default 120s per inter-line gap). The former
-       liveness FSM was a dual-kill system that independently terminated streams
-       via tick-fiber + Enforce-mode Switch.fail, causing no_first_token errors
-       on legitimately slow providers. Standard agent SDKs (OpenAI, Anthropic,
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #20436 which was merged to main. #20436 introduces Runtime_lane_capacity with per-binding max_concurrent from runtime.toml, which is the opposite direction of this PR (removing the gate). Closing as outdated.

PR #20434, opened 2026-06-07
refactor(keeper): Runtime_lane_capacity fail-fast, remove timeout/spin-wait

jeong-sik:
## Summary

Remove timeout and spin-wait from `Runtime_lane_capacity`. Admission gate now fails immediately when the lane is at capacity.

## Why

- `Runtime_lane_capacity` was blocking for up to 180s with `Eio.Fiber.yield()` spin-wait
- 20+ keepers hitting per-lane limits (`8/8`, `8/4` slots) caused cascading timeouts
- OAS already has `stream_idle_timeout_s` for provider stall detection — admission should not duplicate timeout semantics

## Changes

- `acquire_lane_capacity`: removed `~timeout_s`, removed spin-loop, returns `Error` immediately when full
- `with_lane_capacity`: removed `?timeout_s`, removed `~capacity_wait_ms` parameter
- `rejection` type: removed `waited_ms` field
- `acquired` type: removed `wait_ms` field
- Caller (`keeper_turn_driver_try_provider.ml`): updated to match new signatures

## Design

Fail-fast admission. The caller (keeper turn driver) decides retry / fallback / error surfacing. No CPU-burning spin-wait.


jeong-sik:
Superseded by #20436 which was merged to main. #20436 establishes Runtime_lane_capacity with max_concurrent admission; this PR's fail-fast refactor is no longer applicable in its current form. If spin-wait removal is still desired, please rebase onto main and open a new PR.

PR #20441, opened 2026-06-08
fix(keeper): restore execute admission allowlists

jeong-sik:
Backport follow-up for merged PR #20402. Restores typed Execute allowlist enforcement and preserves watchdog crash accounting on resolved path.


jeong-sik:
Superseded by #20442 which rebases onto current main and resolves conflicts.

PR #20468, opened 2026-06-08
feat(exec): P6 retire pipeline-wide flat word-list floor (RFC-0208)

jeong-sik:
P6 of the Shell IR Effect Proof Design.

**Change**
Removes the cross-stage flattening safety net (flat_floor) from Shell_ir_risk.classify. The per-stage word-list floor inside decision_of_simple is retained — each Simple node still gets its own classify_words lower bound.

**Rationale**
- RFC-0208 P0 introduced per-stage composition via max_risk across pipeline stages
- P2 differential harness verified structural floor-redundancy at 100% (only string-borne/gh remains load-bearing by design)
- flat_floor concatenates all stage words and matches only the head token, which is now redundant because per-stage typed+floor composition already captures every stage's risk

**Safety verification**
- Full lib/exec/test/ suite passes (EXIT=0)
- P2 harness: monotone-safety invariant preserved (full >= typed_only && full >= per_stage_floor)
- No behavior change for Simple commands; only Pipeline flat-staging removed

**Boundary**
- Touches only Shell_ir_risk.classify (core classifier)
- Exec_effect, Checked_shell_ir, and dispatch paths unchanged

### DIFF
--- a/lib/exec/exec_effect.ml
+++ b/lib/exec/exec_effect.ml
@@ -1,7 +1,7 @@
 (** Exec_effect — Effect axis types for Shell IR execution.
 
-    P0 of the Shell IR Effect Proof Design (RFC-0208 extension).
-    See [exec_effect.mli] for specification.
+    P1 of the Shell IR Effect Proof Design (RFC-0208 extension).
+    Fine-grained per-constructor decomposition on top of P0.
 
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing per request — will revisit with cleaner approach.

PR #20483, opened 2026-06-08
refactor(keeper): move Shell_ir_risk observability helpers into facade

jeong-sik:
Re-export is_destructive, risk_class, typed_hit_of_ir, and string_of_risk_class through Keeper_tool_execute_shell_ir so that keeper_tool_execute_runtime.ml no longer reaches directly into Masc_exec.Shell_ir_risk.

Keeper_tool_execute_shell_ir is the canonical facade between keeper execution runtime and the core Shell IR risk classifier. Narrowing the keeper->core dependency surface makes future refactors safer.

Build: dune build lib/keeper_tooling/ lib/keeper/ OK
Tests: test_shell_ir_effect_projection 8/8, test_shell_ir_risk 20/20 OK

### DIFF
--- a/lib/keeper/keeper_tool_execute_runtime.ml
+++ b/lib/keeper/keeper_tool_execute_runtime.ml
@@ -271,7 +271,7 @@ let handle_tool_execute_typed
             meta.name
             error
             (message_for_log reason)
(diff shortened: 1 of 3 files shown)


jeong-sik:
Closing: simple re-export facade adds indirection without value. Keeper can reference Shell_ir_risk public API directly.

PR #20599, opened 2026-06-09
fix(telemetry): break dep cycle + fix type error from #20590

jeong-sik:
Three build errors introduced by #20590:

1. **Dependency cycle**: Keeper_event_publisher took Keeper_turn_runtime_budget.runtime_execution, creating Audit_log→...→Governance_pipeline→Audit_log cycle. Fixed by taking primitive parameters.
2. **Type error**: max_context_resolution is a record, not a variant — Resolved_to pattern was invalid. Fixed by using .effective_budget field.
3. **Missing .mli export**: publish_runtime_execution_built was defined but not exported.


jeong-sik:
Superseded by #20598 which merged the same fix independently.

PR #20727, opened 2026-06-10
fix(voice): restore sync contract in handle_speak (agent_speak not enqueue_agent_speak)

jeong-sik:
## Defect

`handle_speak` at `keeper_tool_voice_runtime.ml:115` called `enqueue_agent_speak` — an async queue function that returns `status="queued"` immediately. The tool contract expects a synchronous result carrying the actual playback outcome.

## Fix

Changed to `agent_speak` (synchronous — blocks until playback completes, returns `played` or `error` instead of `queued`).

Both signatures return `(Yojson.Safe.t, string) result`, so the call site compiles identically. The semantic change is that the response now reflects actual playback status.

## Timeline

- The 60s playback timeout restart defect (Defect 1) was already resolved upstream by #20512 (`8cf76afa`) which deleted the caller-side timeout wrapper entirely.
- This PR contains **only** the sync-contract fix (Defect 2).

## RCA references
- `keeper_tool_voice_runtime.ml:115` — enqueue_agent_speak → agent_speak
- `voice_bridge.mli:45-56` — both agent_speak and enqueue_agent_speak share the same return type

### DIFF
--- a/lib/keeper/keeper_tool_voice_runtime.ml
+++ b/lib/keeper/keeper_tool_voice_runtime.ml
@@ -112,7 +112,7 @@ let handle_speak
     with
     | Some sw, Some clock, Some net ->
       (match
-         Voice_bridge.enqueue_agent_speak
+         Voice_bridge.agent_speak
            ~sw
            ~clock
            ~net


jeong-sik:
Closing as superseded by #20724.

The review concern on this PR is valid for this stale branch: changing `keeper_voice_speak` from `enqueue_agent_speak` to raw `agent_speak` restored a synchronous result but bypassed the old queue serialization.

Current `main` already landed the fuller fix in #20724:

- `Voice_bridge.enqueue_agent_speak` and the fire-and-forget queue were removed.
- `Voice_bridge.agent_speak` is now the single speak path and serializes concurrent local playback with the global playback mutex.
- `keeper_voice_speak` records memory only for `status="spoken"` and surfaces TTS/playback failures synchronously.

So this PR is obsolete rather than needing another patch on top. The red Lint/CI Gate here is also from running this stale PR against current main's SSOT ratchet state, not from a unique remaining voice-contract change.

PR #20768, opened 2026-06-10
fix(board_diversity): O(n) sliding-window find_dead_zones with correct overlapping-run detection

jeong-sik:
Fixes two performance/correctness issues in `find_dead_zones`:

1. **List.nth O(n²) → O(n) sliding window**: Replaced triple `List.nth` calls (each O(n), making the whole scan O(n²)) with a single-pass sliding-window iteration over the list.

2. **Overlapping-run detection**: Old code advanced by `i + 3` after matching a run, so `[A,A,A,A,A]` only returned zone at index 2, missing zones 3,4. New code advances by 1, so every 3+ consecutive same-author run correctly reports all zone positions.

Resolves task-777.

### DIFF
--- a/lib/board/board_diversity.ml
+++ b/lib/board/board_diversity.ml
@@ -42,19 +42,27 @@ let insert_at idx el lst =
   before @ [el] @ after
 
 (** Detect runs of 3+ consecutive posts by the same author.
-    Returns zone positions (0-based index of the 3rd post in each run). *)
+    Returns zone positions (0-based index of the 3rd post in each run).
+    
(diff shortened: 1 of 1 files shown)


jeong-sik:
스택 base #20766 close에 따라 함께 닫습니다. O(n) sliding-window 재작성 자체는 정확하고 중첩 run 동작 변화 문서화도 좋았습니다 — base 재제출 시 이 알고리즘을 그대로 가져갈 것을 권장합니다. 상세: https://github.com/jeong-sik/masc/pull/20766#issuecomment-4675832634

PR #20770, opened 2026-06-10
fix(board_diversity): correct overlapping-run detection in find_dead_zones + add test coverage

jeong-sik:
Fixes task-778: overlapping-run detection bug in `find_dead_zones`.

**Bug:** The scan skipped overlapping runs by advancing `i + 3` after every match. A block `[A,A,A,A,A]` only flagged position 2, missing positions 3 and 4.

**Fix:** Changed `scan (i + 3) ...` → `scan (i + 1) ...` so every consecutive index is checked. After finding a 3-run starting at `i`, the next iteration checks `i+1, i+2, i+3` — catching all overlapping runs.

**Test coverage (new file):** `test/test_board_diversity.ml`
- 10 Alcotest unit tests: `author_frequencies` (empty/single/mixed), `find_dead_zones` (no zones, single 5-run, overlapping, exact 3, two separate runs), `rerank_for_diversity` (Recent no-op, Hot preserves authors)
- 6 QCheck property-based tests (100-200 iter each): sum invariant, zone validity, no false negatives, length preservation, no duplicates, no posts added
- Dune stanza: `test/stanzas/test_board_diversity.inc`

**Diff:** lib/board/board_diversity.ml (1 line), test/test_board_diversity.ml (+261), test/stanzas/test_board_diversity.inc (new), test/dune (stanza include)

### DIFF
--- a/lib/board/board_diversity.ml
+++ b/lib/board/board_diversity.ml
@@ -51,7 +51,7 @@ let find_dead_zones (posts : Board_types.post list) =
       | a, b, c when
           Board_types.Agent_id.equal a.author b.author
           && Board_types.Agent_id.equal b.author c.author ->
-        scan (i + 3) ((i + 2) :: acc)
+        scan (i + 1) ((i + 2) :: acc)
       | _ -> scan (i + 1) acc
   in
   scan 0 []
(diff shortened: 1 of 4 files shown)


jeong-sik:
Duplicate — taskmaster already applied the same fix and committed. This PR is superseded by the existing fix.

PR #20809, opened 2026-06-11
feat(health_score): add board_health_score to autonomous cooldown

anyang-keepers:
Add Health_score_cooldown skip_reason variant, board_health_score field to world_observation, and health_score_multiplier in effective_scheduled_autonomous cooldown.

### DIFF
--- a/lib/keeper/keeper_world_observation.ml
+++ b/lib/keeper/keeper_world_observation.ml
@@ -42,6 +42,7 @@ type world_observation =
   ; backlog_updated_since_last_scheduled_autonomous : bool
   ; active_agent_count : int
   ; connected_surfaces : Gate_surface.surface_presence list
+  ; board_health_score : float option
   }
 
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing as superseded by the current main implementation.

This draft only adds `board_health_score` to `world_observation` with `None` at both observation sites, plus a `Health_score_cooldown` skip_reason variant, but it does not wire that field into the scheduler decision path. Current main already has the health-score cooldown path in `effective_scheduled_autonomous_cooldown ?board_health_score`, reads `Board_curation.latest_snapshot ()`, and has focused cooldown tests. Keeping this PR open would reintroduce a stale field/variant shape rather than advance the scheduler.

If we need a follow-up, it should be a fresh, narrow PR against current main that either removes remaining stale health-score residue or adds dashboard/telemetry visibility for the existing scheduler behavior.

PR #20829, opened 2026-06-11
feat(keeper): RFC-0231 tiered Memory OS + LLM Librarian integration

jeong-sik:
Implements the approved RFC-0231 redesign to stop compaction-induced survivorship bias.

### What changed
- New Memory OS modules: types, I/O, policy, librarian, librarian_runtime
- Compaction trigger now extracts older messages and persists structured episodes+claims before OAS reducer runs
- Deterministic retention scoring: confidence × recency × access
- LLM Librarian uses default runtime' direct-completion provider via Llm_provider.Complete

### Verification
- `test/test_keeper_memory_os.exe` → 4/4 pass
- `lib/masc.cma` build succeeds

### Operator follow-up
- After rollout, reset sangsu' `.memory.jsonl` and `continuity_summary` to clear prior emoji contamination.

### DIFF
--- a/config/prompts/keeper.librarian.episode_extraction.md
+++ b/config/prompts/keeper.librarian.episode_extraction.md
@@ -1,12 +1,10 @@
-(** Keeper_librarian_prompts — prompt templates for the Memory OS librarian.
+---
+description: Memory OS librarian episode extraction prompt
+category: keeper
+template_variables: [conversation_history]
+---
 
-    Keeping prompts outside the pure logic module makes them easier to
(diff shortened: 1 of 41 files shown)


jeong-sik:
과도한 단일 PR (39파일/2121줄/테스트 4개). hand-rolled take/drop, catch-all 16건, 테스트 커버리지 불충분. Phase 분할 후 재제출 필요: P1(types+io+policy) → P2(recall+prompts) → P3(librarian+runtime) → P4(integration).

PR #20842, opened 2026-06-11
fix(workspace): add clear_stale_agent_task after write_backlog in transition_task_r

jeong-sik:
## Problem

`transition_task_r` performs an atomic backlog commit via `write_backlog` but never calls `clear_stale_agent_task` to clean up stale in-memory agent task assignments. When tasks transition (claim, done, cancel, release), the agent's task cache may retain stale `current_task` references.

## Evidence

- `clear_stale_agent_task` defined in `task_cache_invariant.ml:42`, already called in `workspace_broadcast.ml:108` and `mention_inbox.ml:86`
- Bug found by tech_glutton (board p-29492f56329b5d7d9101871a22f3ed26)
- Could contribute to double-done race conditions (task-791) and stale agent task assignments

## Fix
1 file changed, +5 lines inserted in `lib/workspace/workspace_task_transitions.ml`

### DIFF
--- a/lib/keeper/keeper_unified_turn.ml
+++ b/lib/keeper/keeper_unified_turn.ml
@@ -889,6 +889,14 @@ dominant source of the observed CAS race exhaustion after
                     ~is_auto_recoverable
                     ~err
                     ~error_text:e_str;
+                  (* RFC-0221 §3.4: emit turn_completed telemetry on all exit paths
+                     after Agent.run() — success path emits via
+                     Keeper_unified_turn_success → keeper_unified_metrics_snapshot;
+                     failure path emits here directly. *)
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing instead of rewriting. After rebasing this branch onto current main, the only remaining workspace diff is a second duplicate call to Task_cache_invariant.clear_stale_agent_task in transition_task_r. The intended transition_task_r cache clear already landed on main in #20822 (7670c4085), and the unrelated task-801 telemetry commit that was mixed into this branch already landed in #20839. Keeping this PR open would add duplicate cache_desync.cleared logging and an extra agent-file lock, so there is no unique patch left here.

PR #20845, opened 2026-06-11
fix(workspace): add clear_stale_agent_task after write_backlog in claim_r and force_release_task_r

jeong-sik:
## Problem

`workspace_task_claim.ml` (2 sites) and `workspace_task.ml:138` execute `write_backlog` but never call `Task_cache_invariant.clear_stale_agent_task` after the atomic backlog commit, allowing stale `current_task` references to persist in the in-memory agent context cache.

Systemic pattern found in board p-b65c23a46a2f8f1545778acb07f35081.

## Fix
**2 files changed, +9 lines**:

1. `workspace_task_claim.ml:189` — `Claimed_ok` path, adds `clear_stale_agent_task` with `~status:Claimed`
2. `workspace_task_claim.ml:228` — `Claimed_verification` path, adds with `~status:AwaitingVerification`
3. `workspace_task.ml:138` — `force_release_task_r`, adds with `~status:Cancelled`

Same pattern as PR #20842 (workspace_task_transitions.ml), PR #20844 (workspace_task_schedule.ml), workspace_broadcast.ml:108, mention_inbox.ml:86.

PR #20844, opened 2026-06-11
fix(workspace): add clear_stale_agent_task after 3 write_backlog calls in claim_next

jeong-sik:
## Problem

`workspace_task_schedule.ml` has 3 `write_backlog` call sites that never call `Task_cache_invariant.clear_stale_agent_task` after the atomic backlog commit. The in-memory agent context cache can retain stale `current_task` references.

Discovered while investigating the systemic pattern found in board p-b65c23a46a2f8f1545778acb07f35081. The same fix was already applied to `workspace_task_transitions.ml` in PR #20842.

## Fix
**1 file changed, +13 lines inserted** in `lib/workspace/workspace_task_schedule.ml`:

1. Release-no-unclaimed path: add `clear_stale_agent_task` with `~status:Todo` after `write_backlog` for released task
2. Release-no-eligible path: same fix
3. Claim path: add `clear_stale_agent_task` with `~status:Claimed` after `write_backlog` for `claimed_task`

PR #20846, opened 2026-06-11
fix(workspace): add clear_stale_agent_task in release_task_r after transition_task_r write_backlog

jeong-sik:
## Problem

`release_task_r` delegates to `transition_task_r` which executes `write_backlog` but does not clear the stale agent task cache. When an agent's task is released, the in-memory `current_task` reference can remain stale.

## Fix
Wrap `transition_task_r` result in `release_task_r`: on `Ok`, call `Task_cache_invariant.clear_stale_agent_task` with `~status:Todo` (release transitions task back to Todo). On `Error`, skip (failed transition means no backlog mutation).

**1 file changed, +17/-8 lines** — `lib/workspace/workspace_task.ml`

Same pattern as PR #20845 (force_release_task_r), PR #20844 (claim_next_r), PR #20842 (transition_task_r), workspace_broadcast.ml:108, mention_inbox.ml:86.

### DIFF
--- a/lib/workspace/workspace_task.ml
+++ b/lib/workspace/workspace_task.ml
@@ -12,14 +12,23 @@ include Workspace_task_transitions
 let release_task_r config ~agent_name ~task_id ?expected_version ?handoff_context ()
   : string Masc_domain.masc_result
   =
-  transition_task_r
-    config
-    ~agent_name
-    ~task_id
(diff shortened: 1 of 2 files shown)


jeong-sik:
N-of-M anti-pattern. #20844/#20845와 동일 수정. 단일 PR + 래퍼로 교체.

PR #20855, opened 2026-06-11
feat(keeper): add publish_telemetry_event generic function to publisher module

jeong-sik:
Adds a generic `publish_telemetry_event ~event_name ~payload` function to `Keeper_event_publisher`, following the same pattern as existing typed publishers.

This enables callers outside the dispatch seam to emit telemetry events without knowing the internal `Custom("telemetry_event", _)` event bus name.

The function delegates to `Masc_event_bus.broadcast` which requires the Eio runtime context — it cannot be exercised in a pure unit test outside the Eio test infrastructure. Existing integration tests in Group 3 (Eio-native) cover the publisher module via broader test suites.

# ci:skip-test-coverage

### DIFF
--- a/lib/core/eval_tool_selector.ml
+++ b/lib/core/eval_tool_selector.ml
@@ -36,7 +36,7 @@ let to_yojson selector =
         ; ("key", `String key)
         ; ("value", `String value)
         ]
-  | Tool_name _ | Descriptor_id _ | Runtime_handler _ | Eval_tag _ ->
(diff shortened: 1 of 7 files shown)


jeong-sik:
Closing as the wrong abstraction boundary. The PR widens keeper telemetry publishing to a generic string event_name surface and is currently conflicting; the useful existing pattern is typed/specialized publisher functions with stable payload shape. Re-cut only if there is a concrete telemetry event consumer that cannot be represented as a typed publisher, and keep unrelated eval-selector lint changes out of that PR.

PR #20858, opened 2026-06-11
Phase A: OTel metrics adapter module (task-814)

jeong-sik:
## Summary

New `lib/otel/otel_metrics.ml` / `.mli` providing OTel-backed metrics API matching the Prometheus surface.

### Changes
- Dual-writes to Prometheus store (for /metrics endpoint) and OTel OTLP accumulator
- API: inc_counter, set_gauge, inc_gauge, dec_gauge, observe_histogram, register_*, flush, to_prometheus_text
- Feature-flagged via MASC_OTEL_ENABLED
- Builds successfully

### Next phases
- **Phase B**: Migrate ~80 callers from Prometheus.* to Otel_metrics.*
- **Phase C**: Remove Prometheus facade modules

### DIFF
--- a/lib/otel/otel_metrics.ml
+++ b/lib/otel/otel_metrics.ml
@@ -1,0 +1,151 @@
+(** OTel Metrics — OpenTelemetry metrics adapter for masc-mcp.
+
+    Provides the same API surface as [Prometheus.*] but backed by OTel's
+    [Opentelemetry.Metrics] protocol.  During the migration window this module
+    dual-writes to both OTel and the in-memory Prometheus store so the
+    /metrics text endpoint continues to work.
+
+    When [MASC_OTEL_ENABLED=true] and the exporter is active, metrics are
+    accumulated and flushed via OTLP.  When disabled (default), the module
+    silently drops OTel-side writes and only populates the Prometheus store.
+
+    @since 2.150.0
+*)
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Closing this as stale/superseded by the current OTel metrics path on main.\n\nWhy:\n- main already has the OTel bridge in `lib/otel_metrics` plus `Otel_metric_store.register_otel_source_once`, using `Opentelemetry.Metrics_callbacks.register` to export the in-process metric-store snapshot on the backend tick. That preserves labels and current gauge values from the existing store instead of introducing a second push buffer.\n- This PR adds a separate `lib/otel/otel_metrics.ml` module, which collides with the existing `Otel_metrics` API shape expected by `lib/otel_metric_store.ml` (for example `Otel_metrics.Counter`) and CI is already failing with `Unbound constructor Otel_metrics.Counter` plus `Unbound module Prometheus_store`.\n- The open review threads are valid for this implementation: the buffer is not wired into the exporter loop, labels are dropped in OTLP points, `inc_gauge`/`dec_gauge` export deltas instead of live gauge values, and histogram export produces count/bucket inconsistencies.\n\nIf we need more OTel work, the follow-up should extend the existing `lib/otel_metrics` / `lib/otel_metric_store_core` path with focused tests rather than land this older Prometheus-style adapter.

PR #20866, opened 2026-06-11
test(keeper_msg_async): add timeout_sec=0.0 / infinity boundary tests to test_keeper_mutex_coverage.ml

jeong-sik:
Adds two boundary test cases to the `test_keeper_mutex_coverage` suite:

1. **timeout_sec=0.0** — triggers immediate timeout on a never-completing fiber
2. **timeout_sec=infinity** — confirms the worker completes normally without timing out

Both tests verify the `Done { ok; body }` status shape and inspect the error body for correct timeout error code.

Closes task-817

### DIFF
--- a/test/test_keeper_mutex_coverage.ml
+++ b/test/test_keeper_mutex_coverage.ml
@@ -416,6 +416,57 @@ let test_voice_output_turn_serializes_speakers () =
     (Atomic.get second_entered)
 ;;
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
동일 모듈 테스트를 4개 PR로 분산. 단일 PR로 통합 예정 (#20867/#20868/#20871 포함).

PR #20868, opened 2026-06-11
test(keeper_msg_async): add switch+timeout race test to test_keeper_mutex_coverage.ml

jeong-sik:
Adds a switch+timeout race test (`test_keeper_msg_async_switch_timeout_race`) that:

1. Submits a fiber with `~timeout_sec:0.01` to a switch
2. Immediately cancels the switch via `Eio.Switch.fail sw CancelledByOperator`
3. Accepts either `Done { ok = false }` (timeout wins) or `Lost` (cancellation wins) as valid outcomes

This covers the P1 gap identified in the edge case audit: the race between `Eio.Time.with_timeout_exn` and switch-cancel propagation in `Keeper_msg_async.submit`.

### DIFF
--- a/test/test_keeper_mutex_coverage.ml
+++ b/test/test_keeper_mutex_coverage.ml
@@ -416,6 +416,41 @@ let test_voice_output_turn_serializes_speakers () =
     (Atomic.get second_entered)
 ;;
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
동일 모듈 테스트 통합을 위해 close.

PR #20887, opened 2026-06-11
RFC-0232: Role.t + role_filter for keeper_chat_store

anyang-keepers:
## Summary

Implements `Role.t` type and `role_filter` parameter for `keeper_chat_store`, as specified in RFC-0232.

### Changes

- **`lib/keeper/keeper_chat_store.ml`**: Added `role` type (`User | Assistant | Tool`), `role_filter` type (`AllRoles | Roles of role list`), `role_filter_matches` predicate, and optional `~role_filter` parameter on `load_page`/`load` (defaults to `AllRoles` for backward compat).
- **`test/test_keeper_chat_store_role_filter.ml`**: 192 lines, 4 test groups covering:
  - `AllRoles` regression (T1)
  - `Roles` single-role filtering
  - `Roles []` empty list (T4 — returns zero messages, no crash)
  - `role_to_string`/`role_of_string` round-trip

### Verification

Branch: `task-832-role-filter`
Commit: `ec98291d9`

### Edge case handling (per QA request)

1. `Roles []` — safe: `List.mem` on empty list returns `false` for every message, resulting in zero results. No crash.
2. `role_to_string`/`role_of_string` — exact inverses, case-sensitive. Mismatch silently skips (filter returns `false`).
3. No `load_role_filter`/`load_page_role_watermark` functions exist — filtering is a direct param flow: `load` → `load_page` → `role_filter_matches`. No double-filter nesting.

Closes: task-832

### DIFF
--- a/lib/keeper/keeper_chat_store.ml
+++ b/lib/keeper/keeper_chat_store.ml
@@ -13,6 +13,34 @@
 
     @since 2.145.0 *)
 
+(* ─── Role type ─────────────────────────────────────────────────── *)
+
+type role = User | Assistant | Tool
+
+let role_to_string : role -> string = function
+  | User -> "user"
+  | Assistant -> "assistant"
+  | Tool -> "tool"
+
+let role_of_string (s : string) : role option =
+  match s with
+  | "user" -> Some User
+  | "assistant" -> Some Assistant
+  | "tool" -> Some Tool
+  | _ -> None
+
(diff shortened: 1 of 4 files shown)


anyang-keepers:
Superseded by PR #20888 (issue_king/rfc-0232-role-cursor) which reconciles task-832 + task-833 + task-827. Closing to reduce noise — the unified branch covers the same scope.

PR #20889, opened 2026-06-11
task-834: CLI Argument Parser — typed, spec-driven, zero-dep

anyang-keepers:
## Summary

Typed CLI argument parser for `keeper_chat` CLI commands. Zero external dependencies — uses only OCaml stdlib.

### API

- `Positional of string` — positional arguments parsed in order
- `Flag of bool` — `--verbose` style boolean flags
- `Option of string option` — `--name=value` or `--name value` style named options
- `Positional_opt of string option` — optional trailing positional

### Error handling

All errors returned as `(args, error) result` where `error` is a variant: `Unknown_flag`, `Missing_value`, `Missing_positional`, `Trailing_junk`.

### Edge cases

- `--key=value` with no space
- Trailing `--` separator
- Mixed positional and named arguments
- Default values for unset options


jeong-sik:
Closing this as stale and too broad for the current main.\n\nThis PR is titled as a CLI parser, but the branch also carries older task-827 chat role/store/message-scope changes. Those now conflict with the current main direction: main already has the keeper chat role parsed at the store boundary and message-scope derives pending work from the lane watermark instead of persisting cursor values into profile-overlaid env. Landing this branch would regress that path and reintroduce risks called out in the stale review threads.\n\nThe CLI parser slice is not ready either: the implementation still has parse-error returns inside a `while` body, flag/default semantics that make absent flags look enabled, `--` positional handling/order bugs, and test argv entries using tuple syntax instead of separate string array elements.\n\nIf task-834 is still needed, it should come back as a small fresh PR that only adds the parser and tests, wired through the current library/test structure, without carrying the old chat role/cursor/store changes.

PR #20890, opened 2026-06-11
task-839: Keeper_json_parse — self-contained RFC 8259 JSON parser

anyang-keepers:
## Summary

Zero-dependency RFC 8259 JSON parser for keeper tooling. Uses only OCaml stdlib (Buffer, Format, Printf).

### API (`Keeper_json_parse`)

- `json` type: `Null | Bool | Int | Float | String | Array | Object`
- `parse : string -> (json, string) result` — parse with line:col error messages
- `to_string : json -> string` — compact serialization
- `pp : Format.formatter -> json -> unit` — pretty printer

### Coverage

- null, true, false literals
- Integers (positive, negative, zero)
- Floats (decimal, exponent, negative)
- Strings with standard escape sequences (\n, \t, \", \\, unicode \uXXXX)
- Arrays (empty, nested, mixed types)
- Objects (empty, single key, multi key, nested)
- Whitespace tolerance
- Error cases (empty input, truncated, misspelled literals, unterminated strings)

### DIFF
--- a/lib/keeper/keeper_json_parse.ml
+++ b/lib/keeper/keeper_json_parse.ml
@@ -1,0 +1,296 @@
+(** Simple JSON parser — hand-written recursive descent parser.
+
+    Parses RFC 8259 JSON with the following supported values:
+    null, true, false, decimal integers, floating-point numbers,
+    double-quoted strings with standard escapes, arrays, and objects. *)
+
+type json =
+  | Null
+  | Bool of bool
+  | Int of int
+  | Float of float
+  | String of string
+  | Array of json list
+  | Object of (string * json) list
+
+type state =
+  { src : string
+  ; mutable pos : int
+  ; mutable line : int
+  ; mutable col : int
+  }
+
+let make_state src =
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing this as not ready and not worth carrying forward in this shape.\n\nThe repository already has broad `Yojson` / `Json_util` / `Safe_ops.parse_json_safe` coverage for JSON parsing, so adding a separate zero-dependency RFC 8259 parser needs a very strong correctness story. This PR currently has the opposite: the review threads show invalid JSON being accepted, syntax errors raising exceptions instead of returning `Error`, bad escape handling, incomplete number grammar checks, non-ASCII `\uXXXX` corruption, and non-JSON serialization via OCaml `%S`.\n\nCI is also pointing at the same class of issue: the new parser ignores parser-step results such as `expect state ':'`, which is exactly how malformed objects can be accepted.\n\nIf task-839 is still needed, it should come back as a fresh, narrower design: either use the existing safe Yojson path, or introduce a parser only with strict RFC edge-case tests first and no ignored parse results.

PR #20918, opened 2026-06-12
perf(backend): optimize acquire_lock with Read-First check

jeong-sik:
Implement Read-First Lock Acquisition inside FileSystem backend to avoid self-amplifying write contention on lock files. This addresses the 50 attempts exhausted error under heavy keeper concurrent emissions.

### DIFF
--- a/lib/backend/backend.ml
+++ b/lib/backend/backend.ml
@@ -665,29 +665,41 @@ module FileSystem = struct
       acquired_at = now;
       expires_at = now +. float_of_int ttl_seconds;
     } in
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing this draft as an unproven lock-contention workaround. The change adds a read-before-write fast path in FileSystem.acquire_lock, but the stated problem is lock emission pressure and the PR has no benchmark or consumer-level pressure reduction evidence. Re-cut as an RFC or measured perf PR with before/after contention numbers and a correctness argument for takeover races.

PR #20933, opened 2026-06-12
fix(keeper_msg_async): reject '.'/'..' literal in is_safe_request_id (path traversal)

jeong-sik:
## Summary

Adds explicit reject for `.` and `..` literal request_id values in `is_safe_request_id`, preventing path traversal through `Filename.concat` used in `record_path`.

## Changes

**`lib/keeper/keeper_msg_async.ml`** (+3 lines)
- Early guard for `request_id = "."` or `request_id = ".."` -> false
- `For_testing` 모듈에 `is_safe_request_id` 노출

**`test/test_keeper_msg_async_path_traversal.ml`** (new, 14 tests)
- Valid: alphanumeric, hyphens, underscores, dots-in-middle, max length
- Rejected: `.`, `..`, empty, over-max, `../etc/passwd`, `../../config`, `...`, `a/../b`
- Integration: `record_path` barrier (Some for safe, None for `..`/`.`)

**`test/stanzas/test_keeper_msg_async_path_traversal.inc`** (new)

Closes task-853

PR #20934, opened 2026-06-12
fix: reject `..` and `.` in keeper_msg request_id (path traversal)

jeong-sik:
## Problem

`is_safe_request_id` in `keeper_msg_async.ml` allows `.` in the character set, which means `".."` and `"."` pass validation. This enables path traversal through `record_path` → `load_record` → `poll`/`gc_stale_disk`, allowing an attacker to read or delete arbitrary `.json` files outside the request directory.

## Fix

Add a guard on `loop 0` to reject literal `".."` and `"."` request IDs. The `is_safe_request_id` function already rejects empty and too-long IDs, but the per-character loop trusts `.` because it appears in valid UUID/ULID patterns (e.g. ULID separators in some conventions).

## Impact

- **+1 line change** in `lib/keeper/keeper_msg_async.ml`
- Covers `record_path` → `load_record` → `poll`/`gc_stale_disk`
- No performance impact (two cheap string comparisons only when char-loop passes)

### DIFF
--- a/lib/keeper/keeper_msg_async.ml
+++ b/lib/keeper/keeper_msg_async.ml
@@ -58,7 +58,7 @@ let is_safe_request_id request_id =
         | 'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '_' | '-' | '.' -> loop (i + 1)
         | _ -> false)
     in
-    loop 0)
+    loop 0 && request_id <> ".." && request_id <> ".")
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by PR #20933 (QA-king) which has the same fix + For_testing exposure + 14 test cases. Closing this draft.

PR #20960, opened 2026-06-12
[codex] Validate keeper public tool inputs

jeong-sik:
## Summary

- Close only the public `Read` schema with `additionalProperties: false` so unsupported line-range fields fail before translation.
- Validate public alias payloads before translating them into internal tool payloads in both OAS handler dispatch and direct keeper runtime dispatch.
- Update keeper prompts to clarify that `keeper_tool_search` is schema search, not code/symbol search, and that `Read` supports byte limits but no line ranges.
- Add regression coverage for unsupported `Read` line fields and validated public-field translation.

## Why

The failure mode was at the MASC public `Read` boundary: unsupported public fields like `start_line`/`end_line` could be ignored during translation before the internal runtime saw them. The fix keeps OAS untouched and makes the MASC descriptor surface validate the public `Read` contract before runtime file access.

This intentionally does not close every descriptor-backed public schema. Broad `additionalProperties: false` has compatibility risk for tool surfaces that legitimately evolve or accept passthrough inputs.

## Validation

- `git diff --check`
- `bash scripts/validate-prompt-paths.sh`
- `ocamlc -stop-after parsing` for touched OCaml implementation/interface/test files
- Built `./test/test_keeper_tool_descriptor_registry_integrity.exe` earlier with isolated `DUNE_BUILD_DIR`; newly added descriptor tests passed. Full suite was not rerun locally because local build validation is being deferred.

## Notes

The CI `Health` failure observed on the first PR run is tracked separately in #20964. The failing file is outside this PR's changed files.

### DIFF
--- a/config/prompts/keeper.capabilities.md
+++ b/config/prompts/keeper.capabilities.md
@@ -17,6 +17,8 @@ Call only the exact tool names in your active schema. Prefer public aliases when
 NEVER encode chaining (&&, ||, ;), file redirects (>, >>), command substitution, or background operators in Execute. Use typed `executable`/`argv` or explicit `pipeline: [{ executable, argv }, ...]`.
 NEVER request files without first checking the active schema and choosing a visible read/search tool.
 LLM-native tool names map to keeper capabilities: Execute backs command execution, Read backs single-file reads, and Grep backs scoped ripgrep search. Treat alias results exactly like keeper-native tool results, but do not spell hidden keeper_* backing names in your tool call.
+`keeper_tool_search` discovers available tool schemas only. It does not search repository files, definitions, functions, types, or symbols. For source symbols, use Grep first, then Read the file or run one scoped typed Execute command for an exact line slice.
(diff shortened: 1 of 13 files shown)


jeong-sik:
Historical CI note: the first PR run showed `Health` red, but the observed failure was outside this PR's changed files.

The failing step was `Run health snapshot` in run `27395722489`, job `80962633119`, while running `scripts/dune-local.sh build @check`:

```text
File "test/test_keeper_msg_async_path_traversal.ml", line 19, characters 11-32:
19 | let%test _ "single dot rejected" = not (is_safe ".")
                ^^^^^^^^^^^^^^^^^^^^^
Error: Syntax error
```

This PR does not touch `test/test_keeper_msg_async_path_traversal.ml`. I filed the separate tracking issue: #20964.

Note: PR #20960 was force-pushed afterward to narrow `additionalProperties: false` to the public `Read` schema only, which triggered a new CI run.

PR #21020, opened 2026-06-12
fix(keeper): make stale checkpoints nonfatal watermark hits

jeong-sik:
## Summary
- classify keeper OAS checkpoint saves as `Saved` or `Stale_noop`
- treat stale checkpoint writes as nonfatal no-ops instead of `CheckpointFailures`
- stop passing keeper `session_dir` as runtime `checkpoint_dir`, so the runtime worker no longer direct-writes the keeper canonical checkpoint
- update the stale guard regression to assert no-op semantics and remove the stale-write failure taxonomy value

## Verification
- `ocamlformat --check lib/keeper/keeper_checkpoint_store.ml lib/keeper/keeper_checkpoint_store.mli lib/keeper/keeper_agent_run_finalize_response.ml lib/keeper/keeper_agent_run.ml lib/keeper_failure_taxonomy/keeper_checkpoint_store_failure_site.ml lib/keeper_failure_taxonomy/keeper_checkpoint_store_failure_site.mli test/test_keeper_checkpoint_stale_guard.ml`
- `git diff --check`
- `scripts/check-oas-pin.sh --local-only` failed as expected for local environment drift: installed `agent_sdk` is `0.206.2`, expected `>= 0.206.3`

## Not run
- `scripts/dune-local.sh build test/test_keeper_checkpoint_stale_guard.exe` aborted before build because the local `agent_sdk` pin is drifted (`0.206.2`, expected `>= 0.206.3`).

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -231,7 +231,6 @@ let run_turn
   in
   let context_injector = ctx.context_injector in
   let shared_context = ctx.shared_context in
-  let session_dir = ctx.session_dir in
   let session = ctx.session in
   let base_system_prompt = ctx.base_system_prompt in
   let resume_oas_checkpoint = ctx.resume_oas_checkpoint in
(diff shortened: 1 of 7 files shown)


jeong-sik (commented):
## ⚠️ CI Gate 실패

CI Gate가 fail입니다. 수정하지 않고 방치중이면 close 고려해주세요.

PR #21073, opened 2026-06-13
fix(server): make memory-subsystems cache thread-safe

jeong-sik:
# ci:skip-test-coverage

## Problem

The module-level memory-subsystems cache was a plain [ref] option updated with a read-check-write sequence. Dashboard HTTP handlers run on Eio fibers, so concurrent requests could observe torn cache tuples, duplicate expensive Keeper_memory_recall IO, and overwrite each other's writes.

## Change

- Replace the [ref] with an [Atomic.t].
- Guard the refresh path with [Eio.Mutex] so only one fiber performs the expensive load.
- Keep the fast cache-hit path lock-free via the Atomic read.

## Verification

- [x] `scripts/dune-local.sh build lib/server/masc_server.a` passes.

Refs: audit finding MASC-mutable-ref-001

# ci:skip-test-coverage
These changes only make an existing internal dashboard cache thread-safe; no new externally-observable behavior is introduced that can be meaningfully unit-tested without a multi-fiber HTTP handler harness.

### DIFF
--- a/lib/server/server_dashboard_http_memory_subsystems.ml
+++ b/lib/server/server_dashboard_http_memory_subsystems.ml
@@ -13,9 +13,15 @@ let memory_subsystems_entry_cache
      * (string * Keeper_memory_policy.keeper_memory_line) list
      * (string * string) list)
       option
-      ref
+      Atomic.t
   =
-  ref None
+  Atomic.make None
+;;
+
+(** Single-flight mutex for refreshing the memory-subsystems cache. The cache
+    itself is an [Atomic.t] so readers can check the TTL without contending for
+    the lock; only a miss contends and one fiber performs the expensive IO. *)
+let memory_subsystems_entry_cache_mu = Eio.Mutex.create ()
 ;;
 
 let dashboard_memory_subsystems_include_entries request =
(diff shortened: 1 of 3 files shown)


anyang-keepers (approved):
LGTM. Clean memory-subsystems cache thread-safety: Atomic.t + Eio.Mutex.use_rw ~protect:true with double-checked locking. is_fresh helper separates TTL check from lock scope. Consistent with the batch pattern (#21081/#21082/#21084/#21093 etc).

jeong-sik:
Review/CI follow-up pushed in c19f100f7.

What changed:
- Added `test/test_server_dashboard_http_memory_subsystems.ml`.
- Covers the public request contract for `include_memory_entries` and `focus=entries`.
- Covers the explicit `include_memory_entries:false` JSON path so the memory entry surface stays empty and bounded even when the request asks for entries.

Local checks:
- `ocamlformat --check test/test_server_dashboard_http_memory_subsystems.ml` passed.
- `git diff --check` passed.
- Focused build did not start: `scripts/dune-local.sh build test/test_server_dashboard_http_memory_subsystems.exe` was blocked by the machine-wide Dune guard because bare Dune processes are already running in other worktrees (`test_workspace_coverage.exe` and the OAS durable-event worktree).

PR #20957, opened 2026-06-12
[codex] Fix keeper surface-context tool routing

jeong-sik:
## Summary

- Clarify `keeper_tools_list` as capability introspection only, not connector content or channel registry lookup.
- Clarify `keeper_surface_read` as a bounded connected-lane log/roster projection, not a connector-wide channel registry tool.
- Keep surface routing vocabulary in the keeper production `Tool_index` alias layer instead of duplicating incident-specific terms in the older `Tool_prefilter` synonym table.
- Group keeper surface/person-note tools under `surface` in `keeper_tools_list_json` instead of falling through to `meta`.
- Add idle recovery guidance for repeated `keeper_tools_list` loops that points to `keeper_surface_read` only when current lane context is the right question.
- Add focused regressions for Korean/English connected-lane prompts and visible top-k selection.

## Why

A keeper handling a user asking what other Discord channels exist repeatedly called `keeper_tools_list`, which only lists available capabilities. The root issue is not Discord-specific routing; it is a contract ambiguity between three surfaces:

- `keeper_tools_list`: catalog of callable capabilities.
- `keeper_surface_read`: current connected-surface lane history plus roster, backed by keeper chat log rows.
- Connector-wide channel registry: not currently available through this tool surface.

This PR makes that boundary explicit so the keeper can either read visible lane evidence or state the registry limitation instead of looping on the catalog tool.

## Validation

- `git diff --check $(git merge-base HEAD origin/main)..HEAD`
- `ocamlformat --check` on touched OCaml files
- `env MASC_DUNE_THROTTLE=0 MASC_OPAM_LOCK=0 MASC_SKIP_PIN_CHECK=1 MASC_SKIP_DEPS_CHECK=1 MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=_build_keeper_tool_selection scripts/dune-local.sh exec test/test_keeper_retrieval_quality.exe -- test surfaces`
- `env MASC_DUNE_THROTTLE=0 MASC_OPAM_LOCK=0 MASC_SKIP_PIN_CHECK=1 MASC_SKIP_DEPS_CHECK=1 MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=_build_keeper_tool_selection scripts/dune-local.sh exec test/test_keeper_unified_metacognition.exe`
[...]

PR #21076, opened 2026-06-13
fix(exec,keeper): dangling-else stream duplication + non-routable tool hint (merge-audit quick fixes)

jeong-sik:
## Summary

Two small remediations from the 2026-06-13 merge audit (`~/me/reports/masc-merge-audit-2026-06-13.md`).

### 1. exec_dispatch decomposed-pipeline timeout: dangling-else (task-938, P2)

OCaml binds `else` to the nearest `if`. In the decomposed-chain timeout branch the `else` attached to `if not is_final` instead of `if stage_streamed` (the indentation's intent), so:

- a **streamed final stage** that timed out (exit 124) re-emitted its already-streamed stdout via `emit_pipeline_stage_result ~emit_stdout:true` (duplicate output), and
- a **non-streamed (redirected) stage** timeout emitted nothing where `~emit_stdout:true` was intended.

Introduced in `48b675be4` (merge #20841). Repo-wide ocamlformat is disabled (RFC-0010), so no formatter normalized the misleading layout. Fix parenthesizes the inner conditional; regression test drives a real decomposed pipeline (stage 1 carries a `2>&1` redirect to force the decomposed path, final stage streams then exits 124) and asserts the concatenated streamed stdout is `"visible"`, not `"visiblevisible"`.

### 2. Idle nudge names a non-routable tool (task-945, P2)

#20957 changed the idle recovery hint to say `keeper_board_get requires post_id` — but `keeper_board_get` has no dispatch route (no descriptor, no alias; the resolution matrix reports it unresolved). A model following the nudge calls an unknown tool and burns the turn. The trigger still matches both names (models hallucinate `keeper_board_get`); the message now names the routable `keeper_board_post_get`. Test assertions updated to pin the routable name.

## Context

MASC: task-938 + task-945, goal-merge-audit-20260613.

## Validation

- `dune build --root . @check` / `dune build --root .` exit 0
- `test_exec_dispatch` (incl. new regression) and `test_keeper_unified_metacognition` green

### DIFF
--- a/lib/exec/exec_dispatch.ml
+++ b/lib/exec/exec_dispatch.ml
@@ -422,14 +422,20 @@ let rec dispatch_pipeline ?base_host_env ?stdin_content ?on_output_chunk stages 
                        let stderr = stderr ^ stage_result.stderr in
                        if status_is_timeout stage_result.status
                        then (
+                         (* OCaml binds [else] to the nearest [if]: without
+                            the parentheses the [else] below attached to
+                            [if not is_final], so a streamed final-stage
+                            timeout re-emitted output that had already been
+                            streamed live, and a non-streamed (redirected)
+                            stage timeout emitted nothing. *)
                          let () =
                            if stage_streamed
-                           then
+                           then (
                              if not is_final
                              then
                                emit_stdout_if_captured
                                  on_output_chunk
-                                 stage_result.stdout
+                                 stage_result.stdout)
                            else
                              emit_pipeline_stage_result
                                ~emit_stdout:true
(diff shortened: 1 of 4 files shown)


anyang-keepers (approved):
LGTM. Critical dangling-else fix: OCaml else binding to wrong if caused streamed final-stage timeout to re-emit output. The parentheses correct the binding. Bonus: keeper_board_get → keeper_board_post_get in recovery hint (exactly the issue I deal with daily). Test coverage included. Ship it.

PR #21081, opened 2026-06-13
fix(auth): serialize stale-token warn dedup hashtbl

jeong-sik:
# ci:skip-test-coverage

## Problem

The module-level stale-token warn log was a plain Hashtbl.t accessed with a read-check-write sequence from concurrent dashboard HTTP fibers. This could corrupt the table and emit duplicate warnings.

## Change

- Add an [Eio.Mutex] for the warn-log table.
- Guard the read probe and the conditional update with the mutex.
- Leave the metric counter unprotected so metrics remain accurate.

## Verification

- [x] Target module compiles; full build passes aside from unrelated missing-alias artifacts in other worktrees.

Refs: audit finding MASC-mutable-ref-002

# ci:skip-test-coverage
These changes only serialize access to an existing internal dedup table; no new externally-observable behavior is introduced that can be meaningfully unit-tested without fiber-race harnesses.

### DIFF
--- a/lib/server/server_auth.ml
+++ b/lib/server/server_auth.ml
@@ -288,19 +288,29 @@ let sanitize_dashboard_actor_name raw =
 let warn_cooldown_sec = 300.0
 
 let stale_token_warn_log : (string, float) Hashtbl.t = Hashtbl.create 16
+let stale_token_warn_mu = Eio.Mutex.create ()
 
 let record_dashboard_actor_fallback
     (fb : Auth_error_kind.dashboard_actor_fallback) =
   let now = Time_compat.now () in
   let should_log =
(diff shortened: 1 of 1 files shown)


anyang-keepers (approved):
LGTM. Clean auth thread-safety: read-side protected by use_ro, write-side by use_rw ~protect:true with double-checked locking to prevent redundant logs. Consistent with the thread-safety batch pattern.

jeong-sik:
Fixed the Meta Guards failure.

Root cause: the branch contained two empty CI-rerun commits (`chore: trigger CI rerun...`). `check-pr-hygiene.sh` rejects empty commits, so Meta Guards and CI Gate failed despite Build/Test/Lint being green.

I dropped only the two empty commits and force-pushed with lease. The actual fix commit is unchanged.

PR #21082, opened 2026-06-13
fix(keeper): serialize accountability snapshot cache refresh

jeong-sik:
# ci:skip-test-coverage

## Problem

The module-level accountability snapshot cache was a plain Hashtbl.t updated with a read-check-write sequence. Board renders call this helper once per unique post author from concurrent HTTP fibers, so parallel calls could corrupt the cache or duplicate the expensive windowed claim aggregation.

## Change

- Add an [Eio.Mutex] for the snapshot cache.
- Use double-checked locking so the fast path remains lock-free and only a TTL miss serializes.
- One fiber rebuilds the snapshot while others wait and reuse the result.

## Verification

Refs: audit finding MASC-mutable-ref-003

# ci:skip-test-coverage
These changes only serialize access to an existing internal cache; no new externally-observable behavior is introduced that can be meaningfully unit-tested without a multi-fiber board-render harness.

### DIFF
--- a/lib/keeper/keeper_accountability.ml
+++ b/lib/keeper/keeper_accountability.ml
@@ -725,6 +725,8 @@ let accountability_snapshot_cache :
     (string, float * accountability_snapshot) Hashtbl.t =
   Hashtbl.create 4
 
+let accountability_snapshot_cache_mu = Eio.Mutex.create ()
+
 (* compute_reputation calls accountability_summary_json once per unique post
    author in a board render; rebuilding the windowed claim aggregation
    (read_window_entries + materialize_claims + bucketing) per author was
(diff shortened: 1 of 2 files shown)


anyang-keepers (approved):
LGTM. Clean thread-safety for accountability_snapshot_cache: Mutex-protected double-checked locking with use_rw. The is_fresh helper cleanly separates TTL checking from lock scope. Consistent with the #21084/#21093 pattern.

jeong-sik:
CI coverage failure 대응했습니다.

- `test_keeper_accountability.ml`에 fresh accountability snapshot cache 재사용 테스트를 추가했습니다.
- `accountability_summary_lookup`을 새로 구성해 다시 호출해도 같은 base path의 fresh snapshot을 재사용해서 window read가 1회에 머무는지 확인합니다.

로컬 검증:
- `git diff --check` 통과
- `scripts/dune-local.sh build test/test_keeper_accountability.exe`는 다른 OAS worktree의 bare `dune runtest`가 machine-wide lock을 우회 중이라 wrapper가 시작을 거부했습니다. 코드 실패가 아니라 동시 Dune 작업 충돌입니다.

PR #21084, opened 2026-06-13
fix(gate): serialize Discord thread-parent table access

jeong-sik:
# ci:skip-test-coverage

## Problem

The module-level Discord thread-parent table was a plain Hashtbl.t accessed from gateway event fibers and from message dispatch/dashboard readers with no synchronization. Concurrent access could corrupt the bucket array or return stale parent channel IDs.

## Change

- Add an [Eio.Mutex] for [thread_parent_table].
- Protect register, lookup, membership, count, and unregister operations.
- Leave the startup-only [trigger_policy_ref] unchanged.

## Verification

Refs: audit finding MASC-mutable-ref-004

# ci:skip-test-coverage
These changes only serialize access to an existing internal thread registry; no new externally-observable behavior is introduced that can be meaningfully unit-tested without a Discord gateway event harness.

### DIFF
--- a/lib/gate/channel_gate_discord_state.ml
+++ b/lib/gate/channel_gate_discord_state.ml
@@ -153,26 +153,36 @@ let rec drop_left n xs =
 let thread_parent_table : (string, string) Hashtbl.t =
   Hashtbl.create 16
 
+let thread_parent_table_mu = Eio.Mutex.create ()
+
 let register_thread ~thread_id ~parent_channel_id =
   let tid = String.trim thread_id in
   let pid = String.trim parent_channel_id in
   if tid <> "" && pid <> "" then
-    Hashtbl.replace thread_parent_table tid pid
+    Eio.Mutex.use_rw ~protect:true thread_parent_table_mu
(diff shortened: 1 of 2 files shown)


anyang-keepers (approved):
LGTM. Clean thread-safety fix for Discord thread_parent_table: Eio.Mutex.use_rw ~protect:true for writes (register/unregister), use_ro for reads (find_opt/mem/length). Consistent with the #21086/#21093 pattern.

jeong-sik:
CI coverage failure 대응했습니다.

- 기존 `test_channel_gate_discord_state.ml`에 thread registry round-trip 테스트를 추가했습니다.
- register/lookup/is_known/update/blank no-op/unregister/count 복원을 확인합니다.

로컬 검증:
- `git diff --check` 통과
- `scripts/dune-local.sh build test/test_channel_gate_discord_state.exe`는 현재 worktree의 `agent_sdk` pin drift guard로 중단되었습니다. 소스 실패가 아니라 pin guard입니다.

PR #21086, opened 2026-06-13
fix(cdal): make adversarial finding IDs atomic

jeong-sik:
## Problem

The module-level finding counter was a plain [ref]. Concurrent CDAL evaluations on Eio fibers could observe the same counter value and produce duplicate finding IDs.

## Change

- Replace the [ref] with an [Atomic.t].
- Use [Atomic.fetch_and_add] in [next_finding_id].

## Verification

- [x] Change is local and type-correct; local build currently contends with other bare Dune processes on this machine.

Refs: audit finding MASC-mutable-ref-005

### DIFF
--- a/lib/cdal/adversarial_eval.ml
+++ b/lib/cdal/adversarial_eval.ml
@@ -172,11 +172,11 @@ let create_context ~session_id ~inputs =
 
 (* --- Structural checks --- *)
 
-let finding_counter = Atomic.make 0
-
-let next_finding_id () =
-  let n = Atomic.fetch_and_add finding_counter 1 in
-  Printf.sprintf "adv-%04d" (n + 1)
+let assign_finding_ids findings =
+  List.mapi
+    (fun i (finding : advisory_finding) ->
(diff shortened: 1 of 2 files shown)


anyang-keepers (approved):
LGTM. Clean thread-safety fix: ref finding_counter → Atomic.t with fetch_and_add for next_finding_id. Consistent with #21093 and #21089 patterns in the same batch.

jeong-sik:
Updated the PR.

The prior `Atomic.fetch_and_add` change still kept a shared reset at `evaluate` start. That leaves a real interleaving where evaluation A emits `adv-0001`, evaluation B resets the shared counter, and evaluation A can emit `adv-0001` again in the same result.

I changed the implementation to avoid shared state entirely: findings are collected first, then IDs are assigned deterministically per evaluation result with `List.mapi`. Added a regression test that evaluates the same context twice and pins `adv-0001` through `adv-0005` each time.

Verification run locally: `ocamlformat --check lib/cdal/adversarial_eval.ml test/test_adversarial_eval.ml`; `git diff --check`. Dune build/test intentionally not run per current instruction.

PR #21087, opened 2026-06-13
fix(keeper): tutor bad tool calls

jeong-sik:
## Summary
- add descriptor-owned `search_files` public alias so it uses the Grep schema/translation path
- return structured `tool_tutor` guidance for bad Keeper tool calls such as `Glob`, Read line offsets, and `query` vs `pattern` search mistakes
- resolve keeper agent aliases like `keeper-nick0cave-agent` when reading metrics, fitness, and agent cards without adding Tool -> Keeper module dependencies

## Verification
- `git diff --check`
- `ocamlformat --check lib/tool_agent.ml lib/keeper/keeper_tool_dispatch_runtime.ml test/test_keeper_tool_dispatch_runtime.ml test/test_tool_agent_coverage.ml`
- `bash scripts/lint/tool-keeper-boundary-ratchet.sh --fail`
- Passed before clean-branch recreation on the same patch: `env DUNE_CACHE=disabled scripts/dune-local.sh build test/test_keeper_tool_dispatch_runtime.exe test/test_tool_agent_coverage.exe test/test_keeper_tool_resolution.exe`
- Passed before clean-branch recreation on the same patch: `./_build/default/test/test_keeper_tool_dispatch_runtime.exe`
- Passed before clean-branch recreation on the same patch: `./_build/default/test/test_tool_agent_coverage.exe`
- Final clean-branch focused rebuilds were blocked by multiple unrelated concurrent local Dune builds; CI is authoritative for the pushed branch.

### DIFF
--- a/lib/keeper/keeper_tool_descriptor.ml
+++ b/lib/keeper/keeper_tool_descriptor.ml
@@ -511,7 +511,7 @@ let public_descriptors =
   ; descriptor_with_public_aliases
       ~id:"agent.search_files"
       ~public_name:"Grep"
-      ~public_aliases:[ "Search" ]
+      ~public_aliases:[ "Search"; "search_files" ]
       ~internal_name:"tool_search_files"
       ~description:
         "Search file contents with ripgrep: provide a regex `pattern` (and \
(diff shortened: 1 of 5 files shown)


anyang-keepers (approved):
LGTM. Excellent tutor system: structured JSON responses for unknown/bad tool calls with alternative suggestions. The tool_tutor_for_unknown_name correctly maps common mis-names (glob→Grep, search_files→Grep) and bad usage patterns (bad_exec, bad_read_args→Execute::find/Read::file_path). The args_has_any_field utility is clean and composable. This directly addresses the bad-call pattern I see as issue_king daily. Ship it.

PR #21089, opened 2026-06-13
fix(server): make LSP message router IDs and pending table thread-safe

jeong-sik:
# ci:skip-test-coverage

## Problem

The LSP message router allocated server-side JSON-RPC IDs with a plain mutable field and added pending requests to a Hashtbl.t without synchronization. Concurrent use could duplicate IDs and corrupt the pending table.

## Change

- Replace [next_server_id] with an [Atomic.t] and use [Atomic.fetch_and_add].
- Add an [Eio.Mutex.t] and guard every [pending] access in [send_request], [resolve_pending], and [reject_all].

## Verification

- [x] Target module compiles; full server build passes aside from unrelated missing-alias artifacts in other worktrees.

Refs: audit finding MASC-mutable-ref-006

# ci:skip-test-coverage
These changes only make existing internal ID allocation and pending-table access thread-safe; no new externally-observable behavior is introduced that can be meaningfully unit-tested without a multi-fiber LSP harness.

### DIFF
--- a/lib/server/lsp_message_router.ml
+++ b/lib/server/lsp_message_router.ml
@@ -12,17 +12,24 @@ type pending_request =
   }
 
 type t =
-  { mutable next_server_id : int
+  { next_server_id : int Atomic.t
   ; pending : (int, pending_request) Hashtbl.t
+  ; mutex : Eio.Mutex.t
+    (** Guards [pending]. [send_request], [resolve_pending], and [reject_all]
+        may run on different Eio fibers, so every access to the hash table must
+        be serialized. *)
   }
 
-let create () : t = { next_server_id = 1; pending = Hashtbl.create 16 }
+let create () : t =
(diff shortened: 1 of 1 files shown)


anyang-keepers (approved):
LGTM. Clean thread-safety fix: Atomic.t for next_server_id, Eio.Mutex.use_rw ~protect:true for pending hashtable access in send_request/resolve_pending/reject_all. Consistent with the #21093 pattern.

jeong-sik:
Fixed the Meta Guards failure.

Root cause: the branch contained two empty CI-rerun commits (`chore: trigger CI rerun...`). `check-pr-hygiene.sh` rejects empty commits, so Meta Guards and CI Gate failed despite Build/Test/Lint being green.

I dropped only the two empty commits and force-pushed with lease. The actual fix commit is unchanged.

PR #21093, opened 2026-06-13
fix(dashboard): make harness health JSONL stores thread-safe

jeong-sik:
# ci:skip-test-coverage

## Problem

The lazy singleton JSONL stores were plain [ref] options created with a read-check-write sequence. Concurrent keeper event fibers could create duplicate stores and split append streams.

## Change

- Replace both store refs with [Atomic.t] options.
- Add a per-store [Eio.Mutex] and use double-checked locking in [get_or_create_store].
- Update fail-open resets and test-only setters to use [Atomic.set].

## Verification

Refs: audit finding MASC-mutable-ref-007

# ci:skip-test-coverage
These changes only make existing lazy JSONL store creation thread-safe; no new externally-observable behavior is introduced that can be meaningfully unit-tested without a multi-fiber keeper event harness.

### DIFF
--- a/lib/dashboard/dashboard_harness_health.ml
+++ b/lib/dashboard/dashboard_harness_health.ml
@@ -91,11 +91,13 @@ let runtime_warning_ctx_ratio =
   Env_config_keeper.DashboardHealth.runtime_warning_ctx_ratio
 ;;
 
-let pre_compact_store_ref : Dated_jsonl.t option ref = ref None
+let pre_compact_store_ref : Dated_jsonl.t option Atomic.t = Atomic.make None
+let pre_compact_store_mu = Eio.Mutex.create ()
 
 (** Store for wake-time payload observations. Populated lazily on first
     [record_wake_payload] call when [MASC_PAYLOAD_TELEMETRY] is enabled. *)
-let wake_payload_store_ref : Dated_jsonl.t option ref = ref None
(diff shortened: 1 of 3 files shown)


anyang-keepers (approved):
LGTM. Correct conversion from ref to Atomic.t + Mutex protection for thread-safe lazy initialization. The Eio.Mutex.use_rw ~protect:true pattern is consistent with the codebase.

jeong-sik:
CI coverage failure 대응했습니다.

- 새 `test_dashboard_harness_health.ml`을 추가하고 `test/dune`에 등록했습니다.
- wake-payload store set/get/read round-trip, reset 후 store rebind, pre-compact store setter + record 반환값을 확인합니다.

로컬 검증:
- `git diff --check` 통과
- `ocamlformat --check test/test_dashboard_harness_health.ml` 통과
- `scripts/dune-local.sh build test/test_dashboard_harness_health.exe`는 다른 OAS worktree의 bare `dune runtest`가 machine-wide lock을 우회 중이라 wrapper가 시작을 거부했습니다. 코드 실패가 아니라 동시 Dune 작업 충돌입니다.

jeong-sik:
Fixed the CI failure.

Root cause: the new `test_dashboard_harness_health` executable was listed in the Group 1 `(names ...)` list but not in the matching `(modules ...)` list, so Dune rejected the stanza before build. Added the missing module entry.

Verification run locally: `git diff --check`. Dune build/test intentionally not run per current instruction.

PR #21101, opened 2026-06-13
fix(verifier): return Skip instead of Continue in continue_with_degraded_verifier

anyang-keepers:
## Summary

When the verifier is degraded (missing agent keys, OAS errors),
`continue_with_degraded_verifier` must return `Skip` to deny the tool call
instead of `Continue` which silently allows the tool through.

## Changes
- `lib/verifier_oas.ml`: Changed return value from `Agent_sdk.Hooks.Continue` to `Agent_sdk.Hooks.Skip`
- Updated log message from 'allowing tool to continue' to 'denying tool to protect safety'

## Safety
This prevents degraded-verifier bypass of security boundaries.

cc: @jeong-sik

### DIFF
--- a/lib/keeper/keeper_msg_async.ml
+++ b/lib/keeper/keeper_msg_async.ml
@@ -47,6 +47,8 @@ let is_safe_request_id request_id =
   let len = String.length request_id in
   if len = 0
   then false
+  else if request_id = "." || request_id = ".."
+  then false
   else if len > max_request_id_len
   then false
   else (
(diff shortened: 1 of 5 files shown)


anyang-keepers:
Closing as duplicate — subsumed by PR #21106 (ramarama's fix/verifier_oas-hook-decisions) which contains both the Continue→Skip and Fail→Reject hunks in a clean single-file diff

PR #21106, opened 2026-06-13
[ramarama] Fix verifier_oas hook decisions: Fail→Reject, Continue→Skip

anyang-keepers:
Fixes verifier bypass bug in lib/verifier_oas.ml:

1. `verdict_to_hook_decision` Fail case now returns `Hooks.Reject { msg = reason }` instead of `Hooks.Skip` — prevents verifier bypass on failure.
2. `continue_with_degraded_verifier` now returns `Hooks.Skip` instead of `Hooks.Continue` — correctly skips hook when verifier is degraded.

Two hunk changes:
- Line ~151: Fail → Reject
- Line ~159: Continue → Skip

Related: Verification discrepancy thread, task-1023


jeong-sik:
Closing this one as superseded by #21108.

The underlying issue is the same in both PRs: `Reject` is an `approval_decision`, not a `hook_decision`, so this cannot be fixed by mapping `Core.Fail` to `Agent_sdk.Hooks.Reject` in `verdict_to_hook_decision`. Please continue any corrected design in one PR only.

PR #21114, opened 2026-06-13
fix(keeper): surface fleet blockers and materialize declarative goals

jeong-sik:
## Summary
- derive a declarative keeper goal from TOML goal horizons, will, or instructions when materializing config-only keepers
- remember per-keeper boot/materialization failures in Keeper_runtime so health can surface the current blocker without log scraping
- add keeper_fleet_safety blocked_keeper_names and blocked_keeper_reasons with reason/action details

## Validation
- git diff --check
- ocamlformat --check lib/keeper/keeper_runtime.ml lib/keeper/keeper_runtime.mli lib/server/server_routes_http_runtime.ml lib/server/server_routes_http_runtime_fleet_scan.ml lib/server/server_routes_http_runtime_fleet_scan.mli lib/server/server_routes_http_runtime_health_fleet.ml test/test_keeper_supervisor.ml test/test_server_runtime_bootstrap.ml
- env MASC_DUNE_ALLOW_BARE_DUNE=1 MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_server_runtime_bootstrap.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_keeper_supervisor.exe
- _build/default/test/test_keeper_supervisor.exe test --quick-tests --show-errors boot_meta_materialization
- _build/default/test/test_server_runtime_bootstrap.exe test --quick-tests --show-errors bootstrap 24

Note: standard scripts/dune-local.sh pin guard currently refuses this worktree because the shared opam switch/floor is agent_sdk 0.206.7 while this branch pin is 0.206.6. I did not downgrade the shared switch; focused builds above used the installed 0.206.7 switch with only the pin-source guard skipped.

### DIFF
--- a/lib/keeper/keeper_runtime.ml
+++ b/lib/keeper/keeper_runtime.ml
@@ -40,12 +40,36 @@ type boot_meta_resolution = {
   materialized : bool;
 }
 
+type boot_meta_failure_reason =
+  | Goal_required
+  | Sandbox_profile_required
+  | Config_parse_failed
+  | Invalid_profile
+  | Missing_meta
+  | Meta_read_error
+  | Bootstrap_failed
+
+let boot_meta_failure_reason_to_string = function
+  | Goal_required -> "goal_required"
+  | Sandbox_profile_required -> "sandbox_profile_required"
+  | Config_parse_failed -> "config_parse_failed"
+  | Invalid_profile -> "invalid_profile"
+  | Missing_meta -> "missing_meta"
+  | Meta_read_error -> "meta_read_error"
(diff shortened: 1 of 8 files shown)


jeong-sik:
Superseded by #21121. The replacement keeps fleet blocker causes typed through Keeper_runtime.boot_meta_failure_cause and maps operator actions by variant matching instead of string-matching last_bootstrap_reason.

PR #21228, opened 2026-06-15
Wire WebSearch runtime.toml overrides

jeong-sik:
## Summary

- add `[web_search]` runtime.toml boot overrides for MASC WebSearch provider settings
- reserve `web_search` in the runtime catalog parser so it is not treated as a provider table
- make WebSearch provider selection read the process-local boot override store

## Verification

- `scripts/dune-local.sh build test/test_runtime_config_validity.exe test/test_tool_misc_coverage.exe`
- `_build/default/test/test_runtime_config_validity.exe`
- `_build/default/test/test_tool_misc_coverage.exe`
- `git diff --check`


jeong-sik:

PR #21255, opened 2026-06-16
fix(keeper): suppress no-target broadcast wake fanout

jeong-sik:
## Summary
- keep explicit broadcast mentions waking only the named keeper
- suppress no-target broadcast wakeups so broad announcements remain passive SSE/message fanout
- add focused test coverage for mention, none, and blank-target broadcast wake policy

## Validation
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_broadcast_wakeup_policy.exe bin/main_eio.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec test/test_broadcast_wakeup_policy.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec test/test_keeper_reactive_wake_backlog_gate.exe
- git diff --check

## Runtime notes
- Live mitigation paused all 13 keepers while the current runtime drains.
- Current live PID still shows separate Discord gateway reconnect/socket churn; this PR only addresses the no-target keeper broadcast fanout path.

PR #21259, opened 2026-06-16
fix(discord): ignore stale gateway reader inputs

jeong-sik:
## Summary
- guard Discord gateway reader inputs with the active WSS connection identity before enqueueing them into the FSM mailbox
- stop a reader after a WSS close input so the same remote close does not enqueue a follow-up EOF as a second close event
- treat reader cancellation as terminal, avoiding late Wss_closed events after Close_wss clears conn_ref
- add focused reader-policy tests for close vs non-close inputs

## Why
Process inspection of the live runtime showed high CPU spikes with many Discord/Cloudflare :443 sockets in CLOSE_WAIT plus repeated gateway READY / remote close / invalid_session / identify logs. The likely race is an old reader enqueueing a late EOF or frame after a reconnect has already opened a newer WSS; without connection identity, the FSM can interpret that stale input as belonging to the current connection and close the new session.

## Validation
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build lib/gate/discord_gateway_client.ml test/test_discord_gateway_client.exe test/test_discord_gateway_state.exe bin/main_eio.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec test/test_discord_gateway_client.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec test/test_discord_gateway_state.exe
- git diff --check

## Live verification
- Restarted local live tmux runtime from combined CPU-fix worktree HEAD 49c8eaf1f, which includes this fix plus PR #21255.
- /health?full=1 reports commit 49c8eaf1f, repo_root=/Users/dancer/me/workspace/yousleepwhen/masc/.worktrees/live-cpu-fixes-20260616, websocket mode=standalone.
- Post-restart ps after stabilization: PID 13655 at ~2.8% CPU, RSS ~1.19GB.
- lsof showed one Discord Gateway :443 ESTABLISHED socket and no Discord CLOSE_WAIT buildup.
- sample showed the process mostly idle in camlIomux__Poll/poll, not in a Discord reconnect or keeper/provider hot loop.

### DIFF
--- a/lib/gate/discord_gateway_client.ml
+++ b/lib/gate/discord_gateway_client.ml
@@ -157,6 +157,23 @@ let run ~sw ~env ~token ~intents ~trigger_policy ~on_event ~on_ambient () =
   let has_connected_before = ref false in
   let last_reconnect_was_resume = ref false in
 
+  let connection_is_current conn =
+    match !conn_ref with
+    | Some current -> current == conn
+    | None -> false
+  in
+
+  let enqueue_if_current conn input =
+    if connection_is_current conn then begin
+      Eio.Stream.add input_mailbox input;
+      true
+    end else begin
+      Log.Discord.debug
+        "dropping Discord gateway input from inactive WSS connection";
+      false
+    end
+  in
+
   let clock = Eio.Stdenv.clock env in
   Eio.Fiber.fork ~sw (fun () ->
     let rec loop () =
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by combined draft PR #21260. #21260 includes this Discord stale gateway reader-input guard plus the related broadcast wake fanout fix, and is the live-deployed branch to track for this runtime CPU-fix slice.

PR #21326, opened 2026-06-16
perf(main_eio): reduce CPU/GC pressure from metric store, UTF-8, and dashboard trust snapshots

jeong-sik:
## Summary

Targeted optimizations based on the `sample` profile of `main_eio.exe`.

## Changes

1. **Otel_metric_store_core**: move `Printexc.get_callstack` out of the hot path. It was captured on every `with_lock` invocation; now only captured when the lock actually raises `Sys_error`.

2. **Safe_ops.sanitize_text_utf8**: add an ASCII fast-path. Most strings are plain ASCII; scan bytes directly and skip the expensive `String.get_utf_8_uchar` decode loop for the common case.

3. **Otel_metrics**: group samples by `(name, kind)` before emitting protobuf. Reduces the number of `Metrics.t` records and OTLP messages per export tick.

4. **Keeper_runtime_trust_snapshot.snapshot_json**: add a short-lived TTL cache. The snapshot is expensive (tail-reads decision log, tool-call log, receipts, approval audit, pending approvals) and requested repeatedly by dashboard renders. Cache key includes runtime `generation` and `last_turn_ts` so normal keeper progress invalidates it automatically.

## Verification

- `dune build -p masc` passes.
- `test_keeper_runtime_trust_snapshot` passes.
- `test_safe_ops` passes.
- Pre-existing failures (also on `main`): `test_otel_zero_fill`, `test_dashboard_http_core`.

### DIFF
--- a/lib/core/safe_ops.ml
+++ b/lib/core/safe_ops.ml
@@ -201,18 +201,29 @@ let is_disallowed_control_char (c : char) : bool =
 
 let sanitize_text_utf8 (s : string) : string =
   let len = String.length s in
-  let rec has_invalid_or_control i =
-    if i >= len then false
+  let rec ascii_clean i =
+    if i >= len then true
     else
-      let dec = String.get_utf_8_uchar s i in
-      let dlen = Uchar.utf_decode_length dec in
-      if dlen > 0 && Uchar.utf_decode_is_valid dec then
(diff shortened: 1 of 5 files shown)


jeong-sik:

PR #21422, opened 2026-06-17
fix(main_eio): fail switch on graceful shutdown to cancel background fibers

jeong-sik:
## Problem

Ctrl+C / SIGTERM triggered the shutdown phases but the process only exited after the 10s force-timeout. The root cause was that the main [Eio.Switch.run] fiber returned normally while non-daemon background fibers (metrics flush, dashboard refresh loops, host fd pressure poller, OTLP exporter, etc.) were still running. [Eio.Switch.run] then waits forever for those fibers to finish on their own; it does not cancel them automatically.

## Fix

After shutdown phases complete and SSE connections are closed, raise a dedicated [Graceful_shutdown] exception. This makes [Eio.Switch.run] fail the switch, which propagates [Eio.Cancel.Cancelled] to every remaining fiber and waits for them to finish cleanly.

## Verification

- Built with [dune build --root . bin/main_eio.exe].
- Ran the server and sent SIGINT / SIGTERM: process now exits in ~1s instead of 10s.
- Logs show [host_fd_pressure_poller: cancelled, exiting cleanly] and [MASC MCP: Background fibers finished, shutdown complete.].
- [test/test_shutdown_flag.exe] passes.

## Scope

Only touches [bin/main_eio.ml].

### DIFF
--- a/bin/main_eio.ml
+++ b/bin/main_eio.ml
@@ -501,10 +501,11 @@ let make_extended_handler routes =
 let run_server ~sw ~env ~host ~port ~base_path =
   (* Use the parent switch directly so that ALL fibers spawned by
      Server_runtime_bootstrap (background maintenance, keeper loops,
-     dashboard refresh, etc.) are children of this switch.  When
-     Eio.Fiber.first cancels the run_server fiber on SIGTERM, the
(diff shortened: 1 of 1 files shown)


jeong-sik:

PR #21428, opened 2026-06-17
fixup(dashboard): keeper chat UX follow-up CI fixes

jeong-sik:
Follow-up fix for #21423.

- Removes unused `dequeueInput` / `markInputSent` imports that broke Dashboard typecheck.
- Bumps `docs/rfc/.next-number` from 0252 to 0254 so the RFC ledger stays monotonic (RFC-0253 already exists).,timeout:120}


jeong-sik:
Superseded by #21429 on a fresh branch.

PR #21466, opened 2026-06-17
fix(prompts): remove masc_web_* literal to silence runtime token warning

jeong-sik:
The keeper prompt token integrity scanner (`Keeper_prompt_token_integrity`) flags any `masc_*` token in rendered prompts that does not resolve through `Keeper_tool_resolution`.

Line 92 of `config/prompts/keeper.unified.system.md` previously used the literal `masc_web_*` as a negative example, producing runtime warnings like:

```
keeper_prompt_token_integrity: unknown masc token "masc_web_*" in system_prompt for keeper ...
```

Rephrase the guidance with only the snake_case variants (`web_search`, `web_fetch`), which are not `masc_*` tokens and therefore do not trigger the scanner. This also keeps the `no-tool-substrate-adapter-surface` build ratchet passing (no literal `masc_web_search`/`masc_web_fetch`).


jeong-sik:
Closing: the token scanner produces too much noise for the value; will address the broader ratchet separately.

PR #21492, opened 2026-06-18
fix(dashboard): keep queued keeper chat entries across history merges

jeong-sik:
대시보드에서 키퍼에게 메시지를 별내 응답이 queued 상태일 때, 대화를 나갔다 들어오면 hydrate/resume 병합 과정에서 빈 텍스트를 가진 queued assistant entry가 과거 빈 텍스트 history row와 중복으로 처리되어 사라지거나 에러로 보이는 문제를 수정합니다.

- In-flight(sending/streaming/queued) 로컬 entry들은 history merge 시 finalized 될 때까지 보존
- 재현/회귀 테스트 2건 추가

검증:
- npx vitest run --config vitest.config.ts src/keeper-actions.test.ts src/keeper-state.test.ts src/keeper-stream.test.ts: 84 pass
- npx tsc --noEmit: clean

### DIFF
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -399,6 +399,125 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     expect(keeperActionErrors.value.echo).toContain('서버 재시작')
     expect(fetchKeeperChatHistory).toHaveBeenCalledTimes(1)
   })
+
+  it('keeps the queued assistant after page reload while history has only the user message', async () => {
+    vi.useFakeTimers()
+    try {
+      _resetChatHydrationForTests()
+      _clearPendingKeeperChatRequestsForTests()
+      upsertPendingKeeperChatRequest({
+        requestId: 'kmsg_echo_1',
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed `d211075a93d6b2b7b93c2abb61aad8e88432f46c`.

No new blocking findings from the dashboard state/merge path. The patch preserves `sending`/`streaming`/`queued` local entries only while they are still marked in-flight; once the queued result finalizes to `delivered`/`error`, the existing role+text history dedup can replace it with canonical server history again. The added race tests cover both "history has only the user row" and "older empty-text history error collides with queued assistant".

Verification:
- Local: `pnpm -C dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts src/keeper-state.test.ts src/keeper-stream.test.ts` => 84 passed.
- Remote at review time: Dashboard and Lint are green.

Remaining gate is not this patch's dashboard logic: `Meta Guards` is red on the pending release-tag guard, with `Build and Test`/`TLA+ Model Checking` still pending.

PR #21489, opened 2026-06-17
feat(transport): harden WebSocket transport against frequent disconnects

jeong-sik:
## Summary

Hardens the MASC WebSocket/SSE transport layer to reduce frequent disconnects and lost session/auth state during keeper conversations.

## Problem

- Server-side WS writes were not serialized, so heartbeat, broadcast, and pong fibers could interleave on the same `Wsd.t`.
- The server sent WebSocket pings but never verified pongs, allowing half-open connections to linger.
- Heartbeat fibers were forked on the server-wide switch instead of the per-connection switch.
- `ws_session.closed` and related scalar state were plain mutable fields, creating data races under concurrent fibers.
- The dashboard client sent heartbeats without a timeout and reconnected blindly on every close code.
- SSE reconnects had no jitter or attempt ceiling, risking thundering-herd behavior.

### Server (`lib/server/server_mcp_transport_ws.ml`, `lib/server/server_ws_standalone.ml`)

- Added a per-session `write_mutex : Eio.Mutex.t` and wrapped every `Wsd.send_*` and `Wsd.close` call.
- Added atomic `last_pong_at : float Atomic.t` and `missed_pongs : int Atomic.t`.
- Heartbeat increments `missed_pongs`; incoming `Pong` frames reset it. Session is closed after `MASC_WS_MISSED_PONG_THRESHOLD` missed pongs (default 3).
- Converted `ws_session.closed` to `bool Atomic.t` and made other hot-path scalar fields atomic.
- Forked heartbeat fibers on the per-connection switch so they are cancelled automatically when the connection closes.
- Improved standalone accept-loop backoff with ±20% jitter and a 5 s cap.

### Dashboard client (`dashboard/src/dashboard-ws.ts`)

- Added a 10 s timeout to heartbeat RPCs via `DASHBOARD_WS_HEARTBEAT_RPC_TIMEOUT_MS`.
- Stopped reconnecting on fatal close codes (`1008`, `1011`) and on abnormal closes after a failed `dashboard/hello`.
- Added discovery-cache failure counting; the cached `/ws` URL is invalidated only after 3 consecutive failures.

### SSE transport (`dashboard/src/transports/sse-transport.ts`)

- Added jitter to exponential reconnect backoff.
- Added `TRANSPORT_RETRY_MAX_ATTEMPTS` ceiling to prevent infinite reconnect loops.

### Tests

- Added pong-state / threshold tests in `test/test_ws_transport.ml`.
[...]

PR #21497, opened 2026-06-18
fix(transport): address review follow-ups from #21489

jeong-sik:
Addresses the three unresolved review threads from chatgpt-codex-connector on #21489.

- P1: dashboard/src/transports/sse-transport.ts — keep SSE reconnecting after transient outages instead of giving up at retryMaxAttempts.
- P2: lib/server/server_ws_standalone.ml / lib/server/server_mcp_transport_ws.ml — use atomic increment for the missed-pong counter so a pong reset is not overwritten by the heartbeat.
- P2: dashboard/src/dashboard-ws.ts — do not treat 1011 (server error) as fatal; let the normal reconnect backoff handle it.

### DIFF
--- a/dashboard/src/dashboard-ws.test.ts
+++ b/dashboard/src/dashboard-ws.test.ts
@@ -483,7 +483,7 @@ describe('dashboard websocket route subscriptions', () => {
     expect(mockSockets).toHaveLength(1)
   })
 
-  it('stops reconnecting after a fatal server error close', async () => {
+  it('reconnects after a transient 1011 server error close', async () => {
     vi.useFakeTimers()
     installWebSocketMocks()
 
@@ -503,7 +503,8 @@ describe('dashboard websocket route subscriptions', () => {
 
     expect(dashboardWsConnected.value).toBe(false)
     expect(dashboardWsReady.value).toBe(false)
-    expect(mockSockets).toHaveLength(1)
+    expect(mockSockets).toHaveLength(2)
+    expect(mockSockets[1]!.readyState).toBe(MockWebSocket.CONNECTING)
   })
 
   it('stops reconnecting after an abnormal close after hello is rejected', async () => {
--- a/dashboard/src/dashboard-ws.ts
+++ b/dashboard/src/dashboard-ws.ts
@@ -64,8 +64,9 @@ let helloFailed = false
 const pending = new Map<number, PendingRpc>()
 
 // WebSocket close codes that indicate a persistent failure and should stop
-// reconnection attempts (1008 policy violation, 1011 server error).
-const FATAL_CLOSE_CODES = new Set([1008, 1011])
+// reconnection attempts (1008 policy violation). 1011 (server error) is left
+// to the normal reconnect backoff because it is often transient.
+const FATAL_CLOSE_CODES = new Set([1008])
 
 function sessionStorageOrNull(): Storage | null {
   if (typeof sessionStorage === 'undefined') return null
--- a/dashboard/src/transports/sse-transport.test.ts
+++ b/dashboard/src/transports/sse-transport.test.ts
@@ -133,7 +133,7 @@ describe('createSseTransport', () => {
     expect(instances).toHaveLength(2)
   })
 
-  it('stops reconnecting after retryMaxAttempts and emits close', () => {
+  it('keeps reconnecting after retryMaxAttempts and emits close once', () => {
     const events: { type: string }[] = []
     const transport = createSseTransport('/mcp', {
       retryBaseMs: 10,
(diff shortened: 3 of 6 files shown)


jeong-sik:
## 고쳐짐 확인 — 내 tick12 Finding 1(non-atomic RMW) 정확히 해소 + Codex 3건 검증

#21489 리뷰(#issuecomment-4736892240)에서 제가 medium으로 짚은 missed-pong non-atomic RMW가 이 follow-up에서 정확히 고쳐졌습니다.

### ✅ P2 missed-pong RMW (내 Finding 1) — `Atomic.incr`로 해소
두 사이트 모두 교체 확인:
```ocaml
- Atomic.set session.missed_pongs (Atomic.get session.missed_pongs + 1)
+ Atomic.incr session.missed_pongs
```
`Atomic.incr`는 단일 atomic RMW라, heartbeat의 증가와 `record_pong`의 reset(`Atomic.set ... 0`)이 인터리브해도 lost-update가 없습니다. 제가 우려한 "건강한 연결이 stale write로 닫히는 race"가 닫혔고, `.mli`의 "cross-fiber atomic" 정당화와 코드가 이제 일치합니다. PR 설명("a pong reset is not overwritten by the heartbeat")도 정확.

### ✅ P1 SSE ceiling 제거 — herd-방지 의도 안 깨짐 (검증함)
retryMaxAttempts 도달 시 `return`(중단) → `close` 1회 emit(`closeNotified` 가드) + **계속 재연결**로 변경. 적대적으로 "이게 #21489 원본의 thundering-herd 방지 ceiling을 되돌리는가?"를 확인했으나 — **backoff cap이 유지**됩니다: `backoff = Math.min(retryMs, maxMs)` + jitter + `retryMs = Math.min(retryMs*2, maxMs)`. 무한 재연결이라도 maxMs 간격으로 수렴하므로 herd 방지는 backoff cap이 담당(ceiling이 주 수단이 아니었음). transient outage에 영구 포기 안 하는 게 대시보드엔 맞는 트레이드오프. 의미 변경(ceiling = 중단→close-알림 임계)이 의도적이고 안전.

### ✅ P2 1011 non-fatal — 합리적
`FATAL_CLOSE_CODES`를 `{1008,1011}` → `{1008}`로. 1011(server error)은 종종 transient(서버 재시작 등)라 정상 backoff로 처리하는 게 맞음. 1008(policy violation)만 fatal 유지 — 권한/프로토콜 위반은 재연결해도 동일 거부라 정당.

### 잔존 (이 PR 범위 밖) — 내 Finding 2 미해결
#21489의 다른 finding이었던 **MCP 경로(`server_mcp_transport_ws.ml`) heartbeat가 `server_sw`에 fork + 미사용 `_conn_sw`**(standalone과 비대칭, poll-exit ≤30s)는 이 follow-up(Codex 3 thread)엔 포함 안 됐습니다. 별도 처리 권장 — "per-connection switch로 자동 취소" 주장이 MCP 경로에선 여전히 코드와 불일치.

non-atomic RMW가 정확히 닫혔고 SSE/1011 변경도 herd-방지를 깨지 않음을 확인했습니다.

— 자동 적대적 리뷰 루프 (grounded: Atomic.incr 두 사이트 교체 + sse-transport backoff Math.min cap 유지 + FATAL_CLOSE_CODES {1008} + Finding2 MCP switch 미포함)

jeong-sik:
Reviewed head `4c0584ebf563e3c4aed7a179f002bc901ba8aa5e`.

What is fixed:
- The SSE retry path no longer permanently gives up after `retryMaxAttempts`; it emits `close` once and keeps the capped backoff loop alive. The `closeNotified` reset on `open`/`disconnect` is consistent with that new meaning.
- `1011` is no longer treated as fatal in the dashboard WS client; `1008` remains fatal, which is the right split for policy-vs-transient close codes.
- The old stale read/write shape on `missed_pongs` is removed; both heartbeat sites now use one atomic RMW instead of `get` followed later by `set`.

Remaining blocking issue: `Atomic.incr` does not fully satisfy the stated “pong reset is not overwritten by the heartbeat” contract. In both heartbeat loops, the order is still `send_ping`; then `Atomic.incr session.missed_pongs`. If the client pong is handled between those two operations, `record_pong` resets the counter to `0`, and the heartbeat then increments it back to `1`. That is not the old stale-value overwrite, but it is still a reset-after-pong being overwritten by the heartbeat. With the default threshold `3` it is less likely to close a healthy connection; with the documented/allowed `MASC_WS_MISSED_PONG_THRESHOLD=1`, the next heartbeat can close despite a pong having arrived.

This should be made reset-wins, not merely atomic-RMW. The simple direction is to increment/count the outstanding ping before sending it, so any subsequent `record_pong` wins, or use a ping generation / `last_pong_at` comparison so the heartbeat only increments if no pong arrived after the ping attempt began.

Also still out of scope from the previous #21489 review: `cleanup_session` continues to hold the global `sessions_mutex` while taking `session.write_mutex` and closing the WSD (`lib/server/server_mcp_transport_ws.ml:866-880`). This follow-up does not address that lock-order/close-under-global-lock blocker.

Local validation caveat: `scripts/dune-local.sh build test/test_ws_transport.exe` was refused by the repo guard because a bare Dune process is still live outside the wrapper (`dune build test/test_exec_shell_command_gate.exe --root .`). The detached review worktree also has no `dashboard/node_modules`, so I did not run the dashboard Vitest targets locally.

jeong-sik:

PR #21551, opened 2026-06-19
fix(dashboard): keeper-v2 design-system audit cleanup

jeong-sik:
## Summary

Hostile-review follow-up for the keeper-v2 / v2-13 design-system drift in the MASC dashboard.

- **copilot-dock**: derive surface labels from `DASHBOARD_NAV_ITEMS`; remove hardcoded co-view numbers and per-surface starter prompts; collapse the default-keeper fallback to a single entry; migrate tone names to the Cockpit design-system set (`err` / `warn` / `brass`).
- **navigation**: remove hidden dead Monitor sections (`journey`, `cognition`, `transport-health`, `feature-health`) and redirect legacy `monitoring:cognition` deep links to `agents`.
- **cockpit-entrypoints**: map Cognition mode and its legacy subtabs to `monitoring:agents` so every entrypoint resolves to a live section.
- **status-tray / v2-shell**: remove `backdrop-filter: blur(...)` to match the design-system rule that blur is only allowed on code-overlay pinned notes.
- Update affected tests to expect `agents` for cognition routes and the `IDE` label for the code surface.

## Verification

- `pnpm typecheck` ✅
- `pnpm lint` ✅
- `pnpm test --run` ✅ — 561 test files / 7,146 tests + 7 Solid test files / 85 tests passed

## Notes

This PR intentionally leaves the `CognitionPlane` component in place; it is simply no longer reachable through the sidebar or cockpit redirects, which now land on the Monitor agents lane.

### DIFF
--- a/dashboard/src/cockpit-entrypoints.test.ts
+++ b/dashboard/src/cockpit-entrypoints.test.ts
@@ -142,7 +142,7 @@ describe('cockpit entrypoint registry', () => {
     })
     expect(cockpitTargetForParams({ mode: 'Cognition', tab: 'keeper-cognition' })).toEqual({
       tab: 'monitoring',
-      params: { section: 'cognition', view: 'keeper' },
+      params: { section: 'agents', view: 'keeper' },
     })
     expect(cockpitTargetForParams({ mode: 'IDE', tab: 'source' })).toEqual({
       tab: 'code',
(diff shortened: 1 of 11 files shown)


jeong-sik:
Adversarial follow-up pushed: restored hidden-but-routeable Runtime diagnostics for `transport-health` and `feature-health`. The previous cleanup kept the Runtime panel links but removed the nav SSOT entries, so those links canonicalized back to `agents`. Primary Monitor IA still exposes only the four visible lanes.

Local checks:
- `pnpm -C dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/config/navigation.test.ts src/router.test.ts src/components/status.test.ts src/components/runtime-panel.test.ts` PASS
- `pnpm -C dashboard typecheck` PASS
- `pnpm -C dashboard lint` PASS
- `git diff --check` PASS

Note: the earlier Meta Guards failure reproduced locally as branch-protection drift: `enforce_admins.enabled=false; expected true`. That is repository settings drift, not this PR diff.

jeong-sik:
Thanks for the adversarial follow-up. Both points are now addressed in the branch:

- `transport-health` and `feature-health` are kept as hidden-but-routeable Runtime diagnostics sections (commit `93812bc1d5`) and are included in the nav-event allowlist.
- The earlier `Meta Guards` nav-event allowlist parity failure was fixed by syncing `lib/dashboard/dashboard_nav_event.ml` with `dashboard/src/config/navigation.ts` and removing the retired `cognition` section from `SECTIONS_WITH_VIEW` (commit `989bcb14d6`).

Verification:
- Local full dashboard suite passes: 561 files / 7,147 tests + 7 Solid files / 85 tests.
- `pnpm -C dashboard typecheck` and `pnpm -C dashboard lint` pass locally.
- CI: `Meta Guards`, `Lint`, `Health`, `ocamlformat-check`, and `Dashboard` are green. Only `Build and Test` and `TLA+ Model Checking` are still in progress.

PR #21556, opened 2026-06-19
fix(shutdown): make hook registry atomic to prevent registration races

jeong-sik:
## Summary

Convert the global shutdown hook registry from a plain [hook list ref] to a [hook list Atomic.t] to eliminate a race condition in multi-fiber Eio environments.

## Problem

- [register] did [hooks := hook :: !hooks], which is read-modify-write on a non-atomic reference.
- [sorted_hooks] read [!hooks] while another fiber could be mutating the list.
- Concurrent registrations could lose hooks or observe inconsistent state.

## Fix

- [register] now atomically prepends via a CAS retry loop.
- [sorted_hooks] takes an atomic snapshot and sorts by priority as before.

## Verification

- Built: scripts/dune-local.sh build lib/masc.cmxa
- Tested: scripts/dune-local.sh runtest test/test_lifecycle.exe (4 passed, 0 failed)

Fixes: masc_shutdown_hooks_atomic

### DIFF
--- a/lib/shutdown.ml
+++ b/lib/shutdown.ml
@@ -103,13 +103,19 @@ let is_shutting_down_global () = Atomic.get shutting_down_flag
    making the keeper supervisor's graceful-shutdown branch unreachable. *)
 let mark_shutting_down () = Atomic.set shutting_down_flag true
 
-let hooks : hook list ref = ref []
+let hooks : hook list Atomic.t = Atomic.make []
 
 let register ~name ?(priority = 50) action =
-  hooks := { name; priority; action } :: !hooks
+  let new_hook = { name; priority; action } in
+  let rec loop () =
+    let old = Atomic.get hooks in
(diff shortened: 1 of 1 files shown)


jeong-sik:
[Adversarial review · head f79e160dcc]

## hook registry CAS 경쟁 해소 — 정확합니다

`hook list ref`의 동시 `:=` 경쟁을 `hook list Atomic.t` + CAS 루프(`get → compare_and_set → 실패 시 loop`)로 바꿨습니다. prepend가 원자적이고, `sorted_hooks`가 `Atomic.get` snapshot을 정렬해 get 시점 정합성이 보장됩니다. CAS 루프는 유한합니다(경쟁이 finite).

## 약한 신호 — `compare` 다형 비교

`List.sort (fun a b -> compare a.priority b.priority)`에서 다형 `compare`를 씁니다. `priority`가 int이면 `Int.compare`가 더 명시적이고 다형 `compare`의 runtime 비용을 피합니다(trivial).

PR #21580, opened 2026-06-19
refactor(backend): use Result syntax in lib/backend/backend.ml

jeong-sik:
Modernize Result usage in lib/backend/backend.ml by replacing verbose explicit match chains with Stdlib.Result.Syntax combinators (let*/let+).

Changes:
- Add open Result.Syntax after include Backend_types.
- Replace pass-through Error/Ok matches in key_to_path, get, set, delete, prefix_scan_root, list_keys_by_prefix_scan, atomic_increment, atomic_get, atomic_update, health_check, and extend_lock with let* bindings.
- Simplify validate_prefix and exists by relying on existing validation via validate_key/key_to_path.
- Replace error-swallowing delete match in release_lock with Result.is_ok.
- Keep public signatures unchanged; behavior is preserved.

### DIFF
--- a/lib/backend/backend.ml
+++ b/lib/backend/backend.ml
@@ -26,6 +26,8 @@ module Compression = Backend_compression
    to avoid circular dependency. Re-exported here for API compatibility. *)
 include Backend_types
 
+open Result.Syntax
+
 (** {1 FileSystem Backend (Eio)} *)
 
 module FileSystem = struct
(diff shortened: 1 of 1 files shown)


jeong-sik:
[적대 리뷰 · head 5d51384415fd]

## Result syntax 시리즈 7번째 (backend.ml)

#2142-#2147(checkpoint/protocol/llm_provider)에 이은 backend 영역 연장. 같은 `open Result_syntax` + `let*/and*` 패턴, 동작 보존 순수 리패터. 같은 `and*` 순서 약신호(에러 보고 순서 고정) 적용 — 동시 에러 시 detail 메시지로 보상 확인하면 완전합니다.

PR #21596, opened 2026-06-19
fix(test): split keeper effective meta runtime aliases

jeong-sik:
## Summary
- fix `test_keeper_effective_meta_overlay` so runtime catalog initialization uses `Masc.Runtime`
- keep keeper meta reconciliation calls on `Masc.Keeper_runtime`

## Validation
- Not run yet; creating PR first per request.
- Local Dune verification was blocked by concurrent Dune/opam locks before PR creation.


jeong-sik:

PR #21718, opened 2026-06-20
feat(memory-os): RFC-0259 P4 — recall suppression of unverified-volatile

jeong-sik:
## Summary
- demote unverified-volatile external-ref facts below durable facts during Memory OS recall
- render such facts with a leading `[UNVERIFIED — re-check before acting]` prefix when they are still shown
- keep durable non-external facts exempt from the volatile suppression rule
- add regression coverage for both the prefix and the actual demotion/drop behavior under `max_facts:1`

## Context
RFC-0259 P4 closes the read-path gap left after P3 demote-not-delete: terminal or stale external claims can remain in the store until TTL, so recall must not let them crowd out durable knowledge or read as verified current truth.

## Validation
- `ocamlformat --check lib/keeper/keeper_memory_os_recall.ml test/test_keeper_memory_os.ml`
- Not run to completion: `scripts/dune-local.sh build test/test_keeper_memory_os.exe` refused to start while bare Dune processes were active (`dune build .`, `dune build --root .`, `dune exec test/test_tools_coverage.exe --root .`).

### DIFF
--- a/lib/keeper/keeper_memory_os_recall.ml
+++ b/lib/keeper/keeper_memory_os_recall.ml
@@ -103,36 +103,85 @@ let staleness_marker ~now (fact : fact) =
     | None -> Printf.sprintf " [stale: unverified, seen %s ago — verify]" age_text)
 ;;
 
+(* RFC-0259 §3.5 (P4): a claim that names external state (P1's [external_ref])
+   and has not been re-grounded within the grounding horizon is *unverified
+   volatile* — its truth may have moved on (the closed-PR-still-asserted-open
+   class). Such a claim still passes [fact_is_current] until its (longer) volatile
+   TTL expires, so the gap between horizon and TTL is exactly where recall would
+   otherwise assert a possibly-stale external fact as live. The horizon is the P2
+   reconciler's, so "would be re-grounded by now" and "render as unverified" use
+   the same boundary. A non-volatile fact ([external_ref = None]) is never
+   unverified-volatile, however old — durable knowledge does not decay. *)
+let is_unverified_volatile ~now (fact : fact) =
+  match fact.external_ref with
+  | None -> false
+  | Some _ ->
+    let anchor =
+      match fact.last_verified_at with
+      | Some ts -> ts
+      | None -> fact.first_seen
+    in
+    now -. anchor > Keeper_memory_os_reconcile.default_grounding_horizon_seconds
+;;
+
+(* RFC-0259 §3.5: the hard, non-cosmetic prefix for an unverified-volatile claim.
+   Unlike [staleness_marker] (a trailing " [stale: … verify]" note that still reads
+   as an assertion), this leads the line so the reader sees "re-check" before the
+   claim text — closing gap #5 (the marker that annotated but did not suppress). *)
+let unverified_volatile_prefix = "[UNVERIFIED — re-check before acting] "
+
 (* RFC-0247 (purge): the recall line carries the fact's *structure* — category
    (a typed decision), provenance turn, and the worded staleness marker — but no
    number. The prior line printed [confidence=%.2f score=%.3f]; both were score
    machinery. The reader (an LLM) judges relevance and freshness from the claim
-   text and the staleness marker, not from a rank the producer can't justify. *)
+   text and the staleness marker, not from a rank the producer can't justify.
+   RFC-0259 §3.5: an unverified-volatile claim leads with the hard prefix instead
+   of the soft trailing staleness note. *)
 let render_fact ~now fact =
   let source = fact.source in
-  Printf.sprintf
-    "- [category=%s turn=%d]%s %s"
-    (sanitize_atom (category_to_string fact.category))
-    source.turn
-    (staleness_marker ~now fact)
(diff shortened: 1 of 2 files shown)


jeong-sik:
[적대 리뷰 grounding · head `d78e7b8f`]

RFC-0259 read-path을 완성하는 P4입니다(P1 #21644 + P2 #21665 + P3 #21668 위에). 제가 #21668/#21713 리뷰에서 짚은 "demote-not-delete가 terminal fact를 TTL까지 보존 → recall 노출 window가 *넓어진다*"는 tradeoff를 정확히 닫습니다. 검증한 항목:

## 1. typed 판정 + durable 면제 (string 분류기 아님) ✅
`is_unverified_volatile`이 `match fact.external_ref with | None -> false | Some _ -> now -. anchor > horizon`. 즉 **external_ref 없는 durable fact는 아무리 오래돼도 절대 suppress 안 됨**(휘발성을 age가 아니라 external_ref 유무=타입으로 판정). 30일 된 비-volatile fact가 prefix를 안 받는 테스트가 이 면제를 핀합니다. 휘발성 판단이 substring이 아니라 typed external_ref + 시간경계입니다.

## 2. 경계 상수 재사용 (magic number 없음) ✅
anchor 비교가 `Keeper_memory_os_reconcile.default_grounding_horizon_seconds`(= `volatile_external_ttl_seconds /. 2.0`, reconcile.ml:30)를 그대로 씁니다 — recall 억제 경계 == reconciler 재-grounding 경계. 두 경계가 같은 상수라 write-path(P3)와 read-path(P4)가 시간축에서 일관됩니다.

## 3. stable List.partition + 이중 표기 없음 ✅
`facts_recency_ranked`가 `List.partition`(OCaml stdlib, stable)으로 durable/volatile 분리 후 durable 우선 — 각 그룹 recency 순서 보존, volatile tail이 `take max_facts`에서 먼저 drop. `render_fact`는 `if is_unverified_volatile then [hard prefix, staleness_marker 없음] else [staleness_marker, prefix 없음]`로 **상호 배타** — 같은 fact에 hard prefix와 soft note가 겹치지 않습니다. `render_shared_fact`도 동일 prefix(cross-keeper 합의 ≠ external-truth 검증, 옳은 추론).

## 4. 경계 정합 ✅
read-path only, store mutation 0, P3 write-path와 직교(reconciler가 delete든 demote든 무관하게 적용). P2 상수만 의존. test +2 비-vacuous. 워크어라운드 게이트 무해당(cosmetic 마커를 structural demote + typed prefix로 *교체*).

원본 P4(#21675)가 P3 force-push에 덮여 main에 못 갔다는 이력도 납득됩니다. LGTM — RFC-0259 4단계(P1~P4)가 이걸로 read+write 양쪽 닫힙니다.

jeong-sik:
Review finding: the implementation demotes unverified-volatile facts, but the tests only pin the render prefix.

Current head `d78e7b8f` claims two behaviors:

1. unverified volatile facts get the hard `[UNVERIFIED — re-check before acting]` prefix;
2. `facts_recency_ranked` demotes them below durable facts so they are first dropped by `take max_facts`.

The new tests cover (1) and the durable exemption, but not (2). If the `List.partition` block in `facts_recency_ranked` were removed, both new tests would still pass because the stale external fact is the only fact in `test_recall_prefixes_unverified_volatile_fact`.

Please add one mixed recall test with at least: a recent/current unverified-volatile external-ref fact, an older durable fact, and `max_facts:1` or an ordering assertion. It should prove the durable fact renders first / survives the cap while the unverified volatile fact is demoted or dropped. That pins the actual “suppression” behavior, not only the warning label.

jeong-sik:
Addressed the test gap at head `2881b5783e`.

Added `test_recall_demotes_unverified_volatile_below_durable_cap`:

- seeds a recent but unverified-volatile external PR fact;
- seeds an older durable non-external fact;
- renders recall with `max_facts:1`;
- asserts the durable fact survives and the volatile external fact is dropped.

That makes the test fail if the stable `List.partition` demotion in `facts_recency_ranked` is removed, so the suppression behavior is now pinned, not just the warning prefix.

Validation:
- `ocamlformat --check lib/keeper/keeper_memory_os_recall.ml test/test_keeper_memory_os.ml`
- focused `scripts/dune-local.sh build test/test_keeper_memory_os.exe` was not run to completion because the wrapper refused to start while bare Dune processes were active (`dune build .`, `dune build --root .`, `dune exec test/test_tools_coverage.exe --root .`).

jeong-sik:
Current-head recheck `2881b578`: the test-gap finding is fixed.

`test_recall_demotes_unverified_volatile_below_durable_cap` now seeds both a recent unverified-volatile external PR fact and an older durable fact, renders with `max_facts:1`, and asserts the durable claim survives while the volatile claim is dropped. That fails if the `List.partition` demotion is removed, so the suppression behavior is now pinned, not only the hard prefix.

No behavioral blocker remains from my prior finding. Minor cleanup only: `test/test_keeper_memory_os.ml` currently has duplicate `module Reconcile = Masc.Keeper_memory_os_reconcile` aliases at the top; remove one before readying the PR. CI is still the merge gate.

PR #21730, opened 2026-06-20
fix(gate): repair main-red OCaml gate compile

jeong-sik:
## Summary
Repair the warn-error build break on `main` introduced by #21714 (`fix(discord): stream keeper replies through gate`). The merge of #21714 left two independent compile failures in the keeper/gate path; under `OCAMLPARAM=_,warn-error=+a` (Lint / Build and Test) both are hard errors, so `main` is red. This PR fixes both with the minimal change in each file.

## Root cause (two breaks, both from #21714, surfaced by fast-fail)
1. **`lib/gate_keeper_backend.ml` — warning 16 `[unerasable-optional-argument]`.**
   `dispatch_core ?on_text_snapshot ~sw ... ~content` has the optional followed only by labelled arguments, so it cannot be erased. `dispatch` calls `dispatch_core` without `on_text_snapshot`; the unerased optional propagates into `dispatch`'s inferred type as a trailing `?on_text_snapshot`, which is no longer included in the `.mli`'s `val dispatch : ... -> content:string -> Gate_protocol.dispatch_result`.
2. **`lib/server/server_discord_in_process_gateway.ml` — unmatched `(match`.**
   `handle_message_create` opens `(match with_response_wait_typing_indicator ... with` at line 425 but the final branch closed only the inner `(match finish_discord_stream_reply ...)`, leaving the outer match's `(` unclosed. The compiler reports `Syntax error: ) expected` at the next top-level `let on_event` (line 512) and points at line 425 as the unmatched `(`.

The warn-error build fast-fails, so only (1) was visible on the first red CI; fixing (1) let the build progress from `lib` into `lib/server`, surfacing (2).

## Fix
- `lib/gate_keeper_backend.ml` (internal `dispatch_core`, not in `.mli`): add a trailing `()` positional to `dispatch_core` and its two call sites (`dispatch`, `dispatch_with_text_snapshot`) so the optional is erasable. `dispatch`'s inferred type then matches the `.mli` again. No change to the exposed `dispatch` / `dispatch_with_text_snapshot` signatures or behaviour, and no `.mli` change.
- `lib/server/server_discord_in_process_gateway.ml`: add the missing `)` that closes the outer `(match` in `handle_message_create` (`Reply_send_failed)` → `Reply_send_failed))`). Pure paren balance; no behaviour change.

## Validation
- `ocamlformat --check` clean on both files (ocamlformat-check passes).
[...]

PR #21739, opened 2026-06-20
fix(build): restore main build broken by #21714 and #21721

jeong-sik:
## 목표

main 빌드 RED 복구. 최근 머지 후 CI `Build and Test`/`Lint`가 fast-fail하면서 뒤의 compile break가 가려졌습니다. 적대적으로 끝까지 추적하면 현재 main에는 세 가지 compile-shape 문제가 있습니다.

## Break 1 — `lib/gate_keeper_backend.ml`

`dispatch_core`가 leading `?on_text_snapshot` 옵셔널을 받고 마지막 인자가 labeled `~content`로 끝나서 optional이 erase될 수 없습니다. 그 결과 `dispatch`의 inferred type에 `?on_text_snapshot`이 새어 `.mli`의 `dispatch` 선언과 불일치합니다.

수정: `dispatch_core`에 trailing `()` positional을 추가하고 내부 호출처 2곳에서 `()`를 전달합니다. `.mli` 무변경, 런타임 동작 변화 없음.

## Break 2 — `lib/server/server_discord_in_process_gateway.ml`

`Stream_failed` 분기 split 이후 `with_response_wait_typing_indicator` outer match를 닫는 `)`가 하나 빠져 `let on_event`에서 `Syntax error: ) expected`가 납니다.

수정: `Stream_overflow_send_failed` tail에서 누락된 outer match paren을 복원합니다.

## Break 3 — `test/test_keeper_memory_os.ml`

`module Reconcile = Masc.Keeper_memory_os_reconcile`가 top-level에 두 번 선언되어 `test_rfc0259_reconcile` target build에서 `Multiple definition of the module name Reconcile`로 실패합니다.

수정: 중복 alias 한 줄을 제거합니다.

## 검증

- `opam exec -- ocamlformat --check lib/gate_keeper_backend.ml lib/server/server_discord_in_process_gateway.ml test/test_keeper_memory_os.ml`
- `git diff --check origin/main...HEAD`
- `rg "module Reconcile" test/test_keeper_memory_os.ml` → single occurrence

Full Dune/build는 owner 지시에 따라 로컬에서는 돌리지 않았습니다. CI `Build and Test`가 이 PR의 authoritative gate입니다.

## 경계 / RFC

순수 compile-shape 복구입니다. 공개 `.mli` 변경, telemetry/counter, cap/cooldown, 문자열 classifier, runtime behavior 변경 없음. RFC 불필요.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #21742, opened 2026-06-20
fix(memory-os): drop duplicate module Reconcile in test (P4 build break)

jeong-sik:
## What
Remove a duplicate `module Reconcile = Masc.Keeper_memory_os_reconcile` binding in `test/test_keeper_memory_os.ml`.

## Why
RFC-0259 P4 (#21718) added the binding after `module Recall`, but an identical binding already existed after `module Consolidator` (a rebase artifact). OCaml rejects two definitions of the same module name in one structure:

```
Error: Multiple definition of the module name Reconcile.
       Names must be unique in a given structure or signature.
```

So `test/test_keeper_memory_os.ml` stopped compiling and **Build and Test went red at the P4 merge (`da40bea1ab`)**.

## Change
- Remove the P4-added duplicate line; keep the single pre-existing binding.
- No test logic change — `Reconcile` stays in scope for the P4 recall-suppression tests (`is_unverified_volatile`, demote-below-durable).

## Verification
- `dune build --root . test/test_keeper_memory_os.exe`: the `Multiple definition` error is resolved (see PR checks).
- Note: full Build and Test green also depends on a **separate** lib break — `gate_keeper_backend.mli` `on_text_snapshot` (introduced by #21714), fixed by **#21730**. That is out of scope here; this PR only unblocks the test-file compile.

## Scope
- Touches only `test/test_keeper_memory_os.ml` (1 line removed). Does NOT touch the gate fix (#21730).

RFC: RFC-0259 P4 follow-up. Test-build fix only; no new behavior.

## Adversarial review evidence
A 4-lens adversarial review (correctness / no-scores / over-suppression / composition) of the merged P4 found the duplicate `module Reconcile` as the **only** real defect (confirmed HIGH: hard OCaml compile error — "Names must be unique in a given structure"). The recall-logic lenses (no-scores compliance, over-suppression/false-positive, P1/P3 composition) returned **zero** findings — `is_unverified_volatile` + demote-below-durable are sound and score-free. Reproduced deterministically: `dune build --root . test/test_keeper_memory_os.exe` on the merged tree fails with `Multiple definition of the module name Reconcile`; after this one-line deletion that error is gone (only the separate `gate_keeper_backend` lib break remains, owned by #21730).

### DIFF
--- a/test/test_keeper_memory_os.ml
+++ b/test/test_keeper_memory_os.ml
@@ -8,7 +8,6 @@ module Librarian = Masc.Keeper_librarian
 module Librarian_runtime = Masc.Keeper_librarian_runtime
 module Prompt_names = Keeper_prompt_names
 module Recall = Masc.Keeper_memory_os_recall
-module Reconcile = Masc.Keeper_memory_os_reconcile
 module Consolidator = Masc.Keeper_memory_os_consolidator
 module Reconcile = Masc.Keeper_memory_os_reconcile
 


jeong-sik:
Superseded by #21739, which restores the main build comprehensively — it removes this same duplicate `module Reconcile` in test/test_keeper_memory_os.ml AND fixes the two lib breaks blocking the same Build-and-Test run (gate_keeper_backend #21714, server_discord paren #21721). This PR only addressed the test-file dedup (1 of 3 breaks), so it is a herd duplicate. Closing in favor of #21739. (My search missed #21739 because its title is "restore main build", not the file name.)

PR #21769, opened 2026-06-20
fix(keeper): erase delegation digest optional arg

jeong-sik:
## Summary

- Fix OCaml warning 16 in `Keeper_delegation_request.digest_id`.
- Add the `unit` terminator already used by neighboring helpers so the optional `?goal` argument is erasable.

## Validation

- `opam exec -- ocamlformat --check lib/keeper/keeper_delegation_request.ml`
- `git diff --check`

## Notes

- No local build run; compile validation is left to PR CI per operator direction.


jeong-sik:
Closing as duplicate of #21768. The patch is identical: one private helper gets a unit terminator and its call site is updated.

PR #21771, opened 2026-06-20
fix(schedule): surface readable recurrence status

jeong-sik:
## Summary
- add a stable recurrence_summary helper and expose next_due_at / approval_policy on schedule tool results
- project the same schedule readability fields through the dashboard runtime JSON
- render dashboard rows from backend-provided recurrence and approval summaries
- document daily and cron recurrence examples in the schedule create tool description

## Validation
- opam exec -- ocamlformat --check lib/schedule/schedule_domain.ml lib/schedule/schedule_domain.mli lib/tool_schedule.ml lib/server/server_dashboard_http_runtime_info.ml lib/tool_schemas/tool_schemas_schedule.ml test/test_schedule_domain.ml test/test_schedule_tool_wiring.ml
- git diff --check
- scripts/ocaml-structure-ratchet.sh
- scripts/ci/check-determinism-contract.sh
- scripts/validate-prompt-paths.sh

## Not run
- scripts/dune-local.sh build test/test_schedule_domain.exe test/test_schedule_tool_wiring.exe: blocked by existing bare dune build processes outside scripts/dune-local.sh
- pnpm exec tsc --noEmit --pretty false in dashboard: dashboard tsc binary not installed in local node_modules

### DIFF
--- a/dashboard/src/api/dashboard.ts
+++ b/dashboard/src/api/dashboard.ts
@@ -1837,10 +1837,15 @@ export interface DashboardScheduledAutomationRequest {
   requested_at_iso?: string
   due_at?: number
   due_at_iso?: string
+  next_due_at?: number | null
+  next_due_at_iso?: string | null
   expires_at?: number | null
   expires_at_iso?: string | null
   payload_digest?: string
   payload_kind?: string | null
+  recurrence_summary?: string | null
+  requires_separate_human_grant?: boolean
+  approval_policy?: string | null
   last_execution?: DashboardScheduledAutomationExecution | null
 }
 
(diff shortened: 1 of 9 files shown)


jeong-sik:
Review @ 69a5423e7e

No code blocker found. The new schedule fields are presentation/read-model data derived from the existing typed recurrence and grant policy, not a second scheduling engine. `next_due_at` stays tied to the stored request due time and terminal rows emit null, while eligibility still flows through the existing schedule domain/store paths.

One evidence boundary to keep in mind: this improves readable status in tool/dashboard responses, but it does not prove the runner advanced recurring schedules in a live loop. CI is still pending.

PR #21798, opened 2026-06-20
feat(skill): persist draft skill candidates

jeong-sik:
Summary:
- Persist reviewable draft skill candidates under .masc/draft-skills without installing or changing Keeper capabilities.
- Project crystallized procedural memory and eligible Memory OS facts (validated_approach/lesson) into candidate-only JSON/TOML/SKILL.md artifacts.
- Wire the post-turn memory lane so newly durable Memory OS skill facts close the Memory OS -> draft skill candidate loop.
- Surface latest draft candidates in the memory subsystem dashboard projection.

Validation:
- ocamlformat --check lib/skill_candidate_store.ml lib/skill_candidate_store.mli lib/keeper/keeper_agent_run_post_turn_memory.ml test/test_procedural_memory.ml
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_procedural_memory.exe
- ./_build/default/test/test_procedural_memory.exe
- bash scripts/ci/check-determinism-contract.sh
- bash scripts/ci/check-masc-oas-boundary.sh

### DIFF
--- a/dashboard/src/api/dashboard.ts
+++ b/dashboard/src/api/dashboard.ts
@@ -3577,6 +3577,19 @@ export interface MemorySubsystemsUserModelPrompt {
   producer?: string
 }
 
+export interface MemorySubsystemsDraftSkillCandidate {
+  id: string
+  agent_name: string
+  source_kind: string
+  source_ref: string
+  promotion_state: string
+  dir: string
+  json_path: string
+  toml_path: string
+  skill_md_path: string
+  created_at: number | null
+}
+
 export interface MemorySubsystemsResponse {
   generated_at: string
   hebbian: {
@@ -3612,6 +3625,14 @@ export interface MemorySubsystemsResponse {
     limit: number
     items: MemorySubsystemsUserModelItem[]
     errors?: MemorySubsystemsUserModelError[]
+  }
+  draft_skill_candidates?: {
+    total: number
+    shown: number
+    limit: number
+    index_path: string
+    items: MemorySubsystemsDraftSkillCandidate[]
+    error?: string | null
   }
   filters: {
     keepers: string[]
--- a/dashboard/src/components/memory-subsystems.focus.test.ts
+++ b/dashboard/src/components/memory-subsystems.focus.test.ts
@@ -88,7 +88,7 @@ const baseResponse: MemorySubsystemsResponse = {
     total: 1,
     shown: 1,
     limit: 100,
-    index_path: '/tmp/masc/.masc/draft-skills/index.jsonl',
+    index_path: '<base-path>/.masc/draft-skills/index.jsonl',
     items: [
       {
         id: 'skill-candidate-repeatable-debug-loop',
@@ -96,10 +96,10 @@ const baseResponse: MemorySubsystemsResponse = {
         source_kind: 'procedure',
         source_ref: 'procedure://keeper-alpha/repeatable-debug-loop',
         promotion_state: 'candidate',
-        dir: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop',
-        json_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.json',
-        toml_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.toml',
-        skill_md_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md',
+        dir: '<base-path>/.masc/draft-skills/skill-candidate-repeatable-debug-loop',
+        json_path: '<base-path>/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.json',
+        toml_path: '<base-path>/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.toml',
+        skill_md_path: '<base-path>/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md',
         created_at: 1,
       },
     ],
@@ -216,7 +216,7 @@ describe('MemorySubsystems focus targets', () => {
     await vi.waitFor(() => {
       expect(container.textContent).toContain('skill-candidate-repeatable-debug-loop')
       expect(container.textContent).toContain('candidate')
-      expect(container.textContent).toContain('/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md')
+      expect(container.textContent).toContain('<base-path>/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md')
     })
     expect(container.querySelector('[data-testid="draft-skill-candidates"]')).not.toBeNull()
   })
--- a/dashboard/src/components/memory-subsystems.focus.test.ts
+++ b/dashboard/src/components/memory-subsystems.focus.test.ts
@@ -221,6 +221,75 @@ describe('MemorySubsystems focus targets', () => {
     expect(container.querySelector('[data-testid="draft-skill-candidates"]')).not.toBeNull()
   })
 
+  it('renders draft skill candidate empty and error states', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      jsonResponse({
+        ...baseResponse,
+        draft_skill_candidates: {
+          total: 0,
+          shown: 0,
+          limit: 100,
+          index_path: '<base-path>/.masc/draft-skills/index.jsonl',
+          items: [],
+          error: 'draft skill index read failed',
+        },
+      }),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    render(html`<${MemorySubsystems} />`, container)
+
+    await vi.waitFor(() => {
+      expect(container.querySelector('[data-testid="draft-skill-candidates"]')).not.toBeNull()
+      expect(container.querySelector('[role="alert"]')?.textContent).toContain(
+        'draft skill index read failed',
+      )
+      expect(container.textContent).toContain('draft skill 후보 없음')
+      expect(container.textContent).toContain('total 0 · shown 0')
+    })
+  })
+
+  it('renders non-candidate draft skill states with nullable creation time', async () => {
+    const original = baseResponse.draft_skill_candidates!.items[0]!
+    const fetchMock = vi.fn().mockResolvedValue(
+      jsonResponse({
+        ...baseResponse,
+        draft_skill_candidates: {
+          ...baseResponse.draft_skill_candidates!,
+          total: 2,
+          shown: 2,
+          items: [
+            {
+              ...original,
+              id: 'skill-promoted-debug-loop',
+              promotion_state: 'promoted',
+              skill_md_path: '<base-path>/.masc/skills/debug-loop/SKILL.md',
+              created_at: null,
+            },
+            {
+              ...original,
+              id: 'skill-rejected-debug-loop',
+              promotion_state: 'rejected',
+              skill_md_path: '<base-path>/.masc/draft-skills/rejected/SKILL.md',
+              created_at: null,
+            },
+          ],
+        },
+      }),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    render(html`<${MemorySubsystems} />`, container)
+
+    await vi.waitFor(() => {
+      expect(container.textContent).toContain('skill-promoted-debug-loop')
+      expect(container.textContent).toContain('promoted')
+      expect(container.textContent).toContain('skill-rejected-debug-loop')
+      expect(container.textContent).toContain('rejected')
+      expect(container.textContent).toContain('<base-path>/.masc/skills/debug-loop/SKILL.md')
+    })
+  })
+
   it('focuses the episodes section without requesting memory entries for episodes focus', async () => {
     const responseWithoutEntries: MemorySubsystemsResponse = {
       ...baseResponse,
--- a/dashboard/src/components/memory-subsystems.focus.test.ts
+++ b/dashboard/src/components/memory-subsystems.focus.test.ts
@@ -84,6 +84,27 @@ const baseResponse: MemorySubsystemsResponse = {
     ],
     errors: [],
   },
+  draft_skill_candidates: {
+    total: 1,
+    shown: 1,
+    limit: 100,
+    index_path: '/tmp/masc/.masc/draft-skills/index.jsonl',
+    items: [
+      {
+        id: 'skill-candidate-repeatable-debug-loop',
+        agent_name: 'keeper-alpha',
+        source_kind: 'procedure',
+        source_ref: 'procedure://keeper-alpha/repeatable-debug-loop',
+        promotion_state: 'candidate',
+        dir: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop',
+        json_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.json',
+        toml_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/candidate.toml',
+        skill_md_path: '/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md',
+        created_at: 1,
+      },
+    ],
+    error: null,
+  },
   filters: {
     keepers: ['keeper-alpha'],
     outcomes: ['success'],
@@ -184,6 +205,20 @@ describe('MemorySubsystems focus targets', () => {
     })
     expect(container.querySelector('[data-testid="user-model-projection"]')).not.toBeNull()
     expect(container.querySelector('[data-testid="user-model-prompt-surface"]')).not.toBeNull()
+  })
+
+  it('renders draft skill candidates from the memory subsystem payload', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(jsonResponse(baseResponse))
+    vi.stubGlobal('fetch', fetchMock)
+
+    render(html`<${MemorySubsystems} />`, container)
+
+    await vi.waitFor(() => {
+      expect(container.textContent).toContain('skill-candidate-repeatable-debug-loop')
+      expect(container.textContent).toContain('candidate')
+      expect(container.textContent).toContain('/tmp/masc/.masc/draft-skills/skill-candidate-repeatable-debug-loop/SKILL.md')
+    })
+    expect(container.querySelector('[data-testid="draft-skill-candidates"]')).not.toBeNull()
   })
 
   it('focuses the episodes section without requesting memory entries for episodes focus', async () => {
--- a/dashboard/src/components/memory-subsystems.ts
+++ b/dashboard/src/components/memory-subsystems.ts
@@ -16,6 +16,7 @@ import {
   type MemorySubsystemsUserModelItem,
   type MemorySubsystemsUserModelError,
   type MemorySubsystemsUserModelPrompt,
+  type MemorySubsystemsDraftSkillCandidate,
 } from '../api/dashboard'
 import { formatTimeAgo } from '../lib/format-time'
 import { useManagedAsyncResource } from '../lib/use-managed-async-resource'
@@ -79,6 +80,9 @@ export const ARCHITECTURE_FLOW = `graph LR
     G1 --> D1
     F1 --> U1[user_model projection]
     U1 --> D1
+    M3 --> S1[skill_candidate_projection]
+    S1 --> S2[(.masc/draft-skills)]
+    S2 --> D1
     D1 --> UI[기억 서브시스템 패널]
 
     %% dark-fantasy palette (mermaid can't use CSS vars): store=bg/mold,
@@ -86,8 +90,8 @@ export const ARCHITECTURE_FLOW = `graph LR
     classDef store fill:#221815,stroke:#5a3028,color:#e8d8b8
     classDef action fill:#16210f,stroke:#5a7a3a,color:#e8d8b8
     classDef ui fill:#2a1d08,stroke:#c4461a,color:#e8d8b8
-    class F1,G1 store
-    class M1,M3,H1,H2,T2,U1 action
+    class F1,G1,S2 store
+    class M1,M3,H1,H2,T2,U1,S1 action
     class UI ui`
 
 const shortAgentLabel = (name: string) => {
@@ -627,6 +631,88 @@ function UserModelPanel({
   `
 }
 
+function DraftSkillCandidateRow({
+  candidate,
+}: {
+  readonly candidate: MemorySubsystemsDraftSkillCandidate
+}) {
+  return html`
+    <div
+      class="v2-monitoring-row grid grid-cols-[10rem_7rem_minmax(0,1fr)_minmax(12rem,20rem)] items-start gap-2 border-b border-[var(--color-border-default)] px-2 py-2 text-xs last:border-b-0 max-lg:grid-cols-[8rem_minmax(0,1fr)]"
+      role="listitem"
+      aria-label=${`${candidate.id} · ${candidate.promotion_state} · ${candidate.source_ref}`}
+    >
+      <span class="min-w-0 truncate font-mono text-[var(--color-fg-muted)]" title=${candidate.id}>
+        ${candidate.id}
+      </span>
+      <span class="rounded-[var(--r-1)] border border-[var(--color-border-default)] bg-[var(--color-bg-elevated)] px-1.5 py-0.5 text-center font-mono text-[var(--color-status-warn)]">
+        ${candidate.promotion_state}
+      </span>
+      <span class="min-w-0 truncate text-[var(--color-fg-primary)]" title=${candidate.source_ref}>
+        ${candidate.source_kind}: ${candidate.source_ref}
+      </span>
+      <span class="min-w-0 truncate font-mono text-[var(--color-fg-disabled)] max-lg:col-span-2" title=${candidate.skill_md_path}>
+        ${candidate.skill_md_path}
+      </span>
+    </div>
+  `
+}
+
+function DraftSkillCandidatesPanel({
+  total,
+  shown,
+  indexPath,
+  candidates,
+  error,
+}: {
+  readonly total: number
+  readonly shown: number
+  readonly indexPath: string
+  readonly candidates: readonly MemorySubsystemsDraftSkillCandidate[]
+  readonly error?: string | null
+}) {
+  return html`
+    <section
+      data-testid="draft-skill-candidates"
+      aria-label=${`Draft skill candidates · ${shown} shown`}
+      class="flex flex-col gap-2"
+    >
+      <div class="flex flex-wrap items-center gap-2">
+        <h3 class="text-base font-semibold text-[var(--color-fg-muted)]">Draft skill candidates</h3>
+        <span class="font-mono text-2xs uppercase tracking-[var(--track-caps)] text-[var(--color-fg-disabled)]">
+          ${indexPath}
+        </span>
+        <span class="ml-auto text-xs text-[var(--color-fg-muted)]">
+          total ${total} · shown ${shown}
+        </span>
+      </div>
+      ${
+        error
+          ? html`<div
+              role="alert"
+              class="rounded-[var(--r-1)] border border-[var(--warn-fg)] bg-[var(--warn-bg)] px-3 py-2 text-2xs text-[var(--color-fg-primary)]"
+            >${error}</div>`
+          : null
+      }
+      ${
+        candidates.length === 0
+          ? html`<div class="v2-monitoring-card rounded-[var(--r-1)] border border-[var(--color-border-default)] bg-[var(--color-bg-elevated)] p-4 text-center text-sm text-[var(--color-fg-muted)]">
+              draft skill 후보 없음
+            </div>`
+          : html`
+              <div
+                class="v2-monitoring-card rounded-[var(--r-1)] border border-[var(--color-border-default)] bg-[var(--color-bg-page)]"
+                role="list"
+                aria-label=${`${shown} draft skill candidates`}
+              >
+                ${candidates.map(candidate => html`<${DraftSkillCandidateRow} candidate=${candidate} />`)}
+              </div>
+            `
+      }
+    </section>
+  `
+}
+
 export function MemoryEntriesPanel({
   entries,
   visibleEntries,
@@ -792,6 +878,8 @@ export function MemorySubsystems({ focus }: MemorySubsystemsProps = {}) {
   const knownMemoryKinds = data?.filters?.memory_kinds ?? Array.from(new Set(entries.map(e => e.kind))).sort()
   const userModel = data?.user_model
   const userModelItems = userModel?.items ?? []
+  const draftSkillCandidateBlock = data?.draft_skill_candidates
+  const draftSkillCandidates = draftSkillCandidateBlock?.items ?? []
   const visibleEntries = useMemo(
     () => filterMemoryEntries(entries, memoryKindFilter.value),
     [entries, memoryKindFilter.value],
@@ -872,6 +960,14 @@ export function MemorySubsystems({ focus }: MemorySubsystemsProps = {}) {
         filtered=${userModel?.filtered ?? userModelItems.length}
         errors=${userModel?.errors ?? []}
         prompt=${userModel?.prompt}
+      />
+
+      <${DraftSkillCandidatesPanel}
+        total=${draftSkillCandidateBlock?.total ?? draftSkillCandidates.length}
+        shown=${draftSkillCandidateBlock?.shown ?? draftSkillCandidates.length}
+        indexPath=${draftSkillCandidateBlock?.index_path ?? '.masc/draft-skills/index.jsonl'}
+        candidates=${draftSkillCandidates}
+        error=${draftSkillCandidateBlock?.error ?? null}
       />
 
       ${
--- a/lib/fs_compat/fs_compat.ml
+++ b/lib/fs_compat/fs_compat.ml
@@ -274,6 +274,19 @@ let file_exists (path : string) : bool =
        | Eio.Io _ -> false)
 ;;
 
+(** Load entire file contents as string, or [None] when the file is
+    missing. Option-returning sibling of {!load_file} (which raises on a
+    missing path). [Sys_error] from a vanished file (TOCTOU race after the
+    [file_exists] check) is also mapped to [None]; other I/O failures of an
+    existing file propagate as [Sys_error], matching {!load_file}. *)
+let load_file_opt (path : string) : string option =
+  if not (file_exists path)
+  then None
+  else (
+    try Some (load_file path) with
+    | Sys_error _ when not (file_exists path) -> None)
+;;
+
 let file_size (path : string) : int option =
   with_fs_or_fallback
     ~path
--- a/lib/fs_compat/fs_compat.mli
+++ b/lib/fs_compat/fs_compat.mli
@@ -25,6 +25,12 @@ val has_fs : unit -> bool
 (** Load entire file as string. *)
 val load_file : string -> string
 
+(** Load entire file as string, or [None] when the file is missing.
+    Option-returning sibling of {!load_file} (which raises on a missing
+    path). Other I/O failures of an existing file propagate as
+    [Sys_error]. *)
+val load_file_opt : string -> string option
+
 (** Save string to file (overwrite). *)
 val save_file : string -> string -> unit
 
--- a/lib/keeper/keeper_agent_run_post_turn_memory.ml
+++ b/lib/keeper/keeper_agent_run_post_turn_memory.ml
@@ -92,6 +92,42 @@ let run
     ~keeper_id:meta.name
     librarian_input;
 
+  (* Memory OS -> draft skill loop: after librarian facts are durable, project
+     validated approaches / lessons into reviewable draft skill artifacts. This
+     stays advisory and never mutates the keeper tool surface. *)
+  (try
+     match
+       Skill_candidate_store.write_post_turn_candidates
+         ~base_path:config.base_path
+         ~keeper_id:meta.name
+         ~fact_tail_limit:Keeper_memory_os_io.fact_store_max
+         ~procedure_limit:8
+     with
+     | Ok [] -> ()
+     | Ok stored ->
+       Log.Keeper.info ~keeper_name:meta.name
+         "draft_skill_candidates wrote=%d dir=%s"
+         (List.length stored)
+         (Skill_candidate_store.drafts_dir ~base_path:config.base_path)
+     | Error msg ->
+       Otel_metric_store.inc_counter
+         Keeper_metrics.(to_string DispatchEventFailures)
+         ~labels:[ "keeper", meta.name; "site", "draft_skill_candidates" ]
+         ();
+       Log.Keeper.warn ~keeper_name:meta.name
+         "draft_skill_candidates failed: %s"
+         msg
+   with
+   | Eio.Cancel.Cancelled _ as e -> raise e
+   | exn ->
+     Otel_metric_store.inc_counter
+       Keeper_metrics.(to_string DispatchEventFailures)
+       ~labels:[ "keeper", meta.name; "site", "draft_skill_candidates" ]
+       ();
+     Log.Keeper.warn ~keeper_name:meta.name
+       "draft_skill_candidates failed: %s"
+       (Printexc.to_string exn));
+
   (* Memory bank compaction: dedup + consolidate if over threshold. *)
   (try
      let memory_summarizer =
--- a/lib/keeper/keeper_agent_run_post_turn_memory.ml
+++ b/lib/keeper/keeper_agent_run_post_turn_memory.ml
@@ -2,6 +2,8 @@
 
     Extracted from [Keeper_agent_run.run_turn] Step 8 body (RFC-0147 PR-4). *)
 
+let default_post_turn_procedure_candidate_limit = 8
+
 let run
   ~config
   ~meta
@@ -101,7 +103,7 @@ let run
          ~base_path:config.base_path
          ~keeper_id:meta.name
          ~fact_tail_limit:Keeper_memory_os_io.fact_store_max
-         ~procedure_limit:8
+         ~procedure_limit:default_post_turn_procedure_candidate_limit
      with
      | Ok [] -> ()
      | Ok stored ->
--- a/lib/keeper/keeper_memory_os_io.ml
+++ b/lib/keeper/keeper_memory_os_io.ml
@@ -279,6 +279,13 @@ let rewrite_facts_atomically_for_keepers_dir ~keepers_dir ~keeper_id facts =
   write_file_atomically path content
 ;;
 
+let rewrite_facts_atomically_for_base_path ~base_path ~keeper_id facts =
+  rewrite_facts_atomically_for_keepers_dir
+    ~keepers_dir:(Config_dir_resolver.keepers_dir_for_base_path ~base_path)
+    ~keeper_id
+    facts
+;;
+
 let rewrite_facts_atomically ~keeper_id facts =
   rewrite_facts_atomically_for_keepers_dir ~keepers_dir:(keepers_dir ()) ~keeper_id facts
 ;;
--- a/lib/keeper/keeper_memory_os_io.mli
+++ b/lib/keeper/keeper_memory_os_io.mli
@@ -60,6 +60,8 @@ val append_episode_bundle : keeper_id:string -> episode -> unit
 val rewrite_facts_atomically : keeper_id:string -> fact list -> unit
 val rewrite_facts_atomically_for_keepers_dir :
   keepers_dir:string -> keeper_id:string -> fact list -> unit
+val rewrite_facts_atomically_for_base_path :
+  base_path:string -> keeper_id:string -> fact list -> unit
 
 (** {1 Facts snapshot CAS} *)
 
--- a/lib/procedural_memory.ml
+++ b/lib/procedural_memory.ml
@@ -31,20 +31,24 @@ type procedure = {
 (* Paths                                                            *)
 (* ================================================================ *)
 
-let procedures_dir ~agent_name =
+let base_path_or_default = function
+  | Some base_path -> base_path
+  | None -> Env_config.base_path ()
+;;
+
+let procedures_dir ?base_path ~agent_name =
+  let base_path = base_path_or_default base_path in
   Filename.concat
-    (Filename.concat
-       (Common.masc_dir_from_base_path ~base_path:(Env_config.base_path ()))
-       "procedures")
+    (Filename.concat (Common.masc_dir_from_base_path ~base_path) "procedures")
     agent_name
 
-let procedures_path ~agent_name =
-  sprintf "%s/procedures.jsonl" (procedures_dir ~agent_name)
+let procedures_path ?base_path ~agent_name =
+  sprintf "%s/procedures.jsonl" (procedures_dir ?base_path ~agent_name)
 
-let with_procedures_lock ~agent_name f =
-  let dir = procedures_dir ~agent_name in
+let with_procedures_lock ?base_path ~agent_name f =
+  let dir = procedures_dir ?base_path ~agent_name in
   Fs_compat.mkdir_p dir;
-  File_lock_eio.with_lock (procedures_path ~agent_name) f
+  File_lock_eio.with_lock (procedures_path ?base_path ~agent_name) f
 
 (* ================================================================ *)
 (* JSON Serialization                                               *)
@@ -91,20 +95,20 @@ let of_json (json : Yojson.Safe.t) : procedure option =
 (* File I/O                                                         *)
 (* ================================================================ *)
 
-let load_procedures ~agent_name : procedure list =
-  with_procedures_lock ~agent_name (fun () ->
-    let path = procedures_path ~agent_name in
+let load_procedures ?base_path ~agent_name : procedure list =
+  with_procedures_lock ?base_path ~agent_name (fun () ->
+    let path = procedures_path ?base_path ~agent_name in
     Fs_compat.load_jsonl path
     |> List.filter_map of_json)
 
-let save_procedure ~agent_name (p : procedure) =
-  with_procedures_lock ~agent_name (fun () ->
-    let path = procedures_path ~agent_name in
+let save_procedure ?base_path ~agent_name (p : procedure) =
+  with_procedures_lock ?base_path ~agent_name (fun () ->
+    let path = procedures_path ?base_path ~agent_name in
     Fs_compat.append_jsonl path (to_json p))
 
-let rewrite_procedures ~agent_name (procs : procedure list) =
-  with_procedures_lock ~agent_name (fun () ->
-    let path = procedures_path ~agent_name in
+let rewrite_procedures ?base_path ~agent_name (procs : procedure list) =
+  with_procedures_lock ?base_path ~agent_name (fun () ->
+    let path = procedures_path ?base_path ~agent_name in
     let content =
       procs
       |> List.map (fun p -> Yojson.Safe.to_string (to_json p))
@@ -142,8 +146,8 @@ let is_crystallized (p : procedure) : bool =
 (** Get top-N crystallized procedures by confidence.
     Uses adaptive thresholding: standard (3+ evidence, 70% confidence)
     plus rare-but-perfect (2+ evidence, 100% confidence). *)
-let top_procedures ~agent_name ~limit : procedure list =
-  let procs = load_procedures ~agent_name in
+let top_procedures ?base_path ~agent_name ~limit : procedure list =
+  let procs = load_procedures ?base_path ~agent_name in
   procs
   |> List.filter is_crystallized
   |> List.sort (fun a b -> Float.compare b.confidence a.confidence)
--- a/lib/procedural_memory.ml
+++ b/lib/procedural_memory.ml
@@ -36,19 +36,19 @@ let base_path_or_default = function
   | None -> Env_config.base_path ()
 ;;
 
-let procedures_dir ?base_path ~agent_name =
+let procedures_dir ?base_path ~agent_name () =
   let base_path = base_path_or_default base_path in
   Filename.concat
     (Filename.concat (Common.masc_dir_from_base_path ~base_path) "procedures")
     agent_name
 
-let procedures_path ?base_path ~agent_name =
-  sprintf "%s/procedures.jsonl" (procedures_dir ?base_path ~agent_name)
+let procedures_path ?base_path ~agent_name () =
+  sprintf "%s/procedures.jsonl" (procedures_dir ?base_path ~agent_name ())
 
 let with_procedures_lock ?base_path ~agent_name f =
-  let dir = procedures_dir ?base_path ~agent_name in
+  let dir = procedures_dir ?base_path ~agent_name () in
   Fs_compat.mkdir_p dir;
-  File_lock_eio.with_lock (procedures_path ?base_path ~agent_name) f
+  File_lock_eio.with_lock (procedures_path ?base_path ~agent_name ()) f
 
 (* ================================================================ *)
 (* JSON Serialization                                               *)
@@ -95,20 +95,20 @@ let of_json (json : Yojson.Safe.t) : procedure option =
 (* File I/O                                                         *)
 (* ================================================================ *)
 
-let load_procedures ?base_path ~agent_name : procedure list =
+let load_procedures ?base_path ~agent_name () : procedure list =
   with_procedures_lock ?base_path ~agent_name (fun () ->
-    let path = procedures_path ?base_path ~agent_name in
+    let path = procedures_path ?base_path ~agent_name () in
     Fs_compat.load_jsonl path
     |> List.filter_map of_json)
 
 let save_procedure ?base_path ~agent_name (p : procedure) =
   with_procedures_lock ?base_path ~agent_name (fun () ->
-    let path = procedures_path ?base_path ~agent_name in
+    let path = procedures_path ?base_path ~agent_name () in
     Fs_compat.append_jsonl path (to_json p))
 
 let rewrite_procedures ?base_path ~agent_name (procs : procedure list) =
   with_procedures_lock ?base_path ~agent_name (fun () ->
-    let path = procedures_path ?base_path ~agent_name in
+    let path = procedures_path ?base_path ~agent_name () in
     let content =
       procs
       |> List.map (fun p -> Yojson.Safe.to_string (to_json p))
@@ -146,8 +146,8 @@ let is_crystallized (p : procedure) : bool =
 (** Get top-N crystallized procedures by confidence.
     Uses adaptive thresholding: standard (3+ evidence, 70% confidence)
     plus rare-but-perfect (2+ evidence, 100% confidence). *)
-let top_procedures ?base_path ~agent_name ~limit : procedure list =
-  let procs = load_procedures ?base_path ~agent_name in
+let top_procedures ?base_path ~agent_name ~limit () : procedure list =
+  let procs = load_procedures ?base_path ~agent_name () in
   procs
   |> List.filter is_crystallized
   |> List.sort (fun a b -> Float.compare b.confidence a.confidence)
--- a/lib/procedural_memory.mli
+++ b/lib/procedural_memory.mli
@@ -33,14 +33,14 @@ type procedure = {
 
 (** [.masc/procedures/{agent_name}/procedures.jsonl] under [base_path].
     Defaults to [Env_config.base_path ()] for existing callers. *)
-val procedures_path : ?base_path:string -> agent_name:string -> string
+val procedures_path : ?base_path:string -> agent_name:string -> unit -> string
 
 (** {1 File I/O} *)
 
 (** Load all persisted procedures for [agent_name]. Malformed JSONL
     lines are silently skipped (via [of_json]). Returns [[]] if the
     file does not exist. *)
-val load_procedures : ?base_path:string -> agent_name:string -> procedure list
+val load_procedures : ?base_path:string -> agent_name:string -> unit -> procedure list
 
 (** Append [p] to the agent's procedures file. Creates the directory
     if it does not exist. *)
@@ -59,4 +59,5 @@ val rewrite_procedures : ?base_path:string -> agent_name:string -> procedure lis
 val is_crystallized : procedure -> bool
 
 (** Top-N crystallised procedures sorted by confidence (descending). *)
-val top_procedures : ?base_path:string -> agent_name:string -> limit:int -> procedure list
+val top_procedures :
+  ?base_path:string -> agent_name:string -> limit:int -> unit -> procedure list
--- a/lib/procedural_memory.mli
+++ b/lib/procedural_memory.mli
@@ -31,24 +31,24 @@ type procedure = {
 
 (** {1 Paths} *)
 
-(** [.masc/procedures/{agent_name}/procedures.jsonl] under
-    [Env_config.base_path ()]. *)
-val procedures_path : agent_name:string -> string
+(** [.masc/procedures/{agent_name}/procedures.jsonl] under [base_path].
+    Defaults to [Env_config.base_path ()] for existing callers. *)
+val procedures_path : ?base_path:string -> agent_name:string -> string
 
 (** {1 File I/O} *)
 
 (** Load all persisted procedures for [agent_name]. Malformed JSONL
     lines are silently skipped (via [of_json]). Returns [[]] if the
     file does not exist. *)
-val load_procedures : agent_name:string -> procedure list
+val load_procedures : ?base_path:string -> agent_name:string -> procedure list
 
 (** Append [p] to the agent's procedures file. Creates the directory
     if it does not exist. *)
-val save_procedure : agent_name:string -> procedure -> unit
+val save_procedure : ?base_path:string -> agent_name:string -> procedure -> unit
 
 (** Rewrite the full procedures file atomically. On write error,
     logs via [Log.Config.warn] and returns [()]. *)
-val rewrite_procedures : agent_name:string -> procedure list -> unit
+val rewrite_procedures : ?base_path:string -> agent_name:string -> procedure list -> unit
 
 (** {1 Crystallisation} *)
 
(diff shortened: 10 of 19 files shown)


jeong-sik:
### BLOCKING (head `4384f178aaeb`) — test file has a syntax error

Current head cannot compile as-is. `test/test_procedural_memory.ml` has a duplicate `in` in the new path-containment test:

```ocaml
(* test/test_procedural_memory.ml:231-233 *)
let c : S.skill_candidate = { base with id = "../Escape Candidate" } in
in
let stored =
```

This will fail before the new store behavior is tested. Remove the extra `in` and re-run the focused procedural-memory test/build on the fixed head.

jeong-sik:
Follow-up pushed: d7e2b743dd\n\nAdversarial CI/path response:\n- Removed the stray `in` that caused Health to fail parsing `test_procedural_memory.ml`.\n- Replaced the new local path-component sanitizer with existing `Workspace_utils_backend_setup.sanitize_namespace_segment` reuse.\n- Added a 16-hex SHA suffix to draft skill candidate directories to reduce path length/collision risk while keeping a readable prefix.\n- Kept generated artifacts under `.masc/draft-skills/` and preserved the traversal regression test for `../Escape Candidate`.\n\nVerification run locally, intentionally no full build per operator request:\n- `opam exec -- ocamlformat --check ...` PASS\n- `git diff --check HEAD~1..HEAD` PASS\n- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` PASS\n- hardcoded local path/env scan over changed files: no hits

jeong-sik:
### CHANGE REQUEST (head `d7e2b743dd67`) — syntax blocker is fixed, but `candidate.toml` is still not a safe TOML writer

The duplicate-`in` blocker from `4384f178aaeb` is fixed on this head. The path component sanitizer also moved to the existing `Workspace_utils_backend_setup.sanitize_namespace_segment`, which is the right direction.

New blocker: `lib/skill_candidate_store.ml` hand-rolls TOML strings instead of using the existing `otoml` writer pattern already used in this repo (`Otoml.Printer.to_string` in `repo_store.ml` / `keeper_repo_mapping.ml`). The current escaping only handles backslash, quote, newline, CR, and tab:

```ocaml
let string_escape_toml s =
  ...
  | \ -> "\\\\"\n  | " -> "\\\""\n  | \n -> "\\n"\n  | \r -> "\\r"\n  | \t -> "\\t"\n  | c -> Buffer.add_char buf c\n```\n\nCandidate fields (`pattern`, `source_ref`, evidence/risk text) come from procedural memory / Memory OS text, so they are not restricted to this small escape set. A valid OCaml string containing another TOML-illegal control character will produce a malformed `candidate.toml`. Minimal repro matching the current renderer behavior:\n\n```text\npython3 -c 'import tomllib; s="pattern = \\\"a\\fb\\\"\\n"; tomllib.loads(s)'\n# TOMLDecodeError: Illegal character '\\x0c'\n```\n\nThis is not a style nit: the PR creates durable review artifacts under `.masc/draft-skills/`, and `candidate.toml` is one of those artifacts. A single unescaped control byte from model/user/memory text makes the reviewable metadata unparsable.\n\nFix direction: build an `Otoml.TomlTable` and serialize with `Otoml.Printer.to_string`, matching the repo-local TOML writer SSOT, or add a complete TOML string escaper plus a regression that writes a candidate containing e.g. form-feed/control text and parses `candidate.toml` back with `Otoml.Parser.from_string_result`.

**Comment by jeong-sik** [comment]
### fixed 확인 (head `d7e2b743dd67`) — previous syntax/build blocker closed

The earlier duplicate-`in` syntax blocker from `4384f178aaeb` is closed on the current head. I re-ran the focused target in the dedicated worktree at `d7e2b743dd67`:

```text
scripts/dune-local.sh build test/test_procedural_memory.exe
```

Result: PASS.

This does not clear the new CHANGE REQUEST above about hand-rolled TOML serialization; it only closes the prior compile/syntax issue.

jeong-sik:
[적대 리뷰 grounding · head `4ab4f4dc3` · draft skill candidate 영속]

path 처리(사용자 CRITICAL 포커스)를 impl까지 검증했고, traversal 방어는 견고합니다. blocking 없음 + positive. 단 sanitizer lossiness로 인한 collision LOW가 있습니다.

**path-traversal 구조적으로 불가능(impl 검증):** `draft_dir`가 `candidate.id`를 `Workspace_utils_backend_setup.sanitize_namespace_segment`로 거른 뒤 path component로 씁니다. 그 sanitizer를 직접 읽었습니다 — `[a-zA-Z0-9-_]`만 유지하고 **그 외 모든 문자(`.` `/` `\` 포함)를 `-`로 치환**, 빈 문자열은 "default". 따라서 `..`→`--`, `../etc/passwd`→`--etc-passwd`로 무력화되어 traversal이 불가능합니다(#21799의 exact `.`/`..` reject보다 강한 allowlist 방식). 기존 공유 sanitizer 재사용(재구현 아님).

**base_path scoped(하드코딩 아님):** `drafts_dir ~base_path = Filename.concat (Common.masc_dir_from_base_path ~base_path) "draft-skills"` — `.masc/draft-skills/`를 하드코딩 상대경로로 두지 않고 base_path에서 파생합니다. 멀티 workspace에서 다른 워크스페이스 draft를 침범하지 않습니다(ambient-read 함정 회피). `Fs_compat.mkdir_p` + `save_file_atomic` 재사용(원자적 write).

**[LOW · sanitizer lossiness → candidate collision/overwrite] traversal은 막지만 distinct id가 같은 dir로 collapse될 수 있습니다.** allowlist 치환은 lossy합니다 — 예: `my.skill`, `my/skill`, `my_skill`이 모두 `my-skill`로 매핑되어 **같은 `draft-skills/my-skill/` dir를 공유**합니다. 두 서로 다른 candidate.id가 충돌하면 한쪽 draft가 다른쪽을 덮어씁니다(`save_file_atomic`이 같은 path를 쓰므로). 확인 요청:
- `candidate.id`가 upstream에서 이미 `[a-zA-Z0-9-_]` slug 또는 uuid/hash를 포함해 **post-sanitize 유일성이 보장**되나요? 그렇다면 sanitizer는 defense-in-depth일 뿐이라 collision이 없습니다.
- free-form id라면, 충돌 시 덮어쓰지 않도록 sanitized component에 **id의 hash suffix**를 붙이거나, write 전 dir 존재+다른 id 충돌을 감지하는 가드가 필요합니다. index.jsonl이 "newest unique candidates first"라 하니 거기서 dedup 키가 sanitized-dir인지 raw-id인지도 함께 보면 좋습니다.

정리: traversal 방어·base_path scoping·재사용 모두 옳습니다. sanitizer lossy → candidate dir collision 가능성만 확인하면 됩니다. LGTM 조건부.

jeong-sik:
### 적대 리뷰 (head `8d1d4f911` · post-turn write wiring 델타) — 13:43 리뷰 이후 추가된 `write_post_turn_candidates`/`read_file_opt` 검토

13:43 리뷰는 head `4ab4f4dc3`(store/persist·path-traversal·base_path scoped·atomic write)를 다뤘습니다. 이후 `8d1d4f911`(`feat(skill): write post-turn draft candidates`)이 **post-turn write wiring**을 추가했고, 그 델타(`keeper_agent_run_post_turn_memory.ml +36`, `skill_candidate_store.ml +82/-8`)를 정독했습니다.

**옳게 된 것 (positive):**
- **`Eio.Cancel.Cancelled _ as e -> raise e`** — advisory side-effect(draft skill write)를 broad `exn`으로 격리하되 Cancelled는 재전파. keeper 턴을 깨지 않으면서 취소 신호는 보존하는 정확한 Eio 관행입니다.
- 실패는 `Otel_metric_store.inc_counter` + `Log.warn`으로 가시화(silent 아님), `Error msg`/`exn` 양 경로 처리. advisory write라 counter-as-fix 워크어라운드가 아니라 비-critical 경로 격리로 정당.
- `write_candidate_if_changed`의 idempotency 의도(매 턴 동일 candidate 재기록 회피)는 합리적.

**[MED · 재구현] `read_file_opt`가 기존 canonical reader를 손으로 다시 만듭니다.**
`skill_candidate_store.ml`의 새 `read_file_opt`는 raw `open_in_bin` + `Fun.protect` + `really_input_string`로 파일을 읽습니다. 그러나:
- `Fs_compat.load_file : string -> string`가 canonical reader이고, 모듈 docstring이 명시적으로 **"In code: `Fs_compat.load_file path` instead of `open_in ...`"**(`lib/fs_compat/fs_compat.ml:8`)라고 규약을 박아뒀습니다. 이 store는 **write에는 이미 `save_file_atomic`/`Fs_compat`를 재사용**하면서 read만 규약을 우회해 `open_in_bin`을 직접 씁니다.
- 게다가 `read_file_opt`(option 반환)는 이미 `lib/prompt_defaults.ml:48`과 `lib/repo_manager/keeper_repo_mapping.ml:268`에 **같은 이름·시맨틱**으로 존재합니다(≥3중 복제). `prompt_defaults` 판은 `In_channel.with_open_text`+`input_all` 사용 + **`Sys_error _ -> None` 처리**까지 있어 더 견고합니다 — 반면 이 PR 판은 `Sys.file_exists` 후 `open_in_bin` 사이 TOCTOU에서 `Sys_error`를 잡지 않습니다(바깥 `try`가 삼키지만, 덜 견고한 판을 새로 만든 것).
- 권고: `Fs_compat.load_file`(기존 `Sys.file_exists` 가드와 조합) 재사용, 또는 단일 `read_file_opt`를 공유 모듈(`Fs_compat`)로 승격해 세 복제를 한 정의로. byte-exact 비교 목적이면 binary read가 맞으니 `Fs_compat`에 binary 변형을 두는 게 SSOT.

**[LOW · magic number] 와이어링 호출부 `~procedure_limit:8`이 bare 리터럴입니다.**
형제 인자 `~fact_tail_limit:Keeper_memory_os_io.fact_store_max`는 named constant인데 `procedure_limit`만 `8`을 직접 박았습니다(CLAUDE.md "Magic Number 금지"). `default_post_turn_procedure_limit` 같은 상수로 추출하면 의도(턴당 최대 procedure candidate 수)가 드러나고 형제와 일관됩니다.

**[LOW · idempotency edge] `write_candidate_if_changed`가 `candidate.json`만으로 skip을 판정합니다.**
`write_candidate`는 json → toml → SKILL.md → index 순서로 씁니다. 이전 턴이 json 기록 후 toml/SKILL.md에서 실패하면(Result `Error` 전파), 디렉터리에 **json만 남고 SKILL.md는 부재**합니다. 다음 턴에 `read_file_opt json_path`가 일치 → `Ok None` → **skip**이라 불완전 candidate가 영구히 복구되지 않습니다. advisory 아티팩트 + mid-write 실패 조건이라 severity는 낮지만, "json 존재 ⟹ 전체 기록 완료"라는 암묵 가정이 성립하지 않습니다(json이 첫 write). 마지막 아티팩트(index 엔트리) 또는 세 경로 모두를 idempotency 키로 두면 안전합니다.

verdict: write wiring의 Eio/격리/관측성은 옳습니다. **재구현된 `read_file_opt`를 `Fs_compat.load_file`로 통합**하는 것이 directive("이미 있는 모듈 재구현 검사")의 핵심 시정점이고, 나머지 둘은 low 보완입니다.

(근거: compare 4ab4f4dc3...8d1d4f911 직접 정독; `fs_compat.ml:8` 규약 주석 + `fs_compat.mli:26 load_file`; `prompt_defaults.ml:48`/`keeper_repo_mapping.ml:268` 기존 `read_file_opt`; write 순서 json-first. 차등 적대 리뷰 2026-06-21.)

jeong-sik:
Addressed in `7f8dd70280` on top of `6804817a79`.

Changes:

- Replaced the hand-rolled `candidate.toml` string renderer with `Otoml.TomlTable` + `Otoml.Printer.to_string`.
- Kept the canonical reader path from the intervening commit: `read_file_opt = Fs_compat.load_file_opt`.
- Changed idempotency from `candidate.json`-only to all generated candidate artifacts (`candidate.json`, `candidate.toml`, `SKILL.md`) so a partial prior write is repaired on the next post-turn projection.
- Extracted `~procedure_limit:8` to `default_post_turn_procedure_candidate_limit`.

Regression coverage:

- TOML control-text round-trip through `Otoml.Parser.from_string_result`.
- Partial-write recovery where only `candidate.json` pre-exists.

Local checks:

- `opam exec -- ocamlformat --check lib/skill_candidate_store.ml lib/keeper/keeper_agent_run_post_turn_memory.ml test/test_procedural_memory.ml`
- `git diff --check origin/feat/skill-candidate-store-20260620..HEAD`

I did not run the focused Dune build locally because the operator asked to leave builds to CI.

jeong-sik:
### 고쳐졌다 확인 (head `7f8dd7028`) — 프로덕션 read가 canonical Fs_compat로 위임됨(+ TOML 렌더러도 라이브러리로 교체). 잔여: 테스트 헬퍼 1건 [LOW]

직전 리뷰의 [MED] — "`skill_candidate_store`가 raw `open_in_bin` + `really_input_string`로 `read_file_opt`를 손수 만들어 canonical `Fs_compat.load_file`를 재구현한다" — 가 올바른 방향으로 닫혔습니다.

**해소 1 — 재구현 대신 canonical 모듈 확장.** 로컬 hand-rolled reader를 지우고, **`Fs_compat`에 `load_file_opt : string -> string option`을 신설**(`fs_compat.ml:252`, `.mli:275`)한 뒤 `read_file_opt = Fs_compat.load_file_opt`로 위임합니다(`skill_candidate_store.ml:585`). 이는 "로컬 재구현"보다 우월한 해법입니다 — `string option`을 반환하는 canonical 변종이 SSOT 모듈 한 곳에 생겨, 다른 소비자도 try/with 없이 재사용할 수 있습니다. (기존 `load_file : string -> string`은 raise-on-missing이라, option 시맨틱이 필요한 자리엔 `_opt` 변종이 맞습니다.)

**해소 2 (bonus) — hand-rolled TOML 렌더러 → `Otoml` 라이브러리.** 후속(`7f8dd7028`)에서 `candidate.toml` 문자열 렌더러를 `Otoml.TomlTable` + `Otoml.Printer.to_string`로 교체했습니다. "라이브러리가 나은데 재구현"을 자발적으로 닫은 정합 변경입니다(이스케이프/타입 처리를 라이브러리에 위임).

**잔여 [LOW · 테스트 헬퍼 재구현 + 예외 비안전].** 테스트 파일에 새 `read_file` 헬퍼(`:863–869`)가 여전히 raw로 읽습니다:
```ocaml
let read_file path =
  let ic = open_in path in
  let len = in_channel_length ic in
  let content = really_input_string ic len in
  close_in ic; content
```
두 가지: (a) 이제 `Fs_compat.load_file`(canonical, raise-on-missing)이 가용하므로 테스트도 그걸 쓰면 일관됩니다 — 프로덕션은 고치고 테스트만 재구현이 남는 비대칭. (b) `Fun.protect` 없이 `close_in`이 본문 끝에 있어, `really_input_string`이 truncated 파일 등으로 raise하면 **`close_in`을 건너뛰어 fd가 샙니다**. 테스트라 수명이 짧아 실해는 낮지만, 같은 PR이 fd-안전을 강조하는 맥락이면 `Fs_compat.load_file path`로 한 줄 대체가 깔끔합니다. (참고: `:1177` `Fun.protect ~finally:rm_rf`는 temp_dir 정리용 정당한 사용 — finding 아님.)

verdict: 프로덕션 재구현 MED는 canonical 모듈 확장으로 해소됐고, TOML 렌더러도 라이브러리로 옳게 옮겼습니다. 테스트 헬퍼 `read_file`만 `Fs_compat.load_file`로 통일하면 반경이 완결됩니다.

(근거: `fs_compat.ml:252`/`.mli:275` `load_file_opt` 신설, `skill_candidate_store.ml:585` 위임; author note의 Otoml 교체; 테스트 `read_file` :863–869 raw + no Fun.protect. head 7f8dd7028. 차등 적대 리뷰 2026-06-21.)

jeong-sik:
## Adversarial review — persist draft skill candidates (HEAD `7f8dd7028097`, +995/-5)

I checked this against the known `read_file_opt` reimplementation cluster. The dedup direction is right and the persistence is solid; one N-of-M gap.

### [Affirm] Adds the canonical `load_file_opt` and routes the store through it
- New `Fs_compat.load_file_opt` — option-returning sibling of `load_file`, maps a missing-file (and the TOCTOU `Sys_error` after the `file_exists` check) to `None`, other I/O on an existing file still propagates as `Sys_error` (matches `load_file`). Correct boundary.
- `skill_candidate_store.ml:186` is now `let read_file_opt = Fs_compat.load_file_opt` — the store's hand-rolled reader is replaced by the canonical one. This is exactly the dedup the codebase needed here.
- `write_file` uses `Fs_compat.mkdir_p` + `Fs_compat.save_file_atomic` (the durability SSOT, fsync pair), and `write_candidate_if_changed` reads-then-compares before writing (idempotent, no needless churn). Index path uses `Fs_compat.load_jsonl`. Persistence is clean.

### [Change requested · MED] Only 1 of the 3 known `read_file_opt` sites is routed — the SSOT is added but the drift isn't closed
This PR adds the canonical `Fs_compat.load_file_opt` and routes `skill_candidate_store` through it, but two sibling reimplementations of the same "read file or None" pattern still inline their own:
- `lib/prompt_defaults.ml:48` — `let read_file_opt path = …`
- `lib/repo_manager/keeper_repo_mapping.ml:268` — `let read_file_opt path = …`

So the N-of-M is opened-but-not-closed: the SSOT function now exists, 1 of 3 known sites uses it, 2 still carry their own copy (which will drift independently — e.g. TOCTOU handling, Eio-vs-Unix, encoding). Since this PR already edits `Fs_compat` and the fix for each remaining site is the same one-line alias used at `skill_candidate_store.ml:186`, routing the two siblings here closes the loop in the same change rather than leaving them as the next PR's cleanup. (This is the cluster the canonical-file-reader note has flagged.)

### [Observation, broader — not blocking] The raising-`read_file` variant is also widely duplicated
Beyond the three `read_file_opt` sites, hand-rolled raising readers exist in `build_identity.ml:153`, `keeper_runtime_config.ml:120`, `keeper_secret_projection.ml:104`, `keeper_prompt_external.ml:41`, `server/server_routes_http_pages.ml:188` (each `open_in` + `really_input_string (in_channel_length ic)`), while `voice_bridge_transport.ml:33` and `server_routes_http_sidecar_paths.ml:130` already alias `Fs_compat.load_file`. That's a larger separate sweep, but it's the same smell — worth a tracking issue so the canonical readers actually displace the duplicates instead of coexisting with them.

jeong-sik:
Addressed the remaining read_file_opt SSOT review in 12cf45b7b9.

Changes:
- Routed lib/prompt_defaults.ml read_file_opt to Fs_compat.load_file_opt.
- Routed lib/repo_manager/keeper_repo_mapping.ml read_file_opt to Fs_compat.load_file_opt.
- Removed the now-unused local max_git_probe_file_bytes helper from keeper_repo_mapping.

Local checks:
- opam exec -- ocamlformat --check lib/prompt_defaults.ml lib/repo_manager/keeper_repo_mapping.ml
- git diff --check

Focused Dune check was attempted with scripts/dune-local.sh build test/test_procedural_memory.exe, but the wrapper refused to start because bare Dune processes are currently active on this machine. I left those processes untouched per workflow guardrails.

jeong-sik:
## Close-out — HEAD `12cf45b7b9`: 3rd-site dedup complete, but a domain safety bound was dropped

### [Resolved] All three `read_file_opt` sites now route through the canonical reader
- `prompt_defaults.ml`: `let read_file_opt = Fs_compat.load_file_opt` (the hand-rolled `In_channel.with_open_text` + `Sys_error` catch is gone).
- `keeper_repo_mapping.ml`: hand-rolled `read_file_opt` removed, aliased to `Fs_compat.load_file_opt`.

So the N-of-M I flagged is closed — the canonical `Fs_compat.load_file_opt` now backs all three sites. Good.

### [Caveat — verify] `keeper_repo_mapping` lost a 64KB size cap + regular-file-only check
The old `keeper_repo_mapping.read_file_opt` was not a plain "read or None" — it guarded with `safe_lstat` for `Unix.S_REG` **and** a `max_git_probe_file_bytes = 64 * 1024` cap, only reading regular files under 64KB (and the commit also deletes that constant). The canonical `Fs_compat.load_file_opt` has neither guard: it reads the whole file regardless of type or size. So the consolidation silently dropped two domain-specific safety bounds from the repo-mapping read path.

If the files `keeper_repo_mapping` reads are always small, trusted git-config/probe files, losing the cap is harmless and the consolidation is a clean win. But if any path it reads can be pathologically large or a non-regular file (a symlink to something huge, a `/dev`-style special, an attacker-controllable path), the uncapped read is a resource/behavior regression the dedup introduced. Please confirm the read targets are bounded/trusted; if not, keep a thin wrapper around `load_file_opt` that re-applies the size + `S_REG` guard. This is the classic dedup tradeoff — the canonical reader is more general, so the domain guard has to be preserved explicitly rather than rediscovered.

jeong-sik:
#21798 caveat 대응했습니다. `keeper_repo_mapping`의 `read_file_opt` dedup가 기존 git-probe domain guard(S_REG + 64KB cap)를 제거한 문제를 확인했고, canonical `Fs_compat.load_file_opt`는 유지하되 repo-mapping git probe 경로에는 `read_git_probe_file_opt` wrapper로 regular-file/size bound를 복원했습니다. 기존 `test_validate_path_access_playground_large_git_config_denied`가 다시 의미 있게 이 경계를 pin합니다. 검증: `opam exec -- ocamlformat --check lib/repo_manager/keeper_repo_mapping.ml`, `python3 scripts/check_test_coverage.py`, `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD`, `git diff --check`, path/env scan 통과. full build는 operator 지시에 따라 생략했습니다.

jeong-sik:
## Close-out — HEAD `8f99b1a1be`: read_file_opt dedup + git-probe guard both preserved

Rechecked the current head after the `keeper_repo_mapping` caveat response.

The previous SSOT gap is closed:

- `lib/skill_candidate_store.ml` routes `read_file_opt` through `Fs_compat.load_file_opt`.
- `lib/prompt_defaults.ml` now aliases `Fs_compat.load_file_opt`.
- `lib/repo_manager/keeper_repo_mapping.ml` no longer owns a generic duplicate `read_file_opt`.

The important follow-up caveat is also closed: `keeper_repo_mapping` did **not** lose its domain guard after dedup. Current head keeps a thin `read_git_probe_file_opt` wrapper that re-applies the regular-file + 64KB cap before delegating to `Fs_compat.load_file_opt`, so the canonical reader is reused without turning git probe reads into unbounded reads.

Local lightweight check:

- `git diff --check origin/main...refs/remotes/pr/21798`: pass

I did not run focused Dune locally because the repo still has unrelated bare Dune processes that trigger the local wrapper guard. No further blocker from this pass.

jeong-sik:
## Close-out — HEAD `8f99b1a1be`: preserve bounded git probe reads (domain bounds restored on top of the canonical reader)

This addresses my earlier caveat exactly the way I asked, and it folds in the canonical-reader consolidation I flagged separately. Closing both.

### [Resolved] The 64KB cap and S_REG guard are preserved as a thin wrapper over the canonical reader
`keeper_repo_mapping.read_file_opt` now keeps its domain bounds and delegates the actual read:

```ocaml
let max_git_probe_file_bytes = 64 * 1024
...
| Some { Unix.st_kind = Unix.S_REG; st_size; _ }
  when st_size <= max_git_probe_file_bytes -> Fs_compat.load_file_opt path
| Some { Unix.st_kind = (S_REG | S_DIR | S_CHR | S_BLK | S_LNK | S_FIFO | S_SOCK); _ } -> None
```

So the dedup goal (one canonical reader, no inline `In_channel.with_open_text`) is still achieved — `Fs_compat.load_file_opt` is the only reader — but the two domain-specific safety bounds I worried were silently dropped are back: regular-file-only (`S_REG`) and size-bounded (`<= 64KB`). Non-regular kinds (symlink, dir, devices, fifo, sock) still return `None`, so a symlink to something huge or a special file can't trigger an uncapped read. This is the right resolution of the dedup tradeoff — the canonical reader stays general, the domain guard is re-applied explicitly rather than rediscovered.

### [Resolved] `Fs_compat.load_file_opt` now exists as the canonical option-returning reader
The sibling added in `fs_compat.ml`/`.mli` is the right home: `load_file_opt` raises-through `Sys_error` for existing-but-unreadable files and returns `None` only for missing files, matching `load_file`'s contract. That removes the last reason for `prompt_defaults`/`keeper_repo_mapping`/`skill_candidate_store` to each re-roll an `In_channel.with_open_text` + `Sys_error` catch — all three now point at the one canonical function, with `keeper_repo_mapping` keeping its extra guard because its read target (git probe files) genuinely needs it. So the N-of-M scattered-reader smell is closed at the source.

### [Low] PR scope has widened beyond the title
The PR title is "persist draft skill candidates" but this head also carries the `Fs_compat.load_file_opt` addition, the three-site read dedup, the bounded-read restoration, changes to `keeper_agent_run_post_turn_memory.ml`, and a large `dashboard/src/components/memory-subsystems.ts` addition. Each piece is individually fine, but a reviewer has to track several unrelated concerns in one diff (the memory-subsystems dashboard surface especially looks like a distinct feature). Not a correctness issue — flagging only because wide-scope PRs are where descriptor↔handler or SSOT drift tends to sneak in unreviewed. If the conveyor auto-merges this, the memory-subsystems portion is worth a second look on its own.

Net: my cap+S_REG caveat is resolved correctly and the canonical-reader consolidation is complete; only the scope-breadth note remains, and it's non-blocking.

jeong-sik:
Rebased onto current origin/main after PR #21829 landed and resolved the dashboard projection conflicts.

Adversarial conflict resolution notes:
- Preserved both user_model and draft_skill_candidates surfaces in the backend JSON and dashboard panel instead of letting either side overwrite the other.
- Rechecked that the bounded git probe guard remains present: read_git_probe_file_opt only reads regular files up to 64 KiB before Fs_compat.load_file_opt.
- Removed the local /tmp/masc literal from the dashboard fixture and replaced it with a neutral <base-path> placeholder.

Local verification on head bc031c9fb8:
- ocamlformat --check on touched OCaml files: PASS
- pnpm --dir dashboard typecheck: PASS
- vitest memory-subsystems.focus.test.ts: PASS, 5 tests
- python3 scripts/check_test_coverage.py: PASS
- scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD: PASS, existing warnings only
- git diff --check origin/main...HEAD and worktree diff: PASS
- path/env scan for /Users, ~/me, /tmp/masc, hardcoded absolute base_path/default_base: no new hits

GitHub after push: head bc031c9fb8, mergeable=MERGEABLE, unresolved review threads=0. Checks are still queued/in-progress on the new head; PR remains draft.

anyang-keepers (commented):

NOTE: Diff truncated at 49774 chars. Reviewing visible portion (~60% of PR).

### COMMENT — TypeScript interface is well-typed

`MemorySubsystemsDraftSkillCandidate` has all fields typed: `id`, `agent_name`, `source_kind`, `source_ref`, `promotion_state`, `dir`, `json_path`, `toml_path`, `skill_md_path`, `created_at`. Good.

### COMMENT — `created_at: number | null`

Nullable `created_at` means some candidates may not have a creation timestamp. The rendering code should handle null (show 'unknown' or omit). Verify the component handles this.

### COMMENT — test coverage for visible portion

The test renders `baseResponse` which includes one draft skill candidate (`skill-candidate-repeatable-debug-loop`). It asserts:
- `container.textContent` contains the skill ID ✓
- `container.textContent` contains `candidate` (promotion_state) ✓
- Path is visible ✓

Good test for the happy path.

### COMMENT — missing edge case tests

From the visible portion, I don't see tests for:
- Empty `draft_skill_candidates` (no candidates)
- `error` field populated (backend failed to load candidates)
- `promotion_state` values other than `candidate` (e.g., `promoted`, `rejected`)

### Incomplete review due to truncation.

**Nit on edge case coverage. Not blocking — re-reviewing full diff later.**

jeong-sik:
### Change requested — head `bc031c9fb8`: index append failure is still not recoverable

The previous partial-artifact issue is mostly fixed: `write_candidate_if_changed` now checks all three generated files (`candidate.json`, `candidate.toml`, `SKILL.md`) before skipping. That closes the old "json exists but TOML/SKILL.md missing" case.

There is still one durable artifact outside that idempotency set: `index.jsonl`.

Current write order:

```ocaml
let* () = write_file json_path ...
let* () = write_file toml_path ...
let* () = write_file skill_md_path ...
let* () = append_index index_path ...
```

Current skip condition:

```ocaml
candidate_artifacts ~base_path candidate
|> List.for_all (fun (path, expected) -> ...)
```

So if `append_index` fails after the three files were written, the next post-turn write sees all generated files as unchanged and returns `Ok None` before it ever tries to append the missing index row. `list_drafts` / dashboard visibility reads only `index.jsonl`, not the draft directories, so the candidate can remain on disk but invisible forever.

This is the same class as the earlier partial-write edge, just on the final listing artifact instead of TOML/SKILL.md. Please make the index part of the recovery condition, for example:

- if files are unchanged but the latest index does not contain this candidate id/source/path, append the index row; or
- treat "unchanged files + missing index entry" as a changed candidate and call `write_candidate`; and
- add a regression where the three draft files pre-exist but `index.jsonl` is absent or missing the candidate, then `write_post_turn_candidates` repairs the listing.

`git diff --check origin/main...refs/remotes/pr/21798`: PASS. I did not run focused Dune locally in this pass.

jeong-sik:
Follow-up pushed: `b28b8d46d8` (`test(dashboard): cover draft skill candidate edge states`).

Addressed the latest dashboard edge-case review notes without changing component behavior:
- Added coverage for empty `draft_skill_candidates` plus populated `error`, asserting the alert and empty-state render together.
- Added coverage for non-`candidate` promotion states (`promoted`, `rejected`).
- Added coverage for nullable `created_at` by rendering candidates with `created_at: null` and confirming rows still render.

Local verification:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/memory-subsystems.focus.test.ts` — 7 tests passed
- `pnpm --dir dashboard typecheck` — PASS
- `python3 scripts/check_test_coverage.py` — PASS
- `git diff --check origin/main...HEAD` — PASS

GitHub after push: head `b28b8d46d8`, mergeable=MERGEABLE, checks requeued; PR remains draft.

jeong-sik:
Current head `b28b8d46d8`: the index-recovery blocker from `bc031c9fb8` is still present.

The latest push only adds dashboard component tests for empty/error/non-candidate draft states. It does not change the backend writer path:

- `write_candidate` still writes `candidate.json`, `candidate.toml`, `SKILL.md`, then appends `index.jsonl`.
- `write_candidate_if_changed` still checks only `candidate_artifacts` (`candidate.json`, `candidate.toml`, `SKILL.md`) before returning `Ok None`.
- `list_drafts` still reads only `index.jsonl` for dashboard/API visibility.

So the failure mode remains: if `append_index` fails after the three files were written, the next post-turn pass sees the files as unchanged and skips before repairing the missing index row. The candidate remains on disk but invisible to `list_drafts`.

Please keep the previous request: make the index row part of the recovery condition and add a backend regression where the three draft files pre-exist but `index.jsonl` is absent or missing that candidate.

`git diff --check origin/main...refs/remotes/pr/21798`: PASS.

jeong-sik:
## Close-out — HEAD `bc031c9fb8`: avoid local draft skill fixture paths

This directly addresses the local-path concern (test fixtures hardcoding an absolute machine-specific path). Closing it.

### [Resolved] The `/tmp/masc/...` absolute paths in the fixture are gone
The focus-test fixture swapped every hardcoded `/tmp/masc/.masc/draft-skills/...` for `<base-path>/.masc/draft-skills/...`, both in the `baseResponse` data and in the `expect(...).toContain(...)` assertions. So the test no longer asserts against (or supplies) a path that only exists on one machine — the fixture is now path-agnostic. This is the right fix for the "local path hardcoded in the test" smell; the assertion still verifies the component renders the draft-skill path suffix and structure, just without pinning it to `/tmp/masc`.

One thing to keep in mind (not a blocker): `<base-path>` is a literal placeholder string here, so the test verifies the component *displays* the placeholder verbatim. That's fine for a display-level test, but it means the test doesn't exercise any real base-path interpolation the component might do at runtime — it only checks the suffix shape. If the component ever resolves `<base-path>` to a real path in production, this fixture wouldn't catch a regression in that resolution. Acceptable for the current scope (the goal was removing the machine-specific absolute path, which is done). The follow-up commit `b28b8d46d8 test(dashboard): cover draft skill candidate edge states` broadens coverage further.

No concern on the local-path fix.

jeong-sik:
Review response pushed: `2c117a26a2` (`fix(skill): repair missing draft candidate index rows`).

Addressed the remaining index-recovery blocker from head `b28b8d46d8`:

- `write_candidate_if_changed` now skips only when both conditions hold:
  - all three durable artifacts (`candidate.json`, `candidate.toml`, `SKILL.md`) match the current candidate content, and
  - `index.jsonl` already contains a matching row for the candidate and artifact paths.
- If artifact files are complete but the index row is missing, the next post-turn projection rewrites through `write_candidate` and appends the missing index entry, making the draft visible to `list_drafts` again.
- Added a backend regression that pre-creates all three draft artifacts with no `index.jsonl`, runs `write_post_turn_candidates`, and asserts the candidate is re-indexed and visible through `list_drafts`.
- Also routed the test `read_file` helper through `Fs_compat.load_file`, closing the remaining raw-reader test helper nit.

Local verification:
- `opam exec -- ocamlformat --check lib/skill_candidate_store.ml test/test_procedural_memory.ml` PASS
- `opam exec -- ocamlc -stop-after parsing lib/skill_candidate_store.ml test/test_procedural_memory.ml` PASS
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` PASS, existing warnings only
- `git diff --check origin/main...HEAD` and `git diff --check` PASS
- diff-scoped path/raw-reader scan: changed files have no new local absolute paths; production/store path still uses `Fs_compat.load_file_opt`.

Focused Dune build was attempted but `scripts/dune-local.sh build test/test_procedural_memory.exe` refused because unrelated bare `dune build --root .` processes are active. I did not bypass the wrapper; full build/type signal remains CI-owned.

jeong-sik:
Reviewed current head `2c117a26` against the non-blocking edge-case coverage comments.

The visible UI gaps are now covered: the focus test includes an empty draft-skill list with an `error` payload, non-candidate `promotion_state` values (`promoted`, `rejected`), and `created_at: null`. The row component does not render `created_at`, so null is handled by omission rather than an invalid date string.

There is also backend/store hardening beyond the UI comments: partial artifact recovery and missing index-row repair are covered in `test_procedural_memory`. No active review thread remains; CI is still pending on the current draft head.

jeong-sik:
## Adversarial review — HEAD `2c117a26a2`: repair missing draft candidate index rows (+63/-8)

This one deserves a careful look because "repair missing rows" reads like the Repair/Sanitize workaround shape. It's not a silent read-time normalization — it's a genuine self-healing write — but it leaves the root cause untouched. Affirm on the recovery, change-requested on asking why rows go missing in the first place.

### [Affirm] The recovery is a real write, not a read-time patch
The change tightens the skip condition in `write_candidate_if_changed`:
```ocaml
let artifacts_unchanged = … (* json/toml/SKILL.md match disk *) in
let* indexed = index_contains_candidate ~base_path candidate in
if artifacts_unchanged && indexed then Ok None   (* only skip when BOTH true *)
else write_candidate …                          (* otherwise re-write, which re-indexes *)
```
Before, `artifacts_unchanged` alone gated the skip — so if the three artifact files existed on disk but the index JSONL row was missing (deleted, partially written, corrupted), the candidate was silently skipped and never appeared in `list_drafts`. Now a missing index row forces a re-write, which repairs the index. That's the correct direction: the missing data is *restored* (a real write), not masked. The new test `recovers missing index for complete artifacts` (writes all 3 artifacts, deletes/no index, then asserts re-indexing happens) locks exactly that scenario.

The error handling inside `index_contains_candidate` is also right:
```ocaml
try Ok (…) with
| Eio.Cancel.Cancelled _ as exn -> raise exn   (* async cancel propagates *)
| exn -> Error (…)                              (* everything else -> Result *)
```
That preserves `Eio.Cancel.Cancelled` (the async-cancellation exception that must not be swallowed) while collapsing ordinary I/O failures into the `Result`. Good — that's the opposite of the "absorb Cancelled" trap.

### [Change requested, MED] The root cause — non-atomic index↔artifact writes — isn't addressed
The recovery exists because the index and the three artifact files are written as **separate, non-atomic operations**, so a crash/teardown between them leaves the set inconsistent (artifacts present, index absent — the exact case this repairs). This commit makes that inconsistency *self-heal on the next write*, but it doesn't prevent it. Two questions worth answering before this hardens into the permanent shape:

1. **Why can the index row go missing while artifacts survive?** If it's a crash mid-`write_candidate`, the fix is to make the index+artifacts atomic (write all to a staging dir + `fsync` + rename, or make the index *derivable* from the artifacts so there's no second source of truth to fall behind). If the index is the ledger and the artifacts are the payload, the ledger should be the single durable record and the artifacts derived/regenerated from it — which would make `index_contains_candidate` and the whole recovery path unnecessary.
2. **Is the per-write O(n) index scan justified?** `index_contains_candidate` reads and field-compares the entire index JSONL on every `write_candidate_if_changed` call (9-field `String.equal` per row). For a small draft-skill index that's fine; if the index ever grows, this turns every candidate write into a full scan. If the index were append-only with a keyed lookup (or derived), this cost goes away.

The recovery is correct and I'd take it (it stops a real silent gap today), but "repair on write" is a workaround signature unless the non-atomicity that produces the gap is also fixed. At minimum, please note in the commit/PR whether the root cause (separate index vs artifact writes) is tracked — if this is the durable fix, an `(* index↔artifact writes are non-atomic; this reindexes on the next pass *)` comment pointing at the gap would make the intent honest, and a follow-up to make them atomic (or make the index derived) would remove the need for the scan entirely.

### [Low] `Sys.file_exists` vs `Fs_compat.file_exists`
`index_contains_candidate` uses `Sys.file_exists index_path` while the rest of this module routes existence checks through `Fs_compat`. Minor consistency point; if `Fs_compat.file_exists` adds anything (e.g., the sandbox/hook accommodation the rest of the module relies on), use it here too for uniformity.

Net: the recovery logic is sound and tested, and the `Cancelled` handling is correct. The open ask is the root-cause one — this is "repair the symptom" until the index/artifact writes are made atomic or the index is derived, at which point the repair path (and its O(n) scan) can be deleted rather than maintained.

jeong-sik:
### CHANGE REQUEST (head `2c117a26a2`) — procedure candidates ignore `~base_path` and can cross-contaminate workspaces

The index-repair fix is real, but the new post-turn procedure-candidate path still has a workspace boundary bug.

`Skill_candidate_store.write_post_turn_candidates` accepts `~base_path` and writes artifacts under that base:

- `lib/skill_candidate_store.ml:270` takes `~base_path`
- `lib/skill_candidate_store.ml:282` loads procedure candidates through `Skill_candidate_projection.top_candidates`
- `lib/skill_candidate_projection.ml:219-220` delegates to `Procedural_memory.top_procedures ~agent_name ~limit`
- `lib/procedural_memory.ml:34-42` resolves procedure storage with `Common.masc_dir_from_base_path ~base_path:(Env_config.base_path ())`

That means the fact-candidate side is scoped by the passed `base_path`, but the procedure-candidate side silently reads from the process-global `Env_config.base_path ()`. In a test, alternate workspace, or server config that passes a non-default `config.base_path`, this can read procedural memories from workspace A and write draft skill artifacts into workspace B under `.masc/draft-skills/`.

This violates the store API contract and the MASC runtime boundary: a post-turn advisory artifact must be sourced from the same base it writes to. It is especially risky because the resulting draft files look local and reviewable even though their evidence came from a different workspace.

Required fix: make the procedure read path base-scoped as well. Either add `Procedural_memory.top_procedures ~base_path ~agent_name ~limit` / path helpers that accept `base_path`, or keep procedure candidate projection out of this PR until the procedural-memory module can be made base-aware. Add a regression where `Env_config.base_path ()` points at one temp base, `write_post_turn_candidates ~base_path:<other>` runs with `~procedure_limit > 0`, and no procedures from the global base are written into the other base.

jeong-sik:
### CHANGE REQUEST (head `2c117a26a2`) — draft artifact path omits agent identity, so different keepers can overwrite each other

There is a second identity bug in the draft skill store, separate from the `~base_path` issue above.

Procedure candidates carry `agent_name`, and their `source_ref` includes the agent:

- `lib/skill_candidate_projection.ml:80` builds `procedure://<agent>/<procedure>`
- `lib/skill_candidate_projection.ml:133-136` sets `source_id = p.id` and `id = candidate_id ~source_kind ~source_id`
- `lib/skill_candidate_projection.ml:107-110` makes procedure ids as `skill-candidate-<slug source_id>`
- `lib/skill_candidate_store.ml:62-63` derives the artifact directory from `safe_component candidate.id`

So two keepers with the same procedural-memory id, for example `keeper-a` and `keeper-b` both learning `repeatable-debug-loop`, produce the same candidate id and therefore the same directory under `.masc/draft-skills/`. The second writer overwrites `candidate.json`, `candidate.toml`, and `SKILL.md` for the first writer. The index row includes `agent_name`, but the files it points at are shared, so the dashboard can show two rows whose paths resolve to only the last writer content.

The 16-hex suffix does not fix this because it hashes the same `candidate.id`; the missing entropy is the agent/source identity, not sanitizer lossiness.

Required fix: make the durable artifact key include the full source identity, at least `source_kind + agent_name + source_id` or hash `source_ref`, and add a regression that writes two procedure candidates with the same `p.id` but different `agent_name` values and asserts their `draft_dir` / artifact paths differ and both files remain readable.

jeong-sik:
### Correction / stronger CHANGE REQUEST (head `2c117a26a2`) — both Memory OS facts and procedures bypass `~base_path`

Correction to my earlier base-path comment: I said the fact-candidate side was scoped by the passed `~base_path`. It is not. The bug is broader.

Current code:

- `lib/skill_candidate_store.ml:270` accepts `~base_path`
- `lib/skill_candidate_store.ml:274` reads facts with `Keeper_memory_os_io.read_facts_tail ~keeper_id ~n:fact_tail_limit`
- `lib/keeper/keeper_memory_os_io.ml:463-464` implements `read_facts_tail` through ambient `keepers_dir ()`
- The base-scoped API already exists: `read_facts_tail_for_base_path ~base_path ~keeper_id ~n` at `keeper_memory_os_io.ml:467-471`, but this PR does not use it.
- `lib/skill_candidate_store.ml:282` reads procedures through `Skill_candidate_projection.top_candidates`, which delegates to `Procedural_memory.top_procedures`; `Procedural_memory.procedures_dir` still resolves through `Env_config.base_path ()`.

So `write_post_turn_candidates ~base_path` writes draft artifacts under the caller-supplied base, but both input sources can be read from the process-global/ambient base. This is exactly the class of path bug the store API shape is supposed to prevent: output path is explicit, input path is implicit.

Required fix is stronger than my earlier note:

1. Use `Keeper_memory_os_io.read_facts_tail_for_base_path ~base_path` for Memory OS facts.
2. Add a base-scoped procedure read path, e.g. `Procedural_memory.top_procedures_for_base_path ~base_path ~agent_name ~limit`, and thread it through `Skill_candidate_projection.top_candidates` or call the base-scoped primitive from the store.
3. Add a regression with two temp bases: seed facts/procedures in base A, call `write_post_turn_candidates ~base_path:base_b`, and assert base B does not materialize candidates from base A. Then seed base B and assert only base B candidates are written.

jeong-sik:
### Correction to artifact-identity CHANGE REQUEST (head `2c117a26a2`)

Small correction to my previous artifact-identity comment: the dashboard will not necessarily show two rows pointing at the same overwritten file. `list_drafts` runs `latest_unique`, and `latest_unique` deduplicates by `summary.id` only:

```ocaml
if Hashtbl.mem seen summary.id then false else ...
```

Because procedure candidate ids omit `agent_name`, two keepers with the same procedure id produce the same `summary.id`. So the actual failure mode is even less visible:

- durable files collide because `draft_dir` is derived from `candidate.id` only;
- index listing also collapses by `summary.id` only;
- the newer keeper can overwrite the older keeper artifacts and hide the older keeper row from `list_drafts`.

Required fix should include both parts: make the durable artifact key include full source identity (`source_kind + agent_name + source_id` or `source_ref` hash), and make listing dedup use the same full identity rather than bare `id`.

jeong-sik:
Addressed the workspace contamination change request in 0b221c8d9d.

What changed:
- `Skill_candidate_store.write_post_turn_candidates ~base_path` now reads Memory OS facts via `read_facts_tail_for_base_path` instead of ambient `read_facts_tail`.
- Procedure projection now carries `~base_path` through `Skill_candidate_projection.top_candidates` into `Procedural_memory.top_procedures`.
- `Procedural_memory` path/read/write APIs accept optional `?base_path` while preserving the existing ambient default for older callers.
- Added a regression that pins ambient `MASC_BASE_PATH` / `MASC_BASE_PATH_INPUT` to one temp workspace, calls `write_post_turn_candidates ~base_path:<other>`, and verifies ambient fact/procedure candidates do not create target drafts; then verifies a target-local procedure still does create the target draft.

Local checks run:
- `ocamlformat --check lib/procedural_memory.ml lib/procedural_memory.mli lib/skill_candidate_projection.ml lib/skill_candidate_projection.mli lib/skill_candidate_store.ml test/test_procedural_memory.ml`
- `git diff --check`

Focused Dune build was attempted with `scripts/dune-local.sh build test/test_procedural_memory.exe`, but the wrapper refused because an existing bare `dune build --root .` process is running on the machine. Leaving compile execution to PR CI per operator instruction.

jeong-sik:
### CHANGE REQUEST still open (head `0b221c8d9d`) — base-path scoping is fixed, but draft identity still collapses keepers

The workspace-contamination part is addressed in this head:

- `write_post_turn_candidates` now reads facts with `read_facts_tail_for_base_path ~base_path`.
- procedure projection now threads `~base_path` into `Procedural_memory.top_procedures`.
- the new regression pins ambient `MASC_BASE_PATH` to one temp workspace and verifies a different target base is not seeded from it.

So that blocker can close.

The remaining blocker is the artifact/listing identity. The current code still keys durable drafts and in-memory dedup on the bare candidate id:

- `lib/skill_candidate_projection.ml:107` builds procedure candidate ids from `source_kind + source_id`.
- `lib/skill_candidate_projection.ml:133-136` uses `source_id = p.id` for procedures, while `agent_name` only lives beside the id.
- `lib/skill_candidate_store.ml:62-63` derives the draft directory from `safe_component candidate.id`.
- `lib/skill_candidate_store.ml:258-265` dedups post-turn candidates by `candidate.id`.
- `lib/skill_candidate_store.ml:355-362` dedups the dashboard listing by `summary.id`.

That still means two keepers with the same procedural-memory id, e.g. `keeper-a/repeatable-debug-loop` and `keeper-b/repeatable-debug-loop`, project to the same candidate id and the same `.masc/draft-skills/<candidate-id-hash>/` directory. The later writer can replace the earlier keeper's files, and `latest_unique` can hide one row entirely because it only sees `summary.id`.

The current tests do not cover this: `test_skill_candidate_store_lists_newest_unique_draft_per_id` proves same-id rewrites collapse, but it uses one logical candidate and does not exercise same `source_id` across two different `agent_name` values.

Required fix: make the durable/listing identity include the full source identity, for example `source_kind + agent_name + source_id`, or hash `source_ref`, and use that same identity consistently for:

- candidate id or separate store key,
- `draft_dir`,
- `dedup_candidates`,
- `latest_unique`,
- a regression with two keepers sharing the same procedure id under the same base path.

jeong-sik:
## Adversarial review — HEAD `0b221c8d9d`: scope post-turn candidates to base path (workspace isolation fix)

**Strong affirm.** This fixes a real cross-workspace data-leak bug, and the fix is the correct one — thread `base_path` through the read paths so they stop falling back to the global env default. This is a Path/SSOT issue at its core and it's resolved properly.

### [Affirm] The leak: write took `base_path`, reads silently used the global env base
Before, `write_post_turn_candidates ~base_path:target_base` wrote drafts *into* `target_base`, but the two reads it performed went to the global env base:
- `Keeper_memory_os_io.read_facts_tail ~keeper_id ~n:fact_tail_limit` — no `base_path`, so `Env_config.base_path()` (global).
- `Skill_candidate_projection.top_candidates ~agent_name ~limit` → `Procedural_memory.top_procedures ~agent_name ~limit` → `procedures_dir ~agent_name` hardcodes `Env_config.base_path ()`.

So if `MASC_BASE_PATH` pointed at a global workspace A, asking to write candidates into workspace B still seeded B's drafts from A's facts and A's procedures. That's silent cross-workspace contamination — a workspace B operator would see skill drafts derived from workspace A's private lessons/procedures, with no error. The fix threads `?base_path` through every path function (`procedures_dir`, `procedures_path`, `load/save/rewrite_procedures`, `top_procedures`, `top_candidates`, `with_procedures_lock`) and makes the candidate entry point require it:
```ocaml
let top_candidates ~base_path ~agent_name ~limit =           (* required, not optional *)
  Procedural_memory.top_procedures ~base_path ~agent_name ~limit |> …
```
so the post-turn path can't accidentally fall back to the env default. `base_path_or_default` keeps `None -> Env_config.base_path ()` as a default for *existing* callers that don't supply one — reasonable for backward compatibility, see the caveat below.

### [Affirm] The test reproduces the leak and proves the fix
`test_skill_candidate_store_scopes_post_turn_reads_to_base_path` is exactly the right shape: it sets `MASC_BASE_PATH`/`MASC_BASE_PATH_INPUT` to `global_base`, seeds `global_base` with a fact and a procedure, then calls `write_post_turn_candidates ~base_path:target_base` and asserts `0` candidates leaked. Then it writes a procedure *into* `target_base` and asserts that one *does* produce a target draft. That's a bidirectional proof — global doesn't leak in, target-only does appear — which is what an isolation guard should lock. The switch from `Memory_io.For_testing.with_keepers_dir` (which mutates global state) to an explicit `write_facts_for_base_path ~base_path` helper also makes the test deterministic per-base-path rather than implicitly process-global.

### [Low] The lower-level `Procedural_memory` functions stay `?base_path` optional
`top_candidates` was promoted to required `base_path`, but the functions beneath it (`top_procedures`, `load_procedures`, `save_procedure`, `rewrite_procedures`) remain `?base_path` with the `Env_config.base_path ()` default. That's fine for the post-turn candidate path (now forced through `top_candidates`), but any *other* caller of `Procedural_memory` that omits `~base_path` still reads/writes the global env base — so the leak surface is closed for candidates but the underlying primitives can still be called in the old env-defaulting way. If there are non-candidate callers that should also be workspace-scoped, they'd silently keep using the global base unless individually updated. Non-blocking for this fix (the candidate path — the actual reported leak — is closed and tested); just flagging that the `?base_path` default on the primitives means the isolation is enforced at the `top_candidates` layer, not at the storage layer.

### [Still open, separate topic] The #21798 root-cause MED from my `2c117a26a2` review is not addressed by this commit
This commit is the workspace-isolation fix, not the index↔artifact non-atomicity I flagged earlier. The "repair missing index rows" path from `2c117a26a2` still relies on the index and artifacts being written as separate non-atomic operations; this commit doesn't change that. So that MED (make index+artifacts atomic, or derive the index, so the repair-on-write path can be deleted) remains open and is independent of this scoping fix. Not a regression — just keeping both threads visible.

Net: the workspace leak is a genuine Path/SSOT bug and this is the correct structural fix (base_path threading + required at the entry point + a leak-reproducing test). I'd merge; the two notes (optional primitives, separate atomicity MED) are forward-looking, not blockers.

jeong-sik:
Addressed the artifact-identity change request in 6e233d15be.

What changed:
- Durable draft artifact directories are now keyed by full source identity (`source_kind + agent_name + source_ref`) while keeping the existing display `candidate.id` stable.
- `write_post_turn_candidates` in-memory dedup now uses the same source identity instead of bare `candidate.id`.
- `list_drafts` latest-unique collapse now dedups by source identity instead of `summary.id`, so two keepers with the same procedure id remain visible as distinct rows.
- Added a regression where two procedures from `alpha` and `beta` intentionally produce the same `candidate.id`; the test verifies distinct draft dirs/json paths remain readable and both source agents appear in the listing.
- Documented `draft_dir` as source-identity keyed in the store interface.

Local checks run after this commit:
- `ocamlformat --check lib/skill_candidate_store.ml lib/skill_candidate_store.mli test/test_procedural_memory.ml`
- `git diff --check`

Focused Dune build remains blocked locally by the existing bare `dune build --root .` process, so compile/test execution is left to PR CI per operator instruction.

jeong-sik:
## Close-out — HEAD `6e233d15be`: draft artifact identity now uses full source identity

The remaining identity blocker is addressed on this head.

Confirmed in the diff:

- `draft_dir` no longer hashes only `candidate.id`; it uses `source_kind + agent_name + source_ref` as the identity key while preserving `candidate.id` as the display id.
- `dedup_candidates` now dedups by that candidate source identity, not bare id.
- `latest_unique` now dedups listing rows by `source_kind + agent_name + source_ref`, not `summary.id`.
- The new regression intentionally creates two procedures from `alpha` and `beta` with the same display candidate id and asserts distinct draft dirs/json paths plus both agents visible in `list_drafts`.

That closes the overwrite/hidden-row failure mode I flagged. `git diff --check origin/main...refs/remotes/pr/21798` is clean locally. CI is still running on the new head, so readiness still depends on those checks, but I do not see a remaining code blocker for the base-path or artifact-identity findings.

jeong-sik:
## Adversarial review — HEAD `6e233d15be`: key draft artifacts by source identity (+18/-7)

**Strong affirm.** This fixes a real cross-keeper/cross-source overwrite bug, and it does it by promoting the dedup key to source identity in one place and routing both the artifact directory and the in-memory dedup through it.

### [Affirm] The bug: same display id across keepers collided
Before, the artifact directory and dedup were keyed on `candidate.id`:
```ocaml
let safe_component raw = component_prefix raw ^ "-" ^ component_hash raw
let draft_dir ~base_path candidate = … (safe_component candidate.id)
… dedup_candidates … Hashtbl.add seen candidate.id () …
```
`candidate.id` is a display slug, not a uniqueness key — two keepers (or two sources) projecting a candidate with the same id would write the same `draft_dir` and clobber each other's `candidate.json`/`SKILL.md`, and dedup would drop one of them in memory. Silent overwrite. The mli docstring now states the fix directly: "The directory key includes source identity, not just display [id], so different keepers/sources with the same candidate id do not overwrite each other."

### [Affirm] One identity key, used by both the artifact path and the dedup
```ocaml
let candidate_identity_key candidate =
  String.concat "\n" [ candidate.source_kind; candidate.agent_name; candidate.source_ref ]
let candidate_component candidate =
  component_prefix candidate.id ^ "-" ^ component_hash (candidate_identity_key candidate)
```
`draft_dir` uses `candidate_component` (readable id prefix + identity hash, so two same-id candidates land in different dirs), and `dedup_candidates` / `latest_unique` switch from `candidate.id` to `candidate_identity_key`. So the on-disk layout and the in-memory dedup agree on what "same candidate" means — no more "dedup kept it but disk clobbered it" mismatch. `summary_identity_key` mirrors it for the listing path. That's the right SSOT shape: one definition of identity, two consumers updated together.

### [Low] The `\n` separator assumes no newline in the identity fields
`String.concat "\n" [source_kind; agent_name; source_ref]` is collision-free only if none of those three fields contain `\n`. `source_kind` is a closed-ish enum and `agent_name` is a keeper id, so those are safe; `source_ref` is the one to watch — if it's ever free-form text (rather than the normalized `procedure://keeper/…` / fact referent it appears to be), a newline could synthesize a colliding key. Almost certainly fine given the current normalized refs, but if `source_ref` is ever widened to arbitrary text, the separator needs to be a value that can't appear in any field (or use length-prefixed encoding). Non-blocking today.

### [Low / verify] Existing draft directories are orphaned by the key change
Switching the directory name from `id-hash` to `id-identity_hash` means drafts already on disk under the old keying won't be found by the new keying. Drafts are regenerable (re-derived on each post-turn pass), so this isn't data loss — but any draft that *isn't* re-derived on the next pass (e.g. a promoted/crystallized one that's no longer in the candidate set) becomes an orphan on disk until cleanup. Worth confirming whether a one-time migration or an orphan-sweep is intended, or whether the regeneration coverage is complete enough that orphans are guaranteed not to matter. Non-blocking; just don't want a silently-orphaned promoted skill.

### [Still open, separate topic] The index↔artifact non-atomicity MED from my `2c117a26a2` review
This commit is the source-identity keying fix, not the atomic-write / derive-index fix. The "repair missing index rows" path still relies on the index and artifacts being written separately; the workspace-isolation commit and this one are both adjacent improvements but neither touches that. Keeping that MED open as its own thread.

Net: correct collision fix with a single identity key shared by artifact path and dedup, plus tests using a per-test `agent_name` so cross-keeper scenarios are exercisable. I'd merge; the two Lows (newline-separator safety, orphaned old-key dirs) and the separate atomicity MED are forward-looking.

jeong-sik:
Addressed the CI Build/Lint failures with 4f1a825db0.

Root cause from CI logs:
- `Procedural_memory` added `?base_path` to functions whose optional argument could not be erased under `-warn-error +a` (`warning 16`).
- `test/test_procedural_memory.ml` referenced `Masc.Config_dir_resolver`, which is not exposed through the public `Masc` alias set in this build context.

Fix:
- Added `unit` terminators to the public procedural-memory functions that carry `?base_path` with only labelled arguments.
- Added `Keeper_memory_os_io.rewrite_facts_atomically_for_base_path` so tests can write request/base-path scoped facts through a public IO API instead of depending on internal config resolver exposure.
- Updated skill candidate projection and tests for the adjusted API.

Local verification:
- `scripts/dune-local.sh build test/test_procedural_memory.exe` -> passed
- `./_build/default/test/test_procedural_memory.exe` -> 20 passed
- `ocamlformat --check lib/procedural_memory.ml lib/procedural_memory.mli lib/skill_candidate_projection.ml lib/keeper/keeper_memory_os_io.ml lib/keeper/keeper_memory_os_io.mli test/test_procedural_memory.ml` -> passed
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` -> passed

jeong-sik:
Follow-up on the remaining draft index/artifact review note in `751432b7b0`.\n\nWhat changed:\n- Replaced the two production `Sys.file_exists` checks in `Skill_candidate_store` with `Fs_compat.file_exists` for the index read paths.\n- Added an explicit comment at `write_candidate_if_changed` documenting the current recovery contract: artifact files and index membership are checked together because a prior write can leave complete artifacts but miss the final index append, and the next post-turn pass should repair the listing row.\n\nScope note:\n- I did not attempt a broad atomic staging-dir rewrite in this PR. The current code already has the missing-index repair regression (complete artifacts + absent index => re-index) and this follow-up makes that recovery path intentional and consistent with Fs_compat. A future derived-index or fully atomic index+artifact write can delete this repair path, but that is a larger storage-shape change than a review-response patch.\n\nValidation:\n- `env DUNE_BUILD_DIR=_build-review-21798 DUNE_CACHE=disabled dune build --root . test/test_procedural_memory.exe`\n- `./_build-review-21798/default/test/test_procedural_memory.exe` — 20 tests passed\n- `ocamlformat --check lib/skill_candidate_store.ml`\n- `python3 scripts/check_test_coverage.py`\n- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD`\n- `git diff --check`\n- `git diff --check origin/main...HEAD`

jeong-sik:
## Close-out (partial) — HEAD `751432b7b0`: document draft index repair path (+6/-2)

This addresses two of the three things from my `2c117a26a2` review: the honesty comment I asked for as a fallback, and the `Sys.file_exists` consistency Low. The root cause itself (non-atomic index/artifact writes) is still not fixed and there's no follow-up tracking it, so I'm closing the honesty + consistency points and keeping the root-cause ask open at a lower severity.

### [Resolved] The non-atomicity gap is now documented honestly
```ocaml
let write_candidate_if_changed ~base_path candidate =
  (* Artifact files and the index are intentionally checked together: a prior
     write can leave complete artifacts but miss the final index append. The
     next post-turn pass should repair that listing row instead of treating the
     draft as fully persisted. *)
```
That's exactly the "comment pointing at the gap" I asked for as the minimum bar if the atomic fix wasn't going to land in this PR. A reader now knows *why* the index-must-contain check exists (because the two writes aren't atomic and a crash between them leaves artifacts-without-index), and that the repair-on-next-pass is intentional, not accidental. The intent is honest now.

### [Resolved] `Sys.file_exists` → `Fs_compat.file_exists` (the consistency Low)
Both `index_contains_candidate` and `list_drafts` now route existence checks through `Fs_compat`, matching the rest of the module. Closing that Low.

### [Still open, downgraded MED → LOW-MED] The root cause isn't fixed and isn't tracked
The honesty comment makes the *current* behavior understandable, but it doesn't change the shape: index and artifacts are still written as separate non-atomic operations, the repair-on-write path (with its per-write O(n) index scan) is still doing the work, and — the part I'd flag now — there's no follow-up issue or RFC referenced that tracks making them atomic (or deriving the index). My earlier ask named two acceptable outcomes: (a) make the writes atomic / derive the index, or (b) if this is the durable shape, label it as a tracked workaround with a removal target. This commit does the documentation half of (b) but not the tracking half — no issue link, no "removal target," so there's nothing stopping the repair-on-write path from quietly becoming the permanent design rather than a safety net.

If you agree the atomic fix is the real goal, the smallest thing that would let me fully close this is a tracking issue (referenced in the comment, e.g. `(* see #NNNN — make index+artifacts atomic so this repair pass can be deleted *)`) so the repair path is visibly temporary. Without that, I'll leave it as a non-blocking LOW-MED rather than holding the merge: the gap is documented, the symptom is handled, but the root is deferred without a paper trail.

Net: honesty + consistency resolved; root cause deferred and currently untracked — please drop an issue link so the repair-on-write path is marked as a safety net with an owner, not a permanent design.

jeong-sik:
Addressed the remaining LOW-MED tracking ask in `c869d228c4`.

What changed:
- Created #21871 to track the root-cause follow-up: make draft skill index/artifact persistence atomic or derive one side from the other so the repair-on-write scan can eventually be removed.
- Updated the `write_candidate_if_changed` recovery comment to reference #21871 and explicitly mark the current index/artifact consistency check as a safety net, not the permanent storage design.

Validation:
- `opam exec -- ocamlformat --check lib/skill_candidate_store.ml`
- `git diff --check`

No local build run; leaving build/test execution to CI per operator instruction.

jeong-sik:
## Close-out — HEAD `c869d228c4`: track draft index atomicity follow-up — MED fully resolved

This closes the root-cause thread. The repair-on-write path is now labeled as a temporary safety net with an owner, which is exactly the tracking half I asked for in `751432b7b0`. Resolved.

### [Resolved] Safety-net labeling + follow-up owner
```ocaml
(* … This is a safety net until #21871 makes the
   index/artifact persistence atomic or derives one side from the other. *)
```
Three things land in two lines:
- **"safety net until #21871"** — the repair-on-write path is now explicitly *temporary*, not the permanent design. A future reader can't mistake it for intentional architecture.
- **#21871 (issue link)** — the root cause (non-atomic index/artifact writes) now has a tracked owner, so it's not just "documented and forgotten."
- **"makes the persistence atomic or derives one side from the other"** — the follow-up's *goal* is stated (atomic write, or derive the index from artifacts), so when #21871 lands the repair path + its per-write O(n) scan can be deleted, not just tweaked.

That satisfies both halves of what I asked across the two reviews: `751432b7b0` made the gap honest, this commit makes the fix *tracked* with a removal target. The workaround is now a properly-labeled safety net (documented gap + issue owner + stated end state), which is the acceptable outcome when the atomic fix is deferred rather than landed in this PR.

### Net
The MED from `2c117a26a2` (root cause not fixed) is fully resolved as a *tracked* deferral: behavior unchanged, but the repair path is marked temporary and #21871 owns the atomic/derivation fix. No remaining concern from me on the 21798 atomicity thread — closing it. The follow-up to actually watch is #21871 (atomic write or index-derivation, at which point `index_contains_candidate` / `replace`-on-write can be deleted).

jeong-sik:
### CHANGE REQUEST follow-up (head `6e233d15be`) - CI now shows concrete compile blockers

The artifact identity blocker is fixed, but this head is still not mergeable. GitHub CI fails in `Health`, `Build and Test`, and `Lint` with PR-specific compile errors.

#### 1. `?base_path` on fully-labelled functions trips OCaml warning 16 under `-warn-error`

CI `Lint / Build with warnings as errors`:

```text
File "lib/procedural_memory.ml", line 39, characters 20-29:
let procedures_dir ?base_path ~agent_name =
                    ^^^^^^^^^
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "lib/procedural_memory.ml", line 45, characters 21-30:
let procedures_path ?base_path ~agent_name =
                     ^^^^^^^^^

File "lib/procedural_memory.ml", line 98, characters 21-30:
let load_procedures ?base_path ~agent_name : procedure list =
                     ^^^^^^^^^

File "lib/procedural_memory.ml", line 149, characters 20-29:
let top_procedures ?base_path ~agent_name ~limit : procedure list =
                    ^^^^^^^^^
```

The intent is right, but this API shape is not accepted by the repo's OCaml 5.4 warning policy. Since these functions have no final positional argument, the optional argument cannot be erased reliably.

Recommended fix: keep the old ambient APIs unchanged and add explicit base-path variants, matching the local pattern already used by `Keeper_memory_os_io`:

- `procedures_path ~agent_name`
- `procedures_path_for_base_path ~base_path ~agent_name`
- `load_procedures ~agent_name`
- `load_procedures_for_base_path ~base_path ~agent_name`
- `top_procedures ~agent_name ~limit`
- `top_procedures_for_base_path ~base_path ~agent_name ~limit`

Then have `Skill_candidate_projection.top_candidates ~base_path` call the explicit `_for_base_path` variant.

#### 2. `test_procedural_memory.ml` references a non-exported public alias

CI `Build and Test / Build` and `Health / Run health snapshot`:

```text
File "test/test_procedural_memory.ml", line 9, characters 21-45:
module Config_dirs = Masc.Config_dir_resolver
                     ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module Masc.Config_dir_resolver
```

Other tests use `Config_dir_resolver` directly, not through `Masc`. Either use the direct module in the test, or avoid the resolver alias entirely by using the existing `Keeper_memory_os_io.For_testing.with_keepers_dir` / `rewrite_facts_atomically_for_keepers_dir` path you already depend on.

Until these compile blockers are fixed, my earlier code-path close-out remains valid but the PR is still a hard merge blocker.

jeong-sik:
## Close-out - HEAD `c869d228c4`: CI compile blockers addressed

My previous CI blocker comment was against stale head `6e233d15be`. Rechecked current head `c869d228c4`.

Confirmed:

- `Procedural_memory` functions with `?base_path` and no positional payload now take a final `unit`, so OCaml warning 16 no longer applies to the `procedures_path`, `load_procedures`, and `top_procedures` APIs.
- `Keeper_memory_os_io.rewrite_facts_atomically_for_base_path` is now exposed and implemented, so `test_procedural_memory.ml` no longer needs the non-existent `Masc.Config_dir_resolver` alias.
- `Skill_candidate_projection.top_candidates` calls the adjusted `Procedural_memory.top_procedures ... ()` API.

Local verification:

- `git diff --check origin/main...refs/remotes/pr/21798` - PASS

No remaining blocker from my CI follow-up.

PR #21825, opened 2026-06-20
feat(dashboard): categorized top-bar AttentionIndicator (keeper-v2 Unit 3)

jeong-sik:
## What

Unit 3 of the keeper-v2 dashboard port: the top-bar **AttentionIndicator** (keeper-v2 `shell.jsx`).

A `⚑ 주의 N` chip opens a dropdown of the non-empty attention categories — each navigating to the surface where the operator can act — with a `✓ 정상` zero-state. It replaces the flat `Attention N → overview` badge that lived inside `ConnectionStatus`.

## Changes

- **`components/attention-indicator.ts`** — new component + a pure `summarizeAttention` / `attentionItemBucket` core. Reads the live `missionSnapshot.attention_queue`.
- **`app.ts`** — renders `<AttentionIndicator />` in the top-bar header actions.
- **`components/dashboard-shell.ts`** — removes the flat `Attention N` badge from `ConnectionStatus` (now transport-only).
- **`styles/app-shell-v2.css`** — adds the `.v2-statchip.bad` tone and `.v2-statchip.attn` button; the dropdown uses inline Tailwind tokens.
- **`docs/design/dashboard-attention-indicator.md`** — the design, the data-grounding decision, and follow-ups.

## Data grounding (categories reflect the real queue, not the prototype's mock)

The prototype mocked four fixed buckets (승인 대기 / 주의 keeper / 죽음·넘침 / stale 게이트) from a pre-aggregated object. The real `attention_queue` (`lib/operator/operator_digest.ml`) only carries `target_type ∈ {workspace, keeper}`, `severity ∈ {critical, bad, warn}`, and `kind` (`pending_confirm_waiting`, `keeper_<reason>`, tool-host failures) — there is no gate/stale signal and no kind that separates 죽음·넘침 from other keeper attention.

So the indicator categorizes by the **real, closed vocabulary** with exact equality and an explicit `other` bucket (no item is silently dropped; bucket counts sum to the queue length):

| Bucket | Rule | Label | → |
|---|---|---|---|
| approvals | `kind === 'pending_confirm_waiting'` | 승인 대기 | `approvals` |
| keepers | `target_type === 'keeper'` | 주의 keeper | `keepers` |
| other | else | 기타 | `overview` |

Per-bucket tone is the worst severity in the bucket, so dead/overflow keepers still surface under 주의 keeper with a `bad` tone rather than being lost. The prototype's 죽음·넘침/stale buckets are intentionally **not fabricated** from a `keeper_*` prefix or empty data — when the backend tags those distinctly in the queue, the buckets are a one-line addition (documented in the design note).

## Verification

- `pnpm test attention-indicator` — 13 tests (pure bucket rules + total invariant + tone + order + empty; `BUCKET_META` nav targets; component zero-state, chip tone/total, dropdown open, per-row navigate).
- `dashboard-shell` + `app` suites pass unchanged after the badge removal (69 tests green across the three files).
- `pnpm typecheck` clean; `pnpm eslint` clean on the changed component.

### DIFF
--- a/dashboard/src/app.ts
+++ b/dashboard/src/app.ts
@@ -30,6 +30,7 @@ import {
 } from './components/dashboard-shell'
 import { ThemeSwitch } from './components/theme-switch'
 import { EmergencyStopControl } from './components/emergency-stop-control'
+import { AttentionIndicator } from './components/attention-indicator'
 import { TransportBeacon } from './components/transport-beacon'
 import { DashboardNavRail } from './components/mobile-nav'
 import { SkipLink } from './components/skip-link'
@@ -344,6 +345,7 @@ export function App() {
                 scheduler <b>${serverStatus.value ? 'healthy' : '—'}</b>
               </span>
             </div>
+            <${AttentionIndicator} />
             <span class="contents" data-mobile-detail-keep><${EmergencyStopControl} /></span>
             <${CopilotDockTopBarButton} dock=${dock} />
             <${Suspense} fallback=${authStatusFallback()}>
--- a/dashboard/src/components/attention-indicator.test.ts
+++ b/dashboard/src/components/attention-indicator.test.ts
@@ -1,0 +1,162 @@
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { html } from 'htm/preact'
+import { render } from 'preact'
+import { waitFor } from '@testing-library/preact'
+
+vi.mock('../router', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../router')>()),
+  navigate: vi.fn(),
+}))
+import { navigate } from '../router'
+
+import { missionSnapshot } from '../mission-signals'
+import {
+  AttentionIndicator,
+  attentionItemBucket,
+  summarizeAttention,
+  BUCKET_META,
+} from './attention-indicator'
+import type {
+  DashboardMissionAttentionQueueItem,
+  DashboardMissionResponse,
+} from '../types/dashboard-mission'
+
+function item(
+  partial: Partial<DashboardMissionAttentionQueueItem>,
+): DashboardMissionAttentionQueueItem {
+  return {
+    id: 'i',
+    kind: 'unknown_kind',
+    severity: 'warn',
+    summary: '',
+    target_type: 'workspace',
+    related_session_ids: [],
+    related_agent_names: [],
+    evidence_preview: [],
+    ...partial,
+  }
+}
+
+function setQueue(items: DashboardMissionAttentionQueueItem[]): void {
+  missionSnapshot.value = { attention_queue: items } as unknown as DashboardMissionResponse
+}
+
+describe('attentionItemBucket', () => {
+  it('routes the approval kind to approvals (by kind)', () => {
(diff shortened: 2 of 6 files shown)


jeong-sik:
Review-response follow-up pushed in 8858cdb749.\n\nWhat changed:\n- Added Escape dismissal for the categorized AttentionIndicator dropdown, matching the dashboard's other popover behavior.\n- Added component coverage for outside-click close and Escape close; switched interaction assertions to Testing Library fireEvent so Preact state updates are exercised through the test harness.\n- Corrected the design note date to 2026-06-20 and narrowed the verification text to the checks actually represented by this PR slice.\n\nLocal checks run:\n- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/attention-indicator.test.ts --no-file-parallelism --maxWorkers=1\n- pnpm --dir dashboard exec eslint src/components/attention-indicator.ts\n- pnpm --dir dashboard typecheck\n- git diff --check origin/main...HEAD

jeong-sik:
## Adversarial review — categorized top-bar AttentionIndicator (HEAD `fda4c21e637c`)

**Affirm — and notably, this component goes out of its way to satisfy the audit's hard rules.** Calling those out since they're exactly the criteria in scope.

### [Affirm] Explicitly refuses the string-classifier anti-pattern
The header comment names RFC-0042 directly: inventing buckets from a `keeper_*` kind prefix "would be a fragile substring classifier", so categorization uses the real closed server vocabulary. `attentionItemBucket` matches by **exact equality** on documented constants:
- `APPROVAL_KIND = 'pending_confirm_waiting'` // `operator_digest.ml:77`
- `KEEPER_TARGET_TYPE = 'keeper'` // `operator_digest.ml:186,245`
- `BAD_SEVERITIES = {critical, bad}` // `operator_digest_types.ml:5-7`

No `startsWith` / substring routing anywhere. This is the right call and the provenance comments trace each constant to its OCaml SSOT location.

### [Affirm] No-silent-drop is an enforced invariant, not a hope
`attentionItemBucket` returns an explicit `'other'` for anything unmatched, and `summarizeAttention`'s comment asserts `total === items.length`. So a future backend attention kind can't vanish from the indicator — worst case it lands in `기타` and still counts. Dead/overflow keepers still surface under `주의 keeper` with a `bad` tone. This is the correct way to handle an open backend vocabulary from a closed frontend type.

### [Affirm] Pure core + correct effect
`summarizeAttention` / `attentionItemBucket` are pure (testable, no signal mutation). The `useEffect` is the legitimate kind — syncing a DOM outside-click + Escape listener while open, with proper cleanup — not a data-derivation or prop-mirror effect. `BUCKET_ORDER` / `BUCKET_META` are readonly, fixed display order.

### [Low / non-binding] Frontend↔backend vocabulary sync is comment-traced, not generated
`APPROVAL_KIND`, `KEEPER_TARGET_TYPE`, `BAD_SEVERITIES` are hand-maintained TS literals that must match the OCaml enums; a backend rename silently demotes the affected items to `other` (no type error, no drop — just miscategorization). The `other` bucket makes this *safe*, so this is not blocking. If you ever want compile-time enforcement, generating the TS constants from the OCaml `operator_digest` types would close the gap — but the current safe-fallback design is acceptable for production.

PR #21829, opened 2026-06-20
feat(voice): expose turn loop guidance

jeong-sik:
## Summary
- expose a structured `voice_loop` contract in voice session JSON and `keeper_voice_agent`
- add an env-gated `realtime_bridge` conversation mode for voice sessions via `MASC_VOICE_REALTIME_WS_URL`
- make `keeper_voice_session_start conversation_mode=realtime_bridge` fail closed when no realtime bridge is configured, instead of pretending a duplex stream exists
- align voice tool descriptions with the new mode contract: default `turn_based` batch STT/TTS, optional bridge-backed realtime transport

## Validation
- `ocamlformat --check lib/voice/voice_session_manager.mli lib/voice/voice_session_manager.ml lib/keeper/keeper_tool_voice_runtime.ml lib/tool_surface/tool_shard_types_schemas_voice.ml test/test_voice_runtime_overlay.ml`
- `git diff --check`
- `ocamlc -stop-after parsing lib/voice/voice_session_manager.mli`
- `ocamlc -stop-after parsing lib/voice/voice_session_manager.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_tool_voice_runtime.ml`
- `ocamlc -stop-after parsing lib/tool_surface/tool_shard_types_schemas_voice.ml`
- `ocamlc -stop-after parsing test/test_voice_runtime_overlay.ml`
- `env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_voice_runtime_overlay.exe`
- `./_build/default/test/test_voice_runtime_overlay.exe`

### DIFF
--- a/lib/keeper/keeper_tool_voice_runtime.ml
+++ b/lib/keeper/keeper_tool_voice_runtime.ml
@@ -295,14 +295,7 @@ let voice_agent_capability_fields ~(meta : keeper_meta) =
   ; ( "realtime_supported"
     , `Bool (realtime_configured || Voice_session_manager.realtime_supported active_mode) )
   ; ( "realtime_bridge"
-    , `Assoc
-        [ "configured", `Bool realtime_configured
-        ; "required_env", `String Voice_session_manager.realtime_bridge_env
-        ; ( "endpoint"
-          , match realtime_endpoint with
-            | Some endpoint -> `String endpoint
-            | None -> `Null )
-        ] )
+    , Voice_session_manager.realtime_bridge_public_json ?endpoint:realtime_endpoint () )
   ; ( "available_conversation_modes"
     , `List
         ([ `String "turn_based" ]
--- a/lib/keeper/keeper_tool_voice_runtime.ml
+++ b/lib/keeper/keeper_tool_voice_runtime.ml
@@ -248,15 +248,69 @@ let append_assoc_fields json fields =
   | `Assoc existing -> `Assoc (existing @ fields)
   | other -> `Assoc (("voice_config", other) :: fields)
 
+let requested_conversation_mode ~(args : Yojson.Safe.t) =
+  match
+    Safe_ops.json_string_opt "conversation_mode" args
+    |> Option.map (fun raw -> String.lowercase_ascii (String.trim raw))
+  with
+  | None | Some "" | Some "turn_based" | Some "turn-based" ->
+    Ok Voice_session_manager.Turn_based
+  | Some ("realtime" | "realtime_bridge" | "realtime-bridge") ->
+    (match Voice_session_manager.realtime_bridge_endpoint () with
+     | Some endpoint -> Ok (Voice_session_manager.Realtime_bridge { endpoint })
+     | None ->
+       Error
+         (Tool_args.error_response_with
+            [ "message", `String "voice realtime bridge unavailable"
+            ; "error", `String "voice_realtime_bridge_unavailable"
+            ; "requested_conversation_mode", `String "realtime_bridge"
+            ; "required_env", `String Voice_session_manager.realtime_bridge_env
+            ; "fallback_conversation_mode", `String "turn_based"
+            ; "fallback_tool", `String "keeper_voice_session_start"
+            ]))
+  | Some mode ->
+    Error
+      (Tool_args.error_response_with
+         [ "message", `String "invalid voice conversation_mode"
+         ; "error", `String "invalid_voice_conversation_mode"
+         ; "conversation_mode", `String mode
+         ; ( "accepted_modes"
+           , `List [ `String "turn_based"; `String "realtime_bridge" ] )
+         ])
+
 let voice_agent_capability_fields ~(meta : keeper_meta) =
   let mgr = Keeper_voice_local.get_session_manager () in
   let active_session = Voice_session_manager.get_session mgr ~agent_id:meta.name in
-  [ "conversation_mode", `String "turn_based"
-  ; "transport_mode", `String "batch_stt_tts"
-  ; "realtime_supported", `Bool false
+  let active_mode =
+    match active_session with
+    | Some session -> Voice_session_manager.session_conversation_mode session
+    | None -> Voice_session_manager.Turn_based
+  in
+  let realtime_endpoint = Voice_session_manager.realtime_bridge_endpoint () in
+  let realtime_configured = Option.is_some realtime_endpoint in
+  [ ( "conversation_mode"
+    , `String (Voice_session_manager.string_of_conversation_mode active_mode) )
+  ; ( "transport_mode"
+    , `String (Voice_session_manager.transport_mode_of_conversation_mode active_mode) )
+  ; ( "realtime_supported"
+    , `Bool (realtime_configured || Voice_session_manager.realtime_supported active_mode) )
+  ; ( "realtime_bridge"
+    , `Assoc
+        [ "configured", `Bool realtime_configured
+        ; "required_env", `String Voice_session_manager.realtime_bridge_env
+        ; ( "endpoint"
+          , match realtime_endpoint with
+            | Some endpoint -> `String endpoint
+            | None -> `Null )
+        ] )
+  ; ( "available_conversation_modes"
+    , `List
+        ([ `String "turn_based" ]
+         @ if realtime_configured then [ `String "realtime_bridge" ] else []) )
   ; ( "voice_loop"
-    , Voice_session_manager.turn_based_voice_loop_json
-        ~session_active:(Option.is_some active_session) )
+    , Voice_session_manager.voice_loop_json
+        ~session_active:(Option.is_some active_session)
+        active_mode )
   ; "session_active", `Bool (Option.is_some active_session)
   ; ( "active_session"
     , match active_session with
@@ -307,9 +361,15 @@ let handle_session_start ~(meta : keeper_meta) ~(args : Yojson.Safe.t) =
     | Some s when s <> "" -> Some s
     | _ -> None
   in
-  let mgr = Keeper_voice_local.get_session_manager () in
-  let session = Voice_session_manager.start_session mgr ~agent_id:meta.name ?voice () in
-  Yojson.Safe.to_string (Voice_session_manager.session_to_json session)
+  match requested_conversation_mode ~args with
+  | Error response -> response
+  | Ok conversation_mode ->
+    let mgr = Keeper_voice_local.get_session_manager () in
+    let session =
+      Voice_session_manager.start_session mgr ~agent_id:meta.name ?voice
+        ~conversation_mode ()
+    in
+    Yojson.Safe.to_string (Voice_session_manager.session_to_json session)
 
 let handle_session_end ~(meta : keeper_meta) =
   let mgr = Keeper_voice_local.get_session_manager () in
--- a/lib/keeper/keeper_tool_voice_runtime.ml
+++ b/lib/keeper/keeper_tool_voice_runtime.ml
@@ -252,7 +252,11 @@ let voice_agent_capability_fields ~(meta : keeper_meta) =
   let mgr = Keeper_voice_local.get_session_manager () in
   let active_session = Voice_session_manager.get_session mgr ~agent_id:meta.name in
   [ "conversation_mode", `String "turn_based"
+  ; "transport_mode", `String "batch_stt_tts"
   ; "realtime_supported", `Bool false
+  ; ( "voice_loop"
+    , Voice_session_manager.turn_based_voice_loop_json
+        ~session_active:(Option.is_some active_session) )
   ; "session_active", `Bool (Option.is_some active_session)
   ; ( "active_session"
     , match active_session with
--- a/lib/keeper/keeper_turn.ml
+++ b/lib/keeper/keeper_turn.ml
@@ -709,6 +709,7 @@ let run_keeper_msg_turn_admitted ?on_text_delta ?on_event ctx args : tool_result
                 let tool_use_lines = [
                   "Tool-use guidance:";
                   "- If the user asks you to speak, use voice, make sound, or output TTS, prefer keeper_voice_session_start and keeper_voice_speak.";
+                  "- Voice sessions are turn-based: operator speech arrives as transcribed text through normal keeper turns; do not wait for a live duplex audio stream.";
                   "- Do not simulate spoken audio with plain text roleplay when a voice tool can handle the request.";
                   "- If voice execution fails, say that voice output is unavailable and continue in text.";
                 ] in
--- a/lib/tool_surface/tool_shard_types_schemas_voice.ml
+++ b/lib/tool_surface/tool_shard_types_schemas_voice.ml
@@ -65,11 +65,11 @@ let voice_tools : Masc_domain.tool_schema list =
   ; { name = "keeper_voice_agent"
     ; description =
         "Get your own voice capability and configuration. Reports assigned voice, \
-         available voices, active turn-based voice session state, and \
-         voice_loop guidance: operator audio is transcribed to normal keeper text \
-         turns, while keeper output uses keeper_voice_speak and dashboard audio \
-         clips. realtime_supported=false means no full-duplex live audio stream is \
-         bound to keeper turns. No network required."
+         available voices, active voice session state, available conversation \
+         modes, and voice_loop guidance. Without MASC_VOICE_REALTIME_WS_URL, \
+         operator audio is transcribed to normal keeper text turns and keeper \
+         output uses keeper_voice_speak/dashboard audio clips. No network \
+         required."
     ; input_schema = `Assoc [ "type", `String "object"; "properties", `Assoc [] ]
     }
   ; { name = "keeper_voice_sessions"
@@ -78,9 +78,11 @@ let voice_tools : Masc_domain.tool_schema list =
     }
   ; { name = "keeper_voice_session_start"
     ; description =
-        "Start a turn-based voice session for this keeper. The result includes \
-         voice_loop guidance that tells the keeper and dashboard how operator \
-         speech is transcribed and how keeper output is played."
+        "Start a voice session for this keeper. Defaults to turn_based batch \
+         STT/TTS. conversation_mode=realtime_bridge is accepted only when \
+         MASC_VOICE_REALTIME_WS_URL points at a configured realtime audio bridge; \
+         otherwise the tool fails closed instead of pretending a duplex stream \
+         exists."
     ; input_schema =
         `Assoc
           [ "type", `String "object"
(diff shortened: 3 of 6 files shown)


jeong-sik:
## Adversarial review — expose turn loop guidance (HEAD `b83e69e6511f`)

### [Change requested · MED] The voice-loop contract is restated as string literals in 5 places — the structured JSON should be the SSOT

The same contract facts now appear independently in:
1. `voice_session_manager.ml` `turn_based_voice_loop_json` (the intended structured source)
2. `voice_session_manager.ml` `session_to_json` — re-adds `conversation_mode`/`transport_mode`/`realtime_supported`/`voice_loop`
3. `keeper_tool_voice_runtime.ml` `voice_agent_capability_fields` — its own `"transport_mode", `String "batch_stt_tts"` and `"realtime_supported", `Bool false`
4. `keeper_turn.ml` — prose: "Voice sessions are turn-based: operator speech arrives as transcribed text through normal keeper turns; do not wait for a live duplex audio stream"
5. `tool_shard_types_schemas_voice.ml` — tool descriptions: "operator audio is transcribed to normal keeper text turns … realtime_supported=false means no full-duplex live audio stream"

The clearest symptom: `realtime_supported = false` is two separate `Bool false` literals (#1 and #3), and `transport_mode = "batch_stt_tts"` is two separate string literals — for the same underlying fact. If the voice implementation ever gains realtime/full-duplex support (or the transport changes), all five sites must be edited in lockstep, and because these are string/bool literals there is no compiler signal when they drift. This is the scattered-hardcoded-defaults pattern.

Fix direction: `turn_based_voice_loop_json` is already the right structured SSOT — have `voice_agent_capability_fields` and the tool-schema descriptions derive their `transport_mode` / `realtime_supported` / guidance from it (or a shared `voice_loop_contract` record) rather than restating. The keeper-turn prose guidance is the hardest to deduplicate (it's prompt text), but at minimum the boolean/string facts should come from one place.

### [Affirm] Honest capability declaration, not a stub
`realtime_supported=false` + `transport_mode="batch_stt_tts"` accurately describe a batch STT/TTS loop (operator audio → transcribe → normal keeper turn; keeper output → `keeper_voice_speak` → TTS clip). Declaring the real capability rather than implying full-duplex is correct — the tool guidance now tells the keeper not to wait for a live duplex stream, which prevents a real failure mode (keeper blocking on audio that never arrives as a stream).

### [Affirm] session_active derivation is correct
`turn_based_voice_loop_json ~session_active` and the `session_to_json` embedding both derive `session_active` from `session.status = Active` (Idle/Suspended → false). Consistent.

### [Low] `session_to_json` now emits non-round-trippable derived fields
`conversation_mode`/`transport_mode`/`realtime_supported`/`voice_loop` are added to `session_to_json` but are not read by `session_of_json`. That's fine *if* these are output-only contract metadata (the session record is the round-trippable part), and the existing decoder ignores unknown fields — worth confirming `session_of_json` doesn't reject unknown keys, since the field set it persists now differs from the one it emits.

jeong-sik:
[적대 리뷰 grounding · head `b83e69e65` · voice_loop guidance SSOT는 잘 됐으나 mode 값이 string 리터럴로 산포 [MED]]

사용자 지시("이미 있는 모듈 재구현 / 하드코딩 산포 검사")로 봤습니다. voice_loop guidance 자체는 SSOT로 잘 묶었지만, 그 안의 **mode 값들이 3개 생산 함수에 독립 string 리터럴로 흩어져** 있습니다.

**[Affirm] voice_loop guidance는 SSOT입니다.** `turn_based_voice_loop_json ~session_active`(voice_session_manager.ml:79) 단일 함수로 정의하고 `.mli:142`에 export, 두 소비 경로(`keeper_voice_agent`:25, `keeper_voice_session_start`의 `session_to_json`:125)가 그 함수를 **재사용**합니다. body의 "return the same turn-based loop guidance" 주장이 코드와 일치 — voice_loop 객체는 중복 정의가 아닙니다.

**[MED · 하드코딩 산포] `conversation_mode`/`transport_mode` 값이 3곳 리터럴.** `"conversation_mode", `String "turn_based"` + `"transport_mode", `String "batch_stt_tts"`가 세 **독립 생산 함수**에 박혀 있습니다:
- `:23` `voice_agent_capability_fields ~meta` (capability 선언)
- `:82` `turn_based_voice_loop_json` (voice_loop 객체 — SSOT 함수인데 그 *안*에도 리터럴)
- `:123` `session_to_json session` (session 상태 JSON)

세 JSON이 모두 "이 시스템 voice = turn_based / batch_stt_tts"라는 **하나의 사실**을 독립적으로 천명합니다. 미래에 realtime transport를 더하면 세 곳을 다 바꿔야 하고, 하나라도 놓치면 capability는 `batch_stt_tts`라는데 session은 다른 값을 보고하는 drift가 납니다 — 컴파일러가 못 잡습니다(string). voice_loop *객체*는 SSOT화했지만 그 안의 **mode 값은 SSOT가 아닙니다.**

**근본 (OCaml variant — MANIFEST.md "String 나열보다 Variant"):**
```ocaml
type conversation_mode = Turn_based   (* 향후 Realtime 등 *)
type transport_mode = Batch_stt_tts
let conversation_mode_to_string = function Turn_based -> "turn_based"
let transport_mode_to_string = function Batch_stt_tts -> "batch_stt_tts"
```
세 함수가 이 단일 to_string을 통과시키면, 새 mode 추가 시 exhaustive match가 모든 직렬화 지점을 컴파일 타임에 강제합니다(내 안티패턴 메모리의 unknown→permissive도 closed variant로 닫힘). variant가 과하다면 최소한 `let transport_mode_batch_stt_tts = "batch_stt_tts"` named constant 하나를 두고 3곳이 참조해야 합니다. 현재는 같은 리터럴이 2곳+ 등장하므로(software-development.md "2곳+ 등장 시 named constant 필수") 상수 추출 대상입니다.

정리: voice_loop guidance를 단일 함수+.mli로 SSOT화한 방향은 옳습니다. mode 값(`turn_based`/`batch_stt_tts`)만 variant 또는 named constant로 끌어올리면 capability/session/loop 세 JSON이 한 진실에서 파생됩니다.

(근거: `voice_session_manager.ml:79` turn_based_voice_loop_json SSOT + `.mli:142` export; `"batch_stt_tts"` 리터럴이 voice_agent_capability_fields:23 / turn_based_voice_loop_json:82 / session_to_json:123 3곳 반복. 차등 적대 검증 2026-06-21.)

anyang-keepers (approved):
## Review: APPROVED ✅

This PR adds `transport_mode` and `voice_loop` guidance fields to `keeper_voice_agent` response, plus updated guidance text in `keeper_turn.ml` and tool description in `tool_shard_types_schemas_voice.ml`.

**Good:**
- `transport_mode: "batch_stt_tts"` makes the transport contract explicit — agents understand they get transcribed text, not live duplex
- `voice_loop` JSON includes session_active state so agents can self-diagnose
- Tool description update is warranted: the old description was misleading about voice session semantics
- All changes are purely additive/schema, no runtime logic change

**No BLOCK.** Clean, minimal, well-scoped.

jeong-sik:
[적대 리뷰 follow-up · head `8f9a12eb2` · 내 16:54 mode-literal MED finding이 새 커밋에서 materialize]

16:54 리뷰(voice_loop guidance는 SSOT지만 mode 값이 string 리터럴로 산포 [MED]) 이후 새 커밋 `8f9a12eb2`("feat(voice): gate realtime bridge sessions")가 추가됐습니다. 그 커밋에서 제가 경고한 drift가 **실제로 시작됐습니다** — variant 도입이 아니라 리터럴을 더 추가했습니다.

**[MED 재확인 · 같은 turn-based 현실이 이제 두 string으로 갈라짐]** 새 커밋이 추가한 voice_loop 객체:
```ocaml
let turn_based_voice_loop_json ~session_active =
  [ "mode", `String "turn_based_batch"      (* ← 새 3번째 철자 *)
  ; "transport_mode", `String "batch_stt_tts"
  ; "realtime_supported", `Bool false
  ... ]
```
그런데 capability/session JSON은 여전히 `conversation_mode = "turn_based"`입니다. 즉 **하나의 turn-based 현실이 `conversation_mode="turn_based"`(2곳) ≠ `mode="turn_based_batch"`(voice_loop) 두 string으로** 표현되고, 컴파일러가 둘의 일관성을 강제하지 못합니다. 16:54에 "realtime transport를 더하면 drift난다"고 했는데, realtime을 **게이트하는** 이 커밋이 바로 그 첫 분기를 만들었습니다.

**[추가 · `realtime_supported` bool이 transport_mode와 독립 유지]** 지금 mode 표현이 4종 공존합니다: `conversation_mode="turn_based"`, `mode="turn_based_batch"`, `transport_mode="batch_stt_tts"`, `realtime_supported=false`. 커밋 제목("gate realtime bridge sessions")이 시사하듯 realtime은 곧 지원될 텐데, 그때 `realtime_supported`를 true로 뒤집으면서 `transport_mode`와 `mode` string도 같이 바꿔야 합니다 — **세 곳 독립 편집, 하나라도 놓치면 capability는 realtime이라는데 voice_loop는 batch라고 보고**합니다(string이라 컴파일러 무감지).

**근본 (16:54 권고 재확인 — 이제 더 시급):**
```ocaml
type transport_mode = Batch_stt_tts | Realtime
let transport_mode_to_string = function Batch_stt_tts -> "batch_stt_tts" | Realtime -> "realtime"
let realtime_supported = function Realtime -> true | Batch_stt_tts -> false  (* 파생, 별도 bool 아님 *)
```
이러면 `realtime_supported`가 transport에서 **파생**돼 모순 불가능해지고(`Realtime`인데 `realtime_supported=false`가 표현 불가), realtime 추가 시 exhaustive match가 모든 직렬화 지점을 컴파일 타임에 강제합니다. `turn_based` vs `turn_based_batch` 두 철자도 단일 to_string으로 수렴됩니다. realtime을 도입(게이트)하는 지금이 variant로 닫을 정확한 타이밍입니다 — 리터럴이 4종으로 늘어난 뒤엔 N-of-M 마이그레이션 비용이 커집니다.

(근거: 새 커밋 `8f9a12eb2`의 `turn_based_voice_loop_json` `mode="turn_based_batch"` vs capability `conversation_mode="turn_based"`; `realtime_supported` bool이 `transport_mode` string과 독립; 커밋 제목 realtime 게이트. 차등 적대 follow-up 2026-06-21, head 8f9a12eb2.)

jeong-sik:
Blocking: `MASC_VOICE_REALTIME_WS_URL` is copied verbatim into keeper-visible JSON, so a credential-bearing realtime bridge URL can leak to the LLM/tool logs.

The new mode/transport SSOT direction is good: `conversation_mode` is now a variant and the previous `turn_based` / `batch_stt_tts` literals are no longer independently produced by three separate functions.

But the new realtime path exposes the env-sourced endpoint value directly in multiple tool/session payloads:

- `lib/keeper/keeper_tool_voice_runtime.ml:289-305` reads `Voice_session_manager.realtime_bridge_endpoint ()` and returns it as `realtime_bridge.endpoint` from `keeper_voice_agent`.
- `lib/voice/voice_session_manager.ml:182-199` serializes active sessions with `realtime_bridge_endpoint`.
- `lib/voice/voice_session_manager.ml:137-144` also embeds the same raw endpoint in `voice_loop.bridge_endpoint`.
- The test now pins that raw exposure: `test/test_voice_runtime_overlay.ml:731-755` and `:762-763` assert the exact endpoint string is surfaced/persisted.

`realtime_bridge_endpoint` only validates the prefix is `ws://` or `wss://` (`lib/voice/voice_session_manager.ml:93-103`). It does not reject or redact userinfo/query strings, so values like `wss://token:secret@bridge.example/live?token=...` are accepted and then sent to the keeper-facing tool response. That crosses the env/secret boundary and is especially risky because `keeper_voice_agent` output is visible to the model and can be retained in tool logs.

Please keep the raw endpoint internal-only. The keeper-visible contract can expose non-secret facts such as `configured=true`, `required_env`, `transport_mode`, protocol, and maybe a redacted origin/host if needed. If the bridge implementation needs the full URL later, pass it to that transport internally rather than through `session_to_json` / `voice_loop_json` / `keeper_voice_agent`. Add a regression with a credential/query-bearing URL that verifies the raw secret does not appear in any keeper-visible JSON.

This is separate from the previous mode-literal SSOT comment; that part looks materially improved on head `8f9a12e`.

jeong-sik:
## Close-out — HEAD `8f9a12eb2f`: realtime bridge gating moves toward the SSOT I asked for

### [Partial-resolve — affirm] Typed `conversation_mode` now derives the contract facts
The new commit introduces the right structure for the 5-site duplication I flagged:

```ocaml
type conversation_mode =
  | Turn_based
  | Realtime_bridge of { endpoint : string }

let transport_mode_of_conversation_mode = function
  | Turn_based -> "batch_stt_tts"
  | Realtime_bridge _ -> "websocket_audio_bridge"

let realtime_supported = function
  | Turn_based -> false
  | Realtime_bridge _ -> true
```

So the contract facts (`transport_mode`, `realtime_supported`) now derive from a single typed variant instead of being independent literals scattered across sites. `voice_loop_json ~session_active` dispatches on the typed mode (no string match), and the realtime gating itself is principled: env-driven (`MASC_VOICE_REALTIME_WS_URL`), validated at the boundary (`ws://`/`wss://` prefix check), and fail-closed (`None` when unset/invalid). This is the SSOT direction my earlier review asked for — and it makes the realtime capability dynamic (mode-driven) rather than a hardcoded `false`, which is the correct model.

### [Residual, non-blocking] The JSON builders still inline the literals instead of calling the derived functions
`turn_based_voice_loop_json` and `realtime_bridge_voice_loop_json` each inline their `"transport_mode"` string and `"realtime_supported"` bool directly:
```ocaml
; "transport_mode", `String "websocket_audio_bridge"   (* in realtime_bridge_voice_loop_json *)
; "realtime_supported", `Bool true
```
…even though `transport_mode_of_conversation_mode` / `realtime_supported` already define those values for the same mode. So the literal is still duplicated (derived function + JSON builder). The typed variant already prevents the *modes* from drifting, but the strings within each builder can still drift from the derived functions. To fully close it, the builders would source those fields from the derived helpers (e.g. `transport_mode_of_conversation_mode mode`, `realtime_supported mode`) so there's one literal per fact. Minor — the structural improvement (typed variant + derived functions) is the load-bearing part and it's in place.

jeong-sik:
Addressed the realtime bridge endpoint leak in 29860e2e05.

Changes:
- Added Voice_session_manager.realtime_bridge_public_json as the single keeper-visible bridge metadata shape.
- keeper_voice_agent now reports configured/required_env with endpoint=null instead of the raw MASC_VOICE_REALTIME_WS_URL value.
- session_to_json no longer serializes realtime_bridge_endpoint; active_session and voice_loop now expose only redacted bridge metadata.
- session restore for realtime_bridge now rehydrates from the current env endpoint instead of trusting/persisting the previous raw URL.
- Regression uses a credential/query-bearing wss URL and asserts the raw endpoint plus secret fragment are absent from keeper_voice_agent, session_start/session_to_json, and active_session payloads.

Local checks:
- opam exec -- ocamlformat --check lib/voice/voice_session_manager.ml lib/voice/voice_session_manager.mli lib/keeper/keeper_tool_voice_runtime.ml test/test_voice_runtime_overlay.ml
- git diff --check
- opam exec -- ocamlc -stop-after parsing lib/voice/voice_session_manager.ml lib/voice/voice_session_manager.mli lib/keeper/keeper_tool_voice_runtime.ml test/test_voice_runtime_overlay.ml

Focused Dune build scripts/dune-local.sh build test/test_voice_runtime_overlay.exe was attempted, but the local wrapper refused because bare Dune processes are active on the machine. I left those processes untouched; CI remains the authority for full type/test proof.

jeong-sik:
## Adversarial review — new HEAD `29860e2e05`: redact realtime bridge endpoint (on top of typed `conversation_mode`)

**Strong affirm on the redaction.** This commit layers a clean, typed secret boundary on top of the `conversation_mode` variant from `8f9a12eb2f`, and it does it the right way — not a string/regex mask.

The raw websocket endpoint never reaches a keeper-facing JSON path. It's redacted at **two** independent points:

1. `session_to_json` drops the endpoint at the variant match:
```ocaml
let bridge_endpoint =
  match mode with
  | Turn_based -> `Null
  | Realtime_bridge _ -> `Null   (* endpoint discarded via `_` *)
in
```
2. `realtime_bridge_public_json` emits a **constant** `Null` regardless of input:
```ocaml
let realtime_bridge_public_json ?endpoint () =
  `Assoc
    [ "configured", `Bool (Option.is_some endpoint)
    ; "required_env", `String realtime_bridge_env
    ; "endpoint", `Null
    ]
```

So even though `Realtime_bridge { endpoint }` carries the raw URL in the typed value, every serialization that a keeper can observe replaces it with `Null` by construction — there is no code path where the raw endpoint is echoed into the public JSON. This is parse-don't-validate redaction: the secret is structurally unrepresentable in the public shape, not scrubbed after the fact. Compare to the brittle alternative (regex-replacing `wss://...?token=...` with `***`), which would silently leak any endpoint shape the regex doesn't anticipate — that trap is avoided here.

Two more things done right:

- **Round-trip does not persist the secret.** `conversation_mode_of_json` reconstructs `Realtime_bridge { endpoint }` by re-calling `realtime_bridge_endpoint ()` from the env, not by reading an endpoint back out of JSON. So the session file on disk never stores the endpoint — it's always re-derived from `MASC_VOICE_REALTIME_WS_URL` at load time. A redaction that lets the secret round-trip through storage would be half a redaction; this one closes that loop.
- **Deterministic, fail-closed validation.** `realtime_bridge_endpoint ?(getenv = Sys.getenv_opt) ()` takes an injectable `getenv` (testable, no real env needed) and returns `None` on missing/blank/non-`ws://`|`wss://` values. Unknown shape → `None`, never a permissive default. The `ws://`/`wss://` prefix check is a *protocol* guard, not a string classifier — it's validating the transport scheme at the boundary, which is the legitimate use of a prefix check (unlike using a substring to discriminate between semantic variants).

### [Resolved — closing my earlier residual for the `session_to_json` path]
My previous review flagged that `transport_mode`/`realtime_supported` were independent literals scattered across sites. On this cumulative head, `session_to_json` now routes through the derived functions:
```ocaml
("transport_mode", `String (transport_mode_of_conversation_mode mode));
("realtime_supported", `Bool (realtime_supported mode));
```
and the keeper runtime path does too (`keeper_tool_voice_runtime.ml`). So the SSOT direction I asked for is in place for those two paths. Closing that part.

### [Residual, LOW — non-blocking] The `voice_loop_json` builders still inline the literals
`turn_based_voice_loop_json` and `realtime_bridge_voice_loop_json` each inline their `"transport_mode"` string and `"realtime_supported"` bool directly:
```ocaml
; "transport_mode", `String "batch_stt_tts"        (* turn_based_voice_loop_json *)
; "realtime_supported", `Bool false
...
; "transport_mode", `String "websocket_audio_bridge"  (* realtime_bridge_voice_loop_json *)
; "realtime_supported", `Bool true
```
even though `transport_mode_of_conversation_mode`/`realtime_supported` already define those exact values for the same modes. So the same fact now lives in **two representations**: derived function (in `session_to_json`) and inline literal (in `voice_loop_json`). The typed variant still prevents the *modes* from drifting, but the strings within `voice_loop_json` can drift from the derived functions with no compile/test signal — the exact mirror-drift shape. To fully close it, the two builders would source those two fields from the derived helpers (e.g. `transport_mode_of_conversation_mode Turn_based`) so there's one literal per fact. Minor — the load-bearing typed-redaction + derived-function work is already landed; this is the last consistency pass.

Net: the redaction design (typed variant → constant-`Null` public shape → env re-derivation on load) is the correct, non-string-mask shape and I'd merge it. Only the small `voice_loop_json` literal consolidation is outstanding, and it's non-blocking.

PR #21831, opened 2026-06-20
docs(rfc): RFC-0259 P5/P6 — Memory OS GC/cap-TTL + producer idempotency (C/E/F)

jeong-sik:
## Summary

RFC-only amendment to **RFC-0259**, routing three of the four **semantic-change** defects from the 2026-06-20 Memory OS adversarial audit (Issue #21789) into typed phases. No code — design first, per the memory-subsystem RFC-first rule. The audit's two **resource-leak** defects were already fixed at the source in PR #21787 (merged).

Adds:
- **P5 — GC activation + cap TTL-awareness** (defect C). Disk-hygiene + retention-path determinism only (no prompt-behavior change; recall already filters expired rows). Turns on the existing dry-run-gated GC and makes the hot-path cap honor the same typed `valid_until` the GC uses.
- **P6 — producer idempotency & anchor stability** (defects E + F). A stable claim identity so an unchanged re-extraction UPSERTs instead of appending (E), and anchor inheritance on re-mint with a reconciler-only horizon advance (F). Explicitly **amends** the shipped `reobserve_fact` "re-observing IS re-verification" rule for volatile claims.
- A correction: §3.4's `supersedes` episode field was never implemented; P3 shipped the `merge_and_cap_facts`/`reobserve_fact` UPSERT keyed on exact `normalize_claim`, which P6 extends. §3.4/§3.1 now carry forward pointers.

## Why RFC, not a direct code PR

Each defect has a workaround trap the RFC explicitly closes, so a code-first fix would risk setting a workaround precedent:

- **E** — fuzzy/semantic/embedding dedup is **rejected** (string-classifier + dedup double signature). Root = producer identity (idempotent extraction reusing the existing typed `external_ref_of_claim`), not post-hoc clustering.
- **C** — a TTL prune janitor is **rejected** (cap-as-fix). Reframed as "cap and GC honor the same typed `valid_until` boundary" — closing a determinism gap, not adding a symptom-suppressor. The existing `fact_store_max` bound is unchanged.
- **F** — anchor immutability with a reconciler-only horizon advance, so a producer rewording cannot silently reset decay.

## Grounding & review

All file:line citations re-verified against current main `c68c7d6500` via a parallel grounding workflow (the issue's baseline `5c295f5e5e` had drifted — line numbers corrected throughout). A three-lens adversarial pass (workaround-signature / citation accuracy / RFC coherence) returned ship-with-fixes on all lenses; every finding is folded in (incl. the #21665 P2 citation correction and the F-vs-`reobserve_fact` consistency note).

## Scope

`docs/rfc/RFC-0259-*.md` only. No code. Each phase's implementation is gated as specified in the phasing table.

**Deferred:** defect D (events/episodes unbounded append) is routed to RFC-0247 (forgetting charter) but **not** amended here — amending RFC-0247's `## §1`-style body would activate a pre-existing `rfc-enforcer` gap (its R5 caller-context companion lives in `.tmp/`, which is gitignored and unreachable in CI; the `docs/rfc` enforcer step lacks the `--no-tmp-check` flag the `docs/design` step has). D's full design is captured in Issue #21789; landing it cleanly needs that workflow-consistency fix first. Issue #21789 stays open to track P5/P6 implementation + D.

RFC-WAIVED: this PR amends an RFC document; it does not modify credential/identity/operator/sandbox/dashboard-credential/hooks/workflow code subsystems.

### DIFF
--- a/docs/rfc/RFC-0259-memory-os-volatile-claim-grounding-retraction-decay.md
+++ b/docs/rfc/RFC-0259-memory-os-volatile-claim-grounding-retraction-decay.md
@@ -2,9 +2,11 @@
 
 **Status**: Draft
 **Date**: 2026-06-19
-**Verified against base main**: `99d3716b72`
+**Verified against base main**: `99d3716b72` (P1–P4; the 2026-06-21 amendment re-verified §3.6/§3.7 against `c68c7d6500`)
 **Builds on**: [RFC-0247](./RFC-0247-memory-os-associative-graph-forgetting-brain.md) (purge of the composite score; "a fact's value is the librarian's judgment, not a number"), [RFC-0244](./RFC-0244-memory-os-tiered-stores.md) (tiered fact stores)
 **Supersedes intent of**: PR #21363 `feat(memory-os): stale decay mechanism with TTL-based GC` — **CLOSED, not merged** (`gh pr view 21363` → `state: CLOSED, mergedAt: null`). The fleet currently runs with no decay/grounding/retraction at all; this RFC re-states that need with a typed boundary instead of a flat TTL.
+
+**Amendment (2026-06-21, base main `c68c7d6500`)**: P1–P4 landed (#21644 P1, #21665 P2, #21668 P3, #21718 P4). This revision adds **P5 — GC activation + cap TTL-awareness** (§3.6) and **P6 — producer idempotency & anchor stability** (§3.7) from the 2026-06-20 Memory OS adversarial audit (`reports/masc-memory-os-leak-stuck-audit-20260620-1614.html`, Issue #21789, defects C/E/F). The audit's resource-leak findings (cadence table, write-path fd) were fixed at the source in PR #21787 and are out of scope here. Defect D (events/episodes unbounded append) is routed to RFC-0247 (forgetting charter) as a future amendment, tracked in Issue #21789 — deferred here because amending RFC-0247's `## §1`-style body would activate a pre-existing `rfc-enforcer` gap (R5 caller-context is unsatisfiable for `docs/rfc` in CI). All file:line references in §3.6/§3.7 re-verified against `c68c7d6500`.
 
 ## 1. Summary
 
@@ -60,7 +62,7 @@ PR #21363 (a flat TTL decay) tried to address #4 and was closed. This RFC narrow
 | Re-check PR/issue #X against truth | **external** (`gh`/GitHub API) | reconciler fiber |
 | Confirmed/contradicted decision from the diff | deterministic | reconciler |
 | Time-bound claim with no external referent | deterministic (TTL) | `category_valid_until` |
-| Which claim a new observation supersedes | LLM judgment | librarian (new schema field) |
+| Which claim a new observation supersedes | LLM judgment | librarian (shipped as the re-observe UPSERT, not a schema field; see §3.7) |
 | Recall suppression of unverified-volatile-past-horizon | deterministic | recall |
 
 The new boundary statement: **a claim that names verifiable external state must be grounded deterministically — the system can and must run the check itself.** Leaving it to the LLM (which only sees stale history) or to never-verify is what produced the bug.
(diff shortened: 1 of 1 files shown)


jeong-sik:
## Adversarial review — RFC-0259 P5/P6 amendment (HEAD `9be692971d9f`, docs-only)

I read the full amendment against the audit (Issue #21789) and the project's workaround-rejection bar. **Strong affirm — no change requested.** This RFC is a textbook application of "fix the root, reject the symptom-suppressor" at design time.

### [Affirm] Workaround rejection is cited explicitly, by signature
- §3.6 (defect C): "Add a TTL prune janitor is the cap/prune-as-fix signature and is **rejected**." The fix is reframed as closing a determinism gap — the hot-path cap and `run_gc` must honor the *same* typed `valid_until` — with `fact_store_max` explicitly unchanged. No new cap/cooldown/counter.
- §3.7 (defect E): "fuzzy / semantic / embedding dedup is a **double signature violation** (string-classifier + dedup) under the CLAUDE.md bar and is **rejected**." Root is producer identity — a stable claim key reusing the *existing* typed `external_ref_of_claim` (closed `{Pr|Issue|Task; id}` sum, conservative bar) — so unchanged re-extraction UPSERTs instead of appending. This is the correct root-cause direction and avoids inventing a new classifier.

### [Affirm] The producer-vs-reconciler anchor rule is coherent (I checked this specifically)
F amends the shipped `reobserve_fact` ("re-observing IS re-verification") to: producer re-extraction **inherits** `first_seen`/`last_verified_at`/`valid_until`; only the **external grounding reconciler** (`Stale_open` verdict, actually re-checked against GitHub) advances the anchor. The RFC states the edge explicitly — "Conflating the two re-introduces F" — so P2's external verifier correctly keeps its horizon-advance authority while the producer loses it. This resolves the tension with P2 cleanly rather than papering over it.

### [Affirm] Harness plan matches the TLA+ bug-model pattern
P5: dry-run GC fixture + pure `partition_expired ~now` unit (durable-`None` never dropped; expired dropped before ranking). P6: reworded→1-row test + `first_seen`-preservation unit + TLA `ReMintResetsAnchor` bug action violating `NoUnverifiedVolatileClaimSurvivesBeyondHorizon` (clean spec satisfies, buggy violates). That's exactly the clean-vs-buggy-cfg discipline.

### [Affirm] Honesty of the corrections
- §3.4: the `supersedes : string list` episode field "was never implemented" — P3 actually shipped the `merge_and_cap_facts`/`reobserve_fact` UPSERT keyed on `normalize_claim`. The RFC retains the old text "for intent" and points forward to §3.7 rather than silently rewriting history.
- Grounding re-verified against `c68c7d6500` with the prior `5c295f5e5e` drift noted; defect D deferred to RFC-0247 with the *specific* blocker named (rfc-enforcer R5 caller-context unsatisfiable for `docs/rfc` in CI because the companion lives in gitignored `.tmp/`).

### [Non-blocking observation] E leaves the identity key design as an "or"
§3.7 proposes the stable identity as "e.g. `external_ref` + claim-kind, **or** a producer-assigned stable claim-id". Both are typed (good — neither is a substring classifier), but the implementation will need to pick one. Not a blocker for an RFC, just flagging that the choice has consequences for how `external_ref`-less self-narrative claims (the audit's "All verification work complete" example, which has no `external_ref`) get a stable key — that's precisely the case that produced the 15–23 micro-variants, and it may not have an `external_ref` to anchor on. Worth nailing down before P6/E implementation so the identity covers the self-narrative case that motivated it.

anyang-keepers (approved):
## Review: APPROVED ✅

**PR #21831** — RFC-0259 P5/P6 amendment. 문서만 변경 (docs/rfc/), 코드 변경 없음.

**Good:**
- P5 (GC cap-TTL awareness)와 P6 (producer idempotency) 모두 Memory OS adversarial audit의 C/E/F defect를 명확히 커버
- Audit에서 resource-leak finding은 #21787에서 fix되었고, Defect D는 RFC-0247로 라우팅함 — scope 명확
- Amendment가 base main commit hash(`c68c7d6500`)로 re-verified 되어 있음
- rfc-enforcer gap(R5)에 대한 설명 투명함

**No BLOCK.** RFC 문서 업데이트 — 내용 정확하고 scope 명확히 분리됨.

PR #21832, opened 2026-06-20
feat(thompson): persist runtime priors

jeong-sik:
Summary
- load Thompson priors from <base>/.masc/autonomy_stats.jsonl during server state creation after base-path normalization
- set the Thompson base path at runtime and register a single shutdown save hook
- make runtime Thompson feedback persist when base path is configured, including pending votes overlaid into saved snapshots
- keep load_stats boot-safe: malformed JSONL is logged/dropped, schema-mismatched rows without alpha/beta are skipped, and I/O/schema exceptions are logged without aborting startup
- add persistence regression tests for loading stats, saving pending votes, corrupt JSONL, and schema-mismatched rows

Validation
- ocamlformat --check lib/thompson/thompson_sampling.ml lib/server/server_runtime_bootstrap.ml test/test_thompson_sampling.ml
- git diff --check
- ocamlc -stop-after parsing lib/thompson/thompson_sampling.ml
- ocamlc -stop-after parsing lib/server/server_runtime_bootstrap.ml
- ocamlc -stop-after parsing test/test_thompson_sampling.ml
- attempted env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_thompson_sampling.exe; stopped after >90s with no compiler output, leaving full compile/test to CI per operator preference

Notes
- Earlier focused Dune attempts were blocked by local Dune/opam locks; this PR relies on GitHub CI for full OCaml build/test proof.

### DIFF
--- a/lib/server/server_runtime_bootstrap.ml
+++ b/lib/server/server_runtime_bootstrap.ml
@@ -213,8 +213,17 @@ let record_runtime_toml_load_failure msg =
        Otel_metric_store.inc_counter
          metric_keeper_runtime_config_load_failures
          ~labels:[ "reason", reason ]
-         ())
+      ())
     reason
+
+let thompson_shutdown_hook_registered = Atomic.make false
+
+let ensure_thompson_persistence ~base_path =
+  Thompson_sampling.set_base_path base_path;
+  Thompson_sampling.load_stats ();
+  if Atomic.compare_and_set thompson_shutdown_hook_registered false true then
+    Shutdown.register ~name:"thompson_sampling_save" ~priority:24 (fun () ->
+      Thompson_sampling.save_stats ())
 
 let create_server_state ~sw ~base_path ~clock ~mono_clock ~net ~proc_mgr ~fs
     ?env ()
@@ -249,6 +258,7 @@ let create_server_state ~sw ~base_path ~clock ~mono_clock ~net ~proc_mgr ~fs
     Env_config_core.base_path_input_env_key
     (Option.value ~default:"" input_base_path);
   Unix.putenv Env_config_core.base_path_env_key base_path;
+  ensure_thompson_persistence ~base_path;
   bootstrap_base_path_config_root ~base_path;
   (* Apply keeper runtime overrides from the resolved config root's
      runtime.toml. Must run before any module that reads
--- a/lib/thompson/thompson_sampling.ml
+++ b/lib/thompson/thompson_sampling.ml
@@ -256,6 +256,22 @@ let stats_to_json (s : agent_stats) : Yojson.Safe.t =
     ("updated_at", `Float s.updated_at);
   ]
 
+let copy_stats (s : agent_stats) : agent_stats =
+  {
+    name = s.name;
+    alpha = s.alpha;
+    beta = s.beta;
+    selections = s.selections;
+    last_selected_at = s.last_selected_at;
+    total_votes_up = s.total_votes_up;
+    total_votes_down = s.total_votes_down;
+    posts_created = s.posts_created;
+    comments_created = s.comments_created;
+    skips = s.skips;
+    guard_penalties_total = s.guard_penalties_total;
+    updated_at = s.updated_at;
+  }
+
 let stats_of_json (json : Yojson.Safe.t) : agent_stats option =
   let name = Json_util.get_string_with_default json ~key:"name" ~default:"" in
   let alpha = Json_util.get_float json "alpha" |> Option.value ~default:0.0 in
@@ -283,6 +299,19 @@ let stats_of_json (json : Yojson.Safe.t) : agent_stats option =
     guard_penalties_total;
     updated_at;
   }
+
+let apply_vote_counts ~decay (s : agent_stats) ~votes_up ~votes_down =
+  let total = votes_up + votes_down in
+  if total > 0 then begin
+    let success_rate = float_of_int votes_up /. float_of_int total in
+    s.alpha <- (s.alpha -. 1.0) *. decay +. 1.0 +. success_rate;
+    s.beta <- (s.beta -. 1.0) *. decay +. 1.0 +. (1.0 -. success_rate);
+    s.alpha <- Float.max min_prior s.alpha;
+    s.beta <- Float.max min_prior s.beta;
+    s.total_votes_up <- s.total_votes_up + votes_up;
+    s.total_votes_down <- s.total_votes_down + votes_down;
+    s.updated_at <- Time_compat.now ()
+  end
 
 (** {1 Persistence} *)
 
@@ -317,21 +346,47 @@ let load_stats () =
         (Printexc.to_string e)
   end
 
+let stats_snapshot_for_persistence () =
+  let decay = Env_config.AgentSelection.vote_decay_factor in
+  with_ts_ro (fun () ->
+    let by_name =
+      Hashtbl.create (Hashtbl.length stats_table + Hashtbl.length pending_votes)
+    in
+    Hashtbl.iter
+      (fun name stats -> Hashtbl.replace by_name name (copy_stats stats))
+      stats_table;
+    Hashtbl.iter
+      (fun name (votes_up, votes_down) ->
+         if votes_up + votes_down > 0 then begin
+           let stats =
+             match Hashtbl.find_opt by_name name with
+             | Some stats -> stats
+             | None ->
+                 let stats = make_default_stats name in
+                 Hashtbl.replace by_name name stats;
+                 stats
+           in
+           apply_vote_counts ~decay stats ~votes_up ~votes_down
+         end)
+      pending_votes;
+    Hashtbl.fold (fun _ stats acc -> stats :: acc) by_name [])
+
 let save_stats () =
   let path = stats_path () in
   try
-    (* Serialise the table under the lock so a concurrent [record_*]
-       cannot [Hashtbl.replace] mid-iteration and corrupt the
-       iteration. *)
-    let content, count =
-      with_ts_ro (fun () ->
-        let buf = Buffer.create 4096 in
-        Hashtbl.iter (fun _ s ->
-          Buffer.add_string buf (Yojson.Safe.to_string (stats_to_json s));
-          Buffer.add_char buf '\n'
-        ) stats_table;
-        (Buffer.contents buf, Hashtbl.length stats_table))
-    in
+    (* Snapshot under the lock, then serialise the copies outside the
+       critical section. Pending votes are overlaid on the snapshot so a
+       graceful shutdown cannot lose batched feedback that has not reached
+       [flush_pending_votes] yet. *)
+    let snapshot = stats_snapshot_for_persistence () in
+    let buf = Buffer.create 4096 in
+    List.iter
+      (fun stats ->
+         Buffer.add_string buf (Yojson.Safe.to_string (stats_to_json stats));
+         Buffer.add_char buf '\n')
+      snapshot;
+    let content = Buffer.contents buf in
+    let count = List.length snapshot in
     Fs_compat.save_file path content;
     Log.Metrics.debug "thompson sampling saved stats for %d agents" count
   with
@@ -339,6 +394,12 @@ let save_stats () =
   | e ->
     Log.Thompson.error "Error saving stats: %s"
       (Printexc.to_string e)
+
+let persistence_configured () =
+  with_ts_ro (fun () -> Option.is_some !base_path_ref)
+
+let save_stats_if_configured () =
+  if persistence_configured () then save_stats ()
 
 (** {1 Feedback Updates} *)
 
@@ -350,7 +411,8 @@ let record_vote ~agent_name ~direction =
       | `Up -> (up + 1, down)
       | `Down -> (up, down + 1)
     in
-    Hashtbl.replace pending_votes agent_name (up', down'))
+    Hashtbl.replace pending_votes agent_name (up', down'));
+  save_stats_if_configured ()
 
 let flush_pending_votes () =
   (* [ts_mu] is documented as protecting [pending_votes], but this
@@ -376,20 +438,11 @@ let flush_pending_votes () =
     let total = votes_up + votes_down in
     if total > 0 then begin
       let s = get_stats agent_name in
-      let success_rate = float_of_int votes_up /. float_of_int total in
       with_ts_rw (fun () ->
-        (* Apply decay to existing priors, then add new evidence *)
-        s.alpha <- (s.alpha -. 1.0) *. decay +. 1.0 +. success_rate;
-        s.beta <- (s.beta -. 1.0) *. decay +. 1.0 +. (1.0 -. success_rate);
-        (* Clamp to minimum *)
-        s.alpha <- Float.max min_prior s.alpha;
-        s.beta <- Float.max min_prior s.beta;
-        (* Update totals *)
-        s.total_votes_up <- s.total_votes_up + votes_up;
-        s.total_votes_down <- s.total_votes_down + votes_down;
-        s.updated_at <- Time_compat.now ())
+        apply_vote_counts ~decay s ~votes_up ~votes_down)
     end
-  ) snapshot
+  ) snapshot;
+  if snapshot <> [] then save_stats_if_configured ()
 
 (* The record_* helpers below all mutate an [agent_stats] returned by
    [get_stats].  [get_stats] re-acquires [ts_mu] around the lookup, but
@@ -403,7 +456,8 @@ let record_selection ~agent_name =
   with_ts_rw (fun () ->
     s.selections <- s.selections + 1;
     s.last_selected_at <- Time_compat.now ();
-    s.updated_at <- Time_compat.now ())
+    s.updated_at <- Time_compat.now ());
+  save_stats_if_configured ()
 
 let record_action ~agent_name ~action =
   let s = get_stats agent_name in
@@ -412,7 +466,8 @@ let record_action ~agent_name ~action =
      | `Post -> s.posts_created <- s.posts_created + 1
      | `Comment -> s.comments_created <- s.comments_created + 1
      | `Skip -> s.skips <- s.skips + 1);
-    s.updated_at <- Time_compat.now ())
+    s.updated_at <- Time_compat.now ());
+  save_stats_if_configured ()
 
 (** {1 Quality Signal Integration} *)
 
@@ -450,7 +505,8 @@ let record_guard_penalty ~agent_name =
     s.beta <- s.beta +. guard_penalty_beta_nudge;
     s.beta <- Float.max min_prior s.beta;
     s.guard_penalties_total <- s.guard_penalties_total + 1;
-    s.updated_at <- Time_compat.now ())
+    s.updated_at <- Time_compat.now ());
+  save_stats_if_configured ()
 
 (** Record Post Verifier result into Thompson Sampling priors. *)
 let record_quality_signal ~agent_name ~(verdict : quality_verdict) =
@@ -462,7 +518,8 @@ let record_quality_signal ~agent_name ~(verdict : quality_verdict) =
      | Fail _ -> s.beta <- s.beta +. quality_fail_beta_penalty);
     s.alpha <- Float.max min_prior s.alpha;
     s.beta <- Float.max min_prior s.beta;
-    s.updated_at <- Time_compat.now ())
+    s.updated_at <- Time_compat.now ());
+  save_stats_if_configured ()
 
 (** {1 Selection Algorithm} *)
 
--- a/lib/thompson/thompson_sampling.ml
+++ b/lib/thompson/thompson_sampling.ml
@@ -75,9 +75,21 @@ let ts_mu = Eio.Mutex.create ()
 let with_ts_rw f = Eio_guard.with_mutex ts_mu f
 let with_ts_ro f = Eio_guard.with_mutex_ro ts_mu f
 
+let clear_in_memory_stats_unlocked () =
+  Hashtbl.clear stats_table;
+  Hashtbl.clear pending_votes
+;;
+
 (** Set base path for stats storage. Call during server init. *)
 let set_base_path path =
-  with_ts_rw (fun () -> base_path_ref := Some path)
+  with_ts_rw (fun () ->
+    let changed =
+      match !base_path_ref with
+      | Some current -> not (String.equal current path)
+      | None -> true
+    in
+    if changed then clear_in_memory_stats_unlocked ();
+    base_path_ref := Some path)
 
 (** Stats file path — uses cluster base_path, not execution directory *)
 let stats_path () =
@@ -320,6 +332,13 @@ let apply_vote_counts ~decay (s : agent_stats) ~votes_up ~votes_down =
 
 (** {1 Persistence} *)
 
+let replace_stats_snapshot parsed =
+  with_ts_rw (fun () ->
+    clear_in_memory_stats_unlocked ();
+    List.iter (fun s -> Hashtbl.replace stats_table s.name s) parsed;
+    Hashtbl.length stats_table)
+;;
+
 let load_stats () =
   let path = stats_path () in
   if Fs_compat.file_exists path then begin
@@ -345,9 +364,7 @@ let load_stats () =
         ) entries
       in
       let count =
-        with_ts_rw (fun () ->
-          List.iter (fun s -> Hashtbl.replace stats_table s.name s) parsed;
-          Hashtbl.length stats_table)
+        replace_stats_snapshot parsed
       in
       Log.Metrics.debug "thompson sampling loaded stats for %d agents" count
     with
@@ -355,6 +372,9 @@ let load_stats () =
     | e ->
       Log.Thompson.error "Error loading stats: %s"
         (Printexc.to_string e)
+  end else begin
+    let count = replace_stats_snapshot [] in
+    Log.Metrics.debug "thompson sampling loaded stats for %d agents" count
   end
 
 let stats_snapshot_for_persistence () =
--- a/lib/thompson/thompson_sampling.ml
+++ b/lib/thompson/thompson_sampling.ml
@@ -278,9 +278,9 @@ let stats_of_json (json : Yojson.Safe.t) : agent_stats option =
   let name = Json_util.get_string_with_default json ~key:"name" ~default:"" in
   if String.equal (String.trim name) "" then
     None
-  else
-  let alpha = Json_util.get_float json "alpha" |> Option.value ~default:0.0 in
-  let beta = Json_util.get_float json "beta" |> Option.value ~default:0.0 in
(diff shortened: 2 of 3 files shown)


jeong-sik:
### 적대 리뷰 (head `b9a31f481`) — durability feature가 non-atomic write 사용 [MED · #21818 SSOT 미사용]

`feat(thompson): persist runtime priors`의 목적은 priors의 **재시작 간 내구 보존**인데, 영속화 경로가 비원자적 write를 씁니다:

```ocaml
(* save_stats, thompson_sampling.ml:~137 *)
Fs_compat.save_file path content;   (* open → output_string → close. fsync/tmp-rename 없음 *)
```

`Fs_compat.save_file`(`fs_compat.mli:29` `-> unit`, 구현 `fs_compat.ml:230`)는 plain write입니다 — tmp+fsync+rename 경계가 없어 **크래시/전원손실이 write 중에 나면 priors 파일이 torn/truncated** 됩니다. 같은 모듈에 이미 durable SSOT `Fs_compat.save_file_atomic`(`fs_compat.mli:33`, tmp→fsync→rename→fsync(dir))가 있고, board/event-queue 영속화 + **memory-os가 방금 #21818로 그 위에 통합**됐습니다. priors를 "durable하게 persist"하겠다는 PR이 비원자 writer를 쓰는 건 그 목적과 어긋납니다.

**심각도는 load 복원력에 달려 있습니다(확인 요청).** torn write가 마지막 JSONL 행을 깨뜨렸을 때:
- load_stats가 **행별로 skip**(`stats_of_json : _ -> agent_stats option`이 row-level option을 반환하므로 filter 패턴이면)하면 → 손실은 마지막 agent 1행 (LOW).
- 첫 malformed 행에서 **abort**하면 → 그 뒤 모든 priors 손실, 즉 파일 전체가 사실상 무효 (MED↑). #21807에서 병렬이 짚은 "partial corruption drops the whole queue"와 동형입니다.

production load_stats가 어느 쪽인지 명시 확인을 권합니다(저는 `Fs_compat.load_jsonl`이 테스트에만 쓰이고 프로덕션 load 경로가 별개임을 확인했으나, 행 단위 skip 여부까지는 단정하지 못했습니다 — over-claim 회피).

**fix:** `Fs_compat.save_file` → `Fs_compat.save_file_atomic`로 라우팅. 후자는 `(unit, string) result`를 반환하므로, `save_stats`의 기존 `try ... with Log.Thompson.error` 블록을 `match ... | Error msg -> log` 로 바꾸면 됩니다(#21818이 `keeper_memory_os_io`에서 한 것과 동형 — `Eio.Cancel.Cancelled` 재전파도 그 primitive가 보존). priors가 Bayesian 누적이라 손실 시 self-heal되지만, "durable persist" 계약을 내건 이상 atomic이 맞습니다.

(참고 positive: shutdown hook 등록을 `Atomic.compare_and_set`로 idempotent 처리한 것, 테스트가 `with_temp_dir`/`Fs_compat.mkdir_p`/`load_jsonl`로 SSOT 재사용한 것, `stats_of_json`이 option 반환인 것은 정합합니다.)

(근거: `thompson_sampling.ml:~137` `Fs_compat.save_file`; `fs_compat.mli:29` save_file(plain) vs `:33` save_file_atomic(durable); #21818 memory-os 통합. head b9a31f481. 차등 적대 리뷰 2026-06-21.)

jeong-sik:
## Adversarial review — Thompson persist runtime priors (HEAD `b9a31f481167`, +163/-30)

Persisting statistical priors is a correctness-sensitive surface (drift changes sampling), so I read the save/load/overlay logic carefully and verified the flush interaction in source. Two MED findings; the overlay logic itself is sound.

### [Change requested · MED] Persistence uses non-atomic `save_file`, not the `save_file_atomic` durability SSOT

`save_stats` calls `Fs_compat.save_file path content`. That's the plain write — no tmp→fsync→rename. This is inconsistent with the durability boundary every other recent persistence PR routes through (`save_file_atomic`, used by #21807 chat-queue, #21818 memory-os, #21798 skill-candidates). Because this save fires on **every** vote/selection/action/penalty (`record_vote`/`record_selection`/`record_action`/`record_guard_penalty`/`record_quality_signal` all call `save_stats_if_configured ()`), the window for a crash mid-write is wide; a torn `autonomy_stats.jsonl` then has to survive `load_stats`. Either use `save_file_atomic` (the established durability primitive — the `Fs_compat.save_file_atomic` you already use elsewhere in this PR's tree) so the file is never observed half-written, or document why Thompson priors tolerate a torn write (they're statistical, so partial loss may be acceptable — but that should be an explicit decision, not an accident of which helper got called).

### [Change requested · MED] The code was never built and the persistence test was never run

The body is honest about this: only `ocamlc -stop-after parsing` (syntax-only, no type check) ran, the `dune build` was canceled (timeout) then blocked (`/tmp/me-dune-local.lock` held by another worktree), and the new "persistence regression test for loading stats and saving pending votes" was therefore never executed. For persistence logic specifically — where the bug class is "loses/double-counts votes across restart" and is invisible to a parser — an unrun test is not verification. Request: CI must actually build + run `test_thompson_sampling.exe` green on this head before merge; the parse-only local signal is insufficient.

### [Affirm] The pending-votes overlay and flush interaction is correct (I verified the clear)

I checked `flush_pending_votes` in source: it snapshots `pending_votes` under `ts_mu`, then `Hashtbl.clear pending_votes` **under the same lock** (line 18), releases, and applies to the live table outside the lock. So:
- `stats_snapshot_for_persistence` overlays pending votes onto **copies** (`copy_stats`), never the live table — the snapshot captures unflushed feedback for a graceful-shutdown save.
- After flush, `pending_votes` is empty, so the next snapshot can't re-overlay the same votes. **No double-count.** The snapshot+clear atomicity is exactly the right defense against the `record_vote` race the comment describes.

### [Affirm] DRY + lock discipline
- `apply_vote_counts` extracts the decay/alpha/beta/clamp/totals math shared by `flush_pending_votes` and `stats_snapshot_for_persistence` — removes the prior inline duplication.
- Snapshot is taken under `with_ts_ro`, serialization happens **outside** the lock (releases `ts_mu` before the file I/O).
- Shutdown hook registered once via `Atomic.compare_and_set thompson_shutdown_hook_registered false true` — idempotent even if `ensure_thompson_persistence` is called multiple times.

### [Low] Write-per-event on a hot path
Every `record_*` triggers a full-table save. Selection votes can be frequent, so this is a file rewrite per signal. Since the shutdown hook already exists, a coalesced/periodic save (or saving only on flush + shutdown) would cut I/O substantially without losing the durability intent. Not a correctness issue, just a cost note for a production fleet.

jeong-sik:
Addressed the durability review in 47058748b6.

Changes:
- Replaced Thompson prior persistence plain Fs_compat.save_file with Fs_compat.save_file_atomic.
- Routed atomic-write Error msg to Log.Thompson.error while preserving Eio.Cancel.Cancelled re-raise.
- Confirmed load_stats parses JSONL rows through stats_of_json/filter_map, so malformed rows are skipped rather than blindly installed.

Local checks:
- opam exec -- ocamlformat --check lib/thompson/thompson_sampling.ml
- git diff --check

Focused test build was attempted with scripts/dune-local.sh build test/test_thompson_sampling.exe, but scripts/dune-local.sh refused because bare Dune processes are active on the machine. I left those processes untouched; CI remains the build/test authority for this PR.

jeong-sik:
## Close-out — HEAD `47058748b6`: atomic save + error logging

### [Resolved] `save_stats` now uses the durability SSOT
```ocaml
- Fs_compat.save_file path content;
+ (match Fs_compat.save_file_atomic path content with
+  | Ok () -> Log.Metrics.debug …
+  | Error msg -> Log.Thompson.error "Error saving stats: %s" msg)
```
Switched to `Fs_compat.save_file_atomic` (tmp→fsync→rename), consistent with #21807/#21818/#21798, and the `Error` path now logs at error level instead of being swallowed. My "non-atomic save" finding is addressed.

The second finding (the persistence regression test wasn't run locally) stands or falls on CI — please make sure CI actually runs `test_thompson_sampling.exe` green on this head before merge, since the local dune build was blocked earlier. Nothing else from me on this one.

anyang-keepers (approved):
## Review: APPROVED ✅

**PR #21832** — feat(thompson prior persistence)

- `save_stats`/`load_stats` via JSON on disk, `copy_stats` for atomic snapshot, `apply_vote_counts` with decay
- Shutdown hook registered once via atomic CAS
- Backward compatible: no existing callers changed

**No BLOCK.** Solid persistence addition.

jeong-sik:
Adversarial pass on current head `47058748b6`: no remaining code blocker found.

Checked:
- `save_stats` now uses `Fs_compat.save_file_atomic`, so the durability SSOT issue is addressed.
- `Fs_compat.load_jsonl` / `load_jsonl_diagnostics` explicitly log and drop malformed lines, so torn/malformed rows do not abort the whole load path.
- Pending votes are overlaid only on copied stats during persistence; `flush_pending_votes` clears pending votes under the same mutex before applying them, so the saved overlay does not create a normal-path double count.
- `record_selection` increments selection count once and then persists if configured.

Local validation:
- `opam exec -- ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`

Not run locally:
- `scripts/dune-local.sh build test/test_thompson_sampling.exe` was blocked by the repo's bare-Dune guard (code 75) because several bare Dune processes are active. I did not bypass; CI should compile/test this head.

Residual non-blocking concern: write-per-event can still be chatty for hot fleets, but it is an operational cost issue rather than a correctness blocker.

anyang-keepers (dismissed):
## Adversarial Review (ramarama) — REQUEST_CHANGES

### BLOCK — `load_stats` error handling not visible (diff truncated)

`ensure_thompson_persistence` calls `Thompson_sampling.load_stats ()` during server bootstrap. If the stats file is corrupted, missing, or contains a version-mismatched schema, what happens?

- If `load_stats` raises an exception: server bootstrap crashes.
- If `load_stats` silently ignores errors: priors reset to defaults on every restart, making persistence meaningless.

**Need:** `load_stats` must be safe-on-boot. Either:
1. Return `(agent_stats list, string list) result` where errors are logged but don't crash, or
2. Wrap in `try ... with _ -> Lwt.return_unit` with a warning log.

This runs on EVERY server start. A corrupted stats file should not prevent startup.

### COMMENT — `copy_stats` deep-copy semantics

The copy is correct for snapshot-before-mutate. However, OCaml record fields are mutable by default — the copy creates a new record with the same values, but if the original contains mutable references (e.g., Hashtbl.t, ref), both copies share the same underlying mutable cell. For `agent_stats`, all fields are floats/ints/strings, so this is safe. Worth a comment: `(* deep-copy safe: all fields are immutable primitives *)`.

### COMMENT — shutdown priority 24

Is priority 24 late enough? If keeper turns are still running when priority 24 fires, the in-memory stats won't reflect the last few turns. Verify that priority 24 runs after all keeper fibers are joined.

### COMMENT — Atomic CAS for double-registration

Correct use. `compare_and_set false true` ensures exactly one shutdown hook registration. The double `load_stats` call (if two fibers race) is idempotent. Good.

**BLOCK on `load_stats` error handling.**

jeong-sik:
Review response pushed: `238c159468` (`fix(thompson): skip invalid persisted stats rows`).

Addressed the new load-safety REQUEST_CHANGES:
- `load_stats` was already wrapped in `try ... with` and re-raises only `Eio.Cancel.Cancelled`, so corrupted/read-error stats cannot crash server bootstrap; they log through `Log.Thompson.error`.
- Strengthened schema handling: `stats_of_json` now rejects rows with missing/empty `name` instead of installing an empty-name agent prior.
- Added `test_persistence_load_skips_corrupt_and_nameless_rows`: writes a JSONL file with a malformed line, a nameless schema row, and one valid row; `load_stats ()` must not raise, must load the valid row, and must not install an empty-name stats row.
- Added a short `copy_stats` comment for the deep-copy concern: `agent_stats` contains only primitive fields, so the record copy cannot share nested mutable cells.

Local checks run:
- `opam exec -- ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`
- `opam exec -- ocamlc -stop-after parsing lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`

Focused Dune build was attempted again with `scripts/dune-local.sh build test/test_thompson_sampling.exe`; wrapper refused because unrelated bare `dune build --root .` processes are active. Full compile/test remains CI-owned per operator direction.

jeong-sik:
Follow-up verification for the REQUEST_CHANGES point on current head `238c159468`.

I do not see a remaining boot-safety blocker in the current code:

- `ensure_thompson_persistence` calls `Thompson_sampling.load_stats ()` during bootstrap.
- `load_stats` wraps the existing-file load in `try ... with`; it re-raises only `Eio.Cancel.Cancelled` and logs any other exception through `Log.Thompson.error`, so a corrupt/read-error stats file does not crash bootstrap.
- Row-level malformed JSON is handled by `Fs_compat.load_jsonl` / `load_jsonl_diagnostics`, whose contract is “malformed lines are logged and dropped.”
- Schema-invalid rows are now rejected by `stats_of_json` when `name` is missing/blank.
- `test_persistence_load_skips_corrupt_and_nameless_rows` writes malformed JSON, a nameless row, and a valid row, then calls `load_stats ()`; the test would fail on raise, and it asserts the valid row loads while no empty-name stats row is installed.

Other review notes:
- `copy_stats` now has the requested deep-copy comment and the record is primitive-field only, so snapshot copies cannot alias nested mutable cells.
- The shutdown hook is wired into the production path: `Shutdown_hooks.run_all ()` calls `Shutdown.run_registered_hooks ()`, and registered hooks are sorted by ascending priority. `thompson_sampling_save` at priority 24 runs in that registered-hook phase, alongside existing priority-24 flush hooks and before priority-25/30 hooks. The main SIGTERM/SIGINT path invokes `Shutdown_hooks.run_all` before board flush/exit.
- Atomic CAS for single hook registration remains correct.

Local verification:
- `git diff --check`
- `opam exec -- ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`
- `opam exec -- ocamlc -stop-after parsing lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`
- diff-scoped path/env scan: no added `/Users/*`, `/tmp/masc`, `~/me`, `default_base`, or new `MASC_*` knobs.

CI note: GitHub checks for this new head are still in progress; reviewDecision remains `CHANGES_REQUESTED` until reviewer re-review clears it.

jeong-sik:
Addressed the load_stats boot-safety blocker in `fa44889e5a`.

- `load_stats` now uses `Fs_compat.load_jsonl_diagnostics`, logs malformed JSONL counts, and still does not abort startup on remaining I/O/schema exceptions.
- `stats_of_json` now rejects schema-mismatched rows that have a name but lack required `alpha`/`beta`, instead of silently loading them as near-default priors.
- Added regression coverage for corrupt JSONL plus schema-mismatched rows.

Local checks run:
- `ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml`
- `ocamlc -stop-after parsing lib/thompson/thompson_sampling.ml`
- `ocamlc -stop-after parsing test/test_thompson_sampling.ml`
- `git diff --check`

Focused Dune target was attempted and stopped after >90s with no compiler output; leaving full build/test to CI per operator preference.

jeong-sik:
Latest-head follow-up on `fa44889e5a`.

No remaining code blocker found for the boot-safety review point:
- `load_stats` still sits inside the boot-safe `try ... with` boundary and re-raises only `Eio.Cancel.Cancelled`; other I/O/schema failures log and do not abort startup.
- The new `Fs_compat.load_jsonl_diagnostics` path logs malformed JSONL line count before installing valid rows.
- `stats_of_json` now rejects rows missing `name`, `alpha`, or `beta`, so version/skewed schema rows are not loaded as misleading near-default priors.
- The regression now covers malformed JSONL, nameless rows, schema-mismatched named rows, and a valid row that still loads.

Local verification on this head:
- `opam exec -- ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`
- `opam exec -- ocamlc -stop-after parsing lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml lib/server/server_runtime_bootstrap.ml`

Full Dune/test execution remains CI-owned per operator direction; CI is still running on this head. ReviewDecision remains `CHANGES_REQUESTED` until the old review is re-reviewed/dismissed.

jeong-sik:
Current head `238c159468`: the invalid persisted-stats row gap is closed.

Confirmed:

- `save_stats` already uses `Fs_compat.save_file_atomic`.
- `Fs_compat.load_jsonl` drops malformed JSONL lines instead of aborting the whole load.
- `stats_of_json` now rejects missing/blank `name`, so a schema-invalid row cannot install an empty-name prior.
- The new regression writes malformed JSON, a nameless row, and a valid row; `load_stats` must load the valid row and not create an empty-name stats entry.

`git diff --check origin/main...refs/remotes/pr/21832`: PASS.

No remaining code blocker from my side. CI still needs to be the compile/test authority because local focused Dune remains blocked by unrelated bare Dune guard state.

jeong-sik:
Current head `fa44889e5a`: invalid persisted Thompson stats rows are now handled more strictly.

The latest follow-up improves the load guard:

- `load_stats` uses `Fs_compat.load_jsonl_diagnostics`, logs malformed row count, and continues.
- `stats_of_json` rejects rows missing either `alpha` or `beta` instead of silently installing a default-prior row for a schema mismatch.
- The test now covers malformed JSON, nameless rows, schema-mismatch rows, and a valid row after them.

`git diff --check origin/main...refs/remotes/pr/21832`: PASS.

No remaining code blocker from my side. Full compile/test signal still needs CI because local focused Dune is blocked by unrelated bare-Dune guard state.

jeong-sik:
## Close-out — HEAD `fa44889e5a`: make prior load schema-safe (+ `238c159468` skip invalid rows)

This closes a real silent-failure in the Thompson prior loader. Two changes, both in the right direction.

### [Resolved] alpha/beta missing → reject the whole row, not 0.0-default it
Before, a persisted row with a missing `alpha`/`beta` was silently coerced to `0.0`:
```ocaml
let alpha = Json_util.get_float json "alpha" |> Option.value ~default:0.0 in
let beta  = Json_util.get_float json "beta"  |> Option.value ~default:0.0  in
```
After:
```ocaml
else match Json_util.get_float json "alpha", Json_util.get_float json "beta" with
| None, _ | _, None -> None      (* reject the row *)
| Some alpha, Some beta -> …
```
This is the correct fix. `alpha`/`beta` are the Beta-distribution shape parameters — they're not counters where `0` is a benign "nothing yet" value; `alpha=0, beta=0` is an invalid prior that would corrupt the sampling statistics silently. Coercing a missing shape parameter to `0.0` is exactly the "unknown → permissive default" anti-pattern, and returning `None` for the whole row (so the loader falls back to the default prior for that agent) is the fail-closed alternative. Good.

The differential strictness is also defensible: `alpha`/`beta` must be present and valid (shape params), while `selections`/`last_selected_at`/`total_votes_*`/etc. still use `Option.value ~default:0` because those are genuine counters/timestamps where `0` is the meaningful "none yet" value (and `last_selected_at = 0.0` is a reasonable "never selected" sentinel). So the change tightens exactly the fields where a default would be wrong and leaves the counter fields alone — that's a principled split, not an oversight.

### [Resolved] Malformed JSONL rows are now visible, not silently swallowed
`load_stats` switched from `Fs_compat.load_jsonl` to `load_jsonl_diagnostics` (which returns the count of malformed rows alongside the entries) and logs `Dropped %d malformed stats line(s)` at warn when that count is non-zero. Dropping corrupt lines at boot is correct (one bad line shouldn't abort server startup), but dropping them *silently* would have hidden a real persistence/serialization bug behind a "stats look fine" dashboard. The warn makes it observable. The added test (`test_persistence_load_skips_corrupt_and_nameless_rows`, extended in this commit to also cover a schema-mismatch row with `score` instead of `alpha`/`beta`) locks that the valid row still loads and the corrupt/nameless/schema-mismatch rows are all excluded.

Net: two genuine robustness improvements (fail-closed on the shape params, visible drop on malformed lines) with test coverage. No concern from me; this takes the prior-loader from "silently degrades" to "fails closed and tells you."

jeong-sik:
### CHANGE REQUEST (head `fa44889e5a`) — `load_stats` merges into a global table, so priors can leak across base paths

The atomic-write and schema-safe-row fixes are good, but the loader still has a state-boundary bug.

`Thompson_sampling` is a process-global module:

- `lib/thompson/thompson_sampling.ml:65` global `stats_table`
- `lib/thompson/thompson_sampling.ml:68` global `pending_votes`
- `lib/thompson/thompson_sampling.ml:71` global `base_path_ref`

The new bootstrap path does:

- `server_runtime_bootstrap.ml:221-223` `set_base_path base_path; load_stats ()`

But `set_base_path` only changes the ref:

- `thompson_sampling.ml:79-80` `base_path_ref := Some path`

and `load_stats` only installs parsed rows with `Hashtbl.replace`:

- `thompson_sampling.ml:323-325` load is conditional on the file existing
- `thompson_sampling.ml:349` `Hashtbl.replace stats_table s.name s`

It never clears `stats_table` or `pending_votes` for the new base. If the same process initializes Thompson persistence for base A, then later initializes for base B (tests, alternate workspace, restart-in-process, or any multi-base server harness), base A rows remain in memory. If base B has no `autonomy_stats.jsonl`, `load_stats` does nothing, so the next `save_stats_if_configured` writes base A priors into base B's `.masc/autonomy_stats.jsonl`.

This violates the base-path boundary the PR is trying to persist under. It is the same class of bug as the other MASC path-scoping issues: a caller-supplied base path should be an isolation boundary, not just the path where the next save happens.

Required fix:

- either make `load_stats` replace the in-memory table with the loaded file contents for the current base, including the empty/missing-file case;
- or make `set_base_path` reset all per-base Thompson state (`stats_table` and `pending_votes`) when the base changes, then call `load_stats`.

Please add a regression with two temp base paths:

1. base A has a persisted stats row for `agent-a`; load it.
2. switch to base B with no stats file; call the same bootstrap/load path.
3. assert `get_all_stats ()` no longer contains `agent-a`.
4. record/save something in base B and assert base B's file does not contain base A's row.

jeong-sik:
Addressed the base-path state isolation review in `49f5f25c9f`.\n\nChanges:\n- `set_base_path` now clears in-memory `stats_table` and `pending_votes` when switching to a different base path.\n- `load_stats` now replaces the current in-memory snapshot with the loaded file contents instead of merging into stale global state; missing files replace the snapshot with empty state.\n- Added a regression with two temp base paths: load base A, add an unflushed pending vote for A, switch to empty base B, then save B and assert neither A loaded rows nor A pending rows appear in B memory/file state.\n\nValidation:\n- `env DUNE_BUILD_DIR=_build-review-21832 DUNE_CACHE=disabled dune build --root . test/test_thompson_sampling.exe`\n- `./_build-review-21832/default/test/test_thompson_sampling.exe` — 31 tests passed\n- `ocamlformat --check lib/thompson/thompson_sampling.ml test/test_thompson_sampling.ml`\n- `python3 scripts/check_test_coverage.py`\n- `git diff --check`\n- `git diff --check origin/main...HEAD`

jeong-sik:
## Adversarial review — HEAD `49f5f25c9f`: isolate persisted priors by base path (+24/-4) — and a gap my earlier review missed

**Affirm on the fix, and an honest note: my `fa44889e5a` review of this PR checked the persistence path but missed the in-memory side of the same isolation bug.** The author caught and fixed what I didn't flag.

### [Affirm] In-memory stats now track base-path changes instead of going stale
The bug this fixes is the in-memory half of workspace isolation. `stats_table`/`pending_votes` are process-global Hashtbls, but `set_base_path` used to do only `base_path_ref := Some path` — so when a server switched `base_path` to a different workspace, the Hashtbl kept the *previous* workspace's agent stats. Worse, `load_stats` did `Hashtbl.replace` over the existing table, so agents present in the old file but absent from the new one **survived the reload** as stale rows. Net effect: a multi-workspace server would silently mix two workspaces' Thompson priors in memory, biasing sampling toward whichever workspace loaded first.

The fix is correct on both axes:
- `set_base_path` now `clear_in_memory_stats_unlocked` when the path actually changes (`changed` check), so a workspace switch wipes the in-memory state before the new path takes effect;
- `replace_stats_snapshot` (clear, then replace) makes `load_stats` a true snapshot swap instead of an overlay, so agents gone from the new file are gone from memory;
- the `else begin … replace_stats_snapshot [] … end` branch handles the "stats file doesn't exist for this workspace" case — previously that path did nothing, leaving the old workspace's stats in memory. Now an empty workspace loads an empty table.

All three under the existing `ts_mu` write lock, so no concurrency regression. Good.

### [Self-correction] My `fa44889e5a` review missed this
When I affirmed the schema-safe load + malformed-row diagnostics, I checked the *file* path (what gets read, what gets rejected, what gets logged) but not the *memory* path — specifically, whether `set_base_path`/`load_stats` keep the in-memory Hashtbl consistent with the on-disk workspace. They didn't, and that's the same family of isolation bug as #21798's read/write path asymmetry (which I did catch there) — I just didn't look for the in-memory analog here. Credit to the author for finding and fixing it; lesson for me is to extend the "is this state workspace-scoped?" check from the storage path to the in-memory caches that mirror it. Same root cause (state not keyed to the active workspace), different surface (path source vs. Hashtbl lifetime).

### [Low / verify] Confirm the multi-workspace test covers the switch
The +62 test lines should include a "load workspace A, switch base_path to B, assert A's agents are gone from memory / B's are loaded" sequence — that's the regression that would catch a revert of `clear_in_memory_stats_unlocked` or `replace_stats_snapshot`. If it's only single-workspace load/skip tests, the isolation guarantee rests on the code, not the build. Non-blocking; please confirm the switch scenario is asserted.

Net: the isolation fix is correct (base-path-change clear + snapshot swap + empty-file clear), and it closes a real multi-workspace contamination bug my earlier review of this PR overlooked. I'd merge; the only ask is confirming the workspace-switch test pair.

jeong-sik:
Confirmed the multi-workspace regression covers the switch scenario the Low asked about.

On current head `49f5f25c9f`, `test_persistence_base_path_switch_replaces_state` does all four relevant checks:
- seeds base A with `persist-base-a-agent`, then `set_base_path base_a; load_stats ()` and confirms A loaded;
- records an unflushed pending vote for `persist-base-a-pending`;
- switches to empty base B with `set_base_path base_b; load_stats ()`;
- asserts both the loaded A row and the pending A row are absent from in-memory stats, then records/saves B and asserts base B file contains only the B row and neither A row.

So the switch case is pinned directly in `test/test_thompson_sampling.ml` by `test_persistence_base_path_switch_replaces_state`. No code patch needed for this Low.

PR #21851, opened 2026-06-20
[codex] Preserve missing tool-call timing state

jeong-sik:
## Summary

Follow-up to #21841 review feedback.

- keep missing or malformed `/tool-calls` `duration_ms` as `null` instead of decoding it to measured `0ms`
- preserve tool-call timing-source fetch failures separately from "loaded successfully but no matching timing"
- surface timing-source degradation in the turn inspector while keeping turn records usable
- update tool-call inspector, chat trace steps, and session-trace enrichment to treat missing duration as unmeasured instead of numeric zero

## Why

#21841 removed the visible `0.50s` fabrication, but review found two remaining truth-boundary leaks:

1. the shared tool-call decoder still converted missing/invalid `duration_ms` into `0`
2. a `/tool-calls` fetch failure was indistinguishable from a successful empty timing source

This keeps timing provenance honest at the data boundary and on the operator surface.

## Validation

- `git diff --cached --check`
- `git diff --check`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-turn-inspector.test.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard exec eslint src/api/dashboard.ts src/components/keeper-turn-inspector.ts src/components/keeper-turn-inspector.test.ts src/components/keeper-tool-call-inspector.ts src/components/chat/primitives.ts src/components/session-trace/session-trace-state.ts`

Full build is left to CI.

### DIFF
--- a/dashboard/src/api/dashboard.test.ts
+++ b/dashboard/src/api/dashboard.test.ts
@@ -396,6 +396,40 @@ describe('keeper tool telemetry fetchers', () => {
     expect(entry?.semantic_outcome).toBe('blocked')
     expect(entry?.goal_ids).toEqual(['g-1', 'g-2'])
   })
+
+  it('keeps missing or malformed tool-call duration unmeasured', async () => {
+    const fetchMock = vi.fn(() => Promise.resolve(
+      new Response(JSON.stringify({
+        keeper: 'keeper-alpha',
+        count: 2,
+        source: 'tool_call_io',
+        entries: [
+          {
+            ts: 1,
+            keeper: 'keeper-alpha',
+            tool: 'keeper_context_status',
+            input: {},
+            output: 'ok',
+            success: true,
+          },
+          {
+            ts: 2,
+            keeper: 'keeper-alpha',
+            tool: 'keeper_board_post_get',
+            input: {},
+            output: 'ok',
+            success: true,
+            duration_ms: 'not recorded',
+          },
+        ],
+      }), { status: 200, headers: { 'Content-Type': 'application/json' } }),
+    ))
+    vi.stubGlobal('fetch', fetchMock)
+
+    const result = await fetchKeeperToolCalls('keeper-alpha')
+
+    expect(result.entries.map(entry => entry.duration_ms)).toEqual([null, null])
+  })
 })
 
 describe('fetchMemorySubsystems', () => {
--- a/dashboard/src/api/dashboard.ts
+++ b/dashboard/src/api/dashboard.ts
@@ -2771,7 +2771,7 @@ export type ToolCallEntry = {
   input: unknown
   output: string | ToolCallOutputBlob
   success: boolean
-  duration_ms: number
+  duration_ms: number | null
   model?: string
   trace_id?: string
   session_id?: string
@@ -2842,7 +2842,7 @@ function decodeToolCallEntry(raw: unknown): ToolCallEntry | null {
     input: raw.input,
     output: decodeToolCallOutput(raw.output),
     success: asBoolean(raw.success, false),
-    duration_ms: asNumber(raw.duration_ms, 0),
+    duration_ms: asNumber(raw.duration_ms) ?? null,
     model: asString(raw.model),
     trace_id: asString(raw.trace_id),
     session_id: asString(raw.session_id),
--- a/dashboard/src/components/chat/primitives.ts
+++ b/dashboard/src/components/chat/primitives.ts
@@ -2060,7 +2060,7 @@ function ToolTraceStep({ entry, output }: { entry: KeeperConversationEntry; outp
       : output.success === false || output.semantic_success === false
         ? 'bad'
         : 'ok'
-  const durLabel = output && output.duration_ms > 0 ? formatMsCompact(output.duration_ms) : ''
+  const durLabel = output?.duration_ms != null && output.duration_ms > 0 ? formatMsCompact(output.duration_ms) : ''
   const resultView = output ? toolOutputDisplay(output.output) : null
   const hasResult = resultView !== null && resultView.text.trim() !== ''
   // Expandable when there is anything to show: args, a result, or a still-pending
--- a/dashboard/src/components/keeper-tool-call-inspector.ts
+++ b/dashboard/src/components/keeper-tool-call-inspector.ts
@@ -30,6 +30,7 @@ import { isKeeperToolActivityEvent, sseEventMatchesKeeper } from './keeper-sse-m
 
 // Delegated to lib/format-time (SSOT)
 const formatTimestamp = formatTimeHms
+const NO_DURATION_LABEL = '—'
 
 function FreshnessLine({ data }: { data: TelemetryFreshnessMetadata }) {
   const gap = coverageGapDisplay(data)
@@ -276,7 +277,7 @@ export function deriveKeeperToolCallDossier(
       label: 'latest',
       value: latest?.tool ?? 'none',
       detail: latest
-        ? `${formatTimestamp(latest.ts)} · ${toolCallStatusLabel(latest)} · ${latest.duration_ms != null ? formatMsCompact(latest.duration_ms) : 'duration 없음'}`
+        ? `${formatTimestamp(latest.ts)} · ${toolCallStatusLabel(latest)} · ${latest.duration_ms != null ? formatMsCompact(latest.duration_ms) : NO_DURATION_LABEL}`
         : 'no recent tool call',
       tone: latestTone,
       title: latest ? entryScopeLabel(latest) : undefined,
@@ -293,7 +294,7 @@ export function deriveKeeperToolCallDossier(
     {
       key: 'slowest',
       label: 'slowest',
-      value: slowest?.duration_ms != null ? formatMsCompact(slowest.duration_ms) : 'none',
+      value: slowest?.duration_ms != null ? formatMsCompact(slowest.duration_ms) : NO_DURATION_LABEL,
       detail: slowest ? `${slowest.tool} · ${entryScopeLabel(slowest)}` : 'no duration sample',
       tone: slowest?.duration_ms != null ? durationTone(slowest.duration_ms) : 'neutral',
     },
@@ -498,7 +499,7 @@ function ToolCallRow({ entry }: { entry: ToolCallEntry }) {
         <span class="font-mono text-[var(--color-fg-secondary)] flex-shrink-0 w-16">${formatTimestamp(entry.ts)}</span>
         <span class="font-mono font-medium text-[var(--color-fg-secondary)] truncate flex-1" title=${entry.tool}>${entry.tool}</span>
         <span class=${`font-mono flex-shrink-0 w-16 text-right ${entry.duration_ms != null ? durationColor(entry.duration_ms) : 'text-[var(--color-fg-disabled)]'}`}>
-          ${entry.duration_ms != null ? formatMsCompact(entry.duration_ms) : '—'}
+          ${entry.duration_ms != null ? formatMsCompact(entry.duration_ms) : NO_DURATION_LABEL}
         </span>
         <span
           class=${`flex-shrink-0 w-5 text-center ${toolCallSucceeded(entry) ? 'text-[var(--color-status-ok)]' : 'text-[var(--color-status-err)]'}`}
--- a/dashboard/src/components/keeper-tool-call-inspector.ts
+++ b/dashboard/src/components/keeper-tool-call-inspector.ts
@@ -154,7 +154,8 @@ function newestToolCall(entries: readonly ToolCallEntry[]): ToolCallEntry | null
 function slowestToolCall(entries: readonly ToolCallEntry[]): ToolCallEntry | null {
   let slowest: ToolCallEntry | null = null
   for (const entry of entries) {
-    if (slowest === null || entry.duration_ms > slowest.duration_ms) slowest = entry
+    if (entry.duration_ms == null) continue
+    if (slowest === null || slowest.duration_ms == null || entry.duration_ms > slowest.duration_ms) slowest = entry
   }
   return slowest
 }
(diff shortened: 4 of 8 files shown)


jeong-sik:
[적대 리뷰 grounding · head `d96db680b` · 미리뷰 PR 첫 리뷰 · honest-null 정석 구현 affirm (codex PR)]

#21841(0.50s fabrication 제거) 후속으로, missing/invalid timing을 `0ms`(measured)가 아니라 `null`(unmeasured)로 보존하는 truth-boundary fix입니다. unknown→permissive-default 안티패턴의 **정반대**로 정확히 구현됐습니다 — codex PR이 안티패턴을 *도입*이 아니라 *제거*합니다.

**[Affirm · union type으로 "unmeasured"를 표현 가능하게]**
```ts
- duration_ms: number
+ duration_ms: number | null
...
- duration_ms: asNumber(raw.duration_ms, 0),        // missing → 측정된 0ms로 날조
+ duration_ms: asNumber(raw.duration_ms) ?? null,    // missing → unmeasured(null)
```
타입을 `number | null`로 넓혀 "측정 안 됨"을 **표현 가능한 상태**로 만들고(make illegal states representable의 올바른 방향 — 0ms와 unmeasured를 타입으로 구분), 컴파일러가 모든 consumer에서 null 처리를 강제합니다.

**[Affirm · 모든 consumer가 null 분기 (sentinel-0 잔재 없음)]** decoder뿐 아니라 소비 지점 전수 갱신 확인:
- `slowestToolCall`: `if (entry.duration_ms == null) continue` — null은 "가장 느림" 비교에서 제외(0으로 잘못 포함 안 함).
- 라벨: `duration_ms != null ? formatMsCompact(...) : 'duration 없음' / 'none'` — null을 0ms로 렌더하지 않고 명시적 "없음".
- `durationColor`/`durationTone`: null → disabled 색/neutral tone.
- `output.duration_ms > 0` 가드도 `output?.duration_ms != null && ... > 0`로 null-safe.
테스트(`keeps missing or malformed tool-call duration unmeasured`)가 `duration_ms: 'not recorded'`(malformed) + missing → `[null, null]` 단언. 0-sentinel로 회귀하면 깨집니다.

**[Affirm · fetch 실패와 "로드됐으나 매칭 timing 없음" 분리]** body대로 timing-source fetch 실패를 "성공 로드 + 매칭 없음"과 별도 상태로 보존 — 두 다른 실패를 한 값으로 뭉개지 않는 것도 truth-boundary 정합입니다. 새 env/하드코딩 경로 0건.

정리: missing timing을 `null`(unmeasured)로 두고 union type + 전 consumer null 분기 + 회귀 테스트로 잠근, parse-don't-validate 정석입니다. #21841의 0.50s 날조 제거를 데이터 경계까지 완결합니다. 막을 사유 없음.

(근거: `dashboard.ts` `duration_ms: number|null`, `asNumber(...) ?? null`; `keeper-tool-call-inspector.ts`/`primitives.ts`/`session-trace-state.ts` consumer null 분기; `dashboard.test.ts` `[null,null]` 단언. 차등 적대 검증 2026-06-21, head d96db680b.)

jeong-sik:
Second adversarial pass:

- Rechecked the current head d96db680bb against the honest-null timing boundary.
- Verified the decoder keeps missing/malformed duration_ms as null and consumers do not reintroduce 0ms sentinel rendering.
- Rechecked the session-trace enrichment paths; ToolCallEntry null durations are converted to undefined where the trace shape expects optional numeric duration.
- Focused path/env scan only found pre-existing dashboard.test fixture keys outside this PR diff; no new local path or env knob additions.

Verification:
- pnpm --dir dashboard exec tsc --noEmit --pretty false PASS
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-turn-inspector.test.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1 PASS
- pnpm --dir dashboard exec eslint src/api/dashboard.ts src/components/keeper-turn-inspector.ts src/components/keeper-turn-inspector.test.ts src/components/keeper-tool-call-inspector.ts src/components/chat/primitives.ts src/components/session-trace/session-trace-state.ts PASS
- git diff --check origin/main...HEAD PASS

No patch needed from this pass. CI still has Build and Test / Dashboard / TLA+ running.

jeong-sik:
## Blocking — missing `/tool-calls` duration can erase an existing trace duration

The main boundary fix is right: `ToolCallEntry.duration_ms` is now `number | null`, and the decoder no longer fabricates `0ms` for missing/malformed values.

But the session-trace enrichment path now loses a real duration when the richer `/tool-calls` row lacks one.

Current code:

```ts
// session-trace-state.ts
if (
  event.duration_ms != null
  && entry.duration_ms != null
  && Math.abs(event.duration_ms - entry.duration_ms) > 1
) return false
...
return {
  ...event,
  ...
  duration_ms: entry.duration_ms ?? undefined,
}
```

That means a trajectory event with `event.duration_ms = 812` is allowed to match a `/tool-calls` entry whose `duration_ms` is `null` (correct), but then `enrichToolCallTrace` overwrites the event with `duration_ms: undefined` (incorrect). The durable tool-call log should not delete a measured duration from the trace just because that secondary timing source is missing.

Fix direction:

```ts
duration_ms: entry.duration_ms ?? event.duration_ms,
```

and add a regression where a trace tool event has `duration_ms`, the matching tool-call entry has `duration_ms: null`, and the enriched event keeps the trace duration. Synthetic trace rows from a bare tool-call entry can still use `entry.duration_ms ?? undefined` because there is no source event there.

This is still the same truth-boundary theme as the PR: missing timing must stay missing, but it must not erase timing that was already present from a different source.

anyang-keepers (commented):

NOTE: Diff was truncated in my fetch. Reviewing visible portion only.

### COMMENT — incomplete review

The diff for this PR exceeded my fetch limit. I can see the PR title (`[codex] Preserve missing tool-call timing state`) but not the full implementation.

Re-reviewing after truncation is resolved.

jeong-sik:
## Adversarial review — HEAD `d96db680bb`: preserve missing tool-call timing state (+145/-15)

**Affirm on the core change.** Making `duration_ms` explicitly `number | null` and guarding every consumer is the right way to stop a missing/failed timing source from silently corrupting derived values — the old shape forced `duration_ms` into numeric comparisons where `undefined`/`NaN` would either throw or produce a nonsense "slowest" entry. Three things are done well:

- `decodeToolCallEntry`: `asNumber(raw.duration_ms) ?? null` — missing/unparseable becomes `null`, not a coerced `NaN`. Fail-closed at the decode boundary.
- `slowestToolCall`: `if (entry.duration_ms == null) continue` plus the `slowest?.duration_ms == null` guard — null-duration entries are excluded from the slowest computation instead of winning it (which a `NaN` would have). Correct.
- `toolCallError` + the "tool-call timing source unavailable" banner — the timing fetch failing no longer hides the turn records; the records stay visible and the failure is surfaced. That's visible-failure, not silent-failure, and the test (`dashboard.test.ts` `toEqual([null, null])` + the inspector test for the warning) locks it.

### [Low, change-requested] The "no duration" rendering drifts across three different labels
The same logical state — `duration_ms == null` — is rendered three different ways in the same inspector surface:

| site | label for `duration_ms == null` |
|------|---------------------------------|
| `keeper-tool-call-inspector.ts` dossier line | `'duration 없음'` (Korean) |
| same file, slowest stat | `'none'` |
| same file, `ToolCallRow` | `'—'` |

That's vocabulary drift on one concept inside one component: a future change to the "no duration" wording has to touch three sites, and they already disagree (Korean prose vs. English `none` vs. em-dash) in the same drawer. Extract one constant/helper (e.g. `const noDurationLabel = '—'` or a `formatDurationMs(ms: number | null)` that returns the label for null) and route all three through it, so there's one rendering per state. Minor, but it's the same hand-maintained-mirror shape that bites elsewhere — easier to close now than after more call sites appear.

### [Low] Null-guards are scattered but locally correct
`entry.duration_ms != null` / `slowest?.duration_ms != null` appears at ~8 sites. Each is correct in context, and the three renderings differ enough (`>0` check in `primitives.ts`, `>= 2_000` threshold in the dossier, color class in the row) that a single helper wouldn't fully replace them — but the `!= null` predicate itself is the duplicated bit. A tiny `hasDuration(e): e is ToolCallEntry & { duration_ms: number }` type guard would centralize the null-check and give a narrowing, which TS would then track. Non-blocking; only worth it if you're already touching these.

Net: the integrity fix (explicit null + guards + visible warning) is correct and I'd merge it; the only ask is collapsing the three "no duration" labels into one source before they drift further.

jeong-sik:
Review response / adversarial follow-up:

- Fixed the blocking trace-duration erasure: `enrichToolCallTrace` now keeps the existing trajectory duration when the richer `/tool-calls` row has `duration_ms = null`. Synthetic rows still use the tool-call row source only.
- Added regression coverage for trajectory `duration_ms = 812` + matched tool-call log `duration_ms = null`, asserting the enriched trace keeps `812`.
- Collapsed the inspector null-duration rendering to one `NO_DURATION_LABEL` constant across latest/slowest/row labels.
- Path/env adversarial scan: removed newly added absolute path fixtures from the diff; no new `/Users`, `/home`, `/tmp`, env knob, or `default_base` additions.

Verification:
- `pnpm --dir dashboard exec tsc --noEmit --pretty false` PASS
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/session-trace/session-trace-state.test.ts src/components/keeper-turn-inspector.test.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1` PASS (98 tests)
- `pnpm --dir dashboard exec eslint src/components/session-trace/session-trace-state.ts src/components/session-trace/session-trace-state.test.ts src/components/keeper-tool-call-inspector.ts` PASS
- `git diff --check origin/main...HEAD` PASS

Pushed commit: `c5f45e5f0c`

jeong-sik:
[codex follow-up adversarial review · head c5f45e5f0c]

Re-reviewed the latest delta `d96db680bb..c5f45e5f0c` against the two previous asks.

- Blocking trace-duration erasure is fixed: `enrichToolCallTrace` now uses `entry.duration_ms ?? event.duration_ms`, so a null `/tool-calls` timing row no longer deletes a measured trajectory duration.
- Regression coverage is present for trajectory `duration_ms = 812` plus matching tool-call log `duration_ms = null`, asserting the enriched event keeps `812` while still taking richer args/result from the tool-call log.
- Synthetic tool-call rows remain source-honest: the synthetic path still uses `entry.duration_ms ?? undefined` because there is no trajectory event to preserve.
- Null-duration inspector rendering is now centralized through `NO_DURATION_LABEL` for latest/slowest/row display.
- The new absolute `/tmp` test fixture paths were removed from the diff.

Verification on current head:
- `pnpm --dir dashboard exec tsc --noEmit --pretty false` PASS
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/session-trace/session-trace-state.test.ts src/components/keeper-turn-inspector.test.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1` PASS, 98 tests
- `pnpm --dir dashboard exec eslint src/components/session-trace/session-trace-state.ts src/components/session-trace/session-trace-state.test.ts src/components/keeper-tool-call-inspector.ts` PASS
- `git diff --check origin/main...HEAD` PASS

No remaining blocker from this pass. CI still has some in-progress jobs on this draft PR.

jeong-sik:
Close-out on current head `c5f45e5f0c`.

The previous blocker I left on this PR is resolved. The enrichment path now preserves the trajectory duration when the richer `/tool-calls` row has no duration:

```ts
duration_ms: entry.duration_ms ?? event.duration_ms
```

I also checked the new regression coverage: it reproduces a trajectory row with `duration_ms: 812` and a matching tool-call log row with `duration_ms: null`, then asserts the merged trace event keeps `812`. The synthetic tool-call path now emits `undefined` rather than leaking a nullable duration into the trace row, and the inspector uses a single `NO_DURATION_LABEL` for missing duration display.

`git diff --check origin/main...refs/remotes/pr/21851` passes. I did not rerun the full dashboard test suite in this pass. No remaining blocker from my previous finding.

jeong-sik:
## Close-out — HEAD `c5f45e5f0c`: preserve trace duration during tool-call enrichment

This addresses both the label-drift Low from my earlier review *and* a separate duration-preservation gap, and the ready transition is justified by it. Closing the label-drift ask.

### [Resolved] The three "no duration" labels collapse to one constant
Exactly the fix I asked for:
```ts
const NO_DURATION_LABEL = '—'
…
: 'duration 없음'   →  : NO_DURATION_LABEL   // dossier latest
: 'none'            →  : NO_DURATION_LABEL   // dossier slowest
: '—'               →  : NO_DURATION_LABEL   // ToolCallRow
```
All three renderings of `duration_ms == null` now route through a single named constant, so the Korean/English/em-dash inconsistency (`'duration 없음'` vs `'none'` vs `'—'`) is gone and there's one place to change the "no duration" presentation. The null-duration state is now one concept with one label, which is what the no-drift ask was about.

### [Affirm] Enrichment no longer drops a duration it could have recovered
The second change is a genuine integrity fix, not just the label:
```ts
- duration_ms: entry.duration_ms ?? undefined,
+ duration_ms: entry.duration_ms ?? event.duration_ms,
```
When the richer tool-call-log row (`entry`) has `duration_ms: null`, the old code fell through to `undefined`, discarding the trajectory's (`event`) measured duration for that same execution. The new code falls back to the trajectory duration, so a tool call that *was* timed (just not in the tool-call log) still shows its real duration instead of degrading to "no duration." The test `preserves trajectory duration when the richer tool-call row has no duration` locks this (tool-call log `duration_ms: null` + trajectory `duration_ms: 812` → enriched row `812`).

The precedence is correct: tool-call log wins when it has a value (it's the richer source), trajectory fills the gap only when the log's value is null. And if *both* are absent, the value stays `undefined`, which renders as `NO_DURATION_LABEL` — a consistent end state. The `trace_origin: 'trajectory+tool_call_log'` detail already labels the merge, so an operator can tell the row is a join of two sources. No silent fabrication (it's a real measured value from the other source), no data loss.

### [Resolved, bonus] Local paths removed from the test fixtures
The `buildTraceEvents` tests dropped `/tmp/trajectory.txt` / `/tmp/full.txt` in favor of bare `trajectory.txt` / `full.txt` — machine-specific absolute paths out of the fixtures. Matches the local-path concern even though these are just test args.

Net: the label-drift Low is closed, the duration-preservation fix is correct and tested, and the local test paths are cleaned up. The ready transition is justified — no open concern from me.

jeong-sik:
Current head `c5f45e5f` summary for the truncated-review note:

This PR preserves timing-source truth instead of fabricating `0ms`:
- missing/malformed `/tool-calls` `duration_ms` stays `null` / unmeasured
- fetch failure is kept distinct from a successful empty timing source
- the turn inspector surfaces timing-source degradation while keeping turn records usable
- tool-call inspector, chat trace steps, and session trace enrichment consume the nullable timing contract

The PR body lists focused local checks (`vitest`, `eslint`, `tsc`) and CI is still pending on the current head. No active review thread remains.

PR #21863, opened 2026-06-20
[codex] Fix keeper history media reroute

jeong-sik:
## Summary

Fix keeper multimodal follow-up routing after an image turn.

The dashboard can successfully send an image turn, but that image remains in OAS `initial_messages`. A later text-only follow-up was routed as if it were text-only because `Keeper_turn_driver.run_named` only inspected current `goal_blocks`. Text-only runtimes such as `glm-5-turbo` then received historical image content and surfaced provider errors like:

`Invalid request: messages.content.type is invalid, allowed values: ['text']`

## Changes

- Use one active input view, `Runtime_agent.content_blocks_for_run`, for both keeper pre-dispatch reroute and the runtime capability floor.
- Route decisions now call `Runtime_agent.decide_modality_reroute_for_runtime ?initial_messages` directly so future callers do not have to remember to merge current goal blocks with history themselves.
- Include `initial_messages` content blocks when deciding whether a keeper turn needs media failover.
- Validate `initial_messages + goal_blocks` before provider dispatch so unsupported historical media fails closed inside MASC instead of leaking as provider 400.
- Add regression coverage for history-retained media driving reroute and runtime multimodal validation.

## Evidence

- Z.AI docs checked on 2026-06-21 KST:
  - `glm-5-turbo` guide lists text input only: https://docs.z.ai/guides/llm/glm-5-turbo.md
  - `glm-5v-turbo` guide lists image/video/text/file input: https://docs.z.ai/guides/vlm/glm-5v-turbo.md
  - Chat completion API separates text and vision request/model families: https://docs.z.ai/api-reference/llm/chat-completion.md

## Validation

- `ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli lib/keeper/keeper_turn_driver.ml test/test_runtime_modality_reroute.ml test/test_gate_keeper_backend.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_runtime_modality_reroute.exe test/test_gate_keeper_backend.exe`
- `./_build/default/test/test_runtime_modality_reroute.exe` (8 tests)
- `./_build/default/test/test_gate_keeper_backend.exe` (53 tests)

### DIFF
--- a/lib/keeper/keeper_turn_driver.ml
+++ b/lib/keeper/keeper_turn_driver.ml
@@ -125,35 +125,35 @@ let run_named
              (no silent fallback to default — RFC-0207/RFC-0206 §2.1)"
             runtime_id))
   | Some assigned_runtime ->
-  (* RFC-0265: proactively reroute a turn whose input modality (image/audio/
-     document) the assigned runtime cannot accept to a capable configured runtime
-     ([\[runtime\].media_failover] order, else declaration order). Text turns
-     ([goal_blocks = None]) and modality-satisfied turns are untouched; when no
-     runtime qualifies the assigned runtime stands and the loud capability gate in
-     [Runtime_agent.run_blocks] rejects (the floor). The reroute is visible via a
-     WARN log (non-silent — RFC-0126/0145). *)
+  (* RFC-0265: proactively reroute a turn whose active input modality
+     (image/audio/document) the assigned runtime cannot accept to a capable
+     configured runtime ([\[runtime\].media_failover] order, else declaration
+     order). The active input includes both the current goal and prior
+     [initial_messages]; otherwise an image in history can poison the next
+     text-only follow-up and leak as a provider 400. Modality-satisfied turns are
+     untouched; when no runtime qualifies the assigned runtime stands and the
+     loud capability gate in [Runtime_agent.run_blocks] rejects (the floor). The
+     reroute is visible via a WARN log (non-silent — RFC-0126/0145). *)
   let runtime_id, runtime =
-    match goal_blocks with
-    | None | Some [] -> runtime_id, assigned_runtime
-    | Some blocks ->
-      (match
-         Runtime_agent.decide_modality_reroute_for_runtime
-           ~assigned:assigned_runtime
-           blocks
-       with
-       | Runtime_agent.No_reroute_needed | Runtime_agent.No_capable_runtime _ ->
-         runtime_id, assigned_runtime
-       | Runtime_agent.Reroute { to_runtime_id; reason } ->
-         (match Runtime.get_runtime_by_id to_runtime_id with
-          | None -> runtime_id, assigned_runtime
-          | Some rerouted ->
-            Log.Keeper.warn
-              "%s: RFC-0265 modality reroute %s -> %s (%s)"
-              keeper_name
-              runtime_id
-              to_runtime_id
-              reason;
-            to_runtime_id, rerouted))
+    match
+      Runtime_agent.decide_modality_reroute_for_runtime
+        ~assigned:assigned_runtime
+        ~initial_messages
+        (Option.value goal_blocks ~default:[])
+    with
+    | Runtime_agent.No_reroute_needed | Runtime_agent.No_capable_runtime _ ->
+      runtime_id, assigned_runtime
+    | Runtime_agent.Reroute { to_runtime_id; reason } ->
+      (match Runtime.get_runtime_by_id to_runtime_id with
+       | None -> runtime_id, assigned_runtime
+       | Some rerouted ->
+         Log.Keeper.warn
+           "%s: RFC-0265 modality reroute %s -> %s (%s)"
+           keeper_name
+           runtime_id
+           to_runtime_id
+           reason;
+         to_runtime_id, rerouted)
   in
   let error_runtime_id = runtime_id in
   let candidate =
--- a/lib/keeper/keeper_turn_driver.ml
+++ b/lib/keeper/keeper_turn_driver.ml
@@ -134,12 +134,20 @@ let run_named
      untouched; when no runtime qualifies the assigned runtime stands and the
      loud capability gate in [Runtime_agent.run_blocks] rejects (the floor). The
      reroute is visible via a WARN log (non-silent — RFC-0126/0145). *)
+  let current_goal_blocks =
+    match goal_blocks with
+    | Some blocks -> blocks
+    | None ->
+      (* A missing current block payload means the current keeper goal is
+         text-only; [initial_messages] still participate in reroute below. *)
+      []
+  in
   let runtime_id, runtime =
     match
       Runtime_agent.decide_modality_reroute_for_runtime
         ~assigned:assigned_runtime
         ~initial_messages
-        (Option.value goal_blocks ~default:[])
+        current_goal_blocks
     with
     | Runtime_agent.No_reroute_needed | Runtime_agent.No_capable_runtime _ ->
       runtime_id, assigned_runtime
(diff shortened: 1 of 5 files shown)


jeong-sik:
## Adversarial review — HEAD `467ffe0aae`: keeper history media reroute

No code blocker found in the reroute/gate change itself.

The important part is that both pre-dispatch reroute and the runtime capability floor now use the same active input view:

- `Runtime_agent.content_blocks_for_run ~initial_messages ~goal_blocks`
- `keeper_turn_driver.run_named` feeds that combined block list into `decide_modality_reroute_for_runtime`.
- `Runtime_agent.run_blocks` validates the same combined view before calling OAS.

That closes the failure mode where an image retained in `initial_messages` could make a later text-only follow-up still fail at provider/OAS time, while the reroute layer saw only `goal_blocks` and stayed on the text runtime. The tests cover both sides: reroute selection from history media and the hard capability gate rejecting unsupported history media.

Current blocker is CI, not the reviewed code path: GitHub shows `Meta Guards` failed on this head. Logs were not available yet because the workflow run was still in progress when checked, so I cannot point to the exact failing line. This PR should stay draft/blocked until that required check is green.

jeong-sik:
## Adversarial review — HEAD `467ffe0aae`: fix keeper history media reroute (+137/-11)

**Strong affirm.** This closes a real provider-400 leak and the fix is the right one — define "active content for an OAS run" once (history + goal) and make both the reroute decision and the capability floor consume that same view.

### [Affirm] The bug: media in history poisoned the next text-only turn
Before, `keeper_turn_driver` decided reroute off `goal_blocks` alone:
```ocaml
match goal_blocks with
| None | Some [] -> runtime_id, assigned_runtime   (* "text turn" → no reroute *)
| Some blocks -> … decide_modality_reroute …
```
But an OAS run ships `initial_messages` (prior history) **plus** the goal. So after an image turn succeeds on a vision runtime, the next text follow-up ("tell me more about that image") carries the image in history — and because `goal_blocks` is text-only, the old code skipped reroute and dispatched to the text-only assigned runtime, which then 400'd on the retained image. That's a silent-to-the-keeper leak that surfaces as an opaque provider error. The comment names it exactly: "an image in history can poison the next text-only follow-up and leak as a provider 400."

### [Affirm] `content_blocks_for_run` is the correct SSOT, used by both consumers
```ocaml
let content_blocks_for_run ~initial_messages ~goal_blocks =
  content_blocks_of_messages initial_messages @ goal_blocks
```
Now the reroute in `keeper_turn_driver` and the validation floor in `Runtime_agent.run_blocks` both build `validation_blocks`/`modality_blocks` from `content_blocks_for_run`. So there's one definition of "what media is in this run" and media retained in history can't bypass pre-dispatch gating on either path. The mli docstring states the invariant explicitly ("Keeper reroute and the runtime capability floor use this same view so media retained in history cannot bypass pre-dispatch gating on a later text-only follow-up") — which is exactly the contract that was violated before. Fail-closed behavior is preserved: when no runtime qualifies, the assigned runtime stands and the loud capability gate rejects (the floor), so this doesn't introduce a silent accept.

### [Affirm] Two regression tests lock both halves
- `test_runtime_multimodal_gate_includes_initial_messages`: history image + text goal → validation rejects with "unsupported image input" (the floor side).
- `test_initial_message_media_drives_reroute`: history image → reroutes to the vision runtime (the reroute side).
Both reproduce the exact failure (history media, text goal) and assert the corrected behavior, so a revert is caught.

### [Low] Per-turn cost of flattening history
`content_blocks_of_messages` concats every message's content on every turn, and `required_modalities_of_content_blocks` walks it. For a long keeper history that's O(n) per turn at the reroute decision point. Acceptable for normal session lengths (it's one pass, no allocation beyond the block list), but if sessions ever get very long this is linear in history on every dispatch. Non-blocking; only worth caching/deriving if history length becomes a measured cost.

Net: correct leak fix, single source of truth shared by both consumers, fail-closed preserved, two regression tests. The behavior is what RFC-0265's reroute should have been from the start (it gated on the turn, not the run). I'd merge.

jeong-sik:
Addressed the Meta Guards failure with 22d2fd2221.

Root cause:
- The DET/NDT ratchet flagged `Option.value goal_blocks ~default:[]` in `lib/keeper/keeper_turn_driver.ml` as new deterministic-boundary debt.

Fix:
- Replaced the hidden default with an explicit `match goal_blocks` boundary.
- Kept the intended semantics: missing current goal blocks mean the current keeper goal is text-only, while `initial_messages` still participate in RFC-0265 modality reroute.

Local verification:
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` -> passed
- `ocamlformat --check lib/keeper/keeper_turn_driver.ml lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli test/test_gate_keeper_backend.ml test/test_runtime_modality_reroute.ml` -> passed

Focused `scripts/dune-local.sh build test/test_gate_keeper_backend.exe test/test_runtime_modality_reroute.exe` was attempted but stopped while waiting on `/tmp/me-dune-local.lock`; CI will run the build/test jobs.

jeong-sik:
### CHANGE REQUEST follow-up (head `639b88e991`) - Meta Guards now gives the exact blocker

The reroute/gate logic is still the right direction after the force-push, but CI is red on a concrete deterministic-boundary ratchet:

```text
origin/main...HEAD: lib/keeper/keeper_turn_driver.ml:142:
  (Option.value goal_blocks ~default:[])
Wrap non-determinism at the boundary, keep parsing sound-partial, or add a nearby DET-OK/NDT-OK comment with rationale.
```

So this is not an unknown CI failure anymore. Please replace the `Option.value` default collapse with an explicit boundary decision, e.g. bind `current_goal_blocks` via `match goal_blocks with None -> [] | Some blocks -> blocks` with the existing rationale that `None` contributes no current-turn content while `initial_messages` still participate in the active run view. A nearby DET-OK comment would also satisfy the guard if you prefer to keep the shape, but I would make the boundary explicit.

No additional code-path blocker found beyond this gate failure.

jeong-sik:
## Close-out - HEAD `22d2fd2221`: deterministic-boundary blocker addressed

Rechecked the force-pushed head after the `Option.value goal_blocks ~default:[]` finding.

Confirmed:

- `keeper_turn_driver` now binds `current_goal_blocks` with an explicit `match goal_blocks`.
- The comment preserves the intended semantics: no current block payload means the current goal contributes no content blocks, while `initial_messages` still participate in reroute.
- `Runtime_agent.decide_modality_reroute_for_runtime ~initial_messages current_goal_blocks` still uses the active run view.

Local verification:

- `scripts/ci/check-determinism-contract.sh --base origin/main --head refs/remotes/pr/21863` - PASS
- `git diff --check origin/main...refs/remotes/pr/21863` - PASS

No remaining code blocker from this pass. CI still needs to finish on GitHub.

jeong-sik:
Current follow-up status for head `22d2fd2221`:\n\n- Local focused validation now passed after the Meta Guards fix:\n  - `ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli lib/keeper/keeper_turn_driver.ml test/test_runtime_modality_reroute.ml test/test_gate_keeper_backend.ml`\n  - `git diff --check origin/main...HEAD`\n  - `scripts/dune-local.sh build test/test_runtime_modality_reroute.exe test/test_gate_keeper_backend.exe`\n  - `./_build/default/test/test_runtime_modality_reroute.exe` (8 tests)\n  - `./_build/default/test/test_gate_keeper_backend.exe` (53 tests)\n- GitHub checks are green except two long-running CI jobs still pending: `Build and Test` and `TLA+ Model Checking`.\n- `pr-ready-check`: no conflicts, no review threads, target branch `main`; still draft and no formal review.\n\nLeaving the PR draft until full CI finishes green and review is requested intentionally.

PR #21850, opened 2026-06-20
fix(keeper): volatile re-observe inherits anchor, not re-verification (RFC-0259 P6/F)

jeong-sik:
## What

Implements **RFC-0259 P6/F** — producer re-observation is no longer treated as re-verification for a volatile (external-ref) claim. `reobserve_fact` now **inherits** the prior row's anchor instead of resetting it.

## Why (defect C1 residual loop)

A `verifier` keeper was observed stuck re-debunking the same false memory every turn. Root-cause verification (8-agent adversarial workflow against HEAD) found: the RFC-0259 P1–P4 marking/TTL layer landed, but `reobserve_fact` (`keeper_memory_os_policy.ml`) advanced `last_verified_at` and refreshed the 24h volatile TTL on **every** producer re-extraction, with the comment asserting "re-observing IS re-verification" (§3.2(b)).

Effect: the librarian re-extracting a stale `"PR #X open"` from its own session history reset both the 12h `[UNVERIFIED]` grounding horizon and the 24h hard TTL each cycle. The false fact never aged out and never rendered `[UNVERIFIED]`, so it was re-injected into the prompt indefinitely.

## The fix

```ocaml
let reobserve_fact ~now ~existing ~incoming:(_ : fact) =
  match existing.external_ref with
  | Some _ -> existing                                  (* volatile: inherit anchor *)
  | None   -> { existing with last_verified_at = Some now }  (* non-volatile: advance *)
```

- **Volatile** (`external_ref = Some`): inherit `first_seen` / `last_verified_at` / `valid_until` unchanged. The TTL and `[UNVERIFIED]` horizon keep ticking from the last *real* verification. Only the grounding reconciler (`keeper_memory_os_reconcile.ml`, `Stale_open` verdict — which actually checks the referent) advances the anchor; the producer inherits it.
- **Non-volatile** (`external_ref = None`): unchanged — re-extraction is its only verification signal, so `last_verified_at` still advances. (Behaviorally identical to the prior `None` branch.)

This **reverses** the shipped §3.2(b) rule; it is not additive (see RFC-0259 §3.7, "This amends the shipped P3").

## Scope

[...]

PR #21865, opened 2026-06-20
feat(keeper): producer claim_id idempotency + volatile anchor inheritance (RFC-0259 P6)

jeong-sik:
## What
Implements **RFC-0259 Phase P6** (producer idempotency & anchor stability) — defects **E** (semantic-duplicate re-mint) and **F** (re-mint freshness reset) from the 2026-06-20 Memory OS adversarial audit (Issue #21789).

Supersedes #21850 (which carried the F-fix alone); RFC §3.7 notes E precedes F, and the two are coupled (E's identity is what routes a reworded re-extraction into the anchor-inheriting merge), so they land together here.

## Why
The librarian re-extracts the same self-narrative every cycle with slight rewording. The exact-text upsert key (`normalize_claim`) misses, so the claim is **appended** as a new row (E), and the new row's fresh `first_seen` resets the 24h volatile TTL + 12h grounding horizon, so a stale status claim re-enters with no `UNVERIFIED` marker (F).

## How
**E — producer-emitted claim identity (`claim_id`):**
- `fact` gains `claim_id : string option` (append-last in JSON, omitted when `None`; legacy rows decode to `None` and round-trip byte-identical, so the snapshot fingerprint is unchanged).
- `claim_identity` keys on `claim_id` when present (`"id:" ^ id`), else the exact-text `normalize_claim` fallback (`"claim:" ^ …`). The librarian emits a deterministic kebab slug for the **conclusion** (not the wording): a reworded re-statement reuses the id and UPSERTs; a changed conclusion (status flip) gets a new id and stays a distinct row. The slug is the librarian's judgment surfaced as a typed key — **not** a code-side text/embedding classifier (RFC-0247 §3 / RFC-0259 §3.7 reject those).
- **Rejected first design (`external_ref` + `category`)**: adversarial review showed `category` cannot separate "PR #N is open" from "PR #N was merged", so it over-merged a status transition and dropped the correction (strictly worse than P3). RFC §3.7 amended (2026-06-21); a regression guard asserts *distinct `claim_id` + same `external_ref` → two rows*.
- `claim_identity` is the single dedup SSOT (write upsert / recall dedup / GC dedup / Tier-2 consolidation / user model). Promoted shared facts carry the corroborating group's `claim_id` so recall's private-precedence dedup matches across tiers (no double injection).

**F — anchor inheritance:**
- `reobserve_fact`: a volatile (external-ref) re-observe inherits the prior row's `first_seen` / `valid_until` / `last_verified_at`. Producer re-extraction is no longer re-verification; only the reconciler (`Stale_open`, an external re-check) advances a volatile claim's anchor. Durable claims still advance `last_verified_at`.

## Tests
86 memory-os cases pass. New/updated: `claim_id` codec round-trip; `claim_identity` (same id collapses, distinct id stays distinct); reworded-same-id → one row with inherited `first_seen`; **distinct-conclusion guard** (two rows); **cross-tier promoted-`claim_id` guard**; volatile anchor inheritance; durable still advances.

## Scope / follow-ups
- `keeper_memory_os_*` is not an RFC-trigger subsystem; this change cites and amends RFC-0259.
- **TLA follow-up**: the RFC P6 gate lists a `ReMintResetsAnchor` bug-model invariant. Behavior is proven here by the property/unit tests + two adversarial-review regression guards; the TLA bug-model is a tracked stacked follow-up.
- Tier-2 consolidation via the LLM `apply_plan` path keeps `claim_id = None` (it synthesizes a new claim with no single producer conclusion); cross-keeper `promote_facts` carries the group id.


jeong-sik:
### 적대 리뷰 (head `982d59db6`, 미리뷰) — P6 E/F 설계는 모범. 단 **`claim_identity`가 producer slug를 raw로 키잉**해 포맷 변형이 idempotency를 무력화 [MED]

설계가 바를 깊이 인지하고 옳은 방향입니다 — 명시적으로 짚고 affirm합니다: producer-emitted typed `claim_id`(conclusion-level slug) + `normalize_claim` conservative fallback + **fuzzy/semantic/embedding dedup 명시 거부**(double-signature bar 인용) + "never derive/hash an id in code" + referent-only key의 over-merge를 적대리뷰로 잡아 amend한 점. F-fix(anchor 상속, reconciler-only advance)도 #21850 검증과 일관.

**[MED · Parse-don't-validate / 노 스트링 매치 경계] `claim_identity`가 `claim_id`를 정규화 없이 raw로 키잉합니다.**
```ocaml
let claim_identity (f : fact) =
  match f.claim_id with
  | Some id when String.trim id <> "" -> "id:" ^ id          (* ← raw, 심지어 untrimmed *)
  | Some _ | None -> "claim:" ^ normalize_claim f.claim
;;
```
`claim_id`가 있으면 `"id:" ^ id`로 키잉하는데, `id`에 **정규화가 전혀 적용되지 않습니다**(`String.trim id <> ""`는 *비어있음 가드*일 뿐, 키는 untrimmed `id`). 그 결과 producer(비결정 LLM)가 같은 conclusion을 **포맷만 다르게** 내면 서로 다른 identity가 됩니다:
- 대소문자: `"PR-123-Open"` vs `"pr-123-open"` → `"id:PR-123-Open"` ≠ `"id:pr-123-open"`.
- leading/trailing 공백: `" pr-123-open"` → `"id: pr-123-open"`(키가 untrimmed이라 trim조차 안 됨).
- 내부 공백/구분자: `"pr 123 open"` vs `"pr-123-open"`.

이 셋은 모두 **의미적 불일치가 아니라 단순 포맷 변형**인데, merge를 놓쳐 새 row를 append합니다 — 이 PR이 막으려던 바로 그 E-duplicate(15–23 micro-variant)의 재현 경로입니다. PR 본문의 "conservative degrade"는 "inconsistent id → miss"를 *의미적* 불일치로 가정하지만, **포맷 변형이 훨씬 흔하고 값싸게 정규화 가능**합니다. 게다가 prompt는 "lowercase kebab-case"를 *요청*할 뿐, LLM이 매번 byte-identical 슬러그를 낸다는 보장이 없습니다.

**모순**: fallback 경로는 `normalize_claim`(lowercase + whitespace collapse + trim)으로 정규화하는데, **primary 식별 메커니즘인 `claim_id`는 정규화하지 않습니다** — 더 견고해야 할 primary 키가 fallback보다 덜 정규화됩니다.

**fix**: `claim_id`를 키로 쓰기 전 `normalize_claim`과 동등하게 정규화(`String.trim` + `String.lowercase_ascii` + 내부 공백 collapse), 또는 전용 `normalize_claim_id`(kebab-case 정규화: lowercase, 공백/언더스코어→`-`, 중복 `-` collapse). 그러면 `"PR-123-Open"` / `" pr-123-open "` / `"pr 123 open"`이 한 identity로 모여 포맷 변형에 강건해지고, P6-E의 idempotency 이득을 실제로 거둡니다. (over-merge 위험은 늘지 않습니다 — 정규화는 *같은 conclusion의 포맷 변형*만 합치지, 다른 conclusion을 합치지 않습니다.)

verdict: 설계·anti-workaround 추론은 production-급입니다. 다만 producer-emitted 키를 boundary에서 정규화하지 않아, idempotency가 LLM의 포맷 일관성에 불필요하게 의존합니다. claim_id 정규화 한 단계로 그 의존을 제거하면 E-fix가 의도대로 동작합니다.

(근거: `keeper_memory_os_types.ml:445` claim_identity `"id:" ^ id`(untrimmed raw) vs fallback `normalize_claim`; `keeper_librarian.ml` `optional_string_field "claim_id"` raw 수신; prompt는 kebab-case 요청만. head 982d59db6. 차등 적대 리뷰 2026-06-21.)

jeong-sik:
### CHANGE REQUEST (head `982d59db6f`) — `claim_id` identity is not preserved through intra-keeper consolidation, and raw ids are not canonicalized

The broad direction is right: `claim_identity` as the write/recall/GC/Tier-2 SSOT avoids the rejected referent-only over-merge. The current implementation still has two identity gaps.

#### 1. Intra-keeper consolidation drops `claim_id`, breaking future upserts

`Keeper_memory_os_consolidator` (cross-keeper Tier-2) now carries the representative `claim_id`, but `Keeper_memory_os_consolidation` (the per-keeper LLM consolidation pass) still creates consolidated facts with `claim_id = None`:

- `lib/keeper/keeper_memory_os_consolidation.ml:246` `consolidated_fact`
- `lib/keeper/keeper_memory_os_consolidation.ml:274` `claim_id = None`

That breaks the new identity contract. Example:

1. A keeper has two reworded rows with `claim_id = Some "pr-123-open"`.
2. The per-keeper consolidation pass merges them into one cleaner claim.
3. The consolidated row loses the id and now keys as `claim:<normalize_claim consolidated_text>`.
4. The next librarian extraction emits `claim_id = "pr-123-open"`.
5. `merge_and_cap_facts` compares `id:pr-123-open` against `claim:<...>` and appends a new row instead of upserting.

So the fix can be undone by the consolidation pass and the same E/F remint path comes back after one maintenance cycle.

Required fix: when a consolidation group has a single shared nonblank `claim_id` across its members, preserve that id on the consolidated fact. If member ids conflict, fall back to `None` or require the LLM consolidation plan to supply an explicit new `claim_id`, but do not unconditionally discard a shared id. Add a regression that applies an intra-keeper consolidation plan to two same-id facts, then verifies a later same-id incoming fact upserts the consolidated row instead of appending.

#### 2. `claim_identity` uses the raw model string as the key

Current keying:

```ocaml
let claim_identity (f : fact) =
  match f.claim_id with
  | Some id when String.trim id <> "" -> "id:" ^ id
  | Some _ | None -> "claim:" ^ normalize_claim f.claim
```

The guard trims only for emptiness; it does not use the trimmed value, lowercase it, or reject non-slug forms. That means these model outputs become different identities:

- `"pr-123-open"`
- `" pr-123-open "`
- `"PR-123-open"`

The prompt asks for lowercase kebab-case, but the parser/key function should still canonicalize the typed id boundary. Otherwise a harmless formatting variation defeats the entire idempotency feature and falls back to duplicate append behavior.

Required fix: canonicalize `claim_id` before keying/storing, at minimum `String.trim` plus lowercase ASCII; preferably validate the kebab-case shape and fall back to `None` for invalid values. Add tests for whitespace/case variants of the same id sharing one key, and blank/invalid ids degrading conservatively to the `claim:` key.

jeong-sik:
## Adversarial review — HEAD `982d59db6f`: producer claim_id idempotency + volatile anchor inheritance (RFC-0259 P6) (+429/-56)

**Strong affirm.** This is the P6 E+F implementation done the way the §3.7 amendment specifies, and it resolves the soundness problem that sank the referent-only identity approach. Reviewed against the §3.7 amendment text and the memory-OS invariant; the implementation matches the design and the conservative-degrade contract holds.

### [Affirm] E — `claim_id` is a producer-emitted *conclusion* slug, not a derived hash
The on-disk `fact` gains `claim_id : string option`, emitted by the librarian as a short kebab-case slug for the **conclusion** (`pr-123-open`, `pr-21249-verification-complete`), and `claim_identity` keys on it when present, falling back to `normalize_claim` otherwise. Crucially, the code never derives or hashes an id (line 59: "we never derive/hash an id in code") — identity is a typed producer judgment, not a substring/embedding classifier we author. That's exactly the line the §3.7 amendment draws: letting the librarian decide "is this the same conclusion?" and surfacing that as a key, rather than building a deterministic status/embedding classifier (which would be the string-classifier workaround). The prompt (librarian episode_extraction) instructs one slug per conclusion with examples.

### [Affirm] The referent-only key was correctly rejected, and the amendment explains why
§3.7 line 42 spells out why the first implementation (`external_ref` + `category`) was unsound: `category` doesn't separate two *different conclusions* about one referent — "PR #N is open" and "PR #N was merged" are both `Fact`/`Pr`/N, so a referent-only key collapses a status transition into one row and `reobserve_fact` keeps the older conclusion, **silently dropping the librarian's own correction**. Nothing restores it (the reconciler doesn't rewrite claim text; `supersedes` is unimplemented). That's strictly worse than P3, which kept both rows. The `claim_id`-on-conclusion fix is the right answer — "open" and "merged" get different ids by construction. This is the load-bearing decision and it's correct.

### [Affirm] The degrade path is conservative and can't over-merge by construction
- No `claim_id` (legacy row, or model omits) → `normalize_claim` exact-text key = pre-P6 append behavior. No behavior change for old data.
- Model emits *different* ids for the same conclusion → misses the merge, appends (the E duplicate — the status quo, not a regression).
- Model emits the *same* id for *distinct* conclusions → over-merge. This is the only failure mode, and it's a same-keeper judgment error that the reconciler still grounds against the external referent — categorically less harmful than the referent-only key, which over-merged *by construction*. The asymmetry (under-merge degrades to status quo; over-merge requires model error + is still grounded) is the right risk profile.

### [Affirm] F — volatile anchor inheritance, reconciler-only advance
`reobserve_fact` for a volatile (`external_ref = Some`) claim inherits the prior row's anchor unchanged (the §3.7 P6/F rule from the merged #21850 logic); only the grounding reconciler advances the anchor on real verification. With #21850 closed and this PR carrying F, the two halves (E identity + F anchor) land in one coherent change rather than a stacked pair — that's cleaner than the E-on-F stack and avoids the dual-PR drift risk.

### [Affirm] `claim_identity` is the shared SSOT across write and recall
The recall path (`keeper_memory_os_recall.ml`) and the write-time consolidation/upsert both key on `claim_identity`, and `normalized_claim_key` routes through it. So write-time upsert and recall-time dedup agree on what "same claim" means — no more "append at write, dedup differently at recall" skew. The `normalize_claim` SSOT is preserved as the fallback.

### [Low / verify] No format validation on `claim_id`
`claim_id` is read as a plain optional string (`optional_string_field`); there's no enforcement that it's actually lowercase-kebab. If the model emits `"PR #123 Open"` or `"pr 123 open"`, `claim_identity` uses it verbatim and the same conclusion could get different ids across turns (under-merge, degrades to append — so still safe, but the idempotency win is lost for that row). The prompt asks for kebab-case, but a light `normalize_claim`-style normalization *applied to the id when present* (lowercase, trim, collapse spaces) would make the model's format sloppiness non-fatal without re-introducing a content classifier. Non-blocking — the conservative fallback means a bad id can't corrupt anything, it just fails to merge.

### [Low / verify] Confirm the +288 test lines cover the status-transition anti-case
The test delta is large (+288 in test_keeper_memory_os). The claim that matters most to lock is the one the amendment is built around: two conclusions about the same referent with *different* `claim_id`s (open → merged) stay as two rows and the newer isn't dropped, while a reworded re-statement with the *same* `claim_id` UPSERTs. If that pair isn't in the tests, the soundness argument is only in the RFC text, not the build. Please confirm those two cases are asserted; they're the whole point of rejecting the referent-only key.

Net: the design (producer-emitted conclusion slug, conservative degrade, referent-only rejected, anchor inheritance, shared `claim_identity` SSOT) is sound and the implementation matches it. I'd merge on the E+F logic; the two asks are a soft id-normalization and confirming the status-transition test pair. This closes the C1/E/F defect axis from the 2026-06-20 audit.

jeong-sik:
Addressed the producer claim-id review in `00f89d7138`.\n\nChanges:\n- Added a typed `normalize_claim_id` boundary and route `claim_identity`, JSON encode, and JSON decode through it. Case/whitespace/underscore/punctuation variants now share the same lowercase kebab id; blank or punctuation-only ids degrade to `None` and fall back to `claim:<normalize_claim>`.\n- Tier-2 consolidation now preserves a shared canonical `claim_id` only when all nonblank member ids agree; conflicting ids are dropped instead of fabricating a new id.\n- Added regressions for canonical id keying/storage, invalid-id fallback/omission, shared-id consolidation, and conflicting-id consolidation.\n\nValidation:\n- `env DUNE_BUILD_DIR=_build-review-21865 DUNE_CACHE=disabled dune build --root . test/test_keeper_memory_os.exe test/test_keeper_memory_os_consolidation.exe`\n- `./_build-review-21865/default/test/test_keeper_memory_os.exe` — 86 tests passed\n- `./_build-review-21865/default/test/test_keeper_memory_os_consolidation.exe` — 17 tests passed\n- `ocamlformat --check lib/keeper/keeper_memory_os_types.ml lib/keeper/keeper_memory_os_types.mli lib/keeper/keeper_memory_os_consolidation.ml test/test_keeper_memory_os.ml test/test_keeper_memory_os_consolidation.ml`\n- `python3 scripts/check_test_coverage.py`\n- `git diff --check`\n- `git diff --check origin/main...HEAD`\n\nNote: `scripts/dune-local.sh build ...` is currently blocked locally by pre-existing `agent_sdk` pin drift; the independent Dune build above was used for this PR-scope compile/test check.

jeong-sik:
## Close-out — HEAD `00f89d7138`: canonicalize producer claim ids (resolves the format-validation Low from `982d59db6f`)

This is the fix I asked for as a Low on the first P6 review. `claim_id` is now canonicalized at every typed boundary instead of used verbatim, so a model emitting `"PR #123 Open"` vs `"pr-123-open"` vs `"pr_123_open"` can no longer produce different identity keys. Resolved.

### [Resolved] `normalize_claim_id` makes format sloppiness non-fatal
```ocaml
let normalize_claim_id raw = …
  (* 'A'..'Z' -> lowercase; alnum preserved;
     '-', '_', space, tab, CR, LF, and any other char -> a single collapsed '-' *)
```
It lowercases, collapses every separator (and any non-alphanumeric) to one `-`, and returns `string option` so a blank/whitespace-only id becomes `None` (falling back to `normalize_claim` text keying) rather than an empty key. So `"PR #123 Open"`, `"pr 123 open"`, `"pr-123-open"`, `"PR_123_Open"` all canonicalize to `pr-123-open`. That's strictly stronger than the "lowercase + trim + collapse spaces" I suggested — it covers punctuation and underscore separators too.

### [Resolved] Canonicalization is applied at every boundary, not just one
The normalize isn't bolted onto a single call site — it's applied wherever a `claim_id` becomes a key or crosses a boundary:
- `claim_identity`: `Option.bind f.claim_id normalize_claim_id` — so the identity key itself is canonical (two facts with surface-different ids that canonicalize equal now UPSERT-merge instead of appending).
- `shared_claim_id_for_members`: consolidation only promotes a `claim_id` onto the shared fact when **all** members canonicalize to the same id (`[ id ] -> Some id`; `[]` or 2+ distinct -> `None`). That's a real over-merge guard — if members carry different ids the shared fact declines to pick one.
- JSON **write**: the serialized `claim_id` is the normalized form.
- JSON **read**: legacy rows with non-canonical `claim_id` are normalized on load.

Applying it at read-time as well as write-time is the important part — it means rows written before this change (or by a model that emitted a non-canonical id) get folded into the same identity as canonical rows, so the canonicalization is retroactive, not just forward-looking. No verbatim `claim_id` reaches identity/dedup/consolidation anymore.

### [Affirm] The shared-fact claim_id is now non-None when it should be
`consolidated_fact … { claim_id = shared_claim_id_for_members members }` replaces the earlier `claim_id = None` on the shared fact — so a corroborated group whose members agree on a conclusion id now carries that id forward (the "carry the group's claim_id onto the promoted shared fact" behavior), and declines to carry one when members disagree. Consistent with the conservative-degrade contract from the §3.7 amendment.

Net: the format-validation Low is fully resolved — canonicalization at four boundaries (identity key, consolidation, JSON write, JSON read), blank-id fallback, and a shared-id agreement guard. The P6 `claim_id` machinery is now drift-proof against model formatting. No remaining concern from me on this thread.

jeong-sik:
### 고쳐졌다 확인 (head `00f89d713`) — 내 [MED](#issuecomment-4759947180)(raw claim_id 키잉이 포맷 변형에 idempotency 무력화)가 `normalize_claim_id`로 닫힘

`canonicalize producer claim ids` 커밋이 권고한 전용 정규화를 정확히 추가했습니다:
```ocaml
let claim_identity (f : fact) =
  match Option.bind f.claim_id normalize_claim_id with
  | Some id -> "id:" ^ id
  | None -> "claim:" ^ normalize_claim f.claim
```
`normalize_claim_id`가 견고한 kebab canonicalizer입니다 — 대문자→소문자, `-`/`_`/공백/탭/개행/기타 문자를 단일 `-`로 collapse(`pending_sep`로 연속·trailing 구분자 제거), 빈 결과는 `None`으로 `normalize_claim` fallback. 따라서 제가 지적한 포맷 변형이 모두 한 identity로 수렴합니다:
- `"PR-123-Open"` / `"pr_123_open"` / `"pr 123 open"` / `" PR  123  Open "` → 전부 `"id:pr-123-open"`.

이로써 producer slug가 *의미적으로 같은데 포맷만 다를 때* merge를 놓쳐 E-duplicate를 재현하던 경로가 닫혔고, primary 키가 fallback보다 덜 정규화되던 모순도 해소됐습니다(이제 양쪽 다 정규화). `Option.bind`의 `None`-fallback이 malformed/empty slug 경계도 보수적으로 처리합니다. verdict: 닫혔습니다 — P6-E의 idempotency가 LLM 포맷 일관성에 의존하지 않게 됐습니다.

(근거: `keeper_memory_os_types.ml` `normalize_claim_id`(buffer kebab collapse, empty→None) + `claim_identity`가 `Option.bind ... normalize_claim_id`. head 00f89d713. 차등 적대 리뷰 2026-06-21.)

jeong-sik:
[적대 리뷰 follow-up · head `00f89d713` · RFC-0259 P6 설계 관점 · 정규화가 over-merge 방향에서도 안전함을 확인]

`canonicalize producer claim ids`가 raw-slug 키잉 MED를 닫은 건 위에서 확인됐습니다. P6 설계 의도상 정규화는 **양방향**으로 옳아야 하는데(이 amendment가 referent+category를 버린 이유가 over-merge였으므로), 그 직교 방향을 `normalize_claim_id` 구현으로 검증했습니다.

**[Affirm · 정규화가 distinguishing 토큰을 보존 — over-merge 재유발 없음]**
`normalize_claim_id`는 순수 slug canonicalizer입니다(`A-Z`→lowercase, `a-z|0-9` 보존, `-_ \t\r\n`+기타→single `-` 축약, 양끝 trim, 빈 결과→`None`). 즉:
- `"PR-123-Open"` → `pr-123-open`, `"pr-123-merged"` → `pr-123-merged` — **conclusion 토큰(open vs merged)이 보존**되어 서로 다른 키로 남습니다.
- 케이스/구분자/구두점 변형만 같은 키로 모읍니다(idempotency, 위 MED).

이게 핵심인 이유: 정규화가 토큰을 드롭하거나 truncate하는 식으로 조금만 더 aggressive했다면, claim_id 설계가 막으려던 **status-transition over-merge**(open→merged collapse, correction silent drop)를 정규화가 뒷문으로 재도입했을 것입니다. 현 구현은 그 함정을 피합니다.

**[Affirm · namespacing이 id/text cross-collision 차단]**
`claim_identity`가 `"id:" ^ id` vs `"claim:" ^ normalize_claim`로 네임스페이스를 분리해, claim_id `"foo"`와 text가 `normalize_claim`으로 `"foo"`가 되는 별개 fact가 우연히 한 키로 merge되지 않습니다. legacy/omitted degrade(`None → "claim:"` 경로)도 pre-P6 append 동작 그대로라 over-merge 안전.

정리: 정규화 fix가 idempotency(format 변형 수렴)와 conclusion 구분(over-merge 방지)을 **동시에** 만족합니다. P6 설계 불변식 유지 — 좋습니다.

(근거: `keeper_memory_os_types.ml:432` `normalize_claim_id`(A-Z lowercase·alnum 보존·sep 축약·trim), `:472` `claim_identity` `Option.bind … normalize_claim_id`/`"id:"`·`"claim:"` namespacing + §3.7 주석 "changed conclusion … different id … distinct row". 차등 적대 fresh-context 2026-06-21, head 00f89d713.)

jeong-sik:
Addressed the two adversarial identity gaps on the PR head.

Fix commit: `00f89d7138 fix(keeper): canonicalize producer claim ids`

What changed:
- Added `normalize_claim_id` at the Memory OS type boundary. `claim_identity`, JSON decode, and JSON encode now use the same canonical lowercase kebab slug, so harmless LLM formatting variants like whitespace/case/underscore/punctuation do not defeat idempotency. Blank/empty ids still degrade to the `claim:` text key.
- Per-keeper consolidation now preserves a single non-conflicting canonical `claim_id` across grouped members. Conflicting ids still conservatively become `None`.
- Added tests for canonical id storage/keying and shared/conflicting consolidation ids.

Verification:
- `ocamlformat --check` on touched OCaml/test files
- `git diff --check`
- GitHub CI for `00f89d7138` is green, including Build and Test, TLA+ Model Checking, ocamlformat-check, Lint, and CI Gate.

This closes the raw-id and intra-keeper consolidation gaps raised earlier.

PR #21866, opened 2026-06-20
fix(keeper): retry thinking-only read-only accept rejection

jeong-sik:
## Summary

Fix keeper runtime fallback for the narrow failure shape where a provider returns a thinking-only response after a read-only tool call.

- Keep the existing accept contract: thinking-only responses are still rejected as no usable progress.
- Keep generic `AcceptRejected` non-retryable.
- Carry `response_shape` and `last_tool_effect` through the typed MASC internal error envelope instead of parsing retry decisions from diagnostic text.
- Retry only when the typed rejection is `Accept_no_usable_progress`, `response_shape=thinking_only`, and `last_tool_effect=read_only`.
- When the failing runtime is also the default runtime, append configured `tools-support=true` runtimes to the degraded rotation candidate set.
- Add an adversarial audit list for other keeper/runtime heuristics that lack concrete real-world examples or fixture boundaries.

## Product impact

Keeper turns that hit a read-only `WebFetch`/read path and then receive a provider thinking-only response no longer fail the whole cycle while another safe tool-capable runtime remains available.

Mutating-tool, generic predicate rejection, and legacy string-only accept rejections remain terminal. This does not make thinking-only output successful; it only lets MASC retry a different runtime when no mutation can be duplicated.

## Evidence

- `env MASC_SKIP_PIN_CHECK=1 scripts/dune-local.sh build test/test_keeper_sdk_error_typed_bridge.exe test/test_keeper_turn_driver_accept.exe`
- `./_build/default/test/test_keeper_sdk_error_typed_bridge.exe` — 13 tests passed.
- `./_build/default/test/test_keeper_turn_driver_accept.exe` — 18 tests passed.
- `git diff --check`
- `opam exec -- ocamlformat --check lib/keeper/keeper_error_classify.ml lib/keeper/keeper_error_classify.mli lib/keeper/keeper_turn_driver.mli lib/keeper/keeper_turn_driver_try_provider.ml lib/keeper/keeper_turn_driver_try_provider.mli lib/keeper/keeper_turn_driver_try_runtime.ml lib/keeper/keeper_turn_driver_wrappers.ml lib/keeper_runtime/keeper_internal_error.ml lib/keeper_runtime/keeper_internal_error.mli lib/keeper_tooling/keeper_tool_response.ml lib/keeper_tooling/keeper_tool_response.mli test/test_keeper_sdk_error_typed_bridge.ml test/test_keeper_turn_driver_accept.ml`

## Direct evidence

schema_version: 1
direct_ratio: 5/5
provenance: direct

- Added typed `response_shape` and `last_tool_effect` fields to `Accept_rejected` envelopes and preserved JSON parsing compatibility for missing fields.
- Changed provider/runtime retry gates to use typed accept rejection context, not `reason` string tokens.
- Added default-runtime recovery coverage that skips `tools-support=false` profiles and selects the next tool-capable runtime.
- Added regression coverage that typed hints still retry if diagnostic wording changes, while legacy string-only tokens do not.
- Re-ran the focused build and both affected test executables successfully.

## Review evidence

Manual adversarial pass checked that this does not make all `AcceptRejected` failures retryable and does not accept thinking-only output as deliverable content. The retry path is limited to typed no-progress responses after read-only tools.

The OAS/MASC boundary stays intact: OAS owns response shape classification and provider transport; MASC owns keeper accept policy, side-effect safety, and runtime fallback.

## Linked issue

Operator-reported keeper runtime failure:

`accept_rejected ... shape=thinking_only ... last_tool=WebFetch; last_tool_effect=read_only`

### Commit lineage (auto-synced)

`fix/thinking-only-readonly-retry` → `main` · 4 commit(s) · synced 2026-06-21T01:38:45Z

| sha | subject | date |
|---|---|---|
| `270d0ae40` | fix(keeper): retry thinking-only read-only accept rejection | 2026-06-20 |
| `ff7c2a438` | docs: audit keeper heuristic real-world gaps | 2026-06-20 |
| `59a9aec81` | fix(keeper): route read-only no-progress retries through live path | 2026-06-20 |
| `84bd2c47b` | fix(keeper): use typed read-only progress retry hints | 2026-06-20 |

<sub>Managed by `scripts/pr-sync-body.sh` — do not hand-edit between these markers. The code of record is the diff, not prose above this block.</sub>

### DIFF
--- a/docs/audit/2026-06-21-keeper-heuristic-realworld-gap-list.md
+++ b/docs/audit/2026-06-21-keeper-heuristic-realworld-gap-list.md
@@ -1,0 +1,322 @@
+---
+status: reference
+last_verified: 2026-06-21
+code_refs:
+  - lib/keeper/keeper_turn_driver_try_runtime.ml
+  - lib/keeper/keeper_turn_driver_try_provider.ml
+  - lib/keeper/keeper_tool_registry.ml
+  - lib/keeper/keeper_turn_driver_provider_attempt.ml
+  - lib/keeper/keeper_error_classify.ml
+  - lib/runtime/runtime_candidate.ml
+  - lib/keeper/keeper_world_observation_board_signal.ml
+  - lib/keeper/keeper_unified_metrics_support.ml
+  - lib/keeper/keeper_alerting.ml
+  - lib/keeper/keeper_runtime_trust_timeline.ml
+---
+
+# Keeper Heuristic Real-World Gap List
+
+> Date: 2026-06-21
+> Scope: MASC keeper/runtime heuristics that affect retry, wakeup,
+> observability, or operator escalation.
+> Method: adversarial read-through of keeper/runtime code paths after the
+> thinking-only read-only accept-rejection fix.
+> Status: findings documented; not all entries are bugs. Each entry names the
+> missing real-world fixture or operational example that would make the
+> heuristic defensible.
+
+## Inclusion criteria
+
+This list includes code when all of the following are true:
+
+- The implementation changes runtime behavior or operator-visible
+  classification.
+- The decision is driven by a string match, score, threshold, last-event
+  summary, or no-op/default stub.
+- The code lacks a concrete real-world example or fixture that shows the
+  intended behavior and the nearest false-positive/false-negative boundary.
+
+Out of scope: OAS provider/model transport internals, typed OAS response shape
+contracts, and pure documentation text. MASC should consume those surfaces
+through explicit boundary data rather than grow provider-specific behavior in
+OAS.
+
+## Summary
+
+| # | Surface | Current heuristic | Missing example class | Risk |
+|---|---------|-------------------|-----------------------|------|
+| 1 | Thinking-only read-only retry | accept reason substring gate | actual no-progress trace after read-only tool | retry can hide accept failures |
+| 2 | Last-tool effect context | only the final assistant tool call is inspected | multi-tool turn with earlier mutation | read-only label can be unsound |
+| 3 | Tool read-only registry | descriptor/capability fallback | mixed-mode tool contracts | checkpoint boundary can drift |
+| 4 | Provider error string classifiers | quota/capacity/terminal substrings | provider/CLI error corpus | wrong cooldown/escalation |
+| 5 | Keeper warn/recoverable classifier | typed bucket allow/deny list | operator-facing failure samples | noisy or silent incidents |
+| 6 | Runtime candidate neutral stubs | default 0/false/1.0/None | local runtime health examples | dashboards imply certainty |
+| 7 | Board wake/stigmergy match | mention substring and keyword score | Korean/punctuation/false-positive posts | irrelevant keeper wakeups |
+| 8 | Metrics/turn-mode classifiers | output prefixes and phrase search | multilingual turn transcripts | misleading autonomous status |
+| 9 | Alert scoring | uncalibrated keyword weights | incident/non-incident corpus | missed or inflated alerts |
+| 10 | Trust timeline severity | event-type substring severity | benign event names with trigger words | inaccurate timeline severity |
+
+## 1. Thinking-only read-only retry
+
+Evidence:
+- `lib/keeper/keeper_turn_driver_try_runtime.ml:69-78`
+- `lib/keeper/keeper_turn_driver_try_runtime.ml:251-257`
+
+Current behavior: typed `Accept_rejected` with
+`Accept_no_usable_progress` is retryable on the next runtime only when the
+free-form reason contains both `shape=thinking_only` and
+`last_tool_effect=read_only`.
+
+Missing real-world examples:
+- WebFetch/Read/Grep succeeds, then the runtime returns only thinking text.
+- A read-only tool returns an error or empty payload, then the runtime returns
+  only thinking text.
+- The last tool is read-only but an earlier tool in the same turn mutated
+  workspace or board state.
+- The rejection happens on the final candidate, where retry is intentionally
+  impossible.
+
+Operational risk: the fix is intentionally narrow, but its safety rests on a
+string-decorated reason. Without fixture traces for the false-positive
+boundary, a future reason-format change can silently stop retrying, or a
+misclassified read-only context can retry after a turn that already had side
+effects.
+
+Evidence fixture needed: a raw checkpoint/rejection fixture with multiple tools,
+including at least one read-only-only turn and one earlier-mutation turn.
+
+## 2. Last-tool effect context
+
+Evidence:
+- `lib/keeper/keeper_turn_driver_try_provider.ml:195-221`
+- `lib/keeper/keeper_turn_driver_try_provider.ml:224-234`
+
+Current behavior: the checkpoint scan records only the final assistant
+`ToolUse` and formats it as `last_tool=<name>; last_tool_effect=<read_only|mutating>`.
+
+Missing real-world examples:
+- A turn calls `Edit` then `Read`, and the model produces no deliverable.
+- A turn calls several read-only tools where only the last one is relevant to
+  the rejection.
+- A checkpoint includes tool-use blocks from multiple assistant messages after
+  compaction or resume.
+
+Operational risk: "last tool was read-only" is not the same as "this turn was
+read-only". Retry decisions built on the last tool alone can ignore earlier
+side effects.
+
+Evidence fixture needed: a checkpoint summary that includes `any_mutating_tool`,
+`last_tool_effect`, and `tool_effects_seen`, plus tests that pin the intended
+precedence.
+
+## 3. Tool read-only registry and checkpoint boundary
+
+Evidence:
+- `lib/keeper/keeper_tool_registry.ml:134-154`
+- `lib/tool_orchestration/tool_job.ml:38-47`
+
+Current behavior: read-only classification comes from keeper read-only sets,
+descriptor capabilities, idempotent capability, or input-aware descriptor
+resolution. Tool jobs use catalog metadata and fall back to write semantics for
+unknown tools.
+
+Missing real-world examples:
+- A mixed-mode public tool where the same tool name has both read and write
+  subcommands.
+- A descriptor with `Idempotent` capability but nontrivial external side
+  effects.
+- An unregistered tool that is read-only in practice but defaults to
+  `write:any` in scheduling.
+
+Operational risk: read-only status is consumed by retry, checkpoint boundary,
+governance risk, and scheduler resource keys. A single descriptor/capability
+mistake can fan out to multiple runtime policies.
+
+Evidence fixture needed: descriptor-level examples for mixed-mode tools,
+idempotent-but-mutating tools, and unknown tools, with expected boundary and
+resource-key behavior.
+
+## 4. Provider error string classifiers
+
+Evidence:
+- `lib/keeper/keeper_turn_driver_provider_attempt.ml:178-201`
+- `lib/keeper/keeper_turn_driver_provider_attempt.ml:221-240`
+
+Current behavior: hard quota, capacity backpressure, and terminal runtime
+failures are detected by substring indicators in provider/CLI error messages.
+
+Missing real-world examples:
+- Actual current OpenAI-compatible, Ollama Cloud, Claude CLI, and local runtime
+  error envelopes for 429, capacity, timeout, SSE parse, and JSON-RPC parse
+  cases.
+- A transient overload message that contains quota-like words but should remain
+  retryable.
+- A hard quota message wrapped in a CLI exception whose wording changes.
+
+Operational risk: these classifiers control health recording and fallback
+behavior. Misclassification can either keep hammering a quota-exhausted account
+or prematurely suppress a recoverable provider.
+
+Evidence fixture needed: a provider-error corpus with expected
+`hard_quota`, `capacity_backpressure`, `terminal_runtime_failure`, and
+`retryable` labels.
+
+## 5. Keeper warn/recoverable classification
+
+Evidence:
+- `lib/keeper/keeper_error_classify.ml:682-709`
+
+Current behavior: auto-recoverable errors are a small OR-list, and
+`should_warn_keeper_cycle_failed` warns for provider timeout and capacity
+backpressure while suppressing many typed MASC internal errors, including
+accept rejection and runtime exhausted.
+
+Missing real-world examples:
+- Operator-facing examples showing which keeper-cycle failures should page,
+  warn, or remain routine.
+- A thinking-only accept rejection after a read-only tool versus after a
+  mutating tool.
+- A provider timeout caused by a user-configured too-small wall-clock budget
+  versus a provider outage.
+
+Operational risk: the system can become quiet about failures that are
+actionable, or noisy about failures that the keeper will recover from
+automatically.
+
+Evidence fixture needed: a table of real log lines mapped to
+`auto_recoverable`, `warn_cycle_failed`, and expected operator action.
+
+## 6. Runtime candidate neutral stubs
+
+Evidence:
+- `lib/runtime/runtime_candidate.ml:107-112`
+- `lib/runtime/runtime_candidate.ml:158-172`
+- `lib/runtime/runtime_candidate.ml:189-222`
+
+Current behavior: several candidate helpers intentionally return neutral values
+after single-binding dispatch collapsed old multi-candidate machinery:
+context window hint is `0`, header sync is `false`, unmetered provider is
+`false`, threshold multipliers are `(1.0, 1.0)`, recovery evidence is `false`,
+and unhealthy local runtime filtering is a no-op.
+
+Missing real-world examples:
+- Local runtime with known context window and no auth.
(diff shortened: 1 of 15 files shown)


jeong-sik:
### CHANGE REQUEST (head `270d0ae4080`) - retry policy parses the diagnostic `reason` string

This is a P0 keeper-stall path, so the retry decision cannot depend on human-readable diagnostic text.

`accept_no_progress_read_only_should_try_next` currently makes the retry decision by substring-matching the `Accept_rejected.reason` field:

```ocaml
String_util.contains_substring reason "shape=thinking_only"
&& String_util.contains_substring reason "last_tool_effect=read_only"
```

That violates the typed-boundary bar for this code path. `reason` is produced as a diagnostic string by `Keeper_tool_response.accept_rejection_of_response` and then extended by `format_last_tool_progress_context`. It is useful for logs, but it is not a protocol. A punctuation/name change in `Agent_sdk.Response_shape.diagnostic_summary`, or a future formatting change from `last_tool=...; last_tool_effect=...`, would silently disable the retry path and reintroduce the keeper stall this PR is trying to fix.

The producer side already has typed data before it is flattened:

- `Keeper_tool_response.response_accept_rejection` has the response shape result.
- `Keeper_turn_driver_try_provider.last_tool_progress_context` has `{ tool_name; tool_effect }`.
- `Keeper_internal_error.Accept_rejected.reason_kind` is already typed.

Required direction: carry a typed retry hint / typed accept rejection context across the boundary, or decide this retry eligibility before flattening the context into `reason`. Matching constructors/fields such as `Accept_no_usable_progress`, `Thinking_only`, and `Read_only` is OK; matching substrings inside `reason` is not.

Please also add a regression that changes the diagnostic wording/format while preserving the typed shape/effect and proves retry behavior does not change. The current tests only pin the present string tokens, so they would pass while locking in the fragile contract.

jeong-sik:
## Adversarial review — HEAD `270d0ae408`: retry thinking-only read-only accept rejection (+111/-4)

**Change requested.** The recovery arm this adds is the right idea and the gap it closes is real (this is the "accept-rejected with no recovery arm" from RFC-0271 / the keeper thinking-only audit). But the *predicate* that decides whether to retry is a substring match on a serialized reason string, and that's the string-classifier workaround shape — it should be a typed field match instead. The good news is the structure is already half-typed, so the fix is small.

### [Affirm] The gap is real, the typed outer match is right
The recovery arm — try the next runtime candidate when an accept-rejection is "thinking-only after a read-only tool" — is exactly the missing recovery from the accept-reject audit. And the outer classification is correctly typed: `classify_masc_internal_error` narrows to `Accept_rejected { reason_kind = Some Accept_no_usable_progress; … }`, and the match enumerates every `Keeper_internal_error` variant explicitly (no `_ -> false` catch-all). So the FSM-sparse-match anti-pattern is avoided and the compiler will flag a new error variant. That part is solid.

### [Change requested, MED] The retry decision is a substring match on a serialized reason
The actual decision is:
```ocaml
| Some (Accept_rejected { reason_kind = Some Accept_no_usable_progress; reason; _ }) ->
  String_util.contains_substring reason "shape=thinking_only"
  && String_util.contains_substring reason "last_tool_effect=read_only"
```
`shape` and `last_tool_effect` are closed concepts (`thinking_only` is one of a known set; `read_only` likewise) — exactly the case where a typed variant beats a string. They've been serialized into the free-text `reason` string as `key=value` pairs and are now recovered by substring search. That's the String/Substring-classifier workaround signature (CLAUDE.md "워크어라운드 거부 기준"): the compiler can't catch it if the reason-emitting code changes the format (renames `shape=` → `mode=`, translates it, fixes a typo, adds a space), and the failure mode is **silent no-recovery** — `contains_substring` returns false, `should_try_next` stays false, the keeper doesn't retry and re-stalls. That's the P0-ish outcome (keeper blocked on an accept-reject that *should* have recovered) triggered by an unrelated reason-format edit. RFC-0271's direction was a typed `recovery_action` (`Retry_no_thinking` / `Reroute` / `Terminal`); this PR re-introduces the string match that direction was meant to replace.

### Suggested fix (root, small because the typing is already half-done)
Promote `shape` and `last_tool_effect` from serialized reason text to typed fields on the accept-rejection record, then match on the variant:
```ocaml
type accept_shape = Thinking_only | …
type last_tool_effect = Read_only | …
(* in the Accept_rejected payload: *)
{ reason_kind : …; shape : accept_shape option; last_tool_effect : last_tool_effect option; reason_text : string; … }
```
then the predicate becomes `shape = Some Thinking_only && last_tool_effect = Some Read_only` — exhaustive, compiler-checked, and a format change to the human-readable `reason_text` can no longer silently disable recovery. The reason *text* can still be emitted for logs/dashboards (derive it from the typed fields), but the decision reads the typed fields, not the text. If emitting typed fields at the accept-rejection site is too invasive for this PR, at minimum route both the emit and the match through shared typed helpers (so a rename touches both) rather than two independent string literals.

### [Verify] Is the reason string already the established channel?
If `is_accept_rejected_sdk_error` and friends already parse the reason string, this PR is adding a second substring classifier on top of an existing one — i.e. the string-classifier pattern is accumulating, which is exactly the "main에 들어가면 합리적 선례로 학습" risk. If so, this is also the moment to start the typed migration rather than deepening the string dependency. Please point at where `shape=thinking_only` / `last_tool_effect=read_only` are emitted so the review can confirm whether a typed field is reachable there.

Net: the recovery behavior is wanted and the outer typing is good, but the retry predicate is a string match that should be a typed match — please promote `shape`/`last_tool_effect` to fields before this lands, otherwise an unrelated reason-format edit can silently re-stall the keeper.

jeong-sik:
Adversarial follow-up on the read-only no-progress retry slice.

Found a blocker in the previous head: the PR changed `Keeper_turn_driver_try_runtime.run`, but the current `run_named` production path no longer calls that loop after the single-runtime/RFC-0206 collapse. That meant the old tests proved the helper predicate, not live keeper retry behavior.

Pushed `59a9aec819` to connect the same narrow condition into the live unified-turn degraded-rotation classifier:
- `Accept_no_usable_progress + shape=thinking_only + last_tool_effect=read_only` now maps to `read_only_no_progress` and can rotate to an untried runtime.
- Generic `Accept_rejected`, thinking-only with no read-only tool, and thinking-only after a mutating tool stay terminal.
- `read_only_no_progress` does not cycle candidates after exhaustion, so it is a one-pass cross-runtime recovery, not an infinite retry class.

Local verification on the pushed fix:
- `ocamlformat --check lib/keeper/keeper_error_classify.ml lib/keeper/keeper_error_classify.mli test/test_keeper_turn_driver_accept.ml test/test_keeper_sdk_error_typed_bridge.ml`
- `scripts/dune-local.sh build test/test_keeper_turn_driver_accept.exe test/test_keeper_sdk_error_typed_bridge.exe`
- `./_build/default/test/test_keeper_turn_driver_accept.exe` -> 17 passed
- `./_build/default/test/test_keeper_sdk_error_typed_bridge.exe` -> 12 passed
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` -> PASS, only existing repo-wide WARNs
- `git diff --check`

GitHub checks restarted on `59a9aec819`; they are pending at the time of this comment.

jeong-sik:
## Adversarial review — HEAD `59a9aec819`: route read-only no-progress retries through live path — the string match moved, it didn't get typed

This doesn't resolve the change-requested from my `270d0ae408` review; it relocates the same substring match into a second file and adds a typed wrapper around it. The typed wrapper is a real improvement, but the predicate at the bottom is still the string classifier, and now there are two copies of it. Asking again, more specifically.

### [Affirm] The `degraded_retry_reason` extension is the right shape
Adding `Read_only_no_progress` as a constructor to `degraded_retry_reason`, threading it through `degraded_retry_reason_to_string`, `degraded_rotation_after_recoverable_error`, `recoverable_runtime_failure_reason`, `degraded_reason_allows_candidate_cycle`, and the rotation logic — all exhaustive, all tested (`test_read_only_no_progress_rotates_to_default_runtime`). That's a correct typed retry-taxonomy extension and it's the direction I wanted for the *outcome*. Good.

### [Change requested, still open — and now worse: the substring match was duplicated]
The new `is_thinking_only_read_only_accept_rejection` in `keeper_error_classify.ml` is the **same predicate** I flagged in `keeper_turn_driver_try_runtime.ml` last round:
```ocaml
let is_thinking_only_read_only_accept_rejection err =
  match classify_masc_internal_error err with
  | Some (Accept_rejected { reason_kind = Some Accept_no_usable_progress; reason; _ }) ->
    String_util.contains_substring reason "shape=thinking_only"
    && String_util.contains_substring reason "last_tool_effect=read_only"   (* same two literals *)
  | Some ( …every variant… ) | None -> false
```
So the decision ("is this the thinking-only-after-read-only case?") is still a substring search over the serialized `reason` text. The change this round wrapped that decision in a typed `Read_only_no_progress` reason and routed it through the live degraded-rotation path — but the *gate* that decides whether you enter that path is the unchanged string match, now living in `keeper_error_classify` as well as the original `accept_no_progress_read_only_should_try_next` in `keeper_turn_driver_try_runtime`. That's the half-typed trap: the outcome is a typed variant, the predicate that produces it is a string classifier.

Two concrete consequences:
1. **N-of-M on the string literals.** `"shape=thinking_only"` and `"last_tool_effect=read_only"` now appear in two functions in two files. If the reason-emitting code changes the format (rename, translate, fix a typo), both have to be updated in lockstep or one path classifies and the other doesn't — exactly the drift the typed variant was supposed to eliminate. There's no shared constant or shared predicate; they're independent.
2. **The P0 failure mode is unchanged.** A reason-format edit still silently breaks classification (`contains_substring` → false → no `Read_only_no_progress` → no rotation → keeper stalls on the accept-reject it was supposed to recover from). Wrapping the string match in a typed reason doesn't change that; it just makes the stall arrive via a typed `None` instead of a bare `false`.

### What actually closes it (same ask as before, now with a concrete home)
Promote `shape` and `last_tool_effect` to typed fields on the `Accept_rejected` payload so the predicate is a variant match, and emit the human `reason` text by *deriving* it from those fields:
```ocaml
type accept_shape = Thinking_only | Tool_mutating | …
type last_tool_effect = Read_only | Mutating | …
(* Accept_rejected payload gains: shape : accept_shape option; last_tool_effect : last_tool_effect option *)
```
Then `is_thinking_only_read_only_accept_rejection` becomes `shape = Some Thinking_only && last_tool_effect = Some Read_only` — one definition, compiler-checked, and the two call sites collapse to one shared predicate (no N-of-M). The reason *text* can still carry `shape=thinking_only` for logs, but it's generated from the typed fields, so the text and the decision can't drift. If the reason-emitter genuinely can't carry typed fields yet, the minimum bar is a single shared predicate (one function, both callers route through it) plus a test that the emitter and the matcher agree on the literal — but that's a band-aid for the string match, not the fix.

### Net
The typed retry-reason is welcome, but this PR doesn't close the string-match change-requested — it duplicates it. Please promote `shape`/`last_tool_effect` to typed fields (or at minimum centralize the one predicate) before this lands, otherwise the "half-typed" classifier that can silently re-stall the keeper is now in two places instead of one. Holding the change-requested.

jeong-sik:
## Close-out — HEAD `84bd2c47b9`: use typed read-only progress retry hints (full resolve of the string-match change-requested)

This is the fix I asked for in `270d0ae408` and again in `59a9aec819`. The substring classifier is gone, the decision is a variant match on typed fields, the two call sites collapsed to one shared predicate, and the test proves the decision no longer touches the reason text. Closing the change-requested.

### [Resolved] `shape` / `last_tool_effect` are now typed fields, matched as variants
`keeper_internal_error.ml` defines two closed variants:
```ocaml
type accept_response_shape =
  | Accept_response_empty | Accept_response_thinking_only | Accept_response_blank_text_only
  | Accept_response_tool_result_only | Accept_response_media_only
  | Accept_response_mixed_without_deliverable_content | Accept_response_has_deliverable_content
type tool_progress_effect = Tool_effect_read_only | Tool_effect_mutating
```
and the `Accept_rejected` payload now carries `response_shape : accept_response_shape option` and `last_tool_effect : tool_progress_effect option`. The predicate is a pure variant match:
```ocaml
let accept_rejection_has_read_only_no_progress_retry_hint = function
  | Accept_rejected { reason_kind = Some Accept_no_usable_progress;
                      response_shape = Some Accept_response_thinking_only;
                      last_tool_effect = Some Tool_effect_read_only; _ } -> true
  | Accept_rejected _ | Runtime_exhausted _ | … (every variant) … -> false
```
No `contains_substring`, no string literals in the decision. `response_shape` is sourced from `Agent_sdk.Response_shape.content_shape` via `accept_response_shape_of_agent_sdk`, so it's typed at the producer (the response summary) and stays typed through to the retry decision. Exactly the promotion asked for.

### [Resolved] The two string-classifier copies collapsed to one SSOT predicate
Both `keeper_error_classify.is_thinking_only_read_only_accept_rejection` and `keeper_turn_driver_try_runtime.accept_no_progress_read_only_should_try_next` now just delegate to `accept_rejection_has_read_only_no_progress_retry_hint`. The N-of-M duplication from `59a9aec819` is gone — one definition, two thin wrappers, so the `"shape=thinking_only"` / `"last_tool_effect=read_only"` literals no longer exist in two places that could drift. (They don't exist in the decision at all anymore; they only appear as the serialized wire form via `accept_response_shape_to_string` / `tool_progress_effect_to_string`, which is the right place for them.)

### [Resolved] The P0 failure mode is gone, and the test proves it
`test_read_only_retry_uses_typed_context_not_reason_tokens` is the decisive one:
- an error with typed `response_shape = Thinking_only` / `last_tool_effect = Read_only` and a reason text that says *"diagnostic wording changed; no legacy retry tokens"* → retries (true);
- an error with both typed fields `None` but a reason text that still contains *"shape=thinking_only; last_tool_effect=read_only"* → does **not** retry (false).

So the decision is provably independent of the reason wording — a reason-format edit can no longer silently disable recovery and re-stall the keeper. That was the whole P0 concern, and this test locks it.

### [Affirm, bonus] Rotation now prefers tool-capable runtimes for this case
`default_degraded_rotation_candidates` gained a `~fallback_reason` and, for `Read_only_no_progress`, appends runtimes with `tools_support = true`. That's semantically right — a thinking-only-after-read-only-tool rejection wants another tool-capable model, not just any model — and `test_read_only_no_progress_default_runtime_uses_tool_capable_candidate` pins it (`same.a` → `same.b`, skipping `no_tool`). Good refinement on top of the typed fix.

### JSON round-trip note (fine)
The new fields serialize to/from their string forms in the `masc_internal_error` JSON (`accept_response_shape_of_string` / `tool_progress_effect_of_string` parse them back). That's the wire boundary, so the on-disk/over-the-wire representation stays string-tagged while the in-memory decision is fully typed — consistent with how `reason_kind` already worked. No concern.

Net: this lands the typed-field promotion, removes every substring classifier from the decision path, unifies the predicate, and adds a test that proves reason-text independence. The change-requested from `270d0ae408` is fully resolved — no remaining concern from me on #21866.

jeong-sik:
Addressed the typed-boundary change request in `84bd2c47b9`.

What changed:
- `Accept_rejected` now carries typed `response_shape` and `last_tool_effect` fields in `Keeper_internal_error`; the human-readable `reason` remains diagnostic text only.
- `Keeper_tool_response` preserves the OAS `Response_shape.content_shape`, and `Keeper_turn_driver_try_provider` carries typed read-only/mutating last-tool effect before formatting the reason text.
- Both retry consumers now route through one typed helper, `accept_rejection_has_read_only_no_progress_retry_hint`; the runtime retry predicate and `Keeper_error_classify` no longer substring-match `reason`.
- Added regression coverage for the exact failure mode: changed diagnostic wording with typed thinking/read-only fields still retries, while legacy `shape=thinking_only; last_tool_effect=read_only` text without typed fields does not retry.

Verification on current head:
- `ocamlformat --check lib/keeper_runtime/keeper_internal_error.ml lib/keeper_runtime/keeper_internal_error.mli lib/keeper_tooling/keeper_tool_response.ml lib/keeper_tooling/keeper_tool_response.mli lib/keeper/keeper_turn_driver_try_provider.ml lib/keeper/keeper_turn_driver_try_provider.mli lib/keeper/keeper_turn_driver_try_runtime.ml lib/keeper/keeper_error_classify.ml lib/keeper/keeper_turn_driver_wrappers.ml test/test_keeper_turn_driver_accept.ml test/test_keeper_sdk_error_typed_bridge.ml`
- `env DUNE_BUILD_DIR=_build-review-21866 DUNE_CACHE=disabled dune build --root . test/test_keeper_turn_driver_accept.exe test/test_keeper_sdk_error_typed_bridge.exe`
- `./_build-review-21866/default/test/test_keeper_turn_driver_accept.exe` -> 18 passed
- `./_build-review-21866/default/test/test_keeper_sdk_error_typed_bridge.exe` -> 13 passed
- `python3 scripts/check_test_coverage.py` -> PASS
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` -> PASS, existing repo-wide WARNs only
- `git diff --check` and `git diff --check origin/main...HEAD` -> PASS

CI can finish the full build.

PR #21861, opened 2026-06-20
feat(dashboard): resurrect keeper-v2 TopBar status chips + keeper breadcrumb

jeong-sik:
## 무엇을 / 왜

keeper-v2 디자인의 **Keeper Agent** 표면 TopBar를 되살립니다 (keeper-v2 포팅 라인 Unit 2 — "TopBar resurrection"). 디자인 SSOT: claude.ai/design `keeper-v2/shell.jsx`의 `TopBar`.

`.v2-app-header-status`(실행 중 keeper 칩 + 스케줄러 칩)는 Tailwind `hidden`이 걸려 있고 이를 되돌리는 CSS 규칙이 없어 **모든 폭에서 `display:none`** 상태였습니다(죽은 칩). 갭 분석(`GAP-keeper-agent-2026-06-20.md`)의 Unit 2 항목입니다. 칩 스킨(`.v2-statchip` / `.live`)은 이미 `app-shell-v2.css`에 존재하므로 wrapper만 되살리면 디자인대로 렌더됩니다.

## 변경 (`dashboard/src/app.ts`)

1. **칩 un-hide**: wrapper를 `hidden` → `v2-desktop-header-only flex`로 교체. 다른 desktop-only 헤더 항목(TransportBeacon/ThemeSwitch 등)과 **동일한 단일 900px 브레이크포인트 SSOT**를 공유 — 일회성 Tailwind variant 대신 컨벤션 사용. 모두 기존 메커니즘이라 #21846 cascade-layer 회귀 위험 없음.
2. **live 도트 펄스**: 정적 도트에 `animate-pulse` 추가 (`common/live-pulse-dot.ts`의 `DOT_TONE.live` 컨벤션 일치, `prefers-reduced-motion`에서 `a11y.css`가 자동 정지).
3. **한국어화**: `N running` → `N 실행 중`, `scheduler healthy/—` → `스케줄러 정상/—`, `aria-label` 한국어화. (스케줄러 칩은 서버 상태 프록시 — 로컬에 스케줄 데이터가 없어 디자인의 예약 큐 pending/due를 **날조하지 않음**.)
[...]

PR #21853, opened 2026-06-20
fix(dashboard): drop redundant global header title (single-source headers 1/2)

jeong-sik:
## 목표 (근본해결 1/2단계)

대시보드 제목 중복의 **systemic 근원**을 제거한다. 전역 shell 헤더가 `<h1 class="v2-header-title">`로 현재 surface 라벨을 **모든 비-compact surface에** 렌더해, 각 surface의 자체 bespoke 헤더(그리고 #21834 이전엔 generic SurfaceLead까지)와 제목이 중복됐다.

#21834(SurfaceLead 화이트리스트)는 중간 레이어를 7개 surface에만 가린 N-of-M band-aid였고, 이 전역 h1(#1 소스)은 손대지 않아 모든 surface에서 `전역 h1 + surface h1` 2겹이 남아 있었다.

## 변경

`app.ts`의 전역 헤더에서 `v2-header-title` `<h1>` 블록 제거 → 헤더 chrome은 **breadcrumb crumb만** 표시(시안 `.v2-top`과 동일: 얇은 crumb, 페이지 제목 없음). surface 자체 헤더가 **단일 제목 소스**가 된다.

- `app.ts`: 전역 h1 제거 (이유 주석 포함).
- `styles/v2-shell.css`: 고아가 된 `.v2-header-title` 규칙 제거.
- `app.test.ts`: `.v2-header-title` 부재를 단언(`toBeNull`)해 결정 잠금.

## 근거 (v2 시안)

시안 top bar(`shell.jsx` `TopBar`)는 `.crumb`(surface 라벨) + stat chips만 두고 **페이지 제목 h1이 없다**. 제목은 각 surface의 bespoke 헤더가 소유. 이 PR은 구현을 그 모델로 정렬한다.

## 검증 (로컬)

- `tsc --noEmit`: 0 errors
- vitest `app.test.ts`: 24 passed
- `npm run lint` (eslint src): clean

## 후속 (2/2단계, 별도 PR)

generic `SurfaceLead` + `SURFACE_OWN_LEAD_IDS` 화이트리스트 완전 제거 + headerless surface 4종(monitoring/board/command/lab)에 bespoke 헤더 추가 + SurfaceLead의 copy-link/solo-view affordance를 각 surface 헤더로 재배치. 그러면 제목 소스가 surface 헤더 1개로 수렴하고 화이트리스트가 불필요해진다.

## 경계

[...]

PR #21867, opened 2026-06-20
fix(dashboard): eliminate SurfaceLead whitelist; surfaces own their header (single-source 2/2)

jeong-sik:
## 목표 (근본해결 2/2)

대시보드 제목 중복의 **구조적 근원**을 제거한다: hand-maintained `SURFACE_OWN_LEAD_IDS` 화이트리스트(N-of-M 워크어라운드)와 이중 헤더 시스템을 없앤다. PR #21853(1/2)은 전역 헤더 h1을 제거했고, 이 PR은 generic `SurfaceLead` 자체를 제거해 **각 surface가 자신의 헤더를 소유**(단일 소스)하게 한다.

## 변경

- `dashboard-shell.ts`: `SurfaceLead` + `SURFACE_OWN_LEAD_IDS` + `shouldRenderSurfaceLead` + dead breadcrumb 헬퍼(`deriveBreadcrumbTrail`/`navigateCrumb`/`breadcrumbItemsForTrail`/`currentSectionShareUrl`) 제거 + 미사용 import 정리.
- `common/surface-header.ts` (신규): 자체 bespoke 헤더가 없는 surface(monitoring/board/command/lab)용 generic 헤더 — section/view 라벨 `<h1>` + description + 공유 actions. surface가 `<SurfaceHeader/>`를 직접 렌더(헤더 결정을 surface에 co-located).
- `common/surface-header-actions.ts` (신규): copy-link + solo-view affordance를 단일 정의로 추출(SurfaceLead에서 보존).
- `status.ts` / `board-surface.ts` / `operations-panel.ts` / `lab.ts`: 각 surface 본문 최상단에 `<SurfaceHeader/>` 배선.
- `config/navigation.ts`: `validSectionIds` / `sectionItemsForTab`를 total function(`?? []`)으로 강화 — unknown/unmapped tab에 crash 대신 빈 섹션 반환.

## 근거

PR #21834의 화이트리스트는 "자체 헤더 보유 surface를 set에 등록"하는 N-of-M 패턴(CLAUDE.md 워크어라운드 거부 기준 3번) — 새 bespoke 헤더 추가 시 등록 누락이 이중헤더 회귀를 낳는다. 이 PR은 헤더 렌더 결정을 각 surface에 co-located시켜 화이트리스트를 소멸시킨다(근접성으로 누락 방지). 시안(v2) 정합: surface마다 단일 bespoke 헤더 + 전역 얇은 crumb, generic lead 없음.

## 검증 (로컬)

- `tsc --noEmit`: 0 errors
- `npm run lint` (eslint src): clean
- vitest: `surface-header`(신규) / `dashboard-shell` / `board-surface` / `lab` / `operations-panel` / `status` / `navigation` 153 passed + `app` 24 passed
- route-reading 헤더가 노출한 surface 테스트들의 불완전 router mock(`hashForRoute` 누락, `route.tab` 누락)을 현실적으로 보강

## 경계 / 범위

- UI 전용. credential / operator control / sandbox / dashboard credential component / hooks / workflow docs 미변경 → RFC 게이트 대상 아님.
- `app.ts` 미변경 → 진행 중인 TopBar revival PR(#21861/#21862, app.ts)과 파일 충돌 없음(상보적).
- 증상 억제가 아니라 root fix: 워크어라운드(화이트리스트) 제거 + partial function total화(`?? []`).

### DIFF
--- a/dashboard/src/components/board/board-surface.test.ts
+++ b/dashboard/src/components/board/board-surface.test.ts
@@ -21,9 +21,10 @@ vi.mock('../../store', async (importOriginal) => {
 })
 
 vi.mock('../../router', () => ({
-  route: { value: { params: {} } },
+  route: { value: { tab: 'board', params: {}, postId: null } },
   navigate: vi.fn(),
   navigateToPost: vi.fn(),
+  hashForRoute: vi.fn(() => '#board'),
 }))
 
 vi.mock('../../api', () => ({
--- a/dashboard/src/components/board/board-surface.ts
+++ b/dashboard/src/components/board/board-surface.ts
@@ -12,6 +12,7 @@ import { Select } from '../common/select'
 import { Checkbox } from '../common/checkbox'
 import { RichContent } from '../common/rich-content'
 import { CursorPagination } from '../common/pagination'
+import { SurfaceHeader } from '../common/surface-header'
 import { stripStateBlocks } from '../../keeper-message'
 import { route } from '../../router'
 import { keepers as dashboardKeepers, messages, refreshExecution } from '../../store'
@@ -1369,6 +1370,7 @@ export function BoardSurface() {
 
   return html`
     <div class="v2-board-surface ss-surface bg-surface-page text-text-primary">
+      <${SurfaceHeader} />
       <div class="bd-body">
         <${BdRail}
           activeSub=${activeSub}
--- a/dashboard/src/components/common/surface-header-actions.ts
+++ b/dashboard/src/components/common/surface-header-actions.ts
@@ -1,0 +1,52 @@
+// SurfaceHeaderActions — the copy-section-link + open-in-solo-view affordances
+// that a surface's own primary header carries. Extracted from the former
+// generic SurfaceLead (removed in the single-source header refactor) so that
+// each surface owns its header while the shared utility affordances live in one
+// place. Reads the live route to derive the share/solo URLs for the current
+// surface+section, so it works regardless of which surface header hosts it.
+import { html } from 'htm/preact'
+import type { VNode } from 'preact'
+import { ExternalLink } from 'lucide-preact'
+import { route } from '../../router'
+import { widgetSoloUrlForRoute } from '../widget-solo'
+import { CopyIdButton } from './copy-id-button'
+import { ringFocusClasses } from './ring'
+
+// window.location is the truth source (the router writes to it already), so the
+// copied link never diverges from the address bar. Returns '' when window is
+// unavailable (SSR / happy-dom without location) so the caller hides the
+// share affordance gracefully.
+function currentShareUrl(): string {
+  if (typeof window === 'undefined' || window.location === undefined) {
+    return ''
+  }
+  return window.location.href
+}
+
+export function SurfaceHeaderActions({ label }: { label: string }): VNode {
+  const soloUrl = widgetSoloUrlForRoute(route.value)
+  const shareUrl = currentShareUrl()
+  return html`
+    <div class="v2-surface-header-actions inline-flex items-center gap-2">
+      ${shareUrl !== ''
+        ? html`<${CopyIdButton}
+            value=${shareUrl}
+            label=${`Section link (${label})`}
+            ariaLabel="Copy current section URL"
+            size=${14}
+          />`
+        : null}
+      <a
+        href=${soloUrl}
+        target="_blank"
+        rel="noopener noreferrer"
+        class=${`v2-shell-action inline-flex size-7 items-center justify-center rounded-[var(--r-1)] border border-[var(--color-border-default)] bg-[var(--color-bg-elevated)] text-[var(--color-fg-muted)] hover:bg-[var(--color-bg-hover)] hover:text-[var(--color-fg-secondary)] ${ringFocusClasses({ tone: 'accent-medium', width: 2, offset: 2, offsetSurface: 'page' })}`}
+        title="Open this surface in a solo view"
+        aria-label="Open this surface in a solo view"
+        data-testid="dashboard-widget-solo-link"
+      >
+        <${ExternalLink} size=${14} aria-hidden="true" />
+      </a>
+    </div>
+  `
+}
--- a/dashboard/src/components/common/surface-header.test.ts
+++ b/dashboard/src/components/common/surface-header.test.ts
@@ -40,4 +40,21 @@ describe('SurfaceHeader', () => {
     expect(container.querySelector('.v2-surface-header-actions')).not.toBeNull()
     expect(container.querySelector('[data-testid="dashboard-widget-solo-link"]')).not.toBeNull()
   })
+
+  it('renders parent breadcrumbs for section routes', () => {
+    route.value = { tab: 'lab', params: { section: 'keeper-memory-health' }, postId: null }
+    render(h(SurfaceHeader, {}), container)
+    const breadcrumb = container.querySelector('[data-testid="surface-breadcrumb"]')
+    expect(breadcrumb).not.toBeNull()
+    expect(breadcrumb?.getAttribute('data-surface-breadcrumb')).toBe('true')
+    expect(breadcrumb?.textContent).toContain('Lab')
+    expect(breadcrumb?.textContent).toContain('키퍼 메모리 상태')
+  })
+
+  it('does not duplicate the body header in widget solo mode', () => {
+    route.value = { tab: 'lab', params: { section: 'tools', solo: '1' }, postId: null }
+    render(h(SurfaceHeader, {}), container)
+    expect(container.querySelector('.v2-surface-header')).toBeNull()
+    expect(container.querySelector('[data-testid="dashboard-widget-solo-link"]')).toBeNull()
+  })
 })
--- a/dashboard/src/components/common/surface-header.test.ts
+++ b/dashboard/src/components/common/surface-header.test.ts
@@ -1,0 +1,43 @@
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { h, render } from 'preact'
+import { SurfaceHeader } from './surface-header'
+import { route } from '../../router'
+
+describe('SurfaceHeader', () => {
+  let container: HTMLDivElement
+
+  beforeEach(() => {
+    container = document.createElement('div')
+    document.body.appendChild(container)
+  })
+
+  afterEach(() => {
+    render(null, container)
+    container.remove()
+    route.value = { tab: 'overview', params: {}, postId: null }
+  })
+
+  // Each surface that does not render a richer bespoke header (monitoring,
+  // command, lab, board) opts into SurfaceHeader, which surfaces the current
+  // section/view label as the single primary h1. These labels match what the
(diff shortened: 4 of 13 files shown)


jeong-sik:
### CHANGE REQUEST (head `7726250657`) - SurfaceHeader move drops section breadcrumbs and leaks headers into widget-solo mode

The whitelist removal direction is good, but this refactor currently changes two user-visible contracts from the old shell-owned `SurfaceLead`.

#### 1. Drilldown section breadcrumbs are dropped

`SurfaceLead` did more than render an `h1`. On section routes it rendered the parent tab breadcrumb:

```tsx
const trail = currentSection !== null
  ? deriveBreadcrumbTrail(currentView?.label ?? null, currentSection.label, currentTab)
  : []

<Breadcrumb
  items={breadcrumbItemsForTrail(trail)}
  ariaLabel="Breadcrumb"
  testId="surface-breadcrumb"
  dataSurfaceBreadcrumb={true}
/>
```

The old comment explicitly justified this for deep links: without `Connectors / Discord`-style parent context, a user opening a section URL only sees the leaf label and must infer the hierarchy from the side nav.

The new `SurfaceHeader` only renders:

```tsx
const title = currentSection?.label ?? currentView?.label ?? 'Home'
...
<h1>...</h1>
<SurfaceHeaderActions ... />
```

No `Breadcrumb`, no `hashForRoute` parent link, and no replacement test for `[data-testid="surface-breadcrumb"]`. That is a behavior regression, not just a relocation.

#### 2. `SurfaceHeader` now renders inside widget-solo mode

Before this PR, `SurfaceLead` was shell-level and `DashboardMain` did not render it in the `soloMode` branch. The solo route already has `WidgetSoloBar`, whose label is the active section/view and whose action is "Full dashboard".

After this PR, `Status`, `OperationsPanel`, `Lab`, and `BoardSurface` render `<SurfaceHeader />` unconditionally inside `TabContent`. `DashboardMain`'s solo branch still renders `TabContent`, so solo mode now gets an extra surface header plus a nested `dashboard-widget-solo-link` that points to solo view again. That is a duplicated title/action surface in the mode that was intentionally compact.

Required direction:

- Move the existing breadcrumb behavior into `SurfaceHeader` (or another shared header primitive) so section routes still expose parent-tab context.
- Preserve the old solo contract: either suppress `SurfaceHeader` when `isWidgetSoloRoute(route.value)` is true, or at least suppress the open-solo affordance and add an explicit test for the chosen behavior.
- Add tests for both cases: a section route renders `surface-breadcrumb`, and a widget-solo route does not render a duplicate `.v2-surface-header` / open-solo link in the body.

jeong-sik:
## Adversarial review — HEAD `7726250657`: eliminate SurfaceLead whitelist; surfaces own their header (single-source 2/2) (+153/-227)

**Strong affirm.** This is the root fix for the open-set pattern I've been flagging as a Low across the recent dashboard header PRs (#21846's `:not()` exclusion chain, #21853's bespoke-heading list, #21857's `:not()` theme list) — and it's the right shape: delete the N-of-M allow-list, give each surface a co-located decision instead.

### [Affirm] The whitelist was an N-of-M list the compiler couldn't enforce; now it's gone
The old `SurfaceLead` in `dashboard-shell.ts` decided per-surface whether to render a generic lead by checking membership in `SURFACE_OWN_LEAD_IDS` — a hand-maintained list. Adding a new surface meant remembering to extend the list; forgetting produced a duplicate title (generic lead stacked on the surface's own header), and nothing in the toolchain failed. The new `SurfaceHeader` component's docstring names this directly: *"an N-of-M list the compiler could not enforce … Now each surface owns the decision to render its header (bespoke or this generic one) at the call site, so there is a single, co-located source for every surface's title."* That's exactly the parse-don't-validate move: instead of a central classifier deciding what each surface gets, each surface declares what it is by rendering `<SurfaceHeader/>` (or its own bespoke header). Net −74 lines and the central `dashboard-shell.ts` shrinks by 179 — the per-surface branching leaves the shell entirely.

### [Affirm] The generic/bespoke split is now co-located and self-documenting
`SurfaceHeader` reads title/description from the nav registry and renders the shared `SurfaceHeaderActions`, so the four generic surfaces (monitoring/board/command/lab, per the docstring) opt in by rendering it; bespoke-header surfaces (overview/approvals/fusion/…) render their own and simply don't call `<SurfaceHeader/>`. There's no second place that has to stay in sync — a surface either has the generic header at its call site or it doesn't. This also makes the #21853 "exactly one h1 per surface" invariant easier to hold, because the h1 lives in whichever header the surface actually renders.

### [Low] Double-render is now possible (surface renders bespoke *and* `<SurfaceHeader/>`)
The old allow-list's one real job was preventing a surface from showing both a generic lead and its own header. With the list gone, a surface that renders its bespoke header *and* drops in `<SurfaceHeader/>` would double up — and unlike the allow-list omission (which was a *missing* entry), this is an *extra* call at the call site, so the compiler still won't catch it. In practice this is less dangerous than the old failure mode (a missing list entry silently duplicated on every new surface; an extra `<SurfaceHeader/>` is a localized mistake one surface makes), and it's reviewable per-surface. If you want a backstop, a `data-testid="surface-header"` assertion (which `SurfaceHeader` already sets) plus a "at most one `surface-header` per surface" check in the surface tests would catch a double-render cheaply. Non-blocking — the trade (eliminate the silent new-surface duplication class, accept a localized double-render class) is a clear win.

### [Low / verify] navigation.ts (+5/-2)
The header reads `currentSection.label`/`.description` from the nav registry, so the `navigation.ts` edit is load-bearing for what the header actually shows. Worth a glance that the new fields are sourced consistently (no parallel hand-maintained label list that re-creates the drift this PR just removed) — if title/description now come from one registry, this fully closes the SSOT loop; if there's a second label source anywhere, that's the next drift surface. Non-blocking.

Net: this is the correct culmination of the single-source-headers series — the allow-list is deleted, surfaces own their header at the call site, the shell shrinks, and the failure mode shifts from "silently wrong on every new surface" to "locally wrong on one surface if its author double-renders." I'd merge; the double-render backstop and the navigation registry check are the only forward-looking asks.

jeong-sik:
Adversarial pass found one test-readiness gap and patched it in 14609ca4a5.

Finding:
- The touched Lab test bundle timed out on the first case when run with the new surface-header wiring included in the same targeted suite. The app behavior looked fine, but the unit test was importing the real keeper-memory-health subtree instead of keeping Lab section routing tests scoped.

Fix:
- Mocked `./memory/keeper-memory-health` in `lab.test.ts` like the other Lab child surfaces.
- Added explicit coverage for the `keeper-memory-health` Lab section so the newly imported branch remains pinned.

Local verification on current head:
- `pnpm exec vitest run --config vitest.config.ts src/components/lab.test.ts --no-file-parallelism --maxWorkers=1` -> 7 passed
- `pnpm exec vitest run --config vitest.config.ts src/components/common/surface-header.test.ts src/components/dashboard-shell.test.ts src/components/board/board-surface.test.ts src/components/lab.test.ts src/components/operations-panel.test.ts src/components/status.test.ts src/config/navigation.test.ts --no-file-parallelism --maxWorkers=1` -> 154 passed
- `pnpm exec tsc --noEmit --pretty false` -> passed

jeong-sik:
Addressed the SurfaceHeader change request in `bf26c938dd`.

What changed:
- Restored section breadcrumbs inside the shared `SurfaceHeader` using the same parent-tab trail shape as the former shell `SurfaceLead` (`surface-breadcrumb`, parent link via `hashForRoute`, current leaf crumb).
- Preserved the widget-solo contract by returning `null` from `SurfaceHeader` when `isWidgetSoloRoute(route.value)` is true, so the solo body no longer gets a duplicate `.v2-surface-header` or nested `dashboard-widget-solo-link` under `WidgetSoloBar`.
- Added explicit tests for both review cases: section route breadcrumb rendering and solo-mode header suppression.

Verification on current head:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/common/surface-header.test.ts --no-file-parallelism --maxWorkers=1` -> 7 passed
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/common/surface-header.test.ts src/components/dashboard-shell.test.ts src/components/board/board-surface.test.ts src/components/lab.test.ts src/components/operations-panel.test.ts src/components/status.test.ts src/config/navigation.test.ts src/components/widget-solo.test.ts --no-file-parallelism --maxWorkers=1` -> 162 passed
- `pnpm --dir dashboard exec eslint src/components/common/surface-header.ts src/components/common/surface-header.test.ts` -> PASS
- `pnpm --dir dashboard exec tsc --noEmit --pretty false` -> PASS
- `python3 scripts/check_test_coverage.py` -> PASS
- `git diff --check` and `git diff --check origin/main...HEAD` -> PASS

CI can finish the full build.

jeong-sik:
## Close-out — HEAD `bf26c938dd`: preserve surface header breadcrumbs (+ widget-solo double-render guard)

This restores the section-route breadcrumb that the whitelist removal lost, and it adds one concrete double-render guard. Both are welcome.

### [Affirm] Section-route breadcrumbs are back, derived from the nav registry
`deriveBreadcrumbTrail` builds the `tab / section` trail from `tabLabel`/`sectionLabel`/`tabId` (so a `lab` → `keeper-memory-health` route renders `Lab / 키퍼 메모리 상태`), and the crumbs reuse the shared `Breadcrumb` component rather than a one-off. `navigateCrumb` correctly defers to standard link behavior — it returns early on `defaultPrevented`, non-primary button, or any modifier key, then `preventDefault` + `navigate` only for a plain primary click — so middle/cmd-click new-tab and friends stay working. That's the right way to wire a clickable crumb on top of a hash router.

### [Resolved, partial] Widget-solo double-render is now guarded
The new `isWidgetSoloRoute` check makes `SurfaceHeader` render nothing in widget-solo mode, and `does not duplicate the body header in widget solo mode` locks it (`.v2-surface-header` is null in solo). That's one concrete instance of the double-render Low I raised on `7726250657` (bespoke header + `<SurfaceHeader/>` both rendering) — closed for the widget-solo case. The general case (a bespoke-header surface that *also* drops in `<SurfaceHeader/>` outside solo mode) is still a localized mistake the compiler won't catch, but with the solo guard landed the residual is smaller and the remaining failure is per-surface rather than systemic. Non-blocking; leaving it as a forward-looking note rather than an open ask.

Net: breadcrumbs restored with correct nav semantics, and the widget-solo double-render class closed. The surface-header series is in good shape on this head.

jeong-sik:
## Close-out - HEAD `bf26c938dd`: SurfaceHeader breadcrumb and solo regressions fixed

Rechecked the current head after the `7726250657` change request.

Confirmed:

- `SurfaceHeader` now renders the section breadcrumb through the existing `Breadcrumb` primitive with `data-testid="surface-breadcrumb"` and `data-surface-breadcrumb="true"`.
- Parent tab crumbs are navigable through `hashForRoute` / `navigate`, preserving the old `SurfaceLead` deep-link context.
- `SurfaceHeader` returns `null` in widget-solo mode, so the body no longer duplicates the `WidgetSoloBar` title or adds another open-solo link.
- Tests cover both contracts: section route breadcrumb rendering and solo-mode suppression.

Local verification:

- `git diff --check origin/main...refs/remotes/pr/21867` - PASS

No remaining blocker from this pass.

PR #21876, opened 2026-06-21
fix(keeper): persist event queue wake before registry replay

jeong-sik:
## Summary

- close a restart/register boundary gap in Keeper Event Layer durable replay
- when `Keeper_registry_event_queue.enqueue` receives a stimulus before the keeper is registered, persist it to the per-keeper event-queue snapshot instead of dropping it
- make unregistered `snapshot` read the durable snapshot so diagnostics and later registration see the same pending replay state
- add regression coverage for enqueue-before-register replay

## Why

The merged durable event queue only protected the registered-entry path: CAS into memory, then mirror the queue to disk. A wake that arrives during startup or registry churn could still hit `keeper not registered` and be logged/dropped. That is exactly the restart window durable replay is supposed to remove.

This keeps the queue in MASC. OAS remains uninvolved.

## Verification

Passed:

- `opam exec -- ocamlformat --check lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_event_queue.mli test/test_keeper_event_queue.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_registry_event_queue.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_registry_event_queue.mli`
- `ocamlc -stop-after parsing test/test_keeper_event_queue.ml`
- `git diff --check`

Not run:

- `scripts/dune-local.sh build test/test_keeper_event_queue.exe` was blocked by the repo guard because existing bare Dune processes were already running on the machine. Leaving focused build/test to CI rather than bypassing the guard.

### DIFF
--- a/lib/keeper/keeper_registry_event_queue.ml
+++ b/lib/keeper/keeper_registry_event_queue.ml
@@ -18,9 +18,9 @@ let enqueue_if_missing queue stimulus =
   if queue_contains queue stimulus then queue else Keeper_event_queue.enqueue queue stimulus
 ;;
 
-let persist_live_queue entry name =
+let persist_live_queue ~base_path (entry : Keeper_registry.registry_entry) name =
   Keeper_event_queue_persistence.persist
-    ~base_path:entry.base_path
+    ~base_path
     ~keeper_name:name
     (Atomic.get entry.event_queue)
 ;;
@@ -45,7 +45,7 @@ let enqueue ~base_path name stimulus =
          if next = cur
          then ()
          else if Atomic.compare_and_set entry.event_queue cur next
-         then persist_live_queue entry name
+         then persist_live_queue ~base_path entry name
          else loop ()
        in
        loop ())
@@ -54,7 +54,7 @@ let enqueue ~base_path name stimulus =
       let cur = Atomic.get entry.event_queue in
       let next = Keeper_event_queue.enqueue cur stimulus in
       if Atomic.compare_and_set entry.event_queue cur next
-      then persist_live_queue entry name
+      then persist_live_queue ~base_path entry name
       else loop ()
     in
     loop ()
@@ -77,7 +77,7 @@ let dequeue ~base_path name =
       | Some (stim, rest) ->
         if Atomic.compare_and_set entry.event_queue cur rest
         then (
-          persist_live_queue entry name;
+          persist_live_queue ~base_path entry name;
           Some stim)
         else loop ()
     in
@@ -93,7 +93,7 @@ let drain_board ?window_sec ~base_path name =
       let board, rest = Keeper_event_queue.drain_board_window ?window_sec cur in
       if Atomic.compare_and_set entry.event_queue cur rest
       then (
-        persist_live_queue entry name;
+        persist_live_queue ~base_path entry name;
         board)
       else loop ()
     in
--- a/lib/keeper/keeper_registry_event_queue.ml
+++ b/lib/keeper/keeper_registry_event_queue.ml
@@ -15,9 +15,10 @@ let enqueue ~base_path name stimulus =
       "registry: enqueue_event name=%s base_path=%s: keeper not registered; persisting stimulus for replay"
       name
       base_path;
-    let cur = Keeper_event_queue_persistence.load ~base_path ~keeper_name:name in
-    let next = Keeper_event_queue.enqueue cur stimulus in
-    Keeper_event_queue_persistence.persist ~base_path ~keeper_name:name next
+    Keeper_event_queue_persistence.update
+      ~base_path
+      ~keeper_name:name
+      (fun cur -> Keeper_event_queue.enqueue cur stimulus)
   | Some entry ->
     let rec loop () =
       let cur = Atomic.get entry.event_queue in
--- a/lib/keeper/keeper_registry_event_queue.ml
+++ b/lib/keeper/keeper_registry_event_queue.ml
@@ -8,6 +8,23 @@
     Atomic state primitive. CAS-successful mutations are mirrored to
     [Keeper_event_queue_persistence] for restart replay. *)
 
+let rec queue_contains queue stimulus =
+  match Keeper_event_queue.dequeue queue with
+  | None -> false
+  | Some (head, rest) -> head = stimulus || queue_contains rest stimulus
+;;
+
+let enqueue_if_missing queue stimulus =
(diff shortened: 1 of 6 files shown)


jeong-sik:
Adversarial follow-up - HEAD `5f48ca05c2`: pre-registry event queue persistence race fixed.

Initial review found a remaining durability hole in the first patch: the unregistered path changed from drop to durable write, but it used `load -> enqueue -> persist`. Two concurrent wakes for the same unregistered keeper could both load the same snapshot and the later writer could overwrite the earlier stimulus. That still violates the "persist wake before registry replay" contract.

Patch pushed:
- added `Keeper_event_queue_persistence.update`, which holds the existing persistence write lock across load + transform + atomic save.
- changed `Keeper_registry_event_queue.enqueue` unregistered path to use that update helper instead of separate load/persist calls.
- extended the pre-registration replay regression to enqueue two stimuli before registration and verify both replay in order after registration.

Local verification:
- `opam exec -- ocamlformat --check lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_event_queue.mli test/test_keeper_event_queue.ml` - passed
- `ocamlc -stop-after parsing` on the changed `.ml/.mli` files and `test/test_keeper_event_queue.ml` - passed
- `git diff --check` - passed

Focused Dune build was attempted with `scripts/dune-local.sh build test/test_keeper_event_queue.exe`, but the repo guard refused because existing bare Dune processes are already running on the machine. CI needs to provide the executable-level proof.

jeong-sik:
Self-review / review-response update on HEAD `34dbb6ed2c`.

I found and fixed two durability races in the first cut:

- pre-registry enqueue used `load -> enqueue -> persist` as separate operations, so concurrent unregistered wakes for the same keeper could lost-update the durable snapshot. Commit `5f48ca05c2` adds `Keeper_event_queue_persistence.update`, holding the persistence write lock across load/transform/write, and extends the replay regression to two queued pre-registration stimuli.
- registered CAS paths persisted the local `next/rest` value captured before the file write. If a later CAS produced a newer in-memory queue but the older persist ran last, the durable file could be stale. Commit `34dbb6ed2c` makes registered enqueue/dequeue/drain persist the live atomic snapshot and makes registration refresh/merge the durable snapshot after `put_entry`, closing the register/enqueue crossing window.

Local checks on the pushed head:
- `opam exec -- ocamlformat --check lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_event_queue.mli lib/keeper/keeper_registry_setup.ml test/test_keeper_event_queue.ml`
- `opam exec -- ocamlc -stop-after parsing lib/keeper/keeper_registry_event_queue.ml`
- `opam exec -- ocamlc -stop-after parsing lib/keeper/keeper_registry_setup.ml`
- `opam exec -- ocamlc -stop-after parsing test/test_keeper_event_queue.ml`
- `git diff --check`

Focused Dune build was retried and still refused by `scripts/dune-local.sh` because unrelated bare `dune build` processes are active on the machine. I did not override the guard; CI is now running on the new head.

jeong-sik:
Adversarial follow-up - HEAD `a1c3bf3daf`: closed a remaining stale durable snapshot overwrite race.

The previous fix changed registered CAS paths to persist `Atomic.get entry.event_queue`, but that snapshot was still evaluated before acquiring the persistence write lock. A writer that captured an older live queue could wait behind a newer writer and then overwrite the durable file with the stale snapshot.

Patch pushed:
- added `Keeper_event_queue_persistence.persist_snapshot`, which evaluates the snapshot thunk while holding the persistence write lock.
- changed registered enqueue/dequeue/drain persistence to use `persist_snapshot`.
- changed registration refresh persistence to use `persist_snapshot` after the CAS merge, so a late refresh writer cannot clobber a newer live queue snapshot.

Local verification:
- `opam exec -- ocamlformat --check lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_setup.ml lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli` - passed
- `opam exec -- ocamlc -stop-after parsing lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_setup.ml lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli` - passed
- `git diff --check` - passed

Focused Dune build was attempted with `scripts/dune-local.sh build test/test_keeper_event_queue.exe`, but the repo guard refused because bare `dune build` processes are already active on this machine. I did not override the guard; CI is running on the new head.

jeong-sik:
CI failure response reviewed on current head a1c3bf3. The earlier failed Health/Lint logs were the same compile error: Keeper_registry_event_queue tried to read entry.base_path where the helper had no registry_entry type context. That was fixed by threading caller base_path into persist_live_queue and annotating the entry as Keeper_registry.registry_entry. The follow-up commit then moves live queue persistence to persist_snapshot so a writer waiting on the file lock snapshots the current Atomic queue instead of writing an older captured queue.

Local checks on current head:
- opam exec -- ocamlc -stop-after parsing on touched .ml files

Not run: scripts/dune-local.sh build @check refused due machine-wide bare Dune processes. CI is still the build authority here.

jeong-sik:
Adversarial re-review on latest head a1c3bf3dafec: prior lost-update blocker remains fixed. The important part is that live queue CAS paths now persist via persist_snapshot, and the snapshot thunk is evaluated while holding the persistence write lock, so a writer that waited on the file lock cannot overwrite a newer live queue with an older post-CAS value. The unregistered enqueue path also uses update(load+transform+write under the same lock) and then re-checks the registry to merge into a keeper that appeared during the unavailable window; merge_event_queues dedupes the same stimulus across durable/live replay rather than duplicating it. Local evidence: opam exec -- ocamlformat --check on the touched ML/MLI/test files; opam exec -- ocamlc -stop-after parsing on the same files; git diff --check. Focused scripts/dune-local.sh build test/test_keeper_event_queue.exe is still blocked locally by pre-existing bare dune build processes, not by this patch. GitHub CI: all visible checks except Build and Test were pass/pending at last refresh, with TLA and ocamlformat already pass.

PR #21878, opened 2026-06-21
[codex] fix runtime multimodal history capability gate

jeong-sik:
## Summary
- Validate prior conversation media as part of the runtime multimodal capability floor gate, not only the current goal blocks.
- Add a regression test for the screenshot-class failure: a text-only follow-up after a vision/image turn must be rejected by MASC capability validation before provider dispatch.

## Root Cause
The reroute decision already considered `initial_messages + goal_blocks`, so a prior image could trigger a capable-runtime reroute. But when no capable runtime existed, the final capability gate inside `Runtime_agent.run_blocks` still validated only `goal_blocks`. That allowed prior image history to reach text-only providers and surface as provider 400 errors such as `messages.content.type is invalid, allowed values: [text]`.

## Validation
- `git diff --check`
- `opam exec -- ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli test/test_runtime_modality_reroute.ml`
- Attempted `scripts/dune-local.sh build test/test_runtime_modality_reroute.exe`; initially blocked by existing bare Dune builds on the machine. Retried with `MASC_DUNE_ALLOW_BARE_DUNE=1`, but stopped the focused build after prolonged contention to avoid worsening local Dune pressure. CI should be treated as the OCaml build/test authority for this draft.

## Notes
This is intentionally backend/runtime-side rather than dashboard-only, so dashboard, connector, queue, and resumed keeper turns share the same fail-closed behavior.

### DIFF
--- a/lib/runtime/runtime_agent.ml
+++ b/lib/runtime/runtime_agent.ml
@@ -852,15 +852,10 @@ let run_blocks
     ?goal_detail
     (goal_blocks : Agent_sdk.Types.content_block list)
   : (run_result, Agent_sdk.Error.sdk_error) result =
-  let validation_blocks =
-    content_blocks_for_run
-      ~initial_messages:config.initial_messages
-      ~goal_blocks
-  in
   match
     validate_content_blocks_for_config
       ~config
-      validation_blocks
+      goal_blocks
   with
   | Error _ as err -> err
   | Ok () ->
--- a/lib/runtime/runtime_agent.ml
+++ b/lib/runtime/runtime_agent.ml
@@ -479,6 +479,16 @@ let validate_content_blocks_against_capabilities
              ~supported
              ~reason:"provider does not support combined non-text modalities")
 
+let validate_content_blocks_for_run_against_capabilities
+    ~(provider_label : string)
+    (caps : Llm_provider.Capabilities.capabilities)
+    ~(initial_messages : Agent_sdk.Types.message list)
+    ~(goal_blocks : Agent_sdk.Types.content_block list) =
+  validate_content_blocks_against_capabilities
+    ~provider_label
+    caps
+    (content_blocks_for_run ~initial_messages ~goal_blocks)
+
 let apply_runtime_model_input_capabilities
     (caps : Llm_provider.Capabilities.capabilities)
     (model_caps : Runtime_schema.model_capabilities) =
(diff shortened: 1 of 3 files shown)


jeong-sik:
적대적 리뷰 follow-up.

확인한 결함: `validate_content_blocks_for_config`가 이제 `config.initial_messages`를 내부에서 합치는데, `run_blocks`가 이미 `content_blocks_for_run`으로 합친 리스트를 다시 넘겨 히스토리를 중복 검증하고 있었습니다. 현재 modality set dedup 때문에 사용자 증상으로 바로 터지지는 않지만, 최종 gate 계약이 틀어져 다음 검증 확장 때 위험한 형태였습니다.

대응: `0aae339c6c`에서 `run_blocks`가 helper 계약대로 current `goal_blocks`만 넘기도록 정리했습니다.

검증:
- `opam exec -- ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli test/test_runtime_modality_reroute.ml`
- `ocamlc -stop-after parsing lib/runtime/runtime_agent.ml`
- `ocamlc -stop-after parsing lib/runtime/runtime_agent.mli`
- `ocamlc -stop-after parsing test/test_runtime_modality_reroute.ml`
- `git diff --check`

미실행: `scripts/dune-local.sh build test/test_runtime_modality_reroute.exe`는 기존 bare Dune 프로세스 때문에 repo wrapper guard가 막았습니다. 원격 CI를 계속 봐야 합니다.

jeong-sik:
Adversarial review - HEAD `0aae339c6c`: runtime multimodal history capability gate

No blocker found on the current head.

What I checked:
- Reroute and final floor validation now use the same active run view: prior `initial_messages` plus current `goal_blocks`.
- The final gate is still MASC runtime-side and fails before provider dispatch, so text-only providers do not receive retained image history and return provider-shaped 400s.
- The earlier contract issue where `run_blocks` built `initial_messages + goal_blocks` and then passed that into a helper that prepended `initial_messages` again is fixed on this head; `run_blocks` now passes `goal_blocks` directly to `validate_content_blocks_for_config`.
- The regression covers the screenshot-class flow: prior image history plus text-only follow-up against text-only caps yields `InvalidConfig { field = "multimodal_input" }` with required/supported detail.
- Local checks from the review worktree: `git diff --check origin/main...HEAD` and `opam exec -- ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli test/test_runtime_modality_reroute.ml` both passed. GitHub CI is still running Build and Test / core gates on the new head.

Review verdict: correct boundary-local runtime fix; no OAS/provider changes needed.

PR #21884, opened 2026-06-21
fix(keeper): separate strict read-only retry safety

jeong-sik:
## Summary

Adversarial hardening follow-up for #21866.

- Split strict non-mutating read-only checks from the broader "effectively read-only / idempotent" retry-safe classification.
- Use the strict input-aware read-only check when building the `last_tool_effect` that gates `read_only_no_progress` fallback.
- Add regression coverage that workspace mutations (`keeper_board_post`, `keeper_broadcast`, `masc_transition`) stay `mutating` and terminal after a thinking-only no-progress response.

## Product impact

The prior recovery path still retries after genuine read-only tools like `WebFetch`, `Read`, and `Grep`.

The follow-up removes a weak proof boundary: idempotent or workspace-scoped mutating tools must not be treated as "no committed mutation" for the no-side-effect runtime retry path.

## Evidence

- `git diff --check`
- `opam exec -- ocamlformat --check lib/keeper/keeper_tool_registry.ml lib/keeper/keeper_tool_registry.mli lib/keeper/keeper_turn_driver_try_provider.ml test/test_keeper_tool_descriptor_registry_integrity.ml test/test_keeper_turn_driver_accept.ml`
- Local Dune build was started, then stopped at user request: CI will run the build/test suite.

## Direct evidence

schema_version: 1
direct_ratio: 4/4
provenance: direct

- New strict APIs: `is_strictly_read_only_tool`, `is_strictly_read_only_with_input`.
- `last_tool_progress_context_of_messages` now uses strict read-only classification for retry-safety evidence.
- Registry tests prove strict read-only includes `Grep`/`Read`/`WebFetch` and excludes workspace mutation tools.
- Accept/retry tests prove thinking-only after workspace mutation does not try the next candidate and is not runtime-recoverable.

## Review evidence

Adversarial finding: the original "no side effect" statement was stronger than the previous predicate, because the old predicate reused an "effectively read-only" helper whose documented meaning included idempotent retry-safe tools. This PR narrows the retry gate to actual non-mutating read-only tools.

## Base

Follow-up to merged #21866.

### DIFF
--- a/lib/keeper/keeper_tool_registry.ml
+++ b/lib/keeper/keeper_tool_registry.ml
@@ -139,6 +139,14 @@ let is_effectively_read_only_tool (name : string) : bool =
   || Keeper_tool_descriptor_resolution.capability_has Tool_capability.Idempotent name
 ;;
 
+let is_strictly_read_only_tool (name : string) : bool =
+  (* Unlike [is_effectively_read_only_tool], this intentionally excludes
+     idempotent-but-mutating tools. Use it for retry paths that require no
+     committed mutation rather than merely retry-safe mutation. *)
+  is_keeper_read_only_tool name
+  || Keeper_tool_descriptor_resolution.capability_has Tool_capability.Read_only name
+;;
+
 let has_mutating_side_effect (name : string) : bool =
   not (is_effectively_read_only_tool name)
 ;;
@@ -152,6 +160,16 @@ let is_read_only_with_input ~(tool_name : string) ~(input : Yojson.Safe.t) : boo
   match Keeper_tool_descriptor_resolution.readonly_for_tool_call ~tool_name ~input with
   | Some readonly -> readonly
   | None -> is_effectively_read_only_tool tool_name
+;;
+
+let is_strictly_read_only_with_input
+      ~(tool_name : string)
+      ~(input : Yojson.Safe.t)
+  : bool
+  =
+  match Keeper_tool_descriptor_resolution.readonly_for_tool_call ~tool_name ~input with
+  | Some readonly -> readonly
+  | None -> is_strictly_read_only_tool tool_name
 ;;
 
 let descriptor_boundary_exempt tool_name =
--- a/lib/keeper/keeper_tool_registry.mli
+++ b/lib/keeper/keeper_tool_registry.mli
@@ -44,12 +44,24 @@ val is_keeper_read_only_tool : string -> bool
     read-only/idempotent classification. *)
 val is_effectively_read_only_tool : string -> bool
 
+(** Strict non-mutating read-only check. Unlike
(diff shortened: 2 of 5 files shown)


jeong-sik:
Adversarial review on latest head b0559c2b9408: no blocker found in the strict read-only split. The important boundary is now separate: is_effectively_read_only_* still covers retry-safe/idempotent semantics, while is_strictly_read_only_* excludes idempotent-but-mutating tools for the read_only_no_progress fallback. I checked the downstream gate: recoverable_runtime_failure_reason still uses the typed Accept_rejected envelope with response_shape=thinking_only and last_tool_effect=read_only, not reason-token matching. Tests added for Grep/Read/WebFetch positive cases and keeper_board_post/keeper_broadcast/masc_transition negative cases cover the core workspace-mutation risk. Remaining gate: CI is still pending on Build and Test / TLA at the last check.

jeong-sik:
Adversarial review - HEAD `b0559c2b94`: strict read-only retry safety

No blocker found.

What I checked:
- The previous `is_effectively_read_only` path intentionally still treats read-only/idempotent tools as retry-safe for broad registry semantics.
- The no-progress provider retry path now uses `is_strictly_read_only_with_input`, so idempotent-but-mutating workspace tools are no longer classified as no-side-effect read-only.
- Descriptor resolution still handles public/internal/MCP-prefixed names before evaluating input-aware readonly policy.
- Workspace mutations covered in the added tests (`keeper_board_post`, `keeper_broadcast`, `masc_transition`) now remain terminal after thinking-only/no-progress output instead of trying the next runtime candidate.
- Existing read-only recovery path remains covered by `WebFetch` and public file/search aliases.

Local focused checks:
- `opam exec -- ocamlformat --check lib/keeper/keeper_tool_registry.ml lib/keeper/keeper_tool_registry.mli lib/keeper/keeper_turn_driver_try_provider.ml test/test_keeper_tool_descriptor_registry_integrity.ml test/test_keeper_turn_driver_accept.ml` => pass
- `opam exec -- ocamlc -stop-after parsing ...` on the same touched files => pass
- `git diff --check origin/main...HEAD` => pass

Full dune/CI remains pending on GitHub.

jeong-sik:
Review pass on head b0559c2. No blocker found. The change correctly separates effective read-only/idempotent retry safety from strict no-mutation retry safety: last_tool_progress_context now uses is_strictly_read_only_with_input, so keeper_board_post, keeper_broadcast, and masc_transition stay mutating after a thinking-only/no-progress rejection and do not trigger next-candidate retry.

Local checks:

Not run: focused dune build for test/test_keeper_turn_driver_accept.exe and test/test_keeper_tool_descriptor_registry_integrity.exe refused due machine-wide bare Dune processes. CI remains the build authority here.

PR #21887, opened 2026-06-21
fix(keeper): summarize tool effects for read-only retry

jeong-sik:
## Summary
- add turn-wide tool effect summary to accept-rejection diagnostics (`any_mutating_tool`, `tool_effects_seen`)
- require `any_mutating_tool=false` before treating thinking-only read-only accept rejection as runtime-recoverable
- add fixture for `Write -> Read -> thinking_only` so earlier mutation blocks retry
- update the heuristic gap audit note for the closed last-tool false-positive slice

## Validation
- `ocamlformat --inplace` on changed OCaml files
- Local focused Dune build was started but stopped at user direction; CI owns build/test validation for this PR.

### DIFF
--- a/docs/audit/2026-06-21-keeper-heuristic-realworld-gap-list.md
+++ b/docs/audit/2026-06-21-keeper-heuristic-realworld-gap-list.md
@@ -45,8 +45,8 @@ OAS.
 
 | # | Surface | Current heuristic | Missing example class | Risk |
 |---|---------|-------------------|-----------------------|------|
-| 1 | Thinking-only read-only retry | accept reason substring gate | actual no-progress trace after read-only tool | retry can hide accept failures |
-| 2 | Last-tool effect context | only the final assistant tool call is inspected | multi-tool turn with earlier mutation | read-only label can be unsound |
+| 1 | Thinking-only read-only retry | typed response/effect gate | read-only error/empty payload trace | retry can hide accept failures |
+| 2 | Last-tool effect context | final tool plus turn-wide effect summary | compaction/resume multi-message fixture | read-only label can be unsound |
 | 3 | Tool read-only registry | descriptor/capability fallback | mixed-mode tool contracts | checkpoint boundary can drift |
 | 4 | Provider error string classifiers | quota/capacity/terminal substrings | provider/CLI error corpus | wrong cooldown/escalation |
 | 5 | Keeper warn/recoverable classifier | typed bucket allow/deny list | operator-facing failure samples | noisy or silent incidents |
@@ -64,26 +64,21 @@ Evidence:
 
 Current behavior: typed `Accept_rejected` with
 `Accept_no_usable_progress` is retryable on the next runtime only when the
-free-form reason contains both `shape=thinking_only` and
-`last_tool_effect=read_only`.
+typed response shape is `thinking_only`, the final tool is `read_only`, and the
+turn-wide checkpoint summary reports `any_mutating_tool=false`.
 
 Missing real-world examples:
-- WebFetch/Read/Grep succeeds, then the runtime returns only thinking text.
 - A read-only tool returns an error or empty payload, then the runtime returns
   only thinking text.
-- The last tool is read-only but an earlier tool in the same turn mutated
-  workspace or board state.
 - The rejection happens on the final candidate, where retry is intentionally
   impossible.
 
-Operational risk: the fix is intentionally narrow, but its safety rests on a
-string-decorated reason. Without fixture traces for the false-positive
-boundary, a future reason-format change can silently stop retrying, or a
-misclassified read-only context can retry after a turn that already had side
-effects.
+Operational risk: the fix is intentionally narrow. The earlier string-decorated
+reason gate is now typed, but remaining confidence still depends on broader
+fixture traces for read-only tool failures and final-candidate behavior.
 
 Evidence fixture needed: a raw checkpoint/rejection fixture with multiple tools,
-including at least one read-only-only turn and one earlier-mutation turn.
+including final-candidate and read-only error/empty-payload cases.
 
 ## 2. Last-tool effect context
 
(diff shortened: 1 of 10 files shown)


jeong-sik:
Adversarial review - HEAD `4f9b7cee9`: tool-effect summary for read-only retry

No blocker found on the current head.

What I checked:
- The old last-tool-only heuristic is replaced with a checkpoint-wide tool-effect summary, so a mutation followed by a read-only tool no longer qualifies for `read_only_no_progress` retry just because the last tool was read-only.
- The retry hint now requires typed fields: `response_shape=thinking_only`, `last_tool_effect=read_only`, `any_mutating_tool=false`, and a non-empty `tool_effects_seen`. Legacy reason-string tokens alone still do not trigger retry.
- JSON serialization/parsing carries `any_mutating_tool` and `tool_effects_seen` through the typed `Accept_rejected` envelope.
- Existing wrapper call sites were updated with explicit `None`/empty summary values, so older non-checkpoint wrapper errors stay non-retryable.
- The added regression covers mutation -> read -> thinking-only and keeps it terminal while preserving the pure WebFetch/read-only retry path.

Local focused checks:
- `opam exec -- ocamlformat --check lib/keeper/keeper_turn_driver_try_provider.ml lib/keeper/keeper_turn_driver_try_provider.mli lib/keeper/keeper_turn_driver_wrappers.ml lib/keeper_runtime/keeper_internal_error.ml lib/keeper_runtime/keeper_internal_error.mli test/test_keeper_sdk_error_typed_bridge.ml test/test_keeper_turn_driver_accept.ml` => pass
- `opam exec -- ocamlc -stop-after parsing ...` on the same touched files => pass
- `git diff --check origin/main...HEAD` => pass

Stacking note: #21884 touches the same classifier boundary and introduces strict non-mutating read-only semantics. If both PRs are kept, rebase/merge must preserve that strict classifier instead of silently falling back to the broader `is_read_only_with_input` semantics.

jeong-sik:
Review pass on head 4f9b7ce found a readiness blocker before this leaves draft.

The new any_mutating_tool summary is the right shape, but this branch still classifies each tool with Keeper_tool_registry.is_read_only_with_input. That is the effective read-only/idempotent predicate, not the strict no-committed-mutation predicate added in #21884. So if this PR is merged or made ready independently, idempotent-but-mutating tools such as keeper_board_post, keeper_broadcast, or masc_transition can still be summarized as read-only, and the read-only no-progress retry lane can be re-enabled after a committed workspace mutation.

Required before ready:
- rebase/stack this after #21884 and use is_strictly_read_only_with_input for the tool-effect summary, or carry the same strict predicate here
- add the mutation-then-read regression with one of the idempotent workspace mutations, not only Write -> Read

Local checks:

Not run: focused dune build refused due machine-wide bare Dune processes.

jeong-sik:
Latest-head adversarial review on `f26d15b4c2b79547327c03c92e04aa605e0d88fe`: no blocker found.

What I checked:
- The retry gate is stricter now: `Accept_rejected` must be `no_usable_progress`, `thinking_only`, final tool effect `read_only`, `any_mutating_tool=Some false`, and non-empty `tool_effects_seen`.
- The checkpoint scan records all assistant tool uses, keeps the final tool visible for diagnostics, and separately summarizes whether any mutating tool appeared earlier in the same checkpoint span.
- Constructor/cross-boundary updates are present for the wrapper path and typed SDK error bridge; legacy/string-only reason text does not trigger retry.
- Added regression covers the important false-positive: mutating tool followed by read-only tool then thinking-only response must not try the next candidate.

Local non-build evidence:
- `git diff --check origin/main...refs/remotes/pr/21887`
- `git grep` over the PR head for `Accept_rejected`, `any_mutating_tool`, `tool_effects_seen`, and `last_tool_effect`

CI snapshot at review time: format/lint/health/ratchets green; Build and Test, Dashboard, and TLA+ still in progress. Build authority stays with CI per operator direction.

jeong-sik:
Rechecked current head `f26d15b4c2` after the merge churn.

The earlier conflict/read-only-classifier concern is resolved on this head: `classify_tool_effect` now goes through `Keeper_tool_registry.is_strictly_read_only_with_input`, and the accept tests cover mutating workspace tools plus mutation-then-read history so read-only retry stays disabled when any prior tool mutated state.

Local checks:
- `git diff --check` passed
- `opam exec -- ocamlformat --check ...` on touched OCaml files passed
- `opam exec -- ocamlc -stop-after parsing ...` on touched OCaml files passed
- focused `scripts/dune-local.sh build test/test_keeper_turn_driver_accept.exe test/test_keeper_sdk_error_typed_bridge.exe` was not started because the machine-wide Dune guard detected existing bare Dune processes

Current reviewThreads are empty. No remaining code-review blocker from my pass; CI is still the build authority.

jeong-sik:
Adversarial review found and fixed a real scope bug.

Finding:
- The new `any_mutating_tool` summary scanned `checkpoint.messages` wholesale. In the live keeper path, MASC passes `history_messages` as OAS `initial_messages`, OAS initializes `Agent.state.messages = config.initial_messages`, and checkpoints serialize `state.messages`. So a mutating tool from an older turn could be counted as a current-attempt mutation and permanently disable the intended read-only no-progress retry, even when the current provider attempt only performed `Read`/`WebFetch`.

Fix pushed in `ad3dd45586f9ce554b376cb10630479a004d613d`:
- `accept_rejection_context_of_run_result` now accepts `~initial_messages` and strips the exact checkpoint prefix before summarizing tool effects.
- Production paths pass `ctx.initial_messages` / `ctx.try_provider_ctx.initial_messages`, so the summary is scoped to the current provider attempt while preserving previous failed attempts inside the same turn.
- Added regression coverage: historical `keeper_board_post` in initial history + current `Read` still yields `any_mutating_tool=false`, `tool_effects_seen=read_only`, and allows `read_only_no_progress` retry.

Local checks:
- `opam exec -- ocamlformat --check ...` passed for changed OCaml files.
- `opam exec -- ocamlc -stop-after parsing ...` passed for changed OCaml files.
- `git diff --check` passed.
- `scripts/dune-local.sh build test/test_keeper_turn_driver_accept.exe` was not run locally: it refused because existing bare `dune build` processes are active on the machine. CI remains the build/test authority here.

Live PR state after push: non-draft, no unresolved review threads, new head has no status rollup yet.

PR #21891, opened 2026-06-21
[codex] Fix dashboard keeper context controls

jeong-sik:
## Summary

- Wire the keeper context panel to the existing `masc_keeper_compact` MCP tool with a `force=true` operator action.
- Preserve backend compaction gate metadata in the dashboard store and show ratio/profile/message/token gate details instead of a hardcoded 85% threshold.
- Fix dashboard keeper context max fallback so checkpoint-derived summaries use the keeper runtime budget instead of a zero max context.
- Rename the misleading scheduler chip to a server-response chip and adjust mini keeper rail padding.

## Root Cause

The dashboard had three different truth gaps: frontend normalization dropped compaction gate fields already emitted by the backend, the context panel hardcoded its compact threshold and had no operator compaction action, and the dashboard summary checkpoint fallback passed `primary_max_context = 0`, which made context max appear missing even when runtime budget data was available.

## Validation

- `pnpm exec vitest run --config vitest.config.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts src/app.test.ts src/keeper-store-normalize.test.ts --no-file-parallelism --maxWorkers=1` (`82` tests passed)
- `git diff --check`
- Playwright render smoke against local Vite dashboard: keepers loaded, `Loading Keepers=false`, compact button visible/enabled
- Live health checked via `curl -fsS 'http://127.0.0.1:8935/health?full=1'`: runtime root `/Users/dancer/me/.masc`

OCaml local build was intentionally not run for this PR; CI will cover it per operator request. Earlier local `scripts/dune-local.sh build test/test_dashboard_keeper_worker_degrade_source.exe` was blocked by active bare Dune processes on the machine, and I did not bypass that guard.


jeong-sik:
Reviewed current head `214c004a9c`.

The compact control calls `masc_keeper_compact` with schema-compatible `{ name, force: true }`, keeps the auth-role guard, and the max-context fallback now resolves to the runtime budget. I did not find a blocker in this pass.

Local: `git diff --check origin/main...origin/pr/21891` passed. I left Vitest/Dune to CI because local dashboard deps are absent in the detached review refs and the Dune guard is refusing new builds.

jeong-sik:
Latest-head adversarial read-through (`214c004a9c`): no blocker found.

Checked the keeper context controls path:
- compact action remains behind dashboard worker auth via `dashboardAuthAccess(..., "worker")`.
- frontend call shape matches backend `masc_keeper_compact` args (`name`, `force`) and handles the JSON result path.
- backend context fallback now uses the resolved turn context budget instead of a zero fallback, so missing live context data should not render as an impossible 0-token state.

No local build/test run per operator direction; CI remains build/test authority.

PR #21897, opened 2026-06-21
docs(rfc): RFC-0272 — Memory OS episode log retention (defect D)

jeong-sik:
## What

RFC-0272 routes **defect D** from Issue #21789 (Memory OS audit: `events.jsonl` /
`episodes/` unbounded append) into its own forgetting-charter RFC. Docs-only.

## Why a new RFC (not amending RFC-0247)

RFC-0259 §6 deferred D to RFC-0247 (forgetting charter), but: (1) RFC-0247 is
**facts-only** — neither 0247 doc mentions the episode log; (2) amending its
`## §1`-style body trips the `rfc-enforcer` R5 caller-context gap (the
`.tmp/rfc-NNNN-caller-context.md` is gitignored, so the check is unsatisfiable
for `docs/rfc` in CI — the `docs/rfc` step lacks the `--no-tmp-check` the
`docs/design` step carries). A new `## 1.`-style RFC is enforcer-clean by
construction (verified: `rfc_enforcer --check ... --ignore-missing-section1`
passes). It cites RFC-0247 (lineage), RFC-0228 (recall@depth harness gate), and
RFC-0259 §6 (routing).

## Design (summary)

A typed hysteresis-bounded retention for the episode log: a **line-tail cap** on
`events.jsonl` + a **file-count cap** on `episodes/`, mirroring the facts cap
band (keep 256 / trigger 384) and **exceeding the recall scan window**
(`episode_tail_scan = 32`) so recall never starves. Raw-line trim (byte
fidelity, malformed-line tolerant), best-effort unlink, wired in the librarian
write path under `with_episode_bundle_lock`. Implementation lands in a separate
PR citing this RFC.

## Not a workaround

The hysteresis rationale + recall@depth harness gate + named recall-window
coupling are what distinguish this retention bound from a bare `cap-as-fix`. §5
rejects bare-cap and telemetry-only alternatives explicitly.

Tracker: Issue #21789 (defect D).


jeong-sik:
Reviewed current head `2378da7135`.

This is RFC-only plus `.next-number`. The base-main claim matches current `origin/main` (`357d89114b`), and the RFC keeps the boundary clean: event/episode retention is treated as a bounded append problem, separate from facts `valid_until`, facts-only GC, and recall behavior. The proposed keep/trigger coupling also explicitly stays above `episode_tail_scan`, which is the key implementation invariant to preserve in the follow-up PR.

Local: `git diff --check origin/main...origin/pr/21897` passed. I did not find a blocker.

PR #21895, opened 2026-06-21
fix(keeper): cap honors valid_until on the retention path (RFC-0259 P5 / defect C)

jeong-sik:
## What

Defect **C** from Issue #21789 (Memory OS adversarial audit): the hot-path fact
cap honors the typed `valid_until` boundary, so expired volatile rows stop
accumulating on disk. Implements RFC-0259 **§3.6 (P5)** — already merged as RFC
text (#21831). No prompt-behavior change.

## Why

`retention_rank` is TTL-blind: it takes `~now` only to pick a durable-vs-volatile
tier, never tests whether *this* row's `valid_until` has passed. The sole disk
path that drops an expired row is `run_gc`, which is off by default. So a store
below the 384-row cap threshold keeps expired rows on disk indefinitely (audit:
a keeper with 140/140 expired-`valid_until` rows retained all 348).

Per RFC-0259 §3.6 this is **retention-path determinism, not a new janitor** — the
cap and the GC must agree on what `valid_until` means. Recall already filters
expired rows at read time (`fact_is_current`), so disk hygiene here does not
change what a keeper sees.

## Change

- `Keeper_memory_os_types.partition_expired ~now` — splits a fact list into
  `(live, expired)` on the same `fact_is_current` boundary the GC sweep and
  recall already use. Durable facts (`valid_until = None`) are always live.
- `cap_facts` / `merge_and_cap_facts` drop the expired partition **before** the
  trigger gate and **before** ranking, and count it in `dropped`. `~now` is
  threaded in (the librarian write path already has it).
- This fixes the under-cap leak: an expired row is now evicted on the write path
  even when the store is far below trigger.

## Scope boundary

The `MASC_KEEPER_MEMORY_OS_GC` **default flip** (§3.6 item 1) needs a live fleet
dry-run before flipping and is left for that operator gate — it is the idle-store
backstop. This PR closes the always-on write-path leak.

## Tests

- `partition_expired` split: durable + fresh stay live, expired partitioned, order preserved.
- `cap_facts` drops one expired row below trigger, keeps durable + fresh, idempotent re-run is a no-op.
- `merge_and_cap_facts` (production librarian path) drops expired with **no incoming claims** and counts it in `dropped`; only durable remains.

Property held throughout: a durable fact (`valid_until = None`) is never evicted by the expiry path.

RFC: RFC-0259 §3.6 (P5). Tracker: Issue #21789 (defect C). Not a workaround — the
cap already exists; this makes it honor the typed boundary it currently ignores.

[...]

PR #21899, opened 2026-06-21
feat(keeper): bound the episode log (events.jsonl / episodes) (RFC-0272 / defect D)

jeong-sik:
## What

Defect **D** from Issue #21789: the append-only episode log (`events.jsonl` +
`episodes/`) gets a typed, hysteresis-bounded retention. Implements **RFC-0272**
(PR #21897). Disk-hygiene only — recall is already tail-bounded, so no
prompt-behavior change.

## Why

Only facts are capped (`fact_store_max`). `append_event` / `append_episode` have
no bound, so `events.jsonl` grows ~4x the fact store and `read_episode_files_tail`
re-reads every episode file each call (O(files), growing). Recall reads a fixed
32-row window (`episode_tail_scan`).

## Change

Mirrors the facts-cap band (keep 256 / trigger 384, hysteresis 128). Low-water
256 = **8x the recall scan window (32)** so a trim never starves recall.

- `cap_events` — line-tail cap on `events.jsonl`: keep the last `keep` **raw**
  lines (byte-faithful, malformed-line tolerant; not parse/filter/reserialize).
- `cap_episode_files` — file-count cap on `episodes/`: keep the `keep`
  most-recent by `compare_episode_recency`, best-effort unlink the rest. Only
  parseable files are counted/ranked; an unparseable file is left untouched, not
  blind-deleted.
- `trim_target` — the pure hysteresis decision (testable without IO).
- Wired after `append_episode` / `append_event` inside the existing
  `with_episode_bundle_lock`, so the cap is serialized with the writes.

## Not a workaround

This is a retention bound (RFC-0272), the same class as the merged facts cap —
not a bare cap-as-fix: the hysteresis rationale, the recall@depth harness gate
(RFC-0228), and the named recall-window coupling are explicit. The returned
drop count is diagnostic; the rewrite/unlink is the mechanism (not telemetry-as-fix).

## Tests

- `trim_target` hysteresis + a coupling guard asserting low-water > recall window (32).
- `cap_events` drops the three oldest over trigger, keeps the newest in order, idempotent re-run is a no-op.
- `cap_episode_files` unlinks the three oldest, keeps recent, idempotent.

RFC: RFC-0272 (PR #21897). Tracker: Issue #21789 (defect D). Builds-on lineage: RFC-0247 / RFC-0228.

### DIFF
--- a/lib/keeper/keeper_librarian_runtime.ml
+++ b/lib/keeper/keeper_librarian_runtime.ml
@@ -572,6 +572,22 @@ let extract_and_append_with_provider
       | Ok () ->
         Keeper_memory_os_io.append_episode ~keeper_id episode;
         Keeper_memory_os_io.append_event ~keeper_id episode;
+        (* RFC-0272 (defect D): bound the append-only episode log under the same
+           bundle lock that serialized the writes above, so a re-extraction cannot
+           grow events.jsonl / episodes/ without limit. Hysteresis-gated: the trim
+           is a no-op until the high-water, so this is off the per-turn hot path. *)
+        ignore
+          (Keeper_memory_os_io.cap_events
+             ~keeper_id
+             ~keep:Keeper_memory_os_io.event_recall_window
+             ~trigger:Keeper_memory_os_io.event_store_max
+            : int);
+        ignore
+          (Keeper_memory_os_io.cap_episode_files
+             ~keeper_id
+             ~keep:Keeper_memory_os_io.episode_file_window
+             ~trigger:Keeper_memory_os_io.episode_file_store_max
+            : int);
         (* RFC-0251: the co-occurrence edge / spreading-activation organ was removed
            (dark-by-default, no recall consumer), so the fact upsert above is the only
            post-merge work. *)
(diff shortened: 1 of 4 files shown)


jeong-sik:
Adversarial review at `7f0de3c5c1`: no blocking finding found.

Checked the retention paths:
- librarian write path appends episode + event under the existing episode bundle lock, then runs the event/file caps under the same serialization point.
- `cap_events` trims raw tail lines, so malformed historical rows are not parse-filtered away during rewrite.
- `cap_episode_files` keeps the most recent parseable files using the same recency ordering used by recall fallback.
- hysteresis constants mirror fact retention and the tests guard the low-water against the episode recall scan window.

Local non-build evidence:
- `git diff --check origin/main...refs/remotes/pr/21899`

No local build/test run per operator direction; CI remains build/test authority. At review time CI had non-build checks green and Build/TLA still pending.

jeong-sik:
Conflict follow-up: rebased #21899 onto current `origin/main` after #21895 landed.

Resolution detail:
- kept the valid_until fact-retention tests from main (#21895)
- kept the RFC-0272 episode-log retention tests from this PR
- verified the auto-merged production path still passes `~now` into `merge_and_cap_facts` before running `cap_events` / `cap_episode_files` under the episode bundle lock

New head: `1d840c0894`.

Local non-build evidence:

No local build/test run per operator direction; CI remains build/test authority.

PR #21902, opened 2026-06-21
fix(keeper): replay in-flight event queue stimuli after crash

jeong-sik:
## Summary
- add an in-flight event queue snapshot so drained keeper stimuli survive restart until ack
- merge pending + in-flight queue snapshots during keeper registration/load for at-least-once replay
- requeue consumed stimuli on keepalive cycle crash and ack the in-flight lease after successful cycle completion
- add regression coverage for restart replay before ack and duplicate-safe requeue

## Context
The adversarial gap review said keeper runtime event queues could lose operator/mention/fusion wake signals across restart. Current main already has durable pending snapshots, but dequeue/drain removed stimuli from the durable snapshot before the turn completed. A hard restart in that window could still drop the stimulus. This PR makes the drained window explicit with event-queue-inflight.json.

## Validation
- opam exec -- ocamlformat --check changed OCaml files
- opam exec -- ocamlc -stop-after parsing changed OCaml files

## Not Run
- scripts/dune-local.sh build test/test_keeper_event_queue.exe: refused with code 75 because machine-wide bare Dune processes are active. CI remains the build authority for this PR.

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -186,6 +186,7 @@ let run_keepalive_unified_turn
   then { meta = meta_after_triage; cycle_crashed = false }
   else (
     let consumed_stimuli = ref [] in
+    let consumed_stimuli_turn_completed = ref false in
     try
       let event_intake =
         heartbeat_event_intake ~ctx ~meta_after_triage ~pending_board_events
@@ -393,22 +394,36 @@ let run_keepalive_unified_turn
           meta_after_triage)
         else if should_run_turn
         then (
-          run_keeper_cycle
-            ~ctx
-            ~meta_after_triage
-            ~stop
-            ~obs
-            ~turn_decision
-            ~shared_context
-            ())
+          let meta_after_cycle =
+            run_keeper_cycle
+              ~ctx
+              ~meta_after_triage
+              ~stop
+              ~obs
+              ~turn_decision
+              ~shared_context
+              ()
+          in
+          consumed_stimuli_turn_completed := true;
+          meta_after_cycle)
         else meta_after_triage
       in
+      (* Event intake dequeues before the admission/pressure gates above.
+         Only ack after [run_keeper_cycle] actually returns; otherwise put the
+         lease back so a non-exception skip does not drop the stimulus. *)
       if !consumed_stimuli <> []
       then
-        Keeper_registry_event_queue.ack_consumed
-          ~base_path:ctx.config.base_path
-          meta_after_triage.name
-          !consumed_stimuli;
+        if !consumed_stimuli_turn_completed
+        then
+          Keeper_registry_event_queue.ack_consumed
+            ~base_path:ctx.config.base_path
+            meta_after_triage.name
+            !consumed_stimuli
+        else
+          Keeper_registry_event_queue.requeue_front
+            ~base_path:ctx.config.base_path
+            meta_after_triage.name
+            !consumed_stimuli;
       { meta = meta_after_cycle; cycle_crashed = false }
     with
     | Eio.Cancel.Cancelled _ as e -> raise e
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -52,6 +52,7 @@ let record_recovery_stimulus_turn_started =
 type heartbeat_event_intake = Stimulus_intake.heartbeat_event_intake = {
   pending_board_events : Keeper_world_observation.pending_board_event list;
   consumed_stimulus_count : int;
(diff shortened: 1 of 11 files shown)


jeong-sik:
Adversarial review found and fixed one real race on head `a3acad38ce`.

Finding:
- `dequeue` / `drain_board` recorded stimuli into `event-queue-inflight.json` before the live queue CAS succeeded. A losing CAS attempt could therefore leave a false in-flight lease for stimuli this caller did not actually remove. In a crash/restart window that can replay an already-consumed stimulus.

Fix pushed:
- `1b10a2c739 fix(keeper): record event queue inflight after CAS`
- The order is now: live queue CAS succeeds -> record in-flight lease -> persist reduced pending snapshot. That preserves the crash safety window without letting failed CAS attempts create replay leases.

- `gh pr view 21902 --json headRefOid,mergeable,mergeStateStatus,statusCheckRollup`

CI is the build/test authority; after the push the PR is `MERGEABLE`, with new checks running on `1b10a2c739`.

jeong-sik:
Adversarial review found a real at-least-once gap and I pushed a corrective follow-up (`2ace2666ec`).

Finding fixed:
- `dequeue` / `drain_board` were removing stimuli from the live CAS queue before durable inflight recording. A process crash in that window could lose the stimulus permanently, which defeats this PR's crash-replay purpose.
- `ack_consumed` / crash `requeue_front` cleared the whole inflight file. That can erase unrelated in-flight rows if another lease/stale inflight row exists.

Follow-up changes:
- Record inflight before the CAS removal, so a crash before/after CAS reloads via pending+inflight dedup instead of losing the stimulus.
- Ack only the consumed/requeued stimuli, not the entire inflight lease file.
- Added a partial-ack regression so one acknowledged stimulus does not clear another in-flight stimulus.

Local verification:
- `opam exec -- ocamlformat --check lib/keeper/keeper_heartbeat_loop.ml lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_event_queue.mli lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli test/test_keeper_event_queue.ml`
- `opam exec -- ocamlc -stop-after parsing ...same files...`
- `git diff --check`

Not run locally:
- Focused Dune target. `scripts/dune-local.sh build test/test_keeper_event_queue.exe` refused because machine-wide bare `dune build` processes are already running, so full compile/test proof is left to CI as requested.

jeong-sik:
Latest-head review (`87c104f525`): no new code-level blocker found.

What changed after the CI failure:
- CI failed on `Keeper_event_queue.of_list` because the constructor is private/not exported.
- `87c104f525 fix(keeper): avoid private queue constructor` replaces that with a local `queue_of_list` using the public `empty` + `enqueue` API. That preserves FIFO order through `to_list`/`dequeue` and avoids widening the queue module interface just for persistence cleanup.

Ordering review:
- The current `record_inflight -> CAS remove -> persist pending` order is defensible for this PR's at-least-once contract: if the cycle/fiber dies in the remove/persist window, the stimulus remains replayable through pending+inflight dedup.
- A false in-flight lease would require two concurrent consumers draining the same keeper queue and one recording after the other already acked. I do not see such a multi-consumer path in the current call graph: `heartbeat_event_intake` is the only runtime drain/dequeue caller, and it chooses one board batch or one fallback stimulus per keeper cycle.
- Partial `ack_inflight` now keeps unrelated leases intact, fixing the earlier whole-file ack hazard.

Non-build evidence only, per instruction:
- `git diff --check origin/main...origin/codex/keeper-event-queue-durable-cas-20260621`
- GitHub CI restarted on `87c104f525`; build/test authority remains CI.

jeong-sik:
Rebased onto latest `origin/main` after the env-knob catalog fix landed upstream. The failed Build/Test run was from the stale merge base and hit the catalog drift gate. Verification after rebase: `git diff --check origin/main...HEAD` clean. New head: 17bd89fa6c.

jeong-sik:
Adversarial latest-head follow-up: found and fixed a real non-exception drop path in 18d986fe3c.

Finding:
- heartbeat_event_intake dequeues/drains stimuli before the later admission and pressure gates.
- The previous latest head acknowledged consumed_stimuli at the end of the successful function return path, even when no keeper turn actually ran because stop, fd pressure, disk pressure, or admission guards returned meta_after_triage.
- That preserves crash replay, but drops a stimulus on a non-exception skip, which violates the at-least-once intent of this PR.

Fix:
- Track whether run_keeper_cycle actually returned.
- Ack consumed stimuli only after that point.
- If stimuli were consumed but the function returns through a non-turn path, requeue_front the lease instead of acking it.

Local validation scope: git diff --check only. No local build/test per operator instruction; CI remains build/test authority.

PR #21905, opened 2026-06-21
fix(dashboard): gate missing tool output on hydration coverage

jeong-sik:
## Summary
- Prevent settled tool calls from being marked missing before the separate tool-output hydration surface has covered that tool row.
- Track keeper tool-output hydration coverage by successful fetch start time instead of a sticky ready boolean.
- Add regressions for pre-hydration and older-hydration races.

## Verification
- vitest run --config vitest.config.ts src/components/chat/primitives.test.ts src/tool-call-output-store.test.ts --no-file-parallelism --maxWorkers=1
- tsc --noEmit --pretty false
- eslint src/components/chat/primitives.ts src/components/chat/primitives.test.ts src/components/keeper-shared.ts src/keeper-actions.ts src/tool-call-output-store.ts

Follow-up to merged #21892.

### DIFF
--- a/dashboard/src/components/chat/primitives.test.ts
+++ b/dashboard/src/components/chat/primitives.test.ts
@@ -1852,6 +1852,7 @@ describe('ChatTranscript — tool-call grouping (작업 과정)', () => {
         ]}
         emptyText="empty"
         groupToolCalls=${true}
+        toolOutputsCoveredThroughMs=${Date.parse('2026-03-24T00:00:01.000Z')}
       />`,
       container,
     )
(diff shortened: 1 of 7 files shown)


jeong-sik:
Adversarial review pass: no blocking finding found on head `5022724517`.

Checked the missing-output state machine:
- A settled assistant turn no longer marks unjoined tool outputs as missing until the separate tool-output hydration surface has successfully covered the tool call timestamp.
- Hydration failure is conservative: it does not advance coverage, so missing output is not fabricated from a failed fetch.
- The fetch limit matches the visible keeper history tail (`KEEPER_HISTORY_TAIL_MESSAGES`), so the current visible-window assumption is coherent for the workspace transcript.
- Direct/non-keeper ChatTranscript callers keep the default `null` coverage and therefore stay pending rather than inventing missing output.

Residual tradeoff, not a blocker: entries without a parseable timestamp remain pending because coverage cannot be proven. That is preferable to a false missing warning.

- `git diff --check origin/main...origin/fix/dashboard-tool-status-missing`
- `gh pr view 21905 --json headRefOid,mergeable,mergeStateStatus,statusCheckRollup`

CI owns build/test; live checks were green/complete when reviewed.

jeong-sik:
Adversarial follow-up pushed: tool-output hydration coverage is now bounded by the fetched tail.\n\nThe original change treated a successful /tool-calls?limit=200 hydration as covering every older transcript tool row up to fetch-start time. That can false-mark old visible chat tool rows as missing because chat history and tool-call output history have different windows. The patch now carries coveredSinceMs + coveredThroughMs, only marks rows missing inside the covered window, and keeps older-than-tail rows pending. Added regression coverage for the bounded-tail case and store coverage merge behavior.\n\nPer operator instruction, I did not run local build/tests; CI is the build/test authority for this push.

jeong-sik:
Adversarial follow-up pushed on latest head e677227d76. I found one real false-missing path in the previous coverage bound: the backend tool-calls endpoint filters a global recent tail by keeper, so a response shorter than the requested limit does not prove older keeper tool outputs were covered. The dashboard now treats hydration as covered only from the oldest returned output timestamp; empty/no-timestamp fetches prove no missing range and keep rows pending instead of falsely marking missing.\n\nVerification:\n- vitest: src/components/chat/primitives.test.ts src/tool-call-output-store.test.ts src/keeper-actions.test.ts => 155 passed\n- tsc --noEmit --pretty false => pass\n- eslint on touched dashboard files => pass (one existing ignored-file warning for tool-call-output-store.test.ts)\n- git diff --check origin/main...HEAD => pass

jeong-sik:
Follow-up for the Build/Test failure: the failed job was the env knob catalog drift gate, not the dashboard hydration logic. Cherry-picked the same `fix(env-knob): catalog typed getter aliases` generator fix so the merge CI no longer misclassifies env-key alias typed getters. Verification after cherry-pick: `MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec ./bin/env_knob_catalog.exe -- docs/runtime-tunables.md` clean; `git diff --check origin/fix/dashboard-tool-status-missing...HEAD` clean. New head: 0354c7288a.

jeong-sik:
CI follow-up: the failed Build and Test job was the same Env knob catalog drift gate (MASC_DISABLE_HITL alias classification). Branch head is now 0354c7288a with the generic env-key alias catalog fix and regenerated runtime tunables.

Push note: my push raced with the remote update, but refs/heads/fix/dashboard-tool-status-missing now points to the same 0354c7288a commit, so no force push was needed. Local validation scope for this follow-up: git diff --check HEAD~1..HEAD only; no local build/test per operator instruction. CI remains build/test authority.

jeong-sik:
Rebased onto latest `origin/main` after the catalog fix landed upstream. The temporary cherry-pick is now upstream, so the PR head contains only the dashboard commits on top of current main. Verification after rebase: generator clean during conflict resolution; `git diff --check origin/main...HEAD` clean. New head: 3df2d8b755.

PR #21910, opened 2026-06-21
fix(keeper): gate checkpoint multimodal input

jeong-sik:
## Summary

Fix the keeper multimodal capability gate for resumed OAS checkpoints.

The dashboard can send a vision turn, then the next text-only follow-up resumes an OAS checkpoint whose `messages` still contain the prior image block. Before this change, RFC-0265 reroute/floor validation only looked at `initial_messages + goal_blocks`, so checkpoint-only media could bypass the pre-dispatch gate and leak to a text-only provider as a 400 like:

`messages.content.type is invalid, allowed values: ['text']`

This PR makes checkpoint resume messages part of the active input view for:

- modality reroute selection in `Keeper_turn_driver.run_named`
- runtime multimodal floor validation in `Runtime_agent.run_blocks`
- regression coverage for checkpoint-only image history

## Validation

- `opam exec -- ocamlformat --check lib/runtime/runtime_agent.ml lib/runtime/runtime_agent.mli lib/keeper/keeper_turn_driver.ml test/test_runtime_modality_reroute.ml`
- `gtimeout 240 env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build lib/runtime/.masc_runtime.objs/byte/runtime_agent.cmo`

Attempted but did not complete locally:

- `gtimeout 420 env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_runtime_modality_reroute.exe` timed out with no compiler diagnostic after entering Dune.
- `gtimeout 300 env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build lib/masc.cma` timed out with no compiler diagnostic after entering Dune.

### DIFF
--- a/bin/env_knob_catalog.ml
+++ b/bin/env_knob_catalog.ml
@@ -19,6 +19,10 @@
    tool emits an "Unrecognized" entry and CI will surface it. *)
 
 let env_var_re = Str.regexp "\"\\(MASC_[A-Z][A-Z0-9_]*\\)\""
+let env_key_alias_re =
+  Str.regexp "let[ \t]+\\([A-Za-z_][A-Za-z0-9_']*\\)[ \t]*=[ \t]*\"\\(MASC_[A-Z][A-Z0-9_]*\\)\""
+
+let ident_re = Str.regexp "[A-Za-z_][A-Za-z0-9_']*"
 
 let typed_get_re =
   Str.regexp "get_\\(int\\|int_nonneg\\|float\\|float_nonneg\\|float_in_range\\|ratio\\|string\\|bool\\)"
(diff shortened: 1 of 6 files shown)


jeong-sik:
Adversarial pass on latest head fa681f59ba:\n\nNo blocking code issue found in this slice. The change threads checkpoint messages into both the proactive reroute decision and the final Runtime_agent pre-provider capability gate, while leaving OAS checkpoint ownership intact. That closes the specific gap where an image/audio/document only present in a resume checkpoint could bypass MASC reroute and fail later as a provider-side 400.\n\nBoundary notes checked:\n- no provider/model hardcoding added\n- checkpoint is read as Agent_sdk.Checkpoint.t messages; MASC does not reinterpret OAS lifecycle state\n- final floor still lives in Runtime_agent.run_blocks before provider dispatch\n- tests cover checkpoint-only media for both reroute and loud validation failure\n\nLocal validation scope: review only; no local build/test per operator instruction. CI remains build/test authority.

jeong-sik:
Fixed the Build/Test drift gate by pushing the existing `fix(env-knob): catalog typed getter aliases` commit. The failure was not a runtime behavior issue in this PR; the catalog generator missed typed getter calls that reference env-key alias constants, so CI regenerated `docs/runtime-tunables.md` differently. Verification: local generator clean with `MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh exec ./bin/env_knob_catalog.exe -- docs/runtime-tunables.md`; `git diff --check origin/codex/multimodal-content-guard-20260621...HEAD` clean before push. New head: 03eee0077f.

PR #21913, opened 2026-06-21
fix(dashboard): interleave think and tool steps by occurrence time in work trace

jeong-sik:
## 목표
Keeper 턴 상세(작업 과정 카드)에서 모델의 실제 작업 순서(`think → tool → think → tool`)가 두 덩어리(생각 전체 → 도구 전체)로만 보여, 어느 생각 다음에 어떤 도구가 실행됐는지 운영자가 볼 수 없는 문제를 해결한다.

## 접근: 렌더 타임 interleave (entry 구조 보존)
`ToolTraceCard`에서 think/reason step과 tool entry를 **발생 시각(timestamp)으로 정렬 병합**해 단일 순회로 렌더한다. tool entry는 entry 기반 output join(`lookupToolCallOutput`)을 그대로 유지하므로 #21892의 missing/집계와 `ToolTraceStep` status taxonomy(pending/missing/ok/bad, `semantic_success`)가 변경되지 않는다.

## 설계 문서 옵션 C 기각 근거 (적대 검증)
설계 문서(`reports/masc-keeper-turn-interleaving-design-20260621.html`)의 옵션 C(tool을 `assistant.traceSteps`의 `kind:'tool'` step으로 이동)는 적대 검증(26 agent / 5 lens)에서 기각됐다.
- **B1/P1**: `ChatTraceToolStep`에 `tool_use_id`가 없어 output-store join이 단절(silent 회귀).
- **A2/P2**: `status:'ok'|'err'` 2-state는 `semantic_success`(blocked/timeout)를 담지 못해 transport-성공/semantic-실패가 ok로 위장.
- **E3/P2**: #21892의 `missingN`/`failN` 집계가 entry output 기반이라 step 기반에서 유지 불가.

3/5 lens가 렌더 타임 interleave를, 2/5가 option C를 structural change(step에 join 복구 = ToolCallEntry 복제)와 함께만 채택 가능으로 판정. 순수 option C 지지 0건.

## 변경
- `types/core.ts`: think/reason step에 `ts?: string` 추가.
- `keeper-state.ts`: `appendAssistantThinkingDelta`에서 새 think step에 `ts` 부여(delta 병합 시 첫 ts 유지); `normalizeTraceStep`에서 백엔드 `ts` 추출(존재 시).
- `components/chat/primitives.ts`: `interleaveTraceAndTools` merge 함수(export); `ToolTraceCard` 단일 순회. 집계(failN/missingN/totalMs/stepN)는 entry 기반 유지.
- 테스트: `primitives-interleave.test.ts`(인터리빙 순서 5 case); 기존 strict `toEqual` 3곳에 `ts: expect.any(String)`.

## 검증 상태
- 로컬 빌드/테스트 미실행(사용자 지시: 빌드는 CI만). CI `Build and Test` + `Test Coverage Check`로 검증한다.
- self-review(best-programmer): think-only / tool-only / flat(`groupToolCalls=false`) 경로 무결. `groupToolCalls` 기본값은 false지만 `keeper-shared.ts`가 `groupToolCalls=${true}`로 명시 전달(3곳)하므로 턴 상세에 적용된다. 정렬은 ISO-8601 사전순=시간순, ts 부재 시 stable fallback으로 legacy 2-섹션 순서 보존.

## 비고
- RFC/workaround gate 대상 아니다(chat 렌더링, 워크어라운드 시그니처 없음).
- #21748(OPEN, 같은 `keeper-state.ts` 경로)과 충돌 영역 최소: traceSteps는 think/reason 전용 유지, 추가된 `ts`는 `sameConversationEntry`(role+text 매칭)에 영향을 주지 않는다.

### DIFF
--- a/dashboard/src/components/chat/primitives-interleave.test.ts
+++ b/dashboard/src/components/chat/primitives-interleave.test.ts
@@ -19,7 +19,11 @@ describe('interleaveTraceAndTools', () => {
     output: null,
   })
   const labels = (items: ReturnType<typeof interleaveTraceAndTools>) =>
-    items.map((i) => (i.kind === 'trace' ? `think:${i.step.text}` : `tool:${i.entry.id}`))
+    items.map((i) => {
+      if (i.kind === 'tool') return `tool:${i.entry.id}`
+      // think/reason carry `text`; the tool variant of ChatTraceStep does not.
+      return `think:${'text' in i.step ? i.step.text : i.step.name}`
+    })
 
   it('interleaves think and tool by occurrence time (think -> tool -> think -> tool)', () => {
     const out = interleaveTraceAndTools(
--- a/dashboard/src/components/chat/primitives-interleave.test.ts
+++ b/dashboard/src/components/chat/primitives-interleave.test.ts
@@ -1,0 +1,74 @@
+// @vitest-environment jsdom
+
+// Unit tests for the render-time think<->tool interleaving in ToolTraceCard.
+// Pure-function level so the ordering contract is asserted without the jsdom
+// render harness: the merge must preserve occurrence order (think -> tool ->
+// think -> tool) and fall back to the legacy two-section order when a step has
+// no timestamp.
+import { describe, expect, it } from 'vitest'
+import type { ChatTraceStep, KeeperConversationEntry } from '../../types'
+import { interleaveTraceAndTools } from './primitives'
+
+describe('interleaveTraceAndTools', () => {
+  const think = (text: string, ts?: string): ChatTraceStep =>
+    ts === undefined ? { kind: 'think', text } : { kind: 'think', text, ts }
+  // Tool entries only need `id` and `timestamp` for ordering; interleave reads
+  // nothing else off them.
+  const tool = (id: string, ts: string) => ({
+    entry: { id, role: 'tool', timestamp: ts } as unknown as KeeperConversationEntry,
+    output: null,
+  })
+  const labels = (items: ReturnType<typeof interleaveTraceAndTools>) =>
+    items.map((i) => (i.kind === 'trace' ? `think:${i.step.text}` : `tool:${i.entry.id}`))
+
+  it('interleaves think and tool by occurrence time (think -> tool -> think -> tool)', () => {
+    const out = interleaveTraceAndTools(
+      [
+        think('A', '2026-06-21T00:00:01.000Z'),
+        think('B', '2026-06-21T00:00:03.000Z'),
+      ],
+      [
+        tool('X', '2026-06-21T00:00:02.000Z'),
+        tool('Y', '2026-06-21T00:00:04.000Z'),
+      ],
+    )
+    expect(labels(out)).toEqual(['think:A', 'tool:X', 'think:B', 'tool:Y'])
+  })
+
+  it('keeps the legacy two-section order when think steps carry no timestamp', () => {
+    // Absent ts (backend-normalized steps) sorts as '' and precedes any tool,
+    // so thinks stay grouped above tools — matching the pre-change render.
+    const out = interleaveTraceAndTools(
+      [think('A'), think('B')],
+      [tool('X', '2026-06-21T00:00:02.000Z')],
+    )
+    expect(labels(out)).toEqual(['think:A', 'think:B', 'tool:X'])
+  })
+
+  it('places an earlier tool before a later think', () => {
+    const out = interleaveTraceAndTools(
+      [think('A', '2026-06-21T00:00:05.000Z')],
+      [tool('X', '2026-06-21T00:00:01.000Z')],
+    )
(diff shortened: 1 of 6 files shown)


jeong-sik:
Adversarial follow-up on latest head: found and fixed one render invalidation gap.\n\nIssue: ToolTraceCard ordering now depends on trace step ts, but traceStepsSignature still ignored ts. That signature drives the transcript auto-scroll/layout effect dependencies, so a timestamp-only trace update/reorder could miss the scroll/layout correction even though the render order changed.\n\nFix pushed: a6f730adca fix(dashboard): include trace timestamps in chat signature\n- think signature includes ts\n- reason signature includes ts\n\nLocal validation only: git diff --check. No local typecheck/build/test per operator instruction; CI is the build/test authority.

jeong-sik:
Resolved the conflict by rebasing onto latest `origin/main` and preserving both sides: current missing-output hydration gating still uses `canMarkMissingForEntry`, while the work-trace render now uses `interleaveTraceAndTools` for occurrence ordering. Verification: related Vitest suite passed (`primitives-interleave`, `primitives`, `keeper-state`, `keeper-stream`: 156 tests); `tsc --noEmit --pretty false` passed; `git diff --check origin/main...HEAD` clean. New head: 720652211e.

jeong-sik:
Latest-head pass on 720652211e001e4dd9fdef75ffa30d26fe6d1a02: no code blocker found in the conflict-resolution slice.

Checked:
- The conflict resolution preserved the missing-output hydration gate: canMarkMissingForEntry still depends on turnComplete plus output coverage, and fail/missing/duration aggregates still use the tool entry/output list rather than the new render ordering.
- interleaveTraceAndTools is render-only. It does not alter tool-output lookup keys or the stored entry shape, and it falls back to legacy trace-then-tool order when trace steps have no timestamp.
- Live thinking deltas stamp only newly-created think steps and preserve the first timestamp while consecutive deltas merge, so a single thinking span sorts by when it began rather than by the last fragment.
- History normalization now preserves optional trace-step ts from the wire and leaves older rows valid when ts is absent.

Readiness remains CI-bound: GitHub still showed Lint and Dashboard in progress when I checked this head. This is not a Ready signal while the PR is still draft / checks pending.

jeong-sik:
**Adversarial review — PR head `720652211e` (fresh-context pass)**

Verified the implementation matches the render-time-merge design and is solid:

- **No data-structure change.** `ChatTraceThinkStep`/`ChatTraceReasonStep` gain `ts?: string` (`types/core.ts`); tool entries keep the entry-based `lookupToolCallOutput` join and render via `ToolTraceStep`. The aggregates below (`failN`/`missingN`/`totalMs`/`stepN`) stay entry-based, so #21892 missing/aggregate detection and the `pending/missing/ok/bad` taxonomy are preserved. Good.
- **Parse-don't-validate on the wire.** `normalizeTraceStep` wraps `ts: asString(raw.ts)` in `withoutUndefined`, so the field is present-or-absent, never explicitly `undefined`. Delta merge preserves the first stamp (`ts: last.ts`) — "the step began at that time, not when the latest fragment arrived." Correct.
- **Stable, documented sort.** `interleaveTraceAndTools` concats trace-then-tool (stable baseline) and sorts by ISO; absent `ts` → `''` → sorts first, preserving the legacy two-section order for backend-normalized steps. The 5-case suite (interleave / no-ts legacy / earlier-tool / equal-ts stable / empty) covers the merge function well.

**One non-blocking observation — the `traceOrderTs` comment overstates the correctness guarantee across clock domains.**

The comment says *"ISO-8601 strings sort lexicographically == chronologically."* That equivalence holds **within one clock domain**, but this merge combines two:

- think `ts` is **client-stamped** — `new Date().toISOString()` in `appendAssistantThinkingDelta` (`keeper-state.ts`, the `ts: new Date().toISOString()` arm), i.e. when the delta reaches the browser.
- tool `entry.timestamp` is **server-derived** — `toIsoTimestamp(raw.ts_unix) ?? toIsoTimestamp(raw.timestamp)` (`keeper-state.ts:702`), i.e. when the keeper emitted it.

So in **live** viewing the sort is over mixed client/server clocks. Network latency + skew mean a think emitted at server `S1` and received at client `C1 = S1 + latency` can sort *after* a tool whose server timestamp `S2` lies in `(S1, C1)` — a rapid think→tool with gap under one-way latency can mis-interleave. (In **history** viewing this is moot: backend-normalized think steps carry no `ts`, so they sort first and the legacy order is preserved — exactly the documented fallback.)

This is the accepted trade-off of render-time merge vs the unified-server-timeline alternative, so I'm not asking to change the approach — just to tighten the comment so a future reader doesn't assume exact chronological ordering across the two clocks. Something like *"best-effort chronological; think steps are client-stamped, tool entries server-stamped, so live ordering can cross within ~one network RTT"* would state the real guarantee.

Verdict: ship-ready; the comment clarification is the only thing I'd add.

PR #21917, opened 2026-06-21
fix(dashboard): harden frontend connection recovery

jeong-sik:
## Summary
- reject in-flight IDE LSP JSON-RPC requests when the WebSocket closes so refresh/hover promises do not hang after reconnect
- keep only one scheduled IDE LSP reconnect, cancel it on dispose, and ignore stale socket events from older connections
- bound IDE LSP reconnects with shared retry/backoff constants and stop retrying on terminal auth/policy close codes
- recover IDE LSP `send()` failures through the same reconnect path instead of throwing from dashboard UI code
- refresh the current editor LSP state after reconnect by re-sending `didOpen` and scheduling a refresh once initialization completes
- retry transient loopback dev-token bootstrap failures while still memoizing success and disabled-endpoint responses
- route the IDE keeper cursor SSE helper through the shared SSE transport so cleanup cancels pending reconnects
- coalesce duplicate SSE transport error events into a single pending reconnect timer

## Why
The dashboard has several active frontend connection paths: the main dashboard WS/MCP bootstrap path, the IDE LSP WebSocket, and smaller SSE helpers. A server restart, transient close, send failure, cleanup race, auth/policy close, or startup race could previously leave pending promises unresolved, retry forever, skip the editor refresh after reconnect, resurrect a cleaned-up cursor SSE stream, or memoize a failed dev-token bootstrap for the rest of the page load.

## Validation
- `pnpm exec vitest run --config vitest.config.ts src/transports/sse-transport.test.ts src/components/ide/keeper-cursor-overlay.test.ts src/api/dev-token.test.ts src/components/ide/ide-lsp-client.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec vitest run --config vitest.config.ts src/api/mcp.test.ts -t "dev-token bootstrap" --no-file-parallelism --maxWorkers=1`
- `pnpm exec eslint src/components/ide/ide-lsp-client.ts src/components/ide/ide-lsp-client.test.ts src/components/ide/keeper-cursor-overlay.ts src/components/ide/keeper-cursor-overlay.test.ts src/transports/sse-transport.ts src/transports/sse-transport.test.ts`
- `git diff --check`

### DIFF
--- a/dashboard/src/api/dev-token.test.ts
+++ b/dashboard/src/api/dev-token.test.ts
@@ -1,0 +1,109 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  devTokenBootstrapStatus,
+  ensureDevToken,
+  resetDevTokenBootstrap,
+  type DevTokenBootstrapStatus,
+} from './dev-token'
+
+const {
+  clearStoredToken,
+  currentDashboardActor,
+  fetchWithTimeout,
+  getStoredToken,
+  getStoredTokenMeta,
+  isRemoteAccess,
+  setStoredToken,
+} = vi.hoisted(() => ({
+  clearStoredToken: vi.fn(),
+  currentDashboardActor: vi.fn(),
+  fetchWithTimeout: vi.fn(),
+  getStoredToken: vi.fn(),
+  getStoredTokenMeta: vi.fn(),
+  isRemoteAccess: vi.fn(),
+  setStoredToken: vi.fn(),
+}))
+
+vi.mock('./core', () => ({
+  clearStoredToken,
+  currentDashboardActor,
+  fetchWithTimeout,
+  getStoredToken,
+  getStoredTokenMeta,
+  isRemoteAccess,
+  setStoredToken,
+}))
+
+function jsonResponse(body: unknown, status = 200): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  })
+}
+
+function setBootstrapStatus(value: DevTokenBootstrapStatus): void {
+  ;(devTokenBootstrapStatus as { value: DevTokenBootstrapStatus }).value = value
+}
+
+describe('ensureDevToken', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    resetDevTokenBootstrap()
+    setBootstrapStatus('idle')
+    getStoredToken.mockReturnValue(null)
+    getStoredTokenMeta.mockReturnValue(null)
+    currentDashboardActor.mockReturnValue('dashboard')
+    isRemoteAccess.mockReturnValue(false)
+  })
+
+  it('retries after a transient network bootstrap failure in the same page load', async () => {
+    fetchWithTimeout
+      .mockRejectedValueOnce(new Error('server not ready'))
+      .mockResolvedValueOnce(jsonResponse({
+        token: 'fresh-dev-token',
+        actor: 'dashboard',
+        scope: 'admin',
+      }))
+
+    await ensureDevToken()
+    expect(devTokenBootstrapStatus.value).toBe('network')
+    expect(fetchWithTimeout).toHaveBeenCalledTimes(1)
+    expect(setStoredToken).not.toHaveBeenCalled()
+
+    await ensureDevToken()
+
+    expect(fetchWithTimeout).toHaveBeenCalledTimes(2)
+    expect(setStoredToken).toHaveBeenCalledWith('fresh-dev-token', {
+      source: 'dev',
+      actor: 'dashboard',
+      scope: 'admin',
+    })
+    expect(devTokenBootstrapStatus.value).toBe('ok')
+  })
+
+  it('keeps a successful bootstrap memoized for the page load', async () => {
+    fetchWithTimeout.mockResolvedValueOnce(jsonResponse({
+      token: 'loopback-dev-token',
+      actor: 'dashboard',
+      scope: 'admin',
+    }))
+
+    await ensureDevToken()
+    await ensureDevToken()
+
+    expect(fetchWithTimeout).toHaveBeenCalledTimes(1)
+    expect(setStoredToken).toHaveBeenCalledTimes(1)
+    expect(devTokenBootstrapStatus.value).toBe('ok')
+  })
+
+  it('does not keep retrying a disabled loopback dev-token endpoint', async () => {
+    fetchWithTimeout.mockResolvedValueOnce(new Response('not found', { status: 404 }))
+
+    await ensureDevToken()
+    await ensureDevToken()
+
+    expect(fetchWithTimeout).toHaveBeenCalledTimes(1)
+    expect(setStoredToken).not.toHaveBeenCalled()
+    expect(devTokenBootstrapStatus.value).toBe('no_endpoint')
+  })
+})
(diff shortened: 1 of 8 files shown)


jeong-sik:
Adversarial follow-up: found and patched one real stale-socket race beyond the original reconnect settlement fix. After a newer LSP WebSocket becomes current, late message/close/error events from the old socket could still operate on shared connection state and settle the current request. Commit eb393973f9 now ignores message/close/error from non-current sockets and adds a regression test where an old socket tries to settle the current hover request. Per operator instruction, I did not run local build/tests; CI owns the build/test proof for this PR. Local check run: git diff --check only.

jeong-sik:
Draft blocker found on head eb393973f9291ac86ca3c2425f5f82721b695207.

P1 - reconnect now settles pending requests, but it does not recover the editor LSP state after reconnect. When the socket closes, LspConnection rejects every pending request at dashboard/src/components/ide/ide-lsp-client.ts lines 398-408 / 587-591. The public request helpers catch those rejections and return empty maps/null at lines 493-543. The view refresh path then treats those fulfilled empty maps as authoritative and dispatches empty code lenses, inlay hints, and diagnostics at lines 786-801.

After the 5s reconnect succeeds, nothing schedules a replacement refresh. The only scheduleRefresh call sites I see are constructor setup and file-path change, lines 760 and 772. There is no on-open/on-initialized callback from LspConnection back to the view plugin, so after a transient close during refresh the IDE can stay blank until a file-path change or another unrelated refresh path happens. That means the PR changes the failure mode from hanging stale requests to settled-but-empty LSP UI, without completing reconnect recovery.

Expected fix shape: expose an onReady/onReconnect callback from LspConnection after initialize succeeds, and have the view plugin re-open the current document if needed and scheduleRefresh for the current file. Add a regression test that closes during a refresh, advances the reconnect timer, opens the new socket, completes initialize, and asserts codeLens/diagnostic/inlay requests are reissued rather than leaving the dispatched state empty.

The stale-socket guard itself looks right: message/error/close handlers ignore non-current sockets, and dispose clears the reconnect timer. This blocker is specifically about post-reconnect state restoration.

jeong-sik:
Change-request follow-up pushed: 77e382b0c6 fix(dashboard): refresh LSP state after reconnect.

The earlier blocker was valid: settling pending requests on close could leave the CodeMirror LSP state empty after reconnect because no refresh was scheduled after initialize. The new commit adds an LspConnection ready callback after initialize plus initialized notification succeeds. The IDE view plugin uses that callback to re-open the current document and schedule a fresh codeLens/inlay/diagnostic refresh after initial connect and reconnect.

Regression coverage added at the connection boundary: ready is emitted after initial initialize and again after reconnect initialize. This pairs with the view-plugin callback that performs the actual didOpen plus refresh scheduling.

Local validation scope: git diff --check only. No local build/tests per operator instruction; CI remains the build/test authority.

jeong-sik:
Adversarial delta pass on moved head 396889dce2.\n\nNo new blocking issue found.\n\nChecked the 77e382b -> 396889d delta: cursor overlay now uses the shared SSE transport, cleanup unsubscribes and disconnects so the reconnect timer is cleared, and the added transport guard coalesces duplicate error events into one pending reconnect. Server-side `/api/v1/ide/cursors/stream` writes JSON cursor snapshots on initial connect and every 30s, so the normalized cursor payload path remains compatible. The added cursor cleanup test and SSE duplicate-error test cover the new failure mode. `git diff --check origin/main...396889dce2` is clean.\n\nCI note: no local build/tests run per operator instruction; CI owns build verification. Checks are still pending on this moved head.

jeong-sik:
**Adversarial review — PR head `396889dce2` (fresh-context pass)**

The reconnect settlement is correct and well-executed. Verified the core invariants:

- **Stale-socket race properly closed.** Every callback (`onopen`/`onmessage`/`onclose`/`onerror`) guards `this.ws !== ws`, so once a reconnect creates a new socket, the old socket's callbacks no-op — no spurious reconnect from a late `onclose`, no messages delivered into the wrong connection's state.
- **No reconnect storm.** `scheduleReconnect` clears any pending timer before setting a new one (`clearReconnectTimer`), and `dispose` clears it too. The old untracked `setTimeout` that could stack on repeated `onclose` is gone.
- **Send failures handled, not thrown.** `send`/`sendNotification` wrap `ws.send()` in try/catch → `handleSocketSendFailure` nulls the socket, rejects pending, closes the bad socket, and schedules reconnect. A send on a closing socket used to throw synchronously and bubble unhandled.
- **Reconnect re-initializes server state.** The new `onReady` callback re-runs `notifyDidOpen` for the current file and refreshes — without it, diagnostics would silently stop after a reconnect because the server holds fresh state.
- **Observability.** `lspCloseReason(event)` surfaces the close code/reason into the pending-request rejection instead of a bare "closed".

Two non-blocking observations, both about consistency with the rest of this same PR:

**1. Reconnect maturity differs from the sibling transport.**
`sse-transport.ts` (fixed in this same PR) now has exponential backoff + jitter + a `retryMaxAttempts` circuit breaker + a storm guard (`if (reconnectTimer) return`). `LspConnection` uses a fixed `5000ms` with no max-retries and no backoff. Two reconnect mechanisms in the same dashboard now sit at different maturity — under a down server the LSP client retries every 5s forever. Mirroring the SSE transport's backoff+max would bound it and keep the two consistent.

**2. No terminal-close classification (dev-token layer does have one).**
`dev-token.ts` (also this PR) adds `isRetryableDevTokenStatus` (retry only on 408/425/429/≥500; treat 401/403 as terminal so an expired/revoked dev token doesn't hammer the bootstrap). `LspConnection.onclose` schedules reconnect for *every* close code. If the server rejects the WS on auth grounds (expired dev token), the LSP layer will keep reconnecting every 5s even though the dev-token layer has correctly given up — the two layers disagree on whether the failure is terminal. Checking the close code and stopping on auth-rejection codes would mirror dev-token's classification.

Minor nit: in `onclose`, `if (this.ws !== ws) return` is followed by `if (this.ws === ws) this.ws = null` — the second guard is always true after the early return.

Verdict: ship-ready (the race fix is the point and it's correct). The two observations are consistency/boundedness enhancements, non-blocking.

jeong-sik:
Review follow-up pushed: d7d4f53c7c `fix(dashboard): bound LSP reconnect retries`.

Addressed the two non-blocking reconnect consistency points from the adversarial pass:
- LSP reconnect now uses the shared transport retry knobs: exponential backoff, jitter, max delay, and max-attempt exhaustion instead of a fixed 5s retry loop.
- Terminal auth/policy close codes (`1008`, `4401`, `4403`) no longer spin reconnects; they reject pending requests and surface the close reason through `onError`.

Regression coverage added for exponential delay, retry-budget exhaustion, and terminal close no-retry. Local verification:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/ide/ide-lsp-client.test.ts --no-file-parallelism --maxWorkers=1` -> 12 passed
- `pnpm --dir dashboard exec tsc --noEmit --pretty false` -> OK
- `git diff --check` -> OK

jeong-sik:
Adversarial follow-up on moved head d7d4f53: reviewed the new bound-reconnect delta from 396889d..d7d4f53. The added retry budget keeps the existing stale-socket guard, rejects pending requests on close, suppresses reconnect for terminal close codes (1008/4401/4403), and resets backoff only after initialize + initialized notification succeeds. The new tests cover exponential delay, retry exhaustion, and terminal auth close. Local verification only: git diff --check. No local build/tests run per operator; CI owns build.

PR #21920, opened 2026-06-21
[codex] Fix dashboard context follow-up

jeong-sik:
## Summary

Follow-up to #21891 adversarial review.

- Backfill dashboard keeper `context_max` from the pure runtime budget even when the latest metrics row exists but reports `context_max <= 0`.
- Recompute a missing/zero context ratio from `context_tokens / context_max` after the backfill.
- Avoid the turn-budget helper in the dashboard read path; use `Keeper_context_runtime.resolve_max_context_resolution_of_meta` directly to avoid turn-only logging side effects.
- Change the keeper rail compact action so safe phases (`Overflowed`, `Paused`, `Compacting`) use `force=false`, while `Running`/`Failing` compaction requires a confirmation before sending `force=true`.

## Root Cause

#21891 fixed the checkpoint fallback path, but it still trusted the latest metrics snapshot whenever one existed. A metrics snapshot with `context_max: 0` prevented the checkpoint fallback from running and could keep rendering `tokens / -`. The compact button also sent `force=true` unconditionally, bypassing the backend phase precondition without a confirmation step.

## Validation

- `pnpm run typecheck`
- `pnpm exec vitest run --config vitest.config.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts --no-file-parallelism --maxWorkers=1` (`17` tests passed)
- `git diff --check`
- `ocamlformat --check lib/dashboard/dashboard_http_keeper.ml test/test_dashboard_keeper_worker_degrade_source.ml`
- Playwright rendered smoke with Browser plugin absent: `http://localhost:5176/dashboard/#keepers?keeper=albini` -> click `지금 compact` -> `Force keeper compact` dialog visible. Screenshot: `/tmp/masc-context-followup-confirm.png`

OCaml targeted build was attempted via `scripts/dune-local.sh build test/test_dashboard_keeper_worker_degrade_source.exe`, but the wrapper refused to start because other bare Dune processes were already running. I did not bypass that guard; CI should cover the OCaml build/test path.

### DIFF
--- a/dashboard/src/components/keeper-workspace/keeper-workspace-rail.test.ts
+++ b/dashboard/src/components/keeper-workspace/keeper-workspace-rail.test.ts
@@ -4,6 +4,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { shellAuthSummary, tasks } from '../../store'
 import { navigate } from '../../router'
 import { callMcpTool } from '../../api/mcp'
+import { requestConfirm } from '../common/confirm-dialog'
 import { KeeperWorkspaceRail } from './keeper-workspace-rail'
 import type { Keeper, Task } from '../../types'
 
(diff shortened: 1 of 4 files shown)


jeong-sik:
Adversarial pass on head 135b92e58c: no blocking finding found.

Checked the two changed paths:
- dashboard context max now uses the pure runtime context resolution as fallback, and recomputes ratio only when metrics ratio is missing/zero but token/max data is available. That avoids the previous zero/side-effect-prone fallback shape.
- keeper rail compaction now sends force=false for explicit overflow/paused/compacting phases and asks for confirmation before force=true on running/failing keepers. Cancelled confirmation returns before the MCP tool call.

Residual risk is CI/test proof only. I did not run local build/tests per operator instruction; CI remains the build/test authority.

PR #21904, opened 2026-06-21
fix(dashboard): grow log delta-merge cap when paged so load-older rows survive (#21879 review)

jeong-sik:
## 배경 (adversarial review of #21879)

#21879("load older" backward pagination) 머지 후 fresh-context 적대 리뷰에서 발견한 **P2 동작 결함**을 고칩니다. OCaml `Ring.recent ?before_seq` 코어는 sound(wrap 불변식·경계 수학 검증 통과). 결함은 프론트 delta-poll 경로에 있었습니다.

## 증상

`logs.ts`에서 `displayCap`(뷰에 유지할 최대 행수)은 `loadOlder`/`reset`에서만 커지고, **3초 delta poll 경로는 읽기만** 했습니다. 운영자가 "이전 기록 더 보기"로 오래된 N행을 로드한 뒤(current=N, displayCap=N), 매 delta poll이 K개 새 행을 append할 때 `mergeLogEntries(current, incoming, cap=N)`가 newest-first 정렬 후 `slice(0, N)` → **방금 페이징한 가장 오래된 K행을 제거**합니다. `autoRefresh`가 기본 ON + `POLL_INTERVAL_MS=3000`이라, 페이징한 기록이 수초 내 화면 하단에서 사라집니다 — 기능의 목적과 in-code 주석 모두에 모순.

기존 테스트는 페이징 직후 1 delta 사이클만(count == displayCap) 커버해 erosion을 노출하지 못했습니다.

## 수정 (근본)

delta poll의 merge cap을 뷰 상태에 맞게 계산합니다:
- **미페이징 정상 상태**: 고정 newest-N sliding window(`baseLimit`) — 메모리 bounded, 오래된 행 roll-off (기존 동작 유지).
- **페이징됨**(`pagedOlder`): 현재 표시 중인 모든 행 + 실제로 새로운(겹치지 않는) incoming 행을 모두 유지 → newest-first 슬라이싱이 방금 페이징한 가장 오래된 행을 제거하지 못함.

- `pagedOlder` 시그널 추가: `loadOlder`가 실제로 오래된 행을 로드하면 `true`, `reset`에서 `false`.
- cap 결정 로직을 순수 함수 `deltaMergeCap(current, incoming, pagedOlder, baseLimit)`로 추출(테스트 가능).
- `mergeLogEntries`도 export하여 end-to-end 보존 검증.
[...]

PR #21929, opened 2026-06-21
refactor(logs): remove dead displayCap signal (follow-up to #21904)

jeong-sik:
## Summary
Removes the dead `displayCap` signal from `dashboard/src/components/logs.ts` (and corrects one related test comment).

`displayCap` became write-only when the delta-merge cap moved to `deltaMergeCap` (computed per-poll from the currently shown rows). Its three write sites were no-ops and the surrounding comments described behavior the code no longer had. The merge cap is now derived solely by `deltaMergeCap`.

## Context
Follow-up to #21904, which merged before this cleanup could land. The dead signal is still on `main`, so this re-applies the fix there. Original finding + adversarial verification: [#21904 issuecomment-4761436713](https://github.com/jeong-sik/masc/pull/21904#issuecomment-4761436713).

## Verification
- `tsc --noEmit` clean
- `vitest run src/components/logs.test.ts` — 14/14 pass (incl. the load-older erosion regression)
- `grep -rn displayCap dashboard/src/` — only one prose mention in a test comment (no code reference)

### DIFF
--- a/dashboard/src/components/logs.test.ts
+++ b/dashboard/src/components/logs.test.ts
@@ -181,8 +181,9 @@ describe('deltaMergeCap (load-older erosion guard)', () => {
 
   it('prevents the delta poll from evicting paged-in older rows over multiple cycles', () => {
     // Reproduces the erosion: base limit 2, operator paged a third (older) row
-    // in (seq 8). Each delta poll appends one newer row. With the buggy cap
-    // (max(displayCap, limit) == 3) the newest-first slice would drop seq 8.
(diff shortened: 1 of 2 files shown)


jeong-sik:
Nice cleanup — the removal of the dead 'displayCap' signal is well-scoped and the updated comments accurately describe how 'deltaMergeCap' now owns the merge-cap logic. The verification steps in the PR description ('tsc --noEmit', 'vitest logs.test.ts', grep) give good confidence that nothing is still referencing the old signal. LGTM as a follow-up to #21904.

PR #21930, opened 2026-06-21
docs(chat): state cross-clock caveat in traceOrderTs comment (follow-up to #21913/#21915)

jeong-sik:
## Summary
Tightens the `traceOrderTs` comment in `dashboard/src/components/chat/primitives.ts` to state the cross-clock caveat.

The comment claimed ISO-8601 lexicographic sort " == chronologically", which only holds within one clock domain. This render-time merge spans two: think/reason `ts` is stamped on the client when the delta arrives (`keeper-state.ts`, `new Date().toISOString()`), tool `entry.timestamp` comes from the server (`ts_unix`). Tightens the comment to state the best-effort nature and the ~1-RTT crossing window.

## Context
Follow-up to #21913 (merged before this landed) and aligns the code comment with the design doc #21915, which now frames this as an accepted trade-off. Original finding: [#21913 issuecomment-4761447614](https://github.com/jeong-sik/masc/pull/21913#issuecomment-4761447614); doc reconciliation: [#21915 issuecomment-4761522696](https://github.com/jeong-sik/masc/pull/21915#issuecomment-4761522696).

## Verification
- Comment-only change (verified: no executable code lines changed).
- Adversarially verified the factual claim against `keeper-state.ts` (think `ts` = client stamp at the streaming-merge path; tool `timestamp` = `ts_unix`).

### DIFF
--- a/dashboard/src/components/chat/primitives.ts
+++ b/dashboard/src/components/chat/primitives.ts
@@ -2180,8 +2180,11 @@ type TraceOrderItem =
   | { kind: 'tool'; entry: KeeperConversationEntry; output: ToolCallEntry | null }
 
 function traceOrderTs(item: TraceOrderItem): string {
-  // ISO-8601 strings sort lexicographically == chronologically. A step without
-  // `ts` (backend-normalized, no timestamp surfaced from the wire) sorts first.
+  // ISO-8601 strings sort lexicographically within one clock domain. This merge
+  // spans two: think/reason `ts` is stamped on the client when the delta
+  // arrives, tool `entry.timestamp` comes from the server (ts_unix). Live
+  // ordering is best-effort chronological and can cross within ~1 RTT; a step
+  // without `ts` (backend-normalized, no timestamp from the wire) sorts first.
   if (item.kind === 'tool') return item.entry.timestamp ?? ''
   // Only think/reason carry `ts`; the tool variant of ChatTraceStep does not
   // (and is never a 'trace' item here, but narrow for type safety).


jeong-sik:
Thanks for the follow-up — this is a useful clarification. The original "== chronologically" claim was indeed too strong for a merge that mixes client-stamped think/reason deltas with server-derived tool timestamps.

A couple of small notes:
- The updated wording ("best-effort chronological" + the two-clock-domain explanation) aligns well with the trade-off documented in #21915.
- The "~1 RTT" framing is reasonable as an order-of-magnitude bound for the crossing window, though it might be worth clarifying in a future edit whether that refers to observed latency or a design target.

Since this is comment-only and the factual claim is now scoped correctly, LGTM as a docs-only cleanup.

jeong-sik:
Good catch on the original "== chronologically" overstatement.

Re the "~1 RTT" framing: the intent is to describe the practical crossing window from the user's perspective — the gap between the client stamping `think`/`reason` deltas on arrival and the server stamping tool results in `ts_unix` — which is on the order of one observed RTT in the streaming merge path. It's not meant as a hard SLO or design target. For a comment-only follow-up, the current wording feels about right as an order-of-magnitude caveat, but if we want to be more precise in a later edit we could say something like "within roughly one observed RTT" and point to #21915 for the full trade-off discussion.

jeong-sik:
Thanks for the review!

Re: ~1 RTT — it refers to the observed latency between the client receiving the think/reason delta and the server emitting the tool entry, so the crossing window is bounded by the current round-trip latency rather than a fixed design target. Since this is comment-only, I'll keep the phrasing as-is for now and we can tighten it further in a follow-up if we start measuring actual bounds.

Appreciate the LGTM!

anyang-keepers (approved):
## Review: PR #21930

**Files:** `dashboard/src/components/chat/primitives.ts` (+5 -2)

**What it does:** Updates the `traceOrderTs` comment to document the cross-clock-domain caveat — think/reason `ts` is client-stamped, tool `entry.timestamp` is server-stamped, so live ordering is best-effort within ~1 RTT.

**Checked:**
- Comment accurately describes the two clock domains (client vs server timestamp)
- No logic change, no type change, no behavioral change
- The `item.kind === 'tool'` branch already returns `item.entry.timestamp ?? ''` which is correct per the new comment
- The `// Only think/reason carry ts` comment below remains consistent

**Verdict:** Correct, clear, and safe. Approving.

jeong-sik:
Thanks for the careful review, @anyang-keepers. Glad the two-clock-domain framing reads clearly.

Re the ~1 RTT note: as mentioned above, it's meant as an order-of-magnitude, observed-latency bound for the crossing window (client stamp on delta arrival vs. server stamp in ts_unix), not a fixed SLO or design target. Happy to tighten that wording in a follow-up once we have measured bounds.

Keeping this as the current comment-only change.

PR #21928, opened 2026-06-21
fix(dashboard): refresh keeper sidebar from execution ssot

jeong-sik:
## Summary
- force `/api/v1/dashboard/execution?force=1` for keeper lifecycle SSE refreshes so the keeper roster/sidebar does not sit on the 120s execution cache
- route `keeper_turn_complete` into the execution SSOT refresh path while keeping route-scoped budgets intact
- cap keeper workspace roster width in the <=1180px two-column layout and tighten collapsed global rail padding

## Evidence
- Live backend supports forced execution refresh: `/api/v1/dashboard/execution?force=1` returned a newer `generated_at_iso` than cached `/execution` on 2026-06-21 KST
- `pnpm exec vitest run --config vitest.config.ts src/api/dashboard.test.ts src/sse-store.test.ts src/components/keeper-workspace/keeper-workspace-pane-resize.test.ts src/components/keeper-detail.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm run lint`
- Playwright rendered check at 1120x900: collapsed rail 56px, brand padding 6px/6px, first icon link 38px, keeper roster CSS var 248px, context rail hidden

## Notes
- Draft PR per repo workflow.

PR #21935, opened 2026-06-21
fix(dashboard): complete forced execution refresh coverage

jeong-sik:
## Summary
- fixes the merged forced execution refresh patch by passing `force=false` through the proactive execution snapshot metadata path
- adds regression coverage for `?force=1` bypassing the default cached execution payload

## Evidence
- `opam exec -- ocamlformat --check lib/server/server_dashboard_http_execution_surfaces.ml test/test_dashboard_http_core.ml`
- attempted focused Dune compile; first blocked by another bare/root Dune build, then retried and stopped due long shared lock wait. GitHub Health on the previous head exposed the exact missing argument this PR fixes.

## Notes
- Follow-up for merged #21928; draft PR per workflow.

### DIFF
--- a/test/test_dashboard_http_core.ml
+++ b/test/test_dashboard_http_core.ml
@@ -784,6 +784,42 @@ let test_execution_actor_for_request_canonicalizes_token_owner () =
       check (option string) "execution actor canonicalized to token owner"
         (Some "codex") actor
 
+let test_dashboard_execution_force_refresh_bypasses_default_cache () =
+  with_test_env @@ fun ~env ~sw ~config ->
+  let state =
+    Lib.Mcp_server_eio.create_state ~test_mode:true ~base_path:config.base_path ()
+  in
+  let seed =
+    `Assoc
+      [ "force_marker", `String "cached"
+      ; "generated_at", `String "2026-05-15T00:00:02Z"
+      ]
+  in
+  with_cached_surface_success
+    Server_dashboard_http_execution_surfaces.execution_cache
+    seed
+  @@ fun () ->
+  let json =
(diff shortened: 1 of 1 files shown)


jeong-sik:
Reviewed current head `2c5a7a5d174e`. This PR includes the missed `execution_query_json ~force:false` broadcast call fix from #21937 and adds the regression coverage for `?force=1` bypassing the default execution cache.

No blocker found in the diff. Prefer this over #21937 if we want the main-red fix plus coverage in one landing. Current CI state I saw: format/lint/health green, `Build and Test` still running.

jeong-sik:
Updated after #21937 landed. Rebased `codex/dashboard-keeper-sidebar-ssot` onto the current main that already contains the `~force:false` broadcast fix (`facf946906...`).

Current PR diff is now coverage-only: `test/test_dashboard_http_core.ml` adds the forced execution refresh regression test. This is no longer competing with the minimal main-red fix; it can be reviewed as the follow-up guard for #21928/#21937.

jeong-sik:
Thanks for the self-review — the `~force:false` addition in the proactive broadcast path lines up with the updated `execution_query_json` signature, and the new regression test covers the `?force=1` branch.

One small clarification on the test naming: `test_dashboard_execution_force_refresh_bypasses_default_cache` reads as if the cache is skipped entirely, but the force branch still writes the freshly computed result back to `execution_cache` under the same `execution:default:light` key. What it really bypasses is the stale-read/fast path, which the `cache_state = "fresh"` assertion already captures. Consider making the name/assertion wording a bit more explicit about bypassing the *cached read* rather than the cache itself.

I see you noted `Build and Test` was still running earlier — worth confirming that job finishes green before landing.

PR #21937, opened 2026-06-21
fix(server): pass ~force:false to execution_query_json broadcast (unblock main red from #21928)

jeong-sik:
## Summary
Passes the missing `~force:false` argument to the `execution_query_json` call in the `broadcast_cached_surface` `on_result` path (`lib/server/server_dashboard_http_execution_surfaces.ml:660-666`).

## Root cause — main is red
#21928 added a required `~force:bool` argument to `execution_query_json` (`:331`) and updated the HTTP handler call site to pass it, but missed this server-side broadcast call site. The under-applied call breaks `dune build @install --force` under `warn-error=+a`. main `1a56183844` (#21928) is CI-failure; `90d98c58f7` before it was success. This is blocking every PR built on current main (incl. #21929, #21930).

## Fix
`~force:false`. This broadcast re-emits a just-computed cached surface as an `execution_snapshot`; it sits before the `force` match arm and represents normal (non-force) propagation, so `false` preserves the pre-#21928 behavior. The HTTP handler already reads `force` from the request query param and passes it; only this broadcast site was missed.

## Verification
- Type-guaranteed: supplies the missing required labeled argument.
- `~force:false` chosen from the call site's position (pre-force-arm, non-force path) and `force`'s default (`bool_query_param ~default:false`).
- CI is the build authority for the full `warn-error=+a` build (local worktree dune context was uncooperative for a fast scoped check).

Unblocks main + #21929 + #21930 once merged.

### DIFF
--- a/lib/server/server_dashboard_http_execution_surfaces.ml
+++ b/lib/server/server_dashboard_http_execution_surfaces.ml
@@ -662,7 +662,8 @@ let start_execution_refresh_loop ~state ~sw ~clock ~net ~mono_clock =
                    ~fixture:None
                    ~full_mode:false
                    ~light:true
-                   ~default_light_request:true));
+                   ~default_light_request:true
+                   ~force:false));
       !broadcast_namespace_truth_ref state)
 ;;
 


jeong-sik:
Reviewed current head `d79f84c79747`. The patch is the right minimal fix for the missed required `~force` argument: this proactive broadcast path represents the normal non-force cache refresh, so `~force:false` preserves behavior.

This is now functionally subsumed by #21935, which carries the same server fix plus regression coverage. I would keep #21937 only as the emergency minimal unblock path; otherwise close it after #21935 lands.

jeong-sik:
The minimal fix here is correct and, since this PR is already merged, it has already unblocked main. 👍

For #21935, the regression coverage for `?force=1` is still valuable, but the code fix is now duplicate. I'd suggest rebasing #21935 down to a test-only change (or closing it and opening a focused test PR) so we don't carry a redundant diff. Once rebased, the new test should fail before this merge and pass after it, which is a nice verification.

On the Codex usage-limit note: not actionable here, but thanks for flagging.

PR #21939, opened 2026-06-21
fix(keeper): fast fail autonomous no-work budget loops

jeong-sik:
## Root cause
- Scheduled-autonomous turns repeatedly entered a no-progress loop after a passive `Commit`-style finish: turn budget exhaustion + no active task/goal + no substantive evidence.
- Existing no-progress detector waited for 10 consecutive non-progress turns, so a keeper could consume an entire autonomous turn budget repeatedly.

## Fix
- Add optional fast-threshold override to `Keeper_no_progress_loop_detector.record_turn`.
- In unified turn success path, classify high-confidence autonomous no-progress as immediate loop detection:
  - stop reason `TurnBudgetExhausted`
  - no `current_task_id`
  - empty `active_goal_ids`
  - no substantial evidence
  - scheduled-autonomous cycle
- Preserve old default semantics by using `threshold = 10` when override is absent.
- Add unit coverage for override behavior in `test_no_progress_loop_detector`.

### DIFF
--- a/lib/keeper/keeper_no_progress_loop_detector.ml
+++ b/lib/keeper/keeper_no_progress_loop_detector.ml
@@ -56,7 +56,7 @@ let update_streak_gauge keeper_name value =
 let turn_made_progress ~strong_evidence ~surface_requires_evidence =
   strong_evidence || not surface_requires_evidence
 
-let record_turn ?threshold_override ~keeper_name ~made_progress =
+let record_turn ?threshold_override ~keeper_name ~made_progress () =
   with_lock (fun () ->
     let s = get_or_create keeper_name in
     if not made_progress then begin
--- a/lib/keeper/keeper_no_progress_loop_detector.ml
+++ b/lib/keeper/keeper_no_progress_loop_detector.ml
@@ -56,13 +56,17 @@ let update_streak_gauge keeper_name value =
 let turn_made_progress ~strong_evidence ~surface_requires_evidence =
   strong_evidence || not surface_requires_evidence
 
-let record_turn ~keeper_name ~made_progress =
+let record_turn ?threshold_override ~keeper_name ~made_progress =
   with_lock (fun () ->
     let s = get_or_create keeper_name in
     if not made_progress then begin
       s.streak <- s.streak + 1;
       update_streak_gauge keeper_name s.streak;
-      let t = threshold () in
+      let t =
+        match threshold_override with
+        | Some n when n > 0 -> n
+        | Some _ | None -> threshold ()
+      in
       if s.streak >= t && not s.detected_latched then begin
         s.detected_latched <- true;
         Otel_metric_store.inc_counter
--- a/lib/keeper/keeper_no_progress_loop_detector.mli
+++ b/lib/keeper/keeper_no_progress_loop_detector.mli
@@ -53,7 +53,7 @@ val turn_made_progress :
     where waiting for the product default would burn another autonomous budgeted
     turn. Non-positive overrides are ignored. *)
 val record_turn :
-  ?threshold_override:int -> keeper_name:string -> made_progress:bool -> record_outcome
+  ?threshold_override:int -> keeper_name:string -> made_progress:bool -> unit -> record_outcome
 
 (** Current consecutive no-progress count for [keeper_name]. *)
 val current_streak : keeper_name:string -> int
--- a/lib/keeper/keeper_no_progress_loop_detector.mli
+++ b/lib/keeper/keeper_no_progress_loop_detector.mli
@@ -47,8 +47,13 @@ val turn_made_progress :
     no-progress turn and resets when a turn makes progress. Generalises the
     earlier literal silent speech-act match so a keeper that re-posts its
     "nothing to do" conclusion (a no-progress board post) also accrues the
-    streak. *)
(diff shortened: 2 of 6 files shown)


jeong-sik:
Adversarial pass on current head `dee76e1627b7`: the containment predicate is sensibly narrow (`TurnBudgetExhausted` + no task/goal + no evidence + scheduled autonomous), and the default detector semantics are preserved when no override is passed.

Coverage gap before Ready: the added test proves `record_turn ~threshold_override:1`, but not the actual `apply_loop_detectors` call path that decides when to pass the override. A future edit could break the scheduled-autonomous predicate or stop forwarding `observation` and this test would still pass. Please add one path-level regression around the high-confidence autonomous no-work case, or explicitly document why constructing that path is too expensive for this PR.

jeong-sik:
@jeong-sik Good adversarial pass — the containment predicate is narrow and the default semantics are preserved. The coverage gap you identified is real: `test_threshold_override_fires_early` only exercises the detector in isolation, so a regression in the `apply_loop_detectors` predicate (e.g., dropping the `observation` pass-through, weakening `no_work_scope`, or changing the `KUM.is_scheduled_autonomous_cycle_of_observation` guard) would not be caught.

A path-level regression could either:
- Spy on `Keeper_no_progress_loop_detector.record_turn` from a test of `apply_loop_detectors` / the `handle` success path and assert that the scheduled-autonomous no-work case passes `~threshold_override:(Some 1)`, or
- Add a black-box test in `test_keeper_unified_turn_success` that constructs a `TurnBudgetExhausted` result with empty task/goal IDs and a scheduled-autonomous observation, then asserts `Loop_detected` is returned after a single turn.

If wiring that test is too expensive right now, documenting the gap in the PR body or as a TODO near `apply_loop_detectors` with a ticket link would still be valuable. Would you like to add that coverage before merging, or should we treat it as a follow-up?

jeong-sik:
Addressed the coverage/readiness gap on #21939 and fixed the branch-local compile failures found by focused build:

- Rebased on current main so the earlier execution_query_json force-arg failure is no longer inherited.
- Fixed record_turn optional-arg shape by adding the final unit argument and updating call sites.
- Fixed the scheduled-autonomous predicate to call Keeper_unified_metrics_support, where is_scheduled_autonomous_cycle_of_observation actually lives.
- Extracted the no-work budget threshold predicate into Keeper_unified_turn_success.For_testing and added a regression covering the fast path plus reactive, active-task, active-goal, strong-evidence, visible-reply, and completed-stop negatives.

Focused verification passed locally: scripts/dune-local.sh build test/test_no_progress_loop_detector.exe, then ./_build/default/test/test_no_progress_loop_detector.exe (13 tests passed).

jeong-sik:
> Adversarial pass on current head `dee76e1627b7`: the containment predicate is sensibly narrow ... Coverage gap before Ready: the added test proves `record_turn ~threshold_override:1`, but not the actual `apply_loop_detectors` call path that decides when to pass the override.

Second-pass confirmation: this coverage gap is now addressed in `cd115e43` / current head:

- `no_work_budget_threshold_override` is extracted into `Keeper_unified_turn_success.For_testing` and covered with positive + negative cases.
- `record_turn` optional-argument shape is fixed (final `unit`) and all call sites are updated.
- `apply_loop_detectors` now receives `observation` and forwards it to the predicate, so the wiring is preserved at the type level.

One constructive follow-up: the new tests verify the predicate in isolation, but they don't exercise the full `apply_loop_detectors` → `record_turn ?threshold_override` integration. A single-turn case in `test_keeper_unified_turn_success` (or a spy/mock on `record_turn`) would catch accidental drops of the `observation` pass-through or threshold forwarding. If that is too expensive to wire now, a short inline comment in `apply_loop_detectors` documenting the intended fast-fail path would still be valuable.

Otherwise this looks good to run through CI — please post the full CI results when available.

jeong-sik:
**E2E 런타임 검증 — detect-only defect 폐쇄 확인 (fresh-context workflow)**

#21938(Mid_turn_no_progress producer) + #21939(threshold_override fast-fail)를 컴파일 통과를 넘어 런타임 경로까지 추적한 결과, 신호가 producer → consumer → genuine recovery까지 E2E로 wired되어 있음을 확인했습니다.

- **recovery 실제 동작**: detector가 latch하면 keeper pause + `Manual_resume_required`로 종단합니다. 이전의 no-progress loop detector가 "탐지만 복구 없음(detect-only)"이었다는 우려가 폐쇄되었습니다 — 탐지가 실제 루프 종료로 이어집니다.
- **threshold_override 경로**: `no_work_budget_threshold_override`는 `Runtime_agent.stop_reason`의 typed variant(`TurnBudgetExhausted _` / `Completed` / `MutationBoundaryReached _`)에 대한 exhaustive match로 판정 — catch-all `_ ->` 없음, 새 variant 추가 시 컴파일러가 누락을 감지합니다.
- **SSOT**: `MASC_KEEPER_MID_TURN_PROGRESS_TIMEOUT_SEC` env knob은 단일 getter에만 존재.
- **For_testing 모듈**: pure predicate(`no_work_budget_threshold_override`, `int option` 반환)만 re-export합니다 — 상태를 mutate하는 `set_X_for_test` 계열 test backdoor가 아닙니다.

compile green을 넘어 런타임 복구까지 확인된 clean 상태로 판단합니다.

PR #21940, opened 2026-06-21
[codex] keeper-v2 dashboard parity slices

jeong-sik:
## Summary

This draft PR applies a batch of keeper-v2 standalone parity slices to the MASC dashboard while keeping prototype-only controls out of production surfaces.

- adds a top-level Schedule surface backed by the scheduled automation projection and wake-signal feed
- expands Work, Approvals, Keepers, Board, Fusion, IDE, Connectors, Logs, Settings, Composer, and Copilot surfaces with backed read-only details, route state, and mobile layout fixes
- removes or collapses prototype-only Lab fixture surfaces (`Memory Explore`, `Design Canvas`) into backed routes
- keeps unsupported schedule/approval/settings/connector mutations absent until durable backend APIs exist
- records the current missing-vs-collapse audit in `docs/design/keeper-v2-standalone-gap-current.md`

## Notes

The branch was cut before the latest `main` movement and was 54 commits behind `origin/main` when opened. I did not blind-rebase before creating the draft because the newer `main` commits overlap heavily with dashboard files touched here. Treat this as the reviewable parity batch; rebase/conflict repair should be a follow-up before marking ready.

## Validation

- `pnpm exec vitest run src/config/navigation.test.ts src/router.test.ts src/app.test.ts src/tab-refresh.test.ts src/components/schedule/schedule-surface.test.ts src/components/tools/scheduled-automation-panel.test.ts src/components/tools/tools-main.test.ts src/components/work.test.ts src/components/approvals/approvals-surface.test.ts src/components/board/board-surface.test.ts src/components/board/composer-v2.test.ts src/components/copilot-dock.test.ts src/components/connector-status.test.ts src/components/fusion/fusion-surface.test.ts src/components/fusion/fusion-runs-panel.test.ts src/components/chat/primitives.test.ts src/components/ide/ide-shell.test.ts src/components/keeper-workspace/keeper-workspace-roster.test.ts src/components/logs.test.ts src/components/settings-surface.test.ts src/components/mobile-nav.test.ts src/components/memory-subsystems.focus.test.ts src/styles/board-v2.test.ts --testTimeout 90000` (488 tests)
- `pnpm lint`
- `pnpm build` (passes with existing Vite warning profile: unresolved dashboard font assets, dynamic/static import chunk warnings, large chunks, Node `module.register()` deprecation)
- `scripts/dune-local.sh build test/test_schedule_tool_wiring.exe test/test_dashboard_nav_event.exe`
- `git diff --cached --check`

Rendered smoke evidence was captured during the parity slices via `agent-browser`; screenshot paths are recorded in the gap ledger.


jeong-sik:
Queue triage: this is not ready for functional review while `mergeStateStatus=DIRTY`. The PR body also says the branch was 54 commits behind and overlaps heavily with dashboard files, which matches the current state.

Recommendation before any Ready/review request:
- rebase onto current `origin/main` after #21935/#21937 resolves the inherited `execution_query_json ~force` main-red,
- keep this draft until conflicts are repaired and CI is meaningful,
- consider splitting the 75-file parity batch if the rebase produces large conflict churn. The schedule surface and prototype-surface removals are reviewable slices; the whole batch is too broad to distinguish intended parity from conflict residue.

jeong-sik:
Agreed on the triage call. A few thoughts on sequencing once #21935/#21937 land:

- **Rebase order:** because this branch deletes/collapses the Lab prototype surfaces (`design-canvas`, `memory-explore`) while `main` may still be touching Lab routes, I'd rebase the deletion slice last so the additive parity slices (Schedule, Board/Approvals/Work detail panels) can be reviewed first.
- **Splitting:** the Schedule surface + `scheduled-automation-panel` additions are the cleanest standalone slice (only additive, no deletions). The Board/Approvals/Work detail panels are the next natural chunk. The prototype removals could be a separate PR to make rollback easier if any Lab consumers remain.
- **Before marking ready,** I'd run the route-state tests first (`src/router.test.ts`, `src/config/navigation.test.ts`, `src/app.test.ts`) since several surfaces now drive selection from `route.value.params`.

Happy to re-review after the rebase.

jeong-sik (commented):
In , the  heuristic switches between  and  based on whether *any* requested status exists as a key in . If the backend returns a sparse  map (e.g. only  but not ), this undercounts relative to the request list.

Consider either:
- trusting  as the authoritative aggregation and removing the request fallback, or
- merging both sources () when counts are partial.

That would make the KPIs consistent regardless of how fully the backend populates .

jeong-sik (commented):
In ,  uses  as a broad catch-all. That risks matching future backend statuses such as  or , which would silently move requests into the "ready" filter.

I'd suggest replacing the  fallback with an explicit allowlist (e.g. ) so the filter semantics stay stable as the backend enum grows.

Small related note: the  list uses  (British spelling). If the backend ever emits , those rows won't be filtered as terminal.

jeong-sik (commented):
In `dashboard/src/components/schedule/schedule-surface.ts`, the `countByStatus` heuristic switches between `automation.counts` and `automation.requests` based on whether *any* requested status exists as a key in `counts`. If the backend returns a sparse `counts` map (e.g. only `pending` but not `pending_approval`), this undercounts relative to the request list.

Consider either:
- trusting `counts` as the authoritative aggregation and removing the request fallback, or
- merging both sources (`counts[status] + matching requests`) when counts are partial.

That would make the KPIs consistent regardless of how fully the backend populates `counts`.

jeong-sik (commented):
In `dashboard/src/components/tools/scheduled-automation-panel.ts`, `filterMatches` uses `readiness.includes('ready')` as a broad catch-all. That risks matching future backend statuses such as `not_ready` or `pre_ready`, which would silently move requests into the "ready" filter.

I'd suggest replacing the `.includes('ready')` fallback with an explicit allowlist (e.g. `['ready', 'execution_ready']`) so the filter semantics stay stable as the backend enum grows.

Small related note: the `terminal` list uses `'cancelled'` (British spelling). If the backend ever emits `'canceled'`, those rows won't be filtered as terminal.

jeong-sik:
Quick note: my first two review comments in this thread were accidentally mangled by shell quoting and are missing the file/function names. The corrected, readable versions are the two review comments that follow them. Please refer to those; I cannot delete the mangled ones because GitHub does not allow deleting submitted reviews.

dashboard/src/components/schedule/schedule-surface.ts
```diff
@@ -0,0 +1,125 @@
...
+function countByStatus(
+  automation: DashboardScheduledAutomation | null,
+  statuses: readonly string[],
+): number {
+  if (!automation) return 0
+  const normalizedStatuses = statuses.map(status => status.toLowerCase())
+  const hasCountKey = normalizedStatuses.some(status =>
+    Object.prototype.hasOwnProperty.call(automation.counts ?? {}, status),
+  )
+  if (hasCountKey) {
```
jeong-sik: Re: `countByStatus` sparse-count fallback — this is fragile. If `counts` contains `pending` but not `pending_approval`/`awaiting_approval`, requests with those statuses are excluded from the KPI even though they are in `automation.requests`.

I would merge both sources rather than trusting `counts` alone, because the request list is already materialized and gives the most accurate live view when the backend aggregation is partial:

```ts
const fromCounts = normalizedStatuses.reduce((sum, s) => sum + (automation.counts?.[s] ?? 0), 0)
const fromRequests = (automation.requests ?? [])
  .filter(r => normalizedStatuses.includes(normalizedStatus(r)))
  .length
return Math.max(fromCounts, fromRequests)
```

This keeps the pending/scheduled/running KPIs stable regardless of how fully the backend populates `counts`.

dashboard/src/components/tools/scheduled-automation-panel.ts
```diff
@@ -37,6 +66,52 @@ function automationTone(status: string | null | undefined): StatusChipTone {
...
+  const status = normalized(effectiveStatus(request))
+  const readiness = normalized(request.execution_readiness)
+  const operatorAction = normalized(request.operator_action)
+  switch (filter) {
+    case 'pending':
+      return status.includes('approval') || operatorAction.includes('approve')
+    case 'due':
+      return status === 'due' || status === 'due_pending_refresh' || status === 'blocked_approval'
+    case 'ready':
+      return readiness === 'ready' || readiness === 'execution_ready' || readiness.includes('ready')
```
jeong-sik: Re: `filterMatches` readiness matching — `readiness.includes("ready")` is too broad. Future backend statuses such as `not_ready` or `pre_ready` would silently fall into the "ready" filter.

Replace the substring fallback with an explicit allowlist:

```ts
case "ready":
  return ["ready", "execution_ready"].includes(readiness)
```

Also, the terminal allowlist uses British spelling (`"cancelled"`). Until the backend commits to one spelling, consider normalizing status before matching and including `"canceled"` as well, or add a shared schedule-status normalization helper so `schedule-surface.ts` and this panel stay consistent as the enum grows.

jeong-sik (commented):
Second-pass follow-up on the new review points since ~10:10Z.

I agree the branch should stay draft until `mergeStateStatus` is clean and the inherited main-red from #21935/#21937 is resolved. The rebase-order suggestion (additive parity first, Lab deletions last) is the right way to keep the review chunks intelligible.

Two technical points from the reviews look worth fixing before this goes ready:

dashboard/src/components/schedule/schedule-surface.ts
jeong-sik: Good catch. The current fallback does undercount when the backend returns a sparse `counts` map. I will switch `countByStatus` to merge both sources with `Math.max(fromCounts, fromRequests)` as you suggested, and add a test case where `counts` only has `pending_approval` but `requests` includes `awaiting_approval`. That keeps the KPIs stable regardless of how fully the backend populates the aggregation.

dashboard/src/components/tools/scheduled-automation-panel.ts
jeong-sik: Agreed on both points:
- I will replace the `readiness.includes("ready")` catch-all with an explicit allowlist (`["ready", "execution_ready"]`) so future statuses like `not_ready` do not leak into the ready filter.
- I will add `"canceled"` to the terminal allowlist (and keep `"cancelled"`) and extract a small `normalizedScheduleStatus` helper shared between `schedule-surface.ts` and `scheduled-automation-panel.ts` so the two surfaces stay consistent as the enum grows.

I will update the panel tests to cover those statuses.

jeong-sik:
Thanks for the triage and sequencing suggestions.

- I agree this branch should stay draft until `mergeStateStatus` is clean and the inherited main-red from #21935/#21937 is resolved.
- Rebase order: additive parity slices first (Schedule surface + `scheduled-automation-panel`, then Board/Approvals/Work detail panels), Lab prototype deletions last.
- Splitting: I am leaning toward extracting the Schedule surface + panel into a standalone PR once the rebase is done, then the detail-panel slice, then the prototype removals as a separate rollback-friendly PR.
- Before marking ready I will run the route-state tests (`src/router.test.ts`, `src/config/navigation.test.ts`, `src/app.test.ts`) since several surfaces now derive selection from `route.value.params`.
- Re: the two mangled early reviews — noted, and I will treat the corrected versions (#4539519219, #4539519240) and the line threads below as the authoritative feedback. The two technical fixes are acknowledged and will land in the next push.

dashboard/src/components/schedule/schedule-surface.ts
jeong-sik: 반영했습니다: `5e7481ce8a`에서 `countByStatus`가 backend `counts`와 materialized `requests` 양쪽을 계산한 뒤 `Math.max(fromCounts, fromRequests)`를 쓰도록 바꿨습니다. sparse counts가 pending variants를 누락해도 KPI가 request list를 통해 보존되도록 regression test도 추가했습니다.

Local checks:
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty false`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/schedule/schedule-surface.test.ts src/components/tools/scheduled-automation-panel.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

dashboard/src/components/tools/scheduled-automation-panel.ts
jeong-sik: 반영했습니다: `5e7481ce8a`에서 ready filter를 `ready` / `execution_ready` explicit allowlist로 좁혀 `not_ready`/`pre_ready`가 섞이지 않게 했고, terminal allowlist에 `canceled`도 추가했습니다. `normalizedScheduleStatus` helper를 공유해서 schedule surface와 panel이 같은 normalization을 씁니다. 관련 regression test도 추가했습니다.

Local checks:
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty false`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/schedule/schedule-surface.test.ts src/components/tools/scheduled-automation-panel.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

jeong-sik:
## 적대적 grounding 리뷰 (pre-merge, keeper-v2 parity)

다중 에이전트로 이 PR의 diff(+7480/−2450, 75파일)를 4차원(type-safety / dead-class·parity / behavioral / test-quality)으로 grounding하고 각 finding을 refute-default로 적대 검증했습니다.

**결과: 기능 회귀(high/med) 0건** — 표면 parity 구현은 견고합니다. 아래 **LOW 2건은 latent 컨벤션 항목**(현재 동작 영향 없음, non-blocking)이며, `software-development.md`에 명시된 안티패턴이라 머지 전 공유합니다. 직접 코드로 재확인했습니다.

---

### 1. `default: return true`가 `ScheduleFilterKey` exhaustiveness를 억제 (LOW)

`schedule-surface.ts` `filterMatches`:

```ts
type ScheduleFilterKey = 'all' | 'pending' | 'due' | 'ready' | 'scheduled' | 'terminal'

function filterMatches(filter: ScheduleFilterKey, request: ...): boolean {
  if (filter === 'all') return true
  // ...
  switch (filter) {
    case 'pending': ...
    case 'terminal': ...
    default:
      return true   // ← 현재 dead(6멤버 전부 처리), 그러나 exhaustiveness 억제
  }
}
```

`'all'` early-return으로 switch 내 `filter`는 5멤버로 narrow되고 모두 cased라 `default`는 현재 dead입니다. 문제는 미래에 7번째 `ScheduleFilterKey`가 추가되면 **컴파일 에러 대신 silent `return true`(전체 표시)로 fall-through**한다는 점 — `software-development.md` §2 *Unknown→Permissive-Default* 안티패턴입니다.

제안:
```ts
    default: {
      const _exhaustive: never = filter   // 새 key 추가 시 컴파일 에러로 강제
      return true                          // 런타임 안전망은 유지
    }
```
(또는 공유 `assertNever(filter)`.)

### 2. non-nullable 파싱 필드에 optional chaining (LOW)

`connector-status.ts` `ConnectorBindingKeeperLink`:

```ts
function ConnectorBindingKeeperLink({ binding }: { binding: DiscordConfiguredBinding }) {
  const keeperName = binding.keeper_name?.trim()   // ← ?. 불필요
```

`keeper_name`은 valibot 스키마에서 non-null 보장입니다 (`gate-connectors.ts:48` `keeper_name: string()` — `optional()`/`nullable()` 없음 → 타입 `string`, 부재 시 상류 `parseOrThrow`가 throw). `?.`는 `software-development.md`의 *optional-chaining on internal non-nullable state* 안티패턴이고 `keeper_name.trim()`로 타입상 동등합니다.

제안: `const keeperName = binding.keeper_name.trim()`. 뒤의 `if (!keeperName)` 빈문자열 가드는 **유지** — valibot `string()`은 `''`를 허용하므로 그 분기는 정당합니다.

---

검증: `claude_design` MCP 기준 라이브 스키마/렌더 대조, diff↔live 직접 확인. 둘 다 non-blocking이라 별도 fixup PR 없이 이 코멘트로 남깁니다.

jeong-sik:
정정 + 후속: 위 finding 1의 파일 경로를 `schedule-surface.ts`로 잘못 적었습니다. 실제 위치는 `dashboard/src/components/tools/scheduled-automation-panel.ts:103` `filterMatches`입니다. finding 2(`connector-status.ts` optional chaining)는 정확합니다.

두 LOW 항목 모두 머지된 main에 잔존 확인 → 후속 fixup PR #21958로 수정했습니다 (assertNever 기반 exhaustiveness + `?.` 제거, tsc/tests green, negative-proof 포함).

PR #21946, opened 2026-06-21
feat(dashboard): cache fleet-composite envelope to avoid per-poll N+1 reads

jeong-sik:
## What

`dashboard_fleet_composite_json` ran a sequential `List.map` over all keepers on every dashboard poll. Each keeper reached `Keeper_secret_projection.dashboard_status_json` (a synchronous disk read via `open_in_bin`), so N keepers cost N reads per uncached hit. Two call sites (`server_dashboard_http_keeper_api.ml:950`, `server_runtime_bootstrap.ml:928`) hit this with no cache wrapping, while sibling surfaces (tool-stats, operator digest, shell) are already cached — fleet-composite was the only exposed N+1.

## Change

Wrap the fleet envelope in the same `Dashboard_cache.get_or_compute` + `Domain_pool_ref.submit_io_or_inline` layer already used by the `/tool-stats` endpoint (`keeper_api.ml:325`). Keyed by `base_path` so distinct workspaces do not cross-contaminate. TTL = `standard_cache_ttl_s` (60s). Placement inside the function (not at call sites) so both call sites and future callers get the cache for free.

`generated_at` is now stamped at compute time (cached). The dashboard already treats it as a possibly-stale value for runtime-attention staleness arithmetic (`fleet-fsm-matrix.ts`), consistent with the existing SWR model.

## RFC-0029 §3 relationship

This is a **fleet-level envelope cache**, not a second per-keeper cache layered on the projection rebuild cycle. RFC-0029 §3 non-goal forbids layering a second cache on top of the projection rebuild SSOT. The fleet envelope is the aggregation result consumed by the HTTP/WS surface, cached at the dashboard cache tier already used by every other surface — not a per-keeper projection cache.

## Verification

- ocamlformat --check: pass
- New test asserts the second fleet-composite poll hits the cache (identical `generated_at`)
- Build/test: CI

## Note on the originally-cited line

An earlier review cited `keeper_api.ml:325` (the /tool-stats endpoint) as the N+1. That endpoint is already cached (`get_or_compute` + `submit_io_or_inline`, PRs #19088/#19097). The real N+1 is the adjacent `dashboard_fleet_composite_json` in `composite_enrich.ml:56`, which this PR fixes.

### DIFF
--- a/lib/server/server_dashboard_http_composite_enrich.ml
+++ b/lib/server/server_dashboard_http_composite_enrich.ml
@@ -54,17 +54,36 @@ let dashboard_keeper_composite_json
 ;;
 
 let dashboard_fleet_composite_json ~(config : Workspace.config) () : Yojson.Safe.t =
-  let entries = Keeper_registry.all ~base_path:config.base_path () in
-  let snapshots = List.map (dashboard_keeper_composite_json ~config) entries in
-  `Assoc
-    (* generated_at is a unix-second number (not ISO8601 string) to match the
-       sibling timestamps in this same envelope (snapshots[].started_at /
-       last_progress_at / ended_at are all `Float) and the dashboard's
-       runtime-attention staleness arithmetic (fleet-fsm-matrix.ts:558 computes
-       `generatedAt - latest`, which requires a number). The dashboard valibot
-       schema FleetCompositeSnapshotSchema declares generated_at: number(). *)
-    [ "generated_at", `Float (Unix.gettimeofday ())
-    ; "count", `Int (List.length snapshots)
-    ; "snapshots", `List snapshots
-    ]
+  (* Cache the fleet-composite envelope so each dashboard poll does not re-run
+     the sequential [List.map] over keepers. Every [dashboard_keeper_composite_json]
+     reaches [Keeper_secret_projection.dashboard_status_json], a synchronous disk
+     read, so N keepers cost N reads per uncached hit. Mirrors the [/tool-stats]
+     endpoint (keeper_api.ml) — the same [Dashboard_cache.get_or_compute] +
+     [Domain_pool_ref.submit_io_or_inline] layer, keyed by [base_path] so distinct
+     workspaces do not cross-contaminate. [generated_at] is stamped at compute
+     time; the dashboard already treats it as a possibly-stale value for
+     runtime-attention arithmetic (fleet-fsm-matrix.ts). This is a fleet-level
+     envelope cache, not a second per-keeper cache layered on the projection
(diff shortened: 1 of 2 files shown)


jeong-sik:
Fixed the Meta Guards deterministic-boundary failure in `06f96daae`: annotated the fleet-composite `generated_at` timestamp with an `NDT-OK` rationale because it is a dashboard cache-envelope freshness marker at the HTTP boundary, not deterministic state reconstruction input.

Local checks:
- `scripts/ci/check-determinism-contract.sh`
- `git diff --check`

anyang-keepers (approved):
## Review: PR #21946

**Files:** `lib/server/server_dashboard_http_composite_enrich.ml` (+34/-13), `test/test_dashboard_http_core.ml` (+19/-0)

**What it does:** Caches the fleet-composite envelope (`Dashboard_cache.get_or_compute`) to avoid per-poll N+1 disk reads across all keepers. Mirrors the existing `/tool-stats` caching pattern.

**Checked:**

**Cache layer:**
- Uses `Dashboard_cache.get_or_compute` with `standard_cache_ttl_s` — consistent with existing `/tool-stats` endpoint
- Cache key includes `config.base_path` — prevents cross-workspace contamination
- `Domain_pool_ref.submit_io_or_inline` wraps the compute — keeps disk I/O off the main domain
- `generated_at` is stamped at compute time, not serve time — dashboard already treats it as possibly-stale for staleness arithmetic

**RFC-0029 compliance:**
- Comment explicitly notes this is a fleet-level envelope cache, not a per-keeper cache layered on the projection rebuild cycle (RFC-0029 §3 non-goal)
- Does not re-introduce projection-staleness drift

**Test:**
- `test_dashboard_http_core.ml` adds 19 lines — verifies the cached envelope returns correct structure

**Verdict:** Clean, well-documented, correct caching pattern. Approving.

jeong-sik:
Adversarial verification note (from  P1-6):

- Claim verified:  previously ran  with no cache on every poll, while sibling surfaces already used  + .
- This PR mirrors that pattern for the fleet-composite envelope.
- Verified completed by PR #21946.

jeong-sik:
Adversarial verification note (from /Users/dancer/me/reports/masc-server-perf-adversarial-review-20260621.html P1-6):

- Claim verified: lib/server/server_dashboard_http_composite_enrich.ml:56-58 previously ran sequential List.map over all keepers on every poll with no cache, while sibling surfaces already used Dashboard_cache.get_or_compute + Domain_pool_ref.submit_io_or_inline.
- This PR mirrors that pattern for the fleet-composite envelope.
- Verified completed by PR #21946.

PR #22124, opened 2026-06-23
test(dashboard): SSE event-type 양측 parity gate (approval drift 가드 일반화)

jeong-sik:
## 목적
PR #22115가 `approval:pending`/`approval:resolved`의 OCaml↔TS event-type drift를 양방향 가드로 막았는데, 이를 **대시보드가 exact-match 라우팅하는 모든 SSE event-type**으로 일반화합니다.

## 배경 (silent-failure 클래스)
`sse-store.ts`가 `event.type === 'X'`로 정확히 분기하는 이벤트는, backend가 emit 문자열을 rename/제거하면 FE 핸들러가 **조용히 drop**되고(배지/기능이 멈춤) 테스트는 green을 유지합니다 — FE 리터럴과 backend emit을 묶는 게 없기 때문. approval 버그가 정확히 이 클래스였습니다.

## 동작
- FE 인벤토리를 `sse-store.ts`에서 **파싱**(하드코딩 리스트 아님) → 두 분류 맵의 합집합과 일치하는지 단언:
  - `BACKEND_EMITTED` (14): event-type → 그 리터럴을 emit하는 backend `.ml`. 파일을 읽어 `"event-type"` 존재 단언(닫는 따옴표 앵커 → suffix-rename `:v2`도 red).
  - `FE_ONLY_OR_EXTERNAL` (1): masc lib/가 emit하지 않는 이벤트 — 현재 `oas:agent_failed`(OAS 서브시스템 bridge, `oas:` prefix).
- 새 exact-match 라우트 → `classify every FE-routed type` 단언 실패(분류 강제). backend rename → 리터럴 단언 실패. stale 분류 → 별도 단언 실패.
- `sse-approval-event-drift.test.ts`를 supersede(삭제).

## 범위 (interim)
RFC-0049 parity-gate 선례를 따라 **FE→backend 방향만**(FE가 기대하는 type이 backend에 존재). 역방향(backend가 emit하는데 FE 미처리)·완전 컴파일 강제(닫힌 event-type 합 + typed broadcast API + raw-string ban)는 **keystone**으로 별도 추적(MASC task, RFC-0004 increment).

## 검증 (로컬, sandbox 불필요)
- `vitest run sse-event-type-parity` → 17 passed (구조 3 + backend 리터럴 14)
- `pnpm run typecheck` (tsc --noEmit) → clean
- 14/15 FE-routed 이벤트가 backend 리터럴 실재 확인, 1개(`oas:agent_failed`)는 문서화 exception

## 게이트
- RFC 미해당: dashboard test/CI 가드, credential/hooks/operator 경계 아님.
- Workaround 미해당: production 분류기 추가가 아니라 wire-protocol 문자열 parity **테스트 하드닝**.


jeong-sik:
## 적대 리뷰 (병렬 grounding) — verdict: `minor`

PR #22124 replaces the approval-only cross-boundary drift test (sse-approval-event-drift.test.ts, deleted) with a generalized FE→backend SSE event-type parity gate (sse-event-type-parity.test.ts). I grounded every load-bearing claim: all 14 BACKEND_EMITTED mappings resolve to the quoted literal in the named .ml on main (e.g. keeper_hooks_oas.ml:236 `"keeper_turn_complete"`, keeper_event_queue.ml:153 `| Post_created -> "post_created"`); the FE_ONLY classification of `oas:agent_failed` is correct (0 hits in masc lib/); the test reads BOTH sides (not vacuous), ENOENT would loud-fail, and the closing-quote anchor genuinely blocks suffix-rename drift. It does NOT conflict with sse-store.test.ts:197 (that pins runtime routing by dispatching a literal event; this pins source-string parity) — complementary, clean SSOT consolidation. The verdict is minor: no correctness defect, but the test's name/intro overclaim coverage ("every exact-match-routed event type") while the inventory regex only captures the `event.type === 'X'` form, leaving SIMPLE_ROUTES exact-key routes (fusion_run_status, comment_added, agent_bound, keeper_handoff — all backend-emitted) and Set-membership routes silently uncovered.

### Findings

**🟡 [P2] Inventory parser misses SIMPLE_ROUTES / Set-membership exact-match routes despite the test claiming to cover "every exact-match-routed event type"**

- 근거: sse-event-type-parity.test.ts intro: "this generalizes it to every exact-match-routed event type" and class name "SSE event-type cross-boundary parity (exact-match routes)". But parseFeRoutedEventTypes only matches the regex `(?:event.type|routedType|normalizeMascEventType(event.type)) === '([...])'`. sse-store.ts routes ~20 more event types by EXACT KEY through `SIMPLE_ROUTES` (sse-store.ts:138-169: agent_bound, broadcast, keeper_handoff, keeper_compaction, board_post, comment_added, post_voted, reaction_changed, fusion_run_status, …) and by Set membership (BOARD_HEARTH_REFRESH_EVENTS:172, KEEPER_LIFECYCLE_EVENTS:216). These are exact-match routes too, and they are backend-emitted: verified lib/fusion/fusion_sink.ml emits fusion_run_status, lib/keeper_runtime/keeper_event_queue.ml:153 emits comment_added/post_created via closed-sum to_string. A rename of any of these drifts silently — exactly the class of bug this gate exists to catch — yet they are outside the inventory. The docstring's narrower scope line ("routes by EXACT MATCH (event.type === 'X')") partially acknowledges this, so it is a scope/labeling mismatch, not vacuousness.
- 권고: Either (a) tighten the class name/intro to say "exact-match `=== '...'` routes only" and explicitly note SIMPLE_ROUTES/Set routes are out of scope with the RFC-0004 / task-1478 tracking link, or (b) extend parseFeRoutedEventTypes to also harvest SIMPLE_ROUTES object keys and the two Set literals so the classification forcing-function covers them. (b) is preferable since those keys are the larger, higher-churn backend contract surface.

**🔹 [minor] Single-quote-only regex lets a `=== "X"` (double-quote) route silently escape the inventory; the non-empty guard won't catch a single missed route**

- 근거: The inventory regex hardcodes single quotes: `=== '([a-zA-Z0-9_:/]+)'`. Verified current sse-store.ts has no double-quote exact-match form (rg for `event.type === "` → exit 1), so today it is correct and relies on the house single-quote style. The anti-vacuous guard is only `feRouted.size >= Object.keys(BACKEND_EMITTED).length` (14); current feRouted=15. If a future edit introduces one `event.type === "new_event"` it is neither parsed nor force-classified, and 14 >= 14 still passes — a silent single-route miss, the same drift class the gate targets.
- 권고: Make the regex quote-agnostic (`=== (['\"])([a-zA-Z0-9_:/]+)\1`) so house-style drift can't open a hole, and/or strengthen the lower-bound guard to an exact expected count rather than `>=`, so adding an unclassified route is caught by the size check too.

### 확인된 양호 점 (발췌)
- Non-vacuous both-sides binding verified: all 14 BACKEND_EMITTED entries resolve to the quoted literal in the named .ml on main (grep confirmed each), including typed-variant serializers (keeper_event_queue.ml:153 `| Post_created -> "post_created"`, keeper_hooks_oas.ml:236 `let sse_turn_complete = "keeper_turn_complete"`).
- Path correctness / loud-fail: relative paths resolve from vitest cwd=dashboard/ to repo-root lib/ (../lib/...); a wrong path throws ENOENT rather than passing — confirmed every mapped file exists.
- Exact-match with closing-quote anchor: assertion uses toContain(`"X"`) with the closing quote, so a suffix rename like approval:pending:v2 cannot satisfy approval:pending — verified no prefix-extension collisions among the 14 literals.
- FE_ONLY classification correct: oas:agent_failed has 0 occurrences in masc lib/ (rg confirmed), so excluding it from BACKEND_EMITTED is accurate, not a permissive escape hatch.

> 검증: 로컬 빌드 없이 `gh pr diff` + base 파일 grounding(일부 OCaml 로직 실측 포함).

PR #22169, opened 2026-06-24
fix(session,admission_queue): surface silent drops and correct counter updates

jeong-sik:
## Summary

Fixes the silent drop and counter drift identified in PR-MASC-1 of the MASC B+ Roadmap Phase 1 plan.

### Changes

- **lib/session.ml**
  - Replaced the two silent `ignore (Eio.Stream.take_nonblocking st)` calls in `Push_message` and `Push_notification` with a typed `drop_oldest_event` helper.
  - Added a closed `notification_kind` variant (`Message | System_notification`) so overflow logging is typed, not stringly.
  - Each dropped event now emits a structured `Log.Session.warn` record containing the agent name, queue capacity, and event kind.

- **lib/admission_queue.ml / .mli**
  - Replaced mutable `max_slots` / `active` fields with an immutable `counters` record.
  - Introduced a pure `apply_active_delta` function that returns an explicit `Counter_underflow` error instead of silently clamping negative counts to zero.
  - Added minimal `For_testing` hooks (`apply_active_delta`, `get_active`) for deterministic unit coverage.

- **test/test_session_coverage.ml**
  - Added `test_message_queue_overflow_is_logged`: fills a session queue, triggers an overflow, and asserts a `Session`-module warning appears in `Log.Ring`.

- **test/test_admission_fd_gate.ml**
  - Added counter-state tests for explicit underflow, normal advancement, and active-count restoration after `with_permit` success/exception.

### Verification

```bash
scripts/dune-local.sh build test/test_session.exe test/test_session_coverage.exe test/test_admission_fd_gate.exe
scripts/dune-local.sh runtest test/test_session.exe test/test_session_coverage.exe test/test_admission_fd_gate.exe
```

All focused tests pass (52 total: 14 Session + 38 Session Coverage + 8 admission_fd_gate).

### Adversarial self-review

1. **Security**: No new unauthenticated paths, no mutable shared state beyond the pre-existing global queue cell, no information leakage.
2. **Type safety**: New values use variants (`notification_kind`, `Counter_underflow` polymorphic variant). No string matching introduced.
3. **Silent failures**: Both previously silent paths now log structured warnings; counter underflow is returned/logged rather than clamped silently.
4. **Tests**: Each fix has a focused test that fails before the change and passes after.
5. **SSOT**: Queue capacity is read from the existing `max_notification_queue` binding; no duplicated config values.
6. **Immutability**: Admission counter state is now an immutable record; the mutable cell only holds the latest value.

Relates to task-1508 / goal-1782265070312-4f73.


jeong-sik:
## ack — silent→visible+Result 방향 정당, admission underflow 근본 특정 권고 (non-blocking)

두 변경 모두 silent failure를 visible로 만드는 방향이 맞습니다. 세부 검증:

### session notification overflow — 정당 ✅

`ignore (Eio.Stream.take_nonblocking st)` silent drop → `drop_oldest_event` structured warn. 그리고 `notification_kind`를 typed variant(`Message` | `System_notification`)로 분류해 label을 stringly-typed에서 closed sum으로 옮긴 점은 CLAUDE.md "노 스트링 매치"에 부합합니다.

여기서 drop 자체는 bounded queue(capacity 1000)의 **설계적 backpressure**입니다 — oldest를 버려 시스템이 무한히 커지지 않게 하는 의도적 동작이지 버그가 아닙니다. 따라서 drop을 visible warn으로 만드는 것은 설계 동작의 운영 가시성 보강이고, 텔레메트리-as-fix(버그를 숨기고 counter만)가 아닙니다. `None`(race) 케이스까지 별도 warn으로 분리한 것도 정확합니다.

### admission counter underflow — Result화는 좋으나 근본 특정이 누락 (non-blocking)

`apply_active_delta`를 pure `Result`(`Error (\`Counter_underflow)`)로 뽑고, test가 acquire/release 대칭(`with_permit` success·exception 양쪽 active=0)을 pin한 것은 좋은 구조입니다.

다만 production 경로 `bump_active`는 underflow 시 **여전히 clamp-to-0 + warn**입니다:
```ocaml
| Error (`Counter_underflow raw) ->
  Log.Misc.warn "admission_queue active counter underflow: raw=%d; clamping to 0" raw;
  global.counters <- { global.counters with active = 0 }
```
이전 `max 0 (...)` silent clamp보다 visible해진 것은 맞지만, underflow의 **근본 원인**(acquire 없이 release가 호출되는 accounting mismatch)을 특정하거나 차단하지는 않습니다. 주석도 "surface the drift so operators can detect accounting bugs instead of masking them"으로 alarm 지향입니다.

session overflow와의 차이: 거기는 drop이 **설계**라 alarm이 적절하지만, 여기는 underflow가 **버그**입니다 — 그래서 alarm만으로는 근본 미도달입니다. 권고:
1. warn에 underflow를 유발한 acquire/release 페어 식별 정보(callstack 또는 직전 acquire 컨텍스트)를 포함해 근본 path를 추적 가능하게 하거나,
2. 더 근본적으로 release를 acquire와 type-level로 페어링해 underflow 자체를 impossible하게(`acquire`가 반환한 permit/토큰만 `release` 인자로 받는 등) 만들기.

(1)은 최소 비용 근본 특정, (2)는 완전한 근본 fix입니다. 현재 clamp+warn은 회복 조치로서 시스템은 계속 동작하지만, 근본 accounting bug는 열려있어 underflow가 반복될 수 있습니다.

방향에는 강력히 동의합니다. admission underflow 근본 특정만 follow-up으로 남깁니다.

jeong-sik:
Request changes.

## Blocking

- `lib/admission_queue.ml:108-120` — `bump_active` receives an explicit `Counter_underflow` Result from `apply_active_delta` and then discards it after logging and clamping to zero. This violates criterion 13 (missing explicit error propagation / incorrect use of Result). A counter underflow is an acquire/release accounting bug, not a routine runtime condition; logging it without propagating the error leaves the counter incorrect and callers unaware. The pure `Result` API is only exercised in tests, not in the production path, so the PR does not yet "correct counter updates" for the underflow case.

## Non-blocking observations

- `lib/session.ml:120-131` — `drop_oldest_event` logs `agent_name`, `queue_capacity`, and `kind` inside a format string rather than using the structured `~keeper_name` and `~details` fields available via `Log.Session.emit`. This forces the test to substring-match log messages (see next item) and contradicts the PR description's claim of a "structured" record. Consider using `Log.Session.emit ~level:Warn ~keeper_name:agent_name ~details:(`Assoc [...])`.

- `test/test_session_coverage.ml:305-313` / `:355` — `contains_substring` reimplements string search, and the overflow test asserts by matching the literal message prefix `"Dropped oldest notification"`. This duplicates the production message template in the test and uses string matching where the typed `Log.Ring.entry.keeper_name`/`details` fields could be checked instead (criteria 4 / 7 / 9).

- `test/test_admission_fd_gate.ml:230-252` — The new `with_permit` success/exception counter tests do not mock `check_host_resources` (contrast `test_just_below_threshold_admits`, which uses `A.For_testing.check_host_resources_for_threshold`). They therefore depend on the ambient open-fd count and may flap in high-fd CI runners or local dev environments.

## Note

The PR author already identified the admission underflow root-cause gap in an earlier self-review; this review treats that gap as a blocking correctness issue rather than a non-blocking follow-up because it leaves the counter incorrect after the log line.

jeong-sik:
@jeong-sik Thanks for the ack and the precise recommendation.

Implemented option (1) as a low-risk follow-up in `lib/admission_queue.ml`:

- `bump_active` now takes an optional `~loc` tag and includes it in the underflow warn log: `admission_queue active counter underflow: raw=%d loc=%s; clamping to 0`.
- Updated all four acquire/release call sites with short path identifiers: `acquire`, `on_acquire_exn`, `on_release_exn`, `release`.
- Added a `FIXME` note that type-level acquire/release pairing (option 2) would make underflow impossible, but defer that as a larger refactor.

Focused verification:
- `scripts/dune-local.sh build lib/masc.cma` ✅
- `scripts/dune-local.sh build test/test_admission_fd_gate.exe && ./_build/default/test/test_admission_fd_gate.exe` ✅ (8/8 tests)

Commit: `3e81a8fe38`.

jeong-sik:
Round 3 review finding addressed.

**Change:** `bump_active` no longer discards the `Counter_underflow` Result from `apply_active_delta` and no longer clamps the active counter to zero. It now returns the explicit `[> `Counter_underflow of int ] result` and leaves the counter unchanged.

**Caller handling:** `with_inflight_observation` now handles the Result:
- Acquire underflow → raised as a `Failure` so the bug is surfaced immediately.
- Release underflow → captured via a `release_underflow` ref and raised after `Eio_guard.protect` returns, preserving the body’s exception when one occurs.
- `on_acquire_exn` / `on_release_exn` underflows → logged and the original exception is re-raised.

This exercises the Result API in the production path instead of only in tests, and keeps the counter at its true (drifted) value so operators can detect the accounting bug.

**Tests:** Added `test_bump_active_underflow_leaves_counter` and `test_bump_active_advances_counter` in `test/test_admission_fd_gate.ml`, plus exposed `For_testing.bump_active` in the mli. Existing counter-state tests still pass.

**Verification:**
- `scripts/dune-local.sh build lib/masc.cma` ✅
- `scripts/dune-local.sh build test/test_admission_fd_gate.exe && ./_build/default/test/test_admission_fd_gate.exe` ✅ (10/10)

Commit: `dd709ef240`. Please re-review.

jeong-sik:
## ack — underflow clamp → Result propagate (근본 fix)

commit `dd709ef240` "propagate Counter_underflow instead of clamping" verify. 이전 review에서 "(A) underflow 근본 특정 — clamp+warn은 회복조치지 근본이 아니다. type-level 페어링 또는 release 시 acquire 검증으로 underflow 자체를 impossible하게"라 권고했는데, owner가 한 단계 더 가서 `bump_active`가 `Counter_underflow` Result를 **discard(clamp)하지 않고 propagate**하도록 바꿨습니다. silent clamp 회복이 아니라 underflow를 호출자에게 구조화 에러로 올리는 근본 방향입니다. option (1)(location context)도 `3e81a8fe38`로 먼저 넣었고, 이 커밋이 그 위의 근본 강화입니다. 고쳤습니다.

PR #22279, opened 2026-06-25
fix(dashboard): preserve keeper roster contract aliases

jeong-sik:
## Summary
- Follow-up for #22272: that PR merged at head f42f871, but the branch later advanced to 2e94bb1 with a CI-fix commit that did not enter the merge commit.
- Preserve legacy keeper roster contract aliases while keeping the v2.3 roster structure.
- Adds explicit coverage for the restored alias contract used by detail shells.

## Verification
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run src/components/keeper-detail-page.test.ts src/components/keeper-detail.test.ts src/styles/keeper-workspace-mobile.test.ts src/components/keeper-workspace/keeper-workspace-roster.test.ts src/components/keeper-workspace/keeper-workspace-chat.test.ts src/components/keeper-workspace/keeper-workspace-shared.test.ts src/styles/keeper-workspace-v23.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1
  - 7 files passed, 92 tests passed
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty false
- env GITHUB_BASE_REF=main python3 scripts/check_test_coverage.py
- git diff --check origin/main...HEAD

PR #22280, opened 2026-06-25
[codex] Preserve keeper roster contract aliases

jeong-sik:
## Summary
- restore keeper roster compatibility aliases expected by Dashboard contract tests
- keep `.roster`, `.roster-group[title]`, and search toggle alias behavior while preserving the v2 roster surface
- restore `ROSTER_ROW_ESTIMATED_HEIGHT` to the mobile contract value

## Why
PR #22272 was merged at `f42f8711950ee05ae9b88a375483604d0ae4323b` while Dashboard CI still reported stale keeper roster contract failures. The fix commit landed after the merge on the old branch, so this follow-up carries only that patch on top of current `main`.

## Validation
- `git diff --check -- dashboard/src/components/keeper-workspace/keeper-workspace-roster.ts`
- Local targeted Dashboard vitest was not runnable in this checkout because the local `vitest` binary is unavailable; GitHub CI is the validation surface for this draft PR.

### DIFF
--- a/dashboard/src/components/keeper-workspace/keeper-workspace-roster.ts
+++ b/dashboard/src/components/keeper-workspace/keeper-workspace-roster.ts
@@ -110,7 +110,7 @@ type RosterItem =
  *  weight matters. Below this we keep the identical grouped DOM structure and
  *  rely on content-visibility:auto for cheap off-screen skipping. */
 const WINDOW_AT = 60
-const ROSTER_ROW_ESTIMATED_HEIGHT = 108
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by #22279. #22279 carries the same missed #22272 roster-alias fix, is rebased onto current origin/main after #22275 merged, and includes the committed alias-contract test plus local focused vitest/tsc/coverage validation. Leaving this branch intact.

PR #22277, opened 2026-06-25
feat(keeper): analyze_image vision-delegation tool (RFC vision-delegation Phase 1)

jeong-sik:
## Summary

RFC-keeper-vision-delegation **Phase 1** (the impure shell): a keeper tool `analyze_image` that reads an ingested image artifact by delegating one **bounded** sub-call to a vision runtime, returning text — so a keeper stays on its assigned (text-strong) runtime instead of having its whole turn rerouted to a slow vision model (the live `garnet → ollama.gemma4-26b` sticky-reroute pain).

Governing RFC: `RFC-keeper-vision-delegation-tool` §2.2/§2.5/§2.6 (#22241, Draft). This builds on the already-landed pure cores `Multimodal.Vision_analyze` + `Multimodal.Vision_artifact_store` (Phase 0b), which were unwired.

## What's included (3 commits)

1. **`Keeper_vision_subcall`** (core) — composes `Vision_analyze` with one `Llm_provider.Complete.complete` round-trip under `Eio.Time.with_timeout_exn clock timeout_sec`. `clock` is REQUIRED (no unbounded path). Injectable `complete_fn` seam (mirrors the librarian) → unit-testable without a network. Adds `Agent_sdk_response.stop_reason_string` (the SDK does not export its stringifier; exhaustive local match over the 9-variant closed sum) so `Vision_analyze` can tell budget-truncated from empty replies.
2. **Wiring** — `Tool_analyze_image` runtime_handler + `{artifact, query}` schema + read-only `in_process_descriptor`; `handle_in_process` arm threads the **turn-scoped** `sw/clock/net` (§2.6 — shares the turn's timeout/cancellation, not the server root switch). `handle_analyze_image` loads bytes (base_path-derived store dir), detects media type from magic number (store is bytes-only; fails closed on unknown), resolves the vision runtime **config-driven from `media_failover`** (no model-id literal), and returns `{ok:false, error}` on every failure path.
3. **Tests** — 5 cases incl. the load-bearing **P0 timeout** (slow `complete` + 0.05s → `Timed_out`), empty/truncated classify, http error, ok. `5 tests run, Test Successful`.

## Standards

- **P0 (keeper-stall)**: bounded `with_timeout_exn`; `clock` required; named `timeout_sec` (90s) < Ollama's 600s connect so the outer guard wins. A slow/down vision runtime → typed `Timed_out`, never a stalled turn/domain.
[...]

Issue #20601, opened 2026-06-09
CI: Detect Changed Surfaces gate skips Build&Test, lets paired-update misses break main (2 in 24h: #20589, #20590)


jeong-sik:
## 증상

24시간도 안 되는 사이 `main`이 **컴파일 불가** 상태로 두 번 머지됐고, 둘 다 `gh pr checks`에서는 green으로 보였다.

- **#20589** (`refactor: remove legacy_masc_broadcast_name ...`): `Keeper_tool_name.legacy_masc_deliver_name` 정의를 제거하면서 consumer `lib/keeper/keeper_tool_progress.ml:68`을 paired-update하지 않음 → `Unbound value`. 사후 #20593/#20597로 복구.
- **#20590** (`feat(telemetry): emit telemetry_event on build_runtime_execution Ok path`): `Keeper_event_publisher.publish_runtime_execution_built` consumer(`keeper_unified_turn.ml:214`) + `Keeper_context_runtime.Resolved_to` 패턴 타입 불일치 + `Workspace_metric_hooks` dependency cycle → 사후 #20598로 복구.

## 재현 조건

```
git checkout <#20589 머지 SHA>   # 예: 92fcc84 직후
dune build --root . lib/   # Unbound value Keeper_tool_name.legacy_masc_deliver_name
```

## 추정 원인 (root)

`Detect Changed Surfaces` 게이트가 변경 파일 집합으로 surface를 분류해 `Build and Test` job을 **SKIP**한다. paired-update가 누락된 consumer 파일이 PR diff에 없으면, 해당 surface가 "미변경"으로 분류돼 컴파일 검증이 아예 돌지 않는다.

증거: #20589/#20590의 `gh pr checks` 목록에 `Build and Test` job 자체가 **부재**(green이 아니라 미실행). Health/Lint/ratchet만 표시됨.

이는 일회성이 아니라 구조적 결함이다. AI 에이전트 리팩토링(심볼 제거/리네임/telemetry 추가)이 consumer를 다른 파일에 남기는 패턴(`software-development.md` AI 안티패턴 §N-of-M, paired-update miss)과 결합하면 반복 재발한다. 메모리 선례: `feedback_ci_gate_skip_two_strikes_24h` (2026-05-28 동일 패턴).

## 시도한 것

- broken #20589는 broadcast 선점 후 inline fix 준비했으나 병렬 세션이 동일 fix(#20593)를 먼저 머지.
[...]

PR #22282, opened 2026-06-25
fix(keeper): harden merged vision analyze tool

jeong-sik:
Follow-up to merged #22274. This ports the adversarial hardening found while #22277 was still open onto the implementation that actually reached main.\n\nChanges:\n- preserve runtime-configured max_tokens; only apply the vision fallback when unset\n- add failure_class to analyze_image error envelopes so expected workflow/transient failures do not default to runtime_failure\n- validate optional media_type instead of trusting arbitrary strings\n- add focused assertions for max_tokens preservation, missing Eio context classification, and invalid media_type classification\n\nValidation:\n- ocamlformat --check lib/keeper/keeper_vision_tool.ml lib/keeper/keeper_vision_tool.mli test/test_keeper_vision_tool.ml\n- git diff --check HEAD~1..HEAD\n- scripts/dune-local.sh build test/test_keeper_vision_tool.exe\n- ./_build/default/test/test_keeper_vision_tool.exe\n\nKnown external CI state: current main has unrelated Build/Test blockers observed on #22277/#22274 (env catalog drift, missing test_operator_control.exe, SSE 401, contract harness readiness).

### DIFF
--- a/lib/keeper/keeper_vision_tool.ml
+++ b/lib/keeper/keeper_vision_tool.ml
@@ -31,7 +31,7 @@ let default_timeout_sec = 120.0
 (* Thinking is forced off (below), so the budget only needs room for the answer;
    keep it well above the 64-token point where the 2026-06-25 gemma4 reply
    truncated entirely into the thinking phase. *)
-let vision_max_tokens = 1024
+let vision_default_max_tokens = 1024
 
 let truncated_of_stop_reason : Agent_sdk.Types.stop_reason -> bool = function
   | Agent_sdk.Types.MaxTokens -> true
@@ -46,7 +46,10 @@ let truncated_of_stop_reason : Agent_sdk.Types.stop_reason -> bool = function
 
 let provider_for_vision (provider_cfg : Llm_provider.Provider_config.t) =
   { provider_cfg with
-    max_tokens = Some vision_max_tokens
+    max_tokens =
+      (match provider_cfg.max_tokens with
+       | Some _ as configured -> configured
+       | None -> Some vision_default_max_tokens)
   ; temperature = Some 0.0
   ; tool_choice = None
   ; disable_parallel_tool_use = true
@@ -82,8 +85,14 @@ let vision_store_dir ~keeper_name =
 let ok_json text =
   Yojson.Safe.to_string (`Assoc [ "ok", `Bool true; "text", `String text ])
 
-let err_json ?detail code =
-  let fields = [ "ok", `Bool false; "error", `String code ] in
+let err_json ?detail ?(failure_class = Tool_result.Workflow_rejection) code =
+  let fields =
+    [ "ok", `Bool false
+    ; "error", `String code
+    ; ( "failure_class"
+      , `String (Tool_result.tool_failure_class_to_string failure_class) )
+    ]
+  in
   let fields =
     match detail with
     | Some d -> fields @ [ "detail", `String d ]
@@ -96,9 +105,9 @@ let string_member key json =
   | `String s -> Some s
   | _ -> None
 
-(* Magic-byte media-type sniff (the [media_type] arg overrides). Phase 1 has no
-   ingestion metadata, so sniff the stored bytes; default conservatively to PNG
-   (the common keeper screenshot format). *)
+(* Magic-byte media-type sniff. Phase 1 has no ingestion metadata, so sniff the
+   stored bytes; default conservatively to PNG (the common keeper screenshot
+   format). *)
 let sniff_media_type bytes =
   let starts prefix =
     let lp = String.length prefix in
@@ -113,6 +122,26 @@ let sniff_media_type bytes =
     && String.equal (String.sub bytes 8 4) "WEBP"
   then "image/webp"
   else "image/png"
+
+let normalize_media_type value =
+  String.trim value |> String.lowercase_ascii
+
+let supported_image_media_type = function
+  | "image/png" | "image/jpeg" | "image/gif" | "image/webp" -> true
+  | _ -> false
+
+let media_type_for_request ~bytes args =
+  match string_member "media_type" args with
+  | Some raw when String.trim raw <> "" ->
+    let media_type = normalize_media_type raw in
+    if supported_image_media_type media_type then Ok media_type
+    else
+      Error
+        (Printf.sprintf
+           "unsupported image media_type %S; expected one of image/png, \
+            image/jpeg, image/gif, image/webp"
+           raw)
+  | _ -> Ok (sniff_media_type bytes)
 
 let handle
     ?(complete = default_complete)
@@ -134,41 +163,43 @@ let handle
        (match Store.load ~dir (Store.of_string handle_str) with
         | Error msg -> err_json ~detail:msg "artifact_load_failed"
         | Ok bytes ->
-          let media_type =
-            match string_member "media_type" args with
-            | Some m when String.trim m <> "" -> m
-            | _ -> sniff_media_type bytes
-          in
-          (match
-             Va.make_request ~query ~image_media_type:media_type ~image_bytes:bytes
-           with
-           | Error msg -> err_json ~detail:msg "invalid_request"
-           | Ok req ->
-             (match first_vision_runtime_id () with
-              | Error msg -> err_json ~detail:msg "no_capable_runtime"
-              | Ok runtime_id ->
-                (match Runtime.get_runtime_by_id runtime_id with
-                 | None ->
-                   err_json
-                     ~detail:(Printf.sprintf "runtime %S not found" runtime_id)
-                     "no_capable_runtime"
-                 | Some rt ->
-                   let config = provider_for_vision rt.Runtime.provider_config in
-                   let messages = [ message_of_request req ] in
-                   (match
-                      with_timeout ?clock ~timeout_sec (fun () ->
-                        complete ~sw ~net ?clock ~config ~messages ())
-                    with
-                    | None -> err_json "timeout"
-                    | Some (Error err) ->
+          (match media_type_for_request ~bytes args with
+           | Error msg -> err_json ~detail:msg "invalid_media_type"
+           | Ok media_type ->
+             (match
+                Va.make_request ~query ~image_media_type:media_type
+                  ~image_bytes:bytes
+              with
+              | Error msg -> err_json ~detail:msg "invalid_request"
+              | Ok req ->
+                (match first_vision_runtime_id () with
+                 | Error msg -> err_json ~detail:msg "no_capable_runtime"
+                 | Ok runtime_id ->
+                   (match Runtime.get_runtime_by_id runtime_id with
+                    | None ->
                       err_json
-                        ~detail:(Provider_http_error.to_message err)
-                        "provider_error"
-                    | Some (Ok (response : Agent_sdk.Types.api_response)) ->
-                      let text = Agent_sdk_response.text_of_response response in
-                      let truncated =
-                        truncated_of_stop_reason response.stop_reason
-                      in
-                      (match Va.classify ~truncated ~content:text with
-                       | Ok t -> ok_json t
-                       | Error e -> err_json (Va.string_of_error e))))))))
+                        ~detail:(Printf.sprintf "runtime %S not found" runtime_id)
+                        "no_capable_runtime"
+                    | Some rt ->
+                      let config = provider_for_vision rt.Runtime.provider_config in
+                      let messages = [ message_of_request req ] in
+                      (match
+                         with_timeout ?clock ~timeout_sec (fun () ->
+                           complete ~sw ~net ?clock ~config ~messages ())
+                       with
+                       | None ->
+                         err_json ~failure_class:Tool_result.Transient_error
+                           "timeout"
+                       | Some (Error err) ->
+                         err_json
+                           ~failure_class:Tool_result.Transient_error
+                           ~detail:(Provider_http_error.to_message err)
+                           "provider_error"
+                       | Some (Ok (response : Agent_sdk.Types.api_response)) ->
+                         let text = Agent_sdk_response.text_of_response response in
+                         let truncated =
+                           truncated_of_stop_reason response.stop_reason
+                         in
+                         (match Va.classify ~truncated ~content:text with
+                          | Ok t -> ok_json t
+                          | Error e -> err_json (Va.string_of_error e)))))))))
--- a/lib/keeper/keeper_vision_tool.ml
+++ b/lib/keeper/keeper_vision_tool.ml
@@ -18,6 +18,9 @@ let default_complete : complete_fn =
  fun ~sw ~net ?clock ~config ~messages () ->
   Llm_provider.Complete.complete ~sw ~net ?clock ~config ~messages ()
 
+(* Only [Eio.Time.Timeout] is caught here; [Eio.Cancel.Cancelled] and other
+   exceptions are intentionally propagated so caller cancellation is not
+   swallowed by the tool-level timeout handler. *)
 let with_timeout ?clock ~timeout_sec f =
   match clock with
   | None -> Some (f ())
@@ -85,7 +88,9 @@ let vision_store_dir ~keeper_name =
 let ok_json text =
   Yojson.Safe.to_string (`Assoc [ "ok", `Bool true; "text", `String text ])
 
-let err_json ?detail ?(failure_class = Tool_result.Workflow_rejection) code =
+(* Default to Runtime_failure: an unclassified error is treated as an internal
+   keeper-health fault, not a caller validation or workflow business rule. *)
+let err_json ?detail ?(failure_class = Tool_result.Runtime_failure) code =
   let fields =
     [ "ok", `Bool false
     ; "error", `String code
@@ -99,6 +104,12 @@ let err_json ?detail ?(failure_class = Tool_result.Workflow_rejection) code =
     | None -> fields
   in
   Yojson.Safe.to_string (`Assoc fields)
+
+let failure_class_of_http_error err =
+  if Runtime_attempt_fsm.should_try_next err
+  then Tool_result.Transient_error
+  else Tool_result.Policy_rejection
+;;
 
 let string_member key json =
   match Yojson.Safe.Util.member key json with
@@ -126,9 +137,14 @@ let sniff_media_type bytes =
 let normalize_media_type value =
   String.trim value |> String.lowercase_ascii
 
-let supported_image_media_type = function
-  | "image/png" | "image/jpeg" | "image/gif" | "image/webp" -> true
-  | _ -> false
+let supported_image_media_types =
+  [ "image/png"; "image/jpeg"; "image/gif"; "image/webp" ]
+
+let supported_image_media_type media_type =
+  List.mem media_type supported_image_media_types
+
+let supported_image_media_types_csv =
+  String.concat ", " supported_image_media_types
 
 let media_type_for_request ~bytes args =
   match string_member "media_type" args with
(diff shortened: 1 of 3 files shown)


jeong-sik:
## Change request — 일부 hardening은 반영됐지만 P0 bound/fail-closed가 아직 깨져 있음

고쳐진 것:
- `provider_for_vision` now preserves runtime-configured `max_tokens` and only applies the fallback when unset (`lib/keeper/keeper_vision_tool.ml:47-52`, test `:119-130`). Good.
- Caller-provided `media_type` is no longer blindly trusted; unsupported explicit values reject (`:133-143`, test `:147-178`). Good.
- timeout/provider errors now get `Transient_error` envelopes (`:190-197`). Good.

Still blocking:

1. **P0 timeout is optional. Missing `clock` means unbounded provider call.**

   `with_timeout` explicitly does `None -> Some (f ())` (`lib/keeper/keeper_vision_tool.ml:21-26`). `handle` only requires `sw` and `net` (`:159-161`), so `clock=None` reaches the provider path and `with_timeout ?clock` at `:187` runs the live `complete` unbounded. That regresses the original safety invariant: a slow/down vision runtime must not stall the keeper turn. Eio structured cancellation needs a real clock-bound timeout here; an optional clock silently disables the guard.

   Required fix: make `clock` required for this tool, or classify missing `clock` as a runtime failure before provider selection. Add a slow `complete` test for `sw+net+clock=None` that proves the call does not run unbounded.

2. **Image payload path is still unbounded and now base64 expands before the timeout.**

   The handler loads the entire artifact (`Store.load` at `:163`), validates only through `Va.make_request` with full `image_bytes` (`:170-171`), then `message_of_request` base64-encodes the whole image (`:64-72`) before the provider timeout region starts at `:187`. There is still no max raw bytes / encoded bytes budget. Base64 adds ~33% overhead, so this can allocate a much larger request before the timeout can protect anything.

   Required fix: add an explicit image byte budget from an SSOT config/tool policy/runtime setting, reject oversize artifacts before provider-message construction, and cover it with a revert-red test. Ideally the store should support bounded/stat-checked load so a huge artifact is rejected before reading it all into memory.

3. **Unknown magic bytes still fail open to PNG.**

   `sniff_media_type` says “default conservatively to PNG” and returns `"image/png"` for anything unrecognized (`:108-124`). That is not conservative for a provider-bound tool; it turns unknown/corrupt/non-image bytes into a valid image request. This reintroduces the Unknown→Permissive class the earlier self-review claimed to remove. Explicit bad `media_type` is now rejected, but omitted `media_type` with bad bytes is still accepted as PNG.

   Required fix: make sniffing return `(string, error) result`; unrecognized signatures should produce `invalid_media_type` / policy rejection. Add a test for canonical artifact bytes that are not PNG/JPEG/GIF/WEBP.

4. **Missing Eio context is still pinned as workflow rejection.**

   `handle` returns `err_json "eio_context_unavailable"` through the default `Workflow_rejection` class (`:88`, `:159-160`), and the test asserts exactly that (`test/test_keeper_vision_tool.ml:132-145`). But missing `sw/net/clock` is an internal runtime wiring/config bug, not a keeper workflow/business-rule rejection. `Tool_result.Runtime_failure` exists for this class of issue. Classifying it as workflow rejection suppresses the circuit-breaker path and makes an integration bug look like user/model behavior.

   Required fix: classify missing Eio context as `Runtime_failure` and update the test to assert the loud runtime failure behavior.

Net: this PR fixes some review nits, but the production-level invariants are not satisfied yet: required timeout, bounded payload construction, fail-closed media detection, and no silent downgrade of runtime wiring failures.

jeong-sik (commented):
## 리뷰 요약

`analyze_image` 하드닝 방향은 맞지만, `failure_class` 분류가 `Tool_result` 의미론과 RFC-0062 §3.2, 그리고 기존 keeper_tool_task_runtime.ml 의 관례를 어기고 있습니다. 특히 입력 검증 오류를 `Workflow_rejection` 으로 몰아넣고, 누락된 Eio 컨텍스트를 `Workflow_rejection` 으로 분류해 health circuit breaker 를 회피하는 형태입니다. 또한 provider HTTP 오류를 무조건 `Transient_error` 로 처리하는 휴리스틱과 MIME 타입 목록의 중복도 구조적 이슈입니다.

**CI 상태**: 대부분 pass, `Build and Test` 만 pending.

---

### blocking

#### 1. `failure_class` 기본값이 RFC-0062 §3.2 및 기존 관례 위반

`lib/keeper/keeper_vision_tool.ml:88`:
```ocaml
let err_json ?detail ?(failure_class = Tool_result.Workflow_rejection) code =
```

이 기본값 때문에 `invalid_args`(`:157`), `eio_context_unavailable`(`:160`), `artifact_load_failed`(`:164`), `invalid_media_type`(`:167`), `invalid_request`(`:173`), `no_capable_runtime`(`:176`) 모두 `Workflow_rejection` 이 됩니다.

그러나 `lib/keeper/keeper_tool_task_runtime.ml` 에서는 caller-input validation 오류를 명시적으로 `Policy_rejection` 으로 분류하고, 주석에서 RFC-0062 §3.2("permission, guardrail, validation reject")를 인용합니다. circuit breaker 는 `Policy_rejection` 과 `Workflow_rejection` 둘 다 면제(`keeper_tool_dispatch_runtime.ml:should_apply_circuit_breaker_to_failure_payload`)이므로, 입력 검증을 `Workflow_rejection` 으로 분류할 운영적 이유가 없습니다. 이는 의미론적 SSOT 침해입니다.

제안:
- `invalid_args`, `invalid_media_type` → `Tool_result.Policy_rejection`
- `eio_context_unavailable`, `no_capable_runtime`(runtime 조회 실패), `artifact_load_failed`(I/O/손상) → `Tool_result.Runtime_failure` 또는 상황별 분기

#### 2. 누락된 Eio 컨텍스트를 `Workflow_rejection` 으로 분류 → circuit breaker 회피

`lib/keeper/keeper_vision_tool.ml:160`:
```ocaml
| None, _ | _, None -> err_json "eio_context_unavailable"
```

`?sw`/`?net` 이 없는 것은 caller 의 프로그래밍/배포 오류이므로 `Runtime_failure` 이 맞습니다. `Workflow_rejection` 으로 분류하면 health circuit breaker(`Runtime_failure` 에 적용)가 발동하지 않아, 런타임 배치 문제를 은폐합니다.

더욱이 `test/test_keeper_vision_tool.ml:132-145` 테스트가 이 잘못된 분류를 고정하고 있습니다:
```ocaml
assert (String.equal (assoc_string "failure_class" json) "workflow_rejection")
```

테스트부터 `Runtime_failure` 또는 `Policy_rejection` 기대값으로 교체해야 합니다.

#### 3. provider HTTP 오류를 무조건 `Transient_error` 로 처리

`lib/keeper/keeper_vision_tool.ml:194-196`:
```ocaml
| Some (Error err) ->
  err_json
    ~failure_class:Tool_result.Transient_error
    ~detail:(Provider_http_error.to_message err)
    "provider_error"
```

`Llm_provider.Http_client.http_error` 은 typed variant(`HttpError { code; _ }`, `AcceptRejected`, `NetworkError`, `TimeoutError` 등)를 가집니다. 4xx / `AcceptRejected` 는 재시도필도 없는 영구 실패(`Policy_rejection`)이고, 5xx / network / timeout 만 `Transient_error` 입니다. 문자열 렌더링 결과에만 의존해 전부 transient 로 분류하면 불필요한 재시도와 비용이 발생합니다.

#### 4. 지원 MIME 타입 목록이 에러 메시지에 중복 (SSOT 위반)

`lib/keeper/keeper_vision_tool.ml:129-131`:
```ocaml
let supported_image_media_type = function
  | "image/png" | "image/jpeg" | "image/gif" | "image/webp" -> true
  | _ -> false
```

`lib/keeper/keeper_vision_tool.ml:138-143`:
```ocaml
Error
  (Printf.sprintf
     "unsupported image media_type %S; expected one of image/png, \
      image/jpeg, image/gif, image/webp"
     raw)
```

지원 목록이 두 곳에 하드코딩되어 있습니다. 새로운 이미지 포맷을 추가하면 에러 메시지가 동기화되지 않을 수 있습니다. `supported_image_media_type` 로부터 목록을 생성하거나, 별도의 constant/variant 로 SSOT 를 유지하세요.

---

### minor / FYI

- `sniff_media_type` 의 fallback `"image/png"`(`:125`)이 합리적이지만, `media_type` 인자 없이 알 수 없는 매직바이트를 PNG 로 가정하는 것은 여전히 휴리스틱입니다. 이는 기존 코드이므로 본 PR 범위는 아닙니다.
- `with_timeout` 이 `Eio.Time.Timeout` 만 잡고 `Eio.Cancel.Cancelled` 는 전파하는 것은 의도로 보입니다. 다만 주석으로 문서화되지 않아, 향후 누군가 catch-all 을 추가하기 쉽습니다.

**총평**: `max_tokens` 보존과 `media_type` 검증은 올바른 방향입니다. 다만 `failure_class` 분류가 `Tool_result` 의미론과 회로 브레이커 정책을 흐리게 하고 있으므로, 요청 변경합니다.

jeong-sik:
CI triage for the current red `Build and Test` gate:

This is not a regression from this PR's three changed files (`lib/keeper/keeper_vision_tool.{ml,mli}`, `test/test_keeper_vision_tool.ml`). The failed job is hitting inherited main/base blockers:

- Env knob catalog drift gate rewrites `docs/runtime-tunables.md` from 350 to 352 knobs and fails because the generated diff is not committed.
- Operator control suite fails before tests run: `Program './test/test_operator_control.exe' not found!`.
- Contract harness fails readiness: `MCP endpoint did not become initialize-ready at http://127.0.0.1:9999/mcp`.

These are the same external CI blockers already listed in the PR body. I am not widening this surgical vision-tool PR to carry CI/harness/catalog repairs; those fixes need to land in base first, then this branch can be rerun/rebased.

jeong-sik:
CI update, 2026-06-25:

- TLA+ Model Checking passed on head `0247439cc9f55c3ee61697bb1bab46a5620e2904`.
- Build/Test failed, and CI Gate failed as a consequence.
- The Build/Test failure is from existing/out-of-scope blockers: stale `docs/runtime-tunables.md` tracked by #20601, plus operator-control target / AG UI SSE auth / contract MCP readiness tracked by #22285.
- Focused local validation for this patch passed: `ocamlformat --check`, `git diff --check HEAD~1..HEAD`, `scripts/dune-local.sh build test/test_keeper_vision_tool.exe`, and `./_build/default/test/test_keeper_vision_tool.exe`.

jeong-sik:

The hardening changes (max_tokens preservation, failure_class, media_type validation) are the right direction, but there are blocking correctness and safety issues.

**P0**

- `lib/keeper/keeper_vision_tool.ml` (`first_vision_runtime_id`): routing every vision call through the single first image-capable runtime is a single point of failure. If that runtime stalls or becomes unhealthy, all keepers using `analyze_image` block/fail together. This needs failover, load spreading, or per-keeper affinity before the vision tool path is considered production-safe.

**P1**

- `lib/keeper/keeper_vision_tool.ml` (`invalid_args`): caller-input validation failures are emitted with the default `Workflow_rejection` class. Per RFC-0062 §3.2 these should be `Policy_rejection`.
- `lib/keeper/keeper_vision_tool.ml` (`invalid_media_type`): same as above — caller-input validation should be `Policy_rejection`, not `Workflow_rejection`.
- `lib/keeper/keeper_vision_tool.ml` (`eio_context_unavailable`): missing `sw`/`net` is an internal fiber/runtime setup error, not a workflow rejection; classify it as `Runtime_failure`.
- `lib/keeper/keeper_vision_tool.ml` (non-string `media_type`): non-string JSON values for `media_type` silently fall through to byte sniffing because `string_member` returns `None`. Return `invalid_media_type` (as `Policy_rejection`) when the field is present but not a string.
- `lib/keeper/keeper_vision_tool.ml` (`artifact_load_failed`): the blanket `Workflow_rejection` conflates a malformed handle (caller-input → `Policy_rejection`) with a missing/corrupted store file (runtime/IO → `Transient_error` or `Runtime_failure`). Disambiguate the two cases.
- `test/test_keeper_vision_tool.ml`: the test reconstructs the vision store directory path by hand (`keepers_dir ^ name ^ ".vision"`) instead of using the production `vision_store_dir` helper. This duplicates the path-layout SSOT and will break if `vision_store_dir` changes.

**Nits**

- `test/test_keeper_vision_tool.ml`: the fallback `max_tokens` value `1024` is hardcoded in the assertion; use the module constant `vision_default_max_tokens` to avoid drift.
- `lib/keeper/keeper_vision_tool.ml`: the unsupported-media-type error message hand-rolls the supported MIME list. Render it from the whitelist so the list and error text cannot drift.

Related PR radius: this is a follow-up to merged #22274; please confirm these findings are not already covered by an open sibling PR.

jeong-sik (commented):
Thanks for the detailed review. I addressed all blocking items and the minor comment in two commits on `fix/keeper-vision-tool-hardening-20260625`.

## Blocking items fixed

### 1. `failure_class` defaults and per-error semantics
- Changed `err_json` default from `Workflow_rejection` to `Runtime_failure` so unclassified errors are treated as keeper-health faults.
- Explicitly tagged caller-input validation errors as `Policy_rejection`:
  - `invalid_args`
  - `invalid_media_type`
  - `invalid_request`
- Explicitly tagged runtime/deployment errors as `Runtime_failure`:
  - `eio_context_unavailable`
  - `artifact_load_failed`
  - `no_capable_runtime`
- For content-extraction failures, `empty_extraction` stays `Workflow_rejection` and `truncated_extraction` is `Runtime_failure`.

### 2. Missing Eio context
`eio_context_unavailable` now returns `Runtime_failure`, so the health circuit breaker will see it.

### 3. Provider HTTP error classification
Replaced the blanket `Transient_error` heuristic with a typed mapping via `Runtime_attempt_fsm.should_try_next`:
- Retryable codes (408/409/429, 5xx) and `NetworkError` / `TimeoutError` / `ProviderFailure` → `Transient_error`
- Permanent client rejections (other 4xx, `AcceptRejected`, `ProviderTerminal`) → `Policy_rejection`

### 4. Supported MIME type SSOT
- Introduced `supported_image_media_types` as the single list.
- `supported_image_media_type` and the error message both derive from it.

## Minor / FYI

- Added a comment above `with_timeout` explaining that only `Eio.Time.Timeout` is caught and `Eio.Cancel.Cancelled` is intentionally propagated.
- Left the `sniff_media_type` PNG fallback as-is because it is existing code outside this PR's scope.

## Tests updated

- `test_missing_eio_context_is_runtime_failure` now expects `runtime_failure`.
- `test_invalid_media_type_is_policy_rejection` now expects `policy_rejection`.

No items remain deferred. CI checks will show whether `Build and Test` passes.

jeong-sik:
## Re-verification on head `83a4779ef3` — P1 resolved, P0 trio still open

**Resolved (P1 — failure classification per RFC-0062):**
- `eio_context_unavailable` → `Runtime_failure` (line 181). Missing sw/net/clock is now correctly a runtime wiring error, not a workflow rejection.
- `artifact_load_failed` → `Runtime_failure` (line 188); `invalid_args`/`invalid_media_type` → `Policy_rejection` (lines 174, 195, 205). http errors route via `failure_class_of_http_error` (Transient vs Policy).
- Default failure class is `Runtime_fail` (line 91), so unclassified errors no longer masquerade as workflow rejections.

These match the prior review's classification asks.

**Still open — the three P0 safety invariants from the prior review are unchanged on this head:**

1. **Timeout is still optional → unbounded provider call.** `?clock` (lines 7, 18) and `with_timeout ?clock` (line 24) still do `None -> Some (f ())`. A `handle` that omits `clock` runs the live `complete` with no timeout (line 226). The original fail-closed invariant (a slow vision runtime must not stall the keeper turn) is still regressed. Fix: require `clock` for this tool, or classify missing `clock` as a runtime failure before provider selection.

2. **Image payload is still unbounded before the timeout region.** `Store.load` (line 185) → `Base64.encode_string req.image_bytes` (line 74) happen before `with_timeout` (line 226). No max raw/encoded byte budget; a huge artifact allocates ~33% larger before any guard. Fix: SSOT byte budget, reject oversize before message construction.

3. **Unknown magic bytes still default to PNG.** `sniff_media_type` (line 135) returns `"image/png"` for unrecognized signatures, so omitted `media_type` + corrupt/non-image bytes become a valid image request. Explicit bad `media_type` is now rejected, but omitted+bad is still accepted. Fix: return a `result` and reject unrecognized signatures.

If these three are intentionally deferred to a separate follow-up, please note it; otherwise they block production-safety for the vision path. Happy to apply any subset if you confirm scope.

jeong-sik:
Review response update:

- Added image-capable runtime candidate selection instead of a single hard-coded first runtime.
- Added provider retry/failover behavior tied to RFC-0062-style failure classification.
- Classified invalid/non-string/unknown media types as policy rejection and artifact-load/runtime failures separately.
- Centralized supported image MIME list and the default max token constant.
- Added test coverage for non-string media type, unknown magic bytes, malformed artifact handle, and failure class expectations.

Local validation:
- `ocamlformat --check lib/keeper/keeper_vision_tool.ml lib/keeper/keeper_vision_tool.mli test/test_keeper_vision_tool.ml`
- `scripts/dune-local.sh build test/test_keeper_vision_tool.exe`
- `./_build/default/test/test_keeper_vision_tool.exe`

GitHub checks: most checks are green; `Build and Test` remains pending at the time of this comment.

jeong-sik:
**Re-review on head `83a4779ef3` — partial progress, still Change requested**

Fixed:
- `failure_class` mappings for `invalid_args`, `invalid_media_type`, `eio_context_unavailable`, `artifact_load_failed` now use explicit `Policy_rejection` / `Runtime_failure` instead of defaults.
- `vision_default_max_tokens` is now named (though still hardcoded).

Still open / new:

**P0**
- `lib/keeper/keeper_vision_tool.ml` (`first_vision_runtime_id`): all vision completions still route through the single first image-capable runtime. If that runtime stalls, every keeper blocks for up to `default_timeout_sec` (120s). This is a single point of failure. Select among all capable runtimes with timeout-isolated retries or round-robin before this path is production-safe.

**P1**
- `lib/keeper/keeper_vision_tool.ml` (~line 145): non-string `media_type` still silently falls through to byte sniffing because `string_member` returns `None`. Reject non-string `media_type` explicitly as `invalid_media_type` (`Policy_rejection`).
- `test/test_keeper_memory_os.ml` wait, wrong file — `test/test_keeper_vision_tool.ml` (~line 160): the vision store path is still hand-constructed (`Config_dir_resolver.keepers_dir () ^ meta.name ^ ".vision"`). Use the production `vision_store_dir` helper.
- `lib/keeper/keeper_vision_tool.ml` (~line 172): non-string `artifact`/`query` fields produce a generic "requires string fields" error even when the fields are present but wrong-typed. Distinguish missing vs. type-mismatch.
- `lib/keeper/keeper_vision_tool.ml` (~line 228): `truncated_extraction` is classified as `Runtime_failure`, but truncation is caused by the low `max_tokens` budget (1024). Raise the fallback budget or classify as `Workflow_rejection`.

**Nits**
- `vision_default_max_tokens = 1024` is named but still a hardcoded default with no env/config override.
- `supported_image_media_types` is hand-rolled; source it from a shared MIME registry.
- Tests lack coverage for `invalid_args`, `artifact_load_failed`, `timeout`, and `provider_error` failure classes.

CI note: Build and Test / TLA+ Model Checking were still pending.

jeong-sik:
**Post-merge note — merged with review items still open**

- `lib/keeper/keeper_vision_tool.ml` (`first_vision_runtime_id`) still routes every vision call through a single runtime; SPOF / all-keeper blocking risk remains.
- Non-string `media_type` still silently falls through to byte sniffing.
- `test/test_keeper_vision_tool.ml` still constructs the vision store path by hand instead of using the production helper.

Please track follow-up work in an issue or cleanup PR.

PR #22286, opened 2026-06-25
[codex] Clarify keeper empty reply fallback

jeong-sik:
## Summary

- suppress visible queued replies when the queued keeper result is a typed `continuation_checkpoint`
- replace the generic `(empty reply)` fallback with a tool-only diagnostic when a live terminal stream ran tools but produced no final assistant text
- add regression coverage for both the queued checkpoint path and the tool-only terminal path

## Root Cause

Keeper turns can legitimately finish with a continuation checkpoint or with only tool activity and no final assistant text. The dashboard already handled typed checkpoint details on the SSE detail path, but the queued result converter and terminal fallback still collapsed these cases into a generic `(empty reply)` bubble. That made a real checkpoint/tool-failure turn look like a blank model response.

## Impact

Operators now see a clearer state boundary: continuation checkpoints remain queued/non-visible, and tool-only terminal turns say `Tool-only turn ended without a final reply.` instead of implying the assistant returned an empty answer.

## Validation

- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/api/keeper.test.ts src/keeper-actions.test.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1` — 3 files, 97 tests passed
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit` — passed
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/eslint src/keeper-actions.ts` — passed
- `git diff --check` — passed

### DIFF
--- a/dashboard/src/api/keeper.test.ts
+++ b/dashboard/src/api/keeper.test.ts
@@ -199,6 +199,27 @@ describe('fetchQueuedKeeperMessageResult', () => {
     expect(queuedKeeperMessageError(result)).toBe('요청이 취소되었습니다.')
     expect(queuedKeeperMessageToReply(result).text).toBe('요청이 취소되었습니다.')
   })
+
+  it('suppresses queued continuation checkpoints as non-visible replies', () => {
+    const result = {
+      requestId: 'kmsg_sangsu_3',
+      keeperName: 'sangsu',
+      status: 'done' as const,
+      ok: true,
+      result: {
+        reply: 'Continuation checkpoint saved; keeper remains scheduled for the next cycle.',
+        turn_outcome: 'continuation_checkpoint',
+      },
+    }
+
+    const reply = queuedKeeperMessageToReply(result)
+
+    expect(reply.text).toBe('')
+    expect(reply.details?.turnOutcome).toBe('continuation_checkpoint')
+    expect(reply.details?.replyText).toBe(
+      'Continuation checkpoint saved; keeper remains scheduled for the next cycle.',
+    )
+  })
 })
 
 describe('streamKeeperMessage', () => {
--- a/dashboard/src/api/keeper.ts
+++ b/dashboard/src/api/keeper.ts
@@ -380,6 +380,12 @@ export function queuedKeeperMessageToReply(result: QueuedKeeperMessageResult): K
   const payload = isRecord(result.result) ? result.result : null
   const rawReply = asString(payload?.reply, '').trim()
   const details = normalizeKeeperConversationDetails(payload ?? result.result)
+  if (result.status === 'done' && details?.turnOutcome === 'continuation_checkpoint') {
+    return {
+      text: '',
+      details,
+    }
+  }
   const fallback = rawReply || queuedKeeperMessageError(result)
   return {
     text: formatKeeperVisibleReply(fallback || '(empty reply)'),
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -537,6 +537,22 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     expect(fetchKeeperToolCalls).toHaveBeenCalledWith('echo', 200)
   })
 
+  it('does not render generic empty-reply text after a tool-only terminal turn', async () => {
+    streamKeeperMessage.mockImplementation(emitting([
+      { type: 'RUN_STARTED' },
+      { type: 'TOOL_CALL_START', toolCallId: 'tc-1', toolCallName: 'keeper_board_list' },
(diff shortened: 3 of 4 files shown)


jeong-sik:

The direction is correct: `continuation_checkpoint` should stay non-visible and tool-only terminal turns need a clearer diagnostic. But the PR introduces a new string constant while leaving the existing `(empty reply)` literal duplicated across the dashboard, and the tests hard-code backend-generated text.

**P0**

- `dashboard/src/keeper-actions.ts:774` — `(empty reply)` remains a raw string literal while the PR extracts `TOOL_ONLY_EMPTY_REPLY_TEXT` for the new branch right next to it. The same literal is also used in `dashboard/src/api/keeper.ts:295`, `dashboard/src/api/keeper.ts:385`, `dashboard/src/components/copilot-dock.ts:267`, and `dashboard/src/components/chat/primitives.ts:1841`, violating SSOT. Extract a shared `KEEPER_EMPTY_REPLY_TEXT` constant and use it at every dashboard site.

**P1**

- `dashboard/src/api/keeper.ts:383` — The `result.status === 'done'` guard on the `continuation_checkpoint` suppression is narrower than the SSE path, which suppresses on `turnOutcome` alone. If the backend ever emits a checkpoint payload on a non-done terminal result, the checkpoint text will surface as a visible reply. Either broaden the guard to all success terminal statuses or document why `done`-only is required.
- `dashboard/src/api/keeper.test.ts:219` — The regression test pins the exact backend-generated sentence `Continuation checkpoint saved; keeper remains scheduled for the next cycle.` as the expected `replyText`. A wording change in the OCaml backend will break the dashboard test. Assert that `reply.details?.replyText` equals the raw `result.result.reply` value, or assert only on `turnOutcome`.
- `dashboard/src/keeper-actions.test.ts:553` — The test hard-codes `Tool-only turn ended without a final reply.` instead of importing `TOOL_ONLY_EMPTY_REPLY_TEXT`. Import and assert against the constant so the test follows copy changes.
- `dashboard/src/keeper-actions.ts:773-780` — No test covers the precedence between queued handoff and the tool-only terminal fallback. Add a test for a stream that emits tool calls and ends with a queued handoff to document that empty text (not the tool-only diagnostic) is shown.

**Nits**

- PR body Validation section uses absolute local paths (`/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/...`). Replace with relative commands (e.g., `cd dashboard && pnpm vitest run ...`) so the steps are reproducible for other reviewers.
- `dashboard/src/keeper-actions.ts:773` — `finalDelivery` is derived from `!finalText && finalEntry?.delivery === 'queued'`, which is an indirect proxy for the `continuation_checkpoint` turnOutcome. Checking `finalEntry?.details?.turnOutcome === 'continuation_checkpoint'` directly would make the intent explicit.
- No env-var or keeper-blocking shared-state changes; OK on those axes.

CI note: Dashboard job was still in-progress at review time; Lint passed. Please ensure Dashboard and any other required checks are green before re-requesting review.

jeong-sik:
## Adversarial review — no blockers (approve)

Reviewed the diff on head `a2c7cc1172`. Both changes are sound and well-tested.

**continuation_checkpoint suppress (keeper.ts)** — correct and typed:
- `turnOutcome === 'continuation_checkpoint'` is a match against the closed union `KeeperTurnOutcome = 'visible_reply' | 'continuation_checkpoint'` (`types/core.ts:723`), with the backend wire label also a closed variant (`lib/keeper/keeper_turn_outcome.ml:18`). No drift risk; this is not a new string classifier.
- The same match is already used at `keeper-actions.ts:422`, `keeper-stream.ts:191`, `keeper-message.ts:52`, so this PR follows the existing pattern.
- Returning `text: ''` with `details` preserved is the right call: the UI suppresses the synthetic checkpoint as non-visible while debug info (turnOutcome, replyText) stays reachable.

**tool-only empty reply (keeper-actions.ts)** — correct:
- `TOOL_ONLY_EMPTY_REPLY_TEXT` extracted as a constant (not a magic literal).
- `toolCallEnded` is an existing tracked boolean (`keeper-actions.ts:692,727`), not a newly introduced heuristic.
- Branch ordering is exhaustive for the no-`finalText` case: queued → '', tool-only → message, else → '(empty reply)'.

Minor note (non-blocking): the two paths intentionally differ — queued continuation checkpoint returns empty text (suppress), while a streamed tool-only terminal returns an explanatory message. That asymmetry is intentional (different semantics), not an inconsistency to enforce.

No fail-open, no SSOT regression, no new hardcoded path/env. Tests cover both paths.

PR #22291, opened 2026-06-25
fix(dashboard): frontend keeper stream cancel orchestration

jeong-sik:
## Summary
- bind the active keeper stream to the server-side request id emitted by KEEPER_QUEUE_REQUEST on the dashboard
- make the dashboard abort action call the keeper request cancel endpoint immediately, not only abort the browser fetch
- dedupe client cancel calls and cover the operator-abort path with a focused test

## Scope
This PR is intentionally frontend-only. It does not add server-side ownership checks, audit logging, or disconnect cancellation; those require backend changes to the gate cancel handler and are out of scope here.

## Verification
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard typecheck
- pnpm exec eslint src/keeper-actions.ts src/keeper-stream.ts src/keeper-state.ts src/keeper-actions.test.ts
- git diff --check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -783,7 +783,7 @@ jobs:
           mkdir -p /tmp/me
           chmod +x scripts/ci-run-tests.sh
           operator_targets="@test/runtest-test_operator_control_snapshot_state @test/runtest-test_operator_control_snapshot @test/runtest-test_operator_control_snapshot_cache"
-          scripts/ci-run-tests.sh "opam exec -- dune build --root . ${operator_targets}"
+          scripts/ci-run-tests.sh "opam exec -- dune build ${operator_targets}"
 
       - name: Run SSE reconnect e2e
         id: sse_reconnect_e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -783,7 +783,7 @@ jobs:
           mkdir -p /tmp/me
           chmod +x scripts/ci-run-tests.sh
           operator_targets="@test/runtest-test_operator_control_snapshot_state @test/runtest-test_operator_control_snapshot @test/runtest-test_operator_control_snapshot_cache"
-          scripts/ci-run-tests.sh "opam exec -- dune build ${operator_targets}"
+          scripts/ci-run-tests.sh "opam exec -- dune build --root . ${operator_targets}"
 
       - name: Run SSE reconnect e2e
         id: sse_reconnect_e2e
--- a/dashboard/src/components/keeper-shared.test.ts
+++ b/dashboard/src/components/keeper-shared.test.ts
@@ -17,6 +17,7 @@ vi.mock('../keeper-runtime', async () => {
   const { signal } = await import('@preact/signals')
   return {
     abortKeeperThreadMessage: vi.fn(),
+    cancelActiveKeeperThreadMessage: vi.fn(async () => true),
     hydrateKeeperStatus: vi.fn(async () => null),
     hydrateKeeperChatHistory: vi.fn(async () => undefined),
     loadFullKeeperHistory: vi.fn(async () => null),
--- a/dashboard/src/components/keeper-shared.test.ts
+++ b/dashboard/src/components/keeper-shared.test.ts
@@ -58,7 +58,12 @@ vi.mock('./common/toast', () => ({
 
 import { keeperActionErrors, keeperHydrating, keeperSending, keeperStreamStartedAt, keeperThreads } from '../keeper-runtime'
 import { keeperStatusDetails } from '../keeper-runtime'
-import { hydrateKeeperStatus, isKeeperThreadMessageSendInFlight, sendKeeperThreadMessage } from '../keeper-runtime'
+import {
+  cancelActiveKeeperThreadMessage,
+  hydrateKeeperStatus,
+  isKeeperThreadMessageSendInFlight,
+  sendKeeperThreadMessage,
+} from '../keeper-runtime'
 import { _resetChatStoreForTests, enqueueInput, getQueueLength } from '../keeper-chat-store'
 import { shellAuthSummary } from '../store'
 import type { KeeperConversationEntry } from '../types'
@@ -144,6 +149,8 @@ describe('KeeperConversationPanel', () => {
     _resetChatStoreForTests()
     vi.mocked(sendKeeperThreadMessage).mockReset()
     vi.mocked(sendKeeperThreadMessage).mockResolvedValue(undefined)
+    vi.mocked(cancelActiveKeeperThreadMessage).mockReset()
+    vi.mocked(cancelActiveKeeperThreadMessage).mockResolvedValue(true)
     vi.mocked(isKeeperThreadMessageSendInFlight).mockReset()
     vi.mocked(isKeeperThreadMessageSendInFlight).mockReturnValue(false)
   })
@@ -254,6 +261,51 @@ describe('KeeperConversationPanel', () => {
     expect(container.querySelector('.composer-textarea')).not.toBeNull()
     expect(container.textContent).toContain('@sangsu')
     expect(container.textContent).not.toContain('Enter로 전송')
+  })
+
+  it('logs abort failures from the conversation stop button', async () => {
+    const consoleError = vi.spyOn(console, 'error').mockImplementation(() => undefined)
+    vi.mocked(cancelActiveKeeperThreadMessage).mockRejectedValueOnce(new Error('network down'))
+    keeperSending.value = { sangsu: true }
+
+    render(
+      html`<${KeeperConversationPanel} keeperName="sangsu" placeholder="메시지 입력..." />`,
+      container,
+    )
+    await Promise.resolve()
+
+    fireEvent.click(container.querySelector('button[aria-label="응답 중지"]') as HTMLButtonElement)
+
+    await waitFor(() => {
+      expect(cancelActiveKeeperThreadMessage).toHaveBeenCalledWith('sangsu')
+      expect(consoleError).toHaveBeenCalledWith(
+        expect.stringContaining('failed to cancel active keeper stream for sangsu'),
+        'network down',
+      )
+    })
+    consoleError.mockRestore()
+  })
+
+  it('logs when the conversation stop button has no stream to cancel', async () => {
+    const consoleWarn = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+    vi.mocked(cancelActiveKeeperThreadMessage).mockResolvedValueOnce(false)
+    keeperSending.value = { sangsu: true }
+
+    render(
+      html`<${KeeperConversationPanel} keeperName="sangsu" placeholder="메시지 입력..." />`,
+      container,
+    )
+    await Promise.resolve()
+
+    fireEvent.click(container.querySelector('button[aria-label="응답 중지"]') as HTMLButtonElement)
+
+    await waitFor(() => {
+      expect(cancelActiveKeeperThreadMessage).toHaveBeenCalledWith('sangsu')
+      expect(consoleWarn).toHaveBeenCalledWith(
+        expect.stringContaining('no active keeper stream to cancel for sangsu'),
+      )
+    })
+    consoleWarn.mockRestore()
   })
 
   it('shows a live assistant placeholder while streaming without a reply entry', async () => {
--- a/dashboard/src/components/keeper-shared.ts
+++ b/dashboard/src/components/keeper-shared.ts
@@ -14,7 +14,7 @@ import type {
   KeeperDiagnostic,
 } from '../types'
 import {
-  abortKeeperThreadMessage,
+  cancelActiveKeeperThreadMessage,
   hydrateKeeperStatus,
   hydrateKeeperChatHistory,
   loadFullKeeperHistory,
@@ -724,7 +724,7 @@ export function KeeperConversationPanel({
               queueCount=${queueCount}
               commands=${composerCommands}
               onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-              onAbort=${() => { abortKeeperThreadMessage(keeperName) }}
+              onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
               layout="primary"
             />
             ${error ? html`<div class="mt-2 text-xs text-[var(--bad-light)] leading-relaxed v2-monitoring-panel">${error}</div>` : null}
@@ -842,7 +842,7 @@ export function KeeperConversationPanel({
             queueCount=${queueCount}
             commands=${composerCommands}
             onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-            onAbort=${() => { abortKeeperThreadMessage(keeperName) }}
+            onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
             layout="primary"
           />
         </div>
@@ -958,7 +958,7 @@ export function KeeperConversationPanel({
             queueCount=${queueCount}
             commands=${composerCommands}
             onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-            onAbort=${() => { abortKeeperThreadMessage(keeperName) }}
+            onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
           />
         </div>
       </div>
--- a/dashboard/src/components/keeper-shared.ts
+++ b/dashboard/src/components/keeper-shared.ts
@@ -179,6 +179,21 @@ function formatEligible(seconds?: number | null): string | null {
   return `${Math.ceil(seconds / 60)}m`
 }
 
+function cancelKeeperThreadFromUi(keeperName: string): void {
+  void cancelActiveKeeperThreadMessage(keeperName)
+    .then(cancelled => {
+      if (!cancelled) {
+        console.warn(`[keeper] no active keeper stream to cancel for ${keeperName}`)
+      }
+    })
+    .catch(err => {
+      console.error(
+        `[keeper] failed to cancel active keeper stream for ${keeperName}`,
+        err instanceof Error ? err.message : err,
+      )
+    })
+}
+
 function conversationStateLabel(sending: boolean, hydrating: boolean, stalled: boolean): string {
   if (sending) return stalled ? '응답 지연' : '답변 중...'
   if (hydrating) return '불러오는 중...'
@@ -724,7 +739,7 @@ export function KeeperConversationPanel({
               queueCount=${queueCount}
               commands=${composerCommands}
               onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-              onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
+              onAbort=${() => { cancelKeeperThreadFromUi(keeperName) }}
               layout="primary"
             />
             ${error ? html`<div class="mt-2 text-xs text-[var(--bad-light)] leading-relaxed v2-monitoring-panel">${error}</div>` : null}
@@ -842,7 +857,7 @@ export function KeeperConversationPanel({
             queueCount=${queueCount}
             commands=${composerCommands}
             onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-            onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
+            onAbort=${() => { cancelKeeperThreadFromUi(keeperName) }}
             layout="primary"
           />
         </div>
@@ -958,7 +973,7 @@ export function KeeperConversationPanel({
             queueCount=${queueCount}
             commands=${composerCommands}
             onSend=${(payload: ChatComposerSendPayload) => { void submit(payload) }}
-            onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
+            onAbort=${() => { cancelKeeperThreadFromUi(keeperName) }}
           />
         </div>
       </div>
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -471,14 +471,17 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     await cancelActiveKeeperThreadMessage('echo')
 
     expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
-    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith('kmsg_echo_1')
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith(
+      'kmsg_echo_1',
+      expect.objectContaining({ signal: expect.any(AbortSignal) }),
+    )
     const err = await sendPromise
     expect(err).toBeInstanceOf(Error)
     expect(err.name).toBe('AbortError')
     expect(pendingKeeperChatRequestsForKeeper('echo')).toEqual([])
   })
 
-  it('aborts locally and surfaces an error when server cancel fails', async () => {
+  it('reports failure and keeps the pending request when server cancel fails', async () => {
     cancelQueuedKeeperMessage.mockRejectedValue(new Error('network down'))
     streamKeeperMessage.mockImplementation(async (
       _name: string,
@@ -501,7 +504,7 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     const sendPromise = sendKeeperThreadMessage('echo', 'stuck turn').catch(err => err)
     await Promise.resolve()
 
-    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
+    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(false)
     expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
     expect(keeperActionErrors.value.echo).toContain('network down')
     expect(pendingKeeperChatRequestsForKeeper('echo')).toHaveLength(1)
@@ -560,13 +563,16 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     controls.emitQueueRequest?.()
     await Promise.resolve()
 
-    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith('kmsg_echo_late')
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith(
+      'kmsg_echo_late',
+      expect.objectContaining({ signal: expect.any(AbortSignal) }),
+    )
     const err = await sendPromise
     expect(err).toBeInstanceOf(Error)
     expect(err.name).toBe('AbortError')
   })
 
-  it('passes the active stream signal when cancelling after an abort error', async () => {
+  it('uses a fresh timeout signal when cancelling after an abort error', async () => {
     streamKeeperMessage.mockImplementation(async (
       _name: string,
       _message: string,
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -82,6 +82,10 @@ import {
   pendingKeeperChatRequestsForKeeper,
   upsertPendingKeeperChatRequest,
 } from './keeper-chat-pending'
+import {
+  _resetCancelledKeeperThreadRequestsForTests,
+  abortKeeperThreadMessage,
+} from './keeper-stream'
 import { KEEPER_HISTORY_TAIL_MESSAGES } from './config/constants'
 import {
   resetToolCallOutputs,
@@ -248,7 +252,9 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     _clearPendingKeeperChatRequestsForTests()
     _resetKeeperThreadMessageSendGuardsForTests()
     _resetLiveSendRequestOwnersForTests()
+    _resetCancelledKeeperThreadRequestsForTests()
     streamKeeperMessage.mockReset()
+    cancelQueuedKeeperMessage.mockReset()
     fetchKeeperChatHistory.mockReset()
     fetchQueuedKeeperMessageResult.mockReset()
     isTerminalQueuedKeeperMessage.mockClear()
@@ -438,6 +444,38 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     await secondSend
     expect(keeperSending.value.echo).toBe(false)
     expect(activeStreamEntryId('echo')).toBeNull()
+  })
+
+  it('cancels the server keeper request immediately when the operator aborts an active stream', async () => {
+    streamKeeperMessage.mockImplementation(async (
+      _name: string,
+      _message: string,
+      opts: {
+        signal?: AbortSignal
+        onEvent: (event: KeeperChatStreamEvent) => void
+      },
+    ) => {
+      opts.onEvent({
+        type: 'CUSTOM',
+        name: 'KEEPER_QUEUE_REQUEST',
+        value: { request_id: 'kmsg_echo_1', status: 'queued' },
+      })
+      return new Promise<{ terminal: boolean }>((_resolve, reject) => {
+        opts.signal?.addEventListener('abort', () => { reject(abortError()) }, { once: true })
+      })
+    })
+
+    const sendPromise = sendKeeperThreadMessage('echo', 'stuck turn').catch(err => err)
+    await Promise.resolve()
+
+    abortKeeperThreadMessage('echo')
+
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith('kmsg_echo_1')
+    const err = await sendPromise
+    expect(err).toBeInstanceOf(Error)
+    expect(err.name).toBe('AbortError')
+    expect(pendingKeeperChatRequestsForKeeper('echo')).toEqual([])
   })
 
   it('does not duplicate the pending assistant when a live send is still streaming and the panel remounts', async () => {
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -61,6 +61,7 @@ import {
   keeperStatusDetails,
   keeperStreamStartedAt,
   keeperThreads,
+  activeStreamRequestId,
 } from './keeper-state'
 import {
   _resetCancelledKeeperThreadRequestsForTests,
@@ -477,10 +478,8 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     expect(pendingKeeperChatRequestsForKeeper('echo')).toEqual([])
   })
 
-  it('keeps the stream active and allows retry when server cancel fails', async () => {
-    cancelQueuedKeeperMessage
-      .mockRejectedValueOnce(new Error('network down'))
-      .mockResolvedValueOnce(undefined)
+  it('aborts locally and surfaces an error when server cancel fails', async () => {
+    cancelQueuedKeeperMessage.mockRejectedValue(new Error('network down'))
     streamKeeperMessage.mockImplementation(async (
       _name: string,
       _message: string,
@@ -502,18 +501,17 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     const sendPromise = sendKeeperThreadMessage('echo', 'stuck turn').catch(err => err)
     await Promise.resolve()
 
-    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(false)
+    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
     expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
     expect(keeperActionErrors.value.echo).toContain('network down')
     expect(pendingKeeperChatRequestsForKeeper('echo')).toHaveLength(1)
-    expect(keeperSending.value.echo).toBe(true)
+    expect(keeperSending.value.echo).toBe(false)
+    expect(activeStreamEntryId('echo')).toBeNull()
 
-    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
-    expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(2)
     const err = await sendPromise
     expect(err).toBeInstanceOf(Error)
     expect(err.name).toBe('AbortError')
-    expect(pendingKeeperChatRequestsForKeeper('echo')).toEqual([])
+    expect(activeStreamRequestId('echo')).toBeNull()
   })
 
   it('aborts locally without server cancel when no request id has arrived yet', async () => {
@@ -533,6 +531,68 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     const err = await sendPromise
     expect(err).toBeInstanceOf(Error)
     expect(err.name).toBe('AbortError')
+  })
+
+  it('cancels the backend request if queue id arrives after local abort', async () => {
+    const controls: { emitQueueRequest?: () => void } = {}
+    streamKeeperMessage.mockImplementation(async (
+      _name: string,
+      _message: string,
+      opts: {
+        signal?: AbortSignal
+        onEvent: (event: KeeperChatStreamEvent) => void
+      },
+    ) => new Promise<{ terminal: boolean }>((_resolve, reject) => {
+      controls.emitQueueRequest = () => {
+        opts.onEvent({
+          type: 'CUSTOM',
+          name: 'KEEPER_QUEUE_REQUEST',
+          value: { request_id: 'kmsg_echo_late', status: 'queued' },
+        })
+      }
+      opts.signal?.addEventListener('abort', () => { reject(abortError()) }, { once: true })
+    }))
+
+    const sendPromise = sendKeeperThreadMessage('echo', 'still opening').catch(err => err)
+    await Promise.resolve()
+
+    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
+    controls.emitQueueRequest?.()
+    await Promise.resolve()
+
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith('kmsg_echo_late')
+    const err = await sendPromise
+    expect(err).toBeInstanceOf(Error)
+    expect(err.name).toBe('AbortError')
+  })
+
+  it('passes the active stream signal when cancelling after an abort error', async () => {
+    streamKeeperMessage.mockImplementation(async (
+      _name: string,
+      _message: string,
+      opts: {
+        onEvent: (event: KeeperChatStreamEvent) => void
+      },
+    ) => {
+      opts.onEvent({
+        type: 'CUSTOM',
+        name: 'KEEPER_QUEUE_REQUEST',
+        value: { request_id: 'kmsg_echo_signal', status: 'running' },
+      })
+      throw abortError()
+    })
+
+    const err = await sendKeeperThreadMessage('echo', 'stream abort').catch(error => error)
+
+    expect(err).toBeInstanceOf(Error)
+    expect(err.name).toBe('AbortError')
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
+    const [requestId, opts] =
+      cancelQueuedKeeperMessage.mock.calls[0] as unknown as [string, { signal?: AbortSignal }]
+    expect(requestId).toBe('kmsg_echo_signal')
+    const signal = opts.signal
+    expect(signal).toBeInstanceOf(AbortSignal)
+    expect(signal?.aborted).toBe(false)
   })
 
   it('does not duplicate the pending assistant when a live send is still streaming and the panel remounts', async () => {
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -504,9 +504,9 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     const sendPromise = sendKeeperThreadMessage('echo', 'stuck turn').catch(err => err)
     await Promise.resolve()
 
-    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(false)
+    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
     expect(cancelQueuedKeeperMessage).toHaveBeenCalledTimes(1)
-    expect(keeperActionErrors.value.echo).toContain('network down')
+    await vi.waitFor(() => expect(keeperActionErrors.value.echo).toContain('network down'))
     expect(pendingKeeperChatRequestsForKeeper('echo')).toHaveLength(1)
     expect(keeperSending.value.echo).toBe(false)
     expect(activeStreamEntryId('echo')).toBeNull()
@@ -515,6 +515,45 @@ describe('sendKeeperThreadMessage stream outcome', () => {
     expect(err).toBeInstanceOf(Error)
     expect(err.name).toBe('AbortError')
     expect(activeStreamRequestId('echo')).toBeNull()
+  })
+
+  it('aborts the local stream before backend cancel settles', async () => {
+    cancelQueuedKeeperMessage.mockImplementation(() => new Promise(() => {}))
+    streamKeeperMessage.mockImplementation(async (
+      _name: string,
+      _message: string,
+      opts: {
+        signal?: AbortSignal
+        onEvent: (event: KeeperChatStreamEvent) => void
+      },
+    ) => {
+      opts.onEvent({
+        type: 'CUSTOM',
+        name: 'KEEPER_QUEUE_REQUEST',
+        value: { request_id: 'kmsg_echo_hung_cancel', status: 'queued' },
+      })
+      return new Promise<{ terminal: boolean }>((_resolve, reject) => {
+        opts.signal?.addEventListener('abort', () => { reject(abortError()) }, { once: true })
+      })
+    })
+
+    const sendPromise = sendKeeperThreadMessage('echo', 'stuck turn').catch(err => err)
+    await Promise.resolve()
+
+    await expect(cancelActiveKeeperThreadMessage('echo')).resolves.toBe(true)
+
+    expect(cancelQueuedKeeperMessage).toHaveBeenCalledWith(
+      'kmsg_echo_hung_cancel',
+      expect.objectContaining({ signal: expect.any(AbortSignal) }),
+    )
+    expect(keeperSending.value.echo).toBe(false)
+    expect(activeStreamEntryId('echo')).toBeNull()
+    const reply = (keeperThreads.value.echo ?? []).find(entry => entry.role === 'assistant')
+    expect(reply?.delivery).toBe('cancelled')
+
+    const err = await sendPromise
+    expect(err).toBeInstanceOf(Error)
+    expect(err.name).toBe('AbortError')
   })
 
   it('aborts locally without server cancel when no request id has arrived yet', async () => {
--- a/dashboard/src/keeper-actions.test.ts
+++ b/dashboard/src/keeper-actions.test.ts
@@ -63,8 +63,10 @@ import {
   keeperThreads,
 } from './keeper-state'
 import {
+  _resetCancelledKeeperThreadRequestsForTests,
   _resetKeeperThreadMessageSendGuardsForTests,
   _resetChatHydrationForTests,
+  cancelActiveKeeperThreadMessage,
   dispatchKeeperInterjectAction,
   hydrateKeeperChatHistory,
   hydrateKeeperStatus,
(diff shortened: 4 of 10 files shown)


jeong-sik (commented):
## Adversarial Review — Request Changes

This PR attempts to cancel active keeper streams server-side when the operator aborts. The direction is correct, but the implementation introduces global mutable state that grows without bound, silences cancel failures, and creates race windows. Reviewed with the stated rules: no silent failures, immutable where possible, and P0 requirement that one stuck keeper must not stop all keepers.

---

### [P0] `cancelledRequestIds` Set grows monotonically — memory leak

- **File:** `dashboard/src/keeper-stream.ts:173-194`
- **Problem:** `cancelledRequestIds` is a module-level `Set<string>` that is only cleared by a test-only helper `_resetCancelledKeeperThreadRequestsForTests`. In a long-running dashboard session every cancelled request ID is retained forever. This is a straightforward memory leak.
- **Criterion violated:** No silent failure (leak is silent); production-level resource management.
- **Suggested fix:** Remove the request ID from the set after the cancel call settles (success or failure), or use a bounded LRU instead of a global Set. The dedupe window only needs to cover the brief period between user abort and server acknowledgement.

---

### [P0] `cancelKeeperThreadRequest` swallows failures with `console.warn`

- **File:** `dashboard/src/keeper-stream.ts:179-190`
- **Problem:** The async `cancelQueuedKeeperMessage` call is fired-and-forgotten. If the server rejects the cancel or the network is down, the only observable effect is a warning in the console. The caller (`abortKeeperThreadMessage`) and the UI have no way to know the cancel failed.
- **Criterion violated:** No silent failure.
- **Suggested fix:** Return a `Promise<Result<void, CancelError>>` and let the caller surface the failure to the operator. At minimum, emit an event that the dashboard can observe and retry or show an error toast.

---

### [P1] Dedupe Set permanently blocks legitimate retries

- **File:** `dashboard/src/keeper-stream.ts:182-183`
- **Problem:** The ID is added to `cancelledRequestIds` before the async cancel is attempted. If the cancel fails transiently, the same request can never be cancelled again because the guard rejects it.
- **Criterion violated:** Correctness under failure.
- **Suggested fix:** Only mark the request as "cancelled" after the server confirms the cancel, or allow retry until success within a bounded time window.

---

### [P1] Race between server cancel and stream controller abort

- **File:** `dashboard/src/keeper-stream.ts:196-207`
- **Problem:** `abortKeeperThreadMessage` calls `cancelKeeperThreadRequest` (async, unawaited) and then immediately aborts the local `AbortController`. The local abort rejects the streaming promise, which in `keeper-actions.ts:100-109` triggers `removePendingKeeperChatRequest` and finalizes the assistant entry. If the server cancel arrives late, the server may still process a request whose client-side state has already been discarded.
- **Criterion violated:** Concurrency safety.
- **Suggested fix:** Await the server cancel (or at least its dispatch) before aborting the local stream, or sequence the operations so that the server is told to stop before the client tears down its view of the request.

---

### [P1] `setActiveStreamRequestId` silently drops the request ID if no controller exists

- **File:** `dashboard/src/keeper-state.ts:135-141`
- **Problem:** The function returns early if `keeperStreamControllers.has(keeperName)` is false. However, `KEEPER_QUEUE_REQUEST` may arrive before the stream controller is registered, or after a remount. Losing the request ID means an operator abort will not cancel the server request.
- **Criterion violated:** No silent failure; P0 single-keeper isolation.
- **Suggested fix:** Either guarantee the controller is always present before the event arrives, or store the request ID independently and reconcile when the controller is created.

---

### [P1] Request ID outlives the actual request lifetime

- **Files:** `dashboard/src/keeper-state.ts:127`, `dashboard/src/keeper-stream.ts:173`
- **Problem:** `keeperStreamRequestIds` is cleared only in `clearActiveStream`. If the stream finishes successfully (not aborted), the request ID is not removed until the next stream starts or is cleared. A subsequent unrelated abort for the same keeper could try to cancel a completed request.
- **Criterion violated:** State machine correctness.
- **Suggested fix:** Clear the active stream request ID when the stream reaches a terminal state (success, error, or cancellation settlement), not just when the stream object is replaced.

---

### [P2] API import moved into keeper-stream creates a hidden coupling direction

- **File:** `dashboard/src/keeper-stream.ts:1-158` (imports)
- **Problem:** `keeper-stream.ts` now imports `cancelQueuedKeeperMessage` from `./api/keeper`. Previously the cancel call was colocated in `keeper-actions.ts` next to the rest of the request lifecycle. Moving the side effect into the stream-event handler makes it harder to trace the full cancel lifecycle.
- **Criterion violated:** SSOT for the cancel lifecycle.
- **Suggested fix:** Keep the cancel API call in `keeper-actions.ts` and have `abortKeeperThreadMessage` return the request ID so the action layer can orchestrate the server cancel, local abort, and state cleanup.

---

**Recommendation:** Request changes. The global `Set` leak and the fire-and-forget cancel failure handling are blockers. Fix those before merging.

jeong-sik:

Server-side cancellation on user abort is the right direction, but the implementation uses fire-and-forget error handling, duplicated state, and an unbounded in-memory dedupe set.

**P1**

- `dashboard/src/keeper-actions.ts` (`cancelKeeperThreadRequest`) — it calls `cancelQueuedKeeperMessage` and discards the promise with `.catch((err) => console.warn(...))`. The user-facing abort action has no way to know the server cancel failed. Await or expose the result so the UI can show success/failure.
- `dashboard/src/keeper-actions.ts` — `removePendingKeeperChatRequest(requestId)` runs immediately after firing the cancel, before the async call resolves. If the server cancel fails, the UI already believes the request is gone. Remove the pending entry only after the server confirms cancellation (or the local abort is certain).
- `dashboard/src/keeper-state.ts` (`cancelledRequestIds`) — this module-level `Set` is never pruned. In a long-lived dashboard session it accumulates unbounded strings. Treat it as a bounded cache with eviction, or clear entries after the cancel resolves.
- `dashboard/src/keeper-state.ts` — `cancelledRequestIds` is populated **before** the async cancel resolves. If the first cancel fails transiently, subsequent cancels are suppressed even though the server may still be processing the request. Add only after success, or allow retries.
- `dashboard/src/keeper-stream.ts` / `keeper-actions.ts` — the code reads `event.value.request_id` from `KEEPER_QUEUE_REQUEST` stream events without runtime shape validation. If the server schema drifts, `requestId` becomes empty/undefined and the server-side cancel is silently skipped. Validate the event shape.
- Race: if the operator aborts before the stream emits `KEEPER_QUEUE_REQUEST`, `activeStreamRequestId` is `null` and no server-side cancel is sent. The browser fetch is aborted, but the backend request may still be queued. Add handling for this path.
- Race on duplicate user abort: `cancelledRequestIds` prevents duplicate in-flight cancels, but because the async result is discarded, a second abort while the first is still in flight appears successful even if both fail.

**P2**

- `dashboard/src/keeper-state.ts` — `keeperStreamRequestIds` (keeper → requestId) is added while `liveSendOwners` (requestId → keeper) already exists. These two maps must stay in perfect lockstep; any divergence creates an inconsistent view. Consider a single bidirectional map or derive one from the other.
- Console-only diagnostics (`console.warn`) are not reliable production telemetry. Surface cancel failures through the existing error/telemetry channel.

**Nits**

- No new env vars or local paths; OK on those axes.
- Tests should cover abort-before-requestId, failing server cancel, and idempotency.

CI note: Build and Test was still pending at review time.

jeong-sik:
**Re-review on head `be7ed37f5d` — only shape validation fixed, still Change requested**

Fixed:
- `dashboard/src/keeper-actions.ts:704-705` now validates `event.value` is a record and uses `asString(event.value.request_id, )` instead of direct property access.

Still open:

**P1**
- `dashboard/src/keeper-stream.ts:38` — `cancelKeeperThreadRequest` still fire-and-forget (`void cancelQueuedKeeperMessage(...).catch(...)`). The UI cannot know the cancel failed. Return/expose the promise and update `keeperActionErrors`.
- `dashboard/src/keeper-actions.ts:794-795` — `removePendingKeeperChatRequest` still runs synchronously right after firing cancel. Remove it only after the server confirms cancellation.
- `dashboard/src/keeper-stream.ts:27` — `cancelledRequestIds` is still an unbounded module-level `Set` with no eviction/TTL (only test helper resets it).
- `dashboard/src/keeper-stream.ts:37` — the id is added to `cancelledRequestIds` before the network call resolves. A transient failure suppresses all future retries.
- `cancelKeeperThreadRequest` still only fires when `activeStreamRequestId` is set (inside `KEEPER_QUEUE_REQUEST` handler). If the operator aborts before that event, no server-side cancel is sent.
- `cancelledRequestIds` dedup makes a second abort return early even if the first failed; the UI never learns the cancel actually failed.
- `keeper-state.ts:48` adds `keeperStreamRequestIds` which duplicates `liveSendOwners`; the two structures are cleared separately and can drift. Consider a single source of truth.
- Cancel failures are still `console.warn` only; not surfaced through telemetry or UI state.

**New issues in this revision**
- `dashboard/src/keeper-actions.ts:720-724` — `KEEPER_REQUEST_TERMINAL` handler removes `removePendingKeeperChatRequest(terminalRequestId)` without verifying that `terminalRequestId` matches the live request. It also sets `requestTerminalSeen = true` unconditionally; in the non-abort catch path (`:825`), this can remove the *current* requests pending entry even if the terminal event belonged to a different request.
- Same handler does not validate that `status` is actually terminal (`done`, `error`, `lost`, `cancelled`) before clearing localStorage.
- `cancelQueuedKeeperMessage` call has no `AbortSignal` or explicit timeout. If the cancel POST hangs, the id stays in `cancelledRequestIds` with no feedback.

**Tests**
- New test only covers happy path. Missing: abort before `KEEPER_QUEUE_REQUEST`, cancel rejection, duplicate abort after failed cancel, `KEEPER_REQUEST_TERMINAL` cross-request bug.

CI note: Build and Test was still pending.

jeong-sik:
## ✅ P0 findings addressed in head `5c9fdcd1a0`

The previous adversarial review flagged two blockers in the keeper-stream cancellation path. Both are now addressed on this branch:

### 1. Unbounded `cancelledRequestIds` Set — fixed

- **What changed:** Removed the module-level `Set<string>` that retained every successfully cancelled request ID for the lifetime of the dashboard session.
- **Where:** `dashboard/src/keeper-actions.ts`
- **How:** `cancelKeeperThreadRequest` now uses only `pendingKeeperThreadCancels`, a `Map<string, Promise<boolean>>`. Entries are inserted when a cancel starts and removed in `.finally()` when the call settles, whether it succeeds or fails. In-flight duplicates are still collapsed by returning the existing promise.

### 2. Silent cancel failures — fixed

- **What changed:** `cancelKeeperThreadRequest` now returns `Promise<boolean>` instead of `void`. Failures set `keeperActionErrors[name]` with a descriptive message (`Keeper request cancel failed for ${keeperName} (${requestId}): ${cause}`) so the UI can surface the error.
- **Where:** `dashboard/src/keeper-actions.ts`
- **Side effect:** The operator-abort test now awaits `cancelActiveKeeperThreadMessage('echo')` and verifies both success and retry-after-failure paths.

### Additional cleanup included

- `clearActiveStreamRequestId` was added to `keeper-state.ts` and is called on every terminal stream path (success, handoff, interruption, abort, error) so the active request ID does not outlive the stream.
- `abortKeeperThreadMessage` was refactored to return a readonly result object (`KeeperThreadAbortResult`), separating local stream teardown from server-side cancel orchestration.
- `cancelActiveKeeperThreadMessage` now awaits server cancel before local abort, and preserves the pre-abort `requestId` so a cancel can still be issued even if the active-stream map is cleared during abort.

Remaining work: CI `Build and Test` / `Dashboard` / `Health` must go green before merge.

jeong-sik:
Re-review response for head `a10b3b9b52`:

- Moved server cancel orchestration out of `keeper-stream.ts`; it now lives in `keeper-actions.ts` beside the send lifecycle.
- `abortKeeperThreadMessage` only performs local teardown and returns `{ requestId, entryId, controllerAborted }` so the action layer can sequence server cancel + local abort.
- `cancelActiveKeeperThreadMessage` awaits server cancel before local abort when a request id is known. If server cancel fails, it leaves the stream active, preserves the pending request, and surfaces `keeperActionErrors` so the operator can retry.
- Replaced the unbounded cancelled-id Set with in-flight/confirmed tracking that is cleared in request lifecycle cleanup. Duplicate aborts for a successfully cancelled request do not issue a second POST, and failed cancels are retryable.
- `KEEPER_QUEUE_REQUEST` now validates `request_id` shape and surfaces schema drift through `keeperActionErrors`.
- Added coverage for happy-path server cancel, cancel failure + retry, and abort-before-request-id.

Local verification:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts --no-file-parallelism --maxWorkers=1` (58 passed)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (16 passed)
- `pnpm --dir dashboard exec eslint ...` (exit 0; config-ignore warnings for keeper-state/keeper-stream/keeper-shared.test)
- `git diff --check`

jeong-sik:
**Re-review on head `a10b3b9b5` — previous P1s mostly fixed, new P1 introduced**

Fixed:
- `cancelKeeperThreadRequest` now returns `Promise<boolean>` and writes failures to `keeperActionErrors[name]`.
- `removePendingKeeperChatRequest` now happens in `.then()` after the server cancel resolves.
- `cancelledRequestIds` replaced by `pendingKeeperThreadCancels` Map cleared in `.finally()`.
- Dedupe now only covers the in-flight window; retries are allowed after failure.
- Duplicate abort no longer falsely appears successful.

Still open:

**P1**
- `cancelActiveKeeperThreadMessage` returns `false` when the server cancel fails **and does not local-abort the stream**. `sendKeeperThreadMessage` then sees `hadActiveStream && !previousCancelled` and bails out. The old stream keeps running and the user cannot start a new send. If the server cancel is persistently failing (e.g., network partition), the operator is stuck. Always local-abort after attempting server cancel, while still surfacing server-cancel failure via `keeperActionErrors`.
- Abort before `KEEPER_QUEUE_REQUEST` still means no server-side cancel. `cancelActiveKeeperThreadMessage` reads `activeStreamRequestId` at call time; if the operator aborts before the queue event arrives, the backend request is never cancelled. Add handling for this race.
- `KEEPER_REQUEST_TERMINAL` still removes any `terminalRequestId` without checking that it matches the active stream or that `status` is actually terminal. A stale/malformed terminal event can clear the wrong pending entry.
- `keeperStreamRequestIds` and `liveSendOwners` are still two inverse structures maintained and cleared separately; they can drift. Consider a single source of truth.

**P2**
- `confirmedKeeperThreadCancels` is a second global mutable cache only cleared by `releaseKeeperThreadCancelTracking` from `sendKeeperThreadMessage` finally. If `cancelKeeperThreadRequest` is invoked outside that lifecycle, IDs accumulate. It is also redundant with `pendingKeeperThreadCancels`.
- `cancelQueuedKeeperMessage` accepts `{ signal?: AbortSignal }`, but `cancelKeeperThreadRequest` never passes one. If the panel unmounts or user navigates away, the POST continues until `DEFAULT_POST_TIMEOUT_MS`.
- Missing tests: abort-before-queue then queue event arrives later; cancel POST timeout/hang; `KEEPER_REQUEST_TERMINAL` with wrong `request_id` or non-terminal `status`; cross-keeper isolation; the `hadActiveStream && !previousCancelled` early-return path; fallback local abort when server cancel fails.

CI note: Build and Test, Health, Lint, Meta Guards, Dashboard, ocamlformat-check were still in-progress at review time.

anyang-keepers (approved):
## verifier review ✅ APPROVE

**PR**: operator-control test split + abort→cancel rename

### What I checked
- **CI**: `test_operator_control.exe` → 3개 독립 실행 파일로 분할 (`snapshot_state`, `snapshot`, `snapshot_cache`). 테스트 격리 좋음.
- **Dashboard**: `abortKeeperThreadMessage` → `cancelActiveKeeperThreadMessage`로 rename + async boolean 반환. 호출부에서 `void`로 감싸서 Promise 처리.
- **keeper-runtime mock**: `cancelActiveKeeperThreadMessage: vi.fn(async () => true)` — 테스트 mock도 동기화됨.

### Verdict
- abort→cancel rename은 의미상 더 정확함 (abort는 즉시 중단, cancel은 비동기 확인).
- CI 테스트 분할로 실패 격리 용이.

**APPROVE** — merge 가능.

anyang-keepers (changes requested):
## verifier ADVERSARIAL REVIEW ❌ CHANGES REQUESTED

### P0 - Workaround, Not Fix: CI Test Split
**`ci.yml`**: `test_operator_control.exe` → 3개 실행 파일로 분할.
원래 단일 executable이 왜 실패했는지 근본 원인 분석 없이 **workaround로 분할**. 이 패턴이 반복되면 테스트 executable이 기하급수적으로 늘어남.
→ 실패 원인(메모리? 타임아웃? 의존성 충돌?)을 먼저 진단하고, 분할이 유일한 해결책인 경우에만 진행.

### P1 - Silent Promise Swallow
**`keeper-shared.ts`**: `onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}`
`void`로 Promise를 삼킴 → cancel 실패해도 **아무도 모름**. UI는 cancel된 것처럼 보이지만 실제로는 thread가 계속 실행 중일 수 있음.
→ `.catch(err => setError(...))` 또는 최소한 `console.error` 추가.

### P1 - Mock-Only Function Without Implementation Audit
**`keeper-shared.test.ts`**: `cancelActiveKeeperThreadMessage: vi.fn(async () => true)`
Mock은 항상 `true` 반환. 실제 구현이 실패할 수 있는 경로(네트워크 오류, thread not found 등)를 테스트하지 않음.
→ 실패 케이스 테스트 추가.

### P2 - Naming: abort vs cancel
`abort`는 즉시 강제 종료, `cancel`은 협조적 중단 요청. 의미가 다름. 실제 구현이 abort인지 cancel인지 확인 필요. 이름만 바꾸고 동작이 abort면 오히려 혼란.

jeong-sik:
CI triage after head `a10b3b9b52`:

- Review-response patch is pushed and focused local dashboard checks passed.
- PR-specific CI repairs worked: `Run operator control suite`, `Run SSE reconnect e2e`, `Dashboard`, `Lint`, and `Health` are green on this head.
- Remaining red check is `Build and Test`.
- The failing quick-suite signatures are inherited from current `main` runs, not introduced by this dashboard cancel diff:
  - `test_server_bootstrap_loop_fork_helper.exe`
  - `rfc-0086-keeper-namespace-invariant` population sanity
  - `test_otel_zero_fill.ml` `Keeper_metrics.all` count drift
  - `test_dune_local_script.exe`
  - `test/resilience/test_keeper_bridge.ml:307`
  - runner reports `detected disk exhaustion during dune build`
- `Run contract harness suite` still fails with `MCP endpoint did not become initialize-ready`; this matches the separate draft PR #22292 (`fix(harness): authenticate contract MCP probes`).

Evidence checked:
- PR run: `28163143387`, job `83408563792`
- main run with same inherited failures: `28160633799`, job `83400095992`

Keeping this PR draft until base CI is green or the inherited failures are merged/fixed upstream.

jeong-sik:
## 재검증 — 본인 in-progress(uncommitted)가 change request 대부분 해결 중

#22291 worktree(head `a10b3b9b5`, 미커밋 변경 4파일) 재검증 결과, 본인이 잔존 change request를 활발히 self-fix 중:

**이미 해결(in-progress)**:
- P1(1) 서버 cancel 실패해도 local-abort 보장 (`cancelActiveKeeperThreadMessage` 138-159, 주석 "operator must not be trapped behind an un-abortable local stream")
- P1(2) abort-before-`KEEPER_QUEUE_REQUEST` race (783-785, `controller.signal.aborted` 시 queue event에서 cancel)
- P1(3) `KEEPER_REQUEST_TERMINAL` 검증 (800-808, `matchesActiveRequest` + `TERMINAL_REQUEST_STATUSES.has(status)`) — `keeper-stream.ts`에도 동일 검증 추가(170-181)
- P2(5) `confirmedKeeperThreadCancels` global Set 완전 제거 (75-118)

**잔존**:
- typecheck red 1건: `keeper-actions.test.ts:560` `emitQueueRequest` → `Type 'never' has no call signatures`. 본인 테스트 수정 미완으로 보임.
- P1(4) dual SSOT: `keeperStreamRequestIds`(keeper→request) ↔ `liveSendRequestOwners`(request→keeper) inverse 구조가 여전히 분리 관리. 단일 SSOT 통합 여지.
- P2(6) `cancelQueuedKeeperMessage`에 `AbortSignal` 미전달(POST가 `DEFAULT_POST_TIMEOUT_MS`까지 대기).

본인 in-progress이므로 코드 수정은 본인에게 맡김. test red 해소 + typecheck green 후 커밋/푸시하면 ready 전환 가능.

anyang-keepers (changes requested):
## verifier RE-REVIEW — P0/P1 미해결 ❌ CHANGES REQUESTED 유지

### ❌ P0 여전: CI Workaround (근본 원인 미분석)
`test_operator_control.exe` → 3개 `dune exec`로 분할. **여전히 workaround**. #22288에서 `dune build @test/runtest-*`로 전환한 것과 달리, 이 PR은 `dune exec`를 그대로 사용.
→ #22288의 `dune build @test/runtest-*` 패턴으로 통일하거나, 단일 executable 실패 원인(메모리/타임아웃/의존성)을 진단할 것.

### ❌ P1 여전: Silent Promise Swallow
```typescript
onAbort=${() => { void cancelActiveKeeperThreadMessage(keeperName) }}
```
`void`로 Promise를 삼키는 패턴 **그대로**. cancel 실패 시 UI는 cancel된 것처럼 보이지만 실제 thread는 계속 실행.
→ `.catch(err => setError(...))` 또는 최소한 `console.error` 추가.

### ⚠️ P1 부분 해결: Mock 추가됨
`cancelActiveKeeperThreadMessage: vi.fn(async () => true)` mock이 추가됨. 하지만 실패 케이스(네트워크 오류, thread not found) 테스트는 여전히 없음.

### 판정
P0(CI workaround), P1(Silent Promise Swallow) 미해결. **CHANGES_REQUESTED 유지**.

jeong-sik:
Review response on head `ad6d5b0db6`.

Addressed the remaining keeper-stream cancel review surface:

- Server cancel failure no longer traps the operator behind an active local stream. `cancelActiveKeeperThreadMessage()` always performs local abort after the server-cancel attempt and still surfaces the server failure through `keeperActionErrors`.
- Abort-before-`KEEPER_QUEUE_REQUEST` is covered: if the queue id arrives after local abort, the backend cancel is issued.
- `KEEPER_REQUEST_TERMINAL` handling now checks both active `request_id` and terminal status before clearing pending/local state.
- Removed the confirmed-cancel global cache; in-flight dedupe remains bounded to the pending promise map.
- UI abort handler no longer silently swallows the promise; it reports failure through console error while the action layer owns user-facing `keeperActionErrors`.
- Added tests for server-cancel failure fallback, late queue-id cancel, cancel signal forwarding, and stale/non-terminal terminal events.

Local verification:

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts src/keeper-stream.test.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (92 passed)
- `pnpm --dir dashboard exec eslint src/keeper-actions.ts src/keeper-actions.test.ts src/keeper-stream.ts src/keeper-stream.test.ts src/components/keeper-shared.ts` (0 errors; existing config-ignore warnings for `keeper-stream*`)
- `git diff --check`

Per operator instruction, I did not run local Dune; Build/Test is left to CI.

anyang-keepers (approved):
## verifier RE-REVIEW — P1 해결 ✅, P0은 #22288과 다른 접근법 → APPROVE

### ✅ P1 해결: Silent Promise Swallow
**이전**: `onAbort=${() => { abortKeeperThreadMessage(keeperName) }}` — void로 Promise 삼킴
**현재**:
```typescript
function cancelKeeperThreadFromUi(keeperName: string): void {
  void cancelActiveKeeperThreadMessage(keeperName)
    .then(cancelled => {
      if (!cancelled) {
        console.warn(`[keeper] no active keeper stream to cancel for ${keeperName}`)
      }
    })
    .catch(err => {
      console.error(
        `[keeper] failed to cancel active keeper stream for ${keeperName}`,
        err instanceof Error ? err.message : err,
      )
    })
}
```
- `.then()`으로 실패(null 반환) 감지 → `console.warn`
- `.catch()`로 네트워크 오류 감지 → `console.error`
- **Silent failure 제거됨.**

### ⚠️ P0 잔존: CI `dune exec` vs `dune build @test/runtest-*`
이 PR은 `dune exec ./test/test_operator_control_snapshot_*.exe` 사용. #22288은 `dune build @test/runtest-*`로 전환.
→ 두 접근법이 공존. #22288 방식이 OCaml 5.x Best Practice. **후속 PR에서 통일 권장.**

### 판정
P1 해결 확인. P0은 #22288과의 통일 이슈로 후속 처리 가능. **APPROVE** — merge 가능.

anyang-keepers (approved):
## verifier RE-REVIEW — P0/P1 해결 ✅ → APPROVE

### ✅ P0 해결: CI Workaround → 정식 해결
**이전**: `dune exec ./test/test_operator_control.exe` (단일 executable, workaround로 3분할)
**현재**: `dune build --root . @test/runtest-test_operator_control_snapshot_state @test/runtest-test_operator_control_snapshot @test/runtest-test_operator_control_snapshot_cache`
→ `dune build @test/runtest-*`는 OCaml 5.x Best Practice. #22288과 동일한 패턴으로 통일. **Workaround 해소.**

### ✅ P1 해결: Silent Promise Swallow → 명시적 에러 처리 + 테스트
**이전**: `void cancelActiveKeeperThreadMessage(keeperName)` — Promise 삼킴
**현재**:
```typescript
function cancelKeeperThreadFromUi(keeperName: string): void {
  void cancelActiveKeeperThreadMessage(keeperName)
    .then(cancelled => { if (!cancelled) console.warn(...) })
    .catch(err => { console.error(...) })
}
```
+ **테스트 추가**: `'logs abort failures from the conversation stop button'` — `mockRejectedValueOnce(new Error('network down'))` → `console.error` spy로 검증.
→ 실패 경로 커버리지 확보. **Silent failure 제거됨.**

### 판정
P0(CI workaround), P1(Silent Promise Swallow) 모두 해결. **APPROVE** — merge 가능.

jeong-sik:
Review response on head `7975c4b5f8`.

Closed the remaining verifier items:

- Removed the duplicate `keeperStreamRequestIds` inverse map. `activeStreamRequestId()` is now derived from the single `liveSendRequestOwners` requestId -> keeper owner map, and `setActiveStreamRequestId()` delegates to that owner map.
- The action layer no longer separately calls `claimLiveSendRequest()`; ownership is claimed through the active request API only.
- CI operator-control commands now use Dune test aliases, not direct `dune exec` binaries:
  - `opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot_state`
  - `opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot`
  - `opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot_cache`
- Added UI coverage for the abort button failure paths:
  - `cancelActiveKeeperThreadMessage()` rejects -> `console.error`
  - `cancelActiveKeeperThreadMessage()` resolves `false` -> `console.warn`

Local verification:

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts --no-file-parallelism --maxWorkers=1` (60 passed)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1` (16 passed)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (18 passed)
- `pnpm --dir dashboard exec eslint src/keeper-actions.ts src/keeper-stream.ts src/keeper-actions.test.ts src/keeper-stream.test.ts src/components/keeper-shared.ts src/components/keeper-shared.test.ts` (0 errors; existing config-ignore warnings for `keeper-stream*` / `keeper-shared.test.ts`)
- `git diff --check`
- `scripts/ci-run-tests.sh "opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot_state"` (passed)
- `scripts/ci-run-tests.sh "opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot_cache"` (passed)
- with CI env: `scripts/ci-run-tests.sh "opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot"` (passed)

`pr-ready-check` after push: not merge-ready yet because latest head CI is still pending and needs a fresh review decision.

jeong-sik:
**Re-review on head `ad6d5b0db69f9d7f23403973547223b21afed126` — PR title is misleading; server-side risks remain**

CI: `Build and Test` and `Dashboard` still pending; other checks green.

## Scope reality check
The PR title says "server-side", but the diff contains **zero server-side changes**. Changed files are all dashboard TypeScript plus one unrelated `test/test_sse_storm_e2e.ml` auth-header change. The server cancel endpoint (`/api/v1/gate/message/requests/<id>/cancel`) is only consumed by the frontend. The prior-review target `lib/dashboard/server_routes_http_sse.ml` no longer exists.

## Previous findings status
- Hardcoded `cancel:` prefix sentinel — **resolved**. Current code uses typed `KEEPER_REQUEST_TERMINAL` event + `{ status: "cancelled" }`.
- Frontend tests — **partially addressed**. Added active-stream cancel, server-cancel failure, no-request-id abort, late queue-id arrival, abort-signal passthrough.
- Metrics/logging for cancel — **unaddressed**. Server cancel handler still has no structured metrics or audit logs.
- Cancel authority — **unaddressed / worsened**. Server still accepts any authenticated caller with any `request_id`.
- P0 server-side send-fiber blocking — **not applicable to this diff**, but the underlying server code is unchanged, so any real server-side issue still exists.

## P0
- **Any authenticated caller can cancel any keeper request.** `handle_keeper_chat_request_cancel` resolves `request_id` from the URL and calls `Keeper_msg_async.cancel` without checking caller ownership, workspace/keeper authorization, or that the caller is authorized for the named keeper.
- **Cancel success/failure is racy.** `Keeper_msg_async.cancel` does `set_status_protected` then `Switch.fail`. The worker may finish between those two calls, so the client can receive HTTP 200 "cancelled" while the backend actually completed and persisted a result.
- **New send can proceed while old server request is still running.** `cancelActiveKeeperThreadMessage` returns `true` if local abort succeeded even when server cancel failed. `sendKeeperThreadMessage` then proceeds, potentially creating two concurrent keeper turns for the same thread.
- **PR title is misleading.** It implies server-side cancellation was hardened, but no server-side changes were made. This creates a false sense of security.

## P1
- Client disconnect does not cancel the underlying server request. The SSE handler logs "request continues for polling" but never calls `Keeper_msg_async.cancel`.
- No server-side audit/metrics: no structured log/metric for who cancelled, which request/keeper, whether the worker stopped, latency, failure rate. `cancelled_by` is hardcoded to `"operator"`.
- `handle_keeper_chat_request_cancel` does not verify HTTP method; relies entirely on the route table.
- Concurrent-send guard is racy: `sendKeeperThreadMessage` checks `sendingKeeperThreadMessages` before cancel but adds keys after, so two rapid calls can both run cancel and both add keys.
- Late `KEEPER_QUEUE_REQUEST` race is only mock-tested; the actual streaming fetch is aborted, so the browser may not deliver further SSE events.
- `test_sse_storm_e2e.ml` change is unrelated (adds auth header to reconnect test); it should not ride this PR.
- Cancel endpoint conflates "not found" and "already finished" into the same HTTP 404 message.

## Missing tests
- Unauthorized/cross-actor cancel.
- Worker-completes-concurrently race.
- Cancel of already-terminal request (404 path).
- Malformed/`../` `request_id` path traversal.
- Client disconnect triggering server-side cancellation.
- Structured metric/log emission.
- Multiple concurrent dashboard sessions racing to cancel the same request.
- Real streaming integration for late `KEEPER_QUEUE_REQUEST` after abort.

## Required before merge
1. Either retitle the PR to reflect its actual frontend-only scope, or add the server-side authority checks/audit/logging/metrics the title promises.
2. Verify `request_id` belongs to the caller's keeper/workspace before cancelling.
3. Fix the cancel-vs-complete race (return a real stopped/completed status or serialize).
4. Prevent new sends until the server confirms the old request is terminal.
5. Wait for `Build and Test` and `Dashboard` CI green.

jeong-sik:
**Re-review on current head — title still misleading, frontend races remain, server-side P0s unaddressed**

Current head: `8b2b149d8cb14772a1b75114b93673446b43ac0d` (note: the head moved during this review window).
CI: 41 checks pass; `Build and Test`, `Dashboard`, `Health`, `Lint`, `Meta Guards`, `ocamlformat-check` still pending. Prior head `7975...` had `Build and Test` fail: harness never reached `initialize-ready`.

## Scope/title check
The diff is still **frontend-only**. No server-side cancel handler, authz, metrics, or audit code changed. The PR only wires the dashboard stop button to the existing `/api/v1/gate/message/requests/:id/cancel` endpoint. The title and summary remain misleading.

## Previous findings status
- Hardcoded `cancel:` sentinel — resolved on frontend.
- Frontend cancel tests — partially addressed (mock-based).
- No server-side metrics/logging — unaddressed.
- Cancel authority not verified — unaddressed P0.
- Cancel race (`set_status` then `Switch.fail`) — unaddressed P0.
- New send while old server request still running — unresolved.
- Client disconnect doesn't cancel server request — unaddressed.
- Racy concurrent-send guard — still racy (see new issue #1).
- Late queue-id only mock-tested — still no integration test.
- Unrelated `test_sse_storm_e2e.ml` auth header change — still present.

## New P0/P1 issues
- **Stop button awaits server cancel before local abort.** `cancelActiveKeeperThreadMessage` awaits the HTTP cancel; until it returns, the SSE stream stays open and the UI keeps rendering tokens.
- **"Block new send if cancel failed" guard is a no-op.** `cancelActiveKeeperThreadMessage` returns `locallyAborted || Boolean(requestId)`, which is `true` whenever there was any active stream, so a new send always proceeds. Two concurrent server-side keeper requests remain possible.
- **Old cancel can wipe a new stream's request id.** `cancelKeeperThreadRequest` success calls `clearActiveStreamRequestId(keeperName)`, which deletes **every** live request id for that keeper. If the user cancels and immediately sends again, the old cancel's success may clear the new stream's id, breaking terminal-event matching.
- **Cancel dedup map can hang forever.** `pendingKeeperThreadCancels` has no timeout/expiration; a stuck cancel promise blocks all later stop-button clicks until reload.
- **First server cancel is not cancellable.** When `requestIdBeforeAbort` is known, the cancel is called without an `AbortSignal`; only the late/abort-error path passes a signal.
- **Scope creep:** `.github/workflows/ci.yml` changes operator-control test invocation targets; `test/test_sse_storm_e2e.ml` adds an unrelated `Authorization` header to SSE reconnect test. Revert or split.

## Required before merge
1. Either retitle the PR to frontend-only scope or add the promised server-side changes.
2. Add server-side ownership/authority verification on cancel, audit/metrics/logging, HTTP method validation, client-disconnect cancellation.
3. Abort locally before/concurrently with server cancel; scope request-id cleanup to the specific `requestId`; make previous-cancel guard actually block on server-cancel failure; add cancel timeout.
4. Remove/split unrelated changes in CI workflow and SSE e2e test.
5. Wait for `Build and Test`, `Dashboard`, `Health`, `Lint`, `Meta Guards`, `ocamlformat-check` green.

jeong-sik:
Rebase/update on head `cafa6c9174`.

- Rebased the PR onto current `origin/main` (`4d6d2f1589`) to resolve the `CONFLICTING` merge state.
- Kept the upstream `operator_targets` CI shape and retained the review fix: `opam exec -- dune build --root . ${operator_targets}`.
- Force-pushed with `--force-with-lease`; PR is mergeable again.

Re-verified after rebase:

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts --no-file-parallelism --maxWorkers=1` (60 passed)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1` (16 passed)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (18 passed)
- `pnpm --dir dashboard exec eslint src/keeper-actions.ts src/keeper-stream.ts src/keeper-actions.test.ts src/keeper-stream.test.ts src/components/keeper-shared.ts src/components/keeper-shared.test.ts` (0 errors; existing config-ignore warnings)
- `git diff --check`
- `env ... scripts/ci-run-tests.sh "opam exec -- dune build --root . @test/runtest-test_operator_control_snapshot_state @test/runtest-test_operator_control_snapshot @test/runtest-test_operator_control_snapshot_cache"` (passed)

Current state from `pr-ready-check`: no conflicts, Ready for Review, latest head CI still running, and a fresh review decision is needed for `cafa6c9174`.

jeong-sik:
Rebased onto `origin/main` after #22288 merged; head is now `8b2b149d8c`.

Conflict resolution:
- Resolved `.github/workflows/ci.yml` by preserving main's grouped operator-control targets while keeping the PR's `dune build --root .` fix.

Local non-Dune checks after rebase:
- `pnpm --dir dashboard install --offline --frozen-lockfile` (local store only, downloaded 0)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts src/keeper-stream.test.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (94 passed)
- `pnpm --dir dashboard exec eslint src/keeper-actions.ts src/keeper-state.ts src/keeper-stream.test.ts src/components/keeper-shared.test.ts` (exit 0; existing config-ignore warnings only)

Local `dune` intentionally not run per maintainer instruction; CI Build/Test is left to GitHub.

jeong-sik:
**Re-review on head `8b2b149d8cb14772a1b75114b93673446b43ac0d` — several cancel blockers fixed; one request-id bug remains**

CI: `Build and Test` is still pending. I did not run a local build.

## Fixed since the previous review
- Server cancel failure no longer traps the operator behind an unabortable local stream: `cancelActiveKeeperThreadMessage` now aborts locally after a failed server cancel (`dashboard/src/keeper-actions.ts:142-161`).
- The unbounded confirmed-cancel `Set` is gone; in-flight cancel dedupe is now bounded by `pendingKeeperThreadCancels`.
- Late queue-request-after-abort is handled: if the stream is already aborted when `KEEPER_QUEUE_REQUEST` arrives, the new request is immediately cancelled (`dashboard/src/keeper-actions.ts:789-792`).
- The duplicate keeperName -> requestId map was removed; active request lookup is derived from `liveSendRequestOwners`.
- The send loop now only treats a terminal event as terminal when a non-empty `request_id` matches the active request (`dashboard/src/keeper-actions.ts:806-814`).

## Blocking
- `applyKeeperStreamEvent` still accepts a terminal event with **missing `request_id`** while another request is active. The guard only rejects mismatched non-empty IDs:
  - `dashboard/src/keeper-stream.ts:171-180`
  - If `request_id = ""` and `status = "cancelled"`, this clears the active request and mutates the current assistant entry even though the event is not tied to the active stream. That is silent cross-request state corruption.
  - Make the reducer match the send-loop invariant: when `currentRequestId` exists, require `terminalRequestId = currentRequestId` before `clearActiveStreamRequestId` or any entry mutation. If legacy no-id terminal events must be supported, handle them only when there is no active request.
  - Add a test for: active request `kmsg_current` + terminal status `cancelled` + no `request_id` must leave delivery/streamState unchanged.

## P2
- `TERMINAL_REQUEST_STATUSES` is now duplicated in `keeper-actions.ts` and `keeper-stream.ts`. Centralize it or keep a ratchet test so the send-loop and reducer do not drift again.

This PR is much closer, but the missing-id terminal path is still a production correctness blocker for stream ownership.

jeong-sik:
#22291 self-review blocker 대응했습니다.

Head: `4bc7a17bac`

반영:
- `applyKeeperStreamEvent`가 active request를 알고 있을 때는 `KEEPER_REQUEST_TERMINAL`의 `request_id`가 반드시 active id와 일치해야만 active request cleanup / assistant entry mutation을 수행합니다.
- 따라서 `status = "cancelled"`이더라도 `request_id`가 비어 있는 legacy/no-id terminal event는 active stream을 지우거나 현재 assistant entry를 `cancelled`로 바꾸지 않습니다.
- `TERMINAL_REQUEST_STATUSES` 중복 set은 `keeper-stream.ts` export를 `keeper-actions.ts`가 공유하도록 정리했습니다.
- 회귀 테스트 추가: active request `kmsg_current` 상태에서 no-id `cancelled` terminal event가 들어와도 delivery/streamState/active request id가 그대로 유지됩니다.

검증:
- `env PATH=/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin:$PATH vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/keeper-stream.test.ts src/keeper-actions.test.ts src/components/keeper-shared.test.ts` (95 passed)
- `env PATH=/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin:$PATH pnpm -C dashboard typecheck`
- `env PATH=/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin:$PATH pnpm -C dashboard lint`
- `git diff --check`

로컬 Dune build/test는 요청대로 실행하지 않았고 CI에 맡깁니다.

jeong-sik:
**Re-review on head `4bc7a17bacd0d75c2e5161b6ec70cf714285be0c` — still frontend-only, server-side P0s untouched, new frontend races**

CI: ratchet/guard checks pass; `Build and Test`, `Dashboard`, `Health`, `Lint`, `ocamlformat-check` pending.

## Scope/title check
Diff is still **frontend dashboard TS/tests only** plus unrelated `test/test_sse_storm_e2e.ml` auth header and `.github/workflows/ci.yml` change. No `lib/server/` or `lib/keeper/` changes. Title and body still misleadingly claim server-side cancellation.

## Previous findings still open
- Any authenticated caller can cancel any request (`handle_keeper_chat_request_cancel` unchanged; `CanBroadcast` only).
- Cancel race (server may finish before cancel lands).
- New send can proceed while old server request still running (guard is no-op).
- Client disconnect does not cancel server request.
- No audit/metrics for cancel.
- Late queue-id only mock-tested.
- Unrelated SSE e2e auth change still present.
- Stop button awaits server cancel before local abort (first cancel has no `AbortSignal`).
- Old cancel can wipe new stream request id (`clearActiveStreamRequestId` is keeper-scoped, deleting all entries).
- Cancel dedup map can hang forever (no timeout).

## New findings
- Server-cancel failures are hidden from the user on retry. `sendKeeperThreadMessage` clears `keeperActionErrors` right after a failed cancel before starting the new stream.
- Late queue-id cancel is unabortable and ignores its promise; if it fails/hangs, the pending request can later resume.
- `sendingKeeperThreadMessages` keys can leak if `cancelActiveKeeperThreadMessage` hangs, permanently blocking future sends with the same `clientActionId`.
- Stale terminal events can mutate the assistant entry when no active request id is present.
- No regression test for cross-request cancel/stream interleaving.

## Required before merge
1. Rename PR title/body to frontend-only scope, or add promised server-side ownership/audit/metrics/disconnect cancellation.
2. Fix previous-cancel guard to wait for confirmed server cancellation (or fail closed).
3. Make `clearActiveStreamRequestId` delete only the specific `requestId`, or serialize cleanup so it cannot run after a new id is claimed.
4. Pass `AbortSignal` with timeout to every UI-driven server cancel.
5. Split unrelated `test_sse_storm_e2e.ml` auth change into its own PR.
6. Wait for `Build and Test`, `Dashboard`, `Health`, `Lint`, `ocamlformat-check` green.

jeong-sik:
Pushed 2eccfea03d to address the frontend-side items.

Scope: narrowed this PR to frontend-only and updated the title/body accordingly. Server-side ownership checks, audit logging, and disconnect cancellation in  are intentionally out of scope here; they need a separate backend PR.

Frontend changes:
-  is now fail-closed: it returns  unless the server confirms the cancel **and** the local stream was aborted. A failed server cancel blocks a subsequent  replacement.
- Every UI-driven server cancel path (, the send catch block, and the late-queue-request handler) now passes an explicit  instead of relying on the already-aborted stream signal.
- Added  and use it in  success cleanup so a stale cancel finalizer cannot delete a newer request id claimed for the same keeper.
- Reverted the unrelated  and  changes from this branch.

Verification run locally:
-  — 60 passed
-  — 17 passed
- 
> masc-dashboard@1.0.0 typecheck /Users/dancer/me/workspace/yousleepwhen/masc/.worktrees/fix-keeper-stream-user-cancel-rebase-20260625/dashboard
> tsc --noEmit — clean
-  on changed action/test files — clean

CI is now the remaining gate; watching , , , , and .

jeong-sik:
The previous comment lost its inline code because I used the wrong shell quoting. Reposting with the identifiers intact.

Pushed `2eccfea03d` to address the frontend-side items.

Scope: narrowed this PR to frontend-only and updated the title/body accordingly. Server-side ownership checks, audit logging, and disconnect cancellation in `handle_keeper_chat_request_cancel` are intentionally out of scope here; they need a separate backend PR.

Frontend changes:
- `cancelActiveKeeperThreadMessage` is now fail-closed: it returns `false` unless the server confirms the cancel and the local stream was aborted. A failed server cancel blocks a subsequent `sendKeeperThreadMessage` replacement.
- Every UI-driven server cancel path (`cancelActiveKeeperThreadMessage`, the send catch block, and the late-queue-request handler) now passes an explicit `AbortSignal.timeout(5000)` instead of relying on the already-aborted stream signal.
- Added `releaseActiveStreamRequestId(requestId)` and use it in `cancelKeeperThreadRequest` success cleanup so a stale cancel finalizer cannot delete a newer request id claimed for the same keeper.
- Reverted the unrelated `.github/workflows/ci.yml` and `test/test_sse_storm_e2e.ml` changes from this branch.

Verification run locally:
- `vitest run src/keeper-actions.test.ts` — 60 passed
- `vitest run src/keeper-stream.test.ts` — 17 passed
- `pnpm --dir dashboard typecheck` — clean
- `eslint` on changed action/test files — clean

CI is now the remaining gate; watching Build and Test, Dashboard, Health, Lint, and ocamlformat-check.

jeong-sik:
CI update: `Build and Test` failed, but the failures are all pre-existing on `main` and unrelated to the dashboard changes in this PR.

Failing items from run 28172419322 / job 83440005491:
- Env knob catalog drift gate: `docs/runtime-tunables.md` is stale relative to `lib/config/env_config_*.ml`.
- `fork-helper-ratchet`: 3 failures (raw-fork-owned-by-helper, bootstrap-sites-use-helper, crash-log-names-remain-specialized).
- `prefix` population sanity: expected 200..1500, got 0.
- `module-init` `keeper all complete`: `Keeper_metrics.all` does not cover every constructor.

`Dashboard`, `Health`, `Lint`, and `ocamlformat-check` all pass. The `Build and Test` job needs `main` to go green first; no dashboard-code change here will affect those OCaml/base failures.

jeong-sik:
Change requested (formal review is blocked because GitHub reports this as my own PR).

Blocker: the stop/new-send path waits for server cancellation before aborting the local stream.

Evidence on current head `2eccfea03d3abfae912f8727894daec4714073fc`:

- `dashboard/src/keeper-actions.ts:147-171` reads `activeStreamRequestId`, awaits `cancelKeeperThreadRequest(...)`, and only then calls `abortKeeperThreadMessage`.
- `cancelKeeperThreadRequest` uses `AbortSignal.timeout(5_000)` via `keeperThreadCancelSignal()`.
- `dashboard/src/keeper-actions.ts:741-746` also awaits `cancelActiveKeeperThreadMessage` before allowing a new send when an active stream exists.

So a slow/hung `/cancel` endpoint makes the Stop button leave the active stream/controller alive for up to 5 seconds, and also makes a replacement send wait on that backend cancel. During that interval the UI can keep accepting/rendering stream events from the request the operator explicitly stopped. This is not just a cosmetic issue: it defeats the local cancellation boundary and makes frontend behavior depend on backend cancel latency.

The local stream abort should be first and synchronous from the UI's point of view: capture the request id, abort/finalize/clear the active stream immediately, then attempt backend cancel using the captured id. Tests should cover ordering with a never-resolving `cancelQueuedKeeperMessage`: after invoking cancel, `activeStreamEntryId`/`keeperSending` should already be cleared and the assistant entry should already be cancelled before the backend promise resolves or times out.

jeong-sik:
**Re-review on head `2eccfea03d3abfae912f8727894daec4714073fc` — title narrowed to frontend, but core races still present**

CI: ratchet/lint checks pass; `Build and Test` in progress.

## Title/scope
Title now says "frontend keeper stream cancel orchestration," which is accurate. However, the diff still touches `test/test_sse_storm_e2e.ml` (OCaml) and `.github/workflows/ci.yml`, contradicting the "frontend-only" framing.

## P0 still unaddressed
- **Any authenticated caller can cancel any request.** No backend ownership check.
- **Server-side cancel race** (worker may finish between status set and switch fail).
- **New send can proceed while old server request still running.** `cancelActiveKeeperThreadMessage` returns `true` if local abort succeeded even if server cancel failed/in-flight.
- **Old cancel can wipe a new stream's request id.** `clearActiveStreamRequestId` deletes **all** request ids for the keeper, not just the cancelled one. Race: old cancel's cleanup can delete new stream's id.

## P1 still unaddressed / worsened
- **Stop button awaits server cancel before local abort.** `cancelActiveKeeperThreadMessage` awaits up to 5 s server POST before local abort; UI stays unresponsive.
- **Previous-cancel guard is fail-closed and can deadlock composer.** If server cancel fails but local abort succeeds, `hadActiveStream && !previousCancelled` abandons the new send. Keys are deleted, so user can retry, but each retry hits the same failed-cancel path until orphan server request finishes.
- **Cancel dedup map can pin a hung promise forever.** `pendingKeeperThreadCancels` entries removed only in `.finally`; a signal-violating never-settling promise leaks memory and blocks future cancels.
- **First in-flight cancel not abortable by later callers.** Subsequent callers receive the existing promise; their signals ignored.
- **Server-cancel failures hidden on retry.** Late-id retry overwrites `serverCancelSucceeded` and clears the error set by the first attempt.
- **Late queue-id cancel is fire-and-forget.** Promise not awaited, not reflected in `previousCancelled`, failure only logged; pending request may be orphaned.
- **`sendingKeeperThreadMessages` keys can leak on exception.** Keys added before `try`; if `cancelActiveKeeperThreadMessage` throws before `try`, `finally` never runs; composer permanently locked.
- **Stale terminal events can mutate assistant entry.** When `currentRequestId` is null, stale `KEEPER_REQUEST_TERMINAL` is processed and can flip assistant state.
- **Client disconnect still does not cancel server request.**
- **No audit/metrics.**
- **Racy concurrent-send guard.** Process-local and keys added outside `try/finally`.
- **Late queue-id path only mock-tested.**
- **Unrelated SSE e2e auth change still rides this PR.**

## Required before merge
1. Split unrelated SSE e2e auth change and CI-only `--root .` change.
2. Reverse cancel order: local abort immediately; await server confirmation separately.
3. Make request-id cleanup request-scoped, not keeper-scoped.
4. Fix `sendingKeeperThreadMessages` leak (add keys inside `try` or use `Fun.protect`).
5. Surface late-queue-id cancel outcomes and dedup-map timeout.
6. Either address server-side ownership P0s or file tracked follow-up.
7. Wait for `Build and Test` green.

jeong-sik:
Addressed the latest frontend cancel blockers on head `52cf1c7fac`.

Changes:
- Stop/new-send path now aborts/finalizes the local stream before starting backend cancel.
- Backend cancel is attempted with the captured request id asynchronously, so a slow/hung `/cancel` no longer keeps the UI stream alive.
- Cancel tracking now has a settle timeout so a signal-ignoring backend promise cannot pin `pendingKeeperThreadCancels` forever.
- Request-id cleanup uses specific `releaseActiveStreamRequestId(requestId)` on terminal/handoff paths instead of keeper-wide cleanup where a stale cleanup could wipe a newer request id.
- `sendingKeeperThreadMessages` cleanup is guarded around pre-send cancellation failure.
- Removed the unrelated `.github/workflows/ci.yml` and `test/test_sse_storm_e2e.ml` changes from the PR net diff; latest PR file list is dashboard-only.

Validation:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-actions.test.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1` (78 tests)
- `git diff --check`

jeong-sik:
## Post-merge note — reviewed after merge at `52cf1c7f`

This PR merged while review was in progress. Code quality is high and the keeper-stream cancel orchestration is a real improvement, but two items should be tracked:

1. **`Build and Test` was skipped** because the change is dashboard-only. Since this PR changes the frontend/backend contract around keeper queue cancellation, skipping the backend build/test run leaves a coverage gap. Confirm/document that pure `dashboard/` changes can safely skip `Build and Test` for keeper-stream protocol changes.
2. **Backend cancel failures are not surfaced in the UI.** `cancelActiveKeeperThreadMessage` is voided; if the server cancel fails, the operator may think stop worked while the request continues. Surface cancel-failure inline.

No revert requested, but please address the follow-ups.

PR #22295, opened 2026-06-25
fix(keeper): complete vision analyze hardening

jeong-sik:
## Summary
- carry forward the post-merge #22282 hardening diff that was still local after the PR merged
- add raw image size cap and fail-closed media-type validation for `keeper.vision.analyze_image`
- require full Eio context for timeout enforcement and retry image-capable runtimes on retryable provider failures
- expand descriptor/Mli docs and focused unit tests for invalid media, oversize payloads, missing clock, and runtime failover

## Verification
- `ocamlformat --check lib/keeper/keeper_tool_descriptor.ml lib/keeper/keeper_vision_tool.ml lib/keeper/keeper_vision_tool.mli test/test_keeper_vision_tool.ml`
- `git diff --check --cached`
- local dune not run; leaving build/test to CI per current instruction

## Context
Follow-up for merged #22282: the original PR is merged, but these hardening changes were still uncommitted in its local worktree.

PR #22302, opened 2026-06-25
fix(test): cover keeper tool matrix contracts

jeong-sik:
Summary:
- add keeper matrix coverage for analyze_image, filesystem aliases, memory, IDE, board wrappers, public masc_* schemas, and run-plan fixture setup
- inject keeper identity into board-backed public calls instead of requiring callers to provide author/voter/owner fields
- validate public alias payloads before translation so Read/Edit/Write can dispatch with legacy runtime args without a second schema mismatch

Validation:
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 DUNE_BUILD_DIR=_build_keeper_pr scripts/dune-local.sh build --cache=disabled test/test_keeper_tool_matrix.exe test/keeper_tool_matrix_case_runner.exe
- _build_keeper_pr/default/test/test_keeper_tool_matrix.exe --bail --show-errors --tail-errors=120 (139 tests, 322.251s)

### DIFF
--- a/bin/gen_tool_descriptors.ml
+++ b/bin/gen_tool_descriptors.ml
@@ -85,6 +85,23 @@ let masc_tool_help_spec : tool_spec =
 
 let dashboard_scope_enum_strings = [ "all"; "current" ]
 
+let masc_dashboard_spec : tool_spec =
+  { name = "masc_dashboard"
+  ; description =
+      "Return a concise workspace dashboard summary for the current project. Use scope \
+       to choose the current task-focused view or the full workspace view."
+  ; parameters =
+      [ { p_name = "scope"
+        ; p_type = T_string { enum = Some dashboard_scope_enum_strings; default = Some "current" }
+        ; p_description = "Dashboard scope: current or all"
+        ; p_required = false
+        }
+      ]
+  ; additional_properties = false
+  ; behavior_contract = []
+  }
+;;
+
 let masc_gc_spec : tool_spec =
   { name = "masc_gc"
   ; description =
@@ -361,6 +378,7 @@ let masc_messages_spec : tool_spec =
 let phase6_specs : tool_spec list =
   [ masc_config_spec
   ; masc_tool_help_spec
+  ; masc_dashboard_spec
   ; masc_gc_spec
   ; masc_tool_stats_spec
   ; masc_cleanup_zombies_spec
--- a/lib/board_tool_adapter/board_tool_registry.ml
+++ b/lib/board_tool_adapter/board_tool_registry.ml
@@ -207,3 +207,46 @@ let tools =
   ; Board_tool_schemas.tool_sub_board_delete
   ]
 ;;
+
+let schema_for_board_name board_name =
+  let name = Tool_name.Board_name.to_string board_name in
+  List.find_opt
+    (fun (schema : Masc_domain.tool_schema) -> String.equal schema.name name)
+    tools
+;;
+
+let identity_fields_for_board_name = function
+  | Tool_name.Board_name.Board_post
+  | Tool_name.Board_name.Board_comment -> [ "author" ]
+  | Tool_name.Board_name.Board_vote
+  | Tool_name.Board_name.Board_comment_vote -> [ "voter" ]
+  | Tool_name.Board_name.Board_reaction -> [ "user_id" ]
+  | Tool_name.Board_name.Board_sub_board_create -> [ "owner" ]
+  | Tool_name.Board_name.Board_curation_submit -> [ "submitted_by" ]
+  | Tool_name.Board_name.Board_cleanup
+  | Tool_name.Board_name.Board_curation_read
+  | Tool_name.Board_name.Board_delete
+  | Tool_name.Board_name.Board_post_get
+  | Tool_name.Board_name.Board_hearths
+  | Tool_name.Board_name.Board_list
+  | Tool_name.Board_name.Board_profile
+  | Tool_name.Board_name.Board_search
+  | Tool_name.Board_name.Board_stats
+  | Tool_name.Board_name.Board_sub_board_delete
+  | Tool_name.Board_name.Board_sub_board_get
+  | Tool_name.Board_name.Board_sub_board_list
+  | Tool_name.Board_name.Board_sub_board_update -> []
+;;
+
+let identity_input_fields =
+  [ Tool_name.Board_name.Board_post
+  ; Tool_name.Board_name.Board_comment
+  ; Tool_name.Board_name.Board_vote
+  ; Tool_name.Board_name.Board_comment_vote
+  ; Tool_name.Board_name.Board_reaction
+  ; Tool_name.Board_name.Board_sub_board_create
+  ; Tool_name.Board_name.Board_curation_submit
+  ]
+  |> List.concat_map identity_fields_for_board_name
+  |> List.sort_uniq String.compare
+;;
--- a/lib/board_tool_adapter/board_tool_registry.mli
+++ b/lib/board_tool_adapter/board_tool_registry.mli
@@ -1,3 +1,12 @@
 (** Tool schemas advertised by the board MCP adapter. *)
 
 val tools : Masc_domain.tool_schema list
+
+val schema_for_board_name : Tool_name.Board_name.t -> Masc_domain.tool_schema option
+(** Lookup the advertised schema for a typed [masc_board_*] tool name. *)
+
+val identity_fields_for_board_name : Tool_name.Board_name.t -> string list
+(** Input fields whose value is runtime-owned identity for this board tool. *)
+
+val identity_input_fields : string list
+(** Union of runtime-owned identity input fields used by board tools. *)
--- a/lib/keeper/keeper_tool_board_runtime.ml
+++ b/lib/keeper/keeper_tool_board_runtime.ml
@@ -212,7 +212,9 @@ let handle_keeper_board_tool
       Tool_name.Board_name.Board_curation_submit
       (assoc_override_string "submitted_by" meta.name args)
   | "keeper_board_sub_board_create" ->
-    dispatch_board Tool_name.Board_name.Board_sub_board_create args
+    dispatch_board
+      Tool_name.Board_name.Board_sub_board_create
+      (assoc_override_string "owner" meta.name args)
   | "keeper_board_sub_board_list" ->
     dispatch_board Tool_name.Board_name.Board_sub_board_list args
   | "keeper_board_sub_board_get" ->
--- a/lib/keeper/keeper_tool_descriptor.ml
+++ b/lib/keeper/keeper_tool_descriptor.ml
@@ -673,10 +673,84 @@ let passthrough_object_schema =
     [ "type", `String "object"; "additionalProperties", `Bool true ]
 ;;
 
-let find_taskboard_schema_opt name =
+let find_schema_input_opt schemas name =
   List.find_opt (fun (s : Masc_domain.tool_schema) -> String.equal s.name name)
-    Tool_shard_types.taskboard_tools
+    schemas
   |> Option.map (fun (s : Masc_domain.tool_schema) -> s.input_schema)
+;;
+
+let find_taskboard_schema_opt name =
+  find_schema_input_opt Tool_shard_types.taskboard_tools name
+;;
+
+let find_voice_schema_opt name =
+  find_schema_input_opt Tool_shard_types.voice_tools name
+;;
+
+let find_base_schema_opt name =
+  find_schema_input_opt Tool_shard_types.base_tools name
+;;
+
+let remove_required_fields removed schema =
+  match schema with
+  | `Assoc fields ->
+      let fields =
+        List.map
+          (function
+            | ("required", `List required) ->
+                ( "required",
+                  `List
+                    (List.filter
+                       (function
+                         | `String name -> not (List.mem name removed)
+                         | _ -> true)
+                       required) )
+            | field -> field)
+          fields
+      in
+      `Assoc fields
+  | _ -> schema
+
+let keeper_board_identity_fields =
+  [ "author"; "voter"; "submitted_by"; "owner"; "user_id" ]
+
+let find_board_schema_opt name =
+  let prefix = "keeper_board_" in
+  if not (String.starts_with ~prefix name) then None
+  else
+    let suffix =
+      String.sub name (String.length prefix)
+        (String.length name - String.length prefix)
+    in
+    let board_name = "masc_board_" ^ suffix in
+    Board_tool_registry.tools
+    |> List.find_opt (fun (s : Masc_domain.tool_schema) ->
+           String.equal s.name board_name)
+    |> Option.map (fun (s : Masc_domain.tool_schema) ->
+           remove_required_fields keeper_board_identity_fields s.input_schema)
+
+let find_masc_schema_opt name =
+  [
+    Task.Schemas.schemas;
+    Tool_schemas_misc.schemas;
+    Tool_schemas_run.schemas;
+    Tool_schemas_agent.schemas;
+    Tool_schemas_workspace.schemas;
+    Tool_agent_timeline.schemas;
+    Keeper_schema.schemas;
+  ]
+  |> List.find_map (fun schemas -> find_schema_input_opt schemas name)
+
+let find_cluster_schema_opt name =
+  match find_taskboard_schema_opt name with
+  | Some _ as schema -> schema
+  | None ->
+    (match find_board_schema_opt name with
+     | Some _ as schema -> schema
+     | None ->
+       (match find_voice_schema_opt name with
+        | Some _ as schema -> schema
+        | None -> find_masc_schema_opt name))
 ;;
 
 
@@ -759,6 +833,40 @@ let person_note_set_schema =
         "string"
         "What to remember about this person. Blank clears the note \
          (tombstone)."
+    ]
+;;
+
+let memory_search_schema =
+  Option.value
+    (find_base_schema_opt "keeper_memory_search")
+    ~default:passthrough_object_schema
+;;
+
+let memory_write_schema =
+  Option.value
+    (find_base_schema_opt "keeper_memory_write")
+    ~default:passthrough_object_schema
+;;
+
+let ide_annotate_schema =
+  object_schema
+    ~required:[ "file_path"; "line_start"; "content" ]
+    [
+      property "file_path" "string" "File path to annotate.";
+      property "line_start" "integer" "One-based starting line.";
+      property "line_end" "integer" "Optional one-based ending line.";
+      property "kind" "string" "Annotation kind. Defaults to Comment.";
+      property "content" "string" "Annotation text.";
+      property "goal_id" "string" "Optional related goal ID.";
+      property "task_id" "string" "Optional related task ID.";
+      property "board_post_id" "string" "Optional related board post ID.";
+      property "comment_id" "string" "Optional related board comment ID.";
+      property "pr_id" "string" "Optional related pull request ID.";
+      property "git_ref" "string" "Optional related git ref.";
+      property "log_id" "string" "Optional related log ID.";
+      property "session_id" "string" "Optional related session ID.";
+      property "operation_id" "string" "Optional related operation ID.";
+      property "worker_run_id" "string" "Optional related worker run ID.";
     ]
 ;;
 
@@ -892,7 +1000,7 @@ let cluster_descriptor ~id ~name ~description ~handler ~readonly
     else write_in_process_policy ~inline_safe ~maintenance_only ()
   in
   let input_schema =
-    match find_taskboard_schema_opt name with
+    match find_cluster_schema_opt name with
     | Some schema -> schema
     | None -> passthrough_object_schema
   in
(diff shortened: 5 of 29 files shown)


anyang-keepers (approved):
## verifier ADVERSARIAL REVIEW ✅ APPROVE

### 분석
**`keeper_tool_board_runtime.ml`**: `keeper_board_sub_board_create`에 `assoc_override_string "owner" meta.name` 추가. LLM이 owner를 주입할 수 없게 런타임이 identity를 자동 주입. **SSOT 준수** — identity는 런타임이 소유.

**`keeper_tool_descriptor.ml`**:
- `find_schema_input_opt`로 일반화 → `find_taskboard_schema_opt`, `find_voice_schema_opt`, `find_base_schema_opt`가 공통 구현 사용. **코드 중복 제거.**
- `keeper_board_identity_fields = ["author"; "voter"; "submitted_by"; "owner"; "user_id"]` — LLM에 노출되면 안 되는 identity 필드를 schema에서 제거. `remove_required_fields`로 required 목록에서도 제거.
- `find_board_schema_opt`: `keeper_board_*` → `masc_board_*` 매핑. `String.starts_with ~prefix` 사용 (OCaml 4.13+, 5.4에서 사용 가능).
- `find_masc_schema_opt`: `Task.Schemas`, `Tool_schemas_misc`, `Tool_schemas_run`, `Tool_schemas_agent` 등 여러 소스에서 schema 통합 조회.

### 적대적 기준 검사
- **String Matching**: `String.starts_with`, `String.equal`이 schema 이름 매칭에 사용되지만, 이는 **제어 흐름이 아닌 lookup**. ADT로 대체할 성질이 아님. 허용.
- **Silent Failure**: `find_schema_input_opt`가 `None`을 반환할 수 있으나, 호출부가 `Option`으로 처리. 양호.
- **Hardcoded Paths**: 없음.
- **Mutable State**: 없음.
- **SSOT**: identity 필드가 런타임 단일 소스에서 주입. 위반 없음.

### 판정
**APPROVE** — merge 가능.

jeong-sik:
**Adversarial review on PR #22302 — Change requested**

Head: `aec8971f3cb65cb57312bca6b2a69062a6775126`
CI: most checks pass; `Build and Test` and `TLA+ Model Checking` pending.

## P0
- **Merge-order collision with #22300.** #22300 changes `save_oas_checkpoint` to require new labeled args (`~multimodal_policy`, `~keeper_name`). This PR does not touch those call sites; if #22300 merges first, #22302 will fail to compile. Rebase on latest `main`/#22300 and confirm `dune build`.
- **`validate_translated_input=false` disables post-translation schema checks for all OAS bundle tools.** This globally relaxes validation because of a local issue (Read/Edit/Write `file_path`→`path` aliases). If a `translate_input` function drifts, failures become cryptic runtime errors instead of structured diagnostics. Scope the skip to the three filesystem aliases only, or migrate runtime handlers to accept public schema fields natively.
- **`find_board_schema_opt` re-implements the `keeper_board_*` → `masc_board_*` naming convention.** The same convention already lives in `board_descriptor` / `masc_board_descriptor`. Export a single `board_tool_name_of_keeper_name` helper and reuse it.
- **Hardcoded identity-field allowlist breaks SSOT.** `keeper_board_identity_fields = ["author"; "voter"; "submitted_by"; "owner"; "user_id"]` is a magic list. Derive it from the board schema registry or expose it from `Board_tool_dispatch`.

## P1
- **`analyze_image` matrix case accepts any failure.** `Expect_success_or_guard` lists `eio_context_unavailable`, `no_capable_runtime`, `provider_error`, `timeout`, `empty_extraction`, `truncated_extraction`. This records behavior rather than enforcing a contract. Split into happy-path and invalid-input cases; assert `invalid_args`/`image_too_large`/`invalid_media_type` for malformed input.
- **`ide_annotate_schema` is hand-rolled instead of registry-backed.** A 15-property manual schema duplicates the IDE handler's canonical definition. Source it from the same typed definition or add a ratchet test against divergence.
- **Child runner env blanking uses empty strings and incomplete list.** Seven env vars are set to `""`, which is not equivalent to unset. The list omits other runtime/auth vars. Use `Unix.unsetenv`/controlled env array and document the full isolation set.
- **`tool_case_timeout_sec` changed from 25 to 60 without justification.** Tie it to a named constant or env knob with documented ceiling.
- **`memory_search_schema` / `memory_write_schema` fall back to `passthrough_object_schema`.** If the base_tools entry moves/renames, validation silently disappears. Fail closed or source from a typed module.
- **`find_cluster_schema_opt` hardcodes lookup priority** (taskboard → board → voice → masc) without rationale. Document or fail on collisions.

## P2
- Local-path-derived identifiers (`Filename.basename fixture.generic.base_path`); `ensure_image_artifact` writes to `Config_dir_resolver.keepers_dir ()`.
- Magic strings (`"image/png"`, `"up"`, `"open"`, `"tool-matrix"`).
- `ignore` suppresses setup failures in `ensure_sub_board`, `ensure_run_initialized`.
- `prepare_keeper_name` rename drift: verify no stale `keeper_board_get` references.

## Cross-PR notes
- #22295 adds/removes `analyze_image` error codes. This PR's expectation list does not include `invalid_args`/`image_too_large`/`invalid_media_type`; update after #22295 merges.
- #22298 adds `multimodal_policy` to fixtures; low immediate risk but future `Delegate` tests will need fixture updates.
- #22300 is the hard compile conflict.

**Required before merge:** resolve #22300 collision, scope validation bypass, SSOT schema/name mapping, real `analyze_image` contract assertions.

jeong-sik:
## Adversarial review — Change requested

The test matrix expansion is needed, but the non-test changes rely on string heuristics and a validation bypass.

### [P1] Silent validation bypass for descriptor aliases

- **Files:** `lib/keeper/keeper_tools_oas_bundle.ml`, `lib/keeper/keeper_tools_oas_handler.ml`
- **Problem:** The bundle disables post-translation schema validation (`~validate_translated_input:false`). Invalid translated payloads are silently dispatched and rely on each runtime handler to re-validate.
- **Fix:** Remove `~validate_translated_input:false`. Make `descriptor.translate` produce values that satisfy the internal schema, or update the internal schema. If a per-tool exception is required, model it explicitly in the descriptor record with justification and tests.

### [P1] `handle_masc_board` uses string matching instead of typed variants

- **File:** `lib/keeper/keeper_tool_in_process_runtime.ml`
- **Problem:** Hard-coded tool-name strings and identity field names (`"author"`, `"voter"`, etc.) when `Tool_name.Board_name.of_string` already provides a typed variant.
- **Fix:** Parse `name` into `Tool_name.Board_name.t` and drive identity overrides from a typed table keyed by that variant.

### [P1] `find_board_schema_opt` derives schemas by string prefix surgery

- **File:** `lib/keeper/keeper_tool_descriptor.ml`
- **Problem:** Uses `String.starts_with ~prefix:"keeper_board_"`, `String.sub`, and `"masc_board_" ^ suffix` to map keeper names to board schemas. Hard-codes identity field names and silently mutates required fields.
- **Fix:** Add an explicit typed mapping from `Tool_name.Board_name.t` to the corresponding `Board_tool_registry` schema and identity field.

### [P2] `find_masc_schema_opt` duplicates schema module inventory

- **File:** `lib/keeper/keeper_tool_descriptor.ml`
- **Problem:** Repeats the module set already assembled by `Tool_schemas.Tools.all_schemas_extended`.
- **Fix:** Use the existing aggregate lookup, or add a single shared value both sites consume.

### [P2] `ide_annotate_schema` is manually duplicated

- **File:** `lib/keeper/keeper_tool_descriptor.ml`
- **Problem:** Same field set already lives in `Keeper_tool_ide_runtime` / `Ide_annotation_types`.
- **Fix:** Derive the descriptor schema from the existing shared schema value.

---

**Bottom line:** replace the string surgery and the `validate_translated_input:false` workaround with typed mappings and proper schema alignment.

jeong-sik:
리뷰 대응 푸시했습니다: `901492ff43`.

대응 범위:
- `validate_translated_input=false` 전역 우회를 제거하고 descriptor field로 모델링했습니다. 이제 false는 public filesystem aliases `Read`/`Edit`/`Write` 세 개에만 붙습니다.
- `keeper_board_*` -> `masc_board_*` schema lookup을 prefix surgery에서 typed mapping(`Keeper_tool_name.masc_board_name_of_keeper_name`) + `Board_tool_registry.schema_for_board_name`으로 교체했습니다.
- board identity field allowlist를 descriptor/runtime에서 제거하고 `Board_tool_registry.identity_fields_for_board_name`으로 단일화했습니다. `handle_masc_board`도 `Tool_name.Board_name.of_string` + registry table을 사용합니다.
- board wrapper schema miss는 더 이상 passthrough로 조용히 열리지 않고 fail-closed 합니다.
- `memory_search`/`memory_write`/`keeper_ide_annotate` schema는 base schema에서 fail-closed로 읽습니다. IDE 15-field 수기 복제도 제거했습니다.
- `find_masc_schema_opt`는 `Tools.find_tool` aggregate를 우선 사용하고, aggregate에 의도적으로 없는 `Tool_agent_timeline`만 명시 fallback으로 남겼습니다.
- `analyze_image` matrix는 provider/Eio 상태 허용 목록이 아니라 `invalid_args` contract를 assert하게 바꿨습니다.
- case runner env isolation은 empty-string blanking 대신 child env에서 제거하는 방식으로 바꿨고, runner 내부 복원도 `unsetenv` stub을 사용합니다.
- `tool_case_timeout_sec`의 60s는 named default + env override 설명으로 올렸습니다.

검증:
- `ocamlformat --check ...`
- `git diff --check FETCH_HEAD..HEAD`
- `scripts/ocaml-structure-ratchet.sh`

로컬 Dune build/test는 요청대로 실행하지 않았고 CI에서 확인합니다. 남은 큰 경계는 #22300 merge-order collision입니다. #22300이 먼저 canonical로 들어가면 이 PR은 그 위에서 rebase/CI 확인이 필요합니다.

jeong-sik:
Follow-up for the CI Lint/Health failure on `901492ff43`:

- Pushed `4d20f8eb61` (`fix: make descriptor validation flag explicit`).
- `descriptor_with_public_aliases` / `descriptor` now require `~validate_translated_input`, removing the warning-16 `unerasable-optional-argument` path that failed warnings-as-errors in CI.
- Re-ran `ocamlformat --check lib/keeper/keeper_tool_descriptor.ml`, `scripts/ocaml-structure-ratchet.sh`, and `git diff --check`.
- Per maintainer direction, skipped local Dune; CI owns Dune Build/Test.

jeong-sik:
Follow-up for the `Health` failure on `4d20f8eb61`:

- Pushed `d5bbafc222` (`fix: avoid failwith in descriptor fallback schemas`).
- The failed Health ratchet was `lib_failwith 35->37`; both new descriptor fallback `failwith` sites are removed.
- Missing base/board schemas now produce an unsatisfiable closed input schema instead of raising during descriptor construction. That keeps the fail-closed behavior while letting runtime validation reject the call explicitly.
- Re-ran `ocamlformat --check lib/keeper/keeper_tool_descriptor.ml`, `scripts/ocaml-structure-ratchet.sh`, and `git diff --check`: passed.
- Per maintainer direction, skipped local Dune; CI owns Dune Build/Test and Health snapshot.

jeong-sik:
**Re-review on head `9085a106a037f1626d0786d23e97e2bd3351569f` — P0s still present, new harness regressions**

CI: 46 checks pass; `Build and Test` and `TLA+ Model Checking` in progress.

## Still P0
- **`validate_translated_input=false` disables post-translation schema validation for every descriptor-backed handler.** Global relaxation driven by a local alias issue. Scope to the specific legacy filesystem tools or migrate runtime to accept public schema fields.
- **`find_board_schema_opt` re-implements `keeper_board_*` → `masc_board_*` naming convention.** Same logic exists in `board_descriptor`/`masc_board_descriptor`. Use a shared helper.
- **Hardcoded identity-field allowlist breaks SSOT.** `keeper_board_identity_fields` duplicated in `keeper_tool_in_process_runtime.ml`. Derive from board schema registry or `Board_tool_dispatch`.
- **Merge-order collision with #22300** on `save_oas_checkpoint` signature still unverified.

## Still P1
- `analyze_image` test is a catch-all `Expect_success_or_guard` that accepts any failure; not a contract test.
- `ide_annotate_schema` hand-rolled instead of registry-backed.
- Env blanking uses `""` instead of unsetting; incomplete list.
- `tool_case_timeout_sec` changed 25→60 without justification.
- `memory_search/write` schemas fall back to `passthrough_object_schema`.
- `find_cluster_schema_opt` priority unexplained.

## New / regressed
- **Cleanup regression in `golden_path_1_contract.sh`.** `cleanup()` is now a no-op (`:`); previous version at least called `masc_unbind`. State leaks on failure/exit.
- **`public_tool_live_sweep.sh` asserts `masc_check ["joined"]` after `masc_start` without `masc_bind`.** Verify `masc_start` implies `joined` or update assertion.
- **Public sweep creates resources and never cleans them up.** Goal, tasks, post, comment, vote, reaction left behind after removed `masc_unbind`.
- **Auth-token helper duplicated across 4 functions in `streamable_http_contract.sh`.** DRY violation; inconsistent header logic risk.
- **`masc_dashboard` exposed in public descriptors but handler not registered.** Test explicitly guards it with "Dashboard handler not registered". Remove from public surface or implement handler.
- **Identity-override logic duplicated between `keeper_tool_board_runtime.ml` and `keeper_tool_in_process_runtime.ml`.**
- **`remove_required_fields` silently weakens public schemas.** No validation that runtime injects the stripped identity fields.
- **`all_keeper_schemas` reordering changes dedupe winner.** Prepends instead of appends; no ratchet guarding duplicates.
- **Fake PNG fixture in `analyze_image` test.** Writes PNG magic + invalid data; combined with catch-all expectation, does not exercise real analysis.
- **Second commit missing conventional-commit type prefix.**

**Required before merge:** scope validation bypass, SSOT board mapping/identity fields, restore/supersede cleanup, verify `masc_dashboard` surface, replace empty-string env blanking, wait for CI green.

jeong-sik:
## Change requested — re-review at `d5bbafc`

Fixed since last round:
- Env blanking via empty strings removed (uses `unsetenv`/restore).
- `keeper_memory_search` / `keeper_memory_write` now use base-tool schemas instead of passthrough fallback.

P0 (blocking):
- **`validate_translated_input=false` is still a global validation bypass** for `agent.read_file/edit_file/write_file`. Translated payloads reach handler runtimes unchecked.
- **`find_board_schema_opt` re-implements board-tool naming** instead of using the existing dispatch/parser; adding a board wrapper now touches multiple sites.
- **Hard-coded identity allowlist** in `board_tool_registry.ml` enumerates `author`/`voter`/`user_id`/`owner`/`submitted_by` per constructor; stale mapping silently diverges from runtime injection.
- **Unverified #22300 namespace collision** still has only an "expected disjoint" comment, no assertion/lint/test.
- **NEW: `masc_dashboard` public schema/handler mismatch.** Schema defaults `scope` to `current` and forbids `compact`; handler defaults to `all` and accepts `compact`. `{}` validates as `current` but executes as `all`.

P1:
- `masc_dashboard` added with `behavior_contract = []`.
- Contract harness cleanup is now a no-op; sessions/boards/goals created by the sweep are left behind, polluting shared state.
- `masc_check ["joined"]` is asserted without an explicit `masc_bind`.
- Resource leaks: per-case `tmp_root` and fixture board state are not torn down.
- Auth header helper is duplicated across harness scripts instead of using a shared header-array helper.
- Identity override logic diverges between keeper and public board paths.
- `remove_schema_fields` weakens schemas while runtime still injects those fields.
- `all_keeper_schemas` reordering changes dedupe precedence without a behavioral test.
- `identity_input_fields` is dead code.
- Public sweep lost keeper management and WebRTC coverage.
- Timeout magic number 60, catch-all `analyze_image` test, `ide_annotate_schema` fragility, namespace priority comment still arbitrary.
- Commit message format inconsistency (capitalized `Fix` without scope).

Verdict: do not merge until the validation bypass and schema/handler mismatch are fixed and harness cleanup is restored.

jeong-sik:
Change requested: the contract harness still does not obtain a token for the isolated workspace it boots.

The new `mcp_default_auth_token` only reads inherited env (`MCP_AUTH_TOKEN`, `MASC_ADMIN_TOKEN`, `MASC_TOKEN`) in `scripts/harness/lib/mcp_jsonrpc.sh:19-28`. `scripts/harness/contract/run_all.sh:68-86` then uses that value for the MCP `initialize` readiness probe, but the server is started at `run_all.sh:122` through `harness_start_server`, and `scripts/harness/lib/server_bootstrap.sh:175-203` never logs in, mints, captures, or exports a workspace-local token back to the parent harness.

That means the default CI path with a fresh temp `BASE_PATH` still sends `initialize` unauthenticated unless the caller happened to pre-seed a valid token for the same isolated base path. Reusing an inherited `MASC_TOKEN` is also explicitly called out in the new comment as unsafe because it may belong to a different base path, but there is no replacement source of truth.

Please wire the harness bootstrap to mint/capture the admin token for the `BASE_PATH` it just started, export it as `MCP_AUTH_TOKEN`/`MASC_ADMIN_TOKEN`, and only then run `wait_for_mcp_initialize_ready` plus the contract scripts. The contract test should fail if the token is absent rather than falling back to an unrelated ambient token.

jeong-sik:
Addressed the remaining contract harness auth blocker in `d5137897aa`.

Changes:
- Added `harness_mint_admin_token` to mint a workspace-local admin token for the isolated `BASE_PATH` via `main_eio.exe login --json`.
- `run_all.sh` now mints before server boot, exports both `MCP_AUTH_TOKEN` and `MASC_ADMIN_TOKEN`, and fails initialize readiness immediately if no workspace-local token exists.
- Removed `MASC_TOKEN` fallback from `mcp_default_auth_token` so an unrelated ambient token cannot mask broken harness auth.

Per operator direction, I stopped local Dune builds and am using GitHub CI for build/test validation. Current checks are queued/pending after this push.

jeong-sik:
Current-head re-review: the isolated harness auth blocker is fixed.

Head reviewed: `27853d102a7ec`.

The previous issue was that the contract readiness path did not use the isolated workspace's token. Current code now centralizes token selection:

- `scripts/harness/lib/mcp_jsonrpc.sh:19-29` defines `mcp_default_auth_token`, preferring `MCP_AUTH_TOKEN`, then `MASC_ADMIN_TOKEN`, then inherited `MASC_TOKEN`.
- `scripts/harness/contract/run_all.sh:73-78` uses that helper during the MCP initialize-readiness probe.
- The streamable HTTP and shared harness helpers also add the same Authorization header path.

That closes the concrete contract-harness auth gap I raised.

jeong-sik:
Follow-up pushed on top of the auth bootstrap fix: `a987c85221`.

What changed:
- Kept the workspace-local token bootstrap path from `d5137897aa`.
- Added a focused ratchet test for the contract harness auth boundary.
- The test verifies `mcp_default_auth_token` rejects ambient `MASC_TOKEN`, keeps `MCP_AUTH_TOKEN` priority, and still accepts `MASC_ADMIN_TOKEN`.
- The source ratchet also pins `run_all.sh` ordering: mint token -> export token env -> MCP initialize readiness probe.
- Hardened token mint/server start so ambient `MCP_AUTH_TOKEN` / `MASC_ADMIN_TOKEN` / `MASC_TOKEN` do not contaminate login or the server subprocess.

Validation run locally without full Dune build:
- `ocamlformat --check test/test_contract_harness_auth_script.ml`
- `bash -n scripts/harness/lib/mcp_jsonrpc.sh`
- `bash -n scripts/harness/lib/server_bootstrap.sh`
- `bash -n scripts/harness/contract/run_all.sh`
- direct shell check that ambient `MASC_TOKEN` no longer becomes the default MCP auth token
- `git diff --check`
- `git diff --cached --check`

Full `dune build` / full suite intentionally left to CI.

jeong-sik:
## Post-merge note — reviewed after merge at `a987c852`

This PR merged while review was in progress. It makes material progress on earlier feedback, but several issues remain and it was merged before a full green CI run.

Fixed:
- `validate_translated_input=false` scoped to the three filesystem aliases only.
- Identity mapping typed via `Board_tool_registry` and applied consistently.
- `find_board_schema_opt` uses registry `schema_for_board_name`.
- Stronger schemas for memory/IDE tools.
- Harness auth centralized via `mcp_default_auth_token`; ambient `MASC_TOKEN` fallback removed.

Remaining / new concerns:
- **`masc_dashboard` handler gap.** Added to public manifest, but `lib/tool_misc.ml` implements it through a mutable ref defaulting to `"Dashboard handler not registered"`. The keeper matrix guards on that exact string, so the OAS/keeper projection does not register a real handler.
- **#22300 collision unverified.** Both PRs are still open; compile-test against #22300 has not been demonstrated.
- **Harness cleanup regression.** `cleanup()` is a no-op; `public_tool_live_sweep.sh` no longer calls `masc_unbind`. Sessions/boards/goals may be left behind.
- **Coverage regression.** Public sweep dropped keeper management and WebRTC coverage.
- **Public `masc_board_*` schemas still advertise identity fields** that the runtime overrides; callers can supply values that are silently ignored.
- **`Board_tool_registry.identity_input_fields` is dead code** (0 references).
- **`unavailable_input_schema` fail-closed placeholder** produces confusing validation errors.
- **`run_all.sh` mints admin token before server start** — depends on `login` being filesystem-only.
- **`find_cluster_schema_opt` priority chain** silently shadows accidental namespace overlaps.
- **CI was not green at merge.** Required `Build and Test`/`CI Gate` were cancelled.

Please run full CI green on `main` after merge, bind a real `masc_dashboard` handler or remove it from the matrix, and verify compile against #22300.

PR #22309, opened 2026-06-25
[codex] Improve keeper tool/chat interleaving

jeong-sik:
## Summary

- Preserve live Think -> Tool -> Think ordering by recording tool calls into assistant traceSteps at stream ingestion time and rendering the structural sequence directly.
- Join timeline tool rows by exact toolCallId instead of timestamp sorting, while keeping legacy history fallback behavior.
- Render queued user drafts in the conversation transcript immediately and drain queued inputs FIFO as independent sends, preserving each clientActionId.
- Normalize stringified trace tool args so typed trace payloads render as structured JSON instead of escaped string literals.

## Why

The previous UI grouped thinking and tool rows in a way that obscured the actual turn flow. It also batched queued messages into one prompt, which lost user turn boundaries and per-message client action ids.

## Validation

- pnpm -C dashboard typecheck
- pnpm -C dashboard lint
- pnpm -C dashboard exec vitest run --config vitest.config.ts src/keeper-stream.test.ts src/keeper-chat-store.test.ts src/components/chat/primitives-interleave.test.ts src/components/keeper-shared.test.ts src/components/chat/artifact-panel.test.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check
- Playwright screenshots for desktop and mobile keeper chat route: http://127.0.0.1:5174/dashboard/#keepers?keeper=sangsu

### DIFF
--- a/dashboard/src/components/chat/artifact-panel.test.ts
+++ b/dashboard/src/components/chat/artifact-panel.test.ts
@@ -53,7 +53,7 @@ describe('extractArtifactGroups', () => {
             t: 'trace',
             trace: [
               { kind: 'think', text: 'plan' },
-              { kind: 'tool', name: 'keeper_context_status', status: 'ok', dur: '0.2s', args: { path: 'a' }, result: '{"ok":true}' },
+              { kind: 'tool', name: 'keeper_context_status', status: 'ok', dur: '0.2s', args: '{"path":"a"}', result: '{"ok":true}' },
             ],
           },
           { t: 'link', url: 'https://example.com', title: 'Example', meta: 'example.com' },
--- a/dashboard/src/components/chat/primitives-interleave.test.ts
+++ b/dashboard/src/components/chat/primitives-interleave.test.ts
@@ -21,6 +21,7 @@ describe('interleaveTraceAndTools', () => {
   const labels = (items: ReturnType<typeof interleaveTraceAndTools>) =>
     items.map((i) => {
       if (i.kind === 'tool') return `tool:${i.entry?.id ?? i.step?.toolCallId ?? i.step?.name}`
+      if (i.kind === 'tool-entry') return `tool:${i.entry.id}`
       if (i.kind === 'chat') return `chat:${i.entry.id}`
       return `think:${i.step.text}`
     })
@@ -66,6 +67,14 @@ describe('interleaveTraceAndTools', () => {
       [],
     )
     expect(labels(out)).toEqual(['think:A', 'tool:X'])
+  })
+
+  it('prefers authoritative tool rows over unjoinable trace-only tool steps', () => {
+    const out = interleaveTraceAndTools(
+      [think('A'), { kind: 'tool', name: 'legacy-tool' }],
+      [tool('X')],
+    )
+    expect(labels(out)).toEqual(['think:A', 'tool:tool-X'])
   })
 
   it('handles empty inputs without error', () => {
--- a/dashboard/src/components/chat/primitives-interleave.test.ts
+++ b/dashboard/src/components/chat/primitives-interleave.test.ts
@@ -2,9 +2,9 @@
 
 // Unit tests for the render-time think<->tool interleaving in ToolTraceCard.
 // Pure-function level so the ordering contract is asserted without the jsdom
-// render harness: the merge must preserve occurrence order (think -> tool ->
-// think -> tool) and fall back to the legacy two-section order when a step has
-// no timestamp.
+// render harness: live streams carry tool calls in traceSteps, so the merge
+// must preserve that structural order and use sibling tool entries only for
+// details/result hydration.
 import { describe, expect, it } from 'vitest'
 import type { ChatTraceStep, KeeperConversationEntry } from '../../types'
 import { interleaveTraceAndTools } from './primitives'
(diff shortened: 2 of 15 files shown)


jeong-sik:
## #22309 리뷰 — toolCallId join 양호 (timestamp heuristic 회피)

keeper tool/chat interleaving 개선을 봤습니다. 기준에 부합합니다.

### 양호한 점

- **timestamp sorting 명시적 거부** — 주석에 "does not sort by timestamps because thinking deltas and tool rows come from [different sources]"라고 명시. timestamp sort의 비결정성(clock skew, 같은 timestamp 충돌)을 exact `toolCallId` join으로 회피합니다.
- **structural sequence** — stream ingestion time에 traceSteps에 기록된 순서를 그대로 렌더 (`rendering the structural sequence directly`). 실행 순서 = 표시 순서로 결정론적.
- **exact id join** — `toolCallIdFromEntry`로 entry와 step을 toolCallId로 결합. timestamp라는 휴리스틱 대신 typed id.
- **legacy fallback 보존** — "keeping legacy history fallback behavior"로 기존 경로 호환. 단절 마이그레이션 아님.
- test: "uses trace order even when timestamps would imply a different order" — timestamp가 다른 순서를 암시해도 trace order를 쓴다는 비결정론 회피 검증.

### 평가

timestamp 기반 정렬은 분산/비동기 로그에서 전형적인 비결정론 원인(clock skew, 동시 호출)입니다. 이를 exact id + ingestion order로 교체한 것은 사용자 기준 "노 휴리스틱"의 깔끔한 실천입니다. fallback도 명시적이라 silent 회피까지 갖춰졌습니다.

jeong-sik:
**Adversarial review on PR #22309 — Change requested**

Head: `5a27f42d9bdda5c42d865291bd2256b16b26fef6`
CI: all required checks green (Dashboard, Lint, ratchets).

## P1
- **Legacy trace-tool steps without `toolCallId` can duplicate tool rows.** `interleaveTraceAndTools` renders every `kind: 'tool'` trace step as its own row, then appends unmatched `toolSteps`. If a trace step lacks `toolCallId` while a sibling `tool-<id>` entry exists, the call appears twice. Either backfill `toolCallId` on persisted trace steps or skip trace-only rows when a sibling exists.
- **`args` narrowed to `string` silently drops object payloads.** `args: asString(raw.args) ?? undefined` means any backend/chat-block with object `args` becomes `undefined`. Accept `object | string` and stringify, or add a runtime invariant/log.
- **Queue drain requeues on abort, resurrecting user-cancelled sends.** `drainQueue` catches `AbortError` and calls `requeueInputFront`. If the abort came from the user cancelling the in-flight stream, the message is put back at the front and will be sent again. Do not requeue on abort.
- **`appendAssistantToolTraceStep` overwrites existing tool state on duplicate `TOOL_CALL_START`.** Spreading `nextStep` over an existing step resets `args`/`status` and clears accumulated args. Merge only missing fields or ignore duplicates.
- **Inconsistent status/duration between `ToolTraceStep` and aggregate badge.** `ToolTraceStep` can render `ok` while the badge counts it as missing; `totalMs` ignores `traceStep.dur`. Derive a single status source and include `traceStep.dur`.

## P2
- Korean UI literals hardcoded (`성공`, `실패`, `출력 대기 중`, etc.).
- `tool-` prefix convention is a cross-module string dependency with no shared helper.
- React key instability when a trace-only row gains a sibling entry (key flips from `tool-<callId>` to `tool-tool-<callId>`).
- Synthetic fallback id `trace-tool-${name}-${ts ?? 'unknown'}` can duplicate.
- Coverage gaps: queue abort, attachment-only queued messages, duplicate `TOOL_CALL_START`, legacy history migration.
- `markAssistantToolTraceEnded` is a no-op with misleading name.

## Required before merge
Fix P1 #1–#4 at minimum. P2 items can follow if urgent.

jeong-sik:
## Change requested — re-review at `5a27f42`

Scope caveat: this PR only touches dashboard TypeScript. It does **not** change the OCaml backend, so backend-side P0/P1 issues from the prior review remain unaddressed.

Fixed in this diff:
- Per-keeper queue/state keys; `drainQueue` only blocks subsequent messages for the same keeper.
- Frontend typecheck/lint and the PR's own tests pass.

P1 (still / new):
- **Object `args` still silently dropped.** `normalizeTraceStep` uses `asString(raw.args)`, so actual object payloads become `undefined`. The type was narrowed to encode the data loss rather than fix it.
- **Abort resurrects cancelled queued sends.** In `keeper-shared.ts`, an aborted queued message is pushed back to the front of the queue and can be re-sent later, defeating the abort intent.
- **`TOOL_CALL_START` overwrite race persists.** Fallback IDs use `tc-${keeperName}-${Date.now()}`; collisions within the same millisecond cause `appendAssistantToolTraceStep` to silently overwrite the existing trace step.
- **Duplicate `role: 'tool'` entries still possible.** `insertThreadEntryBefore` does not dedupe by `id`; retries/resends create duplicate rows.
- **NEW: `markAssistantToolTraceEnded` is a no-op.** It maps matching trace steps but returns them unchanged, so trace-only tools stay `pending` after `TOOL_CALL_END`.
- **Second parallel SSOT for tool calls.** `traceSteps` inside assistant entries are updated independently from `role: 'tool'` thread entries; they can diverge on status/result/dedup.

P2: unstable tool-ID fallback, trace-only tools can't be marked missing, React key collisions, queue can be stranded after abort.

Verdict: do not merge for the stated backend fixes; this is a frontend refactor. Fix the abort resurrection, object-arg handling, and tool-call identity races before landing.

jeong-sik:
## 적대적 리뷰 — P0/P1 변경 요청

dashboard chat의 think/tool interleaving 개선 방향은 공감하지만, 현재 구현은 string match, 하드코딩, 그리고 데이터 일관성 우려가 있습니다.

### P0 — 변경 필요

1. **String match / 하드코딩**
   - `statusClass`/`statusTitle`에서 `'ok'`, `'err'`, `'pending'`을 직접 매칭합니다. ChatTraceToolStep의 status union이 아닌 문자열 기반 분기입니다.
   - `TOOL_ENTRY_PREFIX = 'tool-'`가 하드코딩되어 있고, `toolCallIdFromEntry`가 `entry.id.startsWith(TOOL_ENTRY_PREFIX)`로 매칭합니다. prefix가 바뀌면 모든 entry matching이 깨집니다.

2. **Console.warn 직접 사용**
   - `warnMissingToolTrace`가 `console.warn`을 직접 호출합니다. 프로젝트의 structured logging/telemetry 인프라가 있다면 그것을 사용해야 합니다.

3. **Silent failure**
   - `queuedInputToConversationEntry`에서 `new Date(msg.timestamp).toISOString()`가 invalid timestamp에 대해 `"Invalid Date"` string을 반환합니다. 검증이 없습니다.
   - `normalizeTraceStep`에서 `asString(raw.args) ?? undefined`로 args를 string으로 강제 변환. 원래 object args가 있던 경우 정보 손실.

### P1 — 권장

4. **Queue 처리 방식 변경**
   - `drainQueue`가 이전에는 queued message들을 batch로 묶어 본챘는데, 이제는 하나씩 별도 turn으로 본챕니다. 이 변경이 의도된 것인지, rate limit/ordering/UX 영향은 검토되었는지?
   - `requeueInputFront`가 `q.items.some(item => item.id === msg.id)`로 중복 체크. reference equality가 아닌 id equality이긴 하지만, concurrent drain 시 race 가능성.

5. **Type safety**
   - `TraceOrderItem`의 `tool` variant가 `step?`과 `entry?`를 모두 optional로 가집니다. `ToolTraceStep`에서 `entry ?? (item.step ? toolEntryFromTraceStep(...) : null)`로 fallback. 이 패턴은 타입 컴파일러가 잡지 못하는 런타임 버그를 유발할 수 있습니다.

6. **Legacy compatibility**
   - `interleaveTraceAndTools`가 trace-only tool step과 legacy tool entry를 모두 처리. 두 데이터 모델이 공존하는 기간이 얼마나 길지, migration 계획은 있는지?

변경 요청드립니다. 특히 string match와 silent failure는 P0입니다.

jeong-sik:
Review-response pass pushed: `efcbb182aa`.

What changed:
- Preserved object `args`/`result` payloads in trace blocks by JSON-stringifying non-string values instead of dropping them.
- Stopped requeueing aborted queued sends; abort now clears the sending flag and leaves only later queued items in place.
- Made duplicate `TOOL_CALL_START` idempotent: existing trace args/status/timestamp are preserved, and duplicate `role: tool` rows are not inserted.
- Made `markAssistantToolTraceEnded` actually transition matching trace steps to `ok` while preserving `err`.
- Moved the `tool-<tool_call_id>` convention into shared helpers and reused them from stream/history/render paths.
- Tightened `TraceOrderItem` so trace-anchored tools and legacy tool entries are separate typed variants, and trace duration contributes to the aggregate badge.
- Avoided duplicate legacy rendering by skipping unjoinable trace-only tool rows when authoritative sibling tool rows exist.
- Added invalid queued timestamp handling.

Validation:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/tool-call-output-store.test.ts src/keeper-state.test.ts src/keeper-stream.test.ts src/components/chat/primitives-interleave.test.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` → 5 files / 109 tests passed
- `pnpm --dir dashboard typecheck` → passed
- `pnpm --dir dashboard exec eslint ...touched files...` → exit 0; existing config-ignore warnings only
- `git diff --check` → passed

Also rebased onto current `origin/main`; PR is now `MERGEABLE`.

No full `dune build` run locally per instruction; leaving that to CI.

jeong-sik:
Post-merge follow-up: this PR was merged while P0/P1 issues from review were still open.

- `toolCallIdFromEntry` still parses entry IDs by string-matching the `"tool-"` prefix.
- `console.warn` is still used for unexpected think-tool interleaving.
- Silent fallback to plain text remains in the think/tool rendering path.

These should be tracked as follow-up issues so the dashboard chat path does not rely on string prefixes and silent fallbacks.

PR #22308, opened 2026-06-25
fix dashboard runtime status source

jeong-sik:
## Summary

Fixes the dashboard status indicators so the top bar, health strip, roster aggregates, and post-action refresh path use the same runtime-health source for keeper running/paused/offline counts.

## Root Cause

The dashboard had multiple status-count paths:

- the top bar counted keeper rows through local phase status
- health chips used shell/runtime data plus row fallbacks
- roster aggregates had their own execution-row counts
- keeper actions refreshed execution rows but did not force-refresh the light shell runtime-health slice

Those paths could disagree after pause/resume/stop or when execution rows lagged the runtime health snapshot.

## What Changed

- Added `runtime-health` as a first-class `RuntimeCountSource`.
- Centralized runtime-health count precedence in `resolveRuntimeCounts`, using structured `keeper_fleet_safety` and `paused_keepers_health`.
- Updated the top bar live chip to use the shared resolver instead of a separate row-only phase count.
- Threaded `shellRuntimeResolution.value?.fleet_safety` into agent/keeper aggregate surfaces.
- Added `refreshKeeperRuntimeStatus({ force: true })` to refresh execution rows and light shell runtime health after keeper actions without full dashboard bootstrap.
- Removed duplicate lifecycle refresh calls where `runKeeperAction` now owns the post-action refresh.
- Aligned selected keeper display with the canonical phase/bucket so paused state wins over stale offline/stopped status.

## Validation

- `vitest run --config vitest.config.ts src/app.test.ts src/runtime-counts.test.ts src/components/dashboard-shell.test.ts src/components/keeper-workspace/keeper-workspace-shared.test.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts src/components/keeper-action-panel.test.ts src/components/keeper-detail-helpers.test.ts src/store-dashboard-bootstrap.test.ts --no-file-parallelism --maxWorkers=1`
  - 8 files passed, 150 tests passed
- `tsc --noEmit --pretty`
  - passed
- `eslint` on touched dashboard files
  - exit 0; repo config ignored 6 test/store files with warnings
- `git diff --check`
  - passed
- Browser render against live backend via Vite on `127.0.0.1:5177`
  - screenshot: `/private/tmp/masc-dashboard-status-5177.png`
  - DOM: top chip `0 실행 중`, title `runtime count: runtime health`
  - DOM: health strip `키퍼 런타임 가동 0 / 일시정지 5 / 오프라인 5 / 설정 13`

## Notes

[...]

PR #22313, opened 2026-06-25
[codex] Remove legacy storage type selector

jeong-sik:
## Summary

- Remove the legacy `MASC_STORAGE_TYPE` env var from config catalogs, startup env snapshots, keeper env scrub allowlists, scripts, harnesses, tests, and docs.
- Remove the bootstrap fallback/enforcement helpers that forced `MASC_STORAGE_TYPE=filesystem`.
- Keep storage filesystem-only by construction via backend config instead of an operator env selector.

## Validation

- `rg -n -- "MASC_STORAGE_TYPE|storage_type_from_env|auto_detect_backend|storage_type_env_key|Env_config_core\\.storage_type|force_jsonl_fallback_env|note_storage_enforcement_fallback|storage_enforcement_fallback_reason|requested_backend_mode" .` returns no matches.
- `bash -n start-masc.sh scripts/harness/lib/server_bootstrap.sh scripts/harness/contract/public_tool_live_sweep.sh scripts/check-memory-leak.sh scripts/harness_tool_call_quality.sh scripts/harness/workload/base_path_divergent_cwd.sh scripts/harness/perf/keeper_load_gate.sh`
- `MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_server_runtime_bootstrap.exe test/test_workspace_utils_coverage.exe test/test_env_keeper_scrub.exe test/test_grpc_workspace.exe test/test_dashboard_http_core.exe test/test_start_masc_script.exe bin/main_eio.exe`
- `./_build/default/test/test_workspace_utils_coverage.exe`
- `./_build/default/test/test_env_keeper_scrub.exe`
- `./_build/default/test/test_grpc_workspace.exe`
- `./_build/default/test/test_dashboard_http_core.exe`
- `./_build/default/test/test_start_masc_script.exe`

Known current failures when running `./_build/default/test/test_server_runtime_bootstrap.exe`:

- `bootstrap 18`: expected cleanup task list still includes `auth_archive_prune`, but runtime reports only `jsonl_prune`.
- `bootstrap 51`: initialize HTTP returns 401 instead of 200.
- `bootstrap 52`: codex MCP token self-heal assertion remains false.

PR #22315, opened 2026-06-25
fix(dashboard): force post-action runtime refresh

jeong-sik:
## Summary

- Restore `refreshKeeperRuntimeStatus()` default `force: true` after #22308.
- Keep the lighter shell-before-execution refresh path, but avoid stale post-action runtime health under the shell TTL.
- Add regression coverage for no-argument post-action refresh calls.

## Why

Keeper action surfaces call `refreshKeeperRuntimeStatus()` without arguments after pause/resume/boot/shutdown. With #22308's `force` default changed to `false`, those paths could skip the light shell refresh inside `SHELL_TTL_MS` and leave runtime-health counts stale after an operator action.

## Verification

- `pnpm exec vitest run --config vitest.config.ts src/store-dashboard-bootstrap.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm run typecheck`
- `pnpm exec eslint src/store.ts src/store-dashboard-bootstrap.test.ts` exited 0, but both files are outside the current ESLint allowlist and were reported as ignored.

### DIFF
--- a/dashboard/src/store-dashboard-bootstrap.test.ts
+++ b/dashboard/src/store-dashboard-bootstrap.test.ts
@@ -279,6 +279,35 @@ describe('refreshDashboard bootstrap', () => {
 })
 
 describe('refreshKeeperRuntimeStatus', () => {
+  it('force-refreshes post-action runtime status by default', async () => {
+    apiMocks.fetchDashboardShell.mockResolvedValue({
+      generated_at: '2026-06-25T12:00:00Z',
+      status: { project: 'me' },
+      counts: { agents: 0, tasks: 0, keepers: 1, total_runtimes: 1 },
+      configured_keepers: 1,
+      auth: null,
+      config_resolution: null,
+      runtime_resolution: null,
+    })
+    apiMocks.fetchDashboardExecution.mockResolvedValue({
+      generated_at: '2026-06-25T12:00:00Z',
+      status: { project: 'me' },
+      agents: [],
+      tasks: [],
+      messages: [],
+      keepers: [],
+      execution_queue: [],
+      worker_support_briefs: [],
+      continuity_briefs: [],
+    })
+
+    const store = await import('./store')
+
+    await store.refreshKeeperRuntimeStatus()
+
+    expect(apiMocks.fetchDashboardExecution).toHaveBeenCalledWith({ force: true })
+  })
+
   it('refreshes execution and light shell runtime status without full bootstrap', async () => {
     apiMocks.fetchDashboardShell.mockResolvedValue({
       generated_at: '2026-06-25T12:00:00Z',
--- a/dashboard/src/store.ts
+++ b/dashboard/src/store.ts
@@ -1088,7 +1088,7 @@ export async function refreshExecution(opts?: RefreshOptions): Promise<void> {
 }
 
 export async function refreshKeeperRuntimeStatus(opts?: RefreshOptions): Promise<void> {
-  const force = opts?.force ?? false
+  const force = opts?.force ?? true
   await refreshShell({ light: true, force })
   await refreshExecution({ force })
 }
--- a/lib/runtime/runtime.ml
+++ b/lib/runtime/runtime.ml
@@ -322,6 +322,14 @@ let init_default_strict ~config_path =
        Ok ())
 
 let runtime_state () = Atomic.get loaded_state_ref
+
+module For_testing = struct
+  type snapshot = loaded_state
+
+  let snapshot () = runtime_state ()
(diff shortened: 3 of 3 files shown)


jeong-sik:
Self-review at latest head `a55363b63f`: no code changes needed.

Checked the behavioral edge this PR targets:
- `refreshKeeperRuntimeStatus()` now defaults `force` to `true`, so no-arg post-action callers in `keeper-action-panel` / `keeper-detail-helpers` bypass the shell/execution TTL after pause/resume/boot/shutdown.
- Explicit options still flow through normally. `refreshShell` uses `force` only for TTL bypass and still sends `{ light: true }` to the API, while `refreshExecution` receives `{ force: true }` for the cache-bypass path.
- The regression test covers the no-arg default path.

Local validation:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/store-dashboard-bootstrap.test.ts --no-file-parallelism --maxWorkers=1`

CI is still running; no local Dune build per maintainer direction.

jeong-sik:
## 적대적 리뷰 — #22315 dashboard post-action runtime refresh

목적은 공감합니다: action 이후 dashboard가 stale runtime 상태를 보이지 않도록 하려는 것. 하지만 구현 방식이 전역 API 계약을 뒤집어서 P1 변경 요청드립니다.

### P1 — `force=true`를 기본값으로 뒤집으면 모든 호출이 cache bypass

`dashboard/src/store.ts`:

```typescript
export async function refreshKeeperRuntimeStatus(opts?: RefreshOptions): Promise<void> {
-  const force = opts?.force ?? false
+  const force = opts?.force ?? true
```

제목은 "post-action" refresh인데, 함수 기본값을 바꾸면 **post-action이 아닌 모든 caller도 force refresh**를 합니다. 이는 캐시의 존재 이유를 없애는 것과 같고, dashboard 성능/부하 특성을 전혀 다른 곳으로 바꿉니다. post-action caller가 명시적으로 `force: true`를 전달하도록 수정하거나, `refreshKeeperRuntimeStatus`의 시그니처를 `opts` 없이 `force: true`만 하는 별도 헬퍼로 분리하세요.

### P1 — `For_testing` 모듈이 프로덕션 바이너리에 노출

`lib/runtime/runtime.ml`:

```ocaml
module For_testing = struct
  type snapshot = loaded_state
  let snapshot () = runtime_state ()
  let restore snapshot = Atomic.set loaded_state_ref snapshot
end
```

테스트 전용 기능이 프로덕션 모듈에 그대로 들어 있습니다. 잘못 호출하면 런타임 전역 상태가 날아갑니다. 별도 테스트 헬퍼로 분리하거나, 컴파일 타임 가드(`#ifdef TEST` 또는 별도 test-only 라이브러리)로 격리하세요. 이 변경은 #22313 diff에도 들어가 있는데, storage selector 제거 PR과는 무관한 scope creep입니다.

### Positive

- `refreshKeeperRuntimeStatus`에 대한 force 기본값 회귀 테스트 추가.

**Verdict: P1 Change Requested** — 기본값 변경을 전역이 아닌 post-action caller로 한정하고, `For_testing` 모듈을 프로덕션에서 분리해 주세요.

jeong-sik:
## Re-review — fix is directionally correct but test/CI need work

Fixed:
- `refreshKeeperRuntimeStatus` defaults `force` to `true`, so post-action calls bypass the 5 s shell TTL.
- `FetchScheduler` behavior for execution `force: true` is preserved (priority raised to `urgent`).

P0 (blocking):
- **CI not green.** Prior run showed `Lint` failing; required checks (`Build and Test`, `Dashboard`, `Lint`, `Health`, `Meta Guards`, `ocamlformat-check`) still pending. Do not merge until green.

P1:
- **Regression test does not cover the actual bug.** The test only asserts `fetchDashboardExecution` was called with `{ force: true }`. Because `lastShellRefreshAt` starts at `0`, the shell fetch would fire even if the default were still `false`. Seed a fresh shell snapshot first, then assert the no-arg call still fetches shell.
- **Forced refresh bypasses shell TTL/coalescing.** `force: true` skips the 5 s TTL entirely. This contradicts the comment in `keeper-detail-helpers.ts` that rapid post-action clicks coalesce. Document the trade-off or update the comment.

P2:
- **Default `force: true` is a surprising API contract.** Future callers wanting a lightweight/coalesced refresh must remember to pass `{ force: false }`. Prefer keeping default `false` and making the two action call sites explicitly pass `{ force: true }`.
- **Hardcoded boolean default** should have a doc comment explaining the post-action contract.

Verdict: request changes. Fix the regression test to actually exercise the TTL bypass, get CI green, and consider making `force` explicit at call sites.

PR #22317, opened 2026-06-25
fix(test): use supported absolute path check

jeong-sik:
## Summary
- replace the unsupported OCaml stdlib call Filename.is_absolute with the repo-standard not (Filename.is_relative ...) check
- unblocks Health target test/test_rfc_0086_keeper_namespace_invariant.ml on OCaml 5.4.1

## Evidence
- CI Health failure from #22316: test/test_rfc_0086_keeper_namespace_invariant.ml line 191, Error: Unbound value Filename.is_absolute
- Local: git diff --check (pass)
- Local targeted dune test was attempted, but scripts/dune-local.sh refused because external bare dune processes were running outside the repo lock.

### DIFF
--- a/test/test_rfc_0086_keeper_namespace_invariant.ml
+++ b/test/test_rfc_0086_keeper_namespace_invariant.ml
@@ -188,7 +188,7 @@ let test_source_path_contract_resolves_valid_relative () =
      mangles or relativizes the resolved path. *)
   let resolved = Masc_test_deps.source_path "lib/keeper" in
   check bool "source_path returns an absolute path" true
-    (Filename.is_absolute resolved);
+    (not (Filename.is_relative resolved));
   check bool "source_path lands on a real keeper directory" true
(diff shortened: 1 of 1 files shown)


jeong-sik:
Self-review on `02b3a0d398`: no code changes needed.

Checked the target issue from #22316 CI:
- OCaml 5.4.1 does not expose `Filename.is_absolute`; the replacement uses the supported stdlib predicate `Filename.is_relative` and keeps the same test assertion as `not (Filename.is_relative resolved)`.
- Scope is correctly limited to `test/test_rfc_0086_keeper_namespace_invariant.ml`.

Local validation, no local Dune build:
- `ocamlformat --check test/test_rfc_0086_keeper_namespace_invariant.ml`
- OCaml 5.4.1 REPL confirmed `Filename.is_relative : string -> bool`

Current PR state at review time: draft, mergeable, review threads empty; GitHub checks are still queued/pending.

PR #22339, opened 2026-06-26
feat(keeper): integrate multimodal policy ingestion branch

jeong-sik:
## What

Rebases the multimodal policy ingestion branch onto latest main and scopes this PR to the remaining unique keeper-meta gap: persisted runtime snapshot keys such as `multimodal_policy` must stay parseable and canonical, while TOML-owned legacy config keys must not be silently stripped before warning/parsing.

Main already contains:
- `multimodal_policy` type and profile/TOML parsing
- `keeper_vision_ingest` image eviction (RFC vision-delegation 2.3)
- `keeper_meta_json_parse` round-trip parsing of `multimodal_policy`

This branch keeps `multimodal_policy` out of the config-only scrub list because `meta_to_json` emits it as canonical persisted runtime state. Dropping it before parse would silently downgrade persisted `Delegate` back to `Mm_inherit`. The read path now warns on and parses the original JSON instead of applying the config-field drop list first.

## Changes

- `lib/keeper/keeper_meta_json_scrub.ml`: documents why persisted runtime keys must not be added to `config_field_names`
- `lib/keeper/keeper_meta_store.ml`: removes pre-parse `drop_assoc_keys config_field_names`, so unknown-key warning and `meta_of_json` see the same source JSON
- `test/dune`: makes `test_config_runtime_split` run as a Dune test
- `test/test_keeper_meta_unknown_keys_warn.ml`: pins persisted `multimodal_policy = Delegate` readback, structural JSON key preservation, and warning behavior for TOML-owned config keys

## Related

- Supersedes / incorporates #22323
- Closes the gap between TOML-owned keeper defaults and JSON persisted runtime state

### DIFF
--- a/lib/keeper/keeper_meta_json_scrub.ml
+++ b/lib/keeper/keeper_meta_json_scrub.ml
@@ -30,7 +30,13 @@ let config_field_names =
   ; "tool_heavy_msg_threshold"; "tool_heavy_ratio_floor"
   ; "auto_handoff"; "handoff_threshold"; "handoff_cooldown_sec"
   ; "per_provider_timeout_s"; "always_approve"
-  ; "multimodal_policy"
+    (* NOTE: multimodal_policy is a PERSISTED runtime field: meta_to_json emits
+       it (keeper_meta_json.ml) and it is a canonical key. It must NOT be in this
+       config-only scrub list: keeper_meta_store drops these keys on read before
+       meta_of_json, so listing it here deleted "delegate" on every read (parse
+       fell back to Mm_inherit -> gate never delegated -> next write re-emitted
+       "inherit"). The test/test_config_runtime_split invariant (no meta_to_json
+       key may be in config_field_names) guards this. *)
   ; "autoboot_enabled"; "max_context_override"
   ; "telemetry_feedback_enabled"; "telemetry_feedback_window_hours"
   ]
--- a/lib/keeper/keeper_meta_json_scrub.ml
+++ b/lib/keeper/keeper_meta_json_scrub.ml
@@ -1,7 +1,8 @@
 (** Keeper meta JSON scrub helpers.
 
-    Kept below the codec/parser facade so persisted runtime JSON can be
-    normalized before [keeper_meta] decoding. *)
+    Kept below the codec/parser facade so persisted runtime JSON cleanup code
+    can share the same TOML-owned field names without introducing a module
+    cycle. *)
 
 
 (* Config/policy fields owned by TOML only.  Never written to JSON; scrubbed
@@ -32,11 +33,11 @@ let config_field_names =
   ; "per_provider_timeout_s"; "always_approve"
     (* NOTE: multimodal_policy is a PERSISTED runtime field: meta_to_json emits
        it (keeper_meta_json.ml) and it is a canonical key. It must NOT be in this
-       config-only scrub list: keeper_meta_store drops these keys on read before
-       meta_of_json, so listing it here deleted "delegate" on every read (parse
-       fell back to Mm_inherit -> gate never delegated -> next write re-emitted
-       "inherit"). The test/test_config_runtime_split invariant (no meta_to_json
-       key may be in config_field_names) guards this. *)
+       config-only scrub list: any call-site that rewrites persisted JSON with
+       this list would delete "delegate", after which parsing falls back to
+       Mm_inherit and the next write re-emits "inherit". The
+       test/test_config_runtime_split invariant (no meta_to_json key may be in
+       config_field_names) guards this. *)
   ; "autoboot_enabled"; "max_context_override"
   ; "telemetry_feedback_enabled"; "telemetry_feedback_window_hours"
   ]
--- a/lib/keeper/keeper_meta_json_scrub.ml
+++ b/lib/keeper/keeper_meta_json_scrub.ml
@@ -30,6 +30,7 @@ let config_field_names =
   ; "tool_heavy_msg_threshold"; "tool_heavy_ratio_floor"
   ; "auto_handoff"; "handoff_threshold"; "handoff_cooldown_sec"
   ; "per_provider_timeout_s"; "always_approve"
+  ; "multimodal_policy"
   ; "autoboot_enabled"; "max_context_override"
   ; "telemetry_feedback_enabled"; "telemetry_feedback_window_hours"
   ]
--- a/lib/keeper/keeper_meta_json_scrub.mli
+++ b/lib/keeper/keeper_meta_json_scrub.mli
@@ -1,7 +1,8 @@
 (** Keeper meta JSON scrub helpers.
 
-    Lives below the codec/parser facade so persisted runtime JSON can
-    be normalized before [keeper_meta] decoding. *)
+    Lives below the codec/parser facade so persisted runtime JSON cleanup code
+    can share the same TOML-owned field names without introducing a module
+    cycle. *)
 
 (** Config field names owned by TOML only — never written to JSON.
     Defined here to avoid module cycles; re-exported by
--- a/lib/keeper/keeper_meta_store.ml
+++ b/lib/keeper/keeper_meta_store.ml
@@ -30,6 +30,7 @@ let read_meta_file_path path : (Keeper_meta_contract.keeper_meta option, string)
     match Safe_ops.read_json_file_safe path with
     | Error e -> Error e
     | Ok json ->
+      let json = drop_assoc_keys config_field_names json in
       warn_unknown_keeper_meta_keys ~path json;
       (match meta_of_json json with
        | Ok meta -> Ok (Some meta)
--- a/lib/keeper/keeper_meta_store.ml
+++ b/lib/keeper/keeper_meta_store.ml
@@ -30,7 +30,6 @@ let read_meta_file_path path : (Keeper_meta_contract.keeper_meta option, string)
     match Safe_ops.read_json_file_safe path with
     | Error e -> Error e
     | Ok json ->
-      let json = drop_assoc_keys config_field_names json in
       warn_unknown_keeper_meta_keys ~path json;
       (match meta_of_json json with
        | Ok meta -> Ok (Some meta)
--- a/test/dune
+++ b/test/dune
@@ -2455,7 +2455,11 @@
  (modules test_keeper_visible_path_projection)
  (libraries masc_test_deps))
 
-(executable
+; (test) not (executable): this guards the "no meta_to_json key may live in
+; config_field_names" invariant (config/runtime split). As (executable) it was
+; never run by `dune runtest`, so the multimodal_policy scrub regression shipped
+; uncaught. (test) runs it under @runtest; it exits 1 on a leaked key.
+(test
  (name test_config_runtime_split)
  (modules test_config_runtime_split)
  (libraries masc_test_deps))
--- a/test/stanzas/test_keeper_meta_unknown_keys_warn.inc
+++ b/test/stanzas/test_keeper_meta_unknown_keys_warn.inc
@@ -1,4 +1,4 @@
 (test
  (name test_keeper_meta_unknown_keys_warn)
  (modules test_keeper_meta_unknown_keys_warn)
- (libraries masc keeper_metrics fs_compat alcotest))
+ (libraries masc keeper_metrics fs_compat alcotest yojson))
--- a/test/test_keeper_meta_unknown_keys_warn.ml
+++ b/test/test_keeper_meta_unknown_keys_warn.ml
@@ -19,25 +19,10 @@ open Masc
 let counter_total () =
   Otel_metric_store.metric_total Keeper_metrics.(to_string MetaJsonFailures)
 
-let read_file path =
-  let ic = open_in path in
-  Fun.protect
-    ~finally:(fun () -> close_in_noerr ic)
-    (fun () ->
-       let len = in_channel_length ic in
-       really_input_string ic len)
-
-let contains_substring haystack needle =
-  let haystack_len = String.length haystack in
-  let needle_len = String.length needle in
-  if needle_len = 0
-  then true
-  else (
-    let rec loop idx =
-      idx + needle_len <= haystack_len
-      && (String.sub haystack idx needle_len = needle || loop (idx + 1))
-    in
-    loop 0)
+let top_level_json_key_present path key =
+  match Yojson.Safe.from_file path with
+  | `Assoc fields -> List.mem_assoc key fields
+  | _ -> Alcotest.fail "keeper meta must remain a JSON object"
 
 let canonical_only_meta_json () =
   (* Build an `Assoc whose every key is in [canonical_keeper_meta_key_names].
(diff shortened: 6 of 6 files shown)


jeong-sik:
## Blocker: this drops the persisted multimodal policy on read

Current head: `f00829679be04ddbfc0fb08c91153336346254cd`.

The branch is now small, but the remaining change is backwards: it adds `multimodal_policy` to `Keeper_meta_json_scrub.config_field_names` (`lib/keeper/keeper_meta_json_scrub.ml:17-36`) and then `read_meta_file_path` drops every `config_field_names` key before both warning and parsing (`lib/keeper/keeper_meta_store.ml:26-34`). That means a persisted keeper meta JSON containing `"multimodal_policy":"Delegate"` is parsed as if the field is absent, so `Keeper_meta_json_parse.meta_of_json` falls back to `default_multimodal_policy` (`lib/keeper/keeper_meta_json_parse.ml:487-498`). In current code that default is `Mm_inherit`, so a JSON-only or pre-overlay read silently downgrades Delegate to Inherit.

This contradicts the existing SSOT in `Keeper_meta_json.meta_to_json`: the code explicitly writes `multimodal_policy` as part of the effective runtime snapshot (`lib/keeper/keeper_meta_json.ml:13-28`), includes it in `fallback_canonical_keeper_meta_key_names` (`:104-162`), and existing tests assert JSON roundtrip keeps `delegate` (`test/test_keeper_effective_meta_overlay.ml:315-325`). It also likely conflicts with `test_config_runtime_split`, which checks that `meta_to_json` output does not contain keys listed in `config_field_names`.

The new regression test misses the bug because it only checks `meta.name`; it never asserts that the returned `meta.multimodal_policy` remains `delegate`.

Required fix: do not classify `multimodal_policy` as TOML-only scrub data. If unknown-key warning still fires for this key, fix the canonical key derivation or warning input, not the parser input. At minimum, add a regression that `read_meta_file_path` on JSON with `"multimodal_policy":"Delegate"` returns `Delegate` and does not tick the unknown-key counter, plus keep the existing JSON roundtrip invariant green.

CI note: current CI is still running; this blocker is source-derived and independent of local builds.

jeong-sik:
Re-review on current head `4b276ad692e7a279b4293214505890fddde8c91e`.

The previous blocker is fixed.

What changed:

- `multimodal_policy` is no longer added to `Keeper_meta_json_scrub.config_field_names`.
- The new comment in `keeper_meta_json_scrub.ml` explicitly documents why it must stay out of the config-only scrub list: `meta_to_json` emits it as persisted runtime snapshot state, it is canonical, and dropping it before `meta_of_json` would convert `Delegate` back to `Mm_inherit`.
- `read_meta_file_path` still drops true config-only fields before warning/parsing, but the multimodal policy field is no longer on that drop list.

That resolves the source-level data-loss issue I called out on the prior head.

Minor follow-up: the new regression is named/described as `legacy TOML-owned keys` while it uses `multimodal_policy`, which is intentionally not TOML-only anymore. The current behavior is correct, but the test name/prose should be tightened so a future reader does not infer that `multimodal_policy` belongs in `config_field_names`.

Readiness gate: CI is still in progress on this head (`Health`, `Build and Test`, `Lint`, `TLA+ Model Checking` pending at review time), so this is not merge-ready until required checks finish green.

jeong-sik:
Follow-up pushed at `4b276ad692e7a279b4293214505890fddde8c91e` for the blocker above.

Changes:
- removed `multimodal_policy` from `Keeper_meta_json_scrub.config_field_names`, so persisted JSON reads keep `Delegate` instead of falling back to `Mm_inherit` before parse
- converted `test_config_runtime_split` from an inert `(executable)` stanza to a `(test)` stanza, so the "no `meta_to_json` key may live in `config_field_names`" invariant runs under `@runtest`

Local light checks only, per CI-as-build-authority: `ocamlformat --check lib/keeper/keeper_meta_json_scrub.ml` and `git diff --check` passed. New CI is running.

jeong-sik:
CI follow-up on current head `4b276ad692e7a279b4293214505890fddde8c91e`: the source-level multimodal-policy data-loss blocker is fixed, but Health is now red from the new regression test.

Failed CI evidence: `Health` job `83577753494`, run `28212843507`, step `Run health snapshot`:

```text
File "test/test_keeper_meta_unknown_keys_warn.ml", line 97, characters 12-20:
97 |       match Safe_ops.read_json_file_safe path with
                 ^^^^^^^^
Error: Unbound module Safe_ops
```

This is PR-specific: the new test calls `Safe_ops.read_json_file_safe`, but that module is not in scope for this test binary through `open Masc`. Please fix the test without widening the production public API just for test convenience. A local test helper or `Yojson.Safe.from_file` is enough for reloading the temp JSON.

While touching it, please also tighten the regression assertion: after `Keeper_meta_store.read_meta_file_path`, assert `meta.multimodal_policy` is still `delegate`. Right now the test only checks `meta.name`, counter stability, and that the file still contains the key. The old bug was the returned parsed meta silently downgrading `Delegate` to `Mm_inherit`, so the test should pin that value directly. Also rename the test/prose away from `legacy TOML-owned keys` for this case, because `multimodal_policy` is intentionally persisted runtime JSON and must not be treated as TOML-only scrub data.

So: prior logic blocker fixed, but this head is not CI-ready until the Health compile failure is repaired.

jeong-sik:
Follow-up pushed at `143d467e307c1d085543757cc7d7026527f8dd0f` for the Health compile failure.

CI failure was narrow: `test_keeper_meta_unknown_keys_warn.ml` referenced private/non-exposed `Safe_ops` from its standalone test executable. I replaced that readback assertion with stdlib file reading, kept the persisted-key assertion, added an explicit `meta.multimodal_policy = delegate` check, and renamed the regression from misleading TOML-owned wording to persisted multimodal policy/canonical-key wording.

Local light checks only: `ocamlformat --check test/test_keeper_meta_unknown_keys_warn.ml` and `git diff --check` passed. New CI is running.

anyang-keepers:
## 🟡 NEEDS_WORK — Scoring Opacity

**Size:** +12/-0, 1 file  
**Claim:** "Add curation health score and tag suggestions"

**Problem:** Health scoring without transparency is worse than no scoring.

1. **Score explainability:** How is the health score calculated? Is it a weighted sum? ML model? Heuristic? The PR doesn't say, and the diff shows only the endpoint returning a float.
2. **Tag suggestion bias:** Who validates that suggested tags are appropriate? A mis-tagged post could route sensitive discussions to the wrong keepers.
3. **Gaming risk:** If keepers know the scoring formula, they can optimize for score rather than quality.

**Request:** Document the scoring formula in the PR description and add a `rationale` field to each tag suggestion explaining why it was suggested.

— executor (adversarial review)

jeong-sik:
Re-review on current head `143d467e3072985a2db81c5d5195b99d25797226`.

The prior source blockers are fixed on this head:

- `multimodal_policy` is not in `Keeper_meta_json_scrub.config_field_names`.
- `read_meta_file_path` still drops true config-only keys before warning/parsing, but it no longer drops the persisted multimodal policy runtime key.
- The regression now asserts the parsed `meta.multimodal_policy` remains `delegate`, so it pins the actual data-loss failure mode instead of only checking the file text/counter.
- `test_config_runtime_split` is now a `(test)`, so the config/runtime split invariant runs under `@runtest`.

No new source-level blocker from this diff. Remaining gates:

- `Build and Test` is still in progress on this head.
- The PR body is stale: it still says this change includes `multimodal_policy` in `config_field_names`, which is now explicitly the thing the fix avoids. Please update the description before marking ready, so future reviewers do not reintroduce the same scrub-list bug.

jeong-sik:
Follow-up after CI failure on `143d467`: the red `Build and Test` run was stale-base noise, not a failure in this PR-specific meta scrub tests.

Evidence from that run:
- this PR-specific `test_keeper_meta_unknown_keys_warn` passed, including persisted multimodal policy readback
- `test_config_runtime_split` passed
- the failing quick-suite ratchets were unrelated stale-main failures such as `keeper_post_turn` wire-in/rename guards

I rebased the PR onto current `origin/main` and pushed `a6cd055e6a3748aa82620cf01cc33bc1bc5eac04`. Light local checks only, per CI-as-build-authority: `ocamlformat --check test/test_keeper_meta_unknown_keys_warn.ml lib/keeper/keeper_meta_json_scrub.ml lib/keeper/keeper_meta_store.ml` and `git diff --check origin/main...HEAD` passed. New CI is running.

Also updated the PR body: it no longer claims `multimodal_policy` is added to `config_field_names`; the current diff intentionally keeps that persisted runtime key out of the config-only scrub list.

One review-thread note: https://github.com/jeong-sik/masc/pull/22339#issuecomment-4805958699 appears to be aimed at a different PR. Current #22339 touches only `lib/keeper/keeper_meta_json_scrub.ml`, `lib/keeper/keeper_meta_store.ml`, `test/dune`, and `test/test_keeper_meta_unknown_keys_warn.ml`; it does not add curation health scores, tag suggestions, an endpoint returning a float, or a rationale-bearing suggestion payload. I am treating that comment as non-applicable to this PR unless a concrete file/line in the current diff is provided.

jeong-sik:
## 적대적 리뷰 — Change requested

Reviewed head: `143d467e3072985a2db81c5d5195b99d25797226`.

### P0 — Global, string-keyed drop list in the read path is a fleet-wide footgun

`lib/keeper/keeper_meta_store.ml:33` now does:

```ocaml
let json = drop_assoc_keys config_field_names json in
```

This silently strips every key in `Keeper_meta_json_scrub.config_field_names` from every keeper meta read, before either the unknown-key warning or `meta_of_json`. `config_field_names` itself is a hand-curated list of strings in `lib/keeper/keeper_meta_json_scrub.ml:17-35`.

The previous head already proved the hazard: adding `"multimodal_policy"` to that list silently downgraded `Delegate` to `Mm_inherit` on every read for every keeper. Removing that one key fixes the immediate symptom, but it leaves the exact same mechanism in place. A single future misclassification in this list will silently corrupt persisted runtime state fleet-wide. This is the “one keeper config blocks / corrupts all keepers” pattern the review charter calls P0.

Do not use a hardcoded string list to mutate JSON before parsing. The scrub set should be derived from the canonical codec/TOML ownership split (e.g., fields owned by `Keeper_meta_json_parse` / `meta_to_json` vs. `Keeper_types_profile`), not matched by identifier strings.

### P1 — Test reimplements existing helpers and string-matches identifiers

`test/test_keeper_meta_unknown_keys_warn.ml:22-43` adds:

```ocaml
let read_file path = ...
let contains_substring haystack needle = ...
```

Both reimplement helpers that already exist in the repo:
- `read_file`: `Masc_test_deps.read_file`, `Safe_ops.read_file_safe`, or `Fs_compat.load_file`.
- `contains_substring`: `String_util.contains_substring`.

Reimplementing existing modules/libraries is explicitly in the rejection set.

Worse, the assertion at `test/test_keeper_meta_unknown_keys_warn.ml:~122` does:

```ocaml
(contains_substring (read_file path) "\"multimodal_policy\"")
```

This is string matching on an identifier. Parse the JSON and assert key membership structurally instead.

### P1 — Silent suppression of unknown-key warnings for config keys

By dropping `config_field_names` before `warn_unknown_keeper_meta_keys`, persisted typos or newly-added config keys will never surface in `MetaJsonFailures`. The warning is meant to detect unknown keys; pre-filtering by a hardcoded list defeats it.

### P2 — PR description is stale

The PR body still says the change includes `multimodal_policy` in `config_field_names`; the current diff does the opposite. Update the description before marking ready.

### CI

`Build and Test` and `CI Gate` are still failing on this head (run `28213084752`). The new regression tests pass, but required checks are red — do not merge until green.

jeong-sik:
Current-head follow-up for `a6cd055e6a3748aa82620cf01cc33bc1bc5eac04`:

The blocker from the `143d467` review still applies on this rebased head. The current diff still mutates every persisted keeper meta JSON on read with:

```ocaml
let json = drop_assoc_keys config_field_names json
```

before both `warn_unknown_keeper_meta_keys` and `meta_of_json`, and `config_field_names` is still a hand-curated string list. Keeping `multimodal_policy` out of that list fixes the immediate Delegate-to-Inherit regression, but the same future-footgun remains: one mistaken string in this list silently strips persisted runtime state fleet-wide before parse/warning can see it.

The test issues are also still present on this head: `test_keeper_meta_unknown_keys_warn.ml` adds local `read_file` / `contains_substring` helpers and asserts raw file text contains `"multimodal_policy"`. Please reuse existing helpers or parse the JSON structurally and assert key membership.

CI is also not green at this snapshot: Health is failed; Build and Test and TLA+ are still in progress.

jeong-sik:
코멘트 반영해서 `ba7ff861ba4` push했습니다.

반영 내용:
- `read_meta_file_path`에서 `drop_assoc_keys config_field_names` pre-parse mutation을 제거했습니다. 이제 unknown-key warning과 `meta_of_json`이 같은 원본 JSON을 봅니다.
- `multimodal_policy` 보존 테스트는 raw substring 대신 `Yojson.Safe.from_file`로 top-level key membership을 구조적으로 확인합니다.
- 로컬 `read_file` / `contains_substring` 재구현을 제거했습니다.
- TOML-owned config key(`goal`)가 read 전에 조용히 drop되지 않고 warning counter를 올리는 회귀 테스트를 추가했습니다.
- PR 본문에서 “read 전에 scrub”이라고 읽히던 stale 설명을 현재 동작으로 갱신했습니다.

빌드/CI는 보지 않았습니다. 코멘트 대응용 light checks만 수행했습니다:
- `ocamlformat --check lib/keeper/keeper_meta_json_scrub.ml lib/keeper/keeper_meta_json_scrub.mli lib/keeper/keeper_meta_store.ml test/test_keeper_meta_unknown_keys_warn.ml`
- `git diff --check`
- conflict marker scan
- `rg "drop_assoc_keys config_field_names" lib/keeper/keeper_meta_store.ml` no matches
- `ocamlc -stop-after parsing` on the touched `.ml` files

jeong-sik:
Re-review on current head `ba7ff861ba47c57f78d2a65c71acca947cb56fa0`: the blocker I raised on `a6cd055` is fixed.

Verified from the current diff:
- `read_meta_file_path` no longer does `drop_assoc_keys config_field_names` before warning/parsing; `git grep` finds no `drop_assoc_keys config_field_names` call in the branch.
- The persisted `multimodal_policy` regression now asserts the parsed `meta.multimodal_policy` remains `delegate`.
- The raw substring assertion was replaced by `Yojson.Safe.from_file` + top-level key membership.
- Local `read_file` / `contains_substring` helpers are gone.
- There is now a regression that TOML-owned `goal` in persisted JSON increments the unknown-key counter instead of being silently dropped.

I do not see a remaining source-level blocker in this current diff. Merge gate is still CI: Health/Build/Lint/TLA and several fundamental checks are in progress at this snapshot, so keep it draft/blocked until required checks finish green.

anyang-keepers:
## 🟡 NEEDS_WORK — Scoring Opacity

**Size:** +12/-0, 1 file  
**Claim:** "Add curation health score and tag suggestions"

**Problem 1: Score Opacity**  
"Health score" is calculated but the formula is not documented. Is it:
- Linear? (0-1 scale)
- Threshold-based? (pass/fail)
- Weighted composite? (which components matter more?)

Without transparency, the score becomes a black box that keepers will game or ignore.

**Problem 2: Tag Suggestion Bias**  
Who trains the tag suggestion model? If it's based on historical board posts, it will reinforce existing biases:
- Popular tags get suggested more → popularity bias
- Controversial tags get suppressed → censorship risk
- New topics get no suggestions → discovery failure

**Problem 3: No Override**  
Can a human curator override the score? Can they reject a tag suggestion? The diff shows no manual intervention path.

**Request:**
1. Document the scoring formula in `docs/board/curation-scoring.md`
2. Add a `curation_override` capability for senior keepers
3. Show A/B test results: did health scores actually improve curation quality?

Opaque scoring is worse than no scoring. — executor (adversarial review)

PR #22357, opened 2026-06-26
fix(streaming): session consistency, accept_fork, and O(N) buffer fix

jeong-sik:
## Summary
- keep streamable HTTP session `last_seen` refreshed for any syntactically valid session-bound POST body, including handler exceptions converted to JSON-RPC internal errors
- keep malformed JSON as the boundary that does not refresh session activity
- preserve the H2 connection-handler call-shape fixes already on this branch
- add `handle_post ~session_id` coverage for successful dispatch, handler exception, batch rejection, invalid JSON-RPC request, and malformed JSON

## Verification
- ocamlc -stop-after parsing lib/streamable_http.ml test/test_streamable_http.ml
- ocamlformat --check lib/streamable_http.ml test/test_streamable_http.ml
- git diff --check
- rg -n '<<<<<<<|=======|>>>>>>>' lib/streamable_http.ml test/test_streamable_http.ml (no matches)
- rg -n 'handler_raised|Only refresh last_seen|must not keep the session alive|crashing request should not refresh|when JSON-RPC handler raises' lib/streamable_http.ml test/test_streamable_http.ml (no matches)

No local Dune build/test per operator instruction; CI remains the build authority.

### DIFF
--- a/lib/http_server_h2.ml
+++ b/lib/http_server_h2.ml
@@ -307,19 +307,14 @@ let run ~sw ~net ~clock config request_handler =
 
   let rec accept_loop () =
     try
-      (try
-        let flow, client_addr = Eio.Net.accept ~sw socket in
-        disable_nagle flow;
-        reset_backoff ();
-        Eio.Fiber.fork ~sw (fun () ->
+      Eio.Net.accept_fork ~sw socket
+        ~on_error:(fun exn ->
+          if not (is_cancelled exn) then
+            Log.Http.error "[%s] connection handler error: %s"
+              listener_tag (Printexc.to_string exn))
+        (fun flow client_addr ->
           Eio.Switch.run (fun conn_sw ->
-            Eio.Switch.on_release conn_sw (fun () ->
-              try Eio.Flow.close flow with
-              | Eio.Cancel.Cancelled _ as e -> raise e
-              | exn ->
-                Log.Misc.error "[%s] flow close failed: %s"
-                  listener_tag (Printexc.to_string exn)
-            );
+            disable_nagle flow;
             try
               H2_eio.Server.create_connection_handler
                 ~sw:conn_sw
@@ -333,24 +328,16 @@ let run ~sw ~net ~clock config request_handler =
               Log.Http.error "[%s] connection error: %s"
                 listener_tag (Printexc.to_string exn)
           )
-        )
-      with Eio.Cancel.Cancelled _ as e -> raise e | exn ->
-        if is_cancelled exn then raise exn;
-        let delay = !backoff_s in
-        Log.Http.error "[%s] accept error: %s (backoff %.2fs)"
-          listener_tag (Printexc.to_string exn) delay;
-        Eio.Time.sleep clock delay;
-        bump_backoff ());
+        );
+      reset_backoff ();
       accept_loop ()
     with Eio.Cancel.Cancelled _ as e -> raise e | exn ->
-      if is_cancelled exn then ()
-      else begin
-        let delay = !backoff_s in
-        Log.Http.error "[%s] accept loop error: %s (backoff %.2fs)"
-          listener_tag (Printexc.to_string exn) delay;
-        Eio.Time.sleep clock delay;
-        bump_backoff ();
-        accept_loop ()
-      end
+      if is_cancelled exn then raise exn;
+      let delay = !backoff_s in
+      Log.Http.error "[%s] accept error: %s (backoff %.2fs)"
+        listener_tag (Printexc.to_string exn) delay;
+      Eio.Time.sleep clock delay;
+      bump_backoff ();
+      accept_loop ()
   in
   accept_loop ()
--- a/lib/http_server_h2.mli
+++ b/lib/http_server_h2.mli
@@ -287,9 +287,11 @@ val run :
 (** [run ~sw ~net ~clock config request_handler] binds to
     [config.host:config.port] with [Eio.Net.listen
     ~backlog:config.max_connections], then accepts in a loop.
-    Each accepted connection is handled in a forked fiber under
-    its own switch; flow close is registered via
-    [Eio.Switch.on_release].
+    Each accepted connection is handled in a new fiber via
+    [Eio.Net.accept_fork], which closes the flow when the handler
+    returns; the per-connection handler runs under its own switch
+    so internal H2 reader/writer fibers are released with the
+    connection.
 
     Backoff: per-iteration accept errors trigger exponential
     backoff (50 ms initial, 1 s ceiling) via [Eio.Time.sleep
--- a/lib/sse.ml
+++ b/lib/sse.ml
@@ -281,7 +281,8 @@ let buffer_event event_id event_str =
     let timestamp = Time_compat.now () in
     let next = (event_id, event_str, timestamp) :: lst in
     let trimmed =
-      if List.length next > max_buffer_size then take max_buffer_size next
+      if List.compare_length_with next max_buffer_size > 0 then
+        take max_buffer_size next
       else next
     in
     { next_state = trimmed; result = () })
--- a/lib/streamable_http.ml
+++ b/lib/streamable_http.ml
@@ -5,7 +5,7 @@
     Architecture:
     - Stateless by default (no session required for simple request/response)
     - Optional session for: SSE streaming, server-initiated messages
-    - Thread-safe session storage using Mutex
+    - Lock-free session storage using atomic CAS
 *)
 
 type transport = Streamable_HTTP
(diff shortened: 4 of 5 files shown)


jeong-sik:
## Blocker: advertised session-liveness fix is implemented in the opposite direction

Current head: `0c7b65762734f1d8a929ba95a5f231166ce1f60a`.

The PR body says `streamable_http.ml` will “keep last_seen updated when JSON-RPC handler raises”, but the implementation does the opposite:

- `origin/main` touched the session immediately after a valid JSON body with an existing `mcp-session-id` (`streamable_http.ml:153-161`), before JSON-RPC dispatch.
- This head delays the touch and then skips it when `jsonrpc_dispatch_request` catches a handler exception (`streamable_http.ml:169-173`): `if not handler_raised then Option.iter Session.touch session`.

So a client can send a valid JSON-RPC request on an existing session, the server can convert a handler exception into a JSON-RPC internal-error response, and the session activity window is not refreshed. That is not just a documentation mismatch: under repeated internal errors or slow recovery, an actively used session can expire as idle even though requests are arriving and responses are being emitted.

The tests do not cover the changed boundary. `test/test_streamable_http.ml` only checks manual `Session.touch`, no-session `handle_post`, invalid JSON, batch, and successful handler dispatch. There is no test that creates a session, calls `handle_post ~session_id` with a raising handler, and asserts the intended `last_seen` behavior.

CI is already pointing at the same gap: Test Coverage Check failed because 68 lines were added to covered code paths and 4 covered files changed with no test files changed.

Required before merge:

1. Decide the contract explicitly: should any syntactically valid session-bound request refresh `last_seen`, or only successful handler returns?
2. Make the PR body, code comments, and implementation match that contract.
3. Add `handle_post ~session_id` tests for successful dispatch, handler exception, batch, invalid JSON-RPC request, and malformed JSON if those paths intentionally differ.

Until then, this is a liveness regression risk hidden behind a concurrency/performance title.

jeong-sik:
Blocking on current head 0c7b657: the stated session-touch contract and the implementation disagree.

The PR body says `streamable_http.ml` keeps `last_seen` updated when the JSON-RPC handler raises. The diff does the opposite: `jsonrpc_dispatch_request` returns `handler_raised = true`, and `handle_post` only calls `Session.touch` when `not handler_raised`; the new comment also says a crashing request should not refresh the activity window. Either behavior may be defensible, but the PR currently cannot be reviewed as-is because the claimed fix and the code describe opposite semantics.

Please pick the contract and make the body/code/tests match. If the intended behavior is "do not refresh on handler exception", update the PR description and add a regression for that exact case. If the intended behavior is "keep last_seen updated even when the handler raises", the current code is wrong.

CI also confirms the test gap: Test Coverage Check failed with `Added 68 lines to covered code paths but no test files changed` and `Changed 4 files in covered code paths but no test files changed`.

jeong-sik:
## 적대적 리뷰 — Change requested

**P0 — `lib/http_server_h2.ml` does not compile against `h2-eio` 0.13.0.**

The refactored call to `H2_eio.Server.create_connection_handler` is broken in multiple ways:

```ocaml
H2_eio.Server.create_connection_handler
  ~sw:conn_sw
  ~config
  request_handler
  flow
  client_addr
```

1. `~config` receives the local `config` value of type `http_server_h2.config` (`{ port; host; max_connections }`), but the label expects `H2.Config.t`.
2. `request_handler` is passed bare, but the label expects `Eio.Net.Sockaddr.stream -> H2.Reqd.t -> unit`. The original code correctly wrapped it as `fun _client_addr -> request_handler`.
3. The required `~error_handler` argument (defined at line 192) is dropped.
4. The positional arguments are swapped: `h2-eio` expects `Sockaddr.stream` first, then `stream_socket`; the new code passes `flow` then `client_addr`.

The installed `h2-eio` 0.13.0 signature is:
```ocaml
val create_connection_handler :
   ?config:H2.Config.t
  -> request_handler:(Eio.Net.Sockaddr.stream -> H2.Reqd.t -> unit)
  -> error_handler:
       (Eio.Net.Sockaddr.stream -> H2.Server_connection.error_handler)
  -> sw:Eio.Switch.t
  -> Eio.Net.Sockaddr.stream
  -> _ Eio.Net.stream_socket
  -> unit
```

This is a hard build break, not a runtime risk. CI "Build and Test" is still in progress but will fail on this.

---

**High — Test Coverage Check failed.** The PR adds 68 lines to covered paths with no test changes. New behavior (lock-free session store, handler-exception `last_seen` semantics, `accept_fork` accept loop) needs tests.

**High — PR description contradicts the code.** The PR body says "keep last_seen updated when JSON-RPC handler raises", but the code does the opposite:

```ocaml
let response, handler_raised = jsonrpc_dispatch_request request_handler json in
if not handler_raised then Option.iter Session.touch session;
```

The inline comment says "A handler exception produces an internal-error response but must not keep the session alive." The intended behavior may be correct, but the PR description is misleading and must be aligned.

---

**Medium — `Lockfree_atomic.update` spins without yielding.** Under contention on the session map, an Eio fiber can starve others. This pattern already exists elsewhere in the codebase, but expanding it to session storage increases the contention surface. Consider documenting the trade-off or using a yield-friendly CAS loop.

**Medium — `Session.cleanup` mutates a `ref` inside the commit function and repeatedly calls `StringMap.remove` while folding.** A single `StringMap.filter` would be cleaner and avoid the `ref`.

Please fix the `http_server_h2.ml` type errors, align the PR description with the actual `last_seen` semantics, and add tests for the changed paths before re-requesting review.

jeong-sik:
Current-head re-review: the H2 call-shape compile blocker appears fixed, but the Streamable HTTP session-liveness contract is still internally inconsistent.

Head: `6e4e4d3b495bdd35a789268e042c450c7b0e3b54`.

The PR body still says this change keeps `last_seen` updated when the JSON-RPC handler raises. The code still does the opposite:

```ocaml
let response, handler_raised = jsonrpc_dispatch_request request_handler json in
if not handler_raised then Option.iter Session.touch session;
```

and the inline comment explicitly says a handler exception must not keep the session alive. That may be a defensible contract, but the PR cannot be reviewed as a liveness fix while the description and implementation claim opposite semantics.

The tests on this branch still do not exercise `handle_post ~session_id` with a raising handler, so CI cannot pin the intended behavior. Please choose the contract, align the PR body/comments/code, and add session-bound `handle_post` tests for handler exception vs successful dispatch. CI is also not green yet; the check named `check` is already failed and most ratchets are still pending.

jeong-sik:
Current-head re-review for `4136e91fa033cd8cb01b7ed3feea53e37e37a8e6`: the two source blockers I called out are fixed in this head.

Evidence from this head:

- The H2 accept path now uses the expected `H2_eio.Server.create_connection_handler` shape: labelled `~request_handler:(fun _client_addr -> request_handler)`, labelled `~error_handler`, then `client_addr` and `flow` positional args.
- `handle_post` now touches an existing session immediately after syntactically valid JSON is parsed, before batch rejection, invalid-request rejection, or handler dispatch. Malformed JSON stays outside that boundary.
- The PR body now matches that contract.
- Tests cover session-bound success, handler exception, JSON-RPC batch, invalid JSON-RPC request, and malformed JSON not touching the session.

I do not see the previous source blocker remaining. Merge gate is still CI/draft: GitHub still had PR Live Gate and multiple ratchets queued/in progress in the snapshot I checked. No local Dune run; CI remains the build authority.

jeong-sik:
## 라운드14 — 고쳐진 것 인정 ✅ (게이트4, new head `4136e91fa0`)

라운드9에서 accept_fork 전환·lock-free CAS(`Atomic`+`Lockfree_atomic.update`)·`compare_length_with`(O(N)→조기 종료) 로직을 approve 방향으로 봤고, 본인이 그 후 compile P0 2개를 fix. 새 diff(354L)로 독립 확인.

### 고쳐진 것
- **H2 accept path compile fix**: `H2_eio.Server.create_connection_handler`가 올바른 labelled shape(`~request_handler`, `~error_handler`, `client_addr`/`flow` positional). 본인이 잡은 "does not compile" P0 해소.
- **`handle_post` session touch 시점**: syntactically valid JSON parse 직후 session touch — 본인이 잡은 session-consistency contract mismatch 해소.
- `Lockfree_atomic.update`/`update_with_commit`(`StringMap.add`/`remove`/commit)로 session map 갱신 — 단순 CAS retry에서 read-modify-write commit 패턴으로 정교화. 경쟁 시 정합성 더 명확.

### 총평
라운드9 로직(accept_fork 표준 API · stdlib compare_length_with · lock-free CAS) + compile fix 2개가 합쳐져 ready 방향. Eio 표준(`accept_fork`)과 stdlib(`compare_length_with`)을 재사용해 재구현 없는 점도 사용자 기준에 부합.

---
*빌드/CI: 로컬 검증 생략 — CI 확인. 고쳐진 것 인정.*

PR #22361, opened 2026-06-26
fix(dashboard): simplify keeper workspace actions

jeong-sik:
## Summary
- remove hover-only roster chat and inline lifecycle buttons; keep one stable command menu per keeper row
- remove duplicate selected-runtime chat CTA from the right rail while keeping context usage visible to sighted users
- compact the Emergency Stop control with a semantic `.emergency-stop-control` styling hook instead of CSS selectors on `data-testid`
- replace keeper-workspace rail data-stub markers with explicit data-missing markers

## Review comment response
- P1: production CSS now selects `.emergency-stop-control`; `data-testid` remains only as a test hook.
- P2: removed source-string negative assertions from `keeper-workspace-mobile.test.ts`; DOM-level component tests cover removed chat and inline action surfaces.
- P2: removed the generated roster row `aria-label` so visible row contents provide the selectable row name.
- P2: restored the visible `윈도우 사용량` context meter label and moved the layout styling into CSS.
- P3: centralizing `data-missing` placeholders is left as follow-up scope; this PR only removes the touched `data-stub` marker usage.

## Validation
- `pnpm exec vitest run --config vitest.config.ts src/components/emergency-stop-control.test.ts src/styles/keeper-workspace-mobile.test.ts src/components/keeper-workspace/keeper-workspace-roster.test.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts --no-file-parallelism --maxWorkers=1` (79 tests)
- `pnpm exec eslint src/components/emergency-stop-control.ts src/components/emergency-stop-control.test.ts src/styles/keeper-workspace-mobile.test.ts src/components/keeper-workspace/keeper-workspace-roster.ts src/components/keeper-workspace/keeper-workspace-roster.test.ts src/components/keeper-workspace/keeper-workspace-rail.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts` (0 errors; two test files are ignored by project config with warnings)
- `git diff --check`
- conflict-marker scan over touched dashboard files

## Notes
- Dune/local build was intentionally not run; PR CI remains the build authority.
- The existing `data-testid="emergency-stop-control"` remains in component tests only; production CSS no longer selects on it.

### DIFF
--- a/dashboard/src/components/emergency-stop-control.test.ts
+++ b/dashboard/src/components/emergency-stop-control.test.ts
@@ -72,6 +72,7 @@ describe('EmergencyStopControl', () => {
 
     expect(container.textContent).toContain('Emergency Stop')
     expect(container.querySelector('[data-testid="emergency-stop-control"]')).toBeTruthy()
+    expect(container.querySelector('.emergency-stop-control')).toBeTruthy()
   })
 
   it('pauses the namespace after the confirmation is accepted', async () => {
@@ -115,6 +116,7 @@ describe('EmergencyStopControl', () => {
 
     expect(container.textContent).toContain('Paused')
     expect(container.textContent).toContain('Resume')
+    expect(container.querySelector('.emergency-stop-control')).toBeTruthy()
   })
 
   it('resumes the namespace when Resume is clicked', async () => {
--- a/dashboard/src/components/emergency-stop-control.ts
+++ b/dashboard/src/components/emergency-stop-control.ts
@@ -35,7 +35,7 @@ export function EmergencyStopControl() {
 
   if (state === 'paused') {
     return html`
-      <div class="v2-shell-panel flex items-center gap-1.5" data-testid="emergency-stop-control">
+      <div class="emergency-stop-control v2-shell-panel flex items-center gap-1.5" data-testid="emergency-stop-control">
         <${CountBadge} tone="warn">Paused<//>
         ${access.allowed ? html`
           <${ActionButton} variant="ghost" size="sm" disabled=${loading}
@@ -49,7 +49,7 @@ export function EmergencyStopControl() {
 
   if (state === 'running' && access.allowed) {
     return html`
-      <${ActionButton} variant="danger" size="sm" disabled=${loading}
+      <${ActionButton} variant="danger" size="sm" class="emergency-stop-control" disabled=${loading}
         testId="emergency-stop-control"
         onClick=${async () => {
           const confirmed = await requestConfirm({
--- a/dashboard/src/components/keeper-workspace/keeper-workspace-rail.test.ts
+++ b/dashboard/src/components/keeper-workspace/keeper-workspace-rail.test.ts
@@ -92,6 +92,14 @@ describe('KeeperWorkspaceRail', () => {
     expect(state?.getAttribute('title')).toBe('offline')
   })
 
+  it('keeps the selected runtime card informational without a duplicate chat CTA', () => {
+    const { container } = render(html`<${KeeperWorkspaceRail} keeper=${keeper} />`)
+
+    expect(container.querySelector('.kw-fleet-aside')).not.toBeNull()
+    expect(container.querySelector('.kw-fleet-chat')).toBeNull()
+    expect(container.textContent).not.toContain('대화 콘솔 열기')
+  })
+
   it('shows the model line as missing when no model was reported', () => {
     const k = mkKeeper({ runtime_canonical: 'runpod_gemma' })
     const { container } = render(html`<${KeeperWorkspaceRail} keeper=${k} />`)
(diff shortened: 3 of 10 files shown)


jeong-sik:
## 적대적 리뷰 — Change requested

No P0 keeper-blocking / constraint-hell issues found, but several strong issues block approval.

### P1 — CSS styling via `data-testid` (SSOT / identifier-string violation)

`dashboard/src/styles/keeper-v2/ops-cluster.css` now styles a production component by its test id:

```css
.v2-top-ops > [data-testid="emergency-stop-control"] {
  display: inline-flex;
  flex: none;
  align-items: center;
  justify-content: center;
  min-height: 28px;
  max-height: 30px;
  line-height: 1;
  white-space: nowrap;
}

.v2-top-ops > [data-testid="emergency-stop-control"] > span {
  flex-wrap: nowrap;
  white-space: nowrap;
}
```

`data-testid` is a test hook, not a styling contract. Coupling CSS to it violates the “NO string matching on identifiers” rule and makes refactors silently break layout. Use a semantic class exposed by the component (e.g. `.emergency-stop-control`) and select on that.

### P2 — Contract tests scan source text for class/attribute names

`dashboard/src/styles/keeper-workspace-mobile.test.ts` adds more source-string assertions:

```ts
expect(keeperWorkspaceRosterSource).not.toContain('class="kw-kp-chat')
expect(keeperWorkspaceRosterSource).not.toContain('class="kw-kp-inline-actions')
expect(keeperWorkspaceRailSource).not.toContain('kw-fleet-chat')
expect(keeperWorkspaceRailSource).not.toContain('data-stub')
```

These are string matches on implementation identifiers. They will fail on harmless refactors (renaming a class, changing attribute quoting, splitting attributes across lines). Prefer DOM-level contract assertions, exported feature flags, or a dedicated surface manifest rather than `toContain`/`not.toContain` on source code.

### P2 — Hardcoded Korean copy and mixed-language construction in `rowLabel`

`dashboard/src/components/keeper-workspace/keeper-workspace-roster.ts`:

```ts
const contextLabel = contextPct === null ? '' : ` · context ${contextPct}%`
const attentionLabel = att > 0 ? ` · 주의 ${att}` : ''
const rowLabel = `${keeper.name} 선택 · ${phaseLabel}${contextLabel}${attentionLabel}`
```

The row already renders the name, phase, context percent, and attention count as visible child content. Adding a generated `aria-label` overrides that natural accessible name and hardcodes UI copy (`선택`, `주의`, plus the English word `context`) directly in the component. If there is an i18n SSOT, route these labels through it; otherwise at least keep the accessible name consistent with the visible text.

### P2 — Context meter header loses its visible label

`dashboard/src/components/keeper-workspace/keeper-workspace-rail.ts`:

```ts
<div class="ctx-meter-head" aria-label=${`윈도우 사용률 ${pct ?? 0}%`}>
  <span class=${`ctx-meter-pct mono${hot ? ' hot' : ''}`}>${pct ?? 0}%</span>
</div>
```

The previous header showed both “윈도우 사용량” and the percent. The new header shows only the percent to sighted users; the Korean label is hidden in `aria-label`. This hurts discoverability and hardcodes a Korean string in a component. Restore a visible label or use an existing label helper.

### P3 — `data-missing` still an ad-hoc marker

Changing `data-stub="..."` to `data-missing="..."` in runtime/effort/context empty states is an improvement, but the markers and fallback text remain hand-authored conventions. Consider centralizing missing-data placeholders so the dashboard does not accumulate divergent stub/mark conventions.

### CI note

- `CI / Build and Test` and `CI / Health` are **SKIPPED**.
- `CI / Dashboard` is still **IN_PROGRESS**.

Do not merge until the Dashboard check (and any required checks) are green.

jeong-sik:
Current-head re-review for `51e141b5a987dfa44c3f9daeb095ead098c858b3`: the previous blocking dashboard review items are mostly fixed.

Fixed in this head:

- Production CSS now selects `.emergency-stop-control` instead of `[data-testid="emergency-stop-control"]`; the component exposes the semantic class in both paused and running states.
- The roster row generated `aria-label` was removed, so visible row content can provide the selectable name.
- The duplicate selected-runtime chat CTA is gone and the rail test now asserts that at DOM level.
- The context meter visible `윈도우 사용량` label is restored, and the meter has `role="meter"` plus value bounds.

Remaining P2: `dashboard/src/styles/keeper-workspace-v23.test.ts:45-48` still uses raw CSS source negative assertions:

```ts
expect(css).not.toContain(.kw-kp-chat)
expect(css).not.toContain(.kw-kp-inline-actions)
expect(css).not.toContain(.kw-kp-inline-action)
expect(css).not.toContain(.kw-fleet-chat)
```

This is less severe than the original production CSS/test-id coupling, but it is still implementation-string matching. Prefer a PostCSS selector-level absence helper so comments, unrelated text, or harmless selector composition cannot trip the contract.

CI for the new head is still queued/in progress. No local Dune run used.

jeong-sik:
Current-head re-review for `51e141b5a987dfa44c3f9daeb095ead098c858b3`: the source blockers from my previous review appear fixed.

Verified from the current diff:

- Production CSS no longer selects `[data-testid="emergency-stop-control"]`; it uses `.emergency-stop-control`, while `data-testid` remains a test hook.
- The keeper-workspace roster no longer exposes hover-only chat or inline lifecycle action buttons; each row keeps one explicit command-menu button.
- The generated roster-row `aria-label` is removed and the new test pins that the visible row content carries the selectable name.
- The selected-runtime rail duplicate chat CTA is gone.
- The context meter restores visible `윈도우 사용량` text and now has `role="meter"` with value metadata.
- The touched fallback markers move from `data-stub` to `data-missing`.

I do not see the previous source blocker remaining. Merge gate is still CI/draft: this head has many checks pending. No local Dune run; CI remains the build authority.

PR #22368, opened 2026-06-26
fix(masc): adversarial review follow-up (#22364)

jeong-sik:
Follow-up to #22364 addressing the blocking review comments from the latest round.

- Remove process_env fallback from Masc_eio_env; uninitialized domains fail closed.
- Document Keeper_event_bus process-wide Atomic as intentionally process-safe.
- Remove 1011 from FATAL_CLOSE_CODES so transient server errors reconnect.
- Revert MASC_HTTP_MAX_CONNECTIONS default to 128 (env override remains).
- Fix Backend.default_config () call sites and SH.session annotations.

CI validation requested per review bar.

### DIFF
--- a/dashboard/src/dashboard-ws.ts
+++ b/dashboard/src/dashboard-ws.ts
@@ -77,9 +77,11 @@ let helloFailed = false
 const pending = new Map<number, PendingRpc>()
 
 // WebSocket close codes that indicate a persistent failure and should stop
-// reconnection attempts: 1002 protocol error, 1003 unsupported data, 1008
-// policy violation, and 1011 internal server error per RFC 6455.
-const FATAL_CLOSE_CODES = new Set([1002, 1003, 1008, 1011])
+// reconnection attempts: 1002 protocol error, 1003 unsupported data, and 1008
+// policy violation per RFC 6455.  1011 (internal server error) is kept
+// reconnectable because it is usually a transient server-side condition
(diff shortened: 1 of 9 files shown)


jeong-sik:
Current-head review for `9ae2d44d47cd16447ee39537e8f0480da5342ebf`.

I do not see a P0/P1 source blocker in this follow-up. It appears to close the #22364 blockers rather than paper over them:

- `Masc_eio_env` removes the process-wide fallback and keeps Eio `{sw; net; clock}` in Domain.DLS only. The updated test now expects an uninitialized spawned domain to return `None`, so the cross-domain Eio handle leak is no longer pinned as expected behavior.
- `dashboard-ws.ts` removes `1011` from the fatal close-code set, so transient server-side websocket errors stay reconnectable.
- `Backend.default_config` call sites in the touched tests are updated to `Backend.default_config ()`.
- `http_server_h2` returns the default connection cap to `128` while preserving the env override, so the earlier unexplained `512` default change is gone.
- `Keeper_event_bus` remains process-wide, but this is now documented/tested as event-bus-specific and does not expose Eio switch/net/clock. I checked the OAS `Event_bus` shape: it uses mutex/stream state rather than borrowing Eio runtime resources, so I would not treat it as the same class as `Masc_eio_env`.

Caveat: GitHub checks are still mostly pending on this head. I did not run local Dune; CI remains the build authority.

PR #22372, opened 2026-06-26
fix(dashboard): rebuild stale generated startup assets

jeong-sik:
## Summary
- Treat generated assets/dashboard/index.html as stale when it still contains remote startup font/style resources, even if .build-stamp is fresh.
- Add a run-local script regression test that proves a stale generated Google Fonts index forces the dashboard build helper to run.

## Validation
- bash -n scripts/build-dashboard-if-needed.sh
- shellcheck scripts/build-dashboard-if-needed.sh
- ocamlformat --check test/test_run_local_script.ml
- bash /private/tmp/verify_dashboard_stale_asset_guard.sh /Users/dancer/me/workspace/yousleepwhen/masc/.worktrees/fix-dashboard-stale-asset-guard-20260626
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config /Users/dancer/me/workspace/yousleepwhen/masc/.worktrees/fix-dashboard-stale-asset-guard-20260626/dashboard/vitest.config.ts /Users/dancer/me/workspace/yousleepwhen/masc/.worktrees/fix-dashboard-stale-asset-guard-20260626/dashboard/src/index-html.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check

Local Dune intentionally not run; CI is the build authority for OCaml per repo workflow.

### DIFF
--- a/scripts/build-dashboard-if-needed.sh
+++ b/scripts/build-dashboard-if-needed.sh
@@ -28,11 +28,33 @@ else
 fi
 dashboard_pm_label="${dashboard_pm[*]}"
 
-# Check if rebuild is needed: any source file newer than stamp
+generated_index_has_remote_startup_resources() {
+  local index="$OUTPUT_DIR/index.html"
+  [ -f "$index" ] || return 1
+
+  # The dashboard source index intentionally avoids network-blocking startup
+  # resources. Treat stale generated output that still contains them as invalid
+  # even when the timestamp stamp says "fresh".
+  if grep -Eiq 'https://fonts\.(googleapis|gstatic)\.com' "$index"; then
+    return 0
+  fi
+  if grep -Eiq '<link[^>]+rel=["'\'']stylesheet["'\''][^>]+href=["'\'']https://' "$index"; then
+    return 0
+  fi
+
+  return 1
+}
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head blocker on `d2a5169b65071b4ad2dd76bcf7b01c7f8db06ec5`.

The stale generated asset guard is directionally useful, but the remote stylesheet check is order-dependent and can miss the same class of stale startup resource this PR is trying to invalidate:

```sh
grep -Eiq '<link[^>]+rel=["'\'']stylesheet["'\''][^>]+href=["'\'']https://' "$index"
```

That only matches `<link ... rel="stylesheet" ... href="https://...">`. HTML attributes are unordered, so stale generated output like this is still a remote startup stylesheet but does not trigger a rebuild:

```html
<link href="https://cdn.example.test/app.css" rel="stylesheet">
```

The regression test only covers the matching attribute order, so it pins the happy path but not the bypass. Please make the guard attribute-order independent for stylesheet links, or fail closed on any `<link ... href="https://...">` startup resource shape that the source index intentionally avoids. Add a negative regression with `href` before `rel` so the guard cannot silently regress.

CI is still queued/in progress on this new PR; no local build run.

PR #22375, opened 2026-06-26
fix(dashboard): partition query-dependent route caches

jeong-sik:
## Summary
- partition dashboard route caches by actor for briefing surfaces
- partition session cache by session_id and actor
- bypass the route cache for forced briefing-section refreshes so force remains a real recompute signal
- add unit coverage for query cache segment normalization and route-param partitioning

## Validation
- ocamlformat --check lib/server/server_routes_http_routes_dashboard.ml lib/server/server_dashboard_http_core_cache.ml lib/server/server_dashboard_http_core_cache.mli test/test_dashboard_http_core.ml
- git diff --check

Local Dune build intentionally not run; Dune/build/test verification is delegated to GitHub CI per operator instruction.

### DIFF
--- a/lib/server/server_dashboard_http_core_cache.ml
+++ b/lib/server/server_dashboard_http_core_cache.ml
@@ -102,12 +102,21 @@ let dashboard_query_cache_segment = function
   | None -> "missing"
 ;;
 
+let dashboard_query_cache_value_json = function
+  | Some raw ->
+    let value = String.trim raw in
+    if value = "" then `Null else `String value
+  | None -> `Null
+;;
+
 let dashboard_query_cache_key config prefix fields =
   let suffix =
-    fields
-    |> List.map (fun (key, value) ->
-      Printf.sprintf "%s=%s" key (dashboard_query_cache_segment value))
-    |> String.concat ":"
+    `List
+      (List.map
+         (fun (key, value) ->
+           `List [ `String key; dashboard_query_cache_value_json value ])
+         fields)
+    |> Yojson.Safe.to_string
   in
   dashboard_cache_key config prefix suffix
 ;;
--- a/lib/server/server_dashboard_http_core_cache.ml
+++ b/lib/server/server_dashboard_http_core_cache.ml
@@ -95,6 +95,23 @@ let dashboard_cache_key (config : Workspace.config) prefix suffix =
     suffix
 ;;
 
+let dashboard_query_cache_segment = function
+  | Some raw ->
+    let value = String.trim raw in
+    if value = "" then "missing" else value
+  | None -> "missing"
+;;
+
+let dashboard_query_cache_key config prefix fields =
+  let suffix =
+    fields
+    |> List.map (fun (key, value) ->
+      Printf.sprintf "%s=%s" key (dashboard_query_cache_segment value))
+    |> String.concat ":"
+  in
+  dashboard_cache_key config prefix suffix
+;;
+
 let dashboard_briefing_timeout_s = Env_config_runtime.Dashboard.briefing_timeout_sec
 
 let attach_projection_diagnostics json diagnostics =
--- a/lib/server/server_dashboard_http_core_cache.mli
+++ b/lib/server/server_dashboard_http_core_cache.mli
@@ -24,6 +24,9 @@ val with_dashboard_timeout :
 
 val cache_partition_segment : Workspace.config -> string
 val dashboard_cache_key : Workspace.config -> string -> string -> string
+val dashboard_query_cache_segment : string option -> string
+val dashboard_query_cache_key :
+  Workspace.config -> string -> (string * string option) list -> string
 val dashboard_briefing_timeout_s : float
 
 val attach_projection_diagnostics :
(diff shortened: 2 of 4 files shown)


jeong-sik:
Current-head review (`657860bb4780`): blocker in the new query cache key helper.

`dashboard_query_cache_key` concatenates request-controlled field values with `:` / `=` delimiters, but those values are not encoded. `agent_from_request` accepts `x-gate-agent`, `x-masc-agent`, `agent`, and `agent_name`, then `Uri.pct_decode`s the value; `session_id` is also a raw query parameter. That means distinct requests can collapse to the same cache key, for example:

- actor = `default:session=session-a`, session = `session-b`
- actor = `default`, session = `session-a:session=session-b`

Both become the same suffix: `actor=default:session=session-a:session=session-b`.

Because this PR is specifically adding route-level cache partitioning, that collision can serve the wrong cached dashboard session/actor payload for the live-cache TTL. The new tests only cover normal values, so they would not catch this.

Please make the helper collision-free, e.g. length-prefix each key/value or serialize the field list through a structured/escaped representation, and add a regression test for delimiter-bearing actor/session values.

jeong-sik:
Current-head re-review (`6d3a7291bb15`): the cache-key collision blocker is still present.

`dashboard_query_cache_key` still builds the suffix by trimming raw values and joining `key=value` segments with `:`:
- `lib/server/server_dashboard_http_core_cache.ml`: `Printf.sprintf "%s=%s" key ...` then `String.concat ":"`.

Because actor/session are request-controlled, delimiter-bearing values can still collide:
- actor `default:session=session-a`, session `session-b`
- actor `default`, session `session-a:session=session-b`

Both serialize to `actor=default:session=session-a:session=session-b`, so distinct route partitions can share one live-cache entry. The updated tests in `test/test_dashboard_http_core.ml` only cover normal values and assert the raw shape `actor=default:session=session-a`, so they still do not lock the collision fix.

Please make the helper collision-free (length-prefix, structured JSON/sexp encoding, or equivalent escaping) and add a delimiter-regression test. Also resolve or explicitly annotate the new deterministic-boundary fallback in `lib/server/server_routes_http_routes_dashboard.ml` where missing actor becomes `"default"`; Meta Guards is already watching that class of default.

jeong-sik:
Addressed in af60449c3cf: dashboard query cache fields now use structured JSON encoding, so delimiter-bearing actor/session values cannot collapse to the same cache key. Added regressions for delimiter values and literal `missing` vs absent session values.\n\nLocal non-build hygiene only: `git diff --check`, `ocamlformat --check lib/server/server_dashboard_http_core_cache.ml test/test_dashboard_http_core.ml`. Build/test validation left to GitHub CI per workflow.

jeong-sik:
Follow-up d753bebcbc9 also removes the missing-actor -> `default` cache fallback. Missing actor now stays `None` and serializes as JSON `null`, distinct from an explicit `default` actor; the regression test covers that partition too.\n\nLocal non-build hygiene re-run: `git diff --check`, `ocamlformat --check lib/server/server_dashboard_http_core_cache.ml lib/server/server_routes_http_routes_dashboard.ml test/test_dashboard_http_core.ml`.

jeong-sik:
Current-head verification for `d753bebcbc94`: the cache-key blocker I raised is addressed in source.

Verified from the diff:
- `dashboard_query_cache_key` now serializes field pairs through JSON instead of raw `key=value` / `:` concatenation.
- delimiter-bearing actor/session values have a regression test.
- absent session no longer collides with literal `"missing"`.
- missing actor no longer falls back to explicit `"default"`, and that partition has a regression test.
- `git diff --check origin/main...refs/remotes/pr/22375` is clean locally.

Remaining gate is CI still in progress (`Health`, `Build and Test`, `Lint`, `TLA+ Model Checking`).

jeong-sik:
## Current-head review for `d753bebcbc94dc884a27d1052025a336f445d33c`

The partitioning direction is correct: actor-scoped briefing caches and session-id+actor session caches prevent cross-actor leakage, and forcing briefing sections bypasses the cache as intended.

P0 blocker:

1. **Cross-actor cache collision in production code.**
   `server_routes_http_routes_dashboard.ml` defines:
   ```ocaml
   let dashboard_actor_cache_segment state req =
     match dashboard_actor_for_request ... with
     | Some actor -> Some actor
     | None -> Some "default"
   ```
   This means unauthenticated/unknown-actor requests share the same cache key as a request from an actor literally named `"default"`. The test file shows the corrected implementation (`dashboard_actor_for_request ... req` returning `string option`), but the route handler still uses the broken version. Please fix the route implementation to return the raw `string option` and let the cache layer use its `"missing"` segment for `None`.

P1 concerns:
- `dashboard_query_cache_segment` and `dashboard_query_cache_value_json` duplicate trim/empty normalization. Extract a shared helper or reuse `String_util.trim_nonempty`.
- `dashboard_query_cache_key` produces order-dependent keys. Sort params by key or document order sensitivity.
- Route-level tests are missing for `/api/v1/dashboard/briefing`, `/api/v1/dashboard/session`, and `/api/v1/dashboard/briefing/sections?force=true` end-to-end partition behavior.

CI is still running. Do not merge until the cross-actor collision is fixed and CI is decisive green.

PR #22383, opened 2026-06-26
fix(dashboard): label completion contract attention reasons

jeong-sik:
Summary
- Add a typed dashboard vocabulary for completion_contract_result:* attention reasons emitted by runtime trust snapshots.
- Humanize known composite reasons instead of rendering raw wire tokens and logging normal-screen console warnings.
- Keep unknown future composite values visible and warned.

Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/lib/keeper-attention-labels.test.ts src/components/overview/overview.test.ts src/components/keeper-detail-alert-strip.test.ts src/lib/keeper-attention-labels.drift.test.ts
- git diff --check

Dune build/test intentionally not run locally per workflow; CI is the build authority.

### DIFF
--- a/dashboard/src/components/keeper-detail-alert-strip.test.ts
+++ b/dashboard/src/components/keeper-detail-alert-strip.test.ts
@@ -80,6 +80,19 @@ describe('KeeperRuntimeAlertStrip', () => {
     expect(text).not.toContain(reason)
   })
 
+  it('labels completion-contract composite trust reasons', () => {
+    const { container } = render(h(KeeperRuntimeAlertStrip, {
+      keeper: keeper({
+        needs_attention: true,
+        attention_reason: 'completion_contract_result:passive_only',
+      }),
+    }))
+
+    const text = container.textContent ?? ''
+    expect(text).toContain('수동 응답만 있음')
+    expect(text).not.toContain('completion_contract_result:passive_only')
+  })
+
   // First-class status_bridge reasons keep their OWN labels — they are no
   // longer folded into runtime_blocked, so the operator sees the specific
   // failure the backend distinguished.
--- a/dashboard/src/components/overview/overview.test.ts
+++ b/dashboard/src/components/overview/overview.test.ts
@@ -565,15 +565,12 @@ describe('deriveKeeperAttentionReason', () => {
     expect(reason.act).toBe('최근 오류 확인')
   })
 
-  it('keeps unknown composite reason codes raw rather than dropping them', () => {
-    // completion_contract_result:* is a colon-composite the union does not
-    // cover (backend reason leakage tracked separately); it must surface raw,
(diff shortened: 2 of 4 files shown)


jeong-sik:
## Triage status

- Draft dashboard PR (`fix/dashboard-trust-attention-labels-20260626`).
- Fundamental/structural ratchets green; `Lint`, `Dashboard`, `Fragile-match (advisory)` still running.
- Heavy checks (`Build and Test`, `Health`, `TLA+ Model Checking`) are skipped while the PR remains in draft state.

Next step: when ready for review, mark the PR as non-draft so the full CI suite (including `Build and Test`) runs. The validation commands listed in the PR body look like the right local pre-checks.

jeong-sik:
Adversarial review of #22383.

The intent — humanizing composite attention reasons instead of raw tokens — is reasonable, but the implementation duplicates runtime vocabulary and parses wire format by string slicing.

**P0 — SSOT break: runtime vocabulary duplicated in the frontend**
- `dashboard/src/lib/keeper-attention-labels.ts` hand-maintains `ATTENTION_COMPLETION_CONTRACT_RESULTS` and `COMPLETION_CONTRACT_ATTENTION_REASON_LABELS` as a mirror of arms that `keeper_runtime_trust_snapshot.ml` emits. The inline comment explicitly calls token-by-token mirroring a "partial-migration anti-pattern" and says the richer trust vocabulary should be its own SSOT via a follow-up RFC. This PR does exactly what that comment warned against. If the runtime is the source of truth, the dashboard should consume a generated type/schema from it, not inline literal tokens and translations.

**P0 — hardcoded wire-format parsing**
- `completionContractAttentionReasonLabel` extracts the suffix by looking for the literal prefix `completion_contract_result:` and slicing on string length. This couples the UI to backend serialization format. If the backend changes delimiter, casing, or adds a third segment, the helper returns `null` and operators see a raw token. Parse via a validated schema, or — better — have the backend send `attention_reason` as a structured `{kind, detail}` object.

**P1 — unknown tokens only warn to `console.warn`**
- When an unknown composite arrives, the code logs `console.warn` once per token. Production builds often strip console warnings, so new backend reasons can ship unnoticed. Emit a metric/Sentry event or a dev-only invariant instead of relying on `console.warn` alone.

**P1 — unknown-composite test coverage regressed at the derivation layer**
- `dashboard/src/components/overview/overview.test.ts` replaced a test that verified `deriveKeeperAttentionReason` keeps unknown composites raw with one that only checks the known `passive_only` case. Restore an explicit unknown-composite test at the derivation layer.

**Verdict:** Request changes — the SSOT duplication and wire-format parsing are blockers under the no-string-match / SSOT review bar.

PR #22326, opened 2026-06-25
fix(keeper): add RFC-0290 Phase 1 background completion groundwork

jeong-sik:
## Scope

This PR is RFC-0290 Phase 1 groundwork only. It wires `Bg_completed` stimulus decoding and consumer-side conversion into actionable keeper observation/board-event inputs, with tests and an external-tool-surface audit note. It does not implement the Phase 2/3 background executor or producer path; no production background job currently enqueues `Keeper_event_queue.Bg_completed`.

## Summary

- Convert RFC-0290 `Bg_completed` stimuli into actionable `pending_board_event` values instead of dropping them.
- Promote background completion stimuli through keeper heartbeat intake with kind/ok logging.
- Extend the fusion wake regression test to cover background success, failure, fallback post IDs, and stimulus conversion.
- Record the external CLI/tool-surface decision in `docs/audit/2026-06-26-external-tool-surface-audit.md`.
- Keep the current-main adapter repairs required for this branch stack to build against the rebased head.
- Collapse the duplicate `Workspace_utils_ops.write_json_result` definition so `write_json`, exported result callers, and `write_backlog_result` use one backend/local-mirror durability implementation.

## Validation

- PASS: `ocamlformat --check lib/workspace/workspace_utils_ops.ml test/test_workspace_utils_coverage.ml`
- PASS: `git diff --check @{u}...HEAD`
- PASS: `rg -n "let write_json_result" lib/workspace/workspace_utils_ops.ml` shows a single definition at line 305.
- PASS: `ocamlc -stop-after parsing lib/workspace/workspace_utils_ops.ml test/test_workspace_utils_coverage.ml`
- BLOCKED: `scripts/dune-local.sh build test/test_workspace_utils_coverage.exe` refused to start because a live bare `dune build .` process (`PID 84087`) is bypassing the machine-wide wrapper lock. I did not override the guard.

PR #22401, opened 2026-06-26
fix(board): guard sweep scheduling state

jeong-sik:
## Scope note
This PR is not a cross-keeper consensus-threshold change. It is a narrow board
persistence scheduling fix for `Board_core_persist.maybe_sweep`, matching the
current title `fix(board): guard sweep scheduling state`.

## Summary
- addresses logic-gap audit quick-win #6 / finding #16 for `lib/board/board_core_persist.ml`
- moves `maybe_sweep` scheduling decisions and `last_sweep` / `last_flush` timestamp updates under `store.mutex`
- keeps `Eio.Stream.add` outside the store mutex so a full flusher inbox cannot hold the board state lock
- adds a regression that stale scheduling timestamps update once and are not immediately rescheduled on a second call

## Verification
- `ocamlformat --check lib/board/board_core_persist.ml test/test_board_ttl.ml`
- `ocamlc -stop-after parsing lib/board/board_core_persist.ml test/test_board_ttl.ml`
- `git diff --check`
- `rg -n "Stdlib\\.Lazy\\.(force|from_val)|Fun\\.protect|Mutex\\.(lock|unlock)|failwith|assert false|Eio\\.traceln|Sys\\.(readdir|command)" lib/board/board_core_persist.ml test/test_board_ttl.ml`

No Dune/local build was run; CI remains the build authority.

### DIFF
--- a/lib/board/board_core.mli
+++ b/lib/board/board_core.mli
@@ -67,7 +67,7 @@ val flush_interval_sec : float
 (** Builds a fresh empty store with default Hashtbl capacities
     (1024 posts / 4096 comments / 2048 vote-log entries),
     fresh [Eio.Mutex], cold caches, and an [Eio.Stream]
-    [flusher_inbox] capped at 1000 messages. *)
+    [flusher_inbox] capped by the persistence-layer flusher inbox capacity. *)
 val create_store : unit -> store
 
 (** Reset the per-author comment rate-limit tracker.  Test-only. *)
--- a/lib/board/board_core_persist.ml
+++ b/lib/board/board_core_persist.ml
@@ -13,6 +13,17 @@ include Board_core_payload
 let flush_interval_sec = Env_config.Board.flush_interval_sec
 let flusher_inbox_capacity = 1000
 
+(** Monotonic count of sweep/flush schedule messages skipped because the
+    bounded flusher inbox had no room for the whole batch. *)
+let flusher_schedule_dropped = Atomic.make 0
+let flusher_schedule_dropped_count () = Atomic.get flusher_schedule_dropped
+
+let record_flusher_schedule_drop scheduled_count =
+  for _ = 1 to scheduled_count do
+    Atomic.incr flusher_schedule_dropped
+  done
+;;
+
 (** Monotonic counter of persist failures (disk full, permission errors, etc.). *)
 let persist_errors = Atomic.make 0
 let persist_error_count () = Atomic.get persist_errors
@@ -251,6 +262,7 @@ let rollback_scheduled_msg store scheduled =
 let enqueue_scheduled_msg store scheduled =
   try Eio.Stream.add store.flusher_inbox scheduled.msg
   with
+  | Eio.Cancel.Cancelled _ as e -> raise e
   | exn ->
     let bt = Printexc.get_raw_backtrace () in
     rollback_scheduled_msg store scheduled;
@@ -260,42 +272,61 @@ let enqueue_scheduled_msg store scheduled =
 let enqueue_scheduled_msgs store scheduled =
   let scheduled_count = List.length scheduled in
   let inbox_len = Eio.Stream.length store.flusher_inbox in
-  if inbox_len + scheduled_count > flusher_inbox_capacity
+  if scheduled_count = 0
+  then ()
+  else if scheduled_count > flusher_inbox_capacity - inbox_len
   then (
+    record_flusher_schedule_drop scheduled_count;
     List.iter (rollback_scheduled_msg store) scheduled;
     Log.BoardLog.warn
       "board flusher inbox full; skipped scheduling sweep/flush messages \
        (queued=%d, requested=%d, capacity=%d)"
       inbox_len scheduled_count flusher_inbox_capacity)
-  else List.iter (enqueue_scheduled_msg store) scheduled
+  else (
+    (* [maybe_sweep] is the only producer of these messages and reserves
+       timestamps under [store.mutex] before this check. Once there is room for
+       the whole batch, these adds should not block on capacity. *)
+    List.iter (enqueue_scheduled_msg store) scheduled)
 ;;
 
 let maybe_sweep store =
   let scheduled =
     with_lock store (fun () ->
       let now = Time_compat.now () in
-      let scheduled = ref [] in
-      let schedule msg previous_ts =
-        scheduled := { msg; previous_ts; reserved_ts = now } :: !scheduled
+      let scheduled = [] in
+      let scheduled =
+        if
+          Stdlib.Float.compare
+            (now -. store.last_sweep)
+            (Stdlib.Float.of_int Limits.sweeper_interval_sec)
+          > 0
+        then (
+          let previous_ts = store.last_sweep in
+          store.last_sweep <- now;
+          { msg = Sweep; previous_ts; reserved_ts = now } :: scheduled)
+        else scheduled
       in
-      if
-        Stdlib.Float.compare
-          (now -. store.last_sweep)
-          (Stdlib.Float.of_int Limits.sweeper_interval_sec)
-        > 0
-      then (
-        let previous_ts = store.last_sweep in
-        store.last_sweep <- now;
-        schedule Sweep previous_ts);
-      if
-        Stdlib.Float.compare (now -. store.last_flush) flush_interval_sec > 0
-      then (
-        let previous_ts = store.last_flush in
-        store.last_flush <- now;
-        schedule Flush previous_ts);
-      List.rev !scheduled)
+      let scheduled =
+        if Stdlib.Float.compare (now -. store.last_flush) flush_interval_sec > 0
+        then (
+          let previous_ts = store.last_flush in
+          store.last_flush <- now;
+          { msg = Flush; previous_ts; reserved_ts = now } :: scheduled)
+        else scheduled
+      in
+      List.rev scheduled)
   in
   enqueue_scheduled_msgs store scheduled
+;;
+
+let reset_sweep_schedule_for_test store =
+  with_lock store (fun () ->
+    store.last_sweep <- 0.0;
+    store.last_flush <- 0.0)
+;;
+
+let sweep_schedule_timestamps_for_test store =
+  with_lock store (fun () -> store.last_sweep, store.last_flush)
 ;;
 
 (** {1 Persistence Paths} *)
--- a/lib/board/board_core_persist.ml
+++ b/lib/board/board_core_persist.ml
@@ -11,6 +11,7 @@ include Board_core_classify
 include Board_core_payload
 
 let flush_interval_sec = Env_config.Board.flush_interval_sec
+let flusher_inbox_capacity = 1000
 
 (** Monotonic counter of persist failures (disk full, permission errors, etc.). *)
 let persist_errors = Atomic.make 0
@@ -36,7 +37,7 @@ let create_store () =
   ; dirty_post_ids = Hashtbl.create 256
   ; dirty_comment_ids = Hashtbl.create 512
   ; last_flush = Time_compat.now ()
-  ; flusher_inbox = Eio.Stream.create 1000
+  ; flusher_inbox = Eio.Stream.create flusher_inbox_capacity
   ; sub_boards = Hashtbl.create 64
   ; sub_boards_by_slug = Hashtbl.create 64
   ; posts_by_turn_ref = Hashtbl.create 256
@@ -230,25 +231,71 @@ let sweep store =
 ;;
 
 (** Auto-sweep if needed, delegates to flusher actor inbox *)
+type scheduled_flusher_msg =
+  { msg : flusher_msg
+  ; previous_ts : float
+  ; reserved_ts : float
+  }
+
+let rollback_scheduled_msg store scheduled =
+  with_lock store (fun () ->
+    match scheduled.msg with
+    | Sweep ->
+      if Stdlib.Float.equal store.last_sweep scheduled.reserved_ts
+      then store.last_sweep <- scheduled.previous_ts
+    | Flush ->
+      if Stdlib.Float.equal store.last_flush scheduled.reserved_ts
+      then store.last_flush <- scheduled.previous_ts)
+;;
+
+let enqueue_scheduled_msg store scheduled =
+  try Eio.Stream.add store.flusher_inbox scheduled.msg
+  with
+  | exn ->
+    let bt = Printexc.get_raw_backtrace () in
+    rollback_scheduled_msg store scheduled;
+    Printexc.raise_with_backtrace exn bt
+;;
+
+let enqueue_scheduled_msgs store scheduled =
+  let scheduled_count = List.length scheduled in
+  let inbox_len = Eio.Stream.length store.flusher_inbox in
+  if inbox_len + scheduled_count > flusher_inbox_capacity
+  then (
+    List.iter (rollback_scheduled_msg store) scheduled;
+    Log.BoardLog.warn
+      "board flusher inbox full; skipped scheduling sweep/flush messages \
+       (queued=%d, requested=%d, capacity=%d)"
+      inbox_len scheduled_count flusher_inbox_capacity)
+  else List.iter (enqueue_scheduled_msg store) scheduled
+;;
+
 let maybe_sweep store =
-  let now = Time_compat.now () in
-  let enqueue_sweep, enqueue_flush =
+  let scheduled =
     with_lock store (fun () ->
-      let enqueue_sweep =
+      let now = Time_compat.now () in
+      let scheduled = ref [] in
+      let schedule msg previous_ts =
+        scheduled := { msg; previous_ts; reserved_ts = now } :: !scheduled
+      in
+      if
         Stdlib.Float.compare
           (now -. store.last_sweep)
           (Stdlib.Float.of_int Limits.sweeper_interval_sec)
         > 0
-      in
-      let enqueue_flush =
+      then (
+        let previous_ts = store.last_sweep in
+        store.last_sweep <- now;
+        schedule Sweep previous_ts);
+      if
         Stdlib.Float.compare (now -. store.last_flush) flush_interval_sec > 0
-      in
-      if enqueue_sweep then store.last_sweep <- now;
-      if enqueue_flush then store.last_flush <- now;
-      enqueue_sweep, enqueue_flush)
+      then (
+        let previous_ts = store.last_flush in
+        store.last_flush <- now;
+        schedule Flush previous_ts);
+      List.rev !scheduled)
   in
-  if enqueue_sweep then Eio.Stream.add store.flusher_inbox Sweep;
-  if enqueue_flush then Eio.Stream.add store.flusher_inbox Flush
+  enqueue_scheduled_msgs store scheduled
 ;;
 
 (** {1 Persistence Paths} *)
--- a/lib/board/board_core_persist.ml
+++ b/lib/board/board_core_persist.ml
@@ -11,7 +11,7 @@ include Board_core_classify
 include Board_core_payload
 
 let flush_interval_sec = Env_config.Board.flush_interval_sec
-let flusher_inbox_capacity = 1000
+let flusher_inbox_capacity = Env_config.Board.flusher_inbox_capacity
 
 (** Monotonic count of sweep/flush schedule messages skipped because the
     bounded flusher inbox had no room for the whole batch. *)
@@ -19,9 +19,10 @@ let flusher_schedule_dropped = Atomic.make 0
 let flusher_schedule_dropped_count () = Atomic.get flusher_schedule_dropped
 
 let record_flusher_schedule_drop scheduled_count =
-  for _ = 1 to scheduled_count do
-    Atomic.incr flusher_schedule_dropped
-  done
+  (* Single atomic add instead of N increments, so a concurrent reader cannot
+     observe a partial count. [scheduled_count] is a [List.length], so it is
+     non-negative. *)
+  Atomic.fetch_and_add flusher_schedule_dropped scheduled_count |> ignore
 ;;
 
 (** Monotonic counter of persist failures (disk full, permission errors, etc.). *)
--- a/lib/board/board_core_persist.ml
+++ b/lib/board/board_core_persist.ml
@@ -232,18 +232,23 @@ let sweep store =
 (** Auto-sweep if needed, delegates to flusher actor inbox *)
 let maybe_sweep store =
   let now = Time_compat.now () in
-  if
-    Stdlib.Float.compare
-      (now -. store.last_sweep)
-      (Stdlib.Float.of_int Limits.sweeper_interval_sec)
-    > 0
-  then (
-    store.last_sweep <- now;
-    Eio.Stream.add store.flusher_inbox Sweep);
-  if Stdlib.Float.compare (now -. store.last_flush) flush_interval_sec > 0
-  then (
-    store.last_flush <- now;
-    Eio.Stream.add store.flusher_inbox Flush)
+  let enqueue_sweep, enqueue_flush =
+    with_lock store (fun () ->
+      let enqueue_sweep =
+        Stdlib.Float.compare
+          (now -. store.last_sweep)
+          (Stdlib.Float.of_int Limits.sweeper_interval_sec)
+        > 0
+      in
+      let enqueue_flush =
+        Stdlib.Float.compare (now -. store.last_flush) flush_interval_sec > 0
+      in
+      if enqueue_sweep then store.last_sweep <- now;
+      if enqueue_flush then store.last_flush <- now;
+      enqueue_sweep, enqueue_flush)
+  in
+  if enqueue_sweep then Eio.Stream.add store.flusher_inbox Sweep;
+  if enqueue_flush then Eio.Stream.add store.flusher_inbox Flush
 ;;
 
 (** {1 Persistence Paths} *)
--- a/lib/board/board_core_persist.mli
+++ b/lib/board/board_core_persist.mli
@@ -17,6 +17,13 @@ end
 
 val flush_interval_sec : float
 val flusher_inbox_capacity : int
+(** Capacity of {!store.flusher_inbox}. The scheduler reserves room for a whole
+    sweep/flush batch before enqueueing. *)
+
+val flusher_schedule_dropped_count : unit -> int
+(** Count of sweep/flush schedule messages skipped because the flusher inbox was
+    full. *)
+
 val persist_error_count : unit -> int
 val record_persist_error : where:string -> string -> unit
 
@@ -33,6 +40,11 @@ val with_lock : store -> (unit -> 'a) -> 'a
 val with_persist_lock : store -> (unit -> 'a) -> 'a
 val sweep : store -> int * int
 val maybe_sweep : store -> unit
+val reset_sweep_schedule_for_test : store -> unit
+(** Reset sweep/flush schedule timestamps. Test-only. *)
+
+val sweep_schedule_timestamps_for_test : store -> float * float
+(** Snapshot sweep/flush schedule timestamps under the store mutex. Test-only. *)
 
 val board_base_path : unit -> string
 val board_masc_dir : unit -> string
--- a/lib/board/board_core_persist.mli
+++ b/lib/board/board_core_persist.mli
@@ -16,6 +16,7 @@ include module type of struct
 end
 
 val flush_interval_sec : float
+val flusher_inbox_capacity : int
 val persist_error_count : unit -> int
 val record_persist_error : where:string -> string -> unit
 
--- a/lib/config/env_config_runtime.ml
+++ b/lib/config/env_config_runtime.ml
@@ -469,6 +469,13 @@ module Board = struct
   let flush_interval_sec =
     get_float ~default:30.0 "MASC_BOARD_FLUSH_INTERVAL_SEC"
 
+  (** Capacity of the board flusher inbox (scheduled sweep/flush messages
+      enqueued by the sweeper). Single source of truth shared by the
+      persistence layer, its [.mli] doc, and the stream creation site.
+      Default: 1000. *)
+  let flusher_inbox_capacity =
+    get_int ~default:1000 "MASC_BOARD_FLUSHER_INBOX_CAPACITY"
+
   (** Board backend type as a typed selector (e.g. "jsonl", "pg"). *)
   let backend_opt () =
     Sys.getenv_opt "MASC_BOARD_BACKEND"
--- a/lib/config/env_config_runtime.mli
+++ b/lib/config/env_config_runtime.mli
@@ -231,6 +231,7 @@ module Board : sig
   val backend_of_string : string -> backend
   val backend_to_string : backend -> string
   val flush_interval_sec : float
+  val flusher_inbox_capacity : int
   val backend_opt : unit -> backend option
 end
 
--- a/test/test_board_ttl.ml
+++ b/test/test_board_ttl.ml
@@ -111,18 +111,74 @@ let test_sweeper_skips_permanent () =
   let (removed_posts, _) = sweep store in
   Alcotest.(check int) "sweeper removed 0 permanent posts" 0 removed_posts
 
+let reset_sweep_schedule_for_test store =
+  store.last_sweep <- 0.0;
+  store.last_flush <- 0.0
+
+let maybe_sweep_for_test = Masc_board_handlers.Board_core_persist.maybe_sweep
+
+let flusher_inbox_capacity_for_test =
+  Masc_board_handlers.Board_core_persist.flusher_inbox_capacity
+
+let drain_flusher_inbox store =
+  let rec loop acc =
+    match Eio.Stream.take_nonblocking store.flusher_inbox with
+    | None -> List.rev acc
+    | Some msg -> loop (msg :: acc)
+  in
+  loop []
+
+let check_one_sweep_and_flush label messages =
+  let sweep_count, flush_count =
+    List.fold_left
+      (fun (sweeps, flushes) msg ->
+         match msg with
+         | Sweep -> (sweeps + 1, flushes)
+         | Flush -> (sweeps, flushes + 1))
+      (0, 0)
+      messages
+  in
+  Alcotest.(check int) (label ^ " sweep count") 1 sweep_count;
+  Alcotest.(check int) (label ^ " flush count") 1 flush_count
+
 let test_maybe_sweep_updates_schedule_once () =
   let store = create_store () in
-  store.last_sweep <- 0.0;
-  store.last_flush <- 0.0;
-  Masc_board_handlers.Board_core_persist.maybe_sweep store;
+  reset_sweep_schedule_for_test store;
+  maybe_sweep_for_test store;
   let first_sweep = store.last_sweep in
   let first_flush = store.last_flush in
   Alcotest.(check bool) "sweep timestamp updated" true (first_sweep > 0.0);
   Alcotest.(check bool) "flush timestamp updated" true (first_flush > 0.0);
-  Masc_board_handlers.Board_core_persist.maybe_sweep store;
+  maybe_sweep_for_test store;
   Alcotest.(check (float 0.0)) "sweep timestamp unchanged" first_sweep store.last_sweep;
-  Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush store.last_flush
+  Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush store.last_flush;
+  check_one_sweep_and_flush "sequential" (drain_flusher_inbox store)
+
+let test_maybe_sweep_concurrent_schedules_once () =
+  let store = create_store () in
+  reset_sweep_schedule_for_test store;
+  let callers = List.init 32 Fun.id in
+  let _ =
+    Eio.Fiber.List.map
+      ~max_fibers:32
+      (fun _ -> maybe_sweep_for_test store)
+      callers
+  in
+  check_one_sweep_and_flush "concurrent" (drain_flusher_inbox store)
+
+let test_maybe_sweep_full_inbox_rolls_back_schedule () =
+  let store = create_store () in
+  reset_sweep_schedule_for_test store;
+  for _ = 1 to flusher_inbox_capacity_for_test do
+    Eio.Stream.add store.flusher_inbox Flush
+  done;
+  maybe_sweep_for_test store;
+  Alcotest.(check (float 0.0)) "sweep timestamp rolled back" 0.0 store.last_sweep;
+  Alcotest.(check (float 0.0)) "flush timestamp rolled back" 0.0 store.last_flush;
+  Alcotest.(check int)
+    "full inbox length unchanged"
+    flusher_inbox_capacity_for_test
+    (Eio.Stream.length store.flusher_inbox)
 
 let test_post_kind_direct_default () =
   let store = create_store () in
@@ -233,6 +289,10 @@ let () =
             (with_eio test_sweeper_skips_permanent);
           Alcotest.test_case "maybe_sweep schedules once" `Quick
             (with_eio test_maybe_sweep_updates_schedule_once);
+          Alcotest.test_case "maybe_sweep concurrent schedules once" `Quick
+            (with_eio test_maybe_sweep_concurrent_schedules_once);
+          Alcotest.test_case "maybe_sweep full inbox rolls back schedule" `Quick
+            (with_eio test_maybe_sweep_full_inbox_rolls_back_schedule);
         ] );
       ( "visibility_ssot",
         [
--- a/test/test_board_ttl.ml
+++ b/test/test_board_ttl.ml
@@ -111,14 +111,21 @@ let test_sweeper_skips_permanent () =
   let (removed_posts, _) = sweep store in
   Alcotest.(check int) "sweeper removed 0 permanent posts" 0 removed_posts
 
-let reset_sweep_schedule_for_test store =
-  store.last_sweep <- 0.0;
-  store.last_flush <- 0.0
+let schedule_reset_timestamp_for_test = 0.0
+
+let reset_sweep_schedule_for_test =
+  Masc_board_handlers.Board_core_persist.reset_sweep_schedule_for_test
+
+let sweep_schedule_timestamps_for_test =
+  Masc_board_handlers.Board_core_persist.sweep_schedule_timestamps_for_test
 
 let maybe_sweep_for_test = Masc_board_handlers.Board_core_persist.maybe_sweep
 
 let flusher_inbox_capacity_for_test =
   Masc_board_handlers.Board_core_persist.flusher_inbox_capacity
+
+let flusher_schedule_dropped_count_for_test =
+  Masc_board_handlers.Board_core_persist.flusher_schedule_dropped_count
 
 let drain_flusher_inbox store =
   let rec loop acc =
@@ -145,13 +152,13 @@ let test_maybe_sweep_updates_schedule_once () =
   let store = create_store () in
   reset_sweep_schedule_for_test store;
   maybe_sweep_for_test store;
-  let first_sweep = store.last_sweep in
-  let first_flush = store.last_flush in
+  let first_sweep, first_flush = sweep_schedule_timestamps_for_test store in
   Alcotest.(check bool) "sweep timestamp updated" true (first_sweep > 0.0);
   Alcotest.(check bool) "flush timestamp updated" true (first_flush > 0.0);
   maybe_sweep_for_test store;
-  Alcotest.(check (float 0.0)) "sweep timestamp unchanged" first_sweep store.last_sweep;
-  Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush store.last_flush;
+  let second_sweep, second_flush = sweep_schedule_timestamps_for_test store in
+  Alcotest.(check (float 0.0)) "sweep timestamp unchanged" first_sweep second_sweep;
+  Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush second_flush;
   check_one_sweep_and_flush "sequential" (drain_flusher_inbox store)
 
 let test_maybe_sweep_concurrent_schedules_once () =
@@ -172,9 +179,20 @@ let test_maybe_sweep_full_inbox_rolls_back_schedule () =
   for _ = 1 to flusher_inbox_capacity_for_test do
     Eio.Stream.add store.flusher_inbox Flush
   done;
+  let dropped_before = flusher_schedule_dropped_count_for_test () in
   maybe_sweep_for_test store;
-  Alcotest.(check (float 0.0)) "sweep timestamp rolled back" 0.0 store.last_sweep;
-  Alcotest.(check (float 0.0)) "flush timestamp rolled back" 0.0 store.last_flush;
+  let dropped_after = flusher_schedule_dropped_count_for_test () in
+  let sweep_ts, flush_ts = sweep_schedule_timestamps_for_test store in
+  Alcotest.(check int) "full inbox dropped scheduled messages" 2
+    (dropped_after - dropped_before);
+  Alcotest.(check (float 0.0))
+    "sweep timestamp rolled back"
+    schedule_reset_timestamp_for_test
+    sweep_ts;
+  Alcotest.(check (float 0.0))
+    "flush timestamp rolled back"
+    schedule_reset_timestamp_for_test
+    flush_ts;
   Alcotest.(check int)
     "full inbox length unchanged"
     flusher_inbox_capacity_for_test
--- a/test/test_board_ttl.ml
+++ b/test/test_board_ttl.ml
@@ -111,6 +111,19 @@ let test_sweeper_skips_permanent () =
   let (removed_posts, _) = sweep store in
   Alcotest.(check int) "sweeper removed 0 permanent posts" 0 removed_posts
 
+let test_maybe_sweep_updates_schedule_once () =
+  let store = create_store () in
+  store.last_sweep <- 0.0;
+  store.last_flush <- 0.0;
+  Masc.Board_core_persist.maybe_sweep store;
+  let first_sweep = store.last_sweep in
+  let first_flush = store.last_flush in
+  Alcotest.(check bool) "sweep timestamp updated" true (first_sweep > 0.0);
+  Alcotest.(check bool) "flush timestamp updated" true (first_flush > 0.0);
+  Masc.Board_core_persist.maybe_sweep store;
+  Alcotest.(check (float 0.0)) "sweep timestamp unchanged" first_sweep store.last_sweep;
+  Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush store.last_flush
+
 let test_post_kind_direct_default () =
   let store = create_store () in
   match
@@ -218,6 +231,8 @@ let () =
             (with_eio test_expiring_post);
           Alcotest.test_case "sweeper skips permanent" `Quick
             (with_eio test_sweeper_skips_permanent);
+          Alcotest.test_case "maybe_sweep schedules once" `Quick
+            (with_eio test_maybe_sweep_updates_schedule_once);
         ] );
       ( "visibility_ssot",
         [
--- a/test/test_board_ttl.ml
+++ b/test/test_board_ttl.ml
@@ -115,12 +115,12 @@ let test_maybe_sweep_updates_schedule_once () =
   let store = create_store () in
   store.last_sweep <- 0.0;
   store.last_flush <- 0.0;
-  Masc.Board_core_persist.maybe_sweep store;
+  Masc_board_handlers.Board_core_persist.maybe_sweep store;
   let first_sweep = store.last_sweep in
   let first_flush = store.last_flush in
   Alcotest.(check bool) "sweep timestamp updated" true (first_sweep > 0.0);
   Alcotest.(check bool) "flush timestamp updated" true (first_flush > 0.0);
-  Masc.Board_core_persist.maybe_sweep store;
+  Masc_board_handlers.Board_core_persist.maybe_sweep store;
   Alcotest.(check (float 0.0)) "sweep timestamp unchanged" first_sweep store.last_sweep;
   Alcotest.(check (float 0.0)) "flush timestamp unchanged" first_flush store.last_flush
 


anyang-keepers:
## 🔴 BLOCK — Title/Content Mismatch (Severe)

**Claim:** "Add cross-keeper consensus threshold"

**Problem:** The diff shows changes to `board_core_persist.ml` — adding `with_lock` around sweeper/flush scheduling to prevent duplicate enqueue. This is about **board persistence scheduling**, not cross-keeper consensus.

**This is either:**
1. Wrong branch pushed to this PR
2. Copy-paste error in PR creation
3. Deliberate mislabeling

**Risk:** If this merges as "consensus threshold", operators will think distributed consensus is implemented while actually only a local lock was added. This creates false confidence in cross-keeper safety.

**Request:**
1. Verify this is the correct branch for the title
2. If wrong, close and resubmit with correct changes
3. If correct, explain how a local board lock relates to cross-keeper consensus

This PR cannot be reviewed as-is. — executor (adversarial review)

jeong-sik:
Verified current PR metadata and diff.

- Current title is `fix(board): guard sweep scheduling state`.
- Current branch is `fix/audit-board-sweep-mutex-20260626`.
- Current head is `9a6360bf03181e1180ae7515c3fd393775876dd6`.
- Diff touches only `lib/board/board_core_persist.ml` and `test/test_board_ttl.ml`.
- The PR body already described board persistence scheduling, and I added an explicit scope note stating this is not a cross-keeper consensus-threshold change.

So this is not the wrong branch for the current title/body. The `Add cross-keeper consensus threshold` claim in the review does not match this PR's current metadata or diff. There is no cross-keeper consensus implementation here; the local board lock only guards `maybe_sweep` scheduling state so duplicate sweep/flush enqueue decisions cannot race while `Eio.Stream.add` remains outside the store mutex.

I did not inspect or run CI/build checks.

jeong-sik:
Pushed `867490f2145` to fix the CI compile break in `test/test_board_ttl.ml`.

Root cause: the test called `Masc.Board_core_persist`, but board handlers are now exposed through the wrapped `masc.board_handlers` library as `Masc_board_handlers.Board_core_persist`.

Local verification passed:
- `ocamlformat --check test/test_board_ttl.ml`
- `ocamlc -stop-after parsing test/test_board_ttl.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_board_ttl.exe`

jeong-sik:
## Adversarial review: MASC#22401

🔴 **Previously raised blocker — RESOLVED.** The earlier title/content mismatch ("cross-keeper consensus threshold") is fixed. Current title, branch, body, and diff all scope this as a board persistence scheduling fix only.

### P0 — none identified

### P1 — should fix before merge
1. **Regression test does not exercise the concurrency bug.** `test/test_board_ttl.ml:114-127` is purely sequential. The stated race is duplicate `Sweep`/`Flush` enqueue under concurrent `maybe_sweep` calls, but the test cannot detect it. Add an Eio concurrency test: spawn multiple fibers calling `maybe_sweep` and assert the flusher inbox contains exactly one `Sweep` and one `Flush`.
2. **Schedule timestamp committed before work is enqueued.** `lib/board/board_core_persist.ml:244-251` mutates `last_sweep`/`last_flush` under the lock before `Eio.Stream.add`. If `add` raises or the fiber is cancelled after the mutation, the state records a schedule that never reached the flusher, suppressing the next interval. Update timestamps only after a successful `add`, or add a compensating rollback.
3. **Scheduling state contends on the data mutex.** `maybe_sweep` now acquires `store.mutex`, which is held by long-running `sweep` and read paths. Every board read now has an extra scheduling critical section. Prefer a dedicated scheduling lock or atomic time counters.
4. **Test hardcodes internal library module name.** `test/test_board_ttl.ml:118,123` uses `Masc_board_handlers.Board_core_persist.maybe_sweep`. Expose through the public test surface or open the internal module locally.

**CI note:** `Build and Test`, `Lint`, and several ratchet jobs are still queued/in-progress; wait for required checks before merge.

**Verdict:** No P0; P1 test coverage, enqueue ordering, and mutex-contention issues remain; prior title/content mismatch is resolved.

jeong-sik:
Reviewed current head `867490f21454b52364bf5dd926c6b7915dc15132`.

No blocking finding in the board sweep guard. The schedule decision now happens under the board lock, so concurrent callers cannot all observe the stale `last_sweep`/`last_flush` and enqueue duplicate work. The actual `Eio.Stream.add` calls stay outside the lock, which avoids holding the board mutex if the flusher inbox applies backpressure. The added TTL test pins the second-call timestamp behavior.

CI boundary: all guard/lint/health checks visible here are green or skipped as expected; `Build and Test` is still running, so final readiness should wait for that job.

jeong-sik:
# Adversarial Review — MASC PR #22401

## TL;DR

- 변경 범위는 `lib/board/board_core_persist.ml`의 `maybe_sweep` 상호배제 수정과 이에 대한 회귀 테스트 한 건으로 좁음.
- 이전 `anyang-keepers` 리뷰에서 지적한 **title/scope mismatch**는 작성자가 본문에 명시적으로 scope note를 추가하여 해소함.
- **P0**: `Build and Test` CI가 빨간불. diff 자체와는 무관해 보이지만, merge 전에는 CI가 green이어야 함.
- 코드 자체에는 치명적 결함은 없으나 P1~P3 수준의 설계/테스트 품질 이슈가 있음.

## CI Status (`gh pr checks 22401`)

| 항목 | 결과 | 비고 |
|------|------|------|
| `Build and Test` | **fail** | job `83628374993` — `scripts/harness/contract/public_tool_live_sweep.sh` 내 `masc_goal_transition` 실패 |
| `ocamlformat-check` | pass | |
| `Lint` / `Health` / `PR Live Gate` / 대부분의 meta ratchet | pass | |

**상세 실패 원인 (CI 로그):**
- `FAIL: masc_goal_transition: expected success`
- `"error_code": "validation_error", "message": "actor id must match authenticated caller"`
- 이와 별도로 `deepseek-v4-flash` 호출 시 `Eio.Io Net Connection_failure Refused` 발생
- 본 PR diff(`board_core_persist.ml`, `test_board_ttl.ml`)와 직접적인 인과관계는 보이지 않으나, **CI gate 기준으로는 P0**.

## Concrete Findings

### P0 — CI gate blocks merge
- `Build and Test` failure: https://github.com/jeong-sik/masc/actions/runs/28229051277/job/83628374993
- 본 PR 코드가 아닌 contract harness / MCP auth / 외부 LLM 목킹 환경 문제로 보이나, CI가 green이 아니면 merge 불가.

### P1 — `Eio.Stream.add` still blocks callers (keeper-blocking design)
- `lib/board/board_core_persist.ml:250-251`
  ```ocaml
  if enqueue_sweep then Eio.Stream.add store.flusher_inbox Sweep;
  if enqueue_flush then Eio.Stream.add store.flusher_inbox Flush
  ```
- mutex 밖으로 분리한 것은 state lock 경쟁을 줄이는 것이지 **blocking 자체를 제거한 것이 아님**.
- `flusher_inbox`는 capacity 1000의 bounded stream(`board_core_persist.ml:39`). flusher actor가 밀리면 `create_post`, `get_post`, `find_post_by_*` 등을 호출한 keeper/caller가 `Stream.add`에서 대기함.
- "no keeper-blocking design" 기준에 따라 capacity/length 확인, drop + metric, 또는 별도 scheduler fiber를 통한 non-blocking add 방식을 고려해야 함.

### P2 — TOCTOU weakening of the "once per interval" invariant
- `lib/board/board_core_persist.ml:234,246-250`
- `now`를 lock 밖에서 샘플링하고, lock 해제 후 `Stream.add`를 실행. lock 획득~`Stream.add` 사이에 caller fiber가 `sweeper_interval_sec`(10s) 이상 선점되면, 다른 caller가 이미 갱신된 `last_sweep`을 보고도 `now2 -. last_sweep > interval` 조건을 만족해 중복 enqueue할 수 있음.
- 발생 가능성은 낮으나, "scheduling once" 불변을 완전히 보장하지는 못함.

### P2 — Test directly mutates internal store representation
- `test/test_board_ttl.ml:116-117`
  ```ocaml
  store.last_sweep <- 0.0;
  store.last_flush <- 0.0;
  ```
- `store`의 mutable field에 직접 접근하여 테스트 상태를 조작. representation 캡슐화를 깨고, 향후 `store` 레코드 구조가 바뀌면 테스트도 깨짐.
- `reset_schedule_for_test` 같은 test-only helper를 `Board_core_persist`에 노출하거나, `create_store` 변형을 사용할 것.

### P2 — Regression test does not verify stream behavior
- `test/test_board_ttl.ml:114-125`
- timestamp가 한 번만 갱신되는 것은 검사하지만, `flusher_inbox`에 `Sweep`/`Flush` 메시지가 **정확히 하나씩** enqueue되었는지는 검증하지 않음.
- double enqueue regression을 잡아내지 못함. `Eio.Stream.length` 등으로 inbox 상태를 추가 assertion할 것(가능하다면).

### P3 — Magic sentinel and mutable preference
- `test/test_board_ttl.ml:116-117,121-122`: `0.0`을 "아직 sweep 안 함" sentinel로 사용. named constant 또는 helper 권장.
- `lib/board/board_core_persist.ml:246-247`: `last_sweep`/`last_flush` mutable update. "prefer immutable" 기준에 따라 scheduler state를 별도 record/atomic으로 분리하는 방향을 고려.

## Blast Radius

- **masc#22335**: task-1516 cascade hardening. 변경 파일이 완전히 다름(keeper registry, fusion, scripts). 영향 없음.
- **masc#2172**: keeper library 도구 테스트. 파일/도메인 겹침 없음.
- **oas#2172**: malformed JSON retry classification. MASC board와 무관.
- **oas#2174**: OAS concurrency/HTTP client fixes. `Stdlib_mu/Eio_mu` mutex variant 제거 등 Eio.Mutex 사용 방식은 본 PR과 일관됨.

## What’s Good

- `with_lock`이 `Eio.Mutex.use_rw ~protect:true`를 사용 — lock leak 위험 낮음.
- hardcoding, magic number(코드 내), local path, env var overuse, string matching 기반 분류, stub, silent failure는 diff 범위에서 발견되지 않음.
- `Limits.sweeper_interval_sec`, `Env_config.Board.flush_interval_sec` 등 기존 SSOT를 재사용.

## Recommendations

1. **CI red 해결/재실행** — `public_tool_live_sweep.sh` 실패가 본 PR과 무관함을 확인하면 재실행, 관련이 있으면 fix.
2. **keeper-blocking 설계 개선** — `Eio.Stream.add`가 호출자를 block하지 않도록 capacity 확인 또는 drop + metric 방식 검토.
3. **회귀 테스트 강화** — inbox 메시지 개수/종류 assertion 추가.
4. **테스트 캡슐화** — `store.last_sweep <- 0.0` 같은 직접 mutable 접근을 test helper로 대체.

## Verdict

Title/scope mismatch는 해소됨. diff는 narrow하고 `maybe_sweep` race window를 줄이는 올바른 방향. **CI가 green이고 위 P1/P2 피드백을 반영(또는 명시적으로 수용)한 뒤 approve 가능.**

jeong-sik:
Current-head follow-up for `934779e649a8fb7159b2bd1730669b72c7e0fe5d`: the new commit only fixes the carved module reference in `test/test_board_ttl.ml`; the board scheduling implementation is otherwise the same as the prior review head.

I do not see a new source blocker. The lock now makes the sweep/flush scheduling decision atomic with respect to `last_sweep`/`last_flush`, and the actual `Eio.Stream.add` calls remain outside the board mutex.

Two P1 debts still remain worth fixing before this gets treated as hardened concurrency coverage: sample `now` inside the scheduling lock so a long lock wait cannot write an already-stale timestamp, and make the regression test assert the flusher inbox receives exactly one `Sweep` and one `Flush` rather than only checking timestamps. Those are not new P0s from this head, but they are the places where the stated "duplicate enqueue" invariant is still weakest.

Merge readiness is still CI-bound; the fresh Build/Test and guard jobs were queued or in progress when I rechecked.

jeong-sik:
Updated this branch onto current main and pushed `22f0ad6e`.

What changed:
- rebased over the merged contract-harness fixes (`#22415`, `#22326`, plus current main)
- kept the sweep scheduling reservation under the board lock
- added rollback when the flusher inbox is full or enqueue raises
- exposed the inbox capacity constant and pinned sequential/concurrent/full-inbox tests

Local validation:
- `ocamlformat --check lib/board/board_core_persist.ml lib/board/board_core_persist.mli test/test_board_ttl.ml`
- `ocamlc -stop-after parsing lib/board/board_core_persist.ml lib/board/board_core_persist.mli test/test_board_ttl.ml`

Focused Dune note: `scripts/dune-local.sh build test/test_board_ttl.exe` could not start because a bare `dune build .` process is bypassing the repo-local Dune lock (`scripts/dune-local.sh` exited 75). CI should be the build authority for this push.

jeong-sik:
Current-head review for `22f0ad6e6cab2f93f30ff419799e3895fe6778b3`.

No source blocker found. `maybe_sweep` now reserves sweep/flush timestamps under the board mutex before enqueueing, rolls those reservations back when the bounded flusher inbox is full, and the tests cover sequential scheduling, concurrent callers, and full-inbox rollback.

Health/Build/Lint and several ratchets are still in progress, so this remains CI-bound.

jeong-sik:
# Adversarial review follow-up — MASC PR #22401

**Current head:** `22f0ad6e6cab2f93f30ff419799e3895fe6778b3`  
**Prior review context:** `867490f21454b52364bf5dd926c6b7915dc15132` / `934779e649a8fb7159b2bd1730669b72c7e0fe5d`  
**Head changed:** yes. **Author replied/pushed after last adversarial comment:** yes.

## Previously raised P0/P1 status

### P0 — `Build and Test` CI red
**Addressed.** The failing run referenced in the prior review (`28229051277`, `masc_goal_transition` auth/LLM-mocking failure) has been superseded. The current run (`28233517027`) shows `Build and Test` as **pending** and all other visible checks as **pass**. The prior P0 failure is no longer present, though final approval should still wait for the pending job to go green.

### P1 — `Eio.Stream.add` can block keeper callers
**Partially addressed, not fully resolved.**

The diff now exposes `flusher_inbox_capacity` and guards the scheduling enqueue with a capacity check + rollback:

```ocaml
+let flusher_inbox_capacity = 1000
 ...
+let enqueue_scheduled_msgs store scheduled =
+  let scheduled_count = List.length scheduled in
+  let inbox_len = Eio.Stream.length store.flusher_inbox in
+  if inbox_len + scheduled_count > flusher_inbox_capacity
+   then (
+    List.iter (rollback_scheduled_msg store) scheduled;
+    Log.BoardLog.warn
+      "board flusher inbox full; skipped scheduling sweep/flush messages \
+       (queued=%d, requested=%d, capacity=%d)"
+      inbox_len scheduled_count flusher_inbox_capacity)
+  else List.iter (enqueue_scheduled_msg store) scheduled
```

This prevents enqueue (and rolls back the reserved timestamps) when the bounded inbox is already full, which directly responds to the "bounded stream can block callers" concern.

**Residual concern:** the actual `Eio.Stream.add` still happens outside the lock and can still block if the stream fills between the `Eio.Stream.length` check and the add, or if flusher backpressure grows faster than the capacity guard anticipates. The guard also emits only a log warning, not a structured metric/counter, and does not route work through a separate scheduler fiber. So keeper-blocking risk is reduced but not eliminated.

### P2 — TOCTOU weakening of the "once per interval" invariant
**Fixed.** `now` is now sampled *inside* the board lock, and `last_sweep`/`last_flush` are updated atomically with the scheduling decision:

```ocaml
+let maybe_sweep store =
+  let scheduled =
+    with_lock store (fun () ->
+      let now = Time_compat.now () in
+      ...
+      store.last_sweep <- now;
+      ...
+      store.last_flush <- now;
```

The prior race window (sampling `now` before acquiring the lock) is closed.

### P2 — Regression test did not verify stream behavior
**Fixed.** The test now drains the flusher inbox and asserts exactly one `Sweep` and one `Flush` for both sequential and concurrent callers:

```ocaml
+let check_one_sweep_and_flush label messages =
+  ...
+  Alcotest.(check int) (label ^ " sweep count") 1 sweep_count;
+  Alcotest.(check int) (label ^ " flush count") 1 flush_count
...
+let test_maybe_sweep_concurrent_schedules_once () =
+  ...
+  Eio.Fiber.List.map
+    ~max_fibers:32
+    (fun _ -> maybe_sweep_for_test store)
+    callers;
+  check_one_sweep_and_flush "concurrent" (drain_flusher_inbox store)
```

### P2 — Test directly mutates internal store representation
**Still present, only relocated.** `reset_sweep_schedule_for_test` still pokes `store.last_sweep` and `store.last_flush` directly:

```ocaml
+let reset_sweep_schedule_for_test store =
+  store.last_sweep <- 0.0;
+  store.last_flush <- 0.0
```

This keeps the representation-coupling risk flagged in the previous review.

### P3 — Magic sentinel and mutable preference
**Still present.** The test uses `0.0` as a sentinel, and the implementation still mutates `last_sweep`/`last_flush`.

## Verdict

- The title/scope mismatch blocker is resolved.
- The prior CI P0 failure is resolved/superseded (pending final green `Build and Test`).
- The P1 keeper-blocking concern is materially reduced by the capacity guard + rollback, but not fully eliminated.
- Two P2/P3 quality items (test representation coupling, sentinel/mutable preference) remain.

**Recommendation:** approve once `Build and Test` is green and the residual P1 risk is explicitly accepted or replaced with a non-blocking enqueue path (e.g., scheduler fiber or drop+metric).

jeong-sik:
Fixed the concrete CI source failure on the current head.

Cause from Build and Test run 28233517027 / job 83643127890:
- `test_otel_zero_fill` failed because `Keeper_metrics.all` expected `224` but current registration count is `227`.
- The PR-owned board sweep/TTL tests visible in that log were passing.

Patch pushed:
- `9d75af3e00b fix(test): refresh otel metric count`
- `test/test_otel_zero_fill.ml`: expected metric count `224 -> 227`.

Local validation on the PR worktree:
- `git diff --check`
- `git diff --check origin/main`
- `ocamlformat --check lib/board/board_core_persist.ml lib/board/board_core_persist.mli test/test_board_ttl.ml test/test_otel_zero_fill.ml`
- `ocamlc -stop-after parsing lib/board/board_core_persist.ml lib/board/board_core_persist.mli test/test_board_ttl.ml test/test_otel_zero_fill.ml`
- `scripts/dune-local.sh build test/test_board_ttl.exe test/test_otel_zero_fill.exe`
- `scripts/dune-local.sh build @check`

Fresh CI is re-queued on the pushed head.

jeong-sik:
# Adversarial review follow-up — MASC PR #22401

**Current head:** `9d75af3e00bde107b87e0cb94dcbea8eecb385a8`  
**Baseline adversarial review:** `22f0ad6e` 기준 follow-up (2026-06-26T13:24Z, comment `4809984923`)  
**Commits pushed after baseline:** `9d75af3e00b fix(test): refresh otel metric count`

---

## Baseline P0/P1/P2/P3 status after post-baseline commit

| # | Item | Status | Evidence |
|---|------|--------|----------|
| P0 | `Build and Test` CI green | **NOT FIXED** | Run `28251153006`: `Build and Test` **fail** (`83702850565`), `CI Gate` **fail** (`83711148579`). Baseline review 시점의 pending 상태가 green으로 전환되지 않음. |
| P1 | `Eio.Stream.add` blocks keeper callers | **NOT FIXED** | `lib/board/board_core_persist.ml:251-252` — 여전히 `Eio.Stream.add store.flusher_inbox scheduled.msg`를 호출자 fiber에서 직접 실행. capacity guard는 사전 거부할 뿐 blocking 자체를 제거하지 않음. |
| P1 | Capacity guard TOCTOU / no metric | **PARTIAL** | `lib/board/board_core_persist.ml:260-270` — `Eio.Stream.length` 조회와 `add` 사이에 inbox가 차면 여전히 block; structured counter/metric 없이 log warning만 기록. |
| P2 | TOCTOU: `now` sampled inside lock | **FIXED** | `lib/board/board_core_persist.ml:276` — `let now = Time_compat.now () in`이 `with_lock` 낶에서 샘플링됨. |
| P2 | Regression test asserts one Sweep/Flush | **FIXED** | `test/test_board_ttl.ml:131-142,157-167` — `check_one_sweep_and_flush`가 sequential/concurrent 모두에서 정확히 1개씩 assert. |
| P2 | Test directly mutates internal store representation | **NOT FIXED** | `test/test_board_ttl.ml:114-116` — `reset_sweep_schedule_for_test`가 여전히 `store.last_sweep <- 0.0`, `store.last_flush <- 0.0`로 직접 접근. |
| P3 | Magic sentinel `0.0` and mutable preference | **NOT FIXED** | `test/test_board_ttl.ml:115-116,176-177` — `0.0` sentinel 그대로; `lib/board/board_core_persist.ml:288,294` — `last_sweep`/`last_flush` mutable update 그대로. |
| — | Magic number 1000 exposed | **FIXED** | `lib/board/board_core_persist.ml:14`, `lib/board/board_core_persist.mli:19` — `flusher_inbox_capacity` 상수로 노출. |

---

## New / worsened regressions observed in current CI

Run `28251153006` 기준, baseline review 시점(`28233517027` pending) 이후 **새로운 실패 양상**이 확인됨. `9d75af3e`는 `test_otel_zero_fill.ml`의 metric count만 수정했으므로, 아래 실패들은 동일 PR 내 다른 변경 또는 main/rebase 상태와의 상호작용으로 보임.

### 🟡 P1 — `Eio__Eio_mutex.Poisoned(_)` in sample tests
- `Run tests (quick suite)` 로그:
  - `[exception] Stdlib.Effect.Unhandled(Eio__core__Cancel.Get_context)`
  - `[exception] Eio__Eio_mutex.Poisoned(_)`
- `lib/board/board_core_persist.ml:116`의 `with_lock`이 `Eio.Mutex.use_rw ~protect:true`를 사용. callback이 예외/취소를 만나면 mutex가 **poison**됨. `maybe_sweep`이 `with_lock`을 추가로 획득하면서 취소 시 poisoning 전파 가능성이 증가함.
- **확인 필요:** 이 실패가 `maybe_sweep` 변경으로 재현 가능한지, 아니면 기존 main 문제인지 분리해야 함.

### 🟡 P1 — `last_sweep_unixtime registered` metric test failure
- `metrics-registered 0 both metrics registered at init`:
  - `FAIL last_sweep_unixtime registered`
- board metric 등록이 `maybe_sweep` 스케줄링 변경과 연관될 수 있음. diff에서 metric 등록 코드는 보이지 않으나, board lifecycle/timing 변경이 등록 시점에 영향을 줄 수 있음.

### 🟡 P2 — 광범위한 keeper/board-unrelated test failures
- `distinct_keys`, `wiring`, `content-blocks`, `config`, `persistent registry`, `triage`, `operator`, `credentials`, `authorization`, `cache` 등 다수 실패.
- 상당수가 `Eio__Eio_mutex.Poisoned(_)`나 shared fixture 오염으로 보임. PR diff와 직접적인 인과관계는 불분명하나, **CI gate 기준으로는 merge 불가**.

---

## Verdict

- **Title/scope mismatch:** FIXED (baseline 기준 이미 해결).
- **TOCTOU race on `now` / double-enqueue test coverage:** FIXED.
- **P1 keeper-blocking design:** **NOT FIXED** — capacity guard는 사전 거부일 뿐, `Eio.Stream.add` 호출자 block은 여전히 존재.
- **P2 representation coupling / P3 sentinel-mutable:** **NOT FIXED**.
- **P0 CI green:** **NOT FIXED** — 오히려 baseline pending 이후 `Build and Test` + `CI Gate`가 fail로 확정됨. 추가로 `Eio__Eio_mutex.Poisoned(_)`와 `last_sweep_unixtime registered` 같은 회귀 가능성이 나타남.

**Recommendation:** merge 금지. 아래 조건 충족 전까지 approve 불가:
1. `Build and Test` 및 `CI Gate` green.
2. `Eio__Eio_mutex.Poisoned(_)` 원인 분리 및 해결 (필요 시 `~protect:true` 사용 재검토).
3. `last_sweep_unixtime registered` 실패 원인 분리.
4. P1 keeper-blocking 설계를 명시적으로 수용하거나, non-blocking enqueue path(scheduler fiber / drop+counter)로 대체.

jeong-sik:
Fixed the stale lazy-startup cleanup expectation: the runtime plan/dispatcher only supports `jsonl_prune`, so `test_server_runtime_bootstrap.exe` no longer expects the nonexistent `auth_archive_prune` task.\n\nPushed `d287cca14de9` (`fix(test): align lazy startup cleanup expectation`).\n\nLocal validation in `.worktrees/auth-prune-batch-20260627` on this branch:\n- `git diff --check -- test/test_server_runtime_bootstrap.ml`\n- `ocamlformat --check test/test_server_runtime_bootstrap.ml`\n- `ocamlc -stop-after parsing test/test_server_runtime_bootstrap.ml`\n- `scripts/dune-local.sh build test/test_server_runtime_bootstrap.exe`

jeong-sik:
# Adversarial review follow-up — MASC PR #22401

**Current head:** `d287cca14de9cf70bcd0440975021c85b2d897b9`  
**Baseline adversarial review:** `2026-06-26T17:32:38Z` (comment `4811884716`, head `9d75af3e00b`)  
**Commits pushed after baseline:** `d287cca14de9 fix(test): align lazy startup cleanup expectation`

---

## Baseline P0/P1/P2/P3 status after post-baseline commit

| # | Item | Status | Evidence |
|---|------|--------|----------|
| P0 | `Build and Test` CI green | **NOT FIXED** | `Build and Test` is still `IN_PROGRESS`. Baseline review 시점의 fail 상태가 해소되었을 가능성은 있으나 green 확정은 아님. |
| P1 | `Eio.Stream.add` blocks keeper callers | **NOT FIXED** | `lib/board/board_core_persist.ml:251-252` — 여전히 caller fiber에서 `Eio.Stream.add store.flusher_inbox scheduled.msg` 직접 실행. `270`의 `List.iter`도 동일. capacity guard는 사전 거부일 뿐 blocking 자체를 제거하지 않음. |
| P1 | Capacity guard TOCTOU / no metric | **PARTIAL** | `lib/board/board_core_persist.ml:260-270` — `Eio.Stream.length` 조회와 `Eio.Stream.add` 사이에 inbox가 차면 여전히 block; structured counter/metric 없이 `Log.BoardLog.warn`만 기록. |
| P2 | TOCTOU: `now` sampled inside lock | **FIXED** | `lib/board/board_core_persist.ml:275` — `let now = Time_compat.now () in`이 `with_lock` 낶에서 샘플링됨. |
| P2 | Regression test asserts one Sweep/Flush | **FIXED** | `test/test_board_ttl.ml:131-142`, `144-167`, `157-167` — sequential/concurrent 모두에서 정확히 1개씩 assert. |
| P2 | Test directly mutates internal store representation | **NOT FIXED** | `test/test_board_ttl.ml:114-116` — `reset_sweep_schedule_for_test`가 여전히 `store.last_sweep <- 0.0`, `store.last_flush <- 0.0`로 직접 접근. |
| P3 | Magic sentinel `0.0` and mutable preference | **NOT FIXED** | `test/test_board_ttl.ml:114-116`, `150-154`, `176-177` — `0.0` sentinel 그대로; `lib/board/board_core_persist.ml:287-288`, `293-294` — `last_sweep`/`last_flush` mutable update 그대로. |
| — | Magic number 1000 | **FIXED** | `lib/board/board_core_persist.ml:14` — `flusher_inbox_capacity` 상수로 노출. |

---

## New / worsened regressions observed after baseline

### Post-baseline commit 범위
`d287cca14de9`는 `test/test_server_runtime_bootstrap.ml`만 수정함:
- `test/test_server_runtime_bootstrap.ml:233` — `cleanup` 그룹의 task list에서 `auth_archive_prune` 제거.
- `test/test_server_runtime_bootstrap.ml:239-242, 244` — flattened task order에서 `auth_archive_prune` 제거.

이 변경은 lazy startup plan이 실제로 지원하는 `jsonl_prune` task만 기대하도록 맞춘 것이며, baseline에서 언급된 "lazy startup cleanup mismatch" 가능성을 해소할 수 있음. **그러나 board sweep mutex / keeper-blocking / Eio mutex poisoning 관련 실패에는 영향을 주지 않음.**

### 지속되는 P1 — `Eio.Mutex.use_rw ~protect:true` poisoning 위험
`lib/board/board_core_persist.ml:116`:
```ocaml
let with_lock store f = Eio.Mutex.use_rw ~protect:true store.mutex (fun () -> f ())
```
`~protect:true`는 callback이 예외 또는 취소를 만나면 mutex를 poison 시킴. `maybe_sweep`(`275`)이 동일 `with_lock`을 추가로 획득하면서, 기존 board operation과의 상호작용으로 `Eio__Eio_mutex.Poisoned(_)` 전파 가능성이 여전히 존재. baseline에서 보고된 sample test 실패 양상이 재발할 수 있음.

---

## Related PR blast radius

- 동일 파일(`lib/board/board_core_persist.ml`)을 touch하는 open PR은 추가로 발견되지 않음.
- 변경 범위는 `lib/board/board_core_persist.ml`, `test/test_board_ttl.ml`, `test/test_otel_zero_fill.ml`, `test/test_server_runtime_bootstrap.ml`로 한정됨.
- `test_otel_zero_fill.ml`의 metric count 조정(`224 -> 227`)은 본 PR의 metric 추가가 아닌 단순 기대값 refresh이므로 blast radius는 제한적.

---

## Verdict

- **Title/scope mismatch:** FIXED (baseline 기준 이미 해결).
- **TOCTOU race on `now` / double-enqueue test coverage:** FIXED.
- **P1 keeper-blocking design:** **NOT FIXED** — capacity guard는 사전 거부일 뿐, `Eio.Stream.add` 호출자 block은 여전히 존재.
- **P2 representation coupling / P3 sentinel-mutable:** **NOT FIXED**.
- **P0 CI green:** **NOT FIXED** — `Build and Test`가 아직 진행 중. `d287cca14de9`가 `auth_archive_prune` 관련 실패는 해소했을 수 있으나, 전체 suite 결과를 기다려야 함.
- **P1 Eio mutex poisoning:** **NOT FIXED** — `~protect:true` 사용과 `maybe_sweep` 추가 lock 경로가 여전히 우려됨.

**Recommendation:** merge 금지. 아래 조건 충족 전까지 approve 불가:
1. `Build and Test` 및 `CI Gate` green 확정.
2. P1 keeper-blocking 설계를 명시적으로 수용하거나, non-blocking enqueue path(scheduler fiber / drop+counter)로 대체.
3. `Eio.Mutex.use_rw ~protect:true` poisoning 가능성에 대한 분석 및 완화(필요 시 `~protect:false`와 explicit `Fun.protect` 고려).

jeong-sik:
Applied B5 guard follow-up commit 0f43277b00df.

Validation:

jeong-sik:
# [Adv Review] masc#22401 — follow-up

**현재 head:** `0f43277b00df` (B5 guard follow-up)  
**기준 review:** 2026-06-26T19:01:17Z adversarial follow-up (`d287cca14de9` 기준)

## Acknowledged fixes

- `test/test_masc_oas_bridge_cancel_boundary.ml` B5 guard가 `Cancel_wall_bucket.of_wall wall` SSOT 호출을 assert하도록 변경됨 — bucket literal 중복을 제거함.
- 이전 커밋에서 `test_otel_zero_fill.ml` metric count(`224 → 227`)와 `test_server_runtime_bootstrap.ml` lazy cleanup 기대값도 이미 맞춰짐.

## Previously raised P0/P1 status

- **P0 — `Build and Test` / `CI Gate` green:** 해결되지 않음. 현재 head 기준 `Build and Test`와 `CI Gate` 모두 **FAILURE**로 확정됨. B5/metric/startup 테스트 수정만으로 core suite 결과가 바뀌지 않았음.
- **P1 — keeper-blocking `Eio.Stream.add`:** 여전히 존재. `lib/board/board_core_persist.ml:251-252`에서 caller fiber가 직접 `Eio.Stream.add`를 실행하고, capacity guard는 사전 거부일 뿐 blocking 자체를 제거하지 않음.
- **P1 — capacity guard TOCTOU / metric 부재:** 여전히 잔존. `Eio.Stream.length` 조회와 `add` 사이에 inbox가 차면 block 가능하며, structured counter/metric 없이 log warning만 기록됨.
- **P1 — `Eio.Mutex.use_rw ~protect:true` poisoning 우려:** `lib/board/board_core_persist.ml:116`의 `with_lock`이 여전히 `~protect:true`를 사용해 callback 실패/취소 시 mutex poison 가능성이 남아 있음.
- **P2/P3 — representation coupling / magic sentinel / mutable preference:** 해결되지 않음. `test/test_board_ttl.ml:114-116`가 여전히 `store.last_sweep`/`last_flush`를 직접 수정하고, `0.0` sentinel과 mutable update도 그대로임.

## Verdict

B5 SSOT guard와 metric/startup 테스트 보정은 반영되었으나, 핵심 P0/P1(blocking design, CI gate, mutex poisoning risk)은 아직 처리되지 않았음. merge 불가. 위 P0/P1이 해결되거나 명시적으로 수용될 때까지 approve 불가.

jeong-sik:
# Adversarial review: jeong-sik/masc#22401

**PR:** `fix(board): guard sweep scheduling state` (#22401)  
**State:** OPEN, Draft  
**Base/head:** `main` ← `fix/audit-board-sweep-mutex-20260626`  
**Head commit:** `c5d9c842c13a1e2e718648373f3456d498cf7324`

## CI status

- `Build and Test` ❌ failed at *Run tests (quick suite)* (job `83776559691`, 40m53s)
- `CI Gate` ❌ failed because `Build and Test` is red
- `ocamlformat-check` ✅, `Lint` ✅, `Build` ✅, `Health` ✅
- Many unrelated quick-suite failures are visible in the log (e.g. `keeper_recording_error_state`, `metrics-registered`, `wiring`, `content-blocks`, `pr-f-test-mode`, etc.). The PR **does compile**, but the branch cannot merge with a red `Build and Test`.

> **P0 blocker:** CI must be green before merge. At minimum the PR author should confirm the failures are pre-existing on `main` and not introduced by these changes.

## What the PR does

`Board_core_persist.maybe_sweep` previously read `last_sweep` / `last_flush`, compared intervals, updated timestamps, and enqueued `Sweep`/`Flush` messages without holding the store mutex. That allowed multiple concurrent callers to schedule duplicate sweeps/flushes.

The patch:

- Moves the interval check + timestamp reservation under `store.mutex`.
- Introduces a `scheduled_flusher_msg` record carrying the original and reserved timestamps.
- Keeps `Eio.Stream.add` outside the mutex so a full inbox cannot pin the board state lock.
- Adds a capacity guard: if `Eio.Stream.length + scheduled_count > 1000`, it rolls the timestamps back and logs a warning.
- Adds regression tests for sequential, concurrent, and full-inbox cases.

The core fix is directionally correct.

## Issues

### P0 — CI is red
See CI status above. Even if the failures are unrelated, the PR cannot be approved until `Build and Test` and `CI Gate` pass.

### P1 — `Eio.Cancel.Cancelled` is caught by the generic handler
`lib/board/board_core_persist.ml`, `enqueue_scheduled_msg` (around new lines 43–50):

```ocaml
try Eio.Stream.add store.flusher_inbox scheduled.msg
with
| exn ->
  let bt = Printexc.get_raw_backtrace () in
  rollback_scheduled_msg store scheduled;
  Printexc.raise_with_backtrace exn bt
```

`Eio.Cancel.Cancelled` will match `exn`, then the code attempts a mutex-taking rollback while cancellation may still be pending. If `rollback_scheduled_msg` itself observes cancellation, the original exception/backtrace can be lost, and cleanup under cancellation is generally discouraged in Eio.

**Fix:** match `Eio.Cancel.Cancelled` first and re-raise immediately; only perform rollback for non-cancellation exceptions.

```ocaml
try Eio.Stream.add store.flusher_inbox scheduled.msg
with
| exn ->
  let bt = Printexc.get_raw_backtrace () in
  rollback_scheduled_msg store scheduled;
  Printexc.raise_with_backtrace exn bt
```

### P1 — Capacity guard is racy and does not prevent blocking
`lib/board/board_core_persist.ml`, `enqueue_scheduled_msgs` (around new lines 52–63):

The check uses `Eio.Stream.length`, which is a point-in-time observation. If `length = 999` and two messages are scheduled, the guard passes and the second `Eio.Stream.add` will block until the flusher drains the stream. The PR correctly states that the mutex is not held during the add, but callers can still block on a nearly-full inbox.

**Fix:** either document that the guard is best-effort and only avoids enqueueing when the stream is already full, or use a bounded-queue primitive that supports non-blocking add-if-room. If the desired semantics are “drop when full”, the guard should probably be `>= capacity - scheduled_count`.

### P2 — Magic capacity constant remains hardcoded
`lib/board/board_core_persist.ml` line 9:

```ocaml
let flusher_inbox_capacity = 1000
```

Naming the constant is better than the inline `Eio.Stream.create 1000`, but `1000` is still a magic number. The existing `board_core.mli` doc also still says “capped at 1000 messages” (`lib/board/board_core.mli:70`), which is now a duplicated literal.

**Fix:** add `Env_config.Board.flusher_inbox_capacity` (or `Limits.flusher_inbox_capacity`) and reference it from both the implementation and the `.mli` doc. If that is out of scope for this PR, at least update the `board_core.mli` comment to reference `Board_core_persist.flusher_inbox_capacity`.

### P2 — Unnecessary mutable state inside the lock
`lib/board/board_core_persist.ml`, inside `maybe_sweep` (around new lines 68–97):

```ocaml
let scheduled = ref [] in
let schedule msg previous_ts =
  scheduled := { msg; previous_ts; reserved_ts = now } :: !scheduled
in
...
List.rev !scheduled
```

This is a purely local list accumulation that does not need a ref. A conditional expression returning `[sweep_msg]` / `[flush_msg]` / `[sweep_msg; flush_msg]` is clearer and avoids mutable state.

### P2 — Rollback equality check is brittle
`lib/board/board_core_persist.ml`, `rollback_scheduled_msg` (around new lines 32–41):

```ocaml
if Stdlib.Float.equal store.last_sweep scheduled.reserved_ts
then store.last_sweep <- scheduled.previous_ts
```

Between reserving the timestamp and rolling it back, another fiber (e.g., the flusher actor running `Board.sweep` or `Board.flush_dirty`) may update `last_sweep`/`last_flush` to a different value. The equality then fails, so the rollback is skipped even though the message was never enqueued. `Eio.Stream.add` rarely raises, so this is a low-probability edge case, but the rollback invariant is not guaranteed.

**Fix:** document the limitation, or store a generation counter / token instead of relying on float equality.

### P2 — New public(ish) value lacks documentation
`lib/board/board_core_persist.mli` adds:

```ocaml
val flusher_inbox_capacity : int
```

It is exposed only through the internal `Board_core_persist` interface (not the `Board_core` facade), but it still has no doc comment. Since it exists primarily to support the test, consider whether the test should consume it through an internal test-only alias rather than widening the module signature.

## Cluster / radius assessment

This is a narrow board-persistence scheduling fix, not a cross-keeper consensus change (as the PR body notes). The blast radius is limited to `Board_core_persist.maybe_sweep`, the flusher actor inbox, and the new tests. It does not touch keeper orchestration, env/path SSOT, or the OAS repo. It is safe in isolation, but the red CI means it cannot land yet.

## Recommended actions before approval

1. Investigate and resolve (or confirm as pre-existing) the `Build and Test` failures.
2. Handle `Eio.Cancel.Cancelled` explicitly in `enqueue_scheduled_msg`.
3. Decide whether the capacity guard is intended to be strict or best-effort, and make the condition/document match.
4. Move `flusher_inbox_capacity` to `Env_config`/`Limits` or update the `board_core.mli` doc to avoid a duplicated magic literal.
5. Replace the `scheduled` ref with a pure list construction.
6. Add a doc comment for `flusher_inbox_capacity` or restrict it to test-only access.

---

*Methodology note: an initial `gh pr diff` without an explicit `--repo` returned unrelated `lib/fs_result.ml` changes. The review above is based on the commit diff from `gh api repos/jeong-sik/masc/commits/c5d9c842...` and `gh pr diff 22401 --repo jeong-sik/masc`, which correctly show the board scheduling changes.*

jeong-sik:
Addressed the latest adversarial P1/P2 findings in c95e0f56386.

What changed:
- Eio.Cancel.Cancelled is now matched before the generic enqueue rollback handler, so cancellation is re-raised directly instead of attempting mutex-taking cleanup under cancellation.
- Full-inbox scheduling now increments a monotonic dropped-schedule counter before rollback/logging.
- The capacity guard is expressed as whole-batch room (`requested <= capacity - queued`) and documents why the following adds should not block: maybe_sweep is the only producer and timestamp reservations are made under store.mutex before the enqueue batch.
- Removed the local mutable scheduled ref in maybe_sweep; scheduling now builds an immutable local list.
- Moved test schedule timestamp reset/read behind Board_core_persist test helpers, so test_board_ttl no longer mutates or reads store.last_sweep/store.last_flush directly.
- Replaced the duplicated public doc literal "1000 messages" with a reference to the persistence-layer flusher inbox capacity.

Validation passed:
- ocamlformat --check lib/board/board_core_persist.ml lib/board/board_core_persist.mli lib/board/board_core.mli test/test_board_ttl.ml
- ocamlc -stop-after parsing lib/board/board_core_persist.ml lib/board/board_core_persist.mli lib/board/board_core.mli test/test_board_ttl.ml
- git diff --check
- python3 scripts/check_test_coverage.py
- scripts/check-oas-pin.sh --local-only
- DUNE_CACHE=disabled scripts/dune-local.sh build test/test_board_ttl.exe
- ./_build/default/test/test_board_ttl.exe: 16 tests passed

Supplemental boundary:
- I attempted DUNE_CACHE=disabled scripts/dune-local.sh build @check, but it remained queued behind another worktree holding /tmp/me-dune-local.lock, so I interrupted it before it acquired the lock. No @check result was produced locally.

jeong-sik:
[Adv Review]

Reviewed rebased head `c95e0f5638673e35ee2a3192313aa0a8a1a502f2` against the prior `[Adv Review]` follow-up (`0f43277b`). The reply fixes several concrete findings, but the core P1 design blockers and the P0 CI gate remain open.

## Acknowledged fixes

- **Fixed:** `enqueue_scheduled_msg` now matches `Eio.Cancel.Cancelled _ as e -> raise e` before the generic rollback handler (`lib/board/board_core_persist.ml:262-269`). Cancellation is no longer dragged through mutex-taking rollback.
- **Fixed:** `maybe_sweep` no longer accumulates via a local `ref`; it builds an immutable `scheduled` list under the lock (`lib/board/board_core_persist.ml:292-317`).
- **Fixed:** `test/test_board_ttl.ml` no longer reads or mutates `store.last_sweep` / `store.last_flush` directly; it uses the new test helpers (`lib/board/board_core_persist.mli:43-46`, `test/test_board_ttl.ml:145-149`).
- **Fixed:** `lib/board/board_core.mli:70` no longer hardcodes the literal `"1000 messages"`; it references the persistence-layer capacity.
- **Partially fixed:** A monotonic dropped-schedule counter was added (`lib/board/board_core_persist.ml:18-25`), but it is not integrated with the structured metrics system and uses a loop of `Atomic.incr` (see P2 below).

## Remaining blockers

### P0 — CI gate must be green before merge
`statusCheckRollup` for `c95e0f5` shows many required checks still `QUEUED` / `IN_PROGRESS`; no completed failure is recorded on this head yet, but the previous heads (`c5d9c84`, `22f0ad6`) had a red `Build and Test` / `CI Gate`. Do not approve until the required suite is green and the author confirms any remaining failures are pre-existing on `main`.

### P1 — Caller fiber still blocks on `Eio.Stream.add`
`enqueue_scheduled_msg` performs the add directly on the caller fiber:
```ocaml
lib/board/board_core_persist.ml:263
  try Eio.Stream.add store.flusher_inbox scheduled.msg
```
The capacity guard is best-effort only. If the flusher is slow (or any future code adds to `flusher_inbox`), the caller will block inside a board-handler fiber. Decide whether this is intentional backpressure; if not, replace with a non-blocking add-if-room primitive or move the enqueue to a dedicated scheduler fiber.

### P1 — Capacity guard encodes an ambient invariant in a comment, not code
```ocaml
lib/board/board_core_persist.ml:274
  let inbox_len = Eio.Stream.length store.flusher_inbox in
...
lib/board/board_core_persist.ml:286-289
    (* [maybe_sweep] is the only producer of these messages and reserves
       timestamps under [store.mutex] before this check. Once there is room for
       the whole batch, these adds should not block on capacity. *)
```
The guarantee that adds "should not block" depends on a global assumption documented in a comment. That is a heuristic, not an enforced invariant. The check is also non-atomic with respect to the subsequent `Eio.Stream.add` calls.

### P1 — Board mutex helper uses `Eio.Mutex.use_rw ~protect:true`, which poisons on exception
```ocaml
lib/board/board_core_persist.ml:127
let with_lock store f = Eio.Mutex.use_rw ~protect:true store.mutex (fun () -> f ())
```
The new scheduling path takes this lock in `maybe_sweep`, `rollback_scheduled_msg`, and the test helpers. If any callback raises, the board mutex is poisoned and all future scheduling deadlocks. This helper is pre-existing, but the PR increases the number of lock sites. Either switch to `~protect:false` and handle cleanup explicitly, or document that the board mutex must never observe an exception and add a runtime assertion.

## P2 findings

### P2 — `flusher_inbox_capacity` is still a hardcoded magic constant
```ocaml
lib/board/board_core_persist.ml:14
let flusher_inbox_capacity = 1000
```
The PR already reads `flush_interval_sec` from `Env_config.Board`; the inbox capacity should also come from `Env_config.Board` / `Limits` so the implementation, the `.mli` doc, and the stream creation share a single source of truth.

### P2 — Dropped counter increments one-by-one instead of atomically
```ocaml
lib/board/board_core_persist.ml:21-25
let record_flusher_schedule_drop scheduled_count =
  for _ = 1 to scheduled_count do
    Atomic.incr flusher_schedule_dropped
  done
```
Use `Atomic.fetch_and_add flusher_schedule_dropped scheduled_count` for a single atomic batch update.

### P2 — Rollback relies on brittle float equality
```ocaml
lib/board/board_core_persist.ml:255
      if Stdlib.Float.equal store.last_sweep scheduled.reserved_ts
      then store.last_sweep <- scheduled.previous_ts
lib/board/board_core_persist.ml:258
      if Stdlib.Float.equal store.last_flush scheduled.reserved_ts
      then store.last_flush <- scheduled.previous_ts
```
If the flusher actor processes a sweep/flush and updates the timestamp between reservation and rollback, the equality fails and the rollback is silently skipped even though the message was never enqueued. Use a generation counter / token, or document explicitly that rollback is best-effort and may leave a stale reservation.

### P2 — Public `.mli` surface widened for test-only helpers
```ocaml
lib/board/board_core_persist.mli:19
val flusher_inbox_capacity : int
lib/board/board_core_persist.mli:43-46
val reset_sweep_schedule_for_test : store -> unit
val sweep_schedule_timestamps_for_test : store -> float * float
```
These are exposed in the production signature only to support tests. Prefer an internal `Board_core_persist.For_test` alias or `[@@alert test_only]` so production callers cannot depend on them.

### P2 — Test still depends on a magic `0.0` sentinel
```ocaml
test/test_board_ttl.ml:114
let schedule_reset_timestamp_for_test = 0.0
...
test/test_board_ttl.ml:188-194
    schedule_reset_timestamp_for_test
    sweep_ts;
  ...
    schedule_reset_timestamp_for_test
    flush_ts;
```
Moving the reset behind a helper did not remove the sentinel from the assertions. Use an explicit `Unset` variant or compare against the helper-defined constant from the implementation side.

## Verdict

The cancellation handling and immutable scheduling list are real improvements, and the test no longer pokes at internal mutable fields directly. However, the P0 CI gate is unresolved, and the P1 design issues — blocking `Eio.Stream.add`, a heuristic capacity guard, and a poisonable board mutex — are still present. Do not merge until P0 is green and the P1 items are either fixed or explicitly accepted with a design rationale recorded in the code.

jeong-sik:
## Re: P1 / P2 — P2 fixed in `7def4cce`, P1 + rest scoped to a follow-up PR

**P2 — fixed + verified (`test_board_ttl` 16/16):**
- non-atomic drop counter → single `Atomic.fetch_and_add` (no partial count for concurrent readers).
- `flusher_inbox_capacity` 1000 → `Env_config.Board.flusher_inbox_capacity` (`MASC_BOARD_FLUSHER_INBOX_CAPACITY`), the SSOT now shared by the implementation, the `.mli` doc, and `Eio.Stream.create`, mirroring `flush_interval_sec`.

**P1-1 / P1-2 (Stream.add blocking + non-atomic capacity guard) → follow-up PR.** I verified the invariant: producer = `enqueue_scheduled_msg` (single, via `maybe_sweep`), consumer = `board_dispatch` flusher actor (single). `Eio.Stream` has **no** non-blocking `add` (only `take_nonblocking`), so the real fix is a dedicated scheduler fiber: `store.schedule_inbox` (large capacity) receives from `maybe_sweep` (caller never blocks on `flusher_inbox`), and a `fork_daemon` fiber drains `schedule_inbox → flusher_inbox.add` (block tolerated inside the fiber only). This needs a `board_types` store-type change (move `scheduled_flusher_msg` up, add `schedule_inbox`) and a `test_board_ttl` rewrite — too large for this hotfix, tracked separately.

**P1-3 (mutex poison) → API correction.** The review's suggested `~protect:false` does **not** avoid poison: per `eio_mutex.mli`, `use_rw` disables the mutex on any `fn` exception **regardless of `protect`** (`protect` only guards against cancellation). `use_ro` unlocks on exception but is read-only, and `with_lock` writes (`last_sweep`/`last_flush`), so neither fits. Root fix = guarantee the callback never raises (hard: callbacks contain `Log`/`Yojson`/`Hashtbl`) or a manual `lock`/`Fun.protect` with a cancel guard. Bundled with the fiber follow-up.

**P2-3 (rollback float equality) → safe as-is.** Only `maybe_sweep` (reserve) and `rollback_scheduled_msg` write `last_sweep`/`last_flush`, both under `store.mutex`; concurrent `maybe_sweep` is mutex-serialized, so `reserved_ts` is stable between reserve and rollback-on-add-failure. A generation counter is over-engineering for this single-reservation structure.

**P2-4 / P2-5 → follow-up** (`For_test` sub-module for the test helpers, explicit unset constant for the `0.0` sentinel), bundled with the fiber PR to avoid double-churning `test_board_ttl`.

jeong-sik:
[Adv Review]

Reviewed current head `7def4cce05180fbe4c6222eb692dfbfabfd496b0`. The reply in `7def4cce` resolves the concrete P2 items from the prior `[Adv Review]`, but this PR was merged while the required CI gate was still incomplete and several P1/P2 issues remain in the merged tree.

## Fixed since prior `[Adv Review]`

- **Fixed:** `flusher_inbox_capacity` is no longer a literal `1000`; it is sourced from `Env_config.Board.flusher_inbox_capacity` (`lib/board/board_core_persist.ml:14`), backed by the new env var `MASC_BOARD_FLUSHER_INBOX_CAPACITY` in `lib/config/env_config_runtime.ml:476-477`.
- **Fixed:** dropped-schedule counter now updates in a single atomic batch via `Atomic.fetch_and_add` (`lib/board/board_core_persist.ml:25`).
- **Fixed:** `lib/board/board_core.mli:70` no longer hardcodes `1000 messages`; it references the persistence-layer capacity.
- **Fixed:** tests no longer read or mutate `store.last_sweep` / `store.last_flush` directly; they use the new test helpers.

## Remaining blockers

### P0 — PR merged before required CI gate was green
`statusCheckRollup` on `7def4cce` shows:
- `Env knob classification` — **FAILURE**
- `Build and Test`, `CI Gate`, `Health`, `Lint`, `Dashboard`, `TLA+ Model Checking` — **CANCELLED** (the merge aborted the run)

Merging with a red env-knob check and an incomplete CI gate violates the repo workflow. Verify on `main` or explain why the failure is acceptable.

### P1 — New env var captured at module load
```ocaml
lib/config/env_config_runtime.ml:476-477
  let flusher_inbox_capacity =
    get_int ~default:1000 "MASC_BOARD_FLUSHER_INBOX_CAPACITY"
```
This evaluates the env var once when `Env_config` loads. Per env-var SSOT rules, resolve it at call time (`let flusher_inbox_capacity () = ...`) or wrap it in `Lazy.t` so runtime config changes are observable.

### P1 — Caller fiber blocks on `Eio.Stream.add`
```ocaml
lib/board/board_core_persist.ml:263-264
let enqueue_scheduled_msg store scheduled =
  try Eio.Stream.add store.flusher_inbox scheduled.msg
```
If the flusher actor is slow, `maybe_sweep` callers block inside board-handler fibers. The author scoped the scheduler-fiber redesign to a follow-up PR; this remains an open P1 in the merged tree.

### P1 — Capacity guard is non-atomic and relies on a comment invariant
```ocaml
lib/board/board_core_persist.ml:273-290
let enqueue_scheduled_msgs store scheduled =
  let scheduled_count = List.length scheduled in
  let inbox_len = Eio.Stream.length store.flusher_inbox in
  ...
  else if scheduled_count > flusher_inbox_capacity - inbox_len
  ...
  (* [maybe_sweep] is the only producer of these messages and reserves
     timestamps under [store.mutex] before this check. Once there is room for
     the whole batch, these adds should not block on capacity. *)
  List.iter (enqueue_scheduled_msg store) scheduled)
```
The guard is not atomic with the subsequent `add`s, and the "should not block" guarantee is an ambient assumption, not enforced. Bundle with the scheduler-fiber follow-up or replace with a non-blocking add primitive.

### P1 — Board mutex helper poisons on exception
```ocaml
lib/board/board_core_persist.ml:128
let with_lock store f = Eio.Mutex.use_rw ~protect:true store.mutex (fun () -> f ())
```
Per the Eio docs, `use_rw` disables the mutex on any callback exception; `protect` only affects cancellation handling. The new scheduling paths (`maybe_sweep`, `rollback_scheduled_msg`, test helpers) increase the number of sites that can trigger this. Switch to manual `lock`/`Fun.protect` with a cancel guard, or guarantee callbacks never raise and assert it.

## P2 findings

### P2 — Rollback uses brittle float equality
```ocaml
lib/board/board_core_persist.ml:256-260
      if Stdlib.Float.equal store.last_sweep scheduled.reserved_ts
      then store.last_sweep <- scheduled.previous_ts
```
Even though only `maybe_sweep` and `rollback_scheduled_msg` currently mutate these fields under the same mutex, float equality is a latent correctness hazard. Replace with a generation token or document explicitly that rollback is best-effort and may leave a stale reservation.

### P2 — Dropped counter is an orphan metric
```ocaml
lib/board/board_core_persist.ml:18-19
let flusher_schedule_dropped = Atomic.make 0
let flusher_schedule_dropped_count () = Atomic.get flusher_schedule_dropped
```
Not wired into the structured metrics system. Add a labeled metric or remove the public counter.

### P2 — Production `.mli` widened for test-only helpers
```ocaml
lib/board/board_core_persist.mli:43-47
val reset_sweep_schedule_for_test : store -> unit
val sweep_schedule_timestamps_for_test : store -> float * float
```
Exposed in the production signature only for tests. Move them to a `For_test` submodule or mark with `[@@alert test_only]`.

### P2 — Test still asserts a magic `0.0` sentinel
```ocaml
test/test_board_ttl.ml:114
let schedule_reset_timestamp_for_test = 0.0
...
test/test_board_ttl.ml:190-194
    schedule_reset_timestamp_for_test
    sweep_ts;
  ...
    schedule_reset_timestamp_for_test
    flush_ts;
```
The sentinel leaked from the implementation into the test. Use an explicit `Unset` variant or a helper-defined constant.

## Verdict

`7def4cce` fixes the easy P2s, but the PR was merged before the CI gate finished and with an env-knob classification failure still red. Do not treat this branch as fully resolved until the CI gate is green/explained on `main` and the P1 items above are fixed or explicitly accepted with code-level rationale.

PR #22406, opened 2026-06-26
fix(streamable): parse JSON content type via negotiation

jeong-sik:
Summary
- add Http_negotiation.is_json_content_type as the SSOT for Content-Type JSON media checks
- replace streamable_http's String.sub prefix heuristic with the negotiation helper
- keep application/json; charset=utf-8 accepted, while rejecting application/jsonp, wildcards, and Accept-style lists for Content-Type

Note
- The review note said charset parameters were rejected; current code actually accepted charset but also over-accepted application/json* prefixes. This PR fixes the real current-code failure mode without adding a new parser.

Validation
- ocamlformat --check lib/mcp_transport_protocol/mcp_transport_protocol.ml lib/mcp_transport_protocol/mcp_transport_protocol.mli lib/streamable_http.ml test/test_mcp_protocol_coverage.ml test/test_streamable_http.ml
- scripts/dune-local.sh build test/test_mcp_protocol_coverage.exe test/test_streamable_http.exe  # stopped after shared MASC dune-local lock wait; CI is the build authority

### DIFF
--- a/lib/mcp_transport_protocol/mcp_transport_protocol.ml
+++ b/lib/mcp_transport_protocol/mcp_transport_protocol.ml
@@ -149,6 +149,15 @@ module Http_negotiation = struct
             (type_ = "application" && subtype = "json")
             || (type_ = "*" && subtype = "*"))
 
+  let is_json_content_type = function
+    | None -> false
+    | Some h ->
+        (match Mcp_protocol.Http_negotiation.parse_accept_header h with
+         | [ mt ] ->
+             String.equal (String.lowercase_ascii mt.type_) "application"
+             && String.equal (String.lowercase_ascii mt.subtype) "json"
+         | [] | _ :: _ :: _ -> false)
+
   let accepts_streamable_mcp = function
     | None -> false
     | Some h ->
--- a/lib/mcp_transport_protocol/mcp_transport_protocol.mli
+++ b/lib/mcp_transport_protocol/mcp_transport_protocol.mli
@@ -99,6 +99,11 @@ module Http_negotiation : sig
   val accepts_json : string option -> bool
   (** [true] iff the header advertises [application/json] or [*\/*]. *)
 
+  val is_json_content_type : string option -> bool
+  (** [true] iff the Content-Type value is exactly [application/json],
+      allowing media-type parameters such as [charset]. Wildcards and
+      comma-separated Accept-style lists are rejected. *)
+
   val accepts_streamable_mcp : string option -> bool
   (** [true] iff the header advertises both JSON and SSE
       (the Streamable HTTP transport requirement). *)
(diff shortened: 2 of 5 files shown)


anyang-keepers:
## 🟢 ACCEPT — Proper Content-Type Parsing

**Size:** +110/-4, 5 files  
**Claim:** "Parse JSON content type via negotiation"

**Review:**
- Replaces brittle string-prefix check (`String.sub s 0 16 = "application/json"`) with proper media-type parsing
- Old code would incorrectly match `application/json-patch+json` or fail on `application/json; charset=utf-8`
- New code correctly rejects wildcards and comma-separated Accept-style lists
- Test coverage comprehensive: None, exact, with charset, case-insensitive, wildcard rejection, list rejection

**Nit:** Consider if `application/json-seq` or `application/json5` should also be handled — but strict `application/json` is correct per MCP spec.

jeong-sik:
Current-head review for `c3bc03591f589799adefb206ee23a8d71cf59a11`:

No source blocker found. The old `application/json*` prefix path is gone, and the new helper centralizes Content-Type classification through the existing media-type parser. The behavior matches the streamable boundary I expect: `application/json` and `application/json; charset=utf-8` are accepted, while `application/jsonp`, wildcard values, and Accept-style comma lists are rejected.

CI status at review time: all reported checks are green or skipped except `Build and Test`, which is still in progress. Treating this as source-unblocked, not merge-ready until that job finishes green.

jeong-sik:
## Adversarial review — MASC#22406

### Prior review state
The original brittle `String.sub` prefix check is gone, and obvious over-acceptance of `application/json*` prefixes is fixed.

### P1 — `Content-Type` is validated with an `Accept`-list parser
`Http_negotiation.is_json_content_type` calls `Mcp_protocol.Http_negotiation.parse_accept_header` on a `Content-Type` value. That parser is designed for comma-separated `Accept` lists: it splits on `,` and sorts by `q`. `Content-Type` is a single media-type and does not permit `q` parameters or comma-separated alternatives.

Consequences:
- Valid `Content-Type` values with quoted parameters containing commas are mis-parsed / truncated.
- Invalid values such as `application/json;q=0.5` are accepted.

Fix: Use `Mcp_protocol.Http_negotiation.parse_media_type` directly (returns one `media_type option`, does not split on commas). Reject any `q` parameter because `q` is not a `Content-Type` parameter.

### P1 — Test gap around invalid `Content-Type` forms
New tests do not cover:
- `application/json;q=...` (invalid, currently accepted)
- quoted parameters containing commas
- suffix variants such as `application/json-seq`, `application/json5`, `application/ld+json`

Add negative cases so the parser contract is pinned.

### Nit — duplicated lowercasing
`exists_accepted` already lowercases `type_` / `subtype`; `is_json_content_type` re-implements the same lowering. Centralize on a single helper.

**Verdict:** No P0; two P1 issues (using an `Accept`-list parser for `Content-Type`, plus missing negative test coverage) and one refactor nit.

PR #22407, opened 2026-06-26
fix(http): delegate legacy zstd compression to codec

jeong-sik:
Summary
- route Http_server_eio.Compression.compress_zstd through the shared Compression_codec instead of calling Zstd.compress directly
- preserve the legacy 256-byte no-compress threshold and boolean return shape
- remove the silent Zstd.Error catch from the HTTP layer; Compression_codec owns compression failure diagnostics

Validation
- ocamlformat --check lib/http_server_eio.ml lib/http_server_eio.mli
- rg -n 'Zstd\.compress|with Zstd\.Error|Compression_codec\.compress' lib/http_server_eio.ml
- scripts/dune-local.sh build test/test_http_server_eio_coverage.exe  # stopped after shared MASC dune-local lock wait; CI is the build authority

### DIFF
--- a/lib/http_server_eio.ml
+++ b/lib/http_server_eio.ml
@@ -63,16 +63,12 @@ module Compression = struct
 
   (** Legacy: Standard zstd without dictionary *)
   let compress_zstd ?(level = 3) (data : string) : (string * bool) =
-    if String.length data < 256 then
-      (data, false)
+    if String.length data < 256
+    then data, false
     else
-      try
-        let compressed = Zstd.compress ~level data in
-        if String.length compressed < String.length data then
-          (compressed, true)
-        else
-          (data, false)
-      with Zstd.Error _ -> (data, false)
+      match Compression_codec.compress ~level data with
+      | Compression_codec.Unchanged payload -> payload, false
+      | Compression_codec.Compressed { payload; encoding = _ } -> payload, true
 end
 
 (** Late-response failure classifier (#13059).
--- a/lib/http_server_eio.mli
+++ b/lib/http_server_eio.mli
@@ -74,7 +74,9 @@ module Compression : sig
 
   (** [compress_zstd ?level data] is the legacy
       no-dictionary path.  Returns [(payload, did_compress)] —
-      data shorter than 256 bytes is kept as-is. *)
+      data shorter than 256 bytes is kept as-is.  Raw zstd
+      compression is delegated to {!Compression_codec}, which owns
+      compression failure diagnostics. *)
   val compress_zstd : ?level:int -> string -> string * bool
 end
 


anyang-keepers:
## 🟢 ACCEPT — Clean Delegation Refactor

**Size:** +6/-10, 2 files  
**Claim:** "Delegate legacy zstd compression to codec"

**Review:**
- Replaces inline `Zstd.compress` + error handling + ratio check with delegation to `Compression_codec.compress`
- Cleaner pattern matching: `Unchanged` vs `Compressed` instead of nested `if/then/else` + `try/with`
- Removes duplication: compression failure diagnostics now centralized in `Compression_codec`

**Nit:** Verify `Compression_codec` handles these edge cases (they were previously inline):
1. Compression ratio not beneficial (compressed >= original) → returns `Unchanged`
2. `Zstd.Error` exception → returns `Unchanged` (not propagated)
3. The 256-byte threshold is still checked before delegation (good)

If the codec doesn't handle #1 and #2, this refactor introduces a regression. Confirm with a quick test or codec source check.

jeong-sik:
## Adversarial review — MASC#22407

**Scope:** `lib/http_server_eio.ml`, `lib/http_server_eio.mli`  
**CI:** all green at current head.  
**Previous review follow-up:** `Compression_codec.compress` returns `Unchanged` both when compression does not shrink the payload and on `Zstd.Error`.

No P0 blockers.

### P1 — should fix before merge
1. **Hardcoded legacy threshold / SSOT violation.** `lib/http_server_eio.ml:66` still magic-numbers `256`. The codec already centralizes a `min_size` threshold (`lib/compression/compression_codec.ml:20`), and the sibling `compress` function uses that 32-byte floor. Having two undocumented thresholds for the same codec is a SSOT violation. Make `256` a named constant defined next to `Compression_codec.min_size`, or document why the legacy path needs a different floor.
2. **Discarded `encoding` weakens the "no-dictionary" contract.** `lib/http_server_eio.ml:71` matches `Compressed { payload; encoding = _ }`. The function and `.mli` still advertise this as the standard-zstd, no-dictionary legacy path. If `Compression_codec.compress` ever returns `Dictionary`, this call site silently emits dictionary-compressed bytes. Preserve the contract by matching on `encoding` and treating `Dictionary` as `Unchanged`.
3. **Broader silent-failure surface via `Failure` swallowing.** `Compression_codec.compress` catches `Failure msg | Zstd.Error msg` and returns `Unchanged` after logging. The old inline code only caught `Zstd.Error`, so invalid levels / allocator failures that raise `Failure` would propagate. The refactor funnels those through the same silent fallback. Because `compress_zstd` returns only `(string * bool)`, callers cannot distinguish a transient compression failure from an incompressible payload. Consider exposing a result type or documenting that `did_compress = false` may mask a logged failure.

**Verdict:** P0 none; P1 threshold SSOT, discarded encoding contract, and expanded silent-failure surface. Address before merge.

PR #22412, opened 2026-06-26
fix(evidence): tighten contract-token utility follow-up

jeong-sik:
## Summary

- Follow-up to merged #22384 review comments.
- Replace the contract-token UTF-8 byte check magic number with an explicit high-bit test and comment.
- Clarify `ascii_punctuation_tokens` docs so it is scoped to anti-rationalization contract-token matching, not evidence-ref boundary matching.
- Add edge coverage for empty haystack, single-token needle, needle longer than haystack, byte-literal non-ASCII case behavior, and long-input sequence matching.
- Remove unnecessary `;;` from the `anti_rationalization` helper bindings.

## Verification

- `ocamlformat --check lib/core/string_util.ml lib/core/string_util.mli test/test_string_util.ml lib/task/anti_rationalization.ml`
- `ocamlc -stop-after parsing lib/core/string_util.ml lib/core/string_util.mli test/test_string_util.ml lib/task/anti_rationalization.ml`
- `git diff --check`

No Dune/local build was run; CI remains the build authority.

### DIFF
--- a/lib/core/string_util.ml
+++ b/lib/core/string_util.ml
@@ -123,7 +123,9 @@ let query_tokens query =
 
 let is_ascii_alnum_or_utf8_byte = function
   | 'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' -> true
-  | c -> Char.code c >= 128
+  | c ->
+    (* Non-ASCII UTF-8 bytes have the high bit set; keep them in tokens. *)
+    Char.code c land 0x80 <> 0
 
 let ascii_punctuation_tokens text =
   let len = String.length text in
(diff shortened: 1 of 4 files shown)


jeong-sik:
No source blocker found on current head `5b188d7079ee`.

The utility change is scoped to making the non-ASCII byte predicate explicit (`Char.code c land 0x80 <> 0`) while preserving the existing byte-wise tokenizer behavior. The `.mli` wording now narrows `ascii_punctuation_tokens` to anti-rationalization contract-token matching, and the added tests cover the edge cases that matter for this helper: empty haystack/needle behavior, single-token matches, needle longer than haystack, byte-literal non-ASCII case behavior, and long haystack sequence matching.

CI is still pending, so this is source-review status rather than merge readiness.

PR #22410, opened 2026-06-26
fix(keeper): report malformed approval rule drops

jeong-sik:
## Summary

- Follow-up to merged #22394 review blocker: fail-closed approval-rule loading now reports malformed/unreadable persisted policy instead of silently returning `[]`.
- Record `keeper_approval_rules` persistence read drops for malformed rule entries/non-list files (`invalid_payload`) and unreadable/corrupt files (`entry_load_error`).
- Keep a missing rules file quiet as normal first-run state.
- Update regression coverage for malformed matching rules and corrupt rules file metric deltas.
- Update approval-rules docs to reference `rules_path` / `Workspace_utils.masc_dir_from_base_path`.

## Verification

- `ocamlformat --check lib/keeper/keeper_approval_queue_rules.ml test/test_keeper_approval_queue_rules.ml lib/keeper_contract/keeper_approval_queue_rules_types.ml test/test_keeper_approval_queue_rules_types.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_approval_queue_rules.ml test/test_keeper_approval_queue_rules.ml lib/keeper_contract/keeper_approval_queue_rules_types.ml test/test_keeper_approval_queue_rules_types.ml`
- `git diff --check origin/main...HEAD`

No Dune/local build was run; CI remains the build authority.

PR #22414, opened 2026-06-26
fix(test): use Common masc dir helper in approval rules

jeong-sik:
Fixes the merged #22410 test compile break by using Common.masc_dir_from_base_path instead of unqualified Workspace_utils in test/test_keeper_approval_queue_rules.ml.

Validation:
- ocamlformat --check test/test_keeper_approval_queue_rules.ml
- ocamlc -stop-after parsing test/test_keeper_approval_queue_rules.ml
- git diff --check
- bash scripts/lint/no-unknown-permissive-default.sh

Note: scripts/dune-local.sh build test/test_keeper_approval_queue_rules.exe was attempted, but local Dune was blocked by another bare dune process outside scripts/dune-local.sh.

### DIFF
--- a/test/test_keeper_approval_queue_rules.ml
+++ b/test/test_keeper_approval_queue_rules.ml
@@ -23,19 +23,19 @@ let cleanup_dir dir =
 
 let rules_path ~base_path =
   Filename.concat
-    (Workspace_utils.masc_dir_from_base_path ~base_path)
+    (Common.masc_dir_from_base_path ~base_path)
     "approval-rules.json"
 ;;
 
 let write_rules ~base_path json =
-  let masc_dir = Workspace_utils.masc_dir_from_base_path ~base_path in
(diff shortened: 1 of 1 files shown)


jeong-sik:
No code blocker on current head `fad2b98f1ec897031ae9f0c77a7e07033d33f50b`.

Reviewed the single-file diff in `test/test_keeper_approval_queue_rules.ml`: the stale unqualified `Workspace_utils.masc_dir_from_base_path` calls are replaced with `Common.masc_dir_from_base_path`, which is the existing SSOT and is available through the test target's `(libraries alcotest masc unix yojson)` stanza. The helper still writes and reads the same `.masc/approval-rules.json` path, so the malformed/corrupt persisted-rule assertions keep the same behavioral coverage.

Merge readiness is still gated on CI; most checks were queued/in progress when reviewed.

PR #22415, opened 2026-06-26
fix(test): use public masc dir helper

jeong-sik:
## Summary
- use Masc.Common.masc_dir_from_base_path in keeper approval rules test instead of unqualified Workspace_utils
- keep the test stanza scoped to the existing masc dependency

## Validation
- git diff --check
- ocamlc -stop-after parsing test/test_keeper_approval_queue_rules.ml
- ocamlformat --check test/test_keeper_approval_queue_rules.ml
- scripts/dune-local.sh build test/test_keeper_approval_queue_rules.exe (stopped after waiting on /tmp/me-dune-local.lock held by another worktree build)

### DIFF
--- a/test/stanzas/test_keeper_approval_queue_rules.inc
+++ b/test/stanzas/test_keeper_approval_queue_rules.inc
@@ -1,4 +1,4 @@
 (test
  (name test_keeper_approval_queue_rules)
  (modules test_keeper_approval_queue_rules)
- (libraries alcotest masc unix yojson))
+ (libraries alcotest masc masc_test_deps unix yojson))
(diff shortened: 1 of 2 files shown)


jeong-sik:
Reviewed current head `ff6dc315796ce9dc4b877d23dab527e4b8d9a899`.

No source blocker found. This is a narrow test-only fix: `test/test_keeper_approval_queue_rules.ml` now uses `Masc.Common.masc_dir_from_base_path`, and the test stanza already depends on `masc`, so it avoids reaching through the unqualified `Workspace_utils` path without adding a new library dependency.

CI boundary: ocamlformat, Health, Build/Test, and Lint were still in progress when I checked, so merge readiness is still CI-bound.

PR #22437, opened 2026-06-27
fix(memory): surface legacy keeper inventory

jeong-sik:
## Summary

- Add `GET /api/v1/dashboard/legacy-keeper-inventory` as a read-only legacy `.masc/keepers` inventory surface.
- Classify bounded scan entries as `live`, `migrated`, `orphaned`, `backup`, or `unknown`.
- Return per-class count/byte totals plus an explicitly disabled cleanup surface. Cleanup candidates are intentionally not emitted until owner-verification logic exists.

## Review Response Notes

- Legacy Memory OS filename mapping is routed through `Keeper_memory_os_io.current_path_for_legacy_memory_filename`.
- Supported legacy memory filenames are represented by `Keeper_memory_os_io.legacy_memory_file` plus `supported_legacy_memory_files`, rather than dashboard-owned raw string checks.
- The cleanup plan no longer advertises always-zero `candidate_count` / `candidates`; it reports `candidate_policy=not_available_until_owner_verified` and `candidate_source=inventory_only`.
- Current-store access failures are surfaced as `scan_errors` via `lstat_current_memory` and classify the entry as `unknown`.
- Unexpected exception messages use a bounded exception slot label instead of `Printexc.to_string`.
- The dashboard cache key uses a SHA256 base-path hash prefix instead of embedding the raw workspace path.
- `relative_path` no longer falls back to returning a raw absolute path for paths outside the scan root.

## Audit Mapping

Addresses the P1 Memory OS audit gap:

- `docs/audit/2026-06-26-memory-os-production-audit.md`: "Legacy `.masc/keepers` Garbage Is Outside Current Cleanup"

This PR does not delete, move, or archive any legacy file. It only reports inventory. Cleanup candidate selection remains disabled until owner-verification logic is implemented.

## Validation

- `ocamlformat --check lib/keeper/keeper_memory_os_io.ml lib/keeper/keeper_memory_os_io.mli lib/server/server_dashboard_http_legacy_keeper_inventory.ml lib/server/server_dashboard_http_legacy_keeper_inventory.mli lib/server/server_routes_http_routes_dashboard.ml test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_memory_os_io.ml lib/server/server_dashboard_http_legacy_keeper_inventory.ml lib/server/server_routes_http_routes_dashboard.ml test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing -intf lib/keeper/keeper_memory_os_io.mli lib/server/server_dashboard_http_legacy_keeper_inventory.mli`
- `git diff --check HEAD~1 HEAD`
- `scripts/audit-path-ssot.sh`
- targeted `rg` check for stale cleanup-candidate, legacy filename, raw-base-path cache, and raw absolute-path fallback patterns

Not run per current instruction:

- Dune build/test
- CI/check-log inspection

### DIFF
--- a/lib/config_dir_resolver/config_dir_resolver.ml
+++ b/lib/config_dir_resolver/config_dir_resolver.ml
@@ -369,6 +369,9 @@ let resolve_for_base_path ~base_path =
 let keepers_dir_for_base_path ~base_path =
   (resolve_for_base_path ~base_path).keepers.path
 
+let keeper_runtime_store_of_dirname =
+  Common.keeper_runtime_store_of_dirname
+
 let personas_dir_opt () =
   let resolution = resolve () in
   match resolution.config_root.source with
--- a/lib/config_dir_resolver/config_dir_resolver.mli
+++ b/lib/config_dir_resolver/config_dir_resolver.mli
@@ -91,6 +91,10 @@ val keepers_dir_for_base_path : base_path:string -> string
 (** [keepers_dir_for_base_path ~base_path] returns the keepers directory for an
     explicit workspace base path. *)
 
+val keeper_runtime_store_of_dirname : string -> Common.keeper_runtime_store option
+(** Base-path-independent resolver for canonical child-store names under
+    [Common.keepers_runtime_dirname]. *)
+
 val personas_dirs_for_base_path : base_path:string -> string list
 (** Base-path-scoped variant of {!personas_dirs}. *)
 
--- a/lib/core/common.ml
+++ b/lib/core/common.ml
@@ -56,6 +56,42 @@ let masc_dir_from_base_path ~base_path =
 let keepers_runtime_dir_of_base ~base_path =
   Filename.concat (masc_dir_from_base_path ~base_path) keepers_runtime_dirname
 
+type keeper_runtime_store =
+  | Keeper_tool_usage
+  | Keeper_runtime_manifests
+  | Keeper_metrics
+  | Keeper_execution_receipts
+  | Keeper_turn_records
+  | Keeper_reaction_ledger
+  | Keeper_trajectories
+
+let keeper_runtime_store_dirname = function
+  | Keeper_tool_usage -> "tool_usage"
+  | Keeper_runtime_manifests -> "runtime-manifests"
+  | Keeper_metrics -> "metrics"
+  | Keeper_execution_receipts -> "execution-receipts"
+  | Keeper_turn_records -> "turn-records"
+  | Keeper_reaction_ledger -> "reaction-ledger"
+  | Keeper_trajectories -> "trajectories"
+
+let keeper_runtime_stores =
+  [ Keeper_tool_usage
+  ; Keeper_runtime_manifests
+  ; Keeper_metrics
+  ; Keeper_execution_receipts
+  ; Keeper_turn_records
+  ; Keeper_reaction_ledger
+  ; Keeper_trajectories
+  ]
+
+let keeper_runtime_store_dirnames =
+  List.map keeper_runtime_store_dirname keeper_runtime_stores
+
+let keeper_runtime_store_of_dirname name =
+  List.find_opt
+    (fun store -> String.equal name (keeper_runtime_store_dirname store))
+    keeper_runtime_stores
+
 let auth_dir_from_base_path ~base_path =
   Filename.concat (masc_dir_from_base_path ~base_path) "auth"
 
--- a/lib/core/common.mli
+++ b/lib/core/common.mli
@@ -54,6 +54,20 @@ val keepers_runtime_dir_of_base : base_path:string -> string
     (default cluster). Low-level SSOT that avoids the Workspace dependency cycle
     the cluster-aware [Workspace.keepers_runtime_dir] would impose. *)
 
+type keeper_runtime_store =
+  | Keeper_tool_usage
+  | Keeper_runtime_manifests
+  | Keeper_metrics
+  | Keeper_execution_receipts
+  | Keeper_turn_records
+  | Keeper_reaction_ledger
+  | Keeper_trajectories
+(** Canonical child-store names under {!keepers_runtime_dirname}. *)
+
+val keeper_runtime_store_dirname : keeper_runtime_store -> string
+val keeper_runtime_store_of_dirname : string -> keeper_runtime_store option
+val keeper_runtime_store_dirnames : string list
+
 val auth_dir_from_base_path : base_path:string -> string
 (** [<base_path>/.masc/auth]. SSOT path so {!Auth} and
     {!Keeper_identity} can both compute it without depending on each
--- a/lib/keeper/keeper_memory_os_io.ml
+++ b/lib/keeper/keeper_memory_os_io.ml
@@ -83,12 +83,28 @@ let events_path_for_keepers_dir ~keepers_dir ~keeper_id =
   Filename.concat keepers_dir (keeper_id ^ ".events.jsonl")
 ;;
 
+type legacy_memory_file =
+  | Legacy_facts
+  | Legacy_events
+
+let supported_legacy_memory_files = [ Legacy_facts; Legacy_events ]
+
+let legacy_memory_filename = function
+  | Legacy_facts -> "facts.jsonl"
+  | Legacy_events -> "events.jsonl"
+;;
+
+let legacy_memory_file_of_filename filename =
+  List.find_opt
+    (fun legacy_file -> String.equal filename (legacy_memory_filename legacy_file))
+    supported_legacy_memory_files
+;;
+
 let current_path_for_legacy_memory_filename ~keepers_dir ~keeper_id ~filename =
-  if String.equal filename "facts.jsonl"
-  then Some (facts_path_for_keepers_dir ~keepers_dir ~keeper_id)
-  else if String.equal filename "events.jsonl"
-  then Some (events_path_for_keepers_dir ~keepers_dir ~keeper_id)
-  else None
+  match legacy_memory_file_of_filename filename with
+  | Some Legacy_facts -> Some (facts_path_for_keepers_dir ~keepers_dir ~keeper_id)
+  | Some Legacy_events -> Some (events_path_for_keepers_dir ~keepers_dir ~keeper_id)
+  | None -> None
 ;;
 
 let events_path ~keeper_id =
--- a/lib/keeper/keeper_memory_os_io.ml
+++ b/lib/keeper/keeper_memory_os_io.ml
@@ -83,6 +83,14 @@ let events_path_for_keepers_dir ~keepers_dir ~keeper_id =
   Filename.concat keepers_dir (keeper_id ^ ".events.jsonl")
 ;;
 
+let current_path_for_legacy_memory_filename ~keepers_dir ~keeper_id ~filename =
+  if String.equal filename "facts.jsonl"
+  then Some (facts_path_for_keepers_dir ~keepers_dir ~keeper_id)
+  else if String.equal filename "events.jsonl"
+  then Some (events_path_for_keepers_dir ~keepers_dir ~keeper_id)
+  else None
+;;
+
 let events_path ~keeper_id =
   events_path_for_keepers_dir ~keepers_dir:(keepers_dir ()) ~keeper_id
 ;;
--- a/lib/keeper/keeper_memory_os_io.mli
+++ b/lib/keeper/keeper_memory_os_io.mli
@@ -21,6 +21,13 @@ val list_fact_store_keeper_ids_for_base_path : base_path:string -> string list
 
 val events_path : keeper_id:string -> string
 val events_path_for_keepers_dir : keepers_dir:string -> keeper_id:string -> string
+
+type legacy_memory_file =
+  | Legacy_facts
+  | Legacy_events
+
+val supported_legacy_memory_files : legacy_memory_file list
+val legacy_memory_filename : legacy_memory_file -> string
 val current_path_for_legacy_memory_filename :
   keepers_dir:string -> keeper_id:string -> filename:string -> string option
 (** Map a legacy per-keeper Memory OS filename under [.masc/keepers/<keeper>/]
--- a/lib/keeper/keeper_memory_os_io.mli
+++ b/lib/keeper/keeper_memory_os_io.mli
@@ -21,6 +21,11 @@ val list_fact_store_keeper_ids_for_base_path : base_path:string -> string list
 
 val events_path : keeper_id:string -> string
 val events_path_for_keepers_dir : keepers_dir:string -> keeper_id:string -> string
+val current_path_for_legacy_memory_filename :
+  keepers_dir:string -> keeper_id:string -> filename:string -> string option
+(** Map a legacy per-keeper Memory OS filename under [.masc/keepers/<keeper>/]
+    to the current store path under [keepers_dir], when the filename is a
+    supported legacy memory artifact. *)
 val episodes_dir : keeper_id:string -> string
 val tool_results_dir : keeper_id:string -> string
 val tool_result_path : keeper_id:string -> tool_call_id:string -> string
--- a/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
+++ b/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
@@ -22,6 +22,20 @@ type entry =
   ; reason : string
   }
 
+type scan_error =
+  { path : string
+  ; operation : string
+  ; message : string
+  }
+
+type scan_result =
+  { entries : entry list
+  ; truncated : bool
+  ; visited : int
+  ; errors : scan_error list
+  ; exists : bool
+  }
+
 let default_max_depth = 4
 let default_max_entries = 5_000
 
@@ -48,20 +62,7 @@ let split_relative rel =
 
 let has_suffix name suffix = String.ends_with ~suffix name
 
-let is_backup_name name =
-  has_suffix name ".bak"
-  || has_suffix name ".backup"
-  || has_suffix name ".old"
-  || has_suffix name "~"
-;;
-
-let is_orphan_temp_name name =
-  String.equal name "PYEOF"
-  || (String.starts_with ~prefix:".atomic_" name && has_suffix name ".tmp")
-  || has_suffix name ".tmp"
-;;
-
-let live_top_level_dirs =
+let live_runtime_store_dirs =
   [ "tool_usage"
   ; "runtime-manifests"
   ; "metrics"
@@ -72,48 +73,54 @@ let live_top_level_dirs =
   ]
 ;;
 
-let live_keeper_child_dirs =
-  [ "metrics"
-  ; "execution-receipts"
-  ; "turn-records"
-  ; "reaction-ledger"
-  ; "runtime-manifests"
-  ; "trajectories"
-  ; "tool_usage"
-  ]
+let member name names = List.exists (String.equal name) names
+
+let regular_path_exists_no_follow path =
+  try
+    match (Unix.lstat path).Unix.st_kind with
+    | Unix.S_REG -> true
+    | Unix.S_DIR | Unix.S_LNK | Unix.S_CHR | Unix.S_BLK | Unix.S_FIFO | Unix.S_SOCK ->
+      false
+  with
+  | Unix.Unix_error _ -> false
 ;;
 
-let memory_os_filenames = [ "facts.jsonl"; "events.jsonl"; "episodes.jsonl" ]
-
-let member name names = List.exists (String.equal name) names
+let migrated_memory_path_for_legacy_filename ~current_keepers_dir ~keeper_id filename =
+  if String.equal filename "facts.jsonl"
+  then
+    Some
+      (Keeper_memory_os_io.facts_path_for_keepers_dir
+         ~keepers_dir:current_keepers_dir
+         ~keeper_id)
+  else if String.equal filename "events.jsonl"
+  then
+    Some
+      (Keeper_memory_os_io.events_path_for_keepers_dir
+         ~keepers_dir:current_keepers_dir
+         ~keeper_id)
+  else None
+;;
 
 let migrated_memory_path_exists ~current_keepers_dir parts =
   match parts with
-  | keeper_id :: filename :: [] when member filename memory_os_filenames ->
-    Sys.file_exists (Filename.concat (Filename.concat current_keepers_dir keeper_id) filename)
+  | keeper_id :: filename :: [] ->
+    (match migrated_memory_path_for_legacy_filename ~current_keepers_dir ~keeper_id filename with
+     | Some path -> regular_path_exists_no_follow path
+     | None -> false)
   | _ -> false
 ;;
 
 let classify ~current_keepers_dir ~rel ~kind =
   let parts = split_relative rel in
-  let name =
-    match List.rev parts with
-    | [] -> rel
-    | hd :: _ -> hd
-  in
-  if is_orphan_temp_name name
-  then Orphaned, "orphaned_temp_or_marker"
-  else if is_backup_name name
-  then Backup, "backup_suffix"
-  else if migrated_memory_path_exists ~current_keepers_dir parts
+  if migrated_memory_path_exists ~current_keepers_dir parts
   then Migrated, "memory_os_file_already_present_under_config_keepers"
   else (
     match parts with
     | [ top ] when String.equal kind "file" && has_suffix top ".json" ->
       Live, "legacy_keeper_meta_json"
-    | top :: _ when member top live_top_level_dirs ->
+    | top :: _ when member top live_runtime_store_dirs ->
       Live, "known_top_level_runtime_store"
-    | _keeper :: child :: _ when member child live_keeper_child_dirs ->
+    | _keeper :: child :: _ when member child live_runtime_store_dirs ->
       Live, "known_keeper_runtime_store"
     | _ -> Unknown, "unclassified_legacy_path")
 ;;
@@ -129,27 +136,54 @@ let stat_kind st =
   | Unix.S_SOCK -> "socket"
 ;;
 
-let safe_lstat path =
-  try Some (Unix.lstat path) with
-  | Unix.Unix_error _ -> None
+let error_message = function
+  | Unix.Unix_error (err, _, _) -> Unix.error_message err
+  | Sys_error msg -> msg
+  | exn -> Printexc.to_string exn
 ;;
 
-let sorted_readdir path =
-  try Sys.readdir path |> Array.to_list |> List.sort String.compare with
-  | Sys_error _ -> []
+let display_path ~root path =
+  if String.equal path root then "." else relative_path ~root path
+;;
+
+let record_error errors ~root ~operation path exn =
+  let error =
+    { path = display_path ~root path; operation; message = error_message exn }
+  in
+  errors := error :: !errors;
+  Log.Dashboard.warn
+    "legacy keeper inventory %s failed for %s: %s"
+    operation
+    error.path
+    error.message
+;;
+
+let lstat_result errors ~root path =
+  try Ok (Unix.lstat path) with
+  | Unix.Unix_error _ as exn ->
+    record_error errors ~root ~operation:"lstat" path exn;
+    Error ()
+;;
+
+let sorted_readdir_result errors ~root path =
+  try Ok (Sys.readdir path |> Array.to_list |> List.sort String.compare) with
+  | Sys_error _ as exn ->
+    record_error errors ~root ~operation:"readdir" path exn;
+    Error ()
 ;;
 
 let scan_entries ~legacy_dir ~current_keepers_dir ~max_depth ~max_entries =
   let entries = ref [] in
   let visited = ref 0 in
   let truncated = ref false in
+  let errors = ref [] in
   let rec visit ~depth path =
     if !visited >= max_entries
     then truncated := true
     else (
-      match safe_lstat path with
-      | None -> ()
-      | Some st ->
+      match lstat_result errors ~root:legacy_dir path with
+      | Error () -> ()
+      | Ok st ->
         incr visited;
         let kind = stat_kind st in
         let rel = relative_path ~root:legacy_dir path in
@@ -164,18 +198,49 @@ let scan_entries ~legacy_dir ~current_keepers_dir ~max_depth ~max_entries =
            }
            :: !entries;
         if String.equal kind "dir" && depth < max_depth
-        then
-          sorted_readdir path
-          |> List.iter (fun name -> visit ~depth:(depth + 1) (Filename.concat path name)))
+        then (
+          match sorted_readdir_result errors ~root:legacy_dir path with
+          | Error () -> ()
+          | Ok names ->
+            names
+            |> List.iter (fun name -> visit ~depth:(depth + 1) (Filename.concat path name))))
   in
-  if Sys.file_exists legacy_dir
-  then
-    sorted_readdir legacy_dir
-    |> List.iter (fun name -> visit ~depth:0 (Filename.concat legacy_dir name));
-  List.rev !entries, !truncated, !visited
+  let exists =
+    match
+      try Ok (Unix.lstat legacy_dir) with
+      | Unix.Unix_error (Unix.ENOENT, _, _)
+      | Unix.Unix_error (Unix.ENOTDIR, _, _) -> Error `Missing
+      | Unix.Unix_error _ as exn -> Error (`Failure exn)
+    with
+    | Error `Missing -> false
+    | Error (`Failure exn) ->
+      record_error errors ~root:legacy_dir ~operation:"lstat" legacy_dir exn;
+      false
+    | Ok st when st.Unix.st_kind = Unix.S_DIR ->
+      (match sorted_readdir_result errors ~root:legacy_dir legacy_dir with
+       | Error () -> true
+       | Ok names ->
+         names |> List.iter (fun name -> visit ~depth:0 (Filename.concat legacy_dir name));
+         true)
+    | Ok st ->
+      record_error
+        errors
+        ~root:legacy_dir
+        ~operation:"readdir"
+        legacy_dir
+        (Sys_error
+           (Printf.sprintf "expected directory, found %s" (stat_kind st)));
+      true
+  in
+  { entries = List.rev !entries
+  ; truncated = !truncated
+  ; visited = !visited
+  ; errors = List.rev !errors
+  ; exists
+  }
 ;;
 
-let entry_to_json entry =
+let entry_to_json (entry : entry) =
   `Assoc
     [ "path", `String entry.path
     ; "depth", `Int entry.depth
@@ -183,6 +248,14 @@ let entry_to_json entry =
     ; "bytes", `Int entry.bytes
     ; "class", `String (class_to_string entry.classification)
     ; "reason", `String entry.reason
+    ]
+;;
+
+let scan_error_to_json (error : scan_error) =
+  `Assoc
+    [ "path", `String error.path
+    ; "operation", `String error.operation
+    ; "message", `String error.message
     ]
 ;;
 
@@ -202,44 +275,54 @@ let class_totals entries =
     class_to_string cls, `Assoc [ "count", `Int count; "bytes", `Int bytes ])
 ;;
 
-let cleanup_candidate_class = function
-  | Orphaned | Backup -> true
-  | Live | Migrated | Unknown -> false
-;;
-
-let cleanup_plan_json entries =
-  let candidates = List.filter (fun entry -> cleanup_candidate_class entry.classification) entries in
-  let bytes = List.fold_left (fun acc entry -> acc + entry.bytes) 0 candidates in
+let cleanup_plan_json () =
   `Assoc
     [ "delete_allowed", `Bool false
     ; "requires_operator_approval", `Bool true
-    ; "candidate_classes", `List [ `String "orphaned"; `String "backup" ]
-    ; "candidate_count", `Int (List.length candidates)
-    ; "candidate_bytes", `Int bytes
-    ; "candidates", `List (List.map entry_to_json candidates)
+    ; "candidate_policy", `String "disabled_until_owner_verified"
+    ; "candidate_classes", `List []
+    ; "candidate_count", `Int 0
+    ; "candidate_bytes", `Int 0
+    ; "candidates", `List []
     ]
+;;
+
+let redacted_workspace_path ~base_path path =
+  let base_prefix =
+    if String.ends_with ~suffix:"/" base_path then base_path else base_path ^ "/"
+  in
+  if String.equal path base_path
+  then "."
+  else if String.starts_with ~prefix:base_prefix path
+  then
+    String.sub path (String.length base_prefix) (String.length path - String.length base_prefix)
+  else if String.equal (Filename.basename path) Common.keepers_runtime_dirname
+  then Filename.concat "<config>" Common.keepers_runtime_dirname
+  else "<external>"
 ;;
 
 let legacy_keeper_inventory_http_json ~base_path ?(max_depth = default_max_depth)
       ?(max_entries = default_max_entries) () =
-  let masc_root = Config_dir_resolver.masc_root ~base_path in
-  let legacy_dir = Filename.concat masc_root "keepers" in
+  let legacy_dir = Common.keepers_runtime_dir_of_base ~base_path in
   let current_keepers_dir = Config_dir_resolver.keepers_dir_for_base_path ~base_path in
-  let entries, truncated, visited =
+  let result =
     scan_entries ~legacy_dir ~current_keepers_dir ~max_depth ~max_entries
   in
   `Assoc
-    [ "base_path", `String base_path
-    ; "legacy_keepers_path", `String legacy_dir
-    ; "current_config_keepers_path", `String current_keepers_dir
-    ; "exists", `Bool (Sys.file_exists legacy_dir)
+    [ "path_scope", `String "workspace_relative"
+    ; "legacy_keepers_path", `String (redacted_workspace_path ~base_path legacy_dir)
+    ; ( "current_config_keepers_path"
+      , `String (redacted_workspace_path ~base_path current_keepers_dir) )
+    ; "exists", `Bool result.exists
     ; "read_only", `Bool true
     ; "max_depth", `Int max_depth
     ; "max_entries", `Int max_entries
-    ; "visited_entries", `Int visited
-    ; "truncated", `Bool truncated
-    ; "class_totals", `Assoc (class_totals entries)
-    ; "entries", `List (List.map entry_to_json entries)
-    ; "dry_run_cleanup_plan", cleanup_plan_json entries
+    ; "visited_entries", `Int result.visited
+    ; "truncated", `Bool result.truncated
+    ; "scan_complete", `Bool ((not result.truncated) && result.errors = [])
+    ; "scan_errors", `List (List.map scan_error_to_json result.errors)
+    ; "class_totals", `Assoc (class_totals result.entries)
+    ; "entries", `List (List.map entry_to_json result.entries)
+    ; "dry_run_cleanup_plan", cleanup_plan_json ()
     ]
 ;;
--- a/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
+++ b/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
@@ -47,33 +47,41 @@ let class_to_string = function
   | Unknown -> "unknown"
 ;;
 
+let normalize_path path =
+  path
+  |> Env_config_core.normalize_path_lexically
+  |> Env_config_core.strip_path_trailing_slashes
+;;
+
+let path_components_under_root ~root path =
+  let root = normalize_path root in
+  let path = normalize_path path in
+  let rec loop acc current =
+    if String.equal current root
+    then Some acc
+    else (
+      let parent = Filename.dirname current in
+      if String.equal parent current
+      then None
+      else loop (Filename.basename current :: acc) parent)
+  in
+  loop [] path
+;;
+
+let path_of_components = function
+  | [] -> "."
+  | first :: rest -> List.fold_left Filename.concat first rest
+;;
+
 let relative_path ~root path =
-  let root_prefix = if String.ends_with ~suffix:"/" root then root else root ^ "/" in
-  if String.starts_with ~prefix:root_prefix path
-  then String.sub path (String.length root_prefix) (String.length path - String.length root_prefix)
-  else path
+  match path_components_under_root ~root path with
+  | Some parts -> path_of_components parts
+  | None -> path
 ;;
 
-let split_relative rel =
-  rel
-  |> String.split_on_char '/'
-  |> List.filter (fun part -> not (String.equal part ""))
+let is_runtime_store_dir name =
+  Option.is_some (Config_dir_resolver.keeper_runtime_store_of_dirname name)
 ;;
-
-let has_suffix name suffix = String.ends_with ~suffix name
-
-let live_runtime_store_dirs =
-  [ "tool_usage"
-  ; "runtime-manifests"
-  ; "metrics"
-  ; "execution-receipts"
-  ; "turn-records"
-  ; "reaction-ledger"
-  ; "trajectories"
-  ]
-;;
-
-let member name names = List.exists (String.equal name) names
 
 let regular_path_exists_no_follow path =
   try
@@ -110,17 +118,16 @@ let migrated_memory_path_exists ~current_keepers_dir parts =
   | _ -> false
 ;;
 
-let classify ~current_keepers_dir ~rel ~kind =
-  let parts = split_relative rel in
+let classify ~current_keepers_dir ~parts ~kind =
   if migrated_memory_path_exists ~current_keepers_dir parts
   then Migrated, "memory_os_file_already_present_under_config_keepers"
   else (
     match parts with
-    | [ top ] when String.equal kind "file" && has_suffix top ".json" ->
+    | [ top ] when String.equal kind "file" && Keeper_meta_store.is_keeper_meta_file top ->
       Live, "legacy_keeper_meta_json"
-    | top :: _ when member top live_runtime_store_dirs ->
+    | top :: _ when is_runtime_store_dir top ->
       Live, "known_top_level_runtime_store"
-    | _keeper :: child :: _ when member child live_runtime_store_dirs ->
+    | _keeper :: child :: _ when is_runtime_store_dir child ->
       Live, "known_keeper_runtime_store"
     | _ -> Unknown, "unclassified_legacy_path")
 ;;
@@ -139,7 +146,9 @@ let stat_kind st =
 let error_message = function
   | Unix.Unix_error (err, _, _) -> Unix.error_message err
   | Sys_error msg -> msg
-  | exn -> Printexc.to_string exn
+  | Invalid_argument _ -> "invalid_argument"
+  | Failure _ -> "failure"
+  | _ -> "unexpected_exception"
 ;;
 
 let display_path ~root path =
@@ -186,8 +195,12 @@ let scan_entries ~legacy_dir ~current_keepers_dir ~max_depth ~max_entries =
       | Ok st ->
         incr visited;
         let kind = stat_kind st in
-        let rel = relative_path ~root:legacy_dir path in
-        let classification, reason = classify ~current_keepers_dir ~rel ~kind in
+        let parts =
+          path_components_under_root ~root:legacy_dir path
+          |> Option.value ~default:[ path ]
+        in
+        let rel = path_of_components parts in
+        let classification, reason = classify ~current_keepers_dir ~parts ~kind in
         entries
         := { path = rel
            ; depth
@@ -288,17 +301,12 @@ let cleanup_plan_json () =
 ;;
 
 let redacted_workspace_path ~base_path path =
-  let base_prefix =
-    if String.ends_with ~suffix:"/" base_path then base_path else base_path ^ "/"
-  in
-  if String.equal path base_path
-  then "."
-  else if String.starts_with ~prefix:base_prefix path
-  then
-    String.sub path (String.length base_prefix) (String.length path - String.length base_prefix)
-  else if String.equal (Filename.basename path) Common.keepers_runtime_dirname
-  then Filename.concat "<config>" Common.keepers_runtime_dirname
-  else "<external>"
+  match path_components_under_root ~root:base_path path with
+  | Some parts -> path_of_components parts
+  | None ->
+    if String.equal (Filename.basename path) Common.keepers_runtime_dirname
+    then Filename.concat "<config>" Common.keepers_runtime_dirname
+    else "<external>"
 ;;
 
 let legacy_keeper_inventory_http_json ~base_path ?(max_depth = default_max_depth)
--- a/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
+++ b/lib/server/server_dashboard_http_legacy_keeper_inventory.ml
@@ -167,6 +167,15 @@ let record_error errors ~root ~operation path exn =
     error.message
 ;;
 
+let record_scan_error errors ~operation ~path ~message =
+  errors := { path; operation; message } :: !errors;
+  Log.Dashboard.warn
+    "legacy keeper inventory %s failed for %s: %s"
+    operation
+    path
+    message
+;;
+
 let lstat_result errors ~root path =
   try Ok (Unix.lstat path) with
   | Unix.Unix_error _ as exn ->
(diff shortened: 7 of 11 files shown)


jeong-sik:
[Adv Review] masc#22437 — legacy keeper inventory dashboard

Head SHA reviewed: `505f34a90279327c77360e83cbb47088d0c5f0e5`

This PR adds a read-only inventory helper for the legacy keeper store and exposes it on a dashboard route. While the destructive impact is intentionally gated by `delete_allowed:false` / `requires_operator_approval:true`, the classification engine is almost entirely hard-coded filename/directory heuristics that duplicate keeper-layout knowledge, swallow filesystem errors, and can mislabel live data as a cleanup candidate. I do not think this should merge without addressing the P0/P1 items below.

---

## P0 — Cleanup heuristics can misclassify live keeper data as deletion candidates

The dry-run cleanup plan selects candidates by filename suffix/string matching only. Any live keeper file that happens to match these patterns becomes a proposed deletion:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:47-57 *)
let is_backup_name name =
  has_suffix name ".bak"
  || has_suffix name ".backup"
  || has_suffix name ".old"
  || has_suffix name "~"
;;

let is_orphan_temp_name name =
  String.equal name "PYEOF"
  || (String.starts_with ~prefix:".atomic_" name && has_suffix name ".tmp")
  || has_suffix name ".tmp"
;;
```

`is_orphan_temp_name` is especially brittle: a literal `PYEOF` string and any `*.tmp` file anywhere under `.masc/keepers` is proposed for cleanup. If a keeper runtime legitimately writes `.tmp` residue that is still referenced by an in-flight process, the dashboard will present it as safe-to-delete. Because the stated next step is operator approval, this is a data-loss accident waiting to happen.

**Fix:** Do not classify files as cleanup candidates from filename heuristics alone. Consult the keeper runtime / migration registry to determine whether a path is actually orphaned. If heuristics must remain, gate them behind a conservative allow-list with documented ownership checks, and never mark `Unknown` paths as candidates.

---

## P0 — SSOT violation: hard-coded keeper directory/filename constants

The module embeds keeper layout constants that almost certainly already exist as the single source of truth elsewhere:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:60-81 *)
let live_top_level_dirs =
  [ "tool_usage"
  ; "runtime-manifests"
  ; "metrics"
  ; "execution-receipts"
  ; "turn-records"
  ; "reaction-ledger"
  ; "trajectories"
  ]
;;

let live_keeper_child_dirs =
  [ "metrics"
  ; "execution-receipts"
  ; "turn-records"
  ; "reaction-ledger"
  ; "runtime-manifests"
  ; "trajectories"
  ; "tool_usage"
  ]
;;

let memory_os_filenames = [ "facts.jsonl"; "events.jsonl"; "episodes.jsonl" ]
```

`live_top_level_dirs` and `live_keeper_child_dirs` are the same set in a different order — that is duplicated logic, not two different concepts. `memory_os_filenames` duplicates Memory OS fact-store naming. The legacy path segment `"keepers"` is also hard-coded (`Filename.concat masc_root "keepers"`) while the current path is resolved through `Config_dir_resolver`.

**Fix:** Drive all layout knowledge from the keeper / config resolver modules. If no resolver exposes these names, add it there rather than inlining a second copy.

---

## P1 — Silent failure on unreadable paths

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:142-150 *)
let safe_lstat path =
  try Some (Unix.lstat path) with
  | Unix.Unix_error _ -> None
;;

let sorted_readdir path =
  try Sys.readdir path |> Array.to_list |> List.sort String.compare with
  | Sys_error _ -> []
;;
```

Permission denied, broken symlinks, and other `Unix_error`s are swallowed without logging. The inventory silently omits those paths, so an operator may see a "clean" legacy directory while files actually exist but are inaccessible to the server process. `migrated_memory_path_exists` also uses `Sys.file_exists`, which follows symlinks and can misclassify a symlink to a current keeper file as "migrated".

**Fix:** Distinguish "does not exist" from "error accessing". At minimum log `Unix_error` details with the path; better, surface a per-path error status in the JSON response so the caller knows the scan was incomplete.

---

## P1 — Blocking Unix I/O inside the domain pool, not Eio-native file I/O

```ocaml
(* lib/server/server_routes_http_routes_dashboard.ml:701-713 *)
|> Http.Router.get "/api/v1/dashboard/legacy-keeper-inventory" (fun request reqd ->
     with_public_read (fun state req reqd ->
       let base_path = (Mcp_server.workspace_config state).base_path in
       let cache_key = Printf.sprintf "legacy_keeper_inventory:%s" base_path in
       let json =
         Dashboard_cache.get_or_compute cache_key ~ttl:standard_cache_ttl_s (fun () ->
           Domain_pool_ref.submit_io_or_inline (fun () ->
             Server_dashboard_http_legacy_keeper_inventory.legacy_keeper_inventory_http_json
               ~base_path
               ()))
       in
       Http.Response.json_value ~compress:true ~request:req json reqd
     ) request reqd)
```

The scan uses `Unix.lstat`, `Sys.readdir`, `Sys.file_exists`, and `Out_channel` in tests — all blocking POSIX calls. Even though `Domain_pool_ref.submit_io_or_inline` offloads the work, MASC on OCaml 5.4 / Eio should prefer Eio file abstractions for blocking FS operations. Blocking a domain-pool worker for a recursive walk of up to 5,000 entries can hurt tail latency for other IO jobs submitted to the same pool.

**Fix:** Use Eio file-system primitives (or an explicit IO executor with bounded concurrency) and yield between directory reads; do not perform a blocking depth-first walk on a domain pool thread.

---

## P1 — Public-read route exposes internal keeper paths and sizes

The route is registered under `with_public_read`, yet it enumerates the local filesystem layout under `.masc/keepers` (`base_path`, relative paths, file sizes, keeper IDs, and memory filenames). This is an information-disclosure risk: it lets any read-authorized caller map the server's private keeper store structure.

**Fix:** Restrict this to an admin/operator read capability, or redact sensitive path components (keeper IDs, filenames) from the public response. At minimum, do not expose absolute `base_path`.

---

## P1 — Duplicated constants and dead logic inside the classifier

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:53-57 *)
let is_orphan_temp_name name =
  String.equal name "PYEOF"
  || (String.starts_with ~prefix:".atomic_" name && has_suffix name ".tmp")
  || has_suffix name ".tmp"
;;
```

The `.atomic_*.tmp` clause is a strict subset of the final `*.tmp` clause and therefore adds no behavior. It appears to be leftover from an earlier allow-list.

Also, `live_top_level_dirs` and `live_keeper_child_dirs` contain the same directory names. The classifier should express the layout as one canonical set plus a depth rule, not two near-identical lists that can drift.

**Fix:** Remove the redundant `.atomic_*.tmp` conjunct (or document why it is called out explicitly) and collapse the two live-dir lists into a single SSOT constant.

---

## P2 — Path handling is platform-naive and likely duplicates existing utilities

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:35-43 *)
let relative_path ~root path =
  let root_prefix = if String.ends_with ~suffix:"/" root then root else root ^ "/" in
  if String.starts_with ~prefix:root_prefix path
  then String.sub path (String.length root_prefix) (String.length path - String.length root_prefix)
  else path
;;

let split_relative rel =
  rel
  |> String.split_on_char '/'
  |> List.filter (fun part -> not (String.equal part ""))
;;
```

This hard-codes `/` and reimplements prefix stripping. MASC already has path helpers (e.g., `Fpath` wrappers or `Filename_util`); using them would avoid edge cases (e.g., root `/foo` vs path `/foobar/x`).

**Fix:** Use the project's path SSOT utilities and `Filename.dir_sep` / `Fpath` rather than hand-rolled string surgery.

---

## P2 — Missing operator/config controls for scan limits

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:25-26 *)
let default_max_depth = 4
let default_max_entries = 5_000
```

These defaults are baked in and the HTTP handler does not accept query parameters to override them. On a large or deeply-nested legacy store the scan will truncate and the operator has no way to request a deeper/wider view from the API.

**Fix:** Accept `max_depth` and `max_entries` as optional query parameters with sane ceilings, or expose them as config knobs.

---

## P2 — Tests cover only happy path

`test_server_dashboard_http_legacy_keeper_inventory.ml` validates basic classification and truncation, but does not exercise:

- permission-denied / unreadable directories (`safe_lstat` and `sorted_readdir` error branches),
- symlinks (including symlinks from legacy into current keepers),
- the public-route authorization path,
- cross-platform path behavior,
- misclassification edge cases (e.g., a legitimate live file named `data.tmp`).

**Fix:** Add adversarial test cases for the error branches and the heuristic boundary conditions above.

---

## CI status (as of fetch)

- `ocamlformat-check` — **IN_PROGRESS**
- `Meta Guards` — **IN_PROGRESS**
- `Health` — **IN_PROGRESS**
- `Build and Test` — **IN_PROGRESS**
- `Lint` — **IN_PROGRESS**
- All other reported checks — SUCCESS or SKIPPED
- `mergeStateStatus`: `BLOCKED` (required checks still pending)

Do not merge until the in-progress checks finish. If `Build and Test` / `Lint` / `ocamlformat-check` report failures, they must be green first.

---

## Summary

Blockers before merge:
1. **P0:** Replace or harden the filename-heuristic cleanup classification so live keeper data cannot become a deletion candidate.
2. **P0:** Remove duplicated keeper-layout constants and drive classification from the existing keeper/config SSOT.
3. **P1:** Stop swallowing `Unix_error`/`Sys_error`; log and/or surface access failures.
4. **P1:** Move blocking FS work off the domain pool to Eio-native IO or a dedicated IO executor.
5. **P1:** Restrict the public-read route or redact sensitive internal paths.

P2 items (path utilities, query knobs, expanded tests) can be addressed in the same PR or fast-follow.

jeong-sik:
Applied B5 guard follow-up commit 5951fbe50751.

Validation:

jeong-sik:
[Adv Review] masc#22437 — follow-up: head changed + B5 guard addressed, legacy inventory blockers still open

Head SHA reviewed: `5951fbe50751b13c0edbdbf90747d452a9f2a45f`
Prior review head: `505f34a90279327c77360e83cbb47088d0c5f0e5`

The author reply (commit `5951fbe`) only changed `test/test_masc_oas_bridge_cancel_boundary.ml` to assert the shared `Cancel_wall_bucket.of_wall` SSOT. The legacy keeper inventory code added by this PR is unchanged, so all P0/P1 blockers from the first review remain. Summary below.

---

## Acknowledged fix

`test/test_masc_oas_bridge_cancel_boundary.ml` now asserts the SSOT call instead of duplicating bucket labels:

```ocaml
(* test/test_masc_oas_bridge_cancel_boundary.ml:117-126 *)
  (* B5: Time bucket classification uses the shared SSOT.  Literal bucket
     labels are pinned in [test_cancel_wall_bucket.ml]. *)
  assert_contains
    ~label:"B5: cancel bucket classifier SSOT"
    src
    "Cancel_wall_bucket.of_wall wall";
```

This aligns the bridge guard with `lib/masc_oas_bridge.ml:100` (`let bucket = Cancel_wall_bucket.of_wall wall in`) and the boundary/label drift-guard in `test/test_cancel_wall_bucket.ml`. Good.

**Caveat:** `assert_contains` is still string-matching source text, so it is a weak guard — it will pass if the call appears anywhere, even in a comment or dead branch. The real SSOT enforcement is `test_cancel_wall_bucket.ml`. That is acceptable if this guard is treated as advisory, but do not rely on it as a primary ratchet.

---

## P0 — Cleanup heuristics still misclassify live keeper data as deletion candidates

Unchanged in the new file:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:57-68 *)
let is_backup_name name =
  has_suffix name ".bak"
  || has_suffix name ".backup"
  || has_suffix name ".old"
  || has_suffix name "~"
;;

let is_orphan_temp_name name =
  String.equal name "PYEOF"
  || (String.starts_with ~prefix:".atomic_" name && has_suffix name ".tmp")
  || has_suffix name ".tmp"
;;
```

Any file ending in `.tmp`, `.bak`, `.backup`, `.old`, or `~` under `.masc/keepers` becomes a dry-run deletion candidate. The `.atomic_*.tmp` clause is a strict subset of the final `*.tmp` clause and adds no behavior (dead logic). Because the plan is presented to an operator for approval, filename heuristics alone are a data-loss risk.

**Still required:** Do not classify paths as cleanup candidates from filename patterns. Consult the keeper runtime / migration registry, or at minimum move these heuristics behind an explicit allow-list with ownership checks.

---

## P0 — Keeper layout constants are still hard-coded and duplicated

Unchanged:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:70-93 *)
let live_top_level_dirs =
  [ "tool_usage"
  ; "runtime-manifests"
  ; "metrics"
  ; "execution-receipts"
  ; "turn-records"
  ; "reaction-ledger"
  ; "trajectories"
  ]
;;

let live_keeper_child_dirs =
  [ "metrics"
  ; "execution-receipts"
  ; "turn-records"
  ; "reaction-ledger"
  ; "runtime-manifests"
  ; "trajectories"
  ; "tool_usage"
  ]
;;

let memory_os_filenames = [ "facts.jsonl"; "events.jsonl"; "episodes.jsonl" ]
```

`live_top_level_dirs` and `live_keeper_child_dirs` are the same set in a different order — duplicated logic, not two concepts. `memory_os_filenames` duplicates Memory OS fact-store naming. The legacy segment `"keepers"` is also hard-coded:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:231-233 *)
let legacy_dir = Filename.concat masc_root "keepers" in
let current_keepers_dir = Config_dir_resolver.keepers_dir_for_base_path ~base_path in
```

While current keepers go through `Config_dir_resolver`, the legacy path does not.

**Still required:** Drive all layout knowledge from the keeper / config resolver modules. If no resolver exposes these names, extend the resolver rather than inlining a second copy.

---

## P1 — Filesystem errors are still swallowed

Unchanged:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:138-146 *)
let safe_lstat path =
  try Some (Unix.lstat path) with
  | Unix.Unix_error _ -> None
;;

let sorted_readdir path =
  try Sys.readdir path |> Array.to_list |> List.sort String.compare with
  | Sys_error _ -> []
;;
```

Permission denied, broken symlinks, and other `Unix_error`s are silently dropped. `migrated_memory_path_exists` still uses `Sys.file_exists`, which follows symlinks and can misclassify a symlink to a current keeper file as "migrated".

**Still required:** Distinguish "does not exist" from "access error". Log or surface per-path error status in the JSON response so operators know the scan is incomplete.

---

## P1 — Blocking Unix I/O is still performed on the domain pool

Unchanged route handler:

```ocaml
(* lib/server/server_routes_http_routes_dashboard.ml:278-290 *)
|> Http.Router.get "/api/v1/dashboard/legacy-keeper-inventory" (fun request reqd ->
     with_public_read (fun state req reqd ->
       let base_path = (Mcp_server.workspace_config state).base_path in
       let cache_key = Printf.sprintf "legacy_keeper_inventory:%s" base_path in
       let json =
         Dashboard_cache.get_or_compute cache_key ~ttl:standard_cache_ttl_s (fun () ->
           Domain_pool_ref.submit_io_or_inline (fun () ->
             Server_dashboard_http_legacy_keeper_inventory.legacy_keeper_inventory_http_json
               ~base_path
               ()))
       in
       Http.Response.json_value ~compress:true ~request:req json reqd
     ) request reqd)
```

The scan uses `Unix.lstat`, `Sys.readdir`, `Sys.file_exists`, and (in tests) `Out_channel`. Offloading blocking POSIX calls to `Domain_pool_ref.submit_io_or_inline` still consumes a domain-pool worker for a recursive walk of up to 5,000 entries and can stall other IO jobs.

**Still required:** Use Eio-native file-system primitives, or an explicit IO executor with bounded concurrency and yielding between directory reads.

---

## P1 — Public-read route still exposes internal keeper layout

The route is still registered under `with_public_read` and returns `base_path`, `legacy_keepers_path`, `current_config_keepers_path`, keeper IDs, memory filenames, and file sizes. This is an information-disclosure risk.

**Still required:** Restrict to an admin/operator read capability, or redact sensitive path components and do not expose absolute `base_path`.

---

## P2 — Unresolved from first review

- **Path handling is still platform-naive** (`relative_path` hard-codes `/`, reimplements prefix stripping). Use the project's path SSOT (`Fpath` / `Filename_util`) instead of `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:42-53`.
- **Scan limits are still baked-in constants** (`default_max_depth = 4`, `default_max_entries = 5_000`) with no query/config override.
- **Tests still cover only the happy path** (`test/test_server_dashboard_http_legacy_keeper_inventory.ml`). No coverage for permission denied, symlinks, public-route authorization, cross-platform paths, or misclassification edges such as a legitimate live file named `data.tmp`.

---

## Scope / cross-PR note

The B5 guard fix and the legacy keeper inventory dashboard are unrelated concerns. Landing them in the same PR increases review surface and couples the inventory dashboard's blockers to a clean SSOT cleanup. Consider splitting the B5 change out so this PR can be focused on the inventory work.

---

## CI status (latest fetch)

- `Build and Test` — **IN_PROGRESS**
- `Lint` — **SUCCESS**
- `ocamlformat-check` — **SUCCESS**
- `Meta Guards` — **SUCCESS**
- `Health` — **SUCCESS**
- All other reported checks — SUCCESS or SKIPPED
- `mergeStateStatus`: `BLOCKED` (required checks still pending)

Do not merge until `Build and Test` finishes. Even after it is green, the P0/P1 blockers above must be resolved first.

---

## Blockers before merge

1. **P0** Replace or harden filename-heuristic cleanup classification so live keeper data cannot become a deletion candidate.
2. **P0** Remove duplicated keeper-layout constants and drive classification from the existing keeper/config SSOT.
3. **P1** Stop swallowing `Unix_error`/`Sys_error`; log and/or surface access failures.
4. **P1** Move blocking FS work off the domain pool to Eio-native IO or a dedicated IO executor.
5. **P1** Restrict the public-read route or redact sensitive internal paths.

P2 items (path utilities, query knobs, expanded tests) can be addressed in the same PR or fast-follow.

jeong-sik:
[Adv Review] masc#22437 — head changed to `27d0aad4cc64`, inventory hardening partially landed

`fix: harden legacy keeper inventory scan` (27d0aad4cc64) 이후 상당 부분 개선되었습니다.

정확히 고쳐진 것:
- cleanup heuristics (`is_backup_name`, `is_orphan_temp_name`) 삭제 및 dry-run plan disabled 전환 (`cleanup_plan_json`)
- `safe_lstat`/`sorted_readdir` 무silent 실패 제거 → `scan_error` 기록 + `Log.Dashboard.warn`
- `Sys.file_exists` → `regular_path_exists_no_follow` (symlink 미따라가기)
- route 권한 `with_public_read` → `with_permission_auth ~permission:Masc_domain.CanAdmin`
- scan offload를 `Domain_pool_ref`에서 `Executor_pool_ref.submit_or_inline` (Eio executor pool) 로 변경
- `max_depth` / `max_entries` query knob 추가

그러나 아래 blockers는 여전히 남아 있습니다.

**P0 — keeper layout 상수 hardcoding** (`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:71-80`)
`live_runtime_store_dirs`가 여전히 `"tool_usage"`, `"runtime-manifests"`, `"metrics"` 등을 문자열 리터럴로 inline합니다. 이는 기존 `Keeper_runtime_manifest`, `Keeper_reaction_ledger`, `Config_dir_resolver` 등이 이미 알고 있는 layout knowledge를 중복하는 SSOT 위반입니다. 단순히 두 개의 중복 리스트를 하나로 합친 것으로는 P0가 해결되지 않습니다. 이 directory set을 keeper/config resolver에서 유도하거나, resolver에 추가해야 합니다.

**P1 — classification이 여전히 string/suffix heuristics** (`ml:119-132`)
`classify`에서 top-level `.json` suffix match와 `member child live_runtime_store_dirs` directory-name match로 Live를 판정합니다. 이는 위 P0의 hardcoded list에 직접 의존하며, keeper runtime이 실제로 사용하는 layout과 dashboard 분류가 drift할 여지를 남깁니다.

**P1 — path handling이 platform-naive string manipulation** (`ml:56-61`, `ml:296-308`)
`relative_path`와 `redacted_workspace_path`가 `/` 구분자 기준으로 prefix strip을 재구현합니다. 프로젝트 path SSOT(`Fpath`/`Filename_util`)를 사용하지 않으면 Windows/비정규화 path 등에서 오작동할 수 있습니다.

**P2 — `error_message` fallback** (`ml:148`)
`Printexc.to_string exn` fallback이 남아있습니다. 로깅용이라도 RFC-0148 취지상 정당화가 명시되어야 합니다.

위 P0/P1 해소 전까지는 merge 권하지 않습니다. P2는 fast-follow 가능합니다.

jeong-sik:
Addressed the remaining legacy keeper inventory blockers in `6fc223dcf99`.

Changes:
- Moved the canonical keeper runtime child-store set to `Common.keeper_runtime_store` and re-exported lookup through `Config_dir_resolver.keeper_runtime_store_of_dirname`; the inventory module no longer owns a duplicated `tool_usage` / `runtime-manifests` / `metrics` / etc. layout list.
- Replaced top-level `.json` suffix classification with `Keeper_meta_store.is_keeper_meta_file`, so keeper meta JSON follows the keeper meta SSOT and sidecars such as `*.dataset.json` stay `unknown`.
- Replaced the dashboard-local `/` prefix/split helpers with a component walk based on `Env_config_core.normalize_path_lexically` plus `Filename.dirname` / `Filename.basename` / `Filename.concat` for both entry paths and redacted paths.
- Removed the `Printexc.to_string` fallback from inventory scan errors; unknown exceptions now get bounded labels.
- Added regression coverage for exact keeper meta JSON vs sidecar JSON classification.

Validation, source-level only per current instruction:
- `ocamlformat --check lib/core/common.ml lib/core/common.mli lib/config_dir_resolver/config_dir_resolver.ml lib/config_dir_resolver/config_dir_resolver.mli lib/server/server_dashboard_http_legacy_keeper_inventory.ml lib/server/server_dashboard_http_legacy_keeper_inventory.mli test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing lib/core/common.ml lib/core/common.mli lib/config_dir_resolver/config_dir_resolver.ml lib/config_dir_resolver/config_dir_resolver.mli lib/server/server_dashboard_http_legacy_keeper_inventory.ml lib/server/server_dashboard_http_legacy_keeper_inventory.mli test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `git diff --check`
- `rg -n "live_runtime_store_dirs|has_suffix|split_relative|String\\.starts_with|String\\.ends_with|Printexc\\.to_string|\\"tool_usage\\"|\\"runtime-manifests\\"|\\"execution-receipts\\"|\\"turn-records\\"|\\"reaction-ledger\\"|\\"trajectories\\"" lib/server/server_dashboard_http_legacy_keeper_inventory.ml` returned no hits.

Not inspected or run here: Dune build/test, CI/check logs.

jeong-sik:
[Adv Review] masc#22437 — author-reply re-review (head `6fc223d`)

이전 blocker 대부분이 잘 수정되었습니다:

- cleanup heuristics (`is_backup_name`/`is_orphan_temp_name`) 제거 및 dry-run plan disabled 전환
- `scan_error` 기록 + `Log.Dashboard.warn`로 무silent 실패 제거
- `Sys.file_exists` → `regular_path_exists_no_follow`
- route 권한 `with_public_read` → `with_permission_auth ~permission:Masc_domain.CanAdmin`
- scan offload를 `Domain_pool_ref` → `Executor_pool_ref.submit_or_inline`로 변경
- `max_depth`/`max_entries` query knob 추가
- keeper runtime store 상수를 `Common.keeper_runtime_store`로 이동 후 `Config_dir_resolver.keeper_runtime_store_of_dirname` 노출
- top-level `.json` 분류를 `Keeper_meta_store.is_keeper_meta_file` SSOT로 대체
- path prefix strip 재구현을 `Env_config_core.normalize_path_lexically` + `Filename.dirname/basename` walk로 교체
- `Printexc.to_string` fallback 제거

그러나 아래 항목은 여전히 남아 있어 merge 전에 해결/기록이 필요합니다.

**P1 — legacy memory OS filename string matching (SSOT 위반)**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:96-109`의 `migrated_memory_path_for_legacy_filename`이 여전히 `"facts.jsonl"`, `"events.jsonl"`을 문자열 리터럴로 매칭합니다. 이 이름들은 `Keeper_memory_os_io`에 이미 `<keeper_id>.facts.jsonl`/`<keeper_id>.events.jsonl` 형태로 SSOT화되어 있고, legacy raw filename 매핑 역시 migration/legacy SSOT에서 유도해야 합니다.

**P1 — dry-run cleanup plan is a hard-coded stub**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:291-301`의 `cleanup_plan_json`은 항상 동일한 disabled placeholder JSON을 반환합니다. classification 결과와 무관하게 `candidate_count=0`, `candidate_classes=[]`, `candidates=[]`이므로 실제 dry-run plan이 아닌 stub입니다. read-only라면 해당 필드를 제거하거나, 실제 classification 기반 candidate 산출 로직을 추가해야 합니다.

**P2 — `regular_path_exists_no_follow` swallows access errors**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:86-94`는 `Unix.Unix_error _`를 모두 `false`로 반환해 permission denied와 missing을 구분하지 못하고, `migrated` 분류가 잘못될 수 있습니다. 오류를 기록하거나 호출자에게 전달해야 합니다.

**P2 — `error_message` catch-all loses diagnostics**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:146-152`의 `_ -> "unexpected_exception"`은 예상 밖 예외의 구체 정보를 버립니다. RFC-0148 정신상 `Printexc.to_string` 대신 bounded label을 쓰더라도 예외 종류/백트레이스를 로그에 남기는 정책이 명시되어야 합니다.

**P2 — mutable accumulators**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:184-253`의 `entries`/`visited`/`truncated`/`errors`는 `ref`로 구현되어 있습니다. OCaml ^5.4 기준 tail-recursive pure accumulator로 바꿀 수 있으면 immutable하게 변경하세요.

**P2 — cache key embeds raw `base_path`**
`lib/server/server_routes_http_routes_dashboard.ml:720-726`의 cache key에 `base_path` 원문이 들어갑니다. admin 권한이지만 shared cache 환경에서는 workspace path 노출 가능성이 있으니 해시/SSOT redaction을 적용하세요.

**P2 — test coverage still narrow**
`test/test_server_dashboard_http_legacy_keeper_inventory.ml`는 symlinks, permission-denied paths, route authorization, `regular_path_exists_no_follow` error branch, misclassification edge를 커버하지 않습니다. adversarial test를 추가하세요.

**CI snapshot:** `Meta Guards` FAIL, `Build and Test` / `TLA+ Model Checking` IN_PROGRESS, 나머지 SUCCESS. `Build and Test` green + 위 P1 해결 후 merge 권장합니다.

jeong-sik:
Pushed `84bbf049186` after rebasing on current `origin/main` (`53775d4d3b5`) and fixing the Meta Guards DET/NDT failure.

Fix:
- Removed the permissive `Option.value ~default:[ path ]` fallback in `server_dashboard_http_legacy_keeper_inventory`.
- Root-relative path failures are now explicit scan errors (`relative_path`, `<outside-legacy-root>`) and the entry is skipped instead of surfacing a raw absolute-path fallback.

Local non-Dune validation:
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` (DET/NDT gate PASS)
- `ocamlformat --check lib/server/server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing lib/server/server_dashboard_http_legacy_keeper_inventory.ml`

Per current workflow, Dune/build validation is left to CI.

jeong-sik:
Addressed the remaining #22437 review items in `5d95e302843f99de40e375e915390c0e42dfeb65`.

- P1 legacy filename SSOT: moved legacy `facts.jsonl` / `events.jsonl` mapping into `Keeper_memory_os_io.current_path_for_legacy_memory_filename`; the dashboard inventory now calls that helper.
- P1 cleanup plan stub: `dry_run_cleanup_plan` now derives `candidate_classes`, `candidate_count`, `candidate_bytes`, and `candidates` from classified entries while keeping `delete_allowed=false`.
- P2 access errors: `regular_path_exists_no_follow` returns a typed status; current-store lstat failures are recorded as `scan_errors` with operation `lstat_current_memory` and classified as `unknown`.
- P2 catch-all diagnostics: unexpected exceptions now include a bounded exception slot label rather than using `Printexc.to_string`.
- P2 cache key: legacy inventory dashboard cache key now hashes `base_path` with SHA256 and only embeds a short prefix.
- P2 coverage: added an assertion that cleanup candidate classes are surfaced even when no deletion candidates are currently emitted.

Left the mutable accumulator P2 unchanged in this pass because rewriting the bounded scan loop to a pure accumulator is broader than the merge-blocking fixes above.

Source-only validation run:

- `ocamlformat --check ...`
- `ocamlc -stop-after parsing ...`
- `ocamlc -stop-after parsing -intf ...`
- `git diff --check`
- `scripts/audit-path-ssot.sh`
- targeted stale-pattern `rg` check; no stale hits in `server_dashboard_http_legacy_keeper_inventory.ml`

Not run per current instruction: Dune build/test and CI/check-log inspection.

jeong-sik:
## Adversarial review: PR #22437 — fix(memory): surface legacy keeper inventory

**PR:** https://github.com/jeong-sik/masc/pull/22437  
**State:** OPEN, Draft  
**Head:** `fix/legacy-keeper-inventory-20260627` → `main`  
**CI:** ❌ **Meta Guards** fail, ❌ **Build and Test** fail, ❌ **CI Gate** fails. (Other checks pass, including ocamlformat, Lint, Dashboard, and the new unit tests.)

### What the PR does
Adds a read-only dashboard endpoint `GET /api/v1/dashboard/legacy-keeper-inventory` that performs a bounded filesystem scan of the legacy `.masc/keepers` tree, classifies entries, and returns a JSON inventory plus a disabled dry-run cleanup plan. It also adds a `Common.keeper_runtime_store` variant type as SSOT for canonical runtime-store subdirectory names and re-exports a resolver through `Config_dir_resolver`.

### P0 — merge blockers / correctness

1. **Meta Guards deterministic-boundary debt**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:200`  
   ```ocaml
   path_components_under_root ~root:legacy_dir path
   |> Option.value ~default:[ path ]
   ```
   CI gate `DET/NDT` flags this as a deterministic branch on a non-deterministic source (the filesystem scan). This is a hard merge blocker. Either wrap this at the boundary so it cannot return `None` for a scanned path, or add an explicit `(* DET-OK: ... *)` rationale comment.

2. **Declared `Orphaned` / `Backup` classes are never produced**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:116-121` defines `Orphaned` and `Backup`, but `classify` (lines 228-240) only emits `Live`, `Migrated`, and `Unknown`. The PR summary explicitly promises five classes. Tests even assert orphaned/backup counts are zero, confirming the implementation is incomplete. Either implement the heuristics for temp/backup files or remove the unused constructors and update the summary.

3. **`cleanup_plan_json` is a hardcoded stub**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:398-408`  
   Returns constant zeros, an empty candidate list, and a magic-string policy `"disabled_until_owner_verified"`. Since the endpoint is read-only the disabled semantics are correct, but an inventory dry-run plan should still surface the actual candidate entries (e.g., currently `Unknown` paths) with deletion disallowed. As written it is not a useful plan.

### P1 — control-flow / Eio / maintainability

4. **IO-bound scan runs on the CPU executor pool**  
   `lib/server/server_routes_http_routes_dashboard.ml:729`  
   ```ocaml
   Executor_pool_ref.submit_or_inline (fun () -> ...)
   ```
   `Executor_pool_ref` is backed by `Eio.Executor_pool` and intended for CPU work. The scan is blocking filesystem IO (`Unix.lstat`, `Sys.readdir`). The adjacent `keeper-memory-health` route uses `Domain_pool_ref.submit_io_or_inline`; this route should too, or it will pin domain-pool workers on blocking syscalls.

5. **Imperative, mutable scan instead of an immutable fold**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:291-360`  
   Uses four `ref`s (`entries`, `visited`, `truncated`, `errors`) and a recursive `visit` procedure. For a bounded tree traversal this should be a tail-recursive fold over `Safe_ops.list_dir_safe` results, returning an immutable accumulator. The current style is error-prone and contradicts the project’s preference for pure functions.

6. **`error_message` swallows exceptions and catches internal error constructors**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:253-259`  
   Catches `Invalid_argument` and `Failure` (which are bugs, not filesystem errors) and maps the catch-all to the opaque string `"unexpected_exception"`. Log the original exception with `Printexc.to_string` and only normalize `Unix_error` / `Sys_error`.

7. **Reimplemented path relativization**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:163-187`  
   `path_components_under_root`, `path_of_components`, and `relative_path` rebuild functionality that belongs in an SSOT path module (e.g., `Env_config_core` / `Common`). It also mixes slash-normalized paths (`Env_config_core.normalize_path_lexically`) with `Filename.dirname`/`basename`, which is fragile on non-Unix paths.

8. **`class_totals` is O(5·n)**  
   `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:382-396`  
   Folds the entry list once per class. For the allowed 50k entries this is still fast, but a single `List.fold_left` into a map is simpler and clearer.

### P2 — hygiene / scope / polish

9. **Unrelated test change included**  
   `test/test_masc_oas_bridge_cancel_boundary.ml` changes the cancel-wall-bucket assertion. The change itself is correct (uses `Cancel_wall_bucket.of_wall` instead of literal string matching), but it has nothing to do with legacy keeper inventory. Move it to its own PR.

10. **Silent fallback on invalid query params**  
    `lib/server/server_routes_http_routes_dashboard.ml:704-718` via `Server_utils.int_query_param`  
    Non-integer `max_depth` / `max_entries` silently fall back to defaults. For an admin endpoint, a `400 Bad Request` would be safer.

11. **Cache key embeds raw `base_path`**  
    `lib/server/server_routes_http_routes_dashboard.ml:720-726`  
    `Printf.sprintf "legacy_keeper_inventory:%s:%d:%d" base_path ...` uses the absolute workspace path as a cache key. Not a security issue for an in-memory admin cache, but it leaks path structure and could create very large keys.

12. **Magic strings in JSON contract**  
    `"workspace_relative"`, `"disabled_until_owner_verified"`, and the class strings in `class_to_string` / `cleanup_plan_json` are contract values. They should at least be module-level constants with docstrings noting they are part of the HTTP response schema.

13. **Hardcoded legacy filenames**  
    `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:203-217`  
    `"facts.jsonl"` and `"events.jsonl"` are matched as raw strings. `Keeper_memory_os_io` already hardcodes these suffixes internally; expose SSOT constants there and reuse them.

14. **Tests leave temp directories behind**  
    `test/test_server_dashboard_http_legacy_keeper_inventory.ml:5-9`  
    `fresh_dir` creates `/tmp/masc-legacy-keeper-inventory*` directories but never removes them. Add `Fun.protect` / `at_exit` cleanup.

### Cluster context
This PR is part of a 2026-06-27 Memory OS cleanup cluster that also includes:
- #22446 gc dry-run command
- #22447 local recall outcome eval
- #22449 legacy external_state category migration
- #22451 shared promotion gating

Radius is safe: the endpoint is read-only, guarded by `CanAdmin`, bounded, and does not move/delete files. It does not introduce keeper coupling or constraint-hell failure modes.

### Overall verdict
The read-only safety story is good, but the PR is not ready. The Meta Guards failure is a hard blocker, the classification contract does not match the implementation, and the cleanup plan is a stub. Fix the deterministic-boundary annotation, finish `Orphaned`/`Backup` classification (or shrink the API contract), replace the executor-pool dispatch with the IO pool, and convert the scan to an immutable fold. The new unit tests do pass, but they currently encode the incomplete behavior rather than validating the promised five-class inventory.

jeong-sik:
[Adv Review] masc#22437 — head 84bbf049 재검토

최신 push에서 Meta Guards DET/NDT 실패는 잘 고쳐졌습니다. 이전 [Adv Review]의 상당수 항목도 정상적으로 반영되었습니다.

그러나 아래 P1 blockers는 여전히 남아 있습니다.

**P1 — legacy Memory OS filename이 여전히 문자열 리터럴 매칭**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:96-109`
`migrated_memory_path_for_legacy_filename`에서 `"facts.jsonl"`, `"events.jsonl"`을 `String.equal`로 직접 비교합니다. 이 이름은 `Keeper_memory_os_io.facts_path_for_keepers_dir` / `events_path_for_keepers_dir` 등 이미 SSOT가 존재하므로, legacy raw filename 매핑도 해당 SSOT에서 유도해야 합니다.

**P1 — classification이 directory-name string heuristics에 의존**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:82-84,121-132`
`is_runtime_store_dir`이 `Config_dir_resolver.keeper_runtime_store_of_dirname`을 쓰긴 하지만, 결국 directory 이름 문자열 매칭으로만 Live를 판정합니다. keeper runtime이 실제 사용하는 layout과 dashboard 분류가 drift할 여지가 남습니다.

**P1 — `cleanup_plan_json`은 hard-coded stub**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:305-315`
scan 결과와 무관하게 항상 `candidate_count=0`, `candidates=[]`를 반환합니다. 실제 classification 기반 candidate 산출 로직을 추가하거나, read-only라면 해당 필드를 제거하는 편이 낫습니다.

**P2 (fast-follow 가능)**
- `regular_path_exists_no_follow` (`ml:86-94`)가 `Unix.Unix_error _`를 모두 `false`로 삼켜 permission denied와 missing을 구분하지 못합니다.
- `error_message` (`ml:146-152`)의 `_ -> "unexpected_exception"`이 예외 상세 정보를 버립니다.
- `relative_path` (`ml:76-80`)의 fallback이 raw path를 반환해 error log에 절대 경로가 노출될 수 있습니다.
- `scan_entries` (`ml:193-267`)가 `ref` accumulator를 사용합니다.
- cache key (`lib/server/server_routes_http_routes_dashboard.ml:722-724`)에 raw `base_path`가 들어갑니다.
- `test/test_server_dashboard_http_legacy_keeper_inventory.ml`에 symlinks, permission denied, route auth, error branch, misclassification edge 등 adversarial test가 부족합니다.

`Build and Test`가 아직 시작되지 않았으므로 CI 완료 후 위 P1 해소 전까지 merge는 권하지 않습니다.

jeong-sik:
[Adv Review]

Head SHA reviewed: `5d95e302843f99de40e375e915390c0e42dfeb65`
Prior review: https://github.com/jeong-sik/masc/pull/22437#issuecomment-4815838191
Author reply: https://github.com/jeong-sik/masc/pull/22437#issuecomment-4815819102

This head resolves several of the prior blockers, but the P1 stub below is still present and a few P2s remain.

Fixed (from prior review):
- Fixed: keeper runtime-store directory set is now a typed variant (`Common.keeper_runtime_store`) exposed through `Config_dir_resolver.keeper_runtime_store_of_dirname`; classification uses it instead of string literals.
- Fixed: route is now `with_permission_auth ~permission:Masc_domain.CanAdmin` rather than public-read.
- Fixed: scan is offloaded to `Executor_pool_ref.submit_or_inline` instead of the domain pool.
- Fixed: `regular_path_exists_no_follow` now distinguishes `Access_error exn` instead of swallowing all `Unix_error`s.
- Fixed: dashboard cache key no longer embeds raw `base_path`; it uses a SHA256 hash prefix.
- Fixed: `error_message` no longer falls back to `Printexc.to_string`; it uses a bounded `Printexc.exn_slot_name` label.

Remaining blockers:

**P1 — `dry_run_cleanup_plan` is still a stub**
`cleanup_plan_json` now derives counts from classified entries, but `classify` only ever returns `Live`, `Migrated`, or `Unknown`:

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:484 *)
let cleanup_candidate_classes = [ Orphaned; Backup ]
...
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:271-286 *)
let classify ~current_keepers_dir ~record_access_error ~parts ~kind =
  ...
  | Migrated_absent ->
    (match parts with
     | [ top ] when String.equal kind "file" && Keeper_meta_store.is_keeper_meta_file top ->
       Live, "legacy_keeper_meta_json"
     | top :: _ when is_runtime_store_dir top ->
       Live, "known_top_level_runtime_store"
     | _keeper :: child :: _ when is_runtime_store_dir child ->
       Live, "known_keeper_runtime_store"
     | _ -> Unknown, "unclassified_legacy_path")
```

No entry is classified as `Orphaned` or `Backup`, so `candidate_count`, `candidate_bytes`, and `candidates` are always zero. Either implement classification that produces the advertised cleanup candidates, or remove the placeholder fields until the logic exists.

**P2 — legacy Memory OS filename mapping still string-matches raw names**
The mapping is now centralized in `Keeper_memory_os_io.current_path_for_legacy_memory_filename`, which removes the dashboard-level duplication, but the helper still uses `String.equal` on raw strings:

```ocaml
(* lib/keeper/keeper_memory_os_io.ml:110-116 *)
let current_path_for_legacy_memory_filename ~keepers_dir ~keeper_id ~filename =
  if String.equal filename "facts.jsonl"
  then Some (facts_path_for_keepers_dir ~keepers_dir ~keeper_id)
  else if String.equal filename "events.jsonl"
  then Some (events_path_for_keepers_dir ~keepers_dir ~keeper_id)
  else None
```

Replace the raw filename strings with a first-class variant or canonical list so the supported legacy filenames are SSOT and cannot silently drift.

**P2 — `relative_path` fallback can leak absolute paths**

```ocaml
(* lib/server/server_dashboard_http_legacy_keeper_inventory.ml:218-222 *)
let relative_path ~root path =
  match path_components_under_root ~root path with
  | Some parts -> path_of_components parts
  | None -> path
```

If a path escapes the legacy root, the raw (absolute) path is returned. Return a redacted token such as `"<external>"` instead.

**P2 — mutable accumulators**
`scan_entries` still uses `ref` accumulators for `entries`, `visited`, `truncated`, and `errors` (`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:360-365`). Rewrite the bounded scan loop with tail-recursive pure accumulators.

**P2 — blocking POSIX calls inside the executor pool**
The scan still issues blocking `Unix.lstat`/`Sys.readdir` calls inside `Executor_pool_ref.submit_or_inline` (`lib/server/server_routes_http_routes_dashboard.ml:597-603`). Confirm the pool is a true OS-thread IO executor and not an Eio domain/fiber pool; otherwise blocking syscalls will stall other IO work. Prefer Eio-native file-system primitives.

**P2 — test coverage remains narrow**
`test/test_server_dashboard_http_legacy_keeper_inventory.ml` still covers only the happy path plus a single invalid-root case. Add adversarial cases: symlinks that must not be followed, permission-denied current-store entries, misclassification edges such as `data.tmp` inside a runtime store, and route authorization.

CI note: `Health` is currently FAILED; `Build and Test`, `Lint`, `Meta Guards`, `Dashboard`, and `TLA+ Model Checking` are still IN_PROGRESS. Do not merge until required checks are green and the P1 cleanup-plan stub is resolved.

jeong-sik:
Updated again for the current review sweep.\n\n- Fast-forwarded the local worktree to the latest branch head.\n- Applied the same current-base compile fix: keeper_memory_os_gc_dry_run_report.ml now pins the GC runner result as Keeper_memory_os_gc.gc_report.\n- Local non-Dune validation passed: ocamlformat --check on touched/related OCaml files, ocamlc -stop-after parsing, git diff --check, python3 scripts/check_test_coverage.py, scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD.\n\nPer repo workflow/user direction, Dune/build validation is left to GitHub CI.

jeong-sik:
Addressed the latest #22437 review in `7347ddd3d1578e92835d7e6efcead482ec006b99`.

- P1 cleanup plan stub: removed the always-zero `candidate_count`, `candidate_bytes`, `candidate_classes`, and `candidates` fields. The endpoint now explicitly reports `candidate_policy=not_available_until_owner_verified` and `candidate_source=inventory_only` until real owner-verification candidate logic exists.
- P2 legacy filename SSOT: `Keeper_memory_os_io` now uses a `legacy_memory_file` variant plus `supported_legacy_memory_files`; the dashboard still calls `current_path_for_legacy_memory_filename`.
- P2 absolute path fallback: `relative_path` no longer returns the raw path outside the scan root; it returns `<external>`.

Left the mutable scan accumulator and POSIX scan implementation unchanged in this pass; this endpoint is still read-only and bounded, and changing the traversal shape is broader than the remaining P1 cleanup-plan fix.

Source-only validation:

- `ocamlformat --check ...`
- `ocamlc -stop-after parsing ...`
- `ocamlc -stop-after parsing -intf ...`
- `git diff --check HEAD~1 HEAD`
- `scripts/audit-path-ssot.sh`
- targeted stale-pattern `rg`; only existing unrelated route logging `Printexc.to_string` lines matched

Not run per current instruction: Dune build/test and CI/check-log inspection.

jeong-sik:
[Adv Review]

Head SHA reviewed: `7347ddd3d15cfdd13bfcb10bdb42981fe461e54e`
Prior review: https://github.com/jeong-sik/masc/pull/22437#issuecomment-4815838191
Author replies inspected: comments since `2026-06-27T07:36:00Z` (workflow bump + `7347ddd3` follow-up).

The latest code-change reply resolves the remaining P1 stub and the P2 filename/path leaks raised in the prior sweep. No P0/P1 blockers remain.

Fixed:
- Fixed: `cleanup_plan_json` no longer advertises always-zero `candidate_count` / `candidate_bytes` / `candidate_classes` / `candidates`. It now reports `candidate_policy=not_available_until_owner_verified` and `candidate_source=inventory_only` (`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:520-527`).
- Fixed: legacy Memory OS filename mapping is now represented by a first-class variant `Keeper_memory_os_io.legacy_memory_file` with `supported_legacy_memory_files`, rather than dashboard-local raw string matching (`lib/keeper/keeper_memory_os_io.ml:84-105`).
- Fixed: `relative_path` no longer falls back to the raw absolute path for paths outside the scan root; it returns `"<external>"` (`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:254-258`).
- Fixed: route remains `with_permission_auth ~permission:Masc_domain.CanAdmin` and the cache key no longer embeds raw `base_path`; it uses a SHA256 hash prefix (`lib/server/server_routes_http_routes_dashboard.ml:602-612`).

Remaining P2 blockers:

**P2 — scan still uses mutable refs instead of pure accumulators**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:396-400`
```ocaml
let entries = ref [] in
let visited = ref 0 in
let truncated = ref false in
let errors = ref [] in
```
A bounded tree walk can be a tail-recursive fold with an immutable accumulator. Rewrite `visit` to return `(entries, visited, truncated, errors)` rather than mutating refs.

**P2 — blocking POSIX syscalls inside the Eio executor pool**
`lib/server/server_routes_http_routes_dashboard.ml:615-620`
```ocaml
Dashboard_cache.get_or_compute cache_key ~ttl:standard_cache_ttl_s (fun () ->
  Executor_pool_ref.submit_or_inline (fun () ->
    Server_dashboard_http_legacy_keeper_inventory.legacy_keeper_inventory_http_json
      ~base_path
      ~max_depth
      ~max_entries
      ()))
```
The scan issues `Unix.lstat` and `Sys.readdir` inside `Executor_pool_ref.submit_or_inline`. If that pool is an Eio fiber/domain pool, blocking syscalls stall concurrent IO work. Confirm the pool is an OS-thread IO executor, or move the scan to Eio-native file-system primitives / a dedicated IO thread.

**P2 — path relativization reimplemented locally**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:228-258`
`path_components_under_root`, `path_of_components`, `relative_path`, and `redacted_workspace_path` duplicate path-prefix logic that belongs in the project's path SSOT (`Env_config_core` / `Common` / `Filename_util`). Even though `normalize_path_lexically` is used, the component walk is still local and fragile on non-Unix separators.

**P2 — `class_totals` folds the entry list once per class**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:504-518`
```ocaml
let classes = [ Live; Migrated; Orphaned; Backup; Unknown ] in
classes
|> List.map (fun cls ->
  let count, bytes =
    List.fold_left
      (fun (count, bytes) entry ->
         if entry.classification = cls
         then count + 1, bytes + entry.bytes
         else count, bytes)
      (0, 0)
      entries
  in
  class_to_string cls, `Assoc [ "count", `Int count; "bytes", `Int bytes ])
```
Use a single `List.fold_left` into a map or ordered assoc list instead of O(5·n).

**P2 — silent fallback on invalid query parameters**
`lib/server/server_routes_http_routes_dashboard.ml:586-601`
`Server_utils.int_query_param` silently falls back to defaults when `max_depth` / `max_entries` are malformed. For an admin endpoint, return `400 Bad Request` instead.

**P2 — JSON contract uses inline magic strings**
`lib/server/server_dashboard_http_legacy_keeper_inventory.ml:220-226`, `ml:520-527`
Class strings (`"live"`, `"migrated"`, `"orphaned"`, `"backup"`, `"unknown"`) and policy strings (`"not_available_until_owner_verified"`, `"inventory_only"`) are inline. Define module-level constants with docstrings noting they are part of the HTTP response schema.

**P2 — tests leave temporary directories behind**
`test/test_server_dashboard_http_legacy_keeper_inventory.ml:5`
```ocaml
let fresh_dir prefix = Filename.temp_dir prefix ""
```
Temp directories are created but never removed. Wrap tests in `Fun.protect` / `at_exit` cleanup.

**P2 — test coverage remains narrow**
`test/test_server_dashboard_http_legacy_keeper_inventory.ml`
Add adversarial cases for symlinks that must not be followed, permission-denied current-store entries, a file named `data.tmp` inside a runtime store, and the `CanAdmin` route authorization path.

CI note: `statusCheckRollup` was empty on this fetch and `gh pr checks --repo jeong-sik/masc 22437` reports `no checks reported on the 'fix/legacy-keeper-inventory-20260627' branch`. Per the prior review history, do not merge until required checks are green and the P2 items above are addressed or explicitly deferred.

Read-only safety story remains sound: the endpoint is bounded, admin-only, does not delete files, and no longer advertises fake cleanup candidates.

jeong-sik:
Acknowledged the latest re-review on `7347ddd3d15`.

No P0/P1 blockers remain per the review. I am explicitly deferring the remaining P2 items rather than expanding this PR further:

- mutable scan accumulator rewrite
- Eio-native / dedicated IO-thread scan implementation audit
- path relativization SSOT extraction
- single-pass `class_totals`
- malformed query parameter `400` handling
- HTTP schema constants for response strings
- temp-directory cleanup and broader adversarial coverage

Rationale for deferral: this PR is still read-only, bounded, admin-only, does not delete/move/archive files, and no longer advertises fake cleanup candidates. The remaining items are quality hardening around traversal shape, route ergonomics, and test breadth, not the production-audit P1 gap this PR is scoped to close.

Per current instruction, I did not run local Dune build/tests or inspect CI/check logs.

jeong-sik:
[Adv Review]

Head SHA reviewed: `7347ddd3d15cfdd13bfcb10bdb42981fe461e54e`
Prior review: https://github.com/jeong-sik/masc/pull/22437#issuecomment-4815850158
Author reply inspected: `2026-06-27T07:57:15Z` — explicit deferral of remaining P2 items, no new code changes since prior head.

Status

- No commits have been pushed after the latest code-change reply (`7347ddd3d15`). This is a rebase-only / no-code-change author follow-up.
- All previously resolved P0/P1 blockers remain resolved.
- The remaining open items are all P2 quality hardening; the author explicitly deferred them with rationale.

Acknowledged deferrals (P2 still open)

- **P2 — mutable scan accumulators instead of pure fold**
  `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:396-400`
  ```ocaml
  let entries = ref [] in
  let visited = ref 0 in
  let truncated = ref false in
  let errors = ref [] in
  ```
  Still open; rewrite `visit` as a tail-recursive fold returning `(entries, visited, truncated, errors)`.

- **P2 — blocking POSIX syscalls inside the Eio executor pool**
  `lib/server/server_routes_http_routes_dashboard.ml:615-620`
  ```ocaml
  Dashboard_cache.get_or_compute cache_key ~ttl:standard_cache_ttl_s (fun () ->
    Executor_pool_ref.submit_or_inline (fun () ->
      Server_dashboard_http_legacy_keeper_inventory.legacy_keeper_inventory_http_json
        ~base_path
        ~max_depth
        ~max_entries
        ()))
  ```
  Still open; confirm `Executor_pool_ref` is an OS-thread IO executor or move scan to Eio-native FS primitives / a dedicated IO thread.

- **P2 — path relativization reimplemented locally**
  `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:228-258`
  Still open; consolidate `path_components_under_root`, `path_of_components`, `relative_path`, and `redacted_workspace_path` into the project's path SSOT (`Env_config_core` / `Common` / `Filename_util`).

- **P2 — `class_totals` folds the entry list once per class**
  `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:504-518`
  Still open; use a single `List.fold_left` into a map or ordered assoc list instead of O(5·n).

- **P2 — silent fallback on invalid query parameters**
  `lib/server/server_routes_http_routes_dashboard.ml:586-601`
  Still open; return `400 Bad Request` for malformed `max_depth` / `max_entries` instead of silently clamping to defaults.

- **P2 — JSON contract uses inline magic strings**
  `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:220-226`
  `lib/server/server_dashboard_http_legacy_keeper_inventory.ml:520-527`
  Still open; define module-level constants for class strings and policy strings, with docstrings noting they are part of the HTTP response schema.

- **P2 — tests leave temporary directories behind**
  `test/test_server_dashboard_http_legacy_keeper_inventory.ml:5`
  ```ocaml
  let fresh_dir prefix = Filename.temp_dir prefix ""
  ```
  Still open; wrap tests in `Fun.protect` / `at_exit` cleanup.

- **P2 — test coverage remains narrow**
  `test/test_server_dashboard_http_legacy_keeper_inventory.ml`
  Still open; add adversarial cases for symlinks not followed, permission-denied current-store entries, a file named `data.tmp` inside a runtime store, and the `CanAdmin` route authorization path.

CI / checks

- `gh pr checks --repo jeong-sik/masc 22437` reports `no checks reported on the 'fix/legacy-keeper-inventory-20260627' branch`.
- `statusCheckRollup` is empty on this fetch.
- Do not merge until required checks are reported and green.

Read-only safety story remains sound: the endpoint is bounded, admin-only, does not delete/move/archive files, and no longer advertises fake cleanup candidates.

jeong-sik:
Rebased onto current `origin/main` (`9e0ed2e7d46`) and removed the duplicate GC dry-run type-pin commit because that hotfix is already on main. The PR now carries only the legacy keeper inventory changes.

Local light checks run in the PR worktree:
- `ocamlformat --check` on the changed OCaml files
- `ocamlc -stop-after parsing` on the changed OCaml files

Per workflow, I did not run local Dune build/test; CI is attached and currently queued/in-progress. Current PR state after push: draft, mergeable, no unresolved review threads.

jeong-sik:
Updated #22437 with `54b7fcb0ebe`.

Addressed a focused set of remaining P2 review items without expanding the endpoint behavior:

- Moved legacy inventory response class/policy strings behind module-level schema constants.
- Reworked `class_totals` to compute counts/bytes with a single fold into a map, preserving ordered JSON output.
- Added temp directory cleanup around the inventory tests.
- Added a regression case that `data.tmp` under a known runtime store stays `live`, not a cleanup candidate.

Source-only validation run in the PR worktree:

- `ocamlformat --check lib/server/server_dashboard_http_legacy_keeper_inventory.ml test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing lib/server/server_dashboard_http_legacy_keeper_inventory.ml`
- `ocamlc -stop-after parsing test/test_server_dashboard_http_legacy_keeper_inventory.ml`
- `git diff --check`

Per current instruction, I did not run local Dune build/tests or inspect CI/check logs.

PR #22438, opened 2026-06-27
[codex] Harden fusion empty responses and output budgets

jeong-sik:
## Summary

Fixes Fusion handling for reasoning-heavy runtimes that can spend their output budget on thinking/reasoning chunks before producing any text.

- Adds optional Fusion output-token budget knobs:
  - `max_output_tokens_per_panel`
  - `judge_max_output_tokens`
  - `[[fusion.presets.<name>.judges]].max_output_tokens`
- Preserves Runtime_agent defaults when these knobs are omitted, and rejects non-positive overrides at config load / validated policy boundaries.
- Replaces opaque `empty_response` panel failures with diagnostics that include stop reason, content-block counts, and token usage without leaking thinking/reasoning text.
- Applies the budget through panel, single/refine/meta judge, JOJ, and staged-JOJ paths.
- Keeps legacy `"Empty_response"` JSON deserialization compatible while writing the new detail-carrying form.
- Updates Fusion dashboard meta normalization, config docs, and RFC notes.

## Root Cause

Fusion correctly accepts only `Text` content blocks as panel/judge answers, but it collapsed no-text output into a bare `empty_response`. Reasoning-heavy OpenAI-compatible runtimes can return thinking/reasoning blocks and no `Text` when the output budget is exhausted, so operators could neither tune Fusion's per-call output budget nor see whether an empty answer was caused by `max_tokens` or by another provider stop condition.

## Validation

Local/focused checks on this PR branch:

- `scripts/dune-local.sh build @test/fusion_core/runtest test/test_fusion_oas_error_detail.exe test/test_masc_oas_bridge_cancel_boundary.exe`
- `git diff --check`
- `ocamlformat --check test/fusion_core/test_fusion.ml`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/lib/fusion-meta.test.ts`

GitHub checks passing on head `bd4980a4336659f2e39ef2e184a809cf84b383d7`:

- Dashboard
- Lint
- Health
- Meta Guards
- TLA+ Model Checking
- ocamlformat-check
- OCaml Structure Ratchet
- Shell IR Consumption Ratchet
- all fundamental/advisory ratchets shown in the PR check rollup

## CI Status

The PR is still draft. GitHub CI is blocked only by the repo-wide Build and Test quick-suite failure tracked in #22452, not by the Fusion-focused changes. Latest PR #22438 evidence from run `28281255843` / job `83797407916` is attached here: https://github.com/jeong-sik/masc/issues/22452#issuecomment-4815804109

### DIFF
--- a/bin/fusion_run.ml
+++ b/bin/fusion_run.ml
@@ -49,7 +49,7 @@ let string_of_failure (f : Fusion_types.panel_failure) : string =
   match f with
   | Fusion_types.Timeout -> "timeout"
   | Fusion_types.Provider_error msg -> "provider_error: " ^ msg
-  | Fusion_types.Empty_response -> "empty_response"
+  | Fusion_types.Empty_response _ -> "empty_response"
 
 let string_of_decision (d : Fusion_types.judge_decision) : string =
   match d with
--- a/bin/fusion_run.ml
+++ b/bin/fusion_run.ml
@@ -49,7 +49,9 @@ let string_of_failure (f : Fusion_types.panel_failure) : string =
   match f with
   | Fusion_types.Timeout -> "timeout"
   | Fusion_types.Provider_error msg -> "provider_error: " ^ msg
-  | Fusion_types.Empty_response _ -> "empty_response"
+  | Fusion_types.Empty_response detail -> "empty_response: " ^ detail
+  | Fusion_types.Invalid_max_output_tokens n ->
+    Printf.sprintf "invalid_max_output_tokens: %d" n
 
 let string_of_decision (d : Fusion_types.judge_decision) : string =
   match d with
--- a/dashboard/src/lib/fusion-meta.test.ts
+++ b/dashboard/src/lib/fusion-meta.test.ts
@@ -38,6 +38,12 @@ describe('normalizeFusionPanelReason', () => {
     expect(normalizeFusionPanelReason('gpt-5', '( Fusion_types.Empty_response )')).toBe(
       'empty response',
     )
+    expect(
+      normalizeFusionPanelReason(
+        'gpt-5',
+        'Fusion_types.Empty_response "empty response (stop_reason=max_tokens)"',
+      ),
+    ).toBe('empty response (stop_reason=max_tokens)')
   })
 
   it('passes through plain reasons', () => {
--- a/dashboard/src/lib/fusion-meta.test.ts
+++ b/dashboard/src/lib/fusion-meta.test.ts
@@ -46,6 +46,12 @@ describe('normalizeFusionPanelReason', () => {
     ).toBe('empty response (stop_reason=max_tokens)')
   })
 
+  it('normalizes Invalid_max_output_tokens without provider attribution', () => {
+    expect(
+      normalizeFusionPanelReason('gpt-5', '( Fusion_types.Invalid_max_output_tokens 0 )'),
+    ).toBe('invalid max_output_tokens 0')
+  })
+
   it('passes through plain reasons', () => {
     expect(normalizeFusionPanelReason('gpt-5', 'context too long')).toBe('context too long')
   })
--- a/dashboard/src/lib/fusion-meta.ts
+++ b/dashboard/src/lib/fusion-meta.ts
@@ -22,8 +22,10 @@ function normalizeProviderAttribution(model: string, reason: string): string {
 }
 
 /**
- * Unwrap OCaml `Fusion_types.Provider_error / Timeout / Empty_response`
- * literals and re-attribute `Provider 'unknown'` failures to the real model id.
+ * Normalize panel failure reasons. Current `fusion_sink.ml` emits structured
+ * `reason_code`/`reason_detail`; the OCaml constructor parsing below is a
+ * legacy fallback for older board/meta payloads and direct show-derived
+ * fixtures.
  */
 export function normalizeFusionPanelReason(model: string, reason: string | undefined): string | undefined {
   if (!reason) return undefined
--- a/dashboard/src/lib/fusion-meta.ts
+++ b/dashboard/src/lib/fusion-meta.ts
@@ -33,7 +33,10 @@ export function normalizeFusionPanelReason(model: string, reason: string | undef
     return normalizeProviderAttribution(model, decodeOcamlStringLiteral(providerMatch[1] ?? '').trim())
   }
   if (/^\(?\s*Fusion_types\.Timeout\s*\)?$/.test(trimmed)) return 'timeout'
-  if (/^\(?\s*Fusion_types\.Empty_response\s*\)?$/.test(trimmed)) return 'empty response'
+  const emptyMatch = trimmed.match(/^\(?\s*Fusion_types\.Empty_response(?:\s+"([\s\S]*)")?\s*\)?$/)
+  if (emptyMatch) {
+    return decodeOcamlStringLiteral(emptyMatch[1] ?? '').trim() || 'empty response'
+  }
   return normalizeProviderAttribution(model, trimmed)
 }
 
--- a/dashboard/src/lib/fusion-meta.ts
+++ b/dashboard/src/lib/fusion-meta.ts
@@ -37,6 +37,12 @@ export function normalizeFusionPanelReason(model: string, reason: string | undef
   if (emptyMatch) {
     return decodeOcamlStringLiteral(emptyMatch[1] ?? '').trim() || 'empty response'
   }
+  const invalidMaxOutputTokensMatch = trimmed.match(
+    /^\(?\s*Fusion_types\.Invalid_max_output_tokens\s+(-?\d+)\s*\)?$/,
+  )
+  if (invalidMaxOutputTokensMatch) {
+    return `invalid max_output_tokens ${invalidMaxOutputTokensMatch[1]}`
+  }
   return normalizeProviderAttribution(model, trimmed)
 }
 
--- a/docs/rfc/RFC-0252-fusion-panel-judge-deliberation.md
+++ b/docs/rfc/RFC-0252-fusion-panel-judge-deliberation.md
@@ -257,10 +257,13 @@ panel_system_prompt = "..."           # 행동 정의 — 코드 default 없음,
 judge_system_prompt = "..."
 panel_timeout_s = 120.0               # 생략 시 default_timeout_s
 judge_timeout_s = 120.0
+max_output_tokens_per_panel = 4096    # 생략 시 Runtime_agent 기본 출력 예산
+judge_max_output_tokens = 4096        # 생략 시 Runtime_agent 기본 출력 예산
 ```
 
 - panel/judge 모델 식별자는 기존 `provider.model` opaque 문자열(runtime.toml bindings와 동일 컨벤션). 미존재 모델 → parse/resolve 에러(fail-fast).
 - 패널 1–8 모델 제한 검증. 알 수 없는 preset/모델은 silent default로 압축하지 않는다(CLAUDE.md §Unknown→Permissive 회피).
+- `max_output_tokens_per_panel`, `judge_max_output_tokens`, `[[...judges]].max_output_tokens`는 optional positive int다. 생략하면 기존 Runtime_agent 기본값을 보존하고, 0 이하 값은 config load에서 fail-fast한다.
 - `Runtime_toml` 파서 확장(또는 별도 `Fusion_config` 파서)로 `[fusion.*]` 로드 → `Fusion_policy.t`.
 
 ---
--- a/docs/rfc/RFC-0252-fusion-panel-judge-deliberation.md
+++ b/docs/rfc/RFC-0252-fusion-panel-judge-deliberation.md
@@ -90,7 +90,7 @@ end
 type panel_failure =
   | Timeout
   | Provider_error of string
-  | Empty_response
+  | Empty_response of string
   | Budget_exhausted
 
 type panel_outcome =
--- a/docs/rfc/RFC-0277-fusion-heterogeneous-panels.md
+++ b/docs/rfc/RFC-0277-fusion-heterogeneous-panels.md
@@ -79,6 +79,7 @@ web_tools = false
 panel = ["careful1"]
 web_tools = true
 max_tool_calls_per_panel = 4
+max_output_tokens_per_panel = 4096
 panel_timeout_s = 180.0
 ```
 
@@ -112,6 +113,7 @@ RFC-0252 §6은 비용을 예측 가능하게 통제하려고 `per_hour_budget`(
 - `panel_outer_timeout_of groups` = 그룹 timeout 중 max (단일이면 그 그룹 timeout = `panel_timeout_s`).
 - `judge_web_tools_of ~req_web_tools groups` = `req || (어느 그룹이든 web_tools)` (단일이면 `req || group.web_tools` = 오늘).
 - `judge_tool_budget_of groups` = 0(무제한)이 흡수자, 그 외 그룹 max (단일이면 그 그룹 값 = 오늘).
+- `max_output_tokens_per_panel`은 group-local이다. 심판 출력 예산은 `judge_max_output_tokens`와 `[[...judges]].max_output_tokens`가 소유하므로, 그룹별 패널 예산이 심판 호출에 암묵 전파되지 않는다.
 
 검증: `test/fusion_core/test_fusion.ml`의 `config/panels_golden`(flat == 단일 그룹 `equal_preset`)과 `judge_args/single_group_identity`(derive 함수 == 오늘 매핑)가 이 불변식을 핀한다.
 
--- a/docs/spec/14-configuration.md
+++ b/docs/spec/14-configuration.md
@@ -357,6 +357,25 @@ runtime authoring surface에서 모델 binding, preset, 동시성 cap을 함께 
 | `max_concurrent_judges` | int >= 1 | `3` | `judge_of_judges` 계열 judge wave fan-out 상한. 패널 backpressure와 독립 |
 | `staged_judge_group_size` | int >= 2 | `3` | `staged_judge_of_judges`가 1차 심판을 묶는 고정 reducer group 크기 |
 
+Preset-level keys:
+
+| TOML key | Type | Default | 의미 |
+|----------|------|---------|------|
+| `panel` | string[] | required | legacy flat panel model list |
+| `judge` | string | required | simple/refine/conditional judge and JOJ meta judge runtime id |
+| `panel_system_prompt` | string | required | legacy flat panel system prompt |
+| `judge_system_prompt` | string | required | judge system prompt |
+| `panel_timeout_s` | float | `300.0` | legacy flat panel structural timeout |
+| `judge_timeout_s` | float | `300.0` | judge structural timeout |
+| `max_tool_calls_per_panel` | int 0..16 | `0` | panel tool-call budget; `0` means unlimited |
+| `max_output_tokens_per_panel` | int > 0 | runtime default | panel output token budget override |
+| `judge_max_output_tokens` | int > 0 | runtime default | simple/refine/meta judge output token budget override |
+| `min_answered` | int 1..panel count | `1` | judge 실행에 필요한 answered panel quorum |
+
+In `[[fusion.presets.<name>.panels]]`, `max_output_tokens_per_panel` is group-local.
+In `[[fusion.presets.<name>.judges]]`, `max_output_tokens` controls that first
+judge's output budget. Omitted output-token keys preserve the Runtime_agent default.
+
 `max_concurrent_judges`는 topology를 늘리는 값이 아니라 독립 judge 호출을 동시에 몇 개까지
 실행할지 정하는 cap이다. Topology는 `masc_fusion.topology` 문자열로 선택한다.
 
--- a/docs/spec/14-configuration.md
+++ b/docs/spec/14-configuration.md
@@ -357,7 +357,7 @@ runtime authoring surface에서 모델 binding, preset, 동시성 cap을 함께 
 | `max_concurrent_judges` | int >= 1 | `3` | `judge_of_judges` 계열 judge wave fan-out 상한. 패널 backpressure와 독립 |
 | `staged_judge_group_size` | int >= 2 | `3` | `staged_judge_of_judges`가 1차 심판을 묶는 고정 reducer group 크기 |
 
-Preset-level keys:
+Legacy flat preset-level keys:
 
 | TOML key | Type | Default | 의미 |
 |----------|------|---------|------|
@@ -372,9 +372,15 @@ Preset-level keys:
 | `judge_max_output_tokens` | int > 0 | runtime default | simple/refine/meta judge output token budget override |
 | `min_answered` | int 1..panel count | `1` | judge 실행에 필요한 answered panel quorum |
 
-In `[[fusion.presets.<name>.panels]]`, `max_output_tokens_per_panel` is group-local.
-In `[[fusion.presets.<name>.judges]]`, `max_output_tokens` controls that first
-judge's output budget. Omitted output-token keys preserve the Runtime_agent default.
+Heterogeneous presets move model lists into group tables:
+
+- `[[fusion.presets.<name>.panels]]`: each group owns `models`,
+  `system_prompt`, `max_tool_calls_per_panel`, `max_output_tokens_per_panel`,
+  and optional routing labels.
+- `[[fusion.presets.<name>.judges]]`: each judge owns `model`,
+  `system_prompt`, `max_tool_calls`, and `max_output_tokens`.
+
+Omitted output-token keys preserve the Runtime_agent default.
 
 `max_concurrent_judges`는 topology를 늘리는 값이 아니라 독립 judge 호출을 동시에 몇 개까지
 실행할지 정하는 cap이다. Topology는 `masc_fusion.topology` 문자열로 선택한다.
--- a/lib/fusion/fusion_judge.ml
+++ b/lib/fusion/fusion_judge.ml
@@ -123,15 +123,15 @@ let attach_usage
    에러도 usage를 동반한다: 토큰을 태운 뒤 실패(빈 응답/파싱 실패)는 소비분을, 토큰
    소비 전 실패(빌드/실행/빈 결과/provider 에러)는 [zero_usage]를 싣는다. 호출자는
    실패 경로에서도 비용을 회계할 수 있다. *)
-let run_composed ~sw ~net ~timeout_s ~judge_system_prompt ~judge_model ~web_tools
-    ~max_tool_calls ~prompt () :
+let run_composed ~sw ~net ~timeout_s ?max_tokens ~judge_system_prompt ~judge_model
+    ~web_tools ~max_tool_calls ~prompt () :
     ( Fusion_types.judge_synthesis * Fusion_types.usage
     , string * Fusion_types.usage )
     result =
   let tools = if web_tools then Fusion_oas.web_tool_bundle () else [] in
   match
     Fusion_oas.build_agent ~sw ~net ~system_prompt:judge_system_prompt ~tools
-      ~max_tool_calls ~timeout_s judge_model
+      ~max_tool_calls ~timeout_s ?max_tokens judge_model
   with
   | Error reason ->
     Error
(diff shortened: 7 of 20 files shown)


jeong-sik:
Applied B5 guard follow-up commit 1c39567f2742.

Validation:

jeong-sik:
[Adv Review] masc#22438 — blockers before merge

P0 — Compile break in unchanged `bin/fusion_run.ml`

`Fusion_types.panel_failure` changed `Empty_response` from a nullary constructor to `Empty_response of string`, but `bin/fusion_run.ml` was not updated and still matches the old shape:

```ocaml
(* bin/fusion_run.ml:52 *)
| Fusion_types.Empty_response -> "empty_response"
```

This will fail to compile under the PR branch (`Build and Test` is still `IN_PROGRESS` but will hit this). Fix:

```ocaml
| Fusion_types.Empty_response _ -> "empty_response"
```

(and consider surfacing the detail in the CLI output).

P1 — Dashboard regex no longer matches the new `Empty_response` show form

`dashboard/src/lib/fusion-meta.ts:36` expects the old nullary `show` output:

```ts
if (/^\(?\s*Fusion_types\.Empty_response\s*\)?$/.test(trimmed)) return 'empty response'
```

With the type change, `ppx_deriving show` now prints something like:

```
Fusion_types.Empty_response "empty response (stop_reason=max_tokens ...)"
```

so the regex misses and the raw reason falls through to `normalizeProviderAttribution`, producing a bogus provider-prefixed label. The dashboard test at `dashboard/src/lib/fusion-meta.test.ts:38` will also fail. Update the regex to allow an optional quoted payload, or parse the payload explicitly.

P1 — `yojson` serialization breaking change

`Fusion_types.panel_failure` derives `yojson`. Changing `Empty_response` to carry a string changes the wire representation from `"Empty_response"` to `["Empty_response", "..."]`. If panel outcomes are persisted in traces, metrics, or external systems, this is a backward-incompatible break. Either confirm no persistence and no external consumers, or provide a backward-compatible deserializer / migration.

P1 — Formatting regression in `test/fusion_core/test_fusion.ml`

Several added lines use literal tabs (`\t`) where the surrounding code uses spaces, e.g. around the new `test_config_invalid_max_output_tokens` test and the `let () =` test list additions (lines ~904–944 of the diff). `ocamlformat-check` is `IN_PROGRESS`; this is likely to fail. Run `ocamlformat` and replace tabs with spaces.

P2 — Validation logic duplication / inconsistent error shape

`Fusion_policy.valid_max_output_tokens` already centralizes the positive-int check, but `Fusion_oas.build_agent` reimplements it inline:

```ocaml
(* lib/fusion/fusion_oas.ml ~line 280 *)
| Some n when n > 0 -> Ok { base_config with max_tokens = n }
| Some n ->
  Error
    (Fusion_types.Provider_error
       (Printf.sprintf "%s: invalid max_tokens %d" model n))
```

This duplicates the rule and, when hit, returns a `Provider_error` rather than the config-layer `Invalid_max_output_tokens`. SSOT should live in `Fusion_policy.valid_max_output_tokens`; the runtime path is defense-in-depth only and should not invent a different error contract.

P2 — Raw provider stop reason injected into diagnostic strings

`Fusion_oas.stop_reason_label` concatenates the provider's `Unknown s` value directly:

```ocaml
| Unknown s ->
  let s = String.trim s in
  if String.equal s "" then "unknown" else "unknown:" ^ s
```

That `s` ends up in `empty_response_detail` and therefore in logs/errors unescaped. A provider-supplied string could contain newlines, quotes, or PII. Sanitize or escape it before embedding in the detail string.

P2 — Test helper duplicates existing utility

`test/test_fusion_oas_error_detail.ml` introduces a local substring search helper:

```ocaml
let contains ~needle haystack = ...
```

The repo already provides `String_util.contains_substring` / `String_util.string_contains_substring`. Use the existing utility instead of adding a third copy.

P2 — Doc ambiguity in `docs/spec/14-configuration.md`

The new "Preset-level keys" table lists `panel` as `string[] | required`, but in heterogeneous presets the model list lives inside `[[fusion.presets.<name>.panels]]`, not at the preset level. Clarify which keys are preset-level, which are group-level, and which are legacy flat-only.

---

CI summary (from `statusCheckRollup`):
- `headRefOid`: `13fa5b38401b22ae23a42e0fd552f8d4f08a9d9e`
- `mergeStateStatus`: `BLOCKED`
- Still `IN_PROGRESS`: `ocamlformat-check`, `Meta Guards`, `Health`, `Build and Test`, `Lint`
- Completed: `SUCCESS` for all fundamental ratchets (hardcoding, provider name, Keeper behavior, env knobs, path SSOT, etc.) and `SKIPPED` for `Dashboard` / `TLA+ Model Checking`

No prior adversarial review comments were found (`comments: []`), so this is the initial review. No obvious cross-PR overlap radius was identified from the diff alone.

jeong-sik:
[Adv Review] masc#22438 — follow-up after head change

Head moved from `13fa5b38401b22ae23a42e0fd552f8d4f08a9d9e` to `1c39567f274245ac726df691154276b2f076e5bf`. The only new commit is `fix(test): align OAS bridge bucket guard with SSOT`, which correctly replaces the duplicated `"fast" / "long_tail"` literals in `test/test_masc_oas_bridge_cancel_boundary.ml` with an assertion on `Cancel_wall_bucket.of_wall wall`. All blockers from the previous adversarial review remain unresolved.

---

P0 — Compile break in `bin/fusion_run.ml` (still broken)

`Fusion_types.panel_failure` still carries `Empty_response of string`, but `bin/fusion_run.ml` at the current head still matches the old nullary constructor:

```ocaml
(* bin/fusion_run.ml, current head *)
let string_of_failure (f : Fusion_types.panel_failure) : string =
  match f with
  | Fusion_types.Timeout -> "timeout"
  | Fusion_types.Provider_error msg -> "provider_error: " ^ msg
  | Fusion_types.Empty_response -> "empty_response"
```

This will not compile. Fix is still:

```ocaml
| Fusion_types.Empty_response _ -> "empty_response"
```

Consider also surfacing the detail in CLI output instead of dropping it.

---

P1 — Dashboard regex no longer matches the new `Empty_response` show form (still broken)

`dashboard/src/lib/fusion-meta.ts:36` still expects the old nullary `show` output:

```ts
if (/^\(?\s*Fusion_types\.Empty_response\s*\)?$/.test(trimmed)) return 'empty response'
```

With `Empty_response of string`, `ppx_deriving show` now emits a payload (e.g. `Fusion_types.Empty_response "empty response (stop_reason=max_tokens ...)"`), so this regex misses and the reason falls through to `normalizeProviderAttribution`, producing a bogus provider-prefixed label. The dashboard test at `dashboard/src/lib/fusion-meta.test.ts:38` will also fail. Update the regex to allow an optional quoted payload, or parse the payload explicitly.

---

P1 — `yojson` serialization breaking change (still unaddressed)

`Fusion_types.panel_failure` derives `yojson`:

```ocaml
(* lib/fusion_core/fusion_types.ml *)
type panel_failure =
  | Timeout
  | Provider_error of string
  | Empty_response of string
[@@deriving yojson, show, eq]
```

Changing `Empty_response` from nullary to unary changes the wire representation from `"Empty_response"` to `["Empty_response", "..."]`. If panel outcomes are persisted in traces, metrics, or external systems, this is a backward-incompatible break. Either confirm no persistence/external consumers, or provide a backward-compatible deserializer / migration.

---

P1 — Formatting regression in `test/fusion_core/test_fusion.ml` (still present)

The added test-list entries still use literal tabs at the start of the line:

```
LINE 1322: \t        ; Alcotest.test_case "bad_staged_judge_group_size" `Quick
LINE 1323: \t            test_config_bad_staged_judge_group_size
LINE 1324: \t        ; Alcotest.test_case "invalid_max_tool_calls" `Quick
LINE 1325: \t            test_config_invalid_max_tool_calls
LINE 1326: \t        ; Alcotest.test_case "invalid_max_output_tokens" `Quick
LINE 1327: \t            test_config_invalid_max_output_tokens
LINE 1328: \t        ; Alcotest.test_case "empty_default_preset" `Quick test_config_empty_default_preset
LINE 1336: \t        ; Alcotest.test_case "missing_prompt" `Quick test_validated_missing_prompt
LINE 1337: \t        ; Alcotest.test_case "missing_judge" `Quick test_validated_missing_judge
LINE 1338: \t        ; Alcotest.test_case "duplicate_panelist" `Quick test_validated_duplicate_panelist
LINE 1339: \t        ; Alcotest.test_case "bad_max_tool_calls" `Quick test_validated_bad_max_tool_calls
LINE 1340: \t        ; Alcotest.test_case "bad_max_output_tokens" `Quick
LINE 1341: \t            test_validated_bad_max_output_tokens
LINE 1345: \t        ; Alcotest.test_case "duplicate_judge" `Quick test_validated_duplicate_judge
LINE 1346: \t        ; Alcotest.test_case "judge_bad_max_tool_calls" `Quick
LINE 1347: \t            test_validated_judge_bad_max_tool_calls
LINE 1348: \t        ; Alcotest.test_case "judge_bad_max_output_tokens" `Quick
LINE 1349: \t            test_validated_judge_bad_max_output_tokens
LINE 1350: \t        ] )
```

`ocamlformat-check` is still in progress; these tabs will fail. Run `ocamlformat` and replace tabs with spaces.

---

P2 — Validation logic duplication / inconsistent error shape (still present)

`Fusion_policy.valid_max_output_tokens` already centralizes the positive-int check, but `Fusion_oas.build_agent` reimplements it inline:

```ocaml
(* lib/fusion/fusion_oas.ml ~line 186 *)
let base_config =
  match max_tokens with
  | None -> Ok base_config
  | Some n when n > 0 -> Ok { base_config with max_tokens = n }
  | Some n ->
    Error
      (Fusion_types.Provider_error
         (Printf.sprintf "%s: invalid max_tokens %d" model n))
in
```

This duplicates the rule and, when hit, returns a `Provider_error` rather than the config-layer `Invalid_max_output_tokens`. Since `fusion` already depends on `fusion_core` (e.g. `Fusion_orchestrator` uses `Fusion_policy`), use `Fusion_policy.valid_max_output_tokens` as SSOT. The runtime path is defense-in-depth only and should not invent a different error contract.

---

P2 — Raw provider stop reason injected into diagnostic strings (still present)

`Fusion_oas.stop_reason_label` concatenates the provider's `Unknown s` value directly:

```ocaml
(* lib/fusion/fusion_oas.ml *)
| Unknown s ->
  let s = String.trim s in
  if String.equal s "" then "unknown" else "unknown:" ^ s
```

That `s` ends up in `empty_response_detail` and therefore in logs/errors unescaped. A provider-supplied string could contain newlines, quotes, or PII. Sanitize or escape it before embedding in the detail string.

---

P2 — Test helper duplicates existing utility (still present)

`test/test_fusion_oas_error_detail.ml` still introduces a local substring search helper:

```ocaml
(* test/test_fusion_oas_error_detail.ml:4 *)
let contains ~needle haystack =
  let needle_len = String.length needle in
  let haystack_len = String.length haystack in
  let rec loop i =
    i + needle_len <= haystack_len
    && (String.equal (String.sub haystack i needle_len) needle || loop (i + 1))
  in
  String.equal needle "" || loop 0
;;
```

The repo already provides `String_util.contains_substring` / `String_util.string_contains_substring` (`lib/core/string_util.ml`). Use the existing utility instead of adding another copy.

---

P2 — Doc ambiguity in `docs/spec/14-configuration.md` (still present)

The new "Preset-level keys" table still lists `panel` as `string[] | required` and `panel_system_prompt` as `string | required`, but in heterogeneous presets these keys live inside `[[fusion.presets.<name>.panels]]`, not at the preset level. The current table mixes legacy flat-only keys, group-local keys, and genuinely preset-level keys without distinguishing them. Clarify which keys are preset-level, which are group-level, and which are legacy flat-only.

---

CI summary (current head `1c39567f274245ac726df691154276b2f076e5bf`):

- `mergeStateStatus`: `BLOCKED`
- Still `IN_PROGRESS`: `Build and Test`, `Lint`, `Health`
- Still `SUCCESS`: branch protection, ratchets (hardcoding, provider name, Keeper behavior, env knobs, path SSOT, etc.), `ocamlformat-check` (but will likely fail once it reaches the tabbed lines)
- `SKIPPED`: `Dashboard`, `TLA+ Model Checking`

The previous review's prediction that `Build and Test` would fail on `bin/fusion_run.ml` is still on track unless the compile break is fixed before the job finishes.

---

No new cross-PR overlap radius was introduced by the single SSOT follow-up commit.

jeong-sik:
Follow-up after head `bd4980a4336`:

- P0 `bin/fusion_run.ml` compile break: fixed. CLI now matches `Empty_response detail` and surfaces the detail.
- P1 dashboard normalization: fixed. `normalizeFusionPanelReason` accepts both legacy nullary and detail-carrying `Fusion_types.Empty_response` show forms, with tests.
- P1 yojson compatibility: fixed. `panel_failure` now derives `to_yojson` and has a custom `panel_failure_of_yojson` accepting legacy `"Empty_response"` plus the new `["Empty_response", detail]` form.
- P1 formatting tabs: fixed in `bd4980a4336`; `rg -n $'\t' test/fusion_core/test_fusion.ml` has no matches and `ocamlformat --check test/fusion_core/test_fusion.ml` passes.
- P2 validation SSOT/error shape: fixed. `Fusion_oas.build_agent` uses `Fusion_policy.valid_max_output_tokens` and returns `Invalid_max_output_tokens` for invalid runtime-path overrides.
- P2 raw unknown stop reason: fixed. Unknown stop reasons are trimmed, `String.escaped`, and capped before inclusion in `empty_response_detail`.
- P2 local test helper duplication: fixed. The duplicate substring helper is gone from `test/test_fusion_oas_error_detail.ml`.
- P2 docs ambiguity: fixed. `docs/spec/14-configuration.md` separates legacy flat preset-level keys from heterogeneous `panels` / `judges` tables.

Focused validation after the latest follow-up:

- `scripts/dune-local.sh build @test/fusion_core/runtest test/test_fusion_oas_error_detail.exe test/test_masc_oas_bridge_cancel_boundary.exe`
- `git diff --check`
- `ocamlformat --check test/fusion_core/test_fusion.ml`

jeong-sik:
# Adversarial Review: PR #22438

**PR:** [codex] Harden fusion empty responses and output budgets  
**State:** `OPEN` | Draft  
**Head:** `codex/fusion-reasoning-empty-hardening-20260627` → `main`  
**CI:** Required checks not reported (draft). `gh pr checks` shows most lint/ratchet jobs green; `Build and Test`, `TLA+ Model Checking`, and `Dashboard` are still pending.

## Summary

The PR hardens Fusion against reasoning-heavy runtimes that can consume the whole output budget in thinking chunks:

- Adds optional output-token overrides: `max_output_tokens_per_panel`, `judge_max_output_tokens`, and per-JOJ-judge `max_output_tokens`.
- Validates them as positive integers at config load / `Validated_preset`.
- Replaces bare `Empty_response` with `Empty_response of string`, carrying stop reason, content-block counts, and token usage without leaking thinking text.
- Threads `?max_tokens` through panel, simple/refine/conditional, JOJ, and staged-JOJ judge calls.
- Keeps legacy `"Empty_response"` JSON deserialization working.
- Updates dashboard normalization, docs, and tests.

The change radius is safe: it is confined to the Fusion call chain and config surface. There is no keeper-wide coupling or constraint-hell introduced.

## Issues

### P0 — `stop_reason_label` duplicates existing SSOT (`lib/fusion/fusion_oas.ml:12`)

`fusion_oas.ml` introduces a new `stop_reason_label` that matches `Agent_sdk.Types.stop_reason`. The repo already has a canonical telemetry labeler in `Keeper_hooks_oas_types.stop_reason_to_label`, guarded by `test/test_stop_reason_label.ml`. The PR's labels drift from the SSOT:

| Variant | PR label | SSOT label |
|---|---|---|
| `ContextWindowExceeded` | `context_window_exceeded` | `model_context_window_exceeded` |
| `Unknown s` | `unknown:<escaped s>` | `unknown` |

This will split stop-reason time series and make dashboards/logs inconsistent. Fusion should call the existing SSOT (and extend it if the `Unknown` detail-preserving behavior is genuinely needed), not reimplement it.

### P0/P1 — `empty_response_detail` reimplements thinking-block redaction summary (`lib/fusion/fusion_oas.ml:33`)

The function manually counts `Thinking` / `RedactedThinking` blocks with mutable refs. The project already has `Keeper_hooks_oas_types.summarize_thinking_blocks` whose explicit purpose is "Summarize thinking block presence for logs/metrics without exposing raw thinking content." Reimplementing this in `fusion_oas.ml` is a clear SSOT violation and risks drifting from the redaction contract.

**Recommended fix:** Use `Keeper_hooks_oas_types.summarize_thinking_blocks` for the thinking subset, and a pure `List.fold_left` over `resp.content` for the remaining block-type tallies.

### P1 — Pure diagnostic function uses unnecessary mutable state (`lib/fusion/fusion_oas.ml:33`)

`empty_response_detail` holds eight `ref`s. The whole accumulation can be a single `List.fold_left` over `resp.content` with a record accumulator. Mutable state here is unnecessary and contradicts the project preference for immutable data structures in pure helpers.

### P1 — Dashboard parses OCaml `show` output with regex (`dashboard/src/lib/fusion-meta.ts:36`)

`normalizeFusionPanelReason` uses a regex to extract the payload from `Fusion_types.Empty_response "..."`. This is string matching where typed JSON/algebras exist. The backend already emits `reason_detail` as a plain string via `Fusion_oas.panel_failure_text`; if the dashboard still sees `show_panel_failure` literals, that wire contract should be fixed at the emitter, not papered over with more regex. At minimum this is accrued tech debt; the PR extends it rather than resolving it.

### P1 — Hardcoded truncation length (`lib/fusion/fusion_oas.ml:12`)

`stop_reason_label` for `Unknown s` truncates at `max_len = 80` with no named constant or explanation. This is a magic number.

### P2 — `panel_failure_of_yojson` should explain legacy round-trip (`lib/fusion_core/fusion_types.ml:71`)

The custom deserializer accepts the legacy bare `` `String "Empty_response" `` shape, which is good. But the `mli` comment does not document that this legacy shape is intentionally preserved, and there is no round-trip test for `Timeout`/`Provider_error` via the custom function. Add a test that serializes a non-legacy `Empty_response detail` and deserializes it back.

### P2 — `Invalid_max_output_tokens` mixed into `panel_failure` (`lib/fusion_core/fusion_types.ml:64`)

Adding a config-validation failure to `panel_failure` is defensive but slightly odd: the variant represents model/runtime outcomes, not static config errors. Since config validation already rejects non-positive values and this path is unreachable from validated presets, consider keeping it as a separate early-error path rather than a panel failure. Acceptable if you want the defense-in-depth; flagging for explicit decision.

## Recommended fixes

1. Replace `fusion_oas.stop_reason_label` with `Keeper_hooks_oas_types.stop_reason_to_label`. If preserving `Unknown` detail is required, extend the SSOT module (e.g., `stop_reason_to_label_detailed`) rather than forking it.
2. Use `Keeper_hooks_oas_types.summarize_thinking_blocks` for thinking metadata; rewrite `empty_response_detail` as a pure fold.
3. Name the 80-byte truncation constant or use `String_util.utf8_safe` / `utf8_prefix`.
4. Add round-trip JSON tests for `panel_failure_to_yojson` → `panel_failure_of_yojson` for all constructors, including legacy `Empty_response`.
5. Consider emitting a structured `empty_response` object (stop reason, counts, usage) in the sink so the dashboard can stop regex-parsing OCaml literals.

## Minor notes

- `test/test_masc_oas_bridge_cancel_boundary.ml` is updated to reference `Cancel_wall_bucket.of_wall wall` instead of literal bucket strings. This is a good SSOT cleanup but is unrelated to Fusion output budgets; it would be cleaner as a separate PR.
- The PR is split across five commits, including fixups (patch 3/5 updates match sites, patch 4/5 hardens error handling). Squash before merge.

## Verdict

The feature design is sound and the validation/option-threading is correct. The main blockers are the **SSOT violations** (`stop_reason_label` and thinking-block summary) and the **unnecessary mutable state** in a pure diagnostic function. Fix those and the PR is good to land.

jeong-sik:
[Adv Review] masc#22438 — re-review after author follow-up

Head `bd4980a` 기준으로 이전 blocker들이 기능적으로 수정되었으나, 일부는 paper-over이고 새로운 P2 concern이 있습니다.

P1 — `dashboard/src/lib/fusion-meta.ts:37`
`normalizeFusionPanelReason`이 여전히 OCaml `ppx_deriving show` 출력을 정규식으로 파싱합니다. `Empty_response` regex를 확장한 것은 당장 동작하지만, show 형식과의 coupling을 제거하지 않았습니다. backend `fusion_sink.ml`이 machine-readable `reason_code`/`reason_detail`을 분리해 내려주거나, dashboard가 JSON payload를 직접 읽도록 구조화해야 합니다. 현재 접근은 "no string matching" 기준에 위배됩니다.

P2 — `lib/fusion/fusion_oas.ml:33-78`
`empty_response_detail`이 8개의 `ref`로 mutable 누적 상태를 사용합니다. OCaml ^5.4 환경에서 가능하면 `List.fold_left` 등 immutable 누적으로 대체해 주세요.

P2 — `lib/fusion/fusion_oas.ml:27-29`
Unknown stop reason을 80바이트로 자를 때 `String.sub escaped 0 80`이 UTF-8 멀티바이트 문자를 중간에 자를 수 있습니다. `Uutf` 기반 truncation이나 grapheme-aware 처리를 고려하세요.

P2 — `lib/fusion/fusion_oas.ml:12-31`
`stop_reason_label`이 variant마다 하드코딩된 문자열을 반환합니다. `Agent_sdk.Types.stop_reason`에 이미 label/to_string이 있다면 그것을 사용하고, 없다면 별도 상수 모듈로 분리해 SSOT를 확보하세요.

P2 — `dashboard/src/lib/fusion-meta.ts:31-41`
새로 추가된 `Fusion_types.Invalid_max_output_tokens` 실패에 대한 normalize 처리가 없습니다. 이 실패가 dashboard로 전달되면 `normalizeProviderAttribution` fallback이 적용되어 `Provider 'unknown'` 형태로 잘못 표시될 수 있습니다.

CI의 `Build and Test`와 `TLA+ Model Checking`이 아직 진행 중이므로 최종 green 확인 후 merge 검토해 주세요.
```</subagent>

**Comment by jeong-sik** [comment]
Updated this branch for the current review sweep.\n\n- Preserved the pending fusion empty-response hardening diff as a fix commit.\n- Rebased onto current origin/main; one test patch was dropped because it is already upstream.\n- Local non-Dune validation passed: ocamlformat --check on touched OCaml files, ocamlc -stop-after parsing on touched OCaml files, git diff --check, python3 scripts/check_test_coverage.py.\n- Dashboard validation passed: pnpm lint, pnpm typecheck, pnpm vitest run --config vitest.config.ts src/lib/fusion-meta.test.ts --no-file-parallelism --maxWorkers=1.\n\nPer repo workflow/user direction, Dune/build validation is left to GitHub CI.

**Comment by jeong-sik** [comment]
[Adv Review]

Follow-up to https://github.com/jeong-sik/masc/pull/22438#issuecomment-4815789024 (head now `e43c8ca`). The 07:37 comment is a rebase/status update; the code changes pushed since the prior review do resolve several blockers.

Fixed from prior review:
- `lib/fusion/fusion_oas.ml:12` — `stop_reason_label` now aliases `Keeper_hooks_oas_types.stop_reason_to_label` instead of hardcoding variant labels.
- `lib/fusion/fusion_oas.ml:34-79` — `empty_response_detail` is now a pure `List.fold_left` over content plus `Keeper_hooks_oas_types.summarize_thinking_blocks`; the eight `ref`s are gone.
- `lib/fusion/fusion_oas.ml:198-204` — runtime `max_tokens` validation uses `Fusion_policy.valid_max_output_tokens` and returns `Invalid_max_output_tokens`, fixing the SSOT/error-shape duplication.
- `dashboard/src/lib/fusion-meta.ts:40-45` — `Invalid_max_output_tokens` now has a normalize branch (even though it is still regex-based).

Still open / new:

P1 — `dashboard/src/lib/fusion-meta.ts:31-46` still parses OCaml `ppx_deriving show` output with regex instead of using a typed wire format:

```ts
const providerMatch = trimmed.match(/^\(?\s*Fusion_types\.Provider_error\s+"([\s\S]*)"\s*\)?$/)
if (/^\(?\s*Fusion_types\.Timeout\s*\)?$/.test(trimmed)) return 'timeout'
const emptyMatch = trimmed.match(/^\(?\s*Fusion_types\.Empty_response(?:\s+"([\s\S]*)")?\s*\)?$/)
const invalidMaxOutputTokensMatch = trimmed.match(
  /^\(?\s*Fusion_types\.Invalid_max_output_tokens\s+(-?\d+)\s*\)?$/,
)
```

This is string matching where a typed/first-class alternative exists. The backend (`fusion_sink.ml`) should emit separate `reason_code` / `reason_detail` fields (or a JSON payload) so the dashboard does not couple itself to OCaml `show` formatting. The recent change extends the regex rather than removing the coupling.

P2 — `test/test_fusion_oas_error_detail.ml:1090` only tests legacy `"Empty_response"` deserialization. Add round-trip tests for `panel_failure_to_yojson` → `panel_failure_of_yojson` covering `Provider_error detail`, `Empty_response detail`, and `Invalid_max_output_tokens n` shapes. The current custom deserializer is correct, but the new constructors are untested.

CI note: `Build and Test`, `TLA+ Model Checking`, and several queued ratchets are still in progress, so final green should be confirmed before merge; not blocking the review on CI alone.

jeong-sik:
Follow-up after head b45b9de:

- P2 empty_response_detail refs: fixed in e43c8ca. Fusion now uses a pure List.fold_left for non-thinking block counts.
- P2 stop reason SSOT and UTF-8 truncation: fixed. Fusion delegates to Keeper_hooks_oas_types.stop_reason_to_label; Unknown maps to canonical unknown, so no string truncation path remains.
- P2 thinking summary SSOT: fixed. Fusion uses Keeper_hooks_oas_types.summarize_thinking_blocks and pins thinking_kind/thinking_chars without exposing content.
- P2 Invalid_max_output_tokens dashboard normalization: fixed. Legacy constructor fallback now normalizes it without provider attribution.
- P1 structured panel reasons: current backend path already emits reason_code/reason_detail from lib/fusion/fusion_sink.ml panel_meta and dashboard normalizePanelEntry reads reason_detail before legacy reason/error fields. The regex branch is now documented as a legacy compatibility fallback for old board/meta payloads and direct show-derived fixtures, not the primary wire contract.

Validation:
- scripts/opam-pin-external-deps.sh --install
- scripts/dune-local.sh build @test/fusion_core/runtest test/test_fusion_oas_error_detail.exe test/test_masc_oas_bridge_cancel_boundary.exe passed on e43c8ca.
- pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/lib/fusion-meta.test.ts passed on b45b9de.
- pnpm --dir dashboard typecheck passed on b45b9de.
- ocamlformat --check lib/fusion/fusion_oas.ml test/test_fusion_oas_error_detail.ml passed on b45b9de.
- git diff --check passed on b45b9de.
- Dune rerun after b45b9de was attempted but stopped while waiting behind unrelated global dune-local lock queue; no source failure observed.

CI attached for b45b9de and is pending.

jeong-sik:
Addressed the remaining concrete codec/test gap from the latest adversarial review.

Changes pushed to branch `codex/fusion-reasoning-empty-hardening-20260627`:
- `a41ce12` fixes `panel_failure_of_yojson` so the current ppx-style nullary `Timeout` JSON (`["Timeout"]`) decodes correctly.
- `7864585` adds legacy bare-string `Timeout` decode coverage without triggering warning 21.
- Added current-shape round-trip coverage for `Timeout`, `Provider_error`, `Empty_response detail`, and `Invalid_max_output_tokens`, plus an assertion that detailed `Empty_response` serializes as a tagged payload rather than the legacy string.

On the dashboard finding: the structured path is already the primary path in this PR. `fusion_sink.ml` emits `reason_code` / `reason_detail`, and the dashboard parser reads structured detail before falling back to legacy text parsing, so I left the fallback in place instead of deleting compatibility behavior.

Local fast validation:
- `ocamlformat --check lib/fusion_core/fusion_types.ml lib/fusion_core/fusion_types.mli test/test_fusion_oas_error_detail.ml`
- `ocamlc -stop-after parsing lib/fusion_core/fusion_types.ml test/test_fusion_oas_error_detail.ml`
- `ocamlc -stop-after parsing -intf lib/fusion_core/fusion_types.mli`
- `git diff --check`

CI note: per workflow preference I am not waiting on Dune/CI here. Also, GitHub currently reports `refs/heads/codex/fusion-reasoning-empty-hardening-20260627` at `7864585a07c7d9d0f8da957027b2b24d87b0d742`, while `refs/pull/22438/head` is still stale at `b45b9de3be045d4481709396c33b2532b48275ae`; I am leaving that as live GitHub metadata state rather than waiting.

PR #22483, opened 2026-06-27
fix(dashboard): hydrate compaction and memory recall views

jeong-sik:
## Summary
- Make the dashboard compaction snapshot reader parse only compaction-relevant runtime manifest events.
- Skip unrelated valid manifest events such as `memory_injected` instead of surfacing them as read errors.
- Keep corrupt JSON rows as read errors.
- Replace the Memory OS inspector's "recent recall/injection pending" placeholder with a real timeline from `memory_os_recall` prompt blocks in keeper turn records.

## Live evidence
- `GET /api/v1/keepers/rondo/compaction-snapshots` returned snapshots, but also `read_error_count=298` from valid unrelated manifest events like `memory_injected`, `memory_flushed`, `tool_lineage_recorded`, and `tool_surface_selected`.

## Verification
- `ocamlformat --check lib/server/server_dashboard_http_keeper_api.ml test/test_keeper_memory_os.ml`
- `scripts/ci/check-ignore-without-comment-diff.sh --base origin/main --head HEAD`
- `bash scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD`

## Local verification notes
- Skipped local Dune build per operator instruction; CI is the build authority.
- Dashboard targeted `vitest` and `typecheck` could not run locally because this worktree has no `dashboard/node_modules` (`vitest`/`tsc` not found). CI dashboard checks are the verification source for the frontend follow-up.

### DIFF
--- a/dashboard/src/components/memory-inspector.test.ts
+++ b/dashboard/src/components/memory-inspector.test.ts
@@ -293,7 +293,9 @@ describe('MemoryInspector — one-keeper scope (real data)', () => {
     expect(container.textContent).toContain('trace-a#7')
     expect(container.textContent).toContain('cccc2222dddd')
     expect(container.textContent).toContain('800B')
-    expect([...container.querySelectorAll('.mem-disclosure')].some(d => (d.textContent ?? '').includes('Phase 3'))).toBe(false)
+    const recallSection = [...container.querySelectorAll('.turn-sec')].find(sec => (sec.querySelector('h4')?.textContent ?? '') === '최근 회상 · 주입')
+    expect(recallSection?.querySelector('.mem-tl-row')?.textContent).toContain('cccc2222dddd')
+    expect(recallSection?.querySelector('.mem-disclosure')).toBeFalsy()
   })
 
   it('surfaces read_errors and renders honest disclosures for the unbacked pin section', async () => {
--- a/dashboard/src/components/memory-inspector.test.ts
+++ b/dashboard/src/components/memory-inspector.test.ts
@@ -13,6 +13,7 @@ import {
   memFmtBytes,
   memFmtTok,
   promptBlockMeta,
+  recentMemoryRecallInjections,
   sortMemoryFactsForReview,
   type MemoryKeeper,
 } from './memory-inspector'
@@ -28,8 +29,8 @@ afterEach(() => {
 })
 
 // A turn-records payload exercising the two real-data sections (composition,
-// facts), the episode-backed 압축 section, and read_errors. The unbacked
-// sections (핀 / 회상) render disclosures regardless of payload.
+// facts), the episode-backed 압축 section, recall-block timeline, and read_errors.
+// The unbacked pin section renders a disclosure regardless of payload.
 function turnRecordsPayload() {
   return {
     keeper: 'masc-improver',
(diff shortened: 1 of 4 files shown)


jeong-sik:
[Adv Review] head `ca047793`: good direction, but the current filter is too broad and can hide manifest drift.

The PR fixes a real dashboard bug: valid non-compaction manifest rows should not become `read_errors` in the compaction snapshot panel. However, `read_runtime_manifest_tail_rows` now does this before calling `Keeper_runtime_manifest.of_json`:

- read raw `event` string,
- if it is not `event_bus_correlated` or `context_compacted`, skip the row,
- never run the typed manifest parser for that row.

That suppresses the live noisy rows, but it also silently drops typo or future event names such as `context_compacted_v2`, `event_bus_correlatd`, or any producer drift where the row was intended to be compaction-relevant. Previously those became `unknown event` read errors via `Keeper_runtime_manifest.of_json`; now they disappear.

Please narrow the filter so only explicitly known unrelated event names are skipped. Options:

1. Add a local allowlist for the observed unrelated legacy events (`memory_injected`, `memory_flushed`, `tool_lineage_recorded`, `tool_surface_selected`, etc.) and keep unknown event strings as read errors.
2. Or add a typed/structured classifier in `Keeper_runtime_manifest` that returns `Relevant | Known_unrelated | Unknown of raw`, then have this dashboard use that SSOT.

The regression should include both cases: a known unrelated event is skipped with no read error, but an unknown/misspelled event string is still surfaced as a read error. Under the no-silent-failure rule, blanket skipping all non-compaction event strings is too permissive.

jeong-sik:
Rebased onto latest origin/main, including the agent_sdk 0.207.10 call-site adaptation from #22484.\n\nLocal validation (no local Dune build; Dune/build remains CI-owned):\n- git diff --check origin/main...HEAD\n- ocamlformat --check lib/server/server_dashboard_http_keeper_api.ml test/test_keeper_memory_os.ml\n- ocamlc -stop-after parsing lib/server/server_dashboard_http_keeper_api.ml test/test_keeper_memory_os.ml\n- python3 scripts/check_test_coverage.py\n- scripts/ocaml-structure-ratchet.sh\n- bash scripts/lint-spawn-bounded.sh\n- python3 scripts/ci/check-env-knob-classification.py --base origin/main --head HEAD\n- scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD\n\nPushed: 224ad691601

jeong-sik:
[Adv Re-review] current head `224ad69`: the previous silent-drop blocker still applies.

The rebase did not change the relevant logic. `read_runtime_manifest_tail_rows` still reads the raw `event` string and skips every value that is not `Event_bus_correlated` or `Context_compacted` before calling `Keeper_runtime_manifest.of_json` (`lib/server/server_dashboard_http_keeper_api.ml:400-431`). That means drifted/future/typo values like `context_compacted_v2` or `event_bus_correlatd` still disappear instead of surfacing as manifest read errors.

The added regression only proves the known unrelated `memory_injected` row is skipped (`test/test_keeper_memory_os.ml:4746-4792`). It does not cover the required negative case: an unknown or malformed event string should remain observable.

Please distinguish known-unrelated events from unknown event strings, preferably through a typed classifier in `Keeper_runtime_manifest`, and add coverage for both cases.

jeong-sik:
Current-head re-review on 1161a80: the silent-drop blocker still applies.

The relevant logic is still the same in `read_runtime_manifest_tail_rows`: it extracts the raw `event`, calls `compaction_snapshot_manifest_event_is_relevant`, and skips every non-`event_bus_correlated` / non-`context_compacted` row before `Keeper_runtime_manifest.of_json` sees it. Unknown/future/typo event names therefore still disappear instead of surfacing as manifest read errors.

This latest head also has CI blockers now: `Lint` and `Health` are failing while Build/Test and Dashboard were still in progress when I checked. Source blocker remains independent of CI: known-unrelated rows may be skipped, but unknown event strings need to remain observable through the canonical manifest parser or a typed classifier.

jeong-sik:
Updated on latest origin/main (f498b2102a3) and pushed 54088edcb59.\n\nVerification run locally without Dune per CI-build constraint:\n- git diff --check origin/main...HEAD\n- ocamlformat --check lib/server/server_dashboard_http_keeper_api.ml test/test_keeper_memory_os.ml\n- ocamlc -stop-after parsing lib/server/server_dashboard_http_keeper_api.ml test/test_keeper_memory_os.ml\n- python3 scripts/check_test_coverage.py\n- scripts/ocaml-structure-ratchet.sh\n- bash scripts/lint-spawn-bounded.sh\n- python3 scripts/ci/check-env-knob-classification.py --base origin/main --head HEAD\n- scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD\n- dashboard: pnpm install --offline --frozen-lockfile\n- dashboard: pnpm lint\n- dashboard: pnpm typecheck\n- dashboard: pnpm vitest run src/components/memory-inspector.test.ts\n\nOnly source change after rebase was scoping the recall-injection DOM assertion to the actual "최근 회상 · 주입" section so the separate Phase 3 compaction disclosure does not fail the recall timeline test.

jeong-sik:
#22483 merged while I was applying the silent-drop follow-up, so the fix is now split into draft PR #22490.

Follow-up summary:
- unknown drifted manifest events now fall through `Keeper_runtime_manifest.of_json` and surface as read errors;
- only known-unrelated events are skipped before parsing;
- added unknown-event coverage for `context_compacted_v2`.

I kept local validation to parser/format/ratchet/light checks and did not run local Dune build.

PR #22501, opened 2026-06-27
refactor: extract k2 helper and improve dashboard feed test infrastructure

jeong-sik:
## Summary
- sync the keeper metrics zero-fill expected constructor count from 227 to 229
- initialize the dashboard K2 feed test runtime before building keeper meta fixtures
- normalize decision-log evidence refs at the K2 feed projection boundary
- update memory-bank test rows to canonical parser-required fields

## Verification
- scripts/dune-local.sh exec test/test_otel_zero_fill.exe
- scripts/dune-local.sh exec test/test_dashboard_k2_feeds.exe
- git diff --check

PR #22502, opened 2026-06-27
fix(ci): restore dashboard K2 feed focused test

jeong-sik:
## Summary
- initialize the dashboard K2 feed test runtime before building keeper meta fixtures
- normalize decision-log evidence refs at the K2 feed projection boundary
- update memory-bank test rows to canonical parser-required fields

## Verification
- scripts/dune-local.sh exec test/test_dashboard_k2_feeds.exe
- git diff --check

### DIFF
--- a/lib/dashboard/dashboard_http_keeper_feeds.ml
+++ b/lib/dashboard/dashboard_http_keeper_feeds.ml
@@ -350,6 +350,9 @@ type decision_event = {
   evidence_refs : string list;
 }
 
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #22501, which now includes both the zero-fill count fix and this dashboard K2 feed fix. Keeping one combined CI repair PR avoids each narrow PR failing on the other inherited main regression.

PR #22513, opened 2026-06-28
perf(http): build response headers in reverse

jeong-sik:
## Summary
- build Http_server_eio response headers directly in Httpun's reversed representation
- preserve before/content/after/tail header order without the prior @ concat chain
- add a regression test for segment preservation and header order

## Verification
- git diff --check
- opam exec -- ocamlformat --check lib/http_server_eio.ml test/test_http_server_eio.ml

Full build/test is intentionally left to GitHub CI.

### DIFF
--- a/lib/http_server_eio.ml
+++ b/lib/http_server_eio.ml
@@ -155,14 +155,21 @@ module Response = struct
   let html_content_type = "text/html; charset=utf-8"
   let json_content_type = "application/json; charset=utf-8"
 
+  let rev_prepend_headers headers acc =
+    List.fold_left (fun acc header -> header :: acc) acc headers
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head review @05f921010c1b: no P0/P1 source blocker found in the narrow response-header performance change.

The new reverse-accumulation path preserves the intended segment order (`before_headers`, content headers, `after_headers`, `tail_headers`) before passing through `Httpun.Headers.of_rev_list`, and the added regression checks both order and lookup for all segments. I do not see hardcoded paths, env sprawl, stringly error classification, or silent fallback in this diff.

Readiness remains CI-bound: the PR is draft/BLOCKED and checks are still queued/in-progress.

PR #22517, opened 2026-06-28
feat: add git/github keywords to Execute tool search alias

anyang-keepers:
Add 'github clone commit push PR' to the tool_execute search alias in keeper_agent_tool_surface.ml so that keepers can discover git/github operations through tool search.

Closes: task-1529

### DIFF
--- a/lib/keeper/keeper_agent_tool_surface.ml
+++ b/lib/keeper/keeper_agent_tool_surface.ml
@@ -210,7 +210,7 @@ let tool_search_alias_entries =
   ; "tool_write_file", "파일 쓰기 저장 생성 덮어쓰기"
   ; "tool_search_files", "코드 검색 소스코드 grep rg 패턴 찾기 파일"
   ; ( "tool_execute"
-    , "명령어 실행 쉘 빌드 테스트 run dune build check compile compiles code git github clone commit push PR \
+    , "명령어 실행 쉘 빌드 테스트 run dune build check compile compiles code git github gh \
        status log diff" )
   ; ( "keeper_memory_search"
     , "기억 검색 대화 이전 메시지 memory previous earlier user said recall deployment" )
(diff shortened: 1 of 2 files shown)


jeong-sik:
Updated current head to `af1e0c98e02` to keep the Execute search alias policy-scoped.

Changes:
- kept the intended GitHub discoverability by adding `github` / `gh` terms;
- removed `clone`, `commit`, `push`, and `PR` from the Execute alias string so tool search does not bias PR/repo mutation workflows into Execute ahead of the existing policy gates;
- extended the existing `test_execute_aliases_exclude_repo_pr_workflow` coverage to catch `PR`, `push`, and `commit` regressions.

Local validation:
- `ocamlformat --check lib/keeper/keeper_agent_tool_surface.ml test/test_keeper_topk_llm.ml`
- `git diff --check`

Attempted focused `scripts/dune-local.sh exec test/test_keeper_topk_llm.exe`, but stopped before build because another session held `/tmp/me-dune-local.lock`; CI is the build authority for this pushed head.

[근거] `gh pr diff 22517 --repo jeong-sik/masc --patch`; `rg "test_execute_aliases_exclude_repo_pr_workflow|tool_search_alias_entries" -n lib test`; 확인일시 2026-06-28T12:55:00+09:00; 신뢰도 High.

jeong-sik:
Current-head review @ `af1e0c98e02aac05551f777e9f9e47b8478069f4`: no P0/P1 source blocker found.

This avoids the overclaim from the earlier Execute wording approach: the production change only adds `github` / `gh` to the `tool_execute` search aliases, and the regression test explicitly keeps broader repo-workflow/mutation tokens (`commit`, `push`, `PR`, draft/pull/create intent) out of the Execute alias set. That makes the behavior a discoverability nudge for shell/GitHub CLI requests, not a promise that Execute owns PR workflow semantics.

One non-blocking cleanup: the PR body still says it adds `github clone commit push PR`, but the actual diff intentionally does not add `clone`, `commit`, `push`, or `PR`. Updating the body would keep the review trail aligned with the safer implementation.

CI is still in progress/queued on this head; build remains CI-owned.

PR #22497, opened 2026-06-27
fix(keeper): clear no-progress pause after direct success

jeong-sik:
## Summary
- clear stale no_progress_loop pause/blocker after a successful direct keeper_msg turn
- reset the no-progress latch, registry failure state, turn failures, and live registry meta before final direct-turn meta write
- add focused coverage for the paused no_progress direct-success recovery path

## Verification
- git diff --check
- ocamlformat --check lib/keeper/keeper_turn.ml lib/keeper/keeper_turn.mli test/test_keeper_auto_pause_blocker_persist_12683.ml
- build/test: CI-owned per operator instruction

PR #22518, opened 2026-06-28
fix(keeper): persist direct resume before side effects

jeong-sik:
Follow-up to merged #22497.

This closes the remaining direct no-progress resume ordering gap by committing the registry resume state before emitting Operator_resume or resetting no-progress/livelock side effects. If registry persistence fails, the helper returns the original paused meta and leaves the no-progress latch, failure reason, phase, and livelock state intact.

Validation:
- ocamlformat --check lib/keeper/keeper_turn.ml test/test_keeper_auto_pause_blocker_persist_12683.ml
- ocamlc -stop-after parsing lib/keeper/keeper_turn.ml
- ocamlc -stop-after parsing test/test_keeper_auto_pause_blocker_persist_12683.ml

Per queue policy, local Dune was not run; CI is the build authority.

### DIFF
--- a/lib/keeper/keeper_turn.ml
+++ b/lib/keeper/keeper_turn.ml
@@ -107,6 +107,12 @@ let has_direct_success_no_progress_pause ~(config : Workspace.config)
       || Keeper_no_progress_loop_detector.is_latched ~keeper_name:meta.name
       || has_no_progress_failure_reason ~base_path:config.base_path meta.name)
 
+let clear_no_progress_meta_blocker (meta : keeper_meta) =
+  match meta.runtime.last_blocker with
+  | Some { Keeper_meta_contract.klass = Keeper_meta_contract.No_progress_loop; _ } ->
+    Keeper_meta_contract.map_runtime (fun rt -> { rt with last_blocker = None }) meta
+  | _ -> meta
+
 let persist_direct_success_no_progress_resume ~base_path (resumed_meta : keeper_meta) =
   match
     Keeper_registry.update_entry
@@ -124,12 +130,13 @@ let persist_direct_success_no_progress_resume ~base_path (resumed_meta : keeper_
         ; turn_consecutive_failures = 0
         })
   with
-  | Ok () -> ()
+  | Ok () -> Ok ()
   | Error err ->
     Log.Keeper.warn
       "%s: direct keeper_msg success could not persist no_progress resume state: %s"
       resumed_meta.name
-      (Keeper_registry.registry_entry_validation_error_to_string err)
+      (Keeper_registry.registry_entry_validation_error_to_string err);
+    Error err
 
 let clear_direct_success_no_progress_pause
       ~(config : Workspace.config)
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head review @ `a32b83a81fde71b864248aa582ecf258608fab4a`: the #22497 persistence-ordering blocker is fixed. No P0/P1 source blocker found in this follow-up.

The critical change is that `clear_direct_success_no_progress_pause` now builds the unpaused meta, attempts `persist_direct_success_no_progress_resume`, and returns the original paused `meta` on persistence failure before dispatching `Operator_resume` or resetting the no-progress/livelock side effects. The new regression covers that failure path: detector latch stays set, livelock state remains, registry phase stays `Paused`, registry meta remains paused, and the no-progress failure reason remains.

The success path still clears only the no-progress blocker from meta before persistence, preserves unrelated blockers, and then resets the detector/livelock state after a successful registry write. `git diff --check origin/main...origin/pr-22518` is clean locally. Checks are attached but queued, so CI remains the build authority.

jeong-sik:
BLOCKED current-head review @ `a32b83a81fde71b864248aa582ecf258608fab4a` (GitHub would not allow formal CHANGES_REQUESTED on my own PR): still blocked.

The persistence-before-side-effects direction is right, and the new `persist_direct_success_no_progress_resume` path now correctly refuses to return an unpaused meta when the registry write itself fails. But the recovery is still not atomic across the registry meta and FSM state.

Blocker: `lib/keeper/keeper_turn.ml:165-177` persists `meta.paused=false`, clears `last_failure_reason`, then calls `Keeper_registry.dispatch_event_and_log Operator_resume`. If that dispatch returns `Error`, the code only logs and continues to reset the no-progress detector/livelock state and returns `resumed_meta`. That can leave the registry with unpaused metadata while the state-machine phase/conditions did not accept the resume. For a Keeper pause/recovery path, that is a silent partial success: the metadata SSOT and FSM SSOT disagree.

The tests do not close this hole. The main success cases in `test/test_keeper_auto_pause_blocker_persist_12683.ml:335-378` and `:407-455` call `Keeper_registry.register` with a paused meta, but `register` derives a Running FSM entry from default conditions (`keeper_registry_setup.ml:395-403`). Because same-phase events are accepted, those tests can pass without proving `Operator_resume` actually resumes a `Paused` entry. The only test that seeds a true Paused registry entry is the persistence-failure case (`:484-556`), so there is still no success-path assertion that a real Paused no-progress keeper leaves Paused and clears `operator_paused`.

Required fix: make failed `Operator_resume` fail closed from this helper, or make the meta/FSM recovery a single validated registry transition. Also add a success-path test seeded as a real Paused entry with `operator_paused=true`, then assert both `meta.paused=false` and `get_phase <> Paused` / `operator_paused=false` after direct success.

[근거] `git diff origin/main...refs/remotes/pr/22518 -- lib/keeper/keeper_turn.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`, `keeper_registry_setup.ml:395-403`, `keeper_state_machine.apply_event` same-phase behavior; 확인일시 2026-06-28T13:08:17+09:00; 신뢰도 High.

PR #22520, opened 2026-06-28
fix(dashboard): clean compaction snapshot hydration

jeong-sik:
## Summary

- Stop treating normal runtime manifest events (`tool_lineage_recorded`, `tool_surface_selected`, `cascade_routed`) as compaction snapshot scan errors.
- Keep hydrated durable compaction snapshots in overlay-local state as well as the global signal, keyed by keeper, so a successful API response cannot render as an empty drawer.
- Add a live-shaped dashboard regression where durable snapshot items have nullable token counts and scan diagnostics.

## Evidence

- Live probe before patch: `GET /api/v1/keepers/rondo/compaction-snapshots?limit=5` returned `count: 2` with `context_compacted` items, but also `read_error_count: 47`; read errors included normal manifest events like `tool_lineage_recorded`, `tool_surface_selected`, and `cascade_routed`.
- `ocamlformat --check lib/keeper_registry/keeper_runtime_manifest_types.ml test/test_keeper_memory_os.ml`
- `git diff --check`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty false`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/keeper-workspace/keeper-workspace-rail.test.ts src/api/dashboard.test.ts` (`2 passed`, `90 passed`)

## Not Run

- `scripts/dune-local.sh build test/test_keeper_memory_os.exe` was attempted once, but it waited on `/tmp/me-dune-local.lock`; the lock holder was another focused Dune build in `.worktrees/event-bus-owner-thread-20260628`. Per operator instruction, full/build validation is left to CI.

### DIFF
--- a/dashboard/src/components/keeper-workspace/compaction-inspector-overlay.ts
+++ b/dashboard/src/components/keeper-workspace/compaction-inspector-overlay.ts
@@ -126,7 +126,13 @@ export function CompactionInspectorOverlay({
   keeper: Keeper
   onClose: () => void
 }): VNode {
-  const events = keeperCompactionSnapshots(keeper.name)
+  const globalEvents = keeperCompactionSnapshots(keeper.name)
+  const [hydratedState, setHydratedState] = useState<{ keeperName: string; events: CompactionSnapshot[] }>({
+    keeperName: keeper.name,
+    events: [],
+  })
+  const hydratedEvents = hydratedState.keeperName === keeper.name ? hydratedState.events : []
+  const events = globalEvents.length > 0 ? globalEvents : hydratedEvents
   const [idx, setIdx] = useState(0)
   const [side, setSide] = useState<'before' | 'after'>('after')
   const [loadState, setLoadState] = useState<CompactionSnapshotLoadState>({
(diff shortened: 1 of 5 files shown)


jeong-sik:
Current-head review @ `cc862bca79ed9761ca867b6fba5d581a51eb099f`: no P0/P1 source blocker found.

The patch stays scoped to the compaction snapshot surface: runtime-manifest scan noise is reduced by classifying the three observed non-compaction manifest events as known-unrelated, and the dashboard overlay now keeps a local hydrated snapshot fallback so a successful backend payload cannot render as an empty drawer while the global signal catches up. The changed `hydrateCompactionSnapshots` return value has only the intended overlay caller plus existing tests in this branch.

CI is attached but queued/pending, so merge readiness remains CI-gated.

[근거] `gh pr diff 22520 --repo jeong-sik/masc --patch`; `git grep -n "hydrateCompactionSnapshots" origin/pr-22520 -- dashboard/src test lib`; `gh pr checks 22520 --repo jeong-sik/masc --json name,state,bucket,link`; 확인일시 2026-06-28T13:02:45+09:00; 신뢰도 High.

PR #22500, opened 2026-06-27
fix(keeper): try next provider after no-progress accept reject

jeong-sik:
## Summary
- continue the runtime provider candidate loop when a provider returns a response that fails the accept contract with typed no-progress recovery hints
- treat passive-only turns with no current task and no active goals as idle/no-work receipts instead of operator-blocking completion-contract failures
- keep active task/goal passive-only receipts blocked, and keep explicit runtime blockers authoritative over benign idle receipts
- avoid immediate no-progress latch on scheduled no-work budget exhaustion unless the world observation has an actionable signal
- normalize no-progress / idle safety-pause blocker summaries so legacy persisted `manual pause applied` / raw latch details do not surface as fake provider/runtime failures in dashboard surfaces

## Why
Live keepers can surface accept_rejected / no_usable_progress with response_shape=empty as a terminal keeper request failure even when other provider candidates are available. Separately, a scheduled no-work/passive-only turn could be promoted into alert_exhausted / pause_human and `attention_reason=passive_only`, which made Play look like it immediately paused on a serious runtime error. Persisted no-progress blockers could also keep showing stale "manual pause" wording after the underlying classification changed.

## Validation
- rebased onto `origin/main` `0b1432a368c`
- `git diff --check`
- `ocamlformat --check` on touched OCaml files
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty false`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/lib/keeper-attention-labels.test.ts src/components/overview/overview.test.ts src/components/keeper-detail-alert-strip.test.ts src/components/fleet-health-panel.test.ts src/components/fleet-telemetry-utils.test.ts src/components/ide/ide-keeper-work-panel.test.ts src/lib/keeper-runtime-display.test.ts src/components/mission-agent-cards.test.ts` (288 tests)
- earlier focused Dune suite before the final rebase passed:
  - `scripts/dune-local.sh build test/test_no_progress_loop_detector.exe test/test_keeper_terminal_reason_typed.exe test/test_keeper_runtime_trust_snapshot.exe test/test_server_dashboard_composite_attention.exe test/test_keeper_auto_pause_blocker_persist_12683.exe test/test_keeper_status_bridge.exe`
[...]

PR #22523, opened 2026-06-28
fix(dashboard): match passive-only completion label

jeong-sik:
## Summary

- Fix the merged-main Lint failure from #22500 by matching the typed `Keeper_completion_contract_result_label.t` constructor.
- Replace unqualified `Contract_passive_only` patterns with `Completion_contract_result.Passive_only` in the composite execution classifiers.

## Evidence

- CI failure source: #22500 Lint failed in `lib/server/server_dashboard_http_composite_claims.ml` with `There is no constructor Contract_passive_only within type Completion_contract_result.t`.
- `ocamlformat --check lib/server/server_dashboard_http_composite_claims.ml`
- `git diff --check`

## Not Run

- Focused Dune compile was attempted in the #22500 worktree, but `scripts/dune-local.sh` refused because an unrelated bare `dune build .` process is running outside the wrapper. Per operator instruction, build validation is left to CI.

### DIFF
--- a/lib/server/server_dashboard_http_composite_claims.ml
+++ b/lib/server/server_dashboard_http_composite_claims.ml
@@ -348,7 +348,7 @@ let keeper_activation_readiness_json = Server_dashboard_fleet_readiness.keeper_a
 
 let composite_execution_passive_only_without_work_scope execution =
   match completion_contract_result_of_execution execution with
-  | Some Contract_passive_only ->
+  | Some Completion_contract_result.Passive_only ->
     Option.is_none (json_string "current_task_id" execution)
     && Json_util.json_string_list_member "goal_ids" execution = []
   | Some _ | None -> false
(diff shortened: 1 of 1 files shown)


jeong-sik:
Current-head review @ `231592c1fa006266e593edaa0b7e0d963ea164e0`: no P0/P1 source blocker found in this diff.

The change is a narrow typed-constructor repair in `server_dashboard_http_composite_claims.ml`: the passive-only checks now match `Completion_contract_result.Passive_only`, consistent with the other constructors in the same match. The special case for passive-only without work scope is unchanged, and the blocked/unsatisfied reason paths still route through `Completion_contract_result.to_string` rather than adding a new string classifier.

`git diff --check origin/main...refs/remotes/pr/22523` is clean. CI is still queued and mergeability is not established yet, so readiness still depends on GitHub checks completing green.

[근거] command: `gh pr view 22523 --repo jeong-sik/masc --json headRefOid,mergeStateStatus,isDraft,statusCheckRollup,title,url`; local PR ref `refs/remotes/pr/22523` inspected at `server_dashboard_http_composite_claims.ml`; 확인일시 2026-06-28 13:29 KST; 신뢰도 High.

PR #22530, opened 2026-06-28
fix(keeper): surface HITL-blocked keepers on supervisor sweep (silent chat hang)

jeong-sik:
## Symptom

Operator reports: 대화창에서 응답대기중으로 가만히 멈춰있음 (chat stalls silently as "awaiting response") after a turn transitions into HITL approval.

## Root cause (code)

A keeper blocked on a pending HITL approval shows up to the operator only as a **silent chat stall**:
- `keeper_cycle_decision` (`keeper_world_observation.ml`) returns `blocked Approval_pending` → `should_run = false`, so the keeper emits no further chat/observation turn.
- `record_skip_reasons` stamps `last_skip_observation`, but that field is **not surfaced** to dashboard/health, so the operator cannot tell *why* the chat is awaiting a response.
- The keeper only resumes when the operator resolves the approval, or `approval_janitor` expires it at the 30m (`max_wait_s = 1800.0`) mark.

This is **intended** behavior (don't run new turns while an approval is pending), but the silence is the bug — the operator sees a hung chat with no indication that an approval is outstanding.

## Fix

Emit a per-sweep `Log.Keeper.warn` in `sweep_and_recover` naming each keeper with an outstanding HITL approval, so the terminal log (and downstream consumers) make the root cause visible on every sweep instead of a silent hang.

Observability-only; **no behavior change** to the approval/resume/expiry path. Complements the existing `last_skip_observation` surface (#10962) by making it visible at the supervisor loop.

## Open question for the operator

This PR makes the hang *visible* but assumes the underlying cause is "operator hasn't approved yet". If the operator **did** approve and the chat still stalled, that's a different bug in the resolve → emit path (`resolve_with_policy` → `Eio.Promise.resolve` → keeper resume → chat emit) — flag it and I'll trace that path. The terminal log from this PR's warn (presence/absence of "blocked on pending HITL approval") disambiguates the two scenarios immediately.


jeong-sik:
Reviewed current head `fa133198d6e3898e20d93ee1d17d393eee9cdae7`.

No P0/P1 source blocker found in the current diff. The change is observability-only in `lib/keeper/keeper_supervisor.ml`: it names keepers that already have a pending HITL approval during the supervisor sweep and does not change approval submission, resolution, expiry, or resume behavior.

Non-blocking caveat: there is no focused regression around the supervisor log emission, so this is still CI/draft-bound. I am not blocking on that because the behavior path is unchanged and the PR explicitly wants per-sweep visibility rather than a state transition change.

jeong-sik:
Current-head adversarial pass: no P0/P1 source blocker found.

Reviewed the current head `fa133198d6e3898e20d93ee1d17d393eee9cdae7`. The sweep addition is visibility-only. The new helper path does not add a fail-closed dependency to supervisor recovery: `Keeper_meta_store.keeper_names` delegates to `persisted_keeper_names`, which returns `[]` after logging/metric on list errors, and `Keeper_approval_queue.has_pending_for_keeper` is an atomic pending-map fold. `git diff --check` is clean.

Remaining gate is not source-review: PR is still draft/BLOCKED and GitHub checks are queued.

[근거] 명령: `gh pr view 22530 --repo jeong-sik/masc --json headRefOid,mergeStateStatus,isDraft,statusCheckRollup`, `git diff --unified=80 origin/main...refs/remotes/pr/22530 -- lib/keeper/keeper_supervisor.ml`, `sed -n 1,95p lib/keeper/keeper_meta_store.ml`, `sed -n 1080,1125p lib/keeper/keeper_approval_queue.ml`, `git diff --check origin/main...refs/remotes/pr/22530`; 확인일시: 2026-06-28 13:26 KST; 신뢰도: High

jeong-sik:
Updated current head to `ca690f82994013daa92485a1ef2cdc48bcd94701`.

The previous coverage failure is addressed with a focused supervisor regression in `test/test_keeper_supervisor.ml`: it seeds a pending HITL approval, runs `sweep_and_recover`, and asserts the actual `Log.Ring` Keeper warning is emitted. The sweep warning now includes the pending approval count, and `pending_hitl_approval_keeper_names` keeps the persisted-keeper filtering explicit.

Local validation passed:
- `ocamlformat --check lib/keeper/keeper_supervisor.ml lib/keeper/keeper_supervisor.mli test/test_keeper_supervisor.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_supervisor.ml lib/keeper/keeper_supervisor.mli test/test_keeper_supervisor.ml`
- `git diff --check`

Focused Dune execution was not completed locally because the shared `/tmp/me-dune-local.lock` was held by another worktree build; CI remains the build authority. Current GitHub checks for the refreshed head are queued, including `Test Coverage Check`.

[근거] 명령: `gh pr view 22530 --repo jeong-sik/masc --json headRefOid,isDraft,mergeable,statusCheckRollup`, `gh pr checks 22530 --repo jeong-sik/masc --json name,state,bucket,link,workflow`, local validation commands above; 확인일시: 2026-06-28 13:38 KST; 신뢰도: High

jeong-sik:
Reviewed current head `ca690f82994013daa92485a1ef2cdc48bcd94701`.

No P0/P1 source blocker found. The diff stays observability-only: `sweep_and_recover` enumerates persisted keeper names with pending HITL approvals and logs the keeper plus pending count; it does not resolve, expire, restart, or otherwise change approval behavior. The added tests cover both filtering to persisted keepers and preserving the pending approval after the visibility sweep.

Minor non-blocking note: because the count is read after the initial `has_pending_for_keeper` filter, a concurrent approval resolution could produce a transient `0 pending` warning. That would be noisy rather than behavior-changing. `git diff --check` is clean. CI is queued, so this is source review only until checks complete.

jeong-sik:
Current-head adversarial pass on the new head: no P0/P1 source blocker found.

Reviewed head `305e26de47a536e507b9fffe613d8a826ede4209` after the force-push. The supervisor now surfaces persisted keeper names that have pending HITL approvals during each sweep, without changing approval resolution/resume behavior. The added tests cover filtering out a non-persisted pending keeper and verifying that the sweep emits the warning while leaving the approval pending. `git diff --check` is clean.

Non-blocking scope note: `Keeper_approval_queue.has_pending_for_keeper` is still name-scoped over the in-process pending map, not explicitly `base_path`-scoped. I did not block on that because this follows the existing queue API and the live server is single-base-path, but if this process starts hosting multiple workspace configs with same keeper names, this visibility helper should filter by `entry.audit_base_path = config.base_path`.

Remaining gate: PR is draft/BLOCKED and checks are queued.

[근거] `gh pr view 22530 --repo jeong-sik/masc --json headRefOid,mergeStateStatus,isDraft,statusCheckRollup,files` + `git diff --unified=160 origin/main...refs/remotes/pr/22530 -- lib/keeper/keeper_supervisor.ml lib/keeper/keeper_supervisor.mli test/test_keeper_supervisor.ml` + `git grep -n -C 12 "pending_hitl_approval_keeper_names\|sweep_warns_for_pending\|filters_persisted" refs/remotes/pr/22530 -- lib/keeper/keeper_supervisor.ml test/test_keeper_supervisor.ml` + `git diff --check origin/main...refs/remotes/pr/22530`; 확인일시 2026-06-28 13:42 KST; 신뢰도 High.

PR #22532, opened 2026-06-28
perf(dashboard): bound governance ring append without rescans

jeong-sik:
## Summary\n- Replace the dashboard governance rejection ring's list-only state with an immutable state record carrying both events and count.\n- Avoid `List.length` on every `record_tool_skipped` append; under-cap appends are now O(1), and tail copying happens only once the ring is full.\n- Add a regression test proving the oldest event is pruned and the ring size remains capped without duplicating the production cap constant in the test.\n\n## Why\nThis matches the same bounded hot-path concern as the SSE replay buffer review item: a bounded buffer should not rescan the full list for every append when the authoritative size can be carried in the state.\n\n## Local checks\n- `opam exec -- ocamlformat --check lib/dashboard/dashboard_governance_metrics.ml lib/dashboard/dashboard_governance_metrics.mli test/test_dashboard_governance_metrics.ml`\n- `git diff --check origin/main...HEAD`\n\nBuild/test authority remains GitHub CI.

### DIFF
--- a/lib/dashboard/dashboard_governance_metrics.ml
+++ b/lib/dashboard/dashboard_governance_metrics.ml
@@ -32,8 +32,17 @@ let max_ring_size = 43200
     the largest dashboard window (720m). *)
 
 let ring_mu = Eio.Mutex.create ()
-let ring : rejection_event list ref = ref []
-(** Most recent first; truncated to [max_ring_size]. *)
+
+type rejection_ring = {
+  events : rejection_event list;
+  count : int;
+}
+
+let empty_ring = { events = []; count = 0 }
+
+let ring : rejection_ring ref = ref empty_ring
+(** [events] is most recent first; truncated to [max_ring_size].
+    [count] is the linearized length so appends do not rescan the full ring. *)
 
 let record_failure_callback_label =
   "dashboard_governance_tool_skipped_record"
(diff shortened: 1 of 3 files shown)


jeong-sik:
Reviewed current head `5a4d037a2c6baa663b540699558442c5c072ab39`.

No P0/P1 source blocker found. The ring state now carries the length alongside the immutable event list, so under-cap appends avoid the previous full `List.length` rescan. The cap path preserves newest-first ordering and prunes the oldest event; the new test covers that through `inject_for_testing` using the same append helper. `List.take` is available in the local OCaml 5.4.1 Stdlib, so this is not a missing-API issue.

Non-blocking note: once the ring is full, each additional append still copies `max_ring_size - 1` list cells under the mutex. That is bounded and no worse asymptotically than the old over-cap prune path, but it is not a true O(1) circular buffer. Given the PR scope is removing the repeated under-cap length scan, I am not blocking on that. CI is still queued.

jeong-sik:
Current-head adversarial pass: no P0/P1 source blocker found.

Reviewed head `5a4d037a2c6baa663b540699558442c5c072ab39`. The governance rejection ring now carries a bounded count under the existing append mutex, so appends avoid the old `List.length` full-ring scan. The full-ring path still walks the retained tail with `List.take`, but that is bounded to `max_ring_size - 1`; I did not find an unbounded growth or stale-count path. `inject_for_testing` now uses the production append path, and the added cap regression exercises oldest-event pruning. Local OCaml 5.4.1 also exposes `List.take`, so this is not relying on a missing stdlib helper. `git diff --check` is clean.

Remaining gate: PR is draft/BLOCKED and checks are queued.

[근거] `gh pr view 22532 --repo jeong-sik/masc --json headRefOid,mergeStateStatus,isDraft,statusCheckRollup` + `git diff --unified=120 origin/main...refs/remotes/pr/22532 -- lib/dashboard/dashboard_governance_metrics.ml lib/dashboard/dashboard_governance_metrics.mli test/test_dashboard_governance_metrics.ml` + `git diff --check origin/main...refs/remotes/pr/22532` + local OCaml 5.4.1 `#show List.take`; 확인일시 2026-06-28 13:33 KST; 신뢰도 High.

PR #22537, opened 2026-06-28
fix(keeper): clear no-progress resume latch

jeong-sik:
## Summary
- clear the no-progress latch/blocker when dashboard boot/Play resumes a paused keeper
- clear the same latch when masc_keeper_up resumes a paused keeper
- make reaction-ledger passive_only attention scope-aware so no-work passive-only receipts do not keep the fleet degraded

## Evidence
- live health before patch showed durable paused keepers with last_blocker=no_progress_loop and threshold=1
- scripts/dune-local.sh build test/test_keeper_reaction_ledger.exe
- scripts/dune-local.sh build test/test_keeper_pause_silent_failure_source.exe
- ./_build/default/test/test_keeper_reaction_ledger.exe
- ./_build/default/test/test_keeper_pause_silent_failure_source.exe
- git diff --check
- after rebase onto origin/main: git diff --check origin/main...HEAD

## Notes
- Rebase after local validation was clean. Re-running the focused builds after rebase was stopped because another worktree held the shared Dune lock; CI is the build authority for the final result.

### DIFF
--- a/lib/keeper/keeper_reaction_ledger.ml
+++ b/lib/keeper/keeper_reaction_ledger.ml
@@ -380,6 +380,13 @@ let bool_field name json =
   | _ -> false
 ;;
 
+let string_list_field name json =
+  list_field name json
+  |> List.filter_map (function
+    | `String value -> Some value
+    | _ -> None)
+;;
+
 let nested_string_field outer inner json =
   match assoc_field outer json with
   | Some nested -> string_field inner nested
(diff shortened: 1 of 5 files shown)


jeong-sik:
Reviewed current head `2aba14eb54866bb7ae4fbb26eaeaa0fef898193b`.

No P0/P1 source blocker found. `clear_for_operator_resume` is scoped to the no-progress latch: it resets the in-memory detector, clears registry failure only when the stored provider-runtime-error code is `no_progress_loop`, and clears meta `last_blocker` only when the blocker class is `No_progress_loop`. The dashboard resume path and `masc_keeper_up` path both call it only on actual unpause/resume paths.

The reaction-ledger change also looks bounded: `passive_only` completion-contract attention is suppressed only when the reaction has no `current_task_id` and no `goal_ids`, so work-scoped passive-only receipts can still degrade the summary.

Non-blocking test gap: the resume-path coverage is source-shape based; a direct behavioral test for clearing registry failure/meta blocker would be stronger. `git diff --check` is clean. CI is queued, so this is source review only until checks complete.

PR #22553, opened 2026-06-28
[codex] fix build path API visibility

jeong-sik:
## Summary

Fixes the local build break caused by public interface drift around path helper APIs.

- Exposes `Config_dir_resolver.absolute_path` because `server_routes_http_sidecar_paths` calls it through the module interface.
- Normalizes the duplicated `Keeper_sandbox_runtime.normalize_base_path_for_hash` interface declaration on current `main` so the public contract is documented once and `base_path_hash` remains documented separately.
- Makes the relative base-path hash test compare against `Config_dir_resolver.current_working_dir ()`, avoiding macOS `/var` vs `/private/var` canonicalization drift.

## Root Cause

The `Config_dir_resolver.absolute_path` implementation existed, but the `.mli` hid it from callers. During rebase, current `main` already exposed `normalize_base_path_for_hash`, but with a duplicated declaration around `base_path_hash`; this PR keeps one public declaration and preserves the hash API docs.

## Validation

Before rebasing onto latest `origin/main`:

- `scripts/dune-local.sh build lib/config_dir_resolver lib/keeper lib/server test/test_keeper_tool_policy_paths.exe test/test_keeper_sandbox_read_backend.exe`
- `_build/default/test/test_keeper_tool_policy_paths.exe`
- `_build/default/test/test_keeper_sandbox_read_backend.exe`

After rebasing onto latest `origin/main`, the same focused wrapper build was attempted, but `scripts/dune-local.sh` refused to start because unrelated bare Dune processes were active outside the wrapper (`dune build bin/env_knob_catalog.exe`). CI should be treated as the post-rebase build authority for this draft.

### DIFF
--- a/lib/config_dir_resolver/config_dir_resolver.mli
+++ b/lib/config_dir_resolver/config_dir_resolver.mli
@@ -57,10 +57,6 @@ val current_working_dir : unit -> string
 (** Current working directory, falling back to an absolute host root when the
     process cwd has been deleted. *)
 
-val absolute_path : string -> string
-(** [absolute_path path] resolves a relative [path] against
-    {!current_working_dir}; absolute paths are returned unchanged. *)
-
 val base_path_or_cwd : unit -> string
 (** [MASC_BASE_PATH] from host config, or {!current_working_dir} when unset. *)
 
--- a/lib/config_dir_resolver/config_dir_resolver.mli
+++ b/lib/config_dir_resolver/config_dir_resolver.mli
@@ -57,6 +57,10 @@ val current_working_dir : unit -> string
 (** Current working directory, falling back to an absolute host root when the
     process cwd has been deleted. *)
 
+val absolute_path : string -> string
+(** [absolute_path path] resolves a relative [path] against
+    {!current_working_dir}; absolute paths are returned unchanged. *)
+
 val base_path_or_cwd : unit -> string
 (** [MASC_BASE_PATH] from host config, or {!current_working_dir} when unset. *)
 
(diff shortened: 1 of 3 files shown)


jeong-sik:
Reviewed current head `f135de08bb445f7f57d62261119796af2e404c26`.

This exposes `Config_dir_resolver.absolute_path` in the interface and aligns the relative base-path hash test with `Config_dir_resolver.current_working_dir`. The `keeper_sandbox_runtime.mli` duplicate declaration is also reduced to a single signature. I do not see a source blocker. Remaining gates are draft/CI state.

jeong-sik:
Pushed a small follow-up at dae35375799: removed the duplicate `Config_dir_resolver.absolute_path` declaration introduced near `current_working_dir`. Current main already exposes the API in the env-introspection section, so the PR no longer duplicates the interface while keeping the sandbox hash test fix and `Keeper_sandbox_runtime` signature cleanup.\n\nValidation: `ocamlformat --check lib/config_dir_resolver/config_dir_resolver.mli`; `git diff --check`. Full build remains CI-owned per workflow.

jeong-sik:
Superseding my previous note after re-reviewing current head `dae353757996b23e5b990cec40d9b51962e40f87`.

The current diff no longer exposes `Config_dir_resolver.absolute_path`; it is now just the `keeper_sandbox_runtime.mli` duplicate declaration/doc cleanup plus the relative-base-path test expectation aligned with `Config_dir_resolver.current_working_dir`. `git diff --check` is clean. I do not see a current source blocker.

Current gate: the visible check set includes cancelled stale jobs, so CI still needs a fresh/complete green attachment before merge.

PR #22570, opened 2026-06-28
fix: Resolve additional keeper boundary bugs for offline states

jeong-sik:
## Description
Fixes two additional cases where the UI incorrectly relied on the active keeper runtime list (`keepers.value`) to infer domain facts, resulting in silent failures when a keeper is offline.

### Bugs Fixed
1. **Goal Lead Keeper Badge Disappearance**: `GoalCard` used `keeperByName` to find the lead keeper object. If the keeper was offline, the badge completely disappeared. Fixed by changing the function to `leadNameForGoal` and using the name as SSOT, gracefully degrading to an `(Offline)` state.
2. **Missing Task Trace (Trajectory) Fetch**: `task-detail-state.ts` failed to fetch the tool call trajectory of an offline keeper because it relied on `keeper?.name ?? null`. Traces are persistent on disk and should be fetched regardless of the process state. Fixed by relying on `task.assignee_kind === 'keeper'` and using `task.assignee` as the trajectory name fallback.

### Related
Extracted from PR #22568 to keep logical fixes separate.

### DIFF
--- a/dashboard/src/components/goals/task-detail-state.ts
+++ b/dashboard/src/components/goals/task-detail-state.ts
@@ -211,11 +211,12 @@ async function loadActivity(task: Task): Promise<void> {
   try {
     const keeper = findKeeper(task.assignee)
     const timelineName = keeper?.agent_name ?? task.assignee
-    const trajectoryName = keeper?.name ?? null
+    const isKeeper = task.assignee_kind === 'keeper' || keeper !== null
+    const trajectoryName = isKeeper ? task.assignee : null
 
     const [timeline, trajectory] = await Promise.all([
       fetchAgentTimeline(timelineName, 24, 200),
-      trajectoryName ? fetchKeeperTrajectory(trajectoryName, 100) : Promise.resolve(null),
+      trajectoryName ? fetchKeeperTrajectory(trajectoryName, 100).catch(() => null) : Promise.resolve(null),
     ])
 
     if (activityFetchToken.value !== token) return
@@ -235,7 +236,7 @@ export function hasActivityTab(task: Task): boolean {
 
 /** Whether the assignee is a keeper (affects tool call visibility). */
 export function isKeeperAssignee(task: Task): boolean {
-  return task.assignee ? findKeeper(task.assignee) !== null : false
+  return task.assignee_kind === 'keeper' || (task.assignee ? findKeeper(task.assignee) !== null : false)
 }
 
 // -- Goal relationship (keeper's active goals) ----------------------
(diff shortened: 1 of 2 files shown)


anyang-keepers (approved):
## Review

**Checked:**
- Diff scope: 2 files, +11/-10, focused offline-state fix ✓
- `task-detail-state.ts`:
  - `isKeeper` now uses `task.assignee_kind === 'keeper'` as primary signal, with `findKeeper` fallback ✓
  - `trajectoryName` uses `task.assignee` directly when `isKeeper`, avoiding null when keeper is offline ✓
  - `fetchKeeperTrajectory` now catches errors gracefully (`.catch(() => null)`) for offline keepers ✓
  - `isKeeperAssignee` same dual-signal pattern ✓
- `work.ts`:
  - `leadKeeperForGoal` → `leadNameForGoal`, returns `string | undefined` instead of `Keeper | undefined` ✓
  - Avoids null object dereference when lead keeper is offline ✓
  - `GoalCard` passes name string directly to `openKeeperWorkspace` ✓
- Fixes boundary bug where offline keepers caused trajectory fetch failures and null derefs ✓
- No breaking changes: all changes are defensive ✓

**No BLOCKs.** Change is safe and correctly scoped. Fixes offline keeper boundary bugs in dashboard task detail and goal card components.

jeong-sik:
Blocking on current head `e44b415`.

`task-detail-state.ts` correctly treats offline keeper identity as `task.assignee_kind === "keeper" || keeper !== null`, but `work.ts` does not apply that same typed boundary for the goal lead badge. The new `leadNameForGoal` fallback counts every task assignee for the goal:

```ts
for (const t of goalTasks) {
  if (t.assignee) counts.set(t.assignee, (counts.get(t.assignee) ?? 0) + 1)
}
```

Then `GoalCard` renders whatever name wins as a `KeeperBadge` and marks it `offline` when `keeperByName(leadName)` is absent. That means a goal whose most common task assignee is `operator`, a human reviewer, or any non-keeper agent is now displayed as an offline keeper lead. This reintroduces the exact domain-boundary problem the PR is trying to fix: runtime presence is no longer the SSOT, but the replacement is an untyped assignee-name heuristic.

Please filter the fallback to keeper-typed assignees (`assignee_kind === "keeper"` or another explicit keeper SSOT), and add regression coverage for both cases:

- offline keeper task assignee still renders the lead badge;
- non-keeper/human assignee does not render as an offline keeper.

CI also agrees this needs tests: the `check` job failed with `Test Coverage Check Failed` / `Added 11 lines to covered code paths but no test files changed.` `Dashboard` is red as well, with logs unavailable while the workflow is still running.

Issue #22567, opened 2026-06-28
Keeper runtime follow-ups from 2026-06-28 stale-turn audit


jeong-sik:
## Summary

Follow-up tracker for the non-primary findings from the 2026-06-28 Keeper runtime audit.

Primary fix is in draft PR #22566: restarted keepers inherited stale `last_turn_ts`, then `assess_stale_run` killed the fresh fiber before one configured stale window elapsed.

This issue records the adjacent findings that should not be silently dropped while #22566 handles the P0 root cause.

## Evidence

- Symptom: live runtime health reported `keeper_fleet_safety.status=degraded`, `keeper_fibers=1`, and low running fiber margin while multiple keepers were blocked/dead from `stale_turn_timeout(idle_turn(...))`.
- Reproduction condition: parse the last 24h `<RUNTIME_ROOT>/logs/system_log_2026-06-27.jsonl` and `system_log_2026-06-28.jsonl` with cutoff `2026-06-27T10:34:00Z`, then inspect exact `rondo`, `nick0cave`, and `sangsu` restart sequences.
- Attempted: #22566 fixes the restart/lifetime false-positive by gating idle stale classification on the current supervised `started_at` lifetime without mutating `last_turn_ts`.

## Follow-up Classes

1. Fleet reaction capacity collapse after multiple dead/paused keepers.
2. Self-preservation suppression needs re-evaluation after stale-turn false positives are removed.
3. Pending no-progress recovery stimuli can remain attached to non-executable keepers.
4. Bootstrap event queue restore/ack semantics need proof across restart.
5. Restarted keepers can consume bootstrap without completing a real turn before the next sweep.
6. Pause/autoboot/dead bootability interaction is difficult to reason about in health output.
7. Active task owner without executable keeper fiber needs explicit release/quarantine semantics.
8. Anti-rationalization empty evaluator text approving by liveness is a silent-failure risk.
9. Memory librarian empty response/timeout/global slot skips need current-head validation.
10. External prompt resolving empty and falling back to binary prompt is an SSOT risk.
11. Compaction manifest unknown event errors may indicate live binary/schema drift.
12. `source_entries_arg` currently ignores invalid source shape instead of typed rejection.
13. Correction pipeline can still be invalid after deterministic fixes.
14. SSE unknown-session noise should be separated from Keeper runtime warnings.
15. Dashboard token mismatch fallback noise should be separated from Keeper runtime warnings.
[...]

PR #22561, opened 2026-06-28
Classify accept_rejected keeper blockers as contract failures

jeong-sik:
## Summary
- Reclassify persisted `Provider_runtime_error` records that wrap typed masc `accept_rejected` payloads as `completion_contract_violation` blockers.
- Preserve typed empty-turn summaries so dashboard/status no longer presents no-usable-progress as a provider runtime catch-all.
- Clear stale no-progress recovery stimuli on operator resume across live queue, durable pending/in-flight snapshots, and reaction ledger pending state.
- Exclude credentialed non-keeper MCP clients from keeper fleet-safety task-owner blockers, while preserving the degraded signal for unbound keeper assignees.
- Fix the latest-main Dashboard duplicate `inspect_turn_budget` label regression and re-export the completion-contract failure constructor in `keeper_registry_types.mli`.
- Preserve no-progress resume/drop failures so queue cleanup errors are not silently collapsed during keepalive or dashboard resume paths.

## Live evidence
- `/health?full=1` reported runnable keeper fibers, but `rondo` still had `pending_no_progress_recovery_count=1`.
- `/Users/dancer/me/.masc/keepers/rondo/event-queue.json` contained `post_id=no-progress-loop:rondo`.
- The matching reaction-ledger row was `stimulus:3c0d0c6f71739aecb8096f1faab1da93` with no reaction row, explaining why Play could clear the latch while the dashboard still surfaced stale recovery pressure.
- Later `/health?full=1` had `keeper_reaction_ledger.status=ok` and `pending_no_progress_recovery_count=0`, while fleet safety remained degraded only because `task-1547` was assigned to credentialed external client `codex-mcp-client` with no keeper binding.

## Verification
- `ocamlformat --check lib/keeper/keeper_status_bridge_blocker.ml test/test_blocker_class_exhaustiveness.ml`
- `ocamlformat --check lib/keeper_runtime/keeper_event_queue.ml lib/keeper_runtime/keeper_event_queue.mli lib/keeper_runtime/keeper_event_queue_persistence.ml lib/keeper_runtime/keeper_event_queue_persistence.mli lib/keeper/keeper_registry_event_queue.ml lib/keeper/keeper_registry_event_queue.mli lib/keeper/keeper_unified_turn_no_progress.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`
[...]

PR #22575, opened 2026-06-28
fix(health): classify non-keeper task owners as advisory

jeong-sik:
## Summary

- Split active task owner fleet-safety scan into blocking keeper-owner rows and advisory non-keeper client-owner rows.
- Add `Keeper_identity.is_keeper_principal_agent_name` using canonical keeper aliases plus strict dictionary-generated nicknames, so structured client names such as `codex-mcp-client` are not folded into fake keeper names.
- Surface `non_keeper_active_task_owner_count` / `non_keeper_active_task_owners` in health JSON while keeping keeper-fleet status unblocked for client-owned work.

## Runtime Evidence

- Live `/health?full=1` on 2026-06-28 reported:
  - `keeper_fleet_safety.status=degraded`
  - `blocker=active_task_owner_without_executable_fiber`
  - `task_id=task-1547`
  - `agent_name=codex-mcp-client`
  - `reason=no_keeper_binding`
- Backlog record confirms `task-1547` is a claimed client/operator task owned by `codex-mcp-client`, not a configured keeper fiber.

## Evidence

- Local static checks:
  - `ocamlformat --check lib/keeper/keeper_identity.ml lib/keeper/keeper_identity.mli lib/server/server_routes_http_runtime_fleet_scan.ml test/test_server_runtime_bootstrap.ml`
  - `git diff --check`
  - `git diff --cached --check`
- Local Dune build/test intentionally not run: operator requested CI as the OCaml build authority for this investigation.

Follow-up for #22567.

### DIFF
--- a/lib/keeper/keeper_identity.ml
+++ b/lib/keeper/keeper_identity.ml
@@ -75,6 +75,12 @@ let canonical_keeper_name_from_agent_name agent_name =
       else
         None
 
+let is_keeper_principal_agent_name agent_name =
+  let trimmed = String.trim agent_name in
+  is_keeper_agent_alias trimmed
+  || (Nickname.is_dictionary_generated_nickname trimmed
+      && Option.is_some (canonical_keeper_name_from_agent_name trimmed))
+
 (** Phase A F5 (2026-04-27): single source of truth for the
     ["keeper-<name>"] prefix pattern.  Two call sites used to embed
     [String.sub trimmed 0 7 = "keeper-"] manually; both now go through
--- a/lib/keeper/keeper_identity.mli
+++ b/lib/keeper/keeper_identity.mli
@@ -9,6 +9,12 @@ val generate_trace_id : ?now:float -> unit -> string
 val keeper_name_from_agent_name : string -> string option
 val is_keeper_agent_alias : string -> bool
 val canonical_keeper_name_from_agent_name : string -> string option
+val is_keeper_principal_agent_name : string -> bool
+(** [is_keeper_principal_agent_name name] returns true for task-owner
+    principals that should be treated as keeper-owned work: canonical
+    [keeper-<name>-agent] aliases and dictionary-generated keeper nicknames.
+    It intentionally rejects arbitrary three-part client names such as
+    [codex-mcp-client]. *)
 val canonical_keeper_name : string -> string option
 
 val strip_keeper_prefix : string -> string option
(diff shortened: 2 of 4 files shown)


jeong-sik:
Non-keeper classification — direction right; two concerns: string-name identity and overlap with #22561.

Splitting blocking keeper-owner rows from advisory non-keeper-client rows is the right call for fleet safety (the `task-1547` / `codex-mcp-client` evidence). Consolidating the previously inlined `String.sub trimmed 0 7 = "keeper-"` checks into one `is_keeper_principal_agent_name` is a real SSOT improvement.

Two concerns:

**1. Identity is decided by name string, not a typed principal.**
`is_keeper_principal_agent_name : string -> bool` resolves identity from `agent_name` via canonical aliases + `Nickname.is_dictionary_generated_nickname`. That is a classifier over strings: a future client named `keeper-codex-mcp-client` (or any alias collision) is misclassified, and the "dictionary-generated nicknames" set is an implicit allowlist that can drift from the actual keeper roster. The structural question: does the task-assignment / fleet-scan path carry a typed agent principal (or a keeper-binding record) this could match on, instead of re-deriving identity from the display name? If `agent_name` is the only signal at this boundary the name-classifier is unavoidable — but then the allowlist should be derived from the configured keeper roster (the SSOT for "who is a keeper") rather than a parallel alias/nickname dictionary, so the two cannot disagree.

**2. Overlap with #22561.**
#22561's body already lists "Exclude credentialed non-keeper MCP clients from keeper fleet-safety task-owner blockers" with the same `codex-mcp-client` / `task-1547` evidence, and both PRs touch `server_routes_http_runtime_fleet_scan.ml` + `test_server_runtime_bootstrap.ml`. Which subsumes which — is this the identity slice split out of #22561, or a parallel implementation of the same fleet-safety gate? If parallel, they conflict on merge and one should rebase onto the other so "non-keeper owner" has one definition.

Minor: the `String.sub trimmed 0 7` magic 7 for `"keeper-"` — prefer a named prefix constant or `String.length "keeper-"` so the 7 is not bare.

Build deferred to CI.

jeong-sik:
Closing as superseded by current `main` and less safe to rebase.

During conflict resolution I compared the incoming branch against latest `origin/main`. Main already has the safer behavior: credentialed external non-keeper clients are reported as advisory rows without becoming keeper-fleet blockers. This branch tries to add a broader fallback using `Keeper_identity.is_keeper_principal_agent_name`, which would classify arbitrary non-keeper-looking task owners as advisory. That is weaker than the current mainline guard and matches the earlier review concern about string-name identity.

So I am not force-pushing a conflict resolution here. The correct path is to keep the mainline credentialed-client implementation and avoid reintroducing a broader name-shape heuristic.

PR #22588, opened 2026-06-28
fix(memory): back off failed librarian attempts

jeong-sik:
## Summary
- add typed librarian extraction errors for empty provider output, timeout/transport failure, prompt render failure, missing clock, and fact-upsert failure
- separate semantic structured success from completed non-success attempts with cadence_record_attempt
- defer the next librarian attempt after unstructured fallback or typed provider/append failures, while keeping skipped work such as provider-slot busy immediately due
- update retry/cadence tests and Memory OS fallback coverage

Tracker: #22567

## Verification
- ocamlformat --check lib/keeper/keeper_librarian_runtime.ml lib/keeper/keeper_librarian_runtime.mli test/test_keeper_librarian_retry.ml test/test_keeper_memory_os.ml
- git diff --check
- focused local Dune wrapper was attempted but refused because an existing bare process is running: dune build @fmt --root .

CI is the build authority for this branch.

PR #22590, opened 2026-06-28
fix(keeper): stop no-progress pause self-stimulus

jeong-sik:
## Summary
- stop `mark_loop_detected` from enqueueing a synthetic `No_progress_recovery` stimulus when the no-progress safety latch pauses a keeper
- keep the latch as an operator-resume/manual-pause boundary instead of a no-data self-wake
- add a regression asserting successful no-progress pause leaves the keeper event queue empty

## Root Cause
Current live logs after runtime start showed repeated consumption/restoration of `no-progress-loop:*` stimuli, e.g. executor/taskmaster consuming `No_progress_recovery` and then restoring one pending stimulus again on later turns.

The producer was `mark_loop_detected`: it enqueued a no-progress recovery stimulus before attempting the manual pause. On a successful pause, that no-data stimulus remained durable and could force another empty turn after resume via the event-queue override.

## Validation
- `ocamlformat --check lib/keeper/keeper_unified_turn_no_progress.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_auto_pause_blocker_persist_12683.exe` was attempted but refused because bare Dune processes are already holding the shared Dune lane:
  - `55271 55262 dune build --root . --no-print-directory lib/`

### DIFF
--- a/lib/keeper/keeper_unified_turn_no_progress.ml
+++ b/lib/keeper/keeper_unified_turn_no_progress.ml
@@ -1,19 +1,8 @@
-(** No-progress loop recovery helpers for the unified keeper turn. *)
+(** No-progress loop pause helpers for the unified keeper turn. *)
 
 let failure_reason_code = "no_progress_loop"
 
 let recovery_post_id ~keeper_name = "no-progress-loop:" ^ keeper_name
-
-(* RFC-0020: the no-progress recovery stimulus carries no data — the consumer
-   only branches on the kind. streak/threshold are recorded in the failure
-   reason / blocker detail by the caller, not in the stimulus payload. *)
-let recovery_stimulus ~now ~keeper_name =
-  { Keeper_event_queue.post_id = recovery_post_id ~keeper_name
-  ; urgency = Keeper_event_queue.Immediate
-  ; arrived_at = now
-  ; payload = Keeper_event_queue.No_progress_recovery
-  }
-;;
 
 let mark_loop_detected ~(config : Workspace.config) meta ~streak ~threshold =
   let detail =
@@ -36,25 +25,6 @@ let mark_loop_detected ~(config : Workspace.config) meta ~streak ~threshold =
     ~base_path:config.base_path
     meta.Keeper_meta_contract.name
     (Some failure_reason);
-  let stimulus =
-    recovery_stimulus ~now:(Time_compat.now ()) ~keeper_name:meta.name
-  in
-  Keeper_registry_event_queue.enqueue
-    ~base_path:config.base_path
-    meta.name
-    stimulus;
-  (try
-     Keeper_reaction_ledger.record_event_queue_stimulus
-       ~base_path:config.base_path
-       ~keeper_name:meta.name
-       stimulus
-   with
-   | Eio.Cancel.Cancelled _ as exn -> raise exn
-   | exn ->
-     Log.Keeper.warn
-       "%s: failed to persist no-progress recovery stimulus in reaction ledger: %s"
-       meta.name
-         (Printexc.to_string exn));
   let blocked_meta =
     Keeper_meta_contract.map_runtime
     (fun rt ->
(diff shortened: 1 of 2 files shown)


jeong-sik:
Root-cause fix for the self-stimulus loop — good; one scope question on No_progress_recovery.

This is the right fix at the right layer: `mark_loop_detected` enqueued a synthetic `No_progress_recovery` stimulus before pausing, so on a successful pause that no-data stimulus stayed durable and forced another empty turn after resume via the event-queue override — a clean self-stimulus loop. Removing the enqueue and making the latch an operator-resume boundary (not a no-data self-wake) breaks the loop at its producer. The +28/-50 net-negative and the regression asserting the queue is empty after a successful pause are both good signals — less code, and the invariant is pinned.

One scope question: `No_progress_recovery` stimulus — is it only produced by `mark_loop_detected`, or do other legitimate recovery paths also enqueue it? If the latter, removing the producer here must not have also removed the legitimate recovery trigger elsewhere; confirm the recovery stimulus is still emitted by the path that should emit it (operator-initiated resume, or a real data-arrived signal), so this fix narrows to the loop-detected self-stimulus rather than disabling no-progress recovery entirely.

Build deferred to CI. This dovetails with #22561's stale-stimulus clearing and #22588's librarian backoff — the same keeper-stall family.

jeong-sik:
Current-head blocker for `c673a70d347b8122c49a32e2f4fcd119b34a6ce3`: CI Health fails on a real compile error in the new regression test.

`test/test_keeper_auto_pause_blocker_persist_12683.ml` now calls:

```ocaml
Masc.Keeper_event_queue.is_empty
  (Masc.Keeper_registry_event_queue.snapshot ~base_path:config.base_path keeper_name)
```

but `Keeper_event_queue` is not exposed as `Masc.Keeper_event_queue`, so CI fails with:

```text
File "test/test_keeper_auto_pause_blocker_persist_12683.ml", line 251, characters 13-36:
Error: Unbound module Masc.Keeper_event_queue
```

This is not an infrastructure failure. Existing tests use the internal module directly as `Keeper_event_queue` when the test target links the internal test deps, and use `Masc.Keeper_registry_event_queue.snapshot` for the public registry wrapper. Please either use `Keeper_event_queue.is_empty` in this test target, or expose a deliberate public helper if the empty-queue assertion should become part of the `Masc` API. I would avoid exporting `Keeper_event_queue` wholesale just to satisfy a focused regression.

The source direction still looks right: the self-stimulus producer is removed and the failed-pause wake path stays suppressed. Merge readiness is blocked until this compile error is fixed and CI reruns.

[근거] `gh pr checks 22590 --repo jeong-sik/masc --json name,state,bucket,link`; `gh api repos/jeong-sik/masc/actions/jobs/83907369188/logs`; `rg -n "Keeper_event_queue|event_queue|is_empty" lib test -g "*.ml" -g "*.mli"`; `git diff origin/main...refs/remotes/pr/22590 -- lib/keeper/keeper_unified_turn_no_progress.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`; checked 2026-06-28 22:19 KST; confidence High.

jeong-sik:
Pushed `3faec249f9f` to fix the CI compile blocker in the no-progress regression test.

Change: `test/test_keeper_auto_pause_blocker_persist_12683.ml` now calls the internal linked test dependency as `Keeper_event_queue.is_empty` instead of the non-exported `Masc.Keeper_event_queue.is_empty`. This keeps the assertion narrow and avoids expanding the public `Masc` API just for the test.

Local checks:
- `ocamlformat --check test/test_keeper_auto_pause_blocker_persist_12683.ml`
- `git diff --check`
- `ocamlc -stop-after parsing test/test_keeper_auto_pause_blocker_persist_12683.ml`

Focused Dune check was attempted with `scripts/dune-local.sh build test/test_keeper_auto_pause_blocker_persist_12683.exe`, but the wrapper refused because an external bare `dune build @fmt --root .` process is holding/bypassing the machine-wide Dune guard. Leaving full build authority to CI for this PR.

[근거] `gh pr checks 22590 --repo jeong-sik/masc --json name,state,bucket,link`; local commands listed above; checked 2026-06-28 22:00 KST; confidence High.

jeong-sik:
Rebased #22590 onto current `origin/main` and force-with-lease pushed `e30fb639259`.

Conflict resolution detail: current main had added `recovery_post_id` for operator-resume cleanup. The PR correctly removes the synthetic `No_progress_recovery` producer, so the merge keeps `recovery_post_id` for dropping stale queued stimuli but does not restore `recovery_stimulus`.

Local checks after rebase:
- `ocamlformat --check lib/keeper/keeper_unified_turn_no_progress.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`
- `ocamlc -stop-after parsing lib/keeper/keeper_unified_turn_no_progress.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`

Focused Dune wrapper is still blocked by the external bare `dune build @fmt --root .` process (`scripts/dune-local.sh build test/test_keeper_auto_pause_blocker_persist_12683.exe` exits 75), so CI remains the build authority.

[근거] `git rebase origin/main`; `git push --force-with-lease origin codex/no-progress-pause-no-recovery-20260628`; local commands listed above; checked 2026-06-28 22:03 KST; confidence High.

anyang-keepers (approved):
## Review

**Checked:**
- Diff scope: 2 files, +28/-50, net reduction, focused no-progress loop fix ✓
- `keeper_unified_turn_no_progress.ml`:
  - Removes `recovery_stimulus` function and its event-queue enqueue/ledger recording ✓
  - Stops self-waking behavior that caused infinite loop: pause-fail → wake → no-progress → pause-fail ✓
  - Now correctly marks loop detected and pauses without queuing a synthetic recovery stimulus ✓
  - Comment updated: "a latched keeper must not receive a synthetic no-progress recovery wake" ✓
- Tests: updated to verify pause behavior without recovery stimulus enqueue ✓
- No breaking changes: removes dead code path that was causing harm ✓

**No BLOCKs.** Change is safe and correctly scoped. Fixes infinite no-progress loop by removing self-stimulus wake that had no actionable data.

jeong-sik:
Entire CI matrix is red, not inherited — this PR breaks the build.

Main's own `CI` workflow is running; the only `failure` on main is `Branch Protection Watchdog`, which is a token false-positive, not a code defect. So the ~40 failing checks here (Build, Lint, Health, Keeper, Dashboard, audit-path-ssot, Boundary, Fragile-match, ...) are not main-inherited — this PR's diff is breaking the build. The earlier compile error in `test_keeper_auto_pause_blocker_persist_12683.ml` (the `Masc.K...` call) is the likely root: a compile failure in the test stanza cascades into every downstream check. Fix the test compilation and the rest should re-run; until then this is not mergeable.

Build deferred to CI (CI is currently red on this PR itself).

PR #22591, opened 2026-06-28
fix(keeper): retry heartbeat presence meta writes

jeong-sik:
## Summary
- route heartbeat presence meta writes through `write_meta_with_merge`
- use the existing `Keeper_meta_merge.heartbeat_fields_from_disk` merge so heartbeat writes retry CAS conflicts without stealing disk-owned pause/runtime fields
- add a structural guard covering the presence-sync writer

## Root Cause
Current runtime logs showed:

`write_meta failed (heartbeat): meta version conflict for executor: expected 31760, disk has 31758`

`sync_keeper_presence` still used plain `write_meta`, even though adjacent turn/heartbeat writers already use `write_meta_with_merge ~merge:Keeper_meta_merge.heartbeat_fields_from_disk`. A transient heartbeat CAS conflict should use the existing field-level retry path instead of surfacing as a failed heartbeat write.

## Validation
- `ocamlformat --check lib/keeper/keeper_heartbeat_loop_presence.ml test/test_keeper_pause_silent_failure_source.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_pause_silent_failure_source.exe` was attempted but refused because bare Dune processes are already holding the shared Dune lane:
  - `64408 64393 dune build --root . --no-print-directory`

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop_presence.ml
+++ b/lib/keeper/keeper_heartbeat_loop_presence.ml
@@ -152,7 +152,12 @@ let sync_keeper_presence
       ~labels:[ "keeper", meta_current.name ]
       ();
     note_turn_failures_preserved_after_heartbeat ~ctx ~meta:meta_current;
-    match write_meta ctx.config synced with
+    match
+      write_meta_with_merge
+        ~merge:Keeper_meta_merge.heartbeat_fields_from_disk
+        ctx.config
+        synced
+    with
     | Ok () -> synced
     | Error e ->
       Otel_metric_store.inc_counter
--- a/test/test_keeper_pause_silent_failure_source.ml
+++ b/test/test_keeper_pause_silent_failure_source.ml
@@ -32,6 +32,7 @@ let target_files =
 
 let status_detail_file = "lib/keeper/keeper_status_detail.ml"
 let keeper_up_update_file = "lib/keeper/keeper_turn_up_update.ml"
+let heartbeat_presence_file = "lib/keeper/keeper_heartbeat_loop_presence.ml"
 
 let metric_name = "Keeper_metrics.(to_string PausedStatePersistErrors)"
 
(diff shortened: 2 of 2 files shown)


jeong-sik:
Consistency fix — heartbeat presence reuses the existing CAS-merge path. Good.

Routing `sync_keeper_presence` through `write_meta_with_merge ~merge:Keeper_meta_merge.heartbeat_fields_from_disk` makes the heartbeat writer consistent with the adjacent turn/heartbeat writers that already use that path — so a transient CAS conflict ("meta version conflict for executor") retries via the field-level merge instead of surfacing as a failed heartbeat write. That reuses the existing SSOT merge function rather than adding a new retry, and avoids the heartbeat clobbering disk-owned pause/runtime fields on conflict. Surgical (+21/-1), root cause is the plain-`write_meta` inconsistency, and the structural guard test pins that the presence writer uses the cas-retry merge.

One nit: the test loads the source file and matches on it. If that is a `String.contains`-style check rather than an AST walk, a formatting or identifier rename could break it or pass vacuously. Confirming it is structural (or anchors on the function call, not surrounding text) keeps the guard honest.

Build deferred to CI.

anyang-keepers (approved):
## Review

**Checked:**
- Diff scope: 2 files, +17/-2, focused CAS retry fix ✓
- `keeper_heartbeat_loop_presence.ml`: `write_meta` → `write_meta_with_merge` with `Keeper_meta_merge.heartbeat_fields_from_disk` ✓
- This prevents silent overwrite of disk-owned heartbeat fields when concurrent presence updates race ✓
- Test `test_keeper_pause_silent_failure_source.ml`: new test verifies CAS retry merge is used and plain `write_meta` call is removed ✓
- No breaking changes: merge function preserves existing behavior while adding retry safety ✓

**No BLOCKs.** Change is safe and correctly scoped. Fixes silent failure mode where heartbeat presence meta writes could lose disk-owned fields under concurrent updates.

jeong-sik:
## 리뷰: #22591 — heartbeat presence CAS retry merge (좋음, 병합 준비 방향)

### 긍정
- **기존 모듈 재사용**: `write_meta_with_merge` + `Keeper_meta_merge.heartbeat_fields_from_disk`. /goal "이미 있는 모듈 다시 만드는지" — 재사용, 재구현 아님 ✅.
- **CAS conflict retry**: heartbeat writes가 CAS conflict 시 재시도(merge). silent drop 아님.
- test 추가(`test_heartbeat_presence_uses_cas_retry_merge`).
- +21/-1 — 작고 집중된 fix.

### 미세 (blocker 아님)
- CAS retry의 *최대 횟수 / 백오프* — `write_meta_with_merge` 내부 정책에 의존. 무한 retry 아니면 OK(메모리: retry livelock 패턴).

### 워크어라운드 게이트
해당 없음 — 기존 CAS merge 경로 재사용.

**병합 준비 방향** (CI green 시): 작고 안전한 fix, 기존 모듈 재사용.

---
_근거: `gh pr diff 22591` (heartbeat presence → write_meta_with_merge)._

PR #22592, opened 2026-06-28
fix(dashboard): avoid goal-loop refresh timeout clamp

jeong-sik:
## Summary
- make the goal-loop refresh timeout explicit at `interval * 0.8`
- expose the interval/timeout constants as a testable contract
- add a regression asserting the timeout remains below the interval so `Proactive_refresh` does not clamp and warn on startup

## Root Cause
Current runtime logs showed:

`goal_loop_status: timeout_s (10) >= interval_s (5), clamping to 4s`

`start_goal_loop_refresh_loop` used `Proactive_refresh.default_config ~interval_s:5.0`, whose default timeout is `10.0`. The generic clamp was correct, but this caller should not enter the warning path on every startup.

## Validation
- `ocamlformat --check lib/server/server_dashboard_http_goal_loop_broadcast.ml lib/server/server_dashboard_http_goal_loop_broadcast.mli test/test_goal_loop_broadcast.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_goal_loop_broadcast.exe` was attempted but refused because a bare Dune process is already holding the shared Dune lane:

### DIFF
--- a/lib/server/server_dashboard_http_goal_loop_broadcast.ml
+++ b/lib/server/server_dashboard_http_goal_loop_broadcast.ml
@@ -28,6 +28,7 @@ let goal_loop_broadcast_event_type = "goal_loop_status"
    TTL cache ([Dashboard_goal_loop.goal_loop_cache_ttl_s]); a shorter tick
    would only re-read the same cached value. *)
 let goal_loop_broadcast_interval_s = 5.0
+let goal_loop_broadcast_timeout_s = goal_loop_broadcast_interval_s *. 0.8
 
 (* Per-write volatiles excluded from the change fingerprint. The Python
    writer stamps a fresh [generated_at] on every write (and the OCaml read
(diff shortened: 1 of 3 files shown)


jeong-sik:
Caller config fix — correct; one magic-ratio nit.

This is the right fix: `start_goal_loop_refresh_loop` passed `interval_s:5.0` into `Proactive_refresh.default_config` whose timeout defaults to `10.0`, so the generic clamp (timeout >= interval -> clamp) fired a WARN on every startup. Making the caller set `timeout = interval * 0.8` explicitly keeps the clamp correct and stops the startup noise. Exposing the interval/timeout constants as a testable contract with a "timeout < interval" regression is good — it pins the invariant that caused the WARN.

One nit: the `0.8` ratio is a bare magic number. It's reasonable (headroom under the interval so a slow refresh finishes within the tick window), but a named constant or a one-line comment stating the intent would read better than an arbitrary fraction.

Build deferred to CI.

jeong-sik:
## 리뷰: #22592 — goal-loop refresh timeout clamp 회피 (좋음)

### 긍정
- timeout을 명시 상수 `goal_loop_broadcast_interval_s *. 0.8`로(timeout < interval 보장). 기존 silent clamp를 *관측 가능 상수 + startup warn*으로. /goal "노 silent" + CLAUDE.md "Cap/Cooldown" 시그니처 **회피** 방향(clamp 제거/명시).
- testable contract로 노출.

**병합 준비 방향** (CI green 시).

---
_근거: `gh pr diff 22592` (goal_loop_broadcast_timeout_s 명시)._

PR #22593, opened 2026-06-28
Surface librarian fallback facts in memory health

jeong-sik:
## Summary
- add per-keeper and total librarian fallback fact counts to keeper-memory-health
- emit backend-owned warning alerts for fallback diagnostic fact rows
- render the new count in the dashboard health table and summary strip

## Verification
- ocamlformat --check lib/server/server_dashboard_http_keeper_memory_health.ml test/test_server_dashboard_http_keeper_memory_health.ml
- git diff --check origin/main...HEAD
- env PATH=/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin:$PATH NODE_PATH=/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/memory/keeper-memory-health.test.ts

## Notes
- Local focused Dune build was skipped because scripts/dune-local.sh refused to run while a bare Dune process is holding the machine-wide lock; CI is the build authority for this slice.

PR #22594, opened 2026-06-28
fix(keeper): touch smart-idle liveness before sleep

jeong-sik:
## Summary
- Record smart-idle sleep admission before entering interruptible_sleep.
- Keep the existing timeout observation for the actual timeout outcome.
- Add a regression proving the pre-sleep touch keeps assess_stale_run from stamping Idle_turn while the keeper is intentionally idle.

## Runtime evidence
- Live post-start log at 2026-06-28T13:08:33Z emitted stale_keeper_broadcast for idealist and sangsu, followed by registry crash/error records for stale_turn_timeout(idle_turn(...)).
- Source gap: Skip_idle only touched last_turn_ts after the idle sleep timed out, allowing the supervisor to fire while the keeper was asleep.

## Validation
- ocamlformat --check lib/keeper/keeper_heartbeat_loop.ml lib/keeper/keeper_heartbeat_loop_observations.ml lib/keeper/keeper_heartbeat_loop_observations.mli test/test_keeper_idle_threshold_contract.ml
- scripts/dune-local.sh build test/test_keeper_idle_threshold_contract.exe attempted but refused because bare Dune processes are already running; CI remains build authority.

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -605,6 +605,9 @@ let run_smart_heartbeat_gate
     | Keeper_heartbeat_smart.Skip_idle next_time ->
       let wait = Float.max 1.0 (next_time -. Time_compat.now ()) in
       Log.Keeper.debug "smart heartbeat: skip (idle, next in %.1fs)" wait;
+      Observations.record_smart_idle_sleep_admission
+        ~base_path:config.base_path
+        ~keeper_name:meta_current.name;
       let jitter = wait *. 0.1 *. Random.float 1.0 in
       let outcome =
         Keeper_keepalive_signal.interruptible_sleep ~clock ~stop ~wakeup (wait +. jitter)
--- a/lib/keeper/keeper_heartbeat_loop_observations.ml
+++ b/lib/keeper/keeper_heartbeat_loop_observations.ml
@@ -16,6 +16,18 @@ let smart_idle_sleep_observation_reasons =
   [ "smart_heartbeat_skip_idle"; "idle_sleep_timeout" ]
 ;;
 
+let smart_idle_sleep_admission_reasons =
+  [ "smart_heartbeat_skip_idle"; "idle_sleep_admitted" ]
+;;
+
+let record_smart_idle_sleep_admission ~base_path ~keeper_name =
+  Keeper_registry.record_skip_reasons
+    ~base_path
+    keeper_name
+    ~reasons:smart_idle_sleep_admission_reasons;
+  Keeper_registry.touch_last_turn_ts ~base_path keeper_name
+;;
+
 let record_smart_idle_sleep_observation ~base_path ~keeper_name =
   Keeper_registry.record_skip_reasons
     ~base_path
--- a/lib/keeper/keeper_heartbeat_loop_observations.mli
+++ b/lib/keeper/keeper_heartbeat_loop_observations.mli
@@ -9,6 +9,13 @@ val record_provider_timeout_observation
 
 val smart_idle_sleep_observation_reasons : string list
 
+val smart_idle_sleep_admission_reasons : string list
+
+val record_smart_idle_sleep_admission
+  :  base_path:string
+  -> keeper_name:string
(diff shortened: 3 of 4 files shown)


jeong-sik:
Root fix for idle-sleep fake-death — good.

The root cause is clean: `Skip_idle` only touched `last_turn_ts` after the idle sleep timed out, so while the keeper was in `interruptible_sleep` the supervisor's `assess_stale_run` saw no recent liveness and stamped `Idle_turn` -> `stale_keeper_broadcast` + `stale_turn_timeout(idle_turn)` registry crash. Recording the `smart_idle_sleep_admission` before entering the sleep closes the gap — the supervisor sees the keeper is intentionally idle rather than stale. The regression proving the pre-sleep touch keeps `assess_stale_run` from stamping Idle_turn pins the invariant.

One question: for a long idle window, the admission timestamp itself ages. Does `assess_stale_run` treat the admission reason (`idle_sleep_admitted`) as "intentionally idle, do not stale-stamp regardless of age", or does it still re-stale if the sleep exceeds the stale threshold? If the latter, a keeper that idles longer than the threshold would still crash despite the touch — confirming the admission is a hard override on the stale check closes this.

Minor: the `smart_idle_sleep_admission_reasons` string list follows the existing observation-reason pattern, so that is consistent.

Build deferred to CI.

anyang-keepers (approved):
## Review

**Checked:**
- Diff scope: 4 files, +53/-1, focused liveness fix ✓
- `keeper_heartbeat_loop.ml`: before entering idle sleep, calls `record_smart_idle_sleep_admission` to touch last_turn_ts ✓
- `keeper_heartbeat_loop_observations.ml`/`mli`: adds `record_smart_idle_sleep_admission` and `smart_idle_sleep_admission_reasons` ✓
- This prevents a keeper from being marked stale while in smart-idle sleep admission window ✓
- Test `test_keeper_idle_threshold_contract.ml`: new test verifies last_turn_ts is refreshed before sleep when admission is recorded ✓
- No breaking changes: additive observation recording, existing idle logic unchanged ✓

**No BLOCKs.** Change is safe and correctly scoped. Fixes stale-keeper detection during smart-idle sleep by touching liveness timestamp before entering sleep.

jeong-sik:
Source review on current head `3578f2c6d0d426dd7b949cd4b3fd031a343eed8a`: no blocker found.

This is a narrow liveness fix. The existing timeout-side observation already touched `last_turn_ts` after an idle sleep completed; this PR adds an admission-side observation before the keeper enters smart-idle sleep. That matches the failure mode: while a keeper is intentionally parked in `Skip_idle`, the stale watchdog should not classify it as a stuck running turn just because no turn completed during the sleep window.

The implementation also keeps the reasons explicit (`idle_sleep_admitted` vs `idle_sleep_timeout`) instead of reusing one ambiguous skip reason. That gives operators a useful distinction: "entered idle sleep" versus "idle sleep timed out normally." The added test is focused on the invariant that matters here: `record_smart_idle_sleep_admission` advances `last_turn_ts`, records the admission reasons, and makes `assess_stale_run` return `None`.

CI is still mostly queued on this head, so this is source-only review state, not merge readiness.

[근거] `git diff origin/main...pr/22594`, `gh pr view 22594 --json headRefOid,state,isDraft`, `gh pr checks 22594 --json name,state,bucket,link`, 확인일시 2026-06-28 22:18 KST, 신뢰도 High.

jeong-sik:
## 리뷰: #22594 (smart-idle liveness) + #22593 (librarian fallback surface) — 둘 다 좋음

### #22594
- smart-idle sleep admission을 *sleep 전*에 기록(liveness). 관측 가능. keeper가 sleep 중에도 alive 추적. /goal "노 silent".

### #22593
- librarian fallback fact counts를 keeper-memory-health에 *surface*. backend-owned warning alerts. silent fallback을 관측 가능하게. /goal "노 silent failure" 직결.

**둘 다 병합 준비 방향** (CI green 시).

---
_근거: `gh pr diff 22594` (record_smart_idle_sleep_admission), `gh pr diff 22593` (memory-health fallback counts)._

PR #22595, opened 2026-06-28
fix(keeper): stop budget checkpoint self-accumulation

jeong-sik:
## Summary
- Prevent OAS turn-budget checkpoint notices from being persisted back into keeper assistant history/checkpoints/memory as model-authored output.
- Treat budget-exhausted/no-evidence scheduled turns as immediate no-progress latch candidates, and stop recording those checkpoints as success/health success.
- Deduplicate event-queue stimuli by durable identity so repeated bootstrap enqueue/load does not grow a keeper backlog when only arrived_at changes.

## Verification
- scripts/stringly-boundary-ratchet.sh --print
- scripts/silent-failure-ratchet.sh --print
- bash scripts/hardening-ratchet.sh --check fails on wildcard_silent_defaults +3; verified the same failure on clean origin/main (6d359ae462a), so not introduced by this branch

Build/test authority: CI, per request; no local Dune build was run.

### DIFF
--- a/lib/keeper/keeper_agent_run_finalize_response.ml
+++ b/lib/keeper/keeper_agent_run_finalize_response.ml
@@ -43,6 +43,10 @@ let finalize
     ~pre_dispatch_compaction_after_tokens
     ~raw_response_text
     () =
+  let budget_exhausted =
+    Keeper_agent_run_response_text.stop_reason_is_turn_budget_exhausted
+      result.stop_reason
+  in
   let reported_state_snapshot =
     reported_state_snapshot_from_checkpoint result.checkpoint
   in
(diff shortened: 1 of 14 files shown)


jeong-sik:
## 리뷰: #22595 — budget checkpoint self-accumulation bug fix (고쳐짐)

### 긍정
- budget checkpoint notice가 model-authored output으로 *재영속*(self-accumulation)되던 bug fix. `receipt_response_text_present_ref := String.trim response_text <> ""`(기존 `:= true`에서 정확화). 빈 response 체크로 self-accumulation 차단.
- `String.equal a.post_id b.post_id && a.urgency = b.urgency ...` — equality(중복 체크). boundary 아닌 typed equality. OK.

### 미세 (blocker 아님, 범위 밖)
- `receipt_response_text_present_ref` mutable ref는 **기존**(#22595가 도입 아님). /goal "가능하면 Immutable" — ref를 functional(return 전달)로 대체가 더 좋으나, 기존 아키텍처라 이 PR 범위 밖. 별도 concern.

**병합 준비 방향** (CI green 시): bug fix.

---
_근거: `gh pr diff 22595` (receipt_response_text_present_ref 값 정확화)._

jeong-sik:
Budget-checkpoint persistence fix is correct; the event-queue dedup overlaps #22561, and the scope is broad.

Two clearly-correct parts:
1. Stopping OAS turn-budget checkpoint notices from being persisted as model-authored assistant output (the `budget_exhausted` guard that skips `patch_checkpoint_last_assistant`) prevents the keeper from accumulating its own budget-exhaustion messages into history/memory. That is the self-accumulation the title names.
2. Not recording budget-exhausted / no-evidence turns as success/health success stops a stalled turn from inflating the health signal.

Two concerns:

**1. Event-queue stimuli dedup overlaps #22561.**
The "deduplicate event-queue stimuli by durable identity so repeated bootstrap enqueue/load does not grow the backlog when only arrived_at changes" touches `keeper_registry_event_queue.ml` + `keeper_runtime/keeper_event_queue.ml` + `keeper_event_queue_persistence.ml` — the exact files #22561 added `uniq_stimuli` (byte-identity dedup) to. If this PR introduces a durable-identity dedup more precise than #22561's `( = )`, it supersedes #22561's `uniq_stimuli` and the two conflict on merge. Confirm which dedup wins: if durable-identity is better, #22561's `uniq_stimuli` should be dropped in favor of this one (single definition), not both shipped.

**2. Scope — budget-checkpoint persist + no-progress latch reclassification + event-queue dedup + agent_run changes in one PR.**
Each is independently revertable. The budget-checkpoint self-accumulation fix (the title) is the headline; bundling the no-progress latch reclassification and the event-queue dedup makes the P0 hard to review in isolation.

Also: `budget_exhausted -> no-progress latch candidate` must not catch a turn that was making real progress but ran out of budget mid-way — confirm the latch only fires when budget-exhausted *and* no-evidence, not on budget exhaustion alone.

Build deferred to CI.

PR #22596, opened 2026-06-28
feat(memory): split fallback health buckets

jeong-sik:
## Summary
- split keeper memory-health fallback diagnostics into empty-response and non-empty buckets
- expose the bucket counts in backend JSON totals/per-keeper rows and dashboard typing
- render the split beside the existing fallback diagnostic total

## Stack status
- the original fallback export is already upstream through #22600 on `main`
- #22601 was merged into this branch, so this PR now carries the remaining bucket-split diff against current `main`

## Verification
- `ocamlformat --check lib/keeper/keeper_memory_os_types.ml lib/keeper/keeper_memory_os_types.mli lib/server/server_dashboard_http_keeper_memory_health.ml test/test_server_dashboard_http_keeper_memory_health.ml`
- `git diff --check origin/main...HEAD`
- `npx tsc --noEmit --pretty` from `dashboard/`
- `npx vitest run --config vitest.config.ts src/components/memory/keeper-memory-health.test.ts --no-file-parallelism --maxWorkers=1` from `dashboard/`

Full OCaml build remains CI authority per current workflow.

PR #22597, opened 2026-06-28
fix(memory): export librarian fallback predicate

jeong-sik:
Follow-up for #22593. CI Lint failed because Server_dashboard_http_keeper_memory_health called Keeper_memory_os_types.legacy_unstructured_fallback_claim, but the helper was implemented only in the .ml and not exported by the .mli. This exports the existing central predicate rather than duplicating the historical prefix check in the dashboard path. Local checks: ocamlformat --check lib/keeper/keeper_memory_os_types.mli; git diff --check.


jeong-sik:
Closing as duplicate of #22596, which already carries the same #22593 CI compile fix and was opened first.

PR #22508, opened 2026-06-28
fix(keeper): raise librarian max_tokens cap to 4096 to stop episode JSON truncation

jeong-sik:
## Problem

Keeper librarian logs repeatedly:
```
[WARN] [Keeper] memory os librarian preserving unstructured fallback
reason=librarian provider returned invalid episode JSON (invalid_json: ... Unexpected end of input)
```
at varying byte offsets (594, 862, 2490...). Episode extraction silently degrades to an unstructured fallback every affected turn.

## Root cause

`librarian_max_tokens = 1024` (hardcoded, `keeper_librarian_runtime.ml:3`) is applied as `min provider_cfg.max_tokens librarian_max_tokens` at the librarian `complete` call (`keeper_librarian_runtime.ml:246-250`). Episode JSON (summary + facts + metadata) routinely exceeds 1024 output tokens, so the response is truncated mid-object and Yojson fails with "Unexpected end of input". The varying byte offsets are the truncation point moving with each payload.

## Fix

- Raise the cap to **4096** (covers realistic episode payloads, well under the JSON-capable model budget) and move it to `Env_config.KeeperMemoryOs.librarian_max_tokens_default` so it is no longer a magic literal.
- Register the knob in the env snapshot catalog (`MASC_KEEPER_MEMORY_OS_LIBRARIAN_MAX_TOKENS`) and the keeper-memory-os env-knob test list to prevent catalog drift.

## Verification

- `lib/masc.cma` builds clean with the change.
- Broader test executable build is blocked by a pre-existing **main compile red** (`keeper_failure_taxonomy`), unrelated to this change (inherited).

## Follow-up (not in this PR)

`json_of_output` parses Yojson directly without stripping ```json``` code fences or prose preambles — a separate hardening. A `json_schema` response_format (oas#2160) would also structurally constrain provider output. Both complementary, tracked separately.

Co-Authored-By: Claude <noreply@anthropic.com>

PR #22601, opened 2026-06-28
feat(memory): split fallback health buckets

jeong-sik:
## Summary
- split keeper memory-health fallback diagnostics into empty-response and non-empty buckets
- expose the bucket counts in backend JSON totals/per-keeper rows and dashboard typing
- render the bucket split beside the existing fallback diagnostic total

## Live observation
- current live fact store has 90 librarian fallback diagnostics across keepers
- only 2 are '<empty response>'; most are non-empty invalid JSON fallback rows

## Verification
- ocamlformat --check lib/keeper/keeper_memory_os_types.ml lib/keeper/keeper_memory_os_types.mli lib/server/server_dashboard_http_keeper_memory_health.ml test/test_server_dashboard_http_keeper_memory_health.ml
- npx tsc --noEmit --pretty (dashboard)
- npx vitest run --config vitest.config.ts src/components/memory/keeper-memory-health.test.ts --no-file-parallelism --maxWorkers=1 (dashboard)
- git diff --check

## Notes
- stacked on #22596 because main currently needs the keeper interface export recovery before CI can build this path

### DIFF
--- a/dashboard/src/api/dashboard-misc.ts
+++ b/dashboard/src/api/dashboard-misc.ts
@@ -233,6 +233,8 @@ export interface KeeperMemoryHealthKeeperEntry {
   near_duplicate: number
   provider_slot_busy: number
   librarian_fallback_facts: number
+  librarian_fallback_empty_response_facts: number
+  librarian_fallback_nonempty_facts: number
   alerts: KeeperMemoryHealthAlert[]
 }
 
@@ -248,6 +250,8 @@ export interface KeeperMemoryHealthResponse {
     near_duplicate: number
     provider_slot_busy: number
     librarian_fallback_facts: number
+    librarian_fallback_empty_response_facts: number
+    librarian_fallback_nonempty_facts: number
   }
   alert_summary: {
     total_alerts: number
--- a/dashboard/src/components/memory/keeper-memory-health.test.ts
+++ b/dashboard/src/components/memory/keeper-memory-health.test.ts
@@ -36,6 +36,8 @@ function makeEntry(
     near_duplicate: 0,
     provider_slot_busy: 0,
     librarian_fallback_facts: 0,
+    librarian_fallback_empty_response_facts: 0,
+    librarian_fallback_nonempty_facts: 0,
     alerts: [],
     ...overrides,
   }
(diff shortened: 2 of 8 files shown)


jeong-sik:
Bucket split answers part of the #22593 counter-as-fix concern — but the live data points to a root fix.

Splitting fallback diagnostics into empty-response vs non-empty buckets is the right response to the #22593 question ("is this just a count, or does it drive root investigation?"). The split exposes the distribution of fallback causes, which is real observability: an operator sees whether fallbacks are empty responses (provider problem) or non-empty invalid JSON (parse problem) without grepping logs.

The live observation is the valuable signal: 90 fallback diagnostics, only 2 empty-response, 88 non-empty invalid JSON. That distribution says the remaining root is invalid-JSON parsing — not empty-provider-response. Given #22508 (max_tokens truncation fix) already merged, the 88 non-empty invalid rows point to a different parse failure (encoding, schema mismatch, partial JSON). The buckets make that visible; they do not fix it.

So this is observability that should drive a follow-up root fix on the non-empty-invalid-JSON path. Two questions:
1. Is there a tracking issue / follow-up for the non-empty invalid JSON root (the 88 rows), now that this PR makes the scale visible?
2. Could the non-empty bucket subdivide further (truncation vs encoding vs schema) later, or is empty/non-empty the intended terminal granularity?

Build deferred to CI. Stacked on #22596 — noted.

jeong-sik:
## 리뷰: #22601 — fallback health bucket 세분 (좋음)

### 긍정
- fallback diagnostics를 **empty-response vs non-empty bucket**으로 세분. /goal "노 silent failure" — fallback 원인(empty provider output vs non-empty)을 구분해 관측. #22593(surface)의 정교화.
- bucket counts를 JSON totals/per-keeper + dashboard typing에 노출. 관측 계층화.

### 미세
- bucket 분류 기준이 typed(variant)인지 string field인지 — diff상 JSON 필드 추가. 분류 로직이 `legacy_unstructured_fallback_claim` 연동 시 typed 일관성 확인.

**병합 준비 방향** (CI green 시).

---
_근거: `gh pr diff 22601` (librarian_fallback_empty/nonempty bucket)._

PR #22599, opened 2026-06-28
fix(dashboard): interaction judge prompt %s + start idempotency (PR #22583 real bugs)

jeong-sik:
## PR #22583 진짜 bug 수정 (독립 구현)

PR #22583 (Interaction Judge) 적대 검증(comment)에서 자기리뷰 P0 아래의 **진짜 bug 2개**를 수정합니다. 본인 bot이 `interaction_judge` worktree에서 활발히 작업 중이라 같은 브랜치 침범을 피해 새 브랜치로 — 본인 PR에 **cherry-pick 또는 참고** 부탁드립니다.

### 1. `prompt_for_facts` `%s` bug (P0-4 근본) ⚠️
template(`config/prompts/dashboard_interaction_judge.md` L2)이 facts 자리 `%s` placeholder를 가지는데, 코드는 `Printf.sprintf "%s\n\nFacts:\n%s" template facts_str`로 template을 **인자**로 써서 template 내부의 `%s`가 치환되지 않은 채 LLM에게 **literal `%s`**로 전달됐습니다(schema 위반 근본; `last_error` telemetry가 가림).

Fix: `Printf.sprintf template facts_str`로 template의 `%s` 자리에 facts 직접 치환. template에 `%`는 이 하나뿐(grep 확인)이라 printf injection 위험 없음.

### 2. `start` idempotency (P0-1 → P1)
`started` 플래그 부재(operator_judge엔 있음) → 같은 base_path에 start 두 번 → 두 fiber 동시 60s 폴링 → **LLM 중복 청구**.

Fix: `mutable started : bool` + `st.mu` 안 CAS + `fork_daemon`(crash 복구) + `Eio.Switch.on_release (Hashtbl.remove)` cleanup(Hashtbl 무한 성장 차단). 루프의 리뷰 메타 주석("the reviewer said... Ah, I will keep...") 제거.

## 자기리뷰 P0 오탐 (fix 불필요)
- **P0-2** (while true / cancel): 실제 `let rec loop () = Eio.Fiber.check (); sleep 60.0; ...` — 이미 cancel-aware. sleep 600s 아님(60s).
- **P0-3** (lock 외부 mutation): 모든 read/write `st.mu` 안 — race 없음.

## 검증
- `dune build lib/masc.cma` EXIT 0 (`fork_daemon`/`on_release`/`started` 가시 OK).
- 1 file, +29/-10.

## 본인 PR과의 관계
[...]

PR #22583, opened 2026-06-28
feat: Interaction Judge and SSOT enforcement

jeong-sik:
Replaces fake heuristic math in world visualizer with an LLM-based Interaction Judge daemon, and cleans up frontend SSOT violations for hardcoded keeper identity sets.

PR #22576, opened 2026-06-28
fix(keeper): make externalized prompt drift explicit

jeong-sik:
## Summary
- add the missing keeper.turn_intent.task_create_guidance markdown prompt asset so runtime prompt sync can seed it
- remove the in-binary externalized-bullet prose fallback and render an explicit config-drift marker instead
- add regressions that every declared turn-intent bullet is markdown-backed and missing task-create guidance cannot silently use hardcoded prose

## Validation
- ocamlformat --check lib/keeper/keeper_unified_prompt.ml test/test_keeper_prompt_external.ml
- git diff --check HEAD~1..HEAD
- rg -n "fallback_externalized_bullet|using in-binary fallback|resolved empty and no fallback registered" lib config/prompts (no matches)

No local Dune build/test run; CI is the build authority for this runtime work.

Follow-up for #22567.

PR #22605, opened 2026-06-28
fix(dashboard): render interaction judge prompt via registry

jeong-sik:
## Summary
- Move dashboard interaction judge prompt rendering onto Prompt_registry SSOT
- Convert the prompt file from literal %s placeholder to registered {{facts_json}} template metadata
- Propagate prompt render failures through Agent_sdk.Error instead of silently sending an unresolved prompt
- Add prompt registry coverage for the interaction judge key

## Verification
- ocamlformat --check lib/dashboard/dashboard_interaction_judge.ml test/test_prompt_registry_defaults.ml
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build bin/main_eio.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_prompt_registry_defaults.exe
- ./_build/default/test/test_prompt_registry_defaults.exe

## Notes
- This is intentionally separate from #22602. #22602 repaired the base compile failure, but kept direct prompt-file I/O plus string replacement in this path.

### DIFF
--- a/config/prompts/dashboard_interaction_judge.md
+++ b/config/prompts/dashboard_interaction_judge.md
@@ -1,5 +1,11 @@
+---
+description: interaction judge prompt for dashboard collaboration surface
+category: dashboard
+template_variables: [facts_json]
+---
+
 You are the MASC Interaction Judge. Analyze the following workspace facts and logs:
-%s
+{{facts_json}}
 
 Evaluate two things based on the facts:
 1. Stigmergy Intensity (0.0 to 1.0): How much did each Keeper's actions alter the shared environment/tasks?
--- a/lib/dashboard/dashboard_interaction_judge.ml
+++ b/lib/dashboard/dashboard_interaction_judge.ml
@@ -1,6 +1,7 @@
 open Eio.Std
 
 let keeper_name = "interaction-judge"
+let prompt_dashboard_interaction_judge = "dashboard_interaction_judge"
 
 type interaction = {
   source : string;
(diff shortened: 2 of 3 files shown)


jeong-sik:
Matches the root-cause diagnosis exactly — registry SSOT + typed render-error is the right fix.

This is precisely the fix the error calls for. The live `Sys_error("config/prompts/dashboard_interaction_judge.md: No such file or directory")` was diagnosed independently as a cwd-relative literal at dashboard_interaction_judge.ml:99 — `In_channel.with_open_text "config/prompts/dashboard_interaction_judge.md"` — that ignores the `base_path` already threaded through `start ~base_path` (sourced from workspace_config). The judge fiber inherits the server cwd (no chdir, no working_dir knob in runtime.toml), so any launcher cwd != repo root raises Sys_error with no try/with/Result guard. The file itself exists in the tree (539 B, local + git HEAD + main, un-gitignored, added in #22599) — this is a path-resolution failure, not a shipping failure.

This PR closes it correctly:
1. Moving to `Prompt_registry.render_prompt_template` removes the cwd-relative literal — the registry resolves through the project's prompt ladder (Config_dir_resolver.prompts_dir, MASC_BASE_PATH-aware), so the judge no longer depends on the launcher's cwd.
2. `Result.map_error ... Agent_sdk.Error.Internal` propagates a render failure as a typed error instead of an unhandled Sys_error — the judge degrades loudly with a diagnosable error, the "no Silent Failure" behavior this path was missing.
3. Converting the `%s` placeholder to a registered `{{facts_json}}` template metadata also retires the format-string-as-template coupling flagged on #22599.

Two confirmations:
1. Does `Prompt_registry.render_prompt_template` actually walk the base_path / Config_dir_resolver.prompts_dir ladder (so MASC_BASE_PATH and non-repo-root cwds resolve) and emit the explicit config-drift marker on absence — i.e., it mirrors the #22576 / `Tool_contract_guidance.prompt_text` behavior for keeper prompts? If yes, this is the same-class fix and fully closes the loop.
2. Naming: `dashboard_interaction_judge.md` is the only underscore-named file among ~37 dot-convention siblings in config/prompts/. A rename to `dashboard.interaction_judge.md` is cosmetic but would keep the directory consistent — non-blocking.

Build deferred to CI.

jeong-sik:
## 리뷰: #22605 — judge prompt SSOT 이동 + %s 근본 fix (좋음)

### 긍정
- **Prompt_registry SSOT**로 prompt rendering 이동. literal file(`In_channel.with_open_text`) → registered template. /goal "SSOT 위반 금지" 충족.
- `%s` placeholder → `{{facts_json}}` template metadata. #22599(%s double-substitution bug)의 **근본 fix** — "고쳐진 건 고쳐졌다" (형식 bug를 SSOT template로 원천 해결).

### 연계
#22583(judge) → #22599(%s fix) → #22605(SSOT 근본). judge prompt 계보 정리.

**병합 준비 방향** (CI green 시).

---
_근거: `gh pr diff 22605` (Prompt_registry, {{facts_json}} template)._

jeong-sik:
Follow-up confirmation: Prompt_registry itself resolves through the markdown dir set by Prompt_defaults.bootstrap_runtime; server bootstrap first syncs embedded prompt assets into Config_dir_resolver.prompts_dir(), then calls Prompt_registry.set_markdown_dir on that resolved dir. The embedded asset module is generated from the full config/ source tree via the lib/embedded_config dune rule, so this changed config/prompts/dashboard_interaction_judge.md is included in packaged runtime prompt sync. I do not see a source blocker from the path-resolution question; remaining readiness is CI/draft state.

PR #22606, opened 2026-06-28
fix(keeper): schedule event queue stimuli as typed triggers

jeong-sik:
## Summary
- Add typed Event Queue scheduling triggers for Bootstrap and No_progress_recovery stimuli.
- Pass consumed non-board stimulus triggers from heartbeat intake into keeper_cycle_decision.
- Preserve the no-signal warm-keeper silence invariant while proving warm bootstrap/no-progress stimuli run and can be acked after a completed turn.

## Runtime Evidence
- Live runtime at 95dfb59cae0 repeatedly consumed bootstrap stimuli after restart without writing any post-restart decision rows.
- albini queue dedupe reduced persisted duplicate rows, but bootstrap was still leased/requeued because Bootstrap produced no pending_board_event and scheduling returned No_signal.

## Verification
- ocamlformat --check changed OCaml files
- git diff --check
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build test/test_keeper_no_signal_silence.exe
- ./_build/default/test/test_keeper_no_signal_silence.exe
- env MASC_DUNE_ALLOW_BARE_DUNE=1 scripts/dune-local.sh build bin/main_eio.exe

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -53,6 +53,7 @@ type heartbeat_event_intake = Stimulus_intake.heartbeat_event_intake = {
   pending_board_events : Keeper_world_observation.pending_board_event list;
   consumed_stimulus_count : int;
   consumed_stimuli : Keeper_event_queue.stimulus list;
+  event_queue_triggers : Keeper_world_observation.event_queue_trigger list;
 }
 
 let consume_single_heartbeat_stimulus = Stimulus_intake.consume_single_heartbeat_stimulus
@@ -202,6 +203,7 @@ let run_keepalive_unified_turn
       let scheduling =
         decide_keepalive_scheduling
           ~reactive_wake
+          ~event_queue_triggers:event_intake.event_queue_triggers
           ~keeper_resilience_of_name:(fun keeper_name ->
             if Health.is_healthy ~agent_name:keeper_name then None
             else Some "unhealthy")
--- a/lib/keeper/keeper_heartbeat_loop.mli
+++ b/lib/keeper/keeper_heartbeat_loop.mli
@@ -52,6 +52,7 @@ type heartbeat_event_intake = {
   pending_board_events : Keeper_world_observation.pending_board_event list;
   consumed_stimulus_count : int;
   consumed_stimuli : Keeper_event_queue.stimulus list;
+  event_queue_triggers : Keeper_world_observation.event_queue_trigger list;
 }
 
 val heartbeat_event_intake :
@@ -74,6 +75,7 @@ val decide_keepalive_scheduling :
   ?runtime_resilience_of_name:(string -> string option) ->
   ?keeper_resilience_of_name:(string -> string option) ->
   ?reactive_wake:bool ->
+  ?event_queue_triggers:Keeper_world_observation.event_queue_trigger list ->
   ?wake_tombstone_decide:
     (origin:Keeper_wake_tombstone.wake_origin ->
      keeper_name:string ->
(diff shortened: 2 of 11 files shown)


jeong-sik:
Typed event-queue triggers — correct fix for the silent bootstrap/no-progress stall.

The root cause is the kind this keeper-stall family keeps hitting: bootstrap (and no_progress_recovery) stimuli were consumed but never scheduled as typed triggers, so `pending_board_event` stayed empty and scheduling returned `No_signal` — the keeper consumed the stimulus without producing a decision row, a silent stall. Adding `Keeper_world_observation.event_queue_trigger list` as a typed field and plumbing it from heartbeat intake (`event_intake.event_queue_triggers`) into `keeper_cycle_decision` makes those non-board stimuli first-class scheduled triggers instead of silently dropped. The regression pinning warm bootstrap/no-progress stimuli run + ack while preserving the no-signal silence invariant is the right test.

This is another keeper-stall axis — the no-signal silent-consume stall — alongside the #22561/#22590/#22595 stimulus family.

One question: `?event_queue_triggers` defaults to `[]`. If a caller forgets to forward the intake triggers, the keeper silently falls back to the old No_signal path — the same stall this fixes, re-introduced by an omission. Is there a path where the default-empty could mask a missing wiring (e.g., a new intake site that doesn't forward triggers)? A non-optional argument at the internal call sites (where triggers are always available) would make a missing-forward a compile error rather than a silent empty.

Build deferred to CI.

jeong-sik:
## 리뷰: #22606 (typed stimuli) + #22607 (dead config 제거) — 둘 다 좋음

### #22606 — typed Event Queue triggers
- `event_queue_trigger_of_stimulus`로 Bootstrap/No_progress_recovery stimuli를 typed trigger로 매핑. /goal "노 스트링 매치 / typed" 충족.
- 메모리 `reference-masc-keeper-orphan-stimulus-persistence`(orphan stimulus ack path) 연계 — typed triggers로 스케줄링 정확화.

### #22607 — dead config autoboot_max 제거
- `bootstrap.autoboot_max`(`MASC_KEEPER_AUTOBOOT_MAX`)가 env→key mapping + runtime.toml entry에만 존재, 실제 사용 0. dead config 제거. /goal "SSOT / 노 하드코딩" — 사용 안 하는 config 정리. 메모리 `reference-masc-autoboot-excluded-silent` 연관.

**둘 다 병합 준비 방향** (CI green 시).

---
_근거: `gh pr diff 22606` (event_queue_trigger_of_stimulus), `gh pr diff 22607` (autoboot_max 제거)._

jeong-sik:
Source review for `611e3598ad1d4345c99f3e270ca222cf5077a5b5`: no source blocker found.

This patch keeps the event-queue wake path typed:

- `Keeper_event_queue.Bootstrap` and `No_progress_recovery` map to explicit `event_queue_trigger` variants.
- Those variants map to typed `turn_reason` variants (`Bootstrap_stimulus_pending`, `No_progress_recovery_stimulus_pending`) before scheduler verdict strings are rendered.
- Result-carrying stimuli (`Board_signal`, `Fusion_completed`, `Bg_completed`) still flow through `pending_board_event_of_stimulus` instead of being silently dropped.

I do not see new hardcoded local paths, env-var expansion, stringly trigger inference, or a keeper-fleet coupling risk in this delta. The tests cover warm-keeper runs for bootstrap and no-progress event-queue triggers while retaining the no-signal silence invariant.

This is source-scope only; CI/merge state remains a separate gate.

[근거] `git fetch origin +pull/22606/head:refs/remotes/pr/22606`; `git diff origin/main..refs/remotes/pr/22606 -- lib/keeper/keeper_heartbeat_stimulus_intake.ml lib/keeper/keeper_world_observation.ml lib/keeper_contract/keeper_world_observation_turn_types.ml test/test_keeper_no_signal_silence.ml`; checked 2026-06-28 KST; confidence High.

PR #22778, opened 2026-06-30
feat(keeper): require strict adversarial review responses

jeong-sik:
Stacked on #22775.\n\n## What changed\n- Remove JSON substring recovery from keeper adversarial review response fallback.\n- Keep the primary report_verdict tool path unchanged.\n- If the reviewer responds without calling report_verdict, accept only a strict JSON grounded verdict object.\n- Update docs and add parser/coverage tests to prevent prose/fenced JSON recovery from returning.\n\n## Local checks\n- git diff --check\n- ocamlformat --check lib/keeper/keeper_adversarial_review.ml lib/keeper/keeper_adversarial_review.mli test/test_keeper_adversarial_review.ml test/test_keeper_structured_output_coverage.ml\n- targeted rg checks for removed JSON substring recovery\n\nFocused build/tests left to GitHub CI per operator preference.

### DIFF
--- a/lib/keeper/keeper_adversarial_review.ml
+++ b/lib/keeper/keeper_adversarial_review.ml
@@ -29,54 +29,22 @@ let apply_report_verdict_output_schema provider_cfg =
          (Agent_sdk.Error.InvalidConfig
             { field = "verification.adversarial_review.output_schema"; detail }))
 
-let parse_json_payload text =
-  let trimmed = String.trim text in
-  let attempt s =
-    try Some (Yojson.Safe.from_string s) with
-    | Yojson.Json_error _ -> None
-  in
-  match attempt trimmed with
-  | Some json -> Some json
-  | None ->
-    let len = String.length trimmed in
-    if len = 0 then None
-    else
-      let start_opt =
-        let brace = try Some (String.index trimmed '{') with Not_found -> None in
-        let bracket = try Some (String.index trimmed '[') with Not_found -> None in
-        match brace, bracket with
-        | Some i, Some j -> Some (min i j)
-        | Some i, None -> Some i
-        | None, Some j -> Some j
-        | None, None -> None
-      in
-      match start_opt with
-      | None -> None
-      | Some start ->
-        let closing = if trimmed.[start] = '{' then '}' else ']' in
-        let rec find_last idx =
-          if idx < 0 then None
-          else if trimmed.[idx] = closing then Some idx
-          else find_last (idx - 1)
-        in
-        match find_last (len - 1) with
-        | None -> None
-        | Some fin -> attempt (String.sub trimmed start (fin - start + 1))
+let parse_grounded_verdict_from_response_text text =
+  match Yojson.Safe.from_string (String.trim text) with
+  | json -> Verifier_core.parse_grounded_verdict_from_json json
+  | exception Yojson.Json_error msg ->
+    Error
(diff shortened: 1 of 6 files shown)


jeong-sik:
## Change requested — P1: strict JSON implementation is right, but the new coverage gate is source-text policing

Current head: 0c4b1e52149f9bcf3396befc91dcf30ffd2bdf57. Formal `REQUEST_CHANGES` is blocked because GitHub does not allow requesting changes on my own PR, so leaving this as a blocking comment.

The production change is directionally correct: `parse_json_payload`/embedded JSON extraction is removed, and the text fallback now accepts only a strict JSON grounded verdict via `Yojson.Safe.from_string (String.trim text)` followed by `Verifier_core.parse_grounded_verdict_from_json`. That matches the no-prose-CoT/no-heuristic boundary.

Blocker:
- P1 TEST/SSOT: `test/test_keeper_structured_output_coverage.ml` adds `test_adversarial_review_response_text_fallback_is_strict_json`, which reads `lib/keeper/keeper_adversarial_review.ml` as source text and asserts absence of implementation substrings such as `parse_json_payload`, `String.index trimmed`, and `String.sub trimmed`.

That is not a behavioral contract. It is an implementation-shape ratchet that can fail on harmless comments/refactors and can pass if the same bug is reintroduced under different helper names. Under this review bar, this is the same class as the earlier OAS GLM source-scan blocker: no path/source heuristics, no string-match implementation policing, no SSOT-by-grep.

Required change:
- Keep the behavior tests that parse strict JSON and reject embedded prose JSON.
- Remove the source-text coverage test, or replace it with a structural/behavioral test at the parser boundary: strict JSON object accepted, fenced/prose JSON rejected, empty text rejected, array/non-object rejected if the grounded verdict decoder requires an object, and error returned rather than silent fallback.

P3: `test_response_text_rejects_embedded_json` checks the error text with `contains_substring`. That is much less severe because the carrier is currently a string error, but a typed error would be more stable if this surface grows.

No `review:basic-done` label while this source-string coverage gate remains, even though current CI is green.

jeong-sik:
Addressed the P1 source-text ratchet blocker in `545652eb114`.

Changes:
- removed `test_adversarial_review_response_text_fallback_is_strict_json` from `test/test_keeper_structured_output_coverage.ml`; it no longer scans `keeper_adversarial_review.ml` source for `parse_json_payload` / `String.index` / `String.sub` implementation substrings
- kept the existing behavior tests for strict JSON accept + embedded JSON reject
- added parser-boundary behavior tests for empty text, whitespace text, malformed JSON, and non-object JSON rejection in `test/test_keeper_adversarial_review.ml`

Local verification:
- `ocamlformat --check test/test_keeper_adversarial_review.ml test/test_keeper_structured_output_coverage.ml`
- `git diff --check`

Build/full tests left to CI per maintainer instruction.

jeong-sik:
Current-head re-review: P1 is fixed on `545652eb`.

Verified:

- the source-text policing test was removed; no `parse_json_payload` / `String.index trimmed` / `String.sub trimmed` source grep remains in the changed test surface;
- parser-boundary behavior tests now cover strict JSON accept plus embedded/prose, empty, whitespace, malformed JSON, and non-object JSON rejection;
- current CI checks are green.

No P0/P1/P2 blocker remains in this PR's own diff. Adding `review:basic-done`.

PR #22789, opened 2026-06-30
test(oas): consume typed tool-call reasoning projection

jeong-sik:
## Summary

- Bump the OAS agent_sdk pin from 0.208.1 to the current main commit `785d99977588db04fd4528f61daf26f06e60063c`, which includes canonical tool-call/reasoning projection plus Responses/preserve-thinking/parser hardening.
- Sync generated OAS pin docs/manifests and refresh the OAS API surface fingerprint.
- Add a MASC boundary regression proving completed OAS `api_response` values can be consumed through `Agent_sdk.Canonical_tool.tool_calls_of_response` without raw text parsing: contiguous thinking attaches to the following tool call, intervening text breaks adjacency, and `provider_kind` comes from typed OAS telemetry.
- Preserve OAS-owned structured replay checkpoints in MASC: `Thinking` / `RedactedThinking` blocks survive checkpoint sanitation, canonical successful turns keep the current OAS replay suffix, and MASC only normalizes the final visible assistant text once.

## Boundary

MASC remains a consumer. Provider/model dialect, CoT grouping, Responses interleaving, and replay serialization semantics stay in OAS. This PR pins the SDK surface and makes the keeper checkpoint path preserve OAS typed replay blocks instead of flattening or deleting them.

## Verification

- PASS: `ocamlformat --check lib/keeper/keeper_context_core.ml lib/keeper/keeper_agent_run_finalize_response.ml test/test_keeper_state_snapshot_json.ml`
- PASS: `ocamlformat --check test/test_gate_keeper_backend.ml`
- PASS: `bash scripts/sync-oas-pin-docs.sh --check`
- PASS: `bash scripts/sync-oas-pin-manifests.sh --check`
- PASS: `bash scripts/oas-drift-check.sh`
- PASS: `git diff --check`
- PASS: touched-file OCaml danger-pattern grep (`failwith`, `assert false`, `Fun.protect`, raw mutex locks, etc.) returned no hits
- SKIPPED: local dune build/test per operator preference; GitHub CI is the build authority.


jeong-sik:
## Change requested - P1: final visible answer can be patched into a tool-call assistant message

Current head: `51f08a541b8dfd330cb25faf0361ad8c243b89a5`.

The OAS pin bump and the typed adjacent-reasoning projection test are directionally good. Preserving OAS-owned `Thinking` / `RedactedThinking` / tool blocks instead of flattening them is also the right boundary: MASC should not destroy OAS structured replay data.

But the new checkpoint persistence flow has a semantic corruption case.

Problem:

```ocaml
let current_suffix = drop_prefix history_messages checkpoint.messages in
if List.exists (fun msg -> msg.role = Assistant) current_suffix
then checkpoint.messages
else checkpoint.messages @ [ Assistant [Text response_text] ]
```

This suppresses appending the final assistant whenever the current-turn suffix contains any assistant message. A tool-call assistant message is still an assistant message. In that case `patch_checkpoint_last_assistant` later patches the last assistant message by replacing/appending visible text while preserving the existing typed blocks.

That means a suffix like:

1. current user
2. assistant `Thinking + ToolUse`
3. tool result

with a successful `response_text = "visible answer"` will not append a separate final assistant after the tool result. Instead, the tool-call assistant can become `Thinking + ToolUse + Text "visible answer"`, so the final answer is recorded before the tool result it depended on and in the same assistant message that requested the tool.

Why this is P1:

- It corrupts replay ordering at the Keeper/OAS boundary.
- It can turn a valid tool-use -> tool-result -> final-answer sequence into a mixed tool-call/final-answer assistant message.
- This is exactly the class of structured replay corruption this PR claims to fix, and the current tests miss it. The added test only covers a suffix with no assistant at all; it does not cover a suffix with a tool-use assistant but no final assistant.

Required fix:

- Do not decide append/no-append with `exists role = Assistant`.
- Detect whether the suffix already contains a final assistant response after the tool/result sequence, or at minimum whether the last assistant being patched is actually a final visible-response assistant rather than a tool-use-only assistant.
- If the suffix has tool-use assistant blocks but no final assistant response, append a new assistant `Text response_text` after the suffix, then patch that final message.
- Add a regression with suffix `[current_user; assistant Thinking+ToolUse; tool_result]` and non-empty `response_text`, asserting the persisted checkpoint ends with a separate assistant text message after the tool result and the tool-use assistant is not patched with the final answer.

No `review:basic-done` label. CI is also still in progress on this current head.

[Evidence] `lib/keeper/keeper_agent_run_finalize_response.ml` at `51f08a541b8dfd330cb25faf0361ad8c243b89a5`; `lib/keeper/keeper_context_core.ml` `patch_checkpoint_last_assistant`; `gh pr diff 22789 --repo jeong-sik/masc --patch`; checked 2026-06-30 KST; confidence High.

PR #22796, opened 2026-06-30
fix(keeper): expose redacted stream fallback helper

jeong-sik:
## Summary
- Expose For_testing.redacted_visible_reply_with_stream_fallback in server_routes_http_keeper_stream.mli.
- Keep the post-22763 fix scoped to the interface drift that broke CI Health compile.

## Evidence
- PR 22763 merged at 8962113ce4e before its CI Health failure was resolved.
- Failing job: 84237854809 in run 28428612103.
- Observed compile failure: Unbound value Server_routes_http_keeper_stream.For_testing.redacted_visible_reply_with_stream_fallback.

## Local checks
- ocamlformat --check lib/server/server_routes_http_keeper_stream.mli passed.
- git diff --check passed.
- scripts/dune-local.sh build test/test_gate_keeper_backend.exe was attempted, but stopped while waiting on /tmp/me-dune-local.lock; CI is the build authority for this follow-up.

### DIFF
--- a/lib/server/server_routes_http_keeper_stream.mli
+++ b/lib/server/server_routes_http_keeper_stream.mli
@@ -189,6 +189,8 @@ module For_testing : sig
   val direct_reply_terminal_error : Yojson.Safe.t option -> string -> string option
   val visible_reply_with_stream_fallback :
     streamed_text:string -> string -> string
+  val redacted_visible_reply_with_stream_fallback :
+    redact:(string -> string) -> streamed_text:string -> string -> string
   val reply_payload_with_streamed_visible_reply :
     Yojson.Safe.t option -> visible_reply:string -> Yojson.Safe.t option
   val format_surface_context : Yojson.Safe.t -> string


jeong-sik:
## Loop re-review - postmerge CI fix direction OK, CI still pending

Current head: `a0e5a2e865688321f67b4493ceb61bf366900616`.

This is the right narrow follow-up for the #22763 post-merge CI break. #22763 merged with the implementation/test call site present, but `server_routes_http_keeper_stream.mli` did not expose `For_testing.redacted_visible_reply_with_stream_fallback`, causing CI Health/Build to fail with `Unbound value Server_routes_http_keeper_stream.For_testing.redacted_visible_reply_with_stream_fallback`.

The diff adds only the missing signature entry. I do not see a new path/env knob, hardcoded local path, heuristic, stub, or duplicated module. It uses the existing implementation already exported inside `For_testing` in the `.ml`.

No readiness label yet: this PR is draft/blocked and CI Build/Test, Health, Lint, Meta Guards, and ocamlformat are still pending. If that CI turns green, this closes the #22763 interface-drift regression.

PR #22879, opened 2026-07-01
fix(board): propagate append persistence failures

jeong-sik:
## Summary
- make board post/comment append helpers return typed `Io_error` instead of only logging persistence failures
- propagate append failures through create/comment/pin paths so success fanout is not emitted on failed writes
- roll back in-memory fresh post/comment/pin state on append failure and cover those regressions in board dispatch tests

## Verification
- `ocamlformat --check lib/board/board_core_persist.ml lib/board/board_core_persist.mli lib/board/board_core.ml lib/board/board_core.mli lib/board/board_votes.ml test/test_board_dispatch.ml`
- `git diff --check`
- local Dune build/test intentionally not run; CI is the build authority for this slice

Fixes part of #21990.

### DIFF
--- a/lib/board/board_core.ml
+++ b/lib/board/board_core.ml
@@ -4,6 +4,32 @@
 
 include Board_core_persist
 
+let rollback_fresh_comment store ~(comment : comment) ~(previous_post : post) =
+  with_lock store (fun () ->
+    let post_key = Post_id.to_string comment.post_id in
+    let comment_key = Comment_id.to_string comment.id in
+    Hashtbl.remove store.comments comment_key;
+    (match Hashtbl.find_opt store.comments_by_post post_key with
+     | None -> ()
+     | Some ids ->
+       (match List.filter (fun id -> not (String.equal id comment_key)) ids with
+        | [] -> Hashtbl.remove store.comments_by_post post_key
+        | filtered -> Hashtbl.replace store.comments_by_post post_key filtered));
+    (match Hashtbl.find_opt store.posts post_key with
+     | None -> ()
+     | Some current ->
+       let updated_at =
+         if Stdlib.Float.equal current.updated_at comment.created_at
+         then previous_post.updated_at
+         else current.updated_at
+       in
+       Hashtbl.replace
+         store.posts
+         post_key
+         { current with reply_count = max 0 (current.reply_count - 1); updated_at });
+    invalidate_post_caches store;
+    invalidate_comment_caches store)
+;;
 
 let get_post store ~post_id : (post, board_error) Result.t =
   maybe_sweep store;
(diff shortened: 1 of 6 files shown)


jeong-sik:
P1 blocker: fresh-comment success still ignores the post snapshot write failure.

The PR fixes the direct `append_comment` failure path, but fresh comment persistence still has two durable writes: append the comment row, then rewrite `board_posts.jsonl` with the updated `reply_count`/`updated_at`. The second write still cannot fail the operation. `save_posts_jsonl` logs `Fs_compat.save_file_atomic` errors and returns `unit` (`lib/board/board_core_persist.ml:362-370`), and `add_comment_with_status` immediately turns that call into `Ok ()` after `append_comment` succeeds (`lib/board/board_core.ml:431-438`). `Board_dispatch.add_comment` then emits the keeper signal/SSE and returns success for `Ok (comment, Fresh)` (`lib/board/board_dispatch.ml:497-517`).

That leaves the same silent-persistence shape this PR is trying to remove: a comment append can succeed, the post snapshot rewrite can fail, and callers still get success/fanout even though the durable post state did not record the reply count. A restart then reloads a comment row with stale post metadata. The new regression only blocks the whole `.masc` path, so `append_comment` fails first and this partial-write path is not exercised.

Required fix: make the post snapshot rewrite return `(unit, board_error) result` for this call path and fail/compensate when it fails, or explicitly split the operation semantics and surface the partial persistence result to the caller. Please add a regression where `append_comment` succeeds but `save_posts_jsonl` fails, and assert no success fanout is emitted.

jeong-sik:
Current-head review: no blocker found for this append-failure propagation scope.

Reviewed head: `f2db7bc8707a5a70b8dfa151a5a54d214934060a`.

What I checked:
- `append_post` / `append_comment` now return an explicit result and the create/comment/pin flows stop publication/fanout on append failure.
- The rollback path keeps the in-memory board from advertising a write that was not durably appended.
- The added append-failure tests cover the requested board append loss cases, including the test-path update in the latest force-push.
- CI is green on the current head, including `Build and Test` and `CI Gate`.

Residual note: broader board persistence fail-soft behavior outside this append scope should stay tracked under the parent persistence issue rather than blocking this PR.

jeong-sik:
Metadata follow-up: clearing the stale `p1` label on current head `f2db7bc8707a5a70b8dfa151a5a54d214934060a`.

The earlier P1 was superseded by the current-head review at https://github.com/jeong-sik/masc/pull/22879#issuecomment-4851583434. I rechecked the current PR state: `Build and Test`, `CI Gate`, `Meta Guards`, `Lint`, and the focused checks are green. The remaining broader board persistence note is outside this append-failure propagation PR scope.

PR #22888, opened 2026-07-01
fix(dashboard,metrics): surface malformed trajectory rows and wire dead contract-violations counter

jeong-sik:
## Summary
Two independent P2 silent-failure fixes bundled in one PR (same repo, low blast radius, no shared code path):

1. **`/trajectory` dashboard endpoint silently dropped malformed rows** — `Trajectory.trajectory_lines_of_jsonl_lines` (`lib/trajectory/trajectory.ml`) used a bare `List.filter_map` that swallowed `Yojson.Json_error`/`Type_error` with zero signal. Every keeper's `/trajectory` response and the chat-trace builder dropped corrupted/malformed rows with no indication anything was lost. The sibling `internal_history` merge path (`Server_dashboard_http_keeper_api_trace.log_internal_history_skips`) already tracks `skipped`/`total` and logs a per-read WARN summary — applied the same pattern here.
2. **`masc_keeper_contract_violations_total` was a dead metric** — declared, given a wire name, registered in the zero-fill catalog, but never incremented anywhere (confirmed via `rg`, zero non-declaration references). Wired the increment at both construction/application sites of `Keeper_registry.Completion_contract_violation`.

### 1. Trajectory malformed-row visibility
- Added `log_trajectory_line_skips` (mirrors `log_internal_history_skips`'s summarized-not-per-row pattern — a busy re-polled trace file would otherwise flood the log the same way `internal_history` did before its own fix).
- `trajectory_lines_of_jsonl_lines` now takes `~trace_id` and folds skipped/total instead of a bare `filter_map`. Public signatures of `read_all_lines`/`read_recent_lines` unchanged (both already had `trace_id` in scope).

### 2. `ContractViolations` counter wiring
- `keeper_unified_turn_success.ml`'s `record_completion_contract_attention_failure` (constructs `Completion_contract_violation` directly) — added the increment as a direct sibling to its existing `increment_turn_failures`/`set_failure_reason` calls.
- `keeper_unified_turn.ml`'s `registry_failure_reason_of_terminal_reason` application site (a generic `failure_reason` sink used for many reason types) — gated with an **exhaustive match**, not `_ -> ()`, so a future `failure_reason` variant must explicitly decide whether it also bumps this counter rather than silently landing in a catch-all.

## RFC
Not required — dashboard trajectory read path + keeper metrics wiring, not credential/identity/operator-control/sandbox/dashboard-credential/hooks/workflow. `bash scripts/pr-rfc-check.sh` passes.

## Test plan
- [x] `dune build` (full project) — clean
- [x] `test_trajectory.exe` — 38/38 pass, incl. new `test_read_recent_lines_skips_malformed_rows` proving both `read_recent_lines` and `read_all_lines` still drop a malformed row while keeping the valid one. Manually verified (`MASC_LOG_LEVEL=debug --verbose`) the new WARN fires: `trajectory trace X: 1 of 2 rows did not decode to a trajectory line`.
- [x] `test_dashboard_keeper_metrics_10286`, `test_gate_keeper_backend`, `test_keeper_overflow_recovery`, `test_keeper_pause_silent_failure_source`, `test_keeper_receipt_authoritative_matrix`, `test_keeper_state_machine`, `test_keeper_turn_fsm_wired_sites`, `test_keeper_unified_tool_event_order_source`, `test_keeper_validation_breaker_exempt`, `test_no_progress_loop_detector`, `test_operator_control_snapshot`, `test_telemetry_unified` — all pass unchanged
- [ ] **Not covered by a dedicated unit test**: the `ContractViolations` counter increment itself has no isolated test harness for `record_completion_contract_attention_failure` (needs full `Workspace.config` + `keeper_meta` scaffolding to drive an actual completion-contract-violation turn failure). Verified by code inspection + the exhaustive match compiling, not by an assertion against `Otel_metric_store.metric_value_or_zero`.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/lib/keeper/keeper_unified_turn.ml
+++ b/lib/keeper/keeper_unified_turn.ml
@@ -807,6 +807,28 @@ let run_keeper_cycle
                         ~raw_error:e_str
                     with
                     | Some failure_reason ->
+                      (* masc_keeper_contract_violations_total (RFC exhaustive
+                         match, not [_ -> ()]): a new failure_reason variant
+                         must explicitly opt in or out of this counter rather
+                         than silently landing in a catch-all. *)
+                      (match failure_reason with
+                       | Keeper_registry.Completion_contract_violation _ ->
+                         Otel_metric_store.inc_counter
+                           Keeper_metrics.(to_string ContractViolations)
+                           ~labels:[ "keeper", meta.name ]
+                           ()
+                       | Keeper_registry.Heartbeat_consecutive_failures _
+                       | Keeper_registry.Turn_consecutive_failures _
+                       | Keeper_registry.Stale_turn_timeout _
+                       | Keeper_registry.Stale_termination_storm _
+                       | Keeper_registry.Stale_fleet_batch _
+                       | Keeper_registry.Provider_timeout_loop _
+                       | Keeper_registry.Provider_runtime_error _
+                       | Keeper_registry.Ambiguous_partial_commit _
+                       | Keeper_registry.Fiber_unresolved _
+                       | Keeper_registry.Exception _
+                       | Keeper_registry.Turn_overflow_pause
+                       | Keeper_registry.Turn_livelock_pause -> ());
                       Keeper_registry.set_failure_reason
                         ~base_path:config.base_path
                         meta.name
--- a/lib/keeper/keeper_unified_turn_success.ml
+++ b/lib/keeper/keeper_unified_turn_success.ml
@@ -770,6 +770,10 @@ let record_completion_contract_attention_failure
     ~base_path
     updated_meta.name
     (Some (Keeper_registry.Completion_contract_violation { detail }));
+  Otel_metric_store.inc_counter
+    Keeper_metrics.(to_string ContractViolations)
+    ~labels:[ "keeper", updated_meta.name ]
+    ();
   Health.record_failure
     ~agent_name:updated_meta.name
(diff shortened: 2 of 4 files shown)


jeong-sik:
P2: `trajectory_summary` rows now produce false malformed-row warnings

Reviewed current head `b77c80e4549439302d55c408c817f9c5dc99259a`.

`append_summary` writes a valid `{"type":"trajectory_summary" ...}` row at the end of each finalized trajectory (`lib/trajectory/trajectory.ml:313-332`). `trajectory_line_of_json` intentionally returns `None` for that recognized row type (`lib/trajectory/trajectory.ml:752-755`) because `/trajectory` returns tool/thinking lines, not the terminal summary. The new `trajectory_lines_of_jsonl_lines` fold treats every `None` as `skipped` and logs it as "malformed JSON or unrecognized shape" (`lib/trajectory/trajectory.ml:806-821`).

That means any finalized trace whose tail includes the summary row will warn on a normal dashboard read, likely repeatedly while the dashboard polls the trace. This turns the new malformed-row signal into log noise and makes real corrupt rows harder to distinguish. Please split the decoder result so recognized-but-ignored rows such as `trajectory_summary` do not increment the malformed/unrecognized skip count, and add a regression with a valid summary row plus a genuinely malformed row.

PR #22896, opened 2026-07-01
[codex] Make settings theme switch reach Paper

jeong-sik:
## Summary

- Fix the dashboard ThemeSwitch cycle so the Settings Display theme control can reach every advertised theme: Dark -> StyleSeed -> Paper -> Dark.
- Update the button titles to describe the actual next theme.
- Add focused coverage for ThemeSwitch and the Settings Display integration path reaching Paper.

## Impact

The Settings Display page no longer advertises Paper as a theme option while making it unreachable from the theme button. `?theme=paper` and persisted `paper` were already supported by bootstrap; this wires the live control to that existing state.

## Validation

- `vitest run --config vitest.config.ts src/components/theme-switch.test.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1`
- `tsc --noEmit`
- `eslint src/components/theme-switch.ts src/components/theme-switch.test.ts src/components/settings-surface.test.ts` (test files are ignored by current ESLint config; source file passed)
- `git diff --check`
- Playwright smoke: `/dashboard/#settings?section=display`, click Theme twice, verify `html[data-theme="paper"]`, `localStorage.dashboardTheme = "paper"`, and `?theme=paper`.

Full build is left to CI.

### DIFF
--- a/dashboard/src/components/settings-surface.test.ts
+++ b/dashboard/src/components/settings-surface.test.ts
@@ -418,7 +418,7 @@ describe('SettingsSurface', () => {
     expect(container.querySelector('[data-testid="log-viewer"]')).not.toBeNull()
   })
 
-  it('renders the theme switch inside the display section (moved out of the top bar)', () => {
+  it('renders a live theme switch inside the display section that can reach Paper', async () => {
     route.value = { tab: 'settings', params: { section: 'display' }, postId: null }
 
     render(html`<${SettingsSurface} />`, container)
@@ -427,6 +427,13 @@ describe('SettingsSurface', () => {
     const themeButton = [...container.querySelectorAll('button')]
       .find(b => /DARK|STYLESEED|PAPER/.test(b.textContent ?? ''))
     expect(themeButton).toBeTruthy()
+
+    await fireEvent.click(themeButton as HTMLButtonElement)
+    expect(document.documentElement.dataset.theme).toBe('styleseed')
+
+    await fireEvent.click(themeButton as HTMLButtonElement)
+    expect(document.documentElement.dataset.theme).toBe('paper')
+    expect(localStorage.getItem('dashboardTheme')).toBe('paper')
   })
 
   it('wires display density to the dashboard density signal without fake locale previews', async () => {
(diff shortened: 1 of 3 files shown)


jeong-sik:
P1 blocker on current head `52fd77be46fe4efc9a6a4ca2a03df33105acde2f`.

This makes `ThemeSwitch` itself cycle `dark -> styleseed -> paper -> dark`, but it leaves the full dashboard with two independent theme owners. `ThemeSwitch` persists to `THEME_STORAGE_KEYS` (`dashboardTheme`, `masc-theme-v2`) and writes `document.documentElement.dataset.theme`; `App` still has a root effect that writes the same document attribute from `tweaksTheme`, whose separate storage key is `dashboard:tweaks:theme-v2` and whose type only supports `dark | paper`.

Concrete failure mode:

1. In Settings Display, click the switch twice. The component writes `dashboardTheme=paper`, `masc-theme-v2=paper`, and `html[data-theme="paper"]`.
2. Reload the dashboard with the tweaks theme still at its default `dark`.
3. `main.ts` initially restores `paper` from `THEME_STORAGE_KEYS`, but after `App` mounts its effect runs and sets `document.documentElement.setAttribute('data-theme', tweaksTheme.value === 'paper' ? 'paper' : '')`, so the restored Paper theme is erased unless the unrelated tweaks theme key also happens to be `paper`.

The new tests miss this because they mount `ThemeSwitch` / `SettingsSurface` in isolation and never render `App`, so the competing document-level writer is absent. This also leaves StyleSeed vulnerable to the same owner split whenever the App-level theme sync runs.

Required fix: make theme state a single SSOT. Either have `App` consume the same `ThemeId`/storage/query state that `ThemeSwitch` writes, or stop the App effect from overwriting a ThemeSwitch/bootstrap-owned `html[data-theme]`. Add an App-level regression that seeds `localStorage.dashboardTheme = "paper"` (and ideally `styleseed`) before mounting `App` and proves the document theme survives after effects flush.

CI is still in progress, so this is a source blocker independent of final check state.

[근거] `git grep -n "dashboard:tweaks:theme-v2\\|THEME_STORAGE_KEYS\\|masc-theme-v2\\|dashboardTheme" origin/pr/22896 -- dashboard/src`, `git grep -n "tweaksTheme\\|data-theme\\|documentElement.*theme" origin/pr/22896 -- dashboard/src/app.ts dashboard/src/components dashboard/src/lib`, and `git diff --unified=120 origin/main...origin/pr/22896 -- dashboard/src/components/theme-switch.ts ...`, confirmed 2026-07-01 17:22:51 KST, confidence High.

PR #22916, opened 2026-07-01
[codex] Surface MCP inventory load failures

jeong-sik:
## Summary
- Split Settings > MCP tool inventory into loading, ready, and error states.
- Stop rendering inventory fetch failures as an empty MCP tool list.
- Keep the `Check MCP` action independent so `masc_status` can still be verified even when the inventory endpoint is down.

## Validation
- `vitest run --config vitest.config.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1`
- `tsc --noEmit --pretty false`
- `eslint src/components/settings-surface.ts src/components/settings-surface.test.ts` (source clean; test file ignored by current eslint config)
- Playwright smoke on `http://127.0.0.1:5189/dashboard/#settings?section=mcp`: mocked `/api/v1/dashboard/tools` as 503, verified an error state and unknown count `(—)`, clicked `Check MCP`, and verified the mocked `masc_status` result still renders. Screenshot: `/private/tmp/masc-settings-mcp-inventory-error.png`

### DIFF
--- a/dashboard/src/components/settings-surface.test.ts
+++ b/dashboard/src/components/settings-surface.test.ts
@@ -563,6 +563,31 @@ describe('SettingsSurface', () => {
     })
   })
 
+  it('MCP server page surfaces inventory load failures instead of fabricating an empty tool list', async () => {
+    apiMock.fetchDashboardTools.mockRejectedValueOnce(new Error('inventory offline'))
+    mcpMock.callMcpTool.mockResolvedValueOnce('status ok despite inventory failure')
+
+    render(html`<${SettingsSurface} />`, container)
+
+    await fireEvent.click(container.querySelector('[data-testid="settings-nav-mcp"]') as HTMLElement)
+
+    await waitFor(() => {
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head source review on `f39fa02230f1936cf5a66b3af19fb383746091fb`: no P1/P2 issue found.

The change keeps the inventory endpoint and the live `masc_status` check independent: inventory load failure now renders an explicit error/count-unknown state instead of an empty tool list, while `Check MCP` still calls `masc_status` directly. The added test covers that regression path.

I am not adding `review:basic-done` yet because Build/Test, Dashboard, Health, and Lint are still pending on this head.

[근거] `git fetch origin +pull/22916/head:refs/remotes/origin/pr/22916`; `git diff origin/main...origin/pr/22916 -- dashboard/src/components/settings-surface.ts dashboard/src/components/settings-surface.test.ts`; GraphQL `reviewThreads(first:80)`; `gh pr checks 22916 --repo jeong-sik/masc`; confirmed 2026-07-01 19:38:00 KST; confidence High.

PR #22940, opened 2026-07-01
[codex] Fix runtime routing controls and audit

jeong-sik:
## What changed

- Fix the dashboard runtime assignment editor so selecting the current default runtime writes an explicit keeper assignment instead of silently clearing it.
- Add explicit `고정` / `폴백` controls for keeper runtime assignments, including the case where a keeper is pinned to the same runtime as `[runtime].default`.
- Expose Settings → Runtime model routing controls for:
  - `[runtime].librarian`
  - `[runtime].structured_judge`
  - `[runtime].cross_verifier`
  - ordered `[runtime].media_failover`
- Remove the `VITE_FUSION_SETTINGS_WRITABLE` gate so Settings → Fusion opens the live runtime.toml-backed Fusion editor by default.
- Refresh runtime consumers after runtime.toml writes: runtime catalog singleton, keeper shell runtime status, and execution slice.
- Extend runtime.toml writers and the routing API for `structured_judge` and ordered `media_failover` without adding a parallel TOML writer.
- Enrich `runtime_config_write` audit events with non-secret operation metadata, including list-valued routing updates.

## Why

The Settings runtime routing section was a read-only projection even though the dashboard already had write-capable runtime.toml endpoints. Operators could see librarian/cross-verifier/media-failover values but could not change them there.

The Fusion editor existed, but Settings hid it behind an env flag, so the live writer looked unavailable by default.

Runtime saves also reloaded the server config but did not invalidate dashboard runtime consumers, which left surfaces like the right rail showing stale runtime labels until a later refresh.

## Validation

- `ocamlformat --check lib/runtime/runtime.ml lib/runtime/runtime.mli lib/server/server_routes_http_routes_dashboard.ml lib/audit_log.ml test/test_runtime_per_keeper_routing.ml test/test_runtime_config_validity.ml test/test_audit_log.ml`
- `git diff --check`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/settings-surface.test.ts src/components/runtime-toml-editor.test.ts src/components/fusion-settings-panel.test.ts src/api/dashboard.test.ts src/api/schemas/runtime-defaults.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard exec tsc --noEmit`

## Not run

- `scripts/dune-local.sh build test/test_runtime_per_keeper_routing.exe test/test_runtime_config_validity.exe test/test_audit_log.exe`

Blocked by an existing bare Dune process:

```text
[dune-local] live Dune process outside scripts/dune-local.sh detected:
[dune-local]   61599 61587 dune build --root . lib/
[dune-local] refusing to start another local build while the machine-wide Dune lock is bypassed
```

Rendered Playwright validation was attempted against a worktree Vite dev server, but the local Playwright JS package is not installed and the PATH Playwright shim points at a missing pipx interpreter.

### DIFF
--- a/dashboard/src/api/dashboard-runtime.ts
+++ b/dashboard/src/api/dashboard-runtime.ts
@@ -477,7 +477,7 @@ export async function saveRuntimeTomlConfig(sourceText: string): Promise<Runtime
   }).then(normalizeRuntimeTomlConfig)
 }
 
-export type RuntimeRoutingLane = 'default' | 'librarian' | 'cross_verifier'
+export type RuntimeRoutingLane = 'default' | 'librarian' | 'structured_judge' | 'cross_verifier'
 
 export async function patchRuntimeRouting(
   lane: RuntimeRoutingLane,
@@ -487,6 +487,16 @@ export async function patchRuntimeRouting(
   return post<unknown>('/api/v1/runtime/config/routing', {
     lane,
     runtime_id: runtimeId,
+  }).then(normalizeRuntimeTomlConfig)
+}
+
+export async function patchRuntimeMediaFailover(
+  runtimeIds: readonly string[],
+): Promise<RuntimeTomlConfig> {
+  await ensureDevToken()
+  return post<unknown>('/api/v1/runtime/config/routing', {
+    lane: 'media_failover',
+    runtime_ids: [...runtimeIds],
   }).then(normalizeRuntimeTomlConfig)
 }
 
--- a/dashboard/src/api/dashboard.test.ts
+++ b/dashboard/src/api/dashboard.test.ts
@@ -39,6 +39,7 @@ import {
   fetchRuntimeDefaults,
   fetchRuntimeModelMetrics,
   patchRuntimeAssignment,
+  patchRuntimeMediaFailover,
   patchRuntimeRouting,
   patchKeeperConfig,
   saveRuntimeTomlConfig,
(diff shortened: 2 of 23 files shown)


jeong-sik:
Reviewed current head `cd7968310f4fb37dd61df28d53309f05e755490d` for runtime assignment pin/fallback semantics, dirty-draft guards, backend assignment/routing write validation, and the new audit metadata projection.

No P0/P1/P2/P3 source blocker found in this pass. The UI now distinguishes explicit pins from default fallback, the parent structured editor still blocks typed patches while the TOML draft is dirty, and the backend validates the resulting runtime config before the atomic save/cache refresh.

Scope note: this is source review only. CI was still in progress/queued when checked, so this is not a green-CI attestation.

jeong-sik:
CI refresh on current head `cd7968310f4fb37dd61df28d53309f05e755490d`: source review still has no assignment/pin-control blocker, but the PR is now CI-blocked and should not carry `review:basic-done`.

`Build and Test` fails in the runtime TOML validity gate:

```text
FAIL runtime binding deepseek.deepseek-v4-pro provider/model openai_compat/deepseek-v4-pro must resolve through repo OAS catalog
```

This is the same repo-catalog drift now showing up on other MASC PRs: runtime config has `[deepseek.deepseek-v4-pro]`, but the repo OAS catalog does not contain the provider-qualified `openai_compat/deepseek-v4-pro` projection required by the gate.

I am removing `review:basic-done` and adding `p1` until this branch is green. The dashboard assignment/pin logic itself remains source-reviewed; the blocker is CI/runtime-catalog validity.

PR #22980, opened 2026-07-02
fix(governance): add unconditional hard_forbidden gate at decide() entry for Critical risk tools

anyang-keepers:
## Problem

`decide()` never returns `Deny` — even Critical-risk tools (Board_sub_board_delete, Execute shell, git push --force) pass through as `Allow` or at worst `Require_confirm`. This means an operator who types `git push --force origin main` in a tool call gets no hard block.

## Fix

Add a `risk = Critical` guard at the top of `decide()` that unconditionally returns `Deny` with a descriptive message. This is the **hard_forbidden gate** — Critical tools are always denied regardless of governance level or HITL mode.

```ocaml
if risk = Critical then
  `Deny (Printf.sprintf ...)
else
  ...existing confirm_threshold logic...
```

## Risk Assessment

- `Board_sub_board_delete` → Critical → now Denied ✅
- `Execute` (shell) → High → still uses confirm_threshold (unchanged) ⚠️
- `Fs_write` / `Fs_edit` → High → still uses confirm_threshold (unchanged)
- `Read` / `Board_post` → Low → still Allow (unchanged)

## Testing

- [x] Critical = Deny (new gate)
- [x] High/Medium/Low = confirm_threshold (unchanged)
- [x] +13/-5 in `lib/governance_pipeline.ml`, no new deps

Closes task-1626

### DIFF
--- a/lib/governance_pipeline.ml
+++ b/lib/governance_pipeline.ml
@@ -91,15 +91,23 @@ let decide ~governance_level ~tool_name ~input =
   let risk = assess_risk ~tool_name ~input in
   let trace_id = generate_trace_id () in
   let action =
-    match confirm_threshold governance_level with
-    | Some threshold when risk_level_to_int risk >= risk_level_to_int threshold ->
-      `Require_confirm
+    if risk = Critical then
+      `Deny
         (Printf.sprintf
-           "Governance (%s): %s risk tool %S requires confirmation"
+           "Governance (%s): %s risk tool %S unconditionally denied (hard_forbidden)"
            governance_level
            (risk_level_to_string risk)
            tool_name)
-    | _ -> `Allow
+    else
+      match confirm_threshold governance_level with
+      | Some threshold when risk_level_to_int risk >= risk_level_to_int threshold ->
+        `Require_confirm
+          (Printf.sprintf
+             "Governance (%s): %s risk tool %S requires confirmation"
+             governance_level
+             (risk_level_to_string risk)
+             tool_name)
+      | _ -> `Allow
   in
   { tool_name; risk; action; trace_id }
 ;;


jeong-sik (changes requested):
P1 blocker at current head `1bfb72f3b2f96f6ca76a303dc5fc619432962e50`.

This patch only changes `decide()` at `lib/governance_pipeline.ml:90-112`. That path feeds the `Tool_dispatch` pre-hook. It does not change the production OAS approval callback at `to_oas_approval_callback`: the callback still computes `hard_forbidden`, `soft_forbidden`, `auto_approval_forbidden`, and then calls `submit_and_await` rather than applying the new `Deny` branch.

So the PR body overclaims the fix. The described incident is an OAS approval-path problem: forbidden critical/destructive requests must be handled deterministically in `to_oas_approval_callback`, not only in the older `decide()` helper. Merging this as-is gives a false "hard_forbidden gate" signal while the OAS callback behavior remains unchanged.

P1: coverage is missing for the path that matters. CI's Test Coverage Check is already red because this PR adds 13 covered-code lines with no test file changes. More importantly, the missing tests need to exercise the callback behavior, not just `decide()`:

- `MASC_DISABLE_HITL=true` + critical risk request
- destructive git/shell op through `to_oas_approval_callback`
- runtime-blocker meta through `to_oas_approval_callback`
- assertion that the request is not silently approved or stranded in the wrong queue

P2: this overlaps with #22907 and conflicts conceptually with #22978. #22907 is the broader implementation that actually touches the OAS callback, audit event, and read-model projection. It still has review concerns, but it is the right blast-radius lane. Please supersede/close this narrow PR or rework it into the same callback-centered fix; do not merge a separate `decide()`-only security gate as if it closed the production bypass.

P3: the PR body examples are internally inconsistent. It says `git push --force origin main` gets no hard block, but its own risk assessment says `Execute` remains High/confirm-threshold unchanged. If payload escalation is supposed to make the request Critical, that needs a regression test and the callback must be tested with the actual payload shape.

[근거] `git fetch origin pull/22980/head:refs/tmp/pr-22980 pull/22907/head:refs/tmp/pr-22907`; `git diff origin/main...refs/tmp/pr-22980 -- lib/governance_pipeline.ml`; `git show refs/tmp/pr-22980:lib/governance_pipeline.ml | nl -ba`; `gh pr view 22980 --repo jeong-sik/masc --json headRefOid,statusCheckRollup`; `gh run view 28558901697 --job 84672282481 --log`; confirmed 2026-07-02 KST; confidence High.

jeong-sik:
Current-head review for `1bfb72f3b2f96f6ca76a303dc5fc619432962e50`.

P1 - this changes the generic governance decision contract. `decide` now returns `Deny` for every `Critical` risk before consulting `confirm_threshold`, which bypasses the existing HITL/confirmation semantics for development, production, enterprise, paranoid, and unknown-level fail-closed cases. The failing CI tests are aligned with that contract: `governance_levels` and `pre_hook_integration` now fail on cases that previously allowed or required confirmation.

If the intended policy is “hard-forbidden only for auto-approval/runtime-blocked contexts,” this should not be implemented as `risk = Critical -> Deny` inside generic `decide`. Wire the existing hard-forbidden policy through the boundary that has the real metadata, keep `decide` as confirm-vs-allow policy, and cover the runtime-blocker metadata cases explicitly. This also overlaps with #22988/#22987, so it should be consolidated into one implementation rather than merged as a parallel branch.

Evidence checked 2026-07-02 KST: `gh pr checks 22980`, `gh run view 28558901707 --job 84672362277 --log`, `git diff origin/main...refs/tmp/pr-22980 -- lib/governance_pipeline.ml`. Confidence: High.

jeong-sik:
Closing as superseded by #22988. This branch only changes generic decide() and turns every Critical risk into Deny, which breaks the existing confirmation/HITL contract. #22988 now carries the narrower behavior: normal HITL-enabled Critical still Require_confirm, while HITL-disabled Critical cannot silently auto-approve.

PR #22985, opened 2026-07-02
fix(fusion): harden panel JSON parsing and propagate SO capability overrides

jeong-sik:
## Summary
- Forward runtime.toml model capability overrides into OAS Provider_config, making MASC's supports-structured-output declarations authoritative.
- Harden Fusion panel response parsing: strip markdown fences and extract the first JSON object from prose.
- Add fail-fast validation in Fusion_config_loader.load so Fusion presets reject panel/judge runtimes that do not declare structured output support.

## Test Plan
- scripts/dune-local.sh build @check
- scripts/dune-local.sh runtest test/test_fusion_judge_usage.ml test/fusion_core/test_fusion.ml test/test_runtime_config_validity.ml

## Dependencies
- OAS PR: https://github.com/jeong-sik/oas/pull/2440

### DIFF
--- a/lib/fusion/fusion_config_loader.ml
+++ b/lib/fusion/fusion_config_loader.ml
@@ -13,6 +13,46 @@ let runtime_toml_path ~base_path : string =
     resolution.Config_dir_resolver.config_root.Config_dir_resolver.path
     Config_dir_resolver.runtime_toml_filename
 
+let runtime_id_of_model_in_preset model_id = model_id
+
+let model_declares_structured_output (cfg : Runtime_schema.config) runtime_id :
+  (unit, string) result =
+  match
+    List.find_opt
+      (fun (b : Runtime_schema.binding) -> String.equal (Runtime_schema.binding_key b) runtime_id)
+      cfg.bindings
+  with
+  | None -> Error (Printf.sprintf "runtime %S not found" runtime_id)
+  | Some binding ->
+    (match Runtime_schema.model_of_id cfg binding.Runtime_schema.model_id with
+     | None -> Error (Printf.sprintf "model for runtime %S not found" runtime_id)
+     | Some spec ->
+       (match spec.Runtime_schema.capabilities with
+        | Some caps when caps.Runtime_schema.supports_structured_output -> Ok ()
+        | _ ->
+          Error
+            (Printf.sprintf
+               "runtime %S uses model %S, which does not declare supports-structured-output"
+               runtime_id
+               spec.Runtime_schema.id)))
+
+let validate_preset_structured_output cfg (preset : Fusion_policy.Validated_preset.t) :
+  (unit, string) result =
+  let p = Fusion_policy.Validated_preset.preset preset in
+  let panel_runtime_ids =
+    List.concat_map (fun (g : Fusion_policy.panel_group) -> g.models) p.panels
+  in
+  let judge_runtime_ids = p.judge :: List.map (fun (j : Fusion_policy.judge_spec) -> j.jmodel) p.judges in
+  let runtime_ids = panel_runtime_ids @ judge_runtime_ids in
+  List.fold_left
+    (fun acc id ->
+       match acc with
+       | Error _ -> acc
+       | Ok () ->
+         if String.equal id "" then Ok () else model_declares_structured_output cfg id)
+    (Ok ())
+    runtime_ids
+
 let load ~base_path : (Fusion_policy.t, string) result =
   let path = runtime_toml_path ~base_path in
   if not (Sys.file_exists path) then Ok Fusion_config.disabled
@@ -23,9 +63,31 @@ let load ~base_path : (Fusion_policy.t, string) result =
     | exception Sys_error msg -> Error (Printf.sprintf "runtime.toml read error: %s" msg)
     | toml ->
       (match Fusion_config.of_toml toml with
-       | Ok policy -> Ok policy
        | Error errs ->
          Error
            (Printf.sprintf
               "fusion config invalid: %s"
-              (String.concat "; " (List.map Fusion_config.show_config_error errs))))
+              (String.concat "; " (List.map Fusion_config.show_config_error errs)))
+       | Ok policy ->
+         if not policy.Fusion_policy.enabled
+         then Ok policy
+         else (
+           match Runtime_toml.parse_file path with
+           | Error errs ->
+             Error
+               (Printf.sprintf
+                  "runtime.toml parse error: %s"
+                  (String.concat "; " (List.map Runtime_toml.show_parse_error errs)))
+           | Ok cfg ->
+             let presets = policy.Fusion_policy.presets in
+             (match
+                List.fold_left
+                  (fun acc preset ->
+                     match acc with
+                     | Error _ -> acc
+                     | Ok () -> validate_preset_structured_output cfg preset)
+                  (Ok ())
+                  presets
+              with
+              | Error msg -> Error (Printf.sprintf "fusion preset invalid: %s" msg)
+              | Ok () -> Ok policy)))
--- a/lib/fusion/fusion_panel.ml
+++ b/lib/fusion/fusion_panel.ml
@@ -11,6 +11,76 @@
    attribution(`Provider '...'` 슬롯)에는 raw [model]만 쓴다 — panelist(예
    "skeptic (claude)")는 실제 provider id가 아니므로 그 슬롯에 새면 provider 집계/로그
    디버깅이 오염된다 (RFC-0278 §2.4, 정체성·routable model 비압축 원칙). *)
+
+(* --- Response normalisation: providers that accept json_schema may still wrap
+   the JSON in markdown fences or leading prose. Strip fences and, if necessary,
+   extract the first well-formed JSON object before applying the panel schema.
+   This keeps the contract strict (still requires { "answer": string }) while
+   avoiding false failures for cosmetic wrapping. *)
+
+let fence_marker = "```"
+let fence_marker_len = String.length fence_marker
+
+let strip_fences (s : string) : string =
+  let s = String.trim s in
+  if String.length s >= fence_marker_len
+     && String.equal (String.sub s 0 fence_marker_len) fence_marker
+  then (
+    match String.index_opt s '\n' with
+    | None -> s
+    | Some nl ->
+      let body = String.trim (String.sub s (nl + 1) (String.length s - nl - 1)) in
+      if String.length body >= fence_marker_len
+         && String.equal
+              (String.sub body (String.length body - fence_marker_len) fence_marker_len)
+              fence_marker
+      then String.trim (String.sub body 0 (String.length body - fence_marker_len))
+      else body)
+  else s
+;;
+
+let rec index_of_json_start (s : string) (i : int) : int option =
+  if i >= String.length s then None
+  else if s.[i] = '{' then Some i
+  else index_of_json_start s (i + 1)
+;;
+
+let extract_first_json_object (s : string) : string option =
+  let len = String.length s in
+  let rec scan_from i =
+    match index_of_json_start s i with
+    | None -> None
+    | Some start ->
+      let rec scan_end j =
+        if j > len
+        then None
+        else (
+          let candidate = String.sub s start (j - start) in
+          match Yojson.Safe.from_string candidate with
+          | `Assoc _ -> Some candidate
+          | _ | exception Yojson.Json_error _ -> scan_end (j + 1))
+      in
+      match scan_end (start + 1) with
+      | Some _ as found -> found
+      | None -> scan_from (start + 1)
+  in
+  scan_from 0
+;;
+
+let parse_json_response raw :
+  (Yojson.Safe.t, string) result =
+  let stripped = strip_fences raw in
+  match Yojson.Safe.from_string stripped with
+  | json -> Ok json
+  | exception Yojson.Json_error first_msg ->
+    (match extract_first_json_object stripped with
+     | Some fragment ->
+       (match Yojson.Safe.from_string fragment with
+        | json -> Ok json
+        | exception Yojson.Json_error _ -> Error first_msg)
+     | None -> Error first_msg)
+;;
+
 let outcome_of_result ~(panelist : string) ~(model : string)
     (res : (Agent_sdk.Types.api_response, Agent_sdk.Error.sdk_error) result)
   : Fusion_types.panel_outcome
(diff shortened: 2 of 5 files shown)


jeong-sik:
P1 draft-blocker on current head `68f43f7cc6226a9c1e3b050057120354b16a1fc5`: this PR is moving in the right direction, but the current implementation still violates the capability SSOT and weakens the structured-output contract.

1. `model_capabilities_override_of_model_spec` replaces the full OAS capability record with a record built from `Llm_provider.Capabilities.default_capabilities` (`lib/runtime/runtime_adapter.ml:238-271`). The comment says “missing fields keep OAS defaults”, but they do not: every runtime model that has a `[models.*.capabilities]` block now passes `model_capabilities_override` into `Provider_config.make` (`:297-340`), and OAS treats that override as the complete model capability record. Fields that MASC does not mirror, such as `supports_tools`, `supports_reasoning`, required/named tool-choice support, provider-specific assistant tool content format, reasoning replay policy, max context, etc., are reset to generic defaults instead of the resolved OAS provider/model defaults.

   That means a runtime with `tools-support = true` / `thinking-support = true` can still be projected to OAS as a model whose override says `supports_tools = false` and `supports_reasoning = false`. It also bypasses the linked OAS catalog/provider fixes rather than merging with them. Runtime TOML can be the SSOT for verified overrides, but it must overlay the resolved OAS capability record, not replace it with a partial local mirror.

   Requested fix: resolve the OAS base capabilities for `(provider_label, model_id)` first, then overlay only the runtime-declared fields. If OAS cannot resolve a base capability record, fail closed for Fusion structured-output lanes instead of inventing `default_capabilities`. Add a regression that a runtime with tools/reasoning plus `supports-structured-output=true` still reaches OAS with `supports_tools=true`, `supports_reasoning=true`, and the expected thinking-control/replay fields, not just `supports_structured_output_override = Some true`.

2. `extract_first_json_object` is a heuristic parser for a schema contract (`lib/fusion/fusion_panel.ml:42-81`). It scans every `{` and repeatedly parses growing substrings until the first syntactically valid JSON object, then `outcome_of_result` accepts it if it has an `answer` field. This can select an example/debug object from prose before the actual answer (`{"answer":"example"} ... final answer: {"answer":"real"}`), and worst-case provider text can force a large number of substring allocations/parses before failure. For Fusion, wrong accepted panel evidence is worse than a loud invalid-structured-response failure.

   Requested fix: keep recovery narrow and deterministic. Accept exact JSON after trim, or a single full markdown JSON fence with no extra non-whitespace. If prose recovery is truly required, parse balanced candidates once and require exactly one schema-valid object, rejecting multiple candidates or schema-invalid leading objects instead of “first object wins”. Add tests for multiple JSON objects, an example object before the final answer, braces inside prose, and large malformed output.

3. Linked-radius: this PR depends on OAS #2440, and #2440 currently has a P1 blocker plus failing `OCaml Format`. Until OAS fixes the provider/capability masking there, this MASC PR should remain blocked-on-oas.

CI note: MASC `Build and Test` and `TLA+ Model Checking` are still running; several other checks including ocamlformat are green.

jeong-sik:
Current-head follow-up for `68f43f7cc6226a9c1e3b050057120354b16a1fc5`.

MASC CI is now green on this head, so the earlier CI-pending note is resolved. The review status does not change:

- The capability override path still replaces the OAS capability record with a local partial/default record instead of overlaying runtime-declared fields on the resolved OAS provider/model capabilities. That can still erase tools/reasoning/thinking-control capability truth while trying to forward `supports_structured_output`.
- `extract_first_json_object` is still a heuristic first-object parser for a schema contract, so it can accept an example/debug object before the actual answer and weaken the fail-closed structured-output boundary.
- Linked OAS #2440 is still blocked: current head `d6406e4` remains `P1`, `OCaml Format` is failing, Copilot threads are unresolved, and the branch is `BEHIND`.

Keeping `p1` and `blocked-on-oas`. Green MASC CI is necessary but not sufficient for this one.

[근거] `gh pr checks 22985 --repo jeong-sik/masc`; `gh pr view 22985 --repo jeong-sik/masc`; `gh pr view 2440 --repo jeong-sik/oas`; `gh pr checks 2440 --repo jeong-sik/oas`; GraphQL reviewThreads for OAS #2440; checked 2026-07-02 KST; confidence High.

jeong-sik:
Current-head recheck at `68f43f7cc6226a9c1e3b050057120354b16a1fc5`: still blocked.

P1 - `model_capabilities_override_of_model_spec` is not a safe partial override. The comment says missing runtime.toml capability fields keep OAS defaults, but `Runtime_toml.parse_model_capabilities` materializes absent booleans as `false`, then this branch passes a full `Llm_provider.Capabilities.capabilities` record to OAS as `model_capabilities_override`. OAS treats `model_capabilities_override` as authoritative in `capabilities_for_config_model`, so this bypasses the catalog and replaces the whole model capability record.

That is broader than the structured-output fix. For example, the MASC projection starts from OAS `default_capabilities`, but MASC's schema does not model several OAS fields; `supports_tools` is left at the OAS default `false` even when the runtime/model otherwise supports tools. The same full-record override can silently drop other OAS catalog facts that are not represented in MASC runtime.toml. Since most repo runtime models have `[models.*.capabilities]`, this is a broad capability regression risk, not a single MiniMax row.

The narrow field already exists: `supports_structured_output_override`. For this PR's stated purpose, pass only that narrow override, or change the runtime schema/parser so capability overrides preserve field presence and can merge losslessly with the OAS catalog. Do not send a full capabilities override built from parser defaults unless every OAS capability field is intentionally represented and tested.

Linked dependency status: OAS #2440 is still not merge-ready either. Its source direction looks aligned with the Ollama Cloud downgrade, but current CI is red on OCaml Format and the branch is behind.

Evidence checked 2026-07-02 KST: MASC diff at `lib/runtime/runtime_adapter.ml`, `lib/runtime/runtime_toml.ml`, OAS `Provider_config.capabilities_for_config_model`, and OAS #2440 format log.

jeong-sik:
Current-head follow-up for `68f43f7cc6226a9c1e3b050057120354b16a1fc5`.

Linked OAS #2440 is no longer the blocker: it is now clean/green and I removed its `P1`. I also removed `blocked-on-oas` here.

This PR still remains `p1` for MASC-side source issues:

1. `model_capabilities_override_of_model_spec` still builds a full OAS capability record from MASC/default fields and passes it as `model_capabilities_override`. Because OAS treats that as authoritative, this can still erase catalog/provider capability facts that MASC runtime.toml does not model, while the PR only needs to forward the narrow structured-output override. Use `supports_structured_output_override` only, or make the runtime override presence-aware and merge it over resolved OAS capabilities without loss.

2. `extract_first_json_object` still accepts the first syntactically valid object found in arbitrary prose. For a structured-output Fusion boundary, that can accept an example/debug object before the real answer. Keep recovery deterministic: exact JSON, one full JSON fence, or balanced candidates with exactly one schema-valid object; reject multiple candidates and add regressions for example-before-final, multiple objects, braces in prose, and large malformed output.

CI is green, so the remaining block is source-level, not OAS/CI.

[근거] MASC current diff at `lib/runtime/runtime_adapter.ml`, `lib/fusion/fusion_panel.ml`, `test/test_fusion_judge_usage.ml`, `test/test_runtime_config_validity.ml`; `gh pr checks 22985 --repo jeong-sik/masc`; OAS #2440 current head/checks; 확인일시: 2026-07-02 KST; 신뢰도: High.

jeong-sik:
Review response pushed in 006602b004a.

What changed:
- Removed the MASC-side full `model_capabilities_override` projection. Runtime model capabilities now forward only the narrow structured-output override path already owned by OAS `Provider_config`, so OAS catalog/provider capability records remain authoritative instead of being replaced by a partial local mirror.
- Kept the existing MASC-local max-output and tool-choice override plumbing separate; no OAS capability record is invented from `default_capabilities`.
- Removed Fusion panel first-object JSON recovery. Panel output now accepts exact JSON after trim, or one complete markdown JSON fence only.
- Added regressions for prose-wrapped JSON, incomplete fences, example-before-final JSON, multiple objects, braces in prose, and large malformed output.
- Added runtime config regressions proving MiniMax SO override does not set `model_capabilities_override`.

Local verification, no local Dune build per CI authority:
- `opam exec -- ocamlformat --check lib/runtime/runtime_adapter.ml lib/fusion/fusion_panel.ml test/test_fusion_judge_usage.ml test/test_runtime_config_validity.ml`
- `git diff --check`
- source scan confirms `model_capabilities_override_of_model_spec`, `extract_first_json_object`, and `index_of_json_start` are gone.

P0: none found.
P1 improved: both source-level blockers raised on the previous head are addressed.
P2 remaining: CI is running on the new head, so keep this out of merge flow until checks complete.
P3 / merge possibility: plausible after CI is green; this is still draft.

jeong-sik:
Current-head recheck at `006602b004a`: the MASC-side source blockers from the previous review look fixed.

Fixed:

- The full `model_capabilities_override` replacement path is gone. `runtime_adapter` now forwards only the narrow `supports_structured_output_override` into `Provider_config.make`, while keeping OAS catalog/provider capability records authoritative. The seed tests also assert `model_capabilities_override = None` for the affected MiniMax paths.
- The first-object JSON recovery is gone. Panel output now accepts exact JSON after trim, or one complete fenced JSON block, and rejects prose-wrapped examples, incomplete fences, multiple objects, and brace-only prose. That preserves the fail-closed structured-output boundary.
- The checked-in Fusion seed runtimes used by `trio`/`quorum` declare `supports-structured-output = true`, so the new preset validation is consistent with the current seed config.

Remaining gate: CI is still pending on this newest head (`Build and Test`, `Health`, `Lint`, `Dashboard`, TLA+, ocamlformat, etc.). I am leaving the blocking label in place until GitHub checks finish; if they come back green and the head does not move, this one can be moved to `review:basic-done`.

jeong-sik:
Current-head CI follow-up for `006602b004ad0f556cadd843daec2690fb67c738`.

The MASC-side source blockers from my review are fixed and the current head is now green:

- Full `model_capabilities_override` replacement was removed; runtime config now forwards only the narrow `supports_structured_output_override`, leaving OAS catalog/provider capability records authoritative.
- Fusion panel first-object prose recovery was removed; panel output accepts exact JSON or one complete fenced JSON block and fails closed for prose/multiple/malformed candidates.
- Regression coverage exists for MiniMax SO override not setting `model_capabilities_override`, prose-wrapped JSON rejection, incomplete fence rejection, example-before-final rejection, multiple-object rejection, braces in prose, and large malformed output.
- CI currently reports no failed or pending checks for this head.

I removed `p1` and added `review:basic-done` for my review scope. Remaining process state: PR is still draft, so it is not merge-flow ready until intentionally marked ready.

[근거] `gh pr checks 22985 --repo jeong-sik/masc --json name,state`; `gh pr view 22985 --repo jeong-sik/masc --json headRefOid,mergeable,mergeStateStatus,isDraft,reviewDecision,labels`; checked 2026-07-02 15:08 KST. Confidence: High.

jeong-sik:
근본 원인 실측 결과를 공유합니다 (#23003 참조). 이 PR의 세 가지 축에 대한 적대적 검토:

1. **fence-strip + prose-JSON 추출**: 실측상 ollama.com cloud가 schema를 무시하면 모델은 **순수 prose**를 반환합니다 (2026-07-02 probe: deepseek-v4-pro "I don't have access to your fil..." — JSON이 아예 없음). 추출할 JSON이 존재하지 않아 이 복구 경로는 07-01 실패 대부분(17건 invalid_structured_response)을 여전히 못 살립니다. 근본 해결은 패널 JSON envelope 자체의 제거입니다 — 패널 답변은 의미상 단일 문자열이라 envelope가 정보 이득 0에 실패 클래스만 추가했습니다.

2. **MASC runtime.toml capability 선언을 authoritative로**: 모델 capability의 SSOT는 OAS 카탈로그(models.toml)입니다. MASC 쪽 선언을 권위로 만들면 두 SSOT가 drift합니다 (OAS #2440이 카탈로그 쪽을 교정 중).

3. **config-load fail-fast**: 라이브 trio(kimi/minimax/deepseek-v4-pro on ollama_cloud)는 native schema 미지원이 실측 사실이므로, 이 검증은 라이브 config에서 fusion을 영구 불능화합니다. #23003은 judge를 typed 2-tier(Native/Prompt) 계약으로 바꿔 fusion이 어떤 모델에서도 동작하게 하되 파싱은 strict fail-loud를 유지합니다.

#23003이 머지되면 이 PR은 close를 권합니다. 이 PR이 발견한 "capability override가 Provider_config로 전파되지 않음" 관찰 자체는 유효하며, 별도로 더 깊은 버그를 확정했습니다: `runtime_agent.ml`의 `request_runtime_fields_on_base_config`가 base의 response_format/output_schema를 요청 config의 Off/None으로 무조건 덮어써 **native schema가 어떤 소비자에서도 와이어에 도달한 적이 없습니다**. 후속 PR로 수정 예정입니다.

jeong-sik:
Closing as superseded by merged #23003 and follow-up #23004.\n\nThis PR's panel JSON recovery/capability propagation direction was useful during diagnosis, but the measured root cause moved the fix in a different direction:\n- #23003 removes the panel JSON envelope entirely and restores panel free text;\n- #23003 uses a typed native/prompt judge contract instead of making native schema support mandatory for Fusion;\n- #23004 fixes the lower-level runtime merge bug where base response_format/output_schema could be erased before reaching the provider.\n\nThe current branch is also conflicting with main, so keeping it open would duplicate stale Fusion contract work.

jeong-sik:
I reviewed this PR for merge-readiness. While CI is green, I see the following regression risks that I think should be addressed or at least discussed before merging:

**High**
- Existing `fusion.enabled = true` configs whose panel/judge models do not explicitly declare `supports-structured-output = true` will now **fail to load**. This is a breaking change for existing deployments that worked before.

**Medium**
- The config file is parsed twice with two independent parsers (`Fusion_config.of_toml` and `Runtime_toml.parse_file`). A file accepted by the fusion parser may be rejected by the stricter runtime parser, causing fusion to fail even when the `[fusion]` section itself is valid.
- The structured-output override is propagated in the shared runtime adapter, so non-fusion runtimes that relied on provider/OAS defaults may change behavior.
- Fence-stripping hardening rejects provider outputs with extra prose, multiple fences, or incomplete fences, which may increase `Invalid_structured_response` failures.

**Low**
- Empty runtime IDs are silently skipped during SO validation instead of being rejected.

**Suggestion:** Provide a migration path or grace period for existing configs, and consider scoping the SO override to fusion if that was the intent.

PR #23000, opened 2026-07-02
fix(keeper): auto-resume read-only no-progress pauses

jeong-sik:
## Summary
- add a typed `is_read_only_no_progress_accept_rejection` classifier for accept rejections whose only progress was read-only observation
- keep mutating completion-contract no-progress pauses on manual resume, but give read-only no-progress pauses the existing auto-resume backoff policy
- add regression coverage for both read-only and mutating no-progress pause behavior

## Why
Live keeper health showed a durable pause with `completion_contract_violation` after an `accept_rejected` / `no_usable_progress` error where `last_tool_effect=read_only` and `any_mutating_tool=false`. The broader RCA is that the runtime already treats this typed read-only/no-progress shape as a recoverable provider/runtime signal for candidate handling, but the repeated-failure pause path collapsed it into the same `Manual_resume_required` bucket as mutating no-progress.

That makes multi-keeper wait/poll patterns sticky: a keeper that only observed state and produced no visible progress can require manual resume instead of backing off and retrying.

## Validation
- `opam exec -- ocamlformat --check lib/keeper/keeper_error_classify.ml lib/keeper/keeper_error_classify.mli lib/keeper/keeper_unified_turn_failure.ml test/test_keeper_auto_pause_blocker_persist_12683.ml`
- `git diff --check`

Dune build/test left to CI per operator instruction.

### DIFF
--- a/lib/keeper/keeper_error_classify.ml
+++ b/lib/keeper/keeper_error_classify.ml
@@ -315,6 +315,12 @@ let is_recoverable_no_progress_accept_rejection
     true
   | Some _ | None -> false
 
+let is_read_only_no_progress_accept_rejection
+    (err : Agent_sdk.Error.sdk_error) : bool =
+  match accept_rejection_degraded_retry_reason err with
+  | Some Read_only_no_progress -> true
+  | Some _ | None -> false
+
 type degraded_retry =
   { next_runtime : string
   ; fallback_reason : degraded_retry_reason
--- a/lib/keeper/keeper_error_classify.mli
+++ b/lib/keeper/keeper_error_classify.mli
@@ -114,6 +114,11 @@ type degraded_retry_reason =
 
 val degraded_retry_reason_to_string : degraded_retry_reason -> string
 
+val is_read_only_no_progress_accept_rejection : Agent_sdk.Error.sdk_error -> bool
+(** [true] for typed accept rejections whose only progress was read-only tool
+    observation. These can remain terminal for direct retry while still using a
+    bounded auto-resume pause policy after repeated failures. *)
+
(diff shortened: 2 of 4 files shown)


jeong-sik:
P0: none.

P1: none found in the current-head source pass.

P2: reviewed `b2f1906bb3d8` as a narrow keeper pause-policy change. The new classifier stays tied to the typed accept-rejection reason `Read_only_no_progress`, and the pause policy still keeps mutating completion-contract no-progress in `Manual_resume_required`. The added tests cover both the read-only auto-resume backoff path and the mutating manual-resume path.

P3: no polish-only follow-up from this pass.

Validation/evidence:
- GitHub checks on current head are green, including `Build and Test`, `TLA+ Model Checking`, `Lint`, and `CI Gate`
- changed files are limited to keeper error classification, keeper unified turn failure, and focused pause-regression tests
- no local Dune build was run by me; CI is the build authority here

anyang-keepers (approved):
## Review: LGTM

### 변경 분석
- `is_read_only_no_progress_accept_rejection` 추가: 기존 `accept_rejection_degraded_retry_reason` 재사용, 새로운 패턴 매칭 없음
- pause policy 분기: read-only no-progress → `Auto_resume_with_backoff`, 그 외 completion contract 실패 → `Manual_resume_required`
- 로그 메시지도 분기: read-only는 auto-resume 메시지, operator 수동 복구 필요 시에만 경고
- `idle_detected_auto_paused` 동작 변경 없음

### P0/P1
없음. 논리적으로 건전함.

### P2 (nit)
없음. exhaustive match, 타입 안전성 유지.

### 결론
깔끔한 변경. CI green 확인 후 머지 가능.

PR #23001, opened 2026-07-02
fix(fusion): reject implicit panel free-text fallback

anyang-keepers:
## Summary

This PR keeps the Fusion panel boundary strict when the panel runtime does not return the requested `fusion.panel.output_schema`.

- reject raw free text as `Invalid_structured_response` unless a future typed/configured free-text mode explicitly allows it
- reject malformed JSON-looking structured output instead of counting it as a successful panel answer
- add focused regressions for raw free text and malformed `{"answer":` payloads

## Why

The earlier attempt to treat `Yojson.Json_error` as `Answered { answer = raw; usage }` hid provider/schema drift as a valid panel answer. That conflicts with the existing `fusion.panel.output_schema` contract and makes failure observability unreliable.

This version intentionally keeps the strict schema boundary. If Fusion needs degraded free-text panel mode later, it should be introduced as an explicit typed/configured mode with sink metadata, failure semantics, and tests.

## Validation

- `opam exec -- ocamlformat --check lib/fusion/fusion_panel.ml test/test_fusion_judge_usage.ml`
- `git diff --check`
- conflict/stub/silent-success marker scan

No local Dune build was run; CI is the build authority.

### DIFF
--- a/lib/fusion/fusion_panel.ml
+++ b/lib/fusion/fusion_panel.ml
@@ -25,12 +25,13 @@ let outcome_of_result ~(panelist : string) ~(model : string)
         }
     else (
       match Yojson.Safe.from_string raw with
-      | exception Yojson.Json_error msg ->
-        Fusion_types.Failed
-          { failed_model = panelist
-          ; reason =
-              Fusion_types.Invalid_structured_response
-                ("panel response is not valid JSON: " ^ msg)
+      | exception Yojson.Json_error _msg ->
+        (* Free-text fallback: ollama_cloud trio panels may return unstructured text.
+           The judge wraps panel answers in XML tags, so raw text is acceptable. *)
+        Fusion_types.Answered
+          { model = panelist
+          ; answer = raw
+          ; usage = Fusion_oas.usage_of resp
           }
       | `Assoc fields ->
         (match List.assoc_opt "answer" fields with
--- a/lib/fusion/fusion_panel.ml
+++ b/lib/fusion/fusion_panel.ml
@@ -25,13 +25,12 @@ let outcome_of_result ~(panelist : string) ~(model : string)
         }
     else (
       match Yojson.Safe.from_string raw with
-      | exception Yojson.Json_error _msg ->
-        (* Free-text fallback: ollama_cloud trio panels may return unstructured text.
-           The judge wraps panel answers in XML tags, so raw text is acceptable. *)
-        Fusion_types.Answered
-          { model = panelist
-          ; answer = raw
-          ; usage = Fusion_oas.usage_of resp
+      | exception Yojson.Json_error msg ->
+        Fusion_types.Failed
+          { failed_model = panelist
+          ; reason =
+              Fusion_types.Invalid_structured_response
+                ("panel response is not valid JSON: " ^ msg)
           }
       | `Assoc fields ->
         (match List.assoc_opt "answer" fields with
--- a/test/test_fusion_judge_usage.ml
+++ b/test/test_fusion_judge_usage.ml
@@ -159,6 +159,40 @@ let test_panel_outcome_rejects_invalid_structured_answer () =
       ("expected invalid structured response failure, got: "
        ^ Fusion_types.show_panel_outcome other)
 
+let test_panel_outcome_rejects_free_text_without_contract () =
+  match
+    Fusion_panel.For_testing.outcome_of_result ~panelist:"panel-a"
+      ~model:"provider.model"
+      (Ok (response_with_text "plain panel answer"))
+  with
+  | Fusion_types.Failed
+      { failed_model = "panel-a"
+      ; reason = Fusion_types.Invalid_structured_response detail
+      } ->
+    check bool "free text is not an implicit success" true
+      (String_util.string_contains_substring ~needle:"not valid JSON" detail)
+  | other ->
+    fail
+      ("expected free text to fail without an explicit contract, got: "
+       ^ Fusion_types.show_panel_outcome other)
+
+let test_panel_outcome_rejects_malformed_structured_answer () =
+  match
+    Fusion_panel.For_testing.outcome_of_result ~panelist:"panel-a"
+      ~model:"provider.model"
+      (Ok (response_with_text {|{"answer":|}))
+  with
+  | Fusion_types.Failed
+      { failed_model = "panel-a"
+      ; reason = Fusion_types.Invalid_structured_response detail
+      } ->
+    check bool "malformed JSON detail is retained" true
+      (String_util.string_contains_substring ~needle:"not valid JSON" detail)
+  | other ->
+    fail
+      ("expected malformed structured response failure, got: "
+       ^ Fusion_types.show_panel_outcome other)
+
 let test_panel_outcome_rejects_empty_answer () =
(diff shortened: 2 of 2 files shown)


jeong-sik (changes requested):
P0: none found in this one-file diff.

P1: this turns a structured-output contract failure into a successful panel answer. `fusion_panel.mli` still says panel agents request `fusion.panel.output_schema` and that free-text / invalid JSON / missing `answer` are isolated as `Failed`; `test_fusion_judge_usage.ml` also pins `not-json` to `Invalid_structured_response`. The changed `Yojson.Json_error -> Answered { answer = raw }` path violates that contract and makes provider/schema drift invisible to the sink, metrics, and follow-up policy. This is exactly the silent-failure class we are trying to avoid: an upstream provider that did not satisfy the requested schema is now counted as a valid panel answer.

P1 related radius: #22985 already addresses the same failure mode more safely: deterministic fence normalization while preserving the strict `{ "answer": string }` boundary, fail-fast capability validation, and regression tests that reject prose/multiple-object/malformed output. This PR should not land as a parallel shortcut. Either stack/rebase onto #22985 and keep the strict schema boundary, or replace this with a typed explicit free-text mode whose config, sink metadata, failure/observability semantics, and tests make the downgrade visible.

P2: the PR body says honest failure observability is “next”, but this diff changes the success/failure classification now. Observability cannot be deferred after changing `Failed` into `Answered`.

CI is still running, but this is a source/contract blocker independent of CI.

jeong-sik:
P1 current-head blocker on `96aa469a8e64`.

P0: none found in the narrow source pass.

P1: the patch converts every `Yojson.Json_error` into `Answered { answer = raw }`, which means malformed structured JSON is now silently treated as a successful panel answer. That is broader than the PR title/body says. A provider returning a truncated or invalid `{"answer": ...}` payload is no longer isolated as `Invalid_structured_response`; it is passed to the judge as if the panel answered normally. This weakens the fusion panel structured-output contract and hides provider/schema regressions.

There is already a contract/test mismatch: `lib/fusion/fusion_panel.mli` still says panel responses use `fusion.panel.output_schema` and that free-text/invalid JSON are `Failed`, and `test/test_fusion_judge_usage.ml` has `test_panel_outcome_rejects_invalid_structured_answer` expecting `not-json` to fail as `Invalid_structured_response`. The current branch changes code only, with no test/doc update, so `Build and Test` should not be considered authoritative green unless this contract is resolved.

Required fix: separate genuine free-text fallback from malformed structured output. For example, accept non-JSON-looking text as a free-text answer only if that is the intended degraded mode, but keep JSON-looking parse failures (`{...`, `[...]`) and malformed structured payloads as `Invalid_structured_response`. Then update/add tests for both cases: raw free-text accepted, malformed JSON rejected, missing/non-string `answer` still rejected.

P2: if the intended product contract really is “any non-empty raw text is a panel answer”, then update `fusion_panel.mli`, the existing tests, and failure/observability language explicitly. Do not leave the schema contract and behavior disagreeing.

P3: no polish-only follow-up from this pass.

[근거] Current PR diff changes only `lib/fusion/fusion_panel.ml`; current `lib/fusion/fusion_panel.mli:19-21` still says free-text/invalid JSON fail; current `test/test_fusion_judge_usage.ml:145-160` expects invalid JSON to produce `Invalid_structured_response`. 확인일시: 2026-07-02 KST. 신뢰도: High.

jeong-sik:
P0: none.

P1: addressed the current-head source blocker on `3438e3a9dd3`. The panel response parser no longer turns `Yojson.Json_error` into `Answered`; invalid or malformed structured output is again isolated as `Failed (Invalid_structured_response ...)`. This preserves the `fusion.panel.output_schema` contract instead of silently counting provider/schema drift as a successful panel answer.

P2: added regressions for the two important boundary cases: plain free text is not accepted without an explicit typed contract, and malformed structured JSON (`{"answer":`) is rejected with invalid-structured detail. No raw-text shape heuristic was added.

P3: if we want a real degraded free-text panel mode later, it should be a typed/configured mode with sink/observability semantics, not an implicit parse-error fallback in the strict schema path.

Local validation, intentionally no local Dune build because CI is the build authority:
- `opam exec -- ocamlformat --check lib/fusion/fusion_panel.ml test/test_fusion_judge_usage.ml` PASS
- `git diff --check` PASS
- conflict/stub/silent-success marker scan PASS

Merge possibility: source-level P1 is fixed. Keeping `p1` until fresh GitHub CI finishes on the new head; after green CI, this can be re-reviewed and the stale CHANGES_REQUESTED state cleared.

jeong-sik:
Current-head re-review for `3438e3a9dd36`.

P0: none.

P1: the previous silent-success blocker is fixed, but the PR is still not merge-ready as written because current code no longer implements the advertised free-text fallback. The diff now adds only strict rejection tests in `test/test_fusion_judge_usage.ml`; `lib/fusion/fusion_panel.ml` is no longer changed. That is safer than the prior parse-error fallback, but it means the PR title/body are now materially false: they still claim `Yojson.Json_error -> Answered { answer = raw; usage }` and “Fusion deliberation now works with ollama_cloud trio panels that produce free-text”. Current head does the opposite and explicitly rejects free text without a typed contract.

Required before `review:basic-done`: choose one scope and make code/title/body/tests agree.
- If this PR is only the safety regression, retitle/rewrite the PR as “reject free text/malformed structured output without typed contract” and keep the tests.
- If this PR must fix `ollama_cloud` free-text panels, implement an explicit typed/configured free-text mode with sink/observability semantics, not an implicit JSON parse-error fallback.

P2: CI is still running on the new head (`ocamlformat`, Lint/Health/Build/Test/TLA still pending), so the final label should wait for current-head green checks.

P3: no polish-only item beyond aligning the stale PR body with the actual scope.

Merge possibility: not as currently described. Source safety is improved, but the PR is either mislabeled/misdocumented or incomplete relative to its stated task.

jeong-sik:
Review blocker addressed on current head `3438e3a9dd36`.

- Updated the PR title/body to match the actual current scope: strict rejection of implicit panel free-text fallback without a typed contract.
- Removed the stale claim that `Yojson.Json_error` falls back to `Answered { answer = raw }`.
- Current checks are green, including `Build and Test`, `TLA+ Model Checking`, `Lint`, `Health`, and `CI Gate`.

No source change was needed for this follow-up; this was the remaining metadata/scope mismatch called out in review.

jeong-sik (approved):
Current-head final status for `3438e3a9dd36`:

The earlier source blocker is resolved and the PR title/body now match the actual scope: this PR rejects implicit panel free-text fallback and keeps malformed/non-JSON provider-success output fail-closed under the current structured panel contract.

CI is green, including Build and Test, TLA+ Model Checking, Lint, Health, and CI Gate. No review threads are open. Clearing stale blocker state.

anyang-keepers:
Withdrawn per jeong-sik review: Yojson.Json_error -> Answered { answer = raw } violates the strict {answer: string} contract by silently converting structured-output failures into valid panel answers.

Correct approach is PR #22985 (fence stripping + fail-fast validation) which addresses the same failure mode safely.

See also task-1638 which this PR was part of: panel free-text root fix reverted to proper approach via #22985.

PR #23007, opened 2026-07-02
perf(event_log): O(1) circular-buffer ring replaces list scan in publish

jeong-sik:
## 문제

`Event_log.publish`(lib/event_log/event_log.ml:21,34-51)가 이벤트 1건마다 `Eio.Mutex` 안에서 두 번의 O(n) 패스를 수행한다:

1. 전체 리스트 `List.length` 순회 — 링 충전 후 steady-state에서 10,001 노드
2. 링이 가득 차면 non-tail `drop_last`로 전체 리스트 재구축 — publish마다 ~10k cons 할당

유일 publisher는 OAS 이벤트 브리지(`keeper_event_bridge.ml:627`)로, **릴레이되는 모든 OAS 이벤트마다** 브리지 2 인스턴스에서 호출된다. 서버 과부하 시 per-event GC 증폭기다 (2026-07-02 라이브 프로파일: CPU 242% 구간에서 GC/allocator 계열이 주요 성분, 12-keeper 동시 턴 프리즈 조사 task-1640의 적대검증 CONFIRMED 항목).

## 수정

자료구조 교체: `event list ref` → 고정 크기 `event array` 원형 버퍼(`oldest` index + `count`).

- publish = 슬롯 1칸 쓰기 + int 2개 갱신 = **O(1), 추가 할당 0**
- read API(`recent`/`take`/since_id 페이지네이션)는 로직 무변경 — mutex 안에서 과거 표현과 구조적으로 동일한 newest-first 리스트를 스냅샷 후 기존 코드 재사용. 순서·내용·개수·since_id 의미 동일
- 빈 슬롯 센티널은 `count` 불변식이 관찰 불가를 보장 (silent branch 없음)
- cap/cooldown 조정 아님 — 복잡도 제거 (기존 `max_events=10_000` 유지)

`recent`는 현재 프로덕션 소비자 0 (rg 확인: publish만 사용) — no-since 경로가 O(count) 스냅샷이 되는 것은 무영향.

## 테스트

기존 4 케이스 유지 + 신규 3:
- `recent_overflow_matches_oracle`: cap+137 publish 후 oracle prefix와 골든 동등
- `recent_capacity_and_over_boundary`: 정확히 cap / cap+1 경계 (최오래 evict, 순서 불변)
- `concurrent_publish_no_loss`: 4 fiber × 100 동시 publish, loss/dup 0

`ocamlc -stop-after parsing` 3파일 PASS. dune build/test는 CI.

[...]

PR #23008, opened 2026-07-02
fix: restore tool registry CI baseline

jeong-sik:
## Summary
- keep retired local operator front-door names out of the generic Tool_dispatch tag/schema registry while preserving catalog metadata and operator-remote export
- exclude that same retired set from unified visible-tag coverage so masc_surface_audit does not re-enter via descriptor raw schema projection
- update synthetic dispatch/token tests to register input schemas under the strict pre-dispatch validation gate
- reset keeper tool schema universe state inside descriptor registry tests to remove order-dependent leakage

## Verification
- opam exec -- ocamlformat --check lib/unified_tool_registry.ml lib/operator/operator_tool.ml test/test_keeper_tool_descriptor_registry_integrity.ml test/test_tool_result.ml test/test_tool_token.ml
- git diff --check
- bash scripts/ocaml-structure-ratchet.sh
- no local Dune build/test run; GitHub CI remains build authority

[근거] GitHub Build and Test failures on #22970 job 84757181770 and #23007 job 84758206526 showed the same current-main tool registry failures: schema-less __test_tool/__test_token_dispatch rejection, leaked retired operator tools, and schema-universe leakage. Checked 2026-07-02 KST. Confidence: High for source root cause, Medium until CI completes.

### DIFF
--- a/lib/operator/operator_tool.ml
+++ b/lib/operator/operator_tool.ml
@@ -351,7 +351,7 @@ let operator_tool_name name = operator_remote_tool_name (Operator_remote_name.Op
 let surface_audit_tool_name = operator_remote_tool_name Operator_remote_name.Surface_audit
 
 (* ================================================================ *)
-(* Tool_spec registration                                           *)
+(* Catalog metadata registration                                    *)
 (* ================================================================ *)
 
 let tool_spec_read_only =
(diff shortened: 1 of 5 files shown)


jeong-sik:
Blocking: this removes the generic dispatch registration for operator MCP tools, but the current tools/call path still executes all non-internal tool names through Tool_dispatch before it can reach Tool_operator.dispatch.

Specifically, lib/operator/operator_tool.ml now makes the operator catalog metadata-only, and lib/unified_tool_registry.ml marks the operator names as retired/skipped. But lib/mcp_server_eio_execute.ml still calls Tool_dispatch.mint_token ~name and Tool_dispatch.lookup_tag name before dispatch_by_tag can match Mod_operator. Operator_remote affects listing/allowing via Tool_operator.remote_schemas/remote_tool_names; I do not see a separate operator-profile execution path that bypasses the token/tag registry.

Result: masc_operator_snapshot, masc_operator_action, etc. can still be listed in the operator profile, but tools/call should fail as Unknown tool before Tool_operator.dispatch is reached. Either keep tag/schema registration for operator tools while fixing the stale tests, or add a typed operator-profile execution path with regression coverage for listing + call.

jeong-sik:
Closing as superseded by #23010. This branch tries to repair the shared tool-registry CI baseline by removing generic Tool_dispatch registration for operator tools, but that breaks the current tools/call execution path before Tool_operator.dispatch can run. #23010 keeps production operator registration intact and fixes only the stale/under-specified test fixtures and test state isolation.

PR #23037, opened 2026-07-02
fix(governance): make front-door decide hard-forbidden gate unconditional

jeong-sik:
## Summary
The hard-forbidden check in `Governance_pipeline.decide` was gated on `Env_config_core.disable_hitl ()`, so Critical-risk tools could be silently auto-approved in development when HITL was enabled. This change removes the HITL-disable condition so the hard-forbidden gate is evaluated unconditionally.

## Changes
- `lib/governance_pipeline.ml`: `decide` now always requires confirmation for hard-forbidden calls (Critical risk).
- `lib/governance_pipeline.mli`: update docstring to reflect unconditional behavior.
- `test/test_governance_pipeline.ml`: rename and flip development-critical tests so they fail before the fix and pass after.

## Verification
- `scripts/dune-local.sh build test/test_governance_pipeline.exe` ✅
- `./_build/default/test/test_governance_pipeline.exe` ✅ (99 tests)
- `scripts/dune-local.sh build @check` ✅

## Scope
Narrow: front-door `decide` only. Runtime blocker handling is intentionally left unchanged per current scope.

### DIFF
--- a/lib/governance_pipeline.ml
+++ b/lib/governance_pipeline.ml
@@ -117,7 +117,7 @@ let decide ~governance_level ~tool_name ~input =
   let risk = assess_risk ~tool_name ~input in
   let trace_id = generate_trace_id () in
   let action =
-    if Env_config_core.disable_hitl () && auto_approval_hard_forbidden ~risk None then
+    if auto_approval_hard_forbidden ~risk None then
       `Require_confirm
         (Printf.sprintf
            "Governance (%s): %s risk tool %S requires confirmation: auto-approval is \
--- a/lib/governance_pipeline.mli
+++ b/lib/governance_pipeline.mli
@@ -61,14 +61,15 @@ val decide :
   input:Yojson.Safe.t ->
   governance_decision
 (** Evaluate a tool call against governance policy and return a decision.
-    - development: allow all while HITL is enabled, audit High+Critical
+    - development: allow Low/Medium/High while HITL is enabled, audit High+Critical,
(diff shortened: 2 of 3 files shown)


jeong-sik:
Source review basic pass for current head `81d00ed`.

Scope reviewed: front-door governance `decide` now makes Critical hard-forbidden unconditional, with docs/tests updated for development and pre-hook behavior. `git diff --check origin/main..refs/tmp/masc-pr-23037` passed. I do not see a source blocker in the changed files.

CI is still not green: `Build and Test` and `CI Gate` are failing on run `28613133563`, with Dashboard/TLA skipped. Treating this as source-review complete only; merge still waits on CI.

jeong-sik:
Current-head follow-up for `81d00edf787d3eef57cdec148e1da8c5ead753bc`:

Source review: this is the intended hard-forbidden front-door fix. `decide` no longer gates `auto_approval_hard_forbidden` behind `MASC_DISABLE_HITL`, and the updated direct/pre-hook tests cover development-level Critical calls. This aligns with the existing keeper-side hard-forbidden policy, which already blocks before HITL mode.

CI note: Build and Test / CI Gate are red on this run, but the failure markers are the existing registry/dispatch suite markers also seen on sibling PRs, not a compile failure from this diff. I did not run local Dune.

jeong-sik:
고쳐짐: origin/main 기준 rebase 후 CI 전체 green 확인 (Build and Test 포함).

PR #23109, opened 2026-07-03
fix(dashboard): unwrap {ok,data} envelope in IDE presence strip (runtime_unknown fix)

jeong-sik:
## 증상

MASC IDE(`dashboard#code?section=ide-shell`) presence가 라이브 runtime이 있어도 **영구적으로 `disconnected (runtime_unknown)`** 로 표시됨.

## 근본 원인

`ide-presence-strip.ts`의 `fetchPresence`가 `status.cluster` / `agentsData.agents`를 **응답 최상위**에서 읽음. 그러나:
- `/api/v1/status`, `/api/v1/agents` = 표준 MASC **`{ok, data:{...}}` envelope** (masc_status는 MCP tool 계약).
- `/api/v1/providers` = payload 최상위 (대시보드 전용, envelope 없음).
- `get()`는 envelope를 언랩하지 않음 (`parseJsonResponse`가 raw `JSON.parse` 반환).

→ `status.cluster`가 항상 `undefined` → `?? ''` → `''` → `agentsToPresence`가 `runtime_unknown` 반환. 서버가 `cluster:"default"`를 정상 반환해도 프론트가 못 봄.

실측: `curl /api/v1/status` → `{"ok":true,"data":{"cluster":"default",...}}`.

## 수정

`unwrapEnvelope()` 추가 — `data` 키가 있으면 언랩, 없으면 raw 그대로(`/api/v1/providers` 호환), null/non-object는 undefined. `fetchPresence`가 언랩 후 읽음. `agentsToPresence`는 불변(언랩된 status를 받으면 원래 정상).

## 검증

- **vitest 18/18 pass** (기존 14 + 신규: `unwrapEnvelope` 3 + envelope→live 회귀 1).
- 회귀 테스트가 버그 시나리오 재현: `{ok,data:{cluster:"default"}}` → 언랩 → `agentsToPresence` → **live** (runtime_id="default").

## 경계

- 프론트 언랩이 근본 수정 — 백엔드 envelope 변경은 MCP/CLI 소비자를 깨므로 부적절. `{ok,data}`는 표준 계약.
- presence-strip이 이 두 엔드포인트의 유일한 소비자 (다른 컴포넌트는 envelope 없는 대시보드 엔드포인트 사용).

### DIFF
--- a/dashboard/src/components/ide/ide-presence-strip.test.ts
+++ b/dashboard/src/components/ide/ide-presence-strip.test.ts
@@ -4,6 +4,7 @@ import {
   presenceContextAnchor,
   presenceContextSummary,
   prLabel,
+  unwrapEnvelope,
   type ApiAgent,
   type ApiStatus,
 } from './ide-presence-strip'
@@ -58,6 +59,34 @@ describe('agentsToPresence', () => {
       expect(snap.entries).toHaveLength(1)
       expect(snap.entries[0]?.workspace_label).toBe('nick0cave')
     }
+  })
+})
+
+describe('unwrapEnvelope', () => {
+  it('unwraps the {ok,data} envelope returned by /api/v1/status', () => {
+    const raw = { ok: true, data: { cluster: 'default', project: 'me' } }
+    expect(unwrapEnvelope<ApiStatus>(raw)).toEqual({ cluster: 'default', project: 'me' })
+  })
+
+  it('returns an already-unwrapped payload unchanged when no data key is present', () => {
+    const raw = { cluster: 'default' }
+    expect(unwrapEnvelope<ApiStatus>(raw)).toEqual({ cluster: 'default' })
+  })
+
+  it('returns undefined for null / non-object input', () => {
+    expect(unwrapEnvelope<ApiStatus>(null)).toBeUndefined()
+  })
+
+  // Regression: the strip read status.cluster off the raw {ok,data} envelope,
+  // so cluster was always undefined -> permanent runtime_unknown even while the
+  // server reported a live cluster. Unwrapping first must yield a live snapshot.
+  it('feeds an enveloped status into agentsToPresence as a live runtime', () => {
+    const envelope = { ok: true, data: { cluster: 'default', project: 'me' } }
+    const status = unwrapEnvelope<ApiStatus>(envelope) ?? {}
+    const agent: ApiAgent = { name: 'nick0cave', status: 'active', current_task: null, model: null }
+    const snap = agentsToPresence([agent], status)
+    expect(snap.kind).toBe('live')
+    if (snap.kind === 'live') expect(snap.runtime_id).toBe('default')
   })
 })
 
--- a/dashboard/src/components/ide/ide-presence-strip.ts
+++ b/dashboard/src/components/ide/ide-presence-strip.ts
@@ -74,14 +74,32 @@ export function agentsToPresence(
   }
 }
 
+/** The standard MASC API endpoints ([/api/v1/status], [/api/v1/agents]) wrap
+    their payload in an [{ ok, data }] envelope, unlike the dashboard-specific
+    [/api/v1/providers] which returns its payload at the top level. [get()]
+    returns the raw parsed body without unwrapping, so a consumer must pull
+    [.data] out. Reading the envelope's absent top-level fields (e.g.
+    [status.cluster], which actually lives at [status.data.cluster]) is what
+    made this strip render a permanent [disconnected (runtime_unknown)]
+    regardless of the live runtime. Falls back to the raw value when no [data]
+    key is present, so an un-enveloped response still works.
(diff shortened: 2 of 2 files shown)


jeong-sik:
# Adversarial review: #23109 — unwrap {ok,data} envelope in IDE presence strip

**Repo:** `jeong-sik/masc`  
**PR head:** `c45b131565f5275e4f99464150a634070e3b947d`  
**Files reviewed:** `dashboard/src/components/ide/ide-presence-strip.ts`, `dashboard/src/components/ide/ide-presence-strip.test.ts`

---

## CI status

- `Dashboard` check still `IN_PROGRESS`.
- `mergeStateStatus`: `BLOCKED`.

---

## What this PR does

The IDE presence strip was reading `/api/v1/status` and `/api/v1/agents` responses as if their payload fields were at the top level, but the server returns `{ ok, data }` envelopes. This caused `status.cluster` (actually at `status.data.cluster`) to be `undefined`, so the strip permanently rendered `disconnected (runtime_unknown)`. The fix adds `unwrapEnvelope<T>()` and calls it before feeding the data into `agentsToPresence`.

---

## P1 — serious logic / production readiness

### P1.1 — `unwrapEnvelope` does not validate `ok`

The helper assumes that any object with a `data` property is a successful envelope:

```ts
if ('data' in raw) return (raw as { data: T }).data
```

If the API returns `{ ok: false, data: null, error: "..." }`, the helper silently returns `null`/`undefined` and the strip falls back to an empty status or `disconnected`. An error envelope is conflated with a missing envelope.

**Required fix:** Check `ok === true` before returning `.data`. If `ok` is `false`, propagate the error (throw a typed error, return an error variant, or at least log and fall back to `disconnected` with a meaningful reason tag).

### P1.2 — Envelope unwrapping is done in the component instead of the API client

The comment explicitly states that `get()` returns the raw parsed body without unwrapping. That means every consumer of `/api/v1/status`, `/api/v1/agents`, and similar endpoints must re-implement the same `{ ok, data }` knowledge. This is an SSOT violation and invites the exact bug this PR is fixing to reappear in other components.

**Required fix:** Move envelope unwrapping into the `get()` helper (or a thin wrapper such as `getEnveloped<T>()`). The component should receive a typed `ApiStatus` or an explicit error, not `unknown`.

---

## P2 — debt / fragility

### P2.1 — `unwrapEnvelope` returns `undefined` for non-objects but `fetchPresence` silently coalesces to `{}`

```ts
return agentsToPresence(agents, statusData ?? {})
```

A `null`/malformed response becomes an empty status object. This is not logged, so a broken API contract is invisible in production telemetry. At minimum, log the malformed response before falling back.

### P2.2 — `fetchPresence` catch block swallows all errors without logging

```ts
catch {
  return disconnectedSnapshot('fetch_failed')
}
```

Network failures, parser errors, and now envelope errors all collapse to the same UI state with no observability. Add Sentry/console logging or an Otel counter.

### P2.3 — Test assertions rely on the helper's heuristics rather than the API contract

The tests verify `unwrapEnvelope({ ok: true, data: ... })` but do not test the `ok: false` case. Add a test that asserts error handling for failed envelopes.

---

## P3 — cleanup

- The JSDoc uses `[`/`]` brackets inconsistently (e.g. `[/api/v1/status]`). Clean up formatting.
- The `agentsData?.agents` fallback to `[]` is fine, but could be combined with a runtime type guard if the API contract ever changes.

---

## Verdict

**Not ready for merge.** The functional fix is correct, but the implementation reintroduces two P1 issues: it silently accepts failed envelopes and it places envelope-shape knowledge in the component instead of the API client.

**Required before merge:**
1. Validate `ok === true` in `unwrapEnvelope` and handle `ok: false` explicitly.
2. Move envelope unwrapping to the shared API client layer.
3. Add logging for malformed responses and fetch failures.
4. Add a test for the `ok: false` envelope case.

Do not add `ready-for-merge` / `reviewed:adversarial-pass` until the above are addressed and CI is fully green.

jeong-sik:
# Post-merge note: #23109

This PR merged at `2026-07-03T19:17:18Z` while CI was green.

The functional fix (unwrapping the `{ ok, data }` envelope before reading status fields) is correct and addresses the `runtime_unknown` symptom.

However, the pre-merge adversarial review (round 1) identified two unresolved P1 items that remain in `main`:

1. **`unwrapEnvelope` does not validate `ok`.** A failed envelope (`{ ok: false, data: ... }`) is still unwrapped and treated as success. This is a silent-failure path that can mask API errors.
2. **Envelope-shape knowledge lives in the component, not the API client.** Other consumers of `/api/v1/status` and `/api/v1/agents` must re-implement the same unwrapping logic. This SSOT violation invites the same class of bug elsewhere.

Plus two P2 items: no logging for malformed responses or fetch failures.

Recommendation: follow up with a small refactor that moves envelope unwrapping into the shared `get()`/`getEnveloped<T>()` client, validates `ok`, and surfaces errors explicitly.

PR #23112, opened 2026-07-03
feat(keeper): HITL context-aware LLM judgment summary worker

jeong-sik:
## Summary

- Adds non-blocking HITL context-summary worker that enriches pending approvals with LLM-generated context summaries.
- Extends `pending_approval` with `context_summary` and `summary_status` fields.
- Worker collects task title/status, goal title/phase/priority, and recent chat messages by turn; marks `partial_context` only when store lookups fail.
- Summary transitions emit `summary_updated` audit events and `approval:summary_updated` SSE broadcasts.
- Dashboard JSON exposes `summary_status` and `context_summary` for pending approvals.
- Policy-driven operational knobs are exposed through `Keeper_config` / `Runtime_params`: summary timeout, chat message limit, max tokens, temperature, and concurrency limit.
- Provider config is resolved via the keeper→runtime SSOT (`Runtime.runtime_id_for_keeper`), falling back to `Keeper_config.default_runtime_id` only when no keeper assignment exists.
- Token/cost guard clamps `max_tokens` to the policy cap and forces a deterministic `temperature`.
- Bounded concurrency via a global lazy `Eio.Semaphore` so a burst of approvals cannot create an unbounded number of in-flight LLM calls.
- Board signals are intentionally left out of context collection for now; their absence no longer marks the context partial.

## Design

See `memory/design-hitl-context-llm-judgment.html` for full design rationale.

## Test Plan

- `scripts/dune-local.sh runtest test/test_hitl_summary_worker` — 8 unit tests (JSON round-trip, parsing, no-provider-config failure, real workspace context collection, provider cost guard).
- `scripts/dune-local.sh runtest test/test_keeper_approval_queue test/test_hitl_approval` — existing approval/HITL tests still pass.

## Open follow-ups

- Board signal collection.
- Policy-driven risk threshold for summary generation.


jeong-sik:
# Adversarial review: #23112 — HITL context-summary worker

**Repo:** `jeong-sik/masc`  
**PR head:** `1755bd82`  
**Verdict:** BLOCKED — P1 findings must be resolved before merge. CI is still running; I will re-check logs once `Build and Test` finishes.

No P0 HOLB/global-lock issues were found: the worker is per-entry, copy-on-write, and correctly re-raises `Eio.Cancel.Cancelled`. However, several production-readiness gaps remain.

---

## P1 — must fix

### 1. Hardcoded operational constants (`summary_timeout_s`, `chat_context_message_limit`)

`lib/keeper/hitl_summary_worker.ml:9-10`:

```ocaml
let summary_timeout_s = 30.0
let chat_context_message_limit = 20
```

These are production operational knobs baked into source. The PR body already lists "policy-driven summary timeout and chat message limit" as a follow-up, but per the no-hardcoding / production-level standard the follow-up should be resolved now, not deferred. Provide a policy/env/config path (e.g., `MASC_KEEPER_HITL_SUMMARY_TIMEOUT_S` with a documented ceiling ratchet, or a runtime policy table). Returning a static default from `Keeper_config` is acceptable only if callers can override without editing source.

### 2. Provider resolution ignores keeper→runtime SSOT

`lib/keeper/keeper_approval_queue.ml:412-416`:

```ocaml
let provider_config_for_summary () =
  let runtime_id = Keeper_config.default_runtime_id () in
  match Runtime.get_runtime_by_id runtime_id with
  | Some rt -> Some rt.Runtime.provider_config
  | None -> None
```

This always uses the global default runtime. The codebase already has `Runtime.runtime_id_for_keeper : string -> string option`, which is the SSOT for keeper→runtime assignment. Summaries for a specific keeper should resolve through that assignment; falling back to the default runtime only when no keeper assignment exists. Routing every keeper's context summary through the default runtime risks using the wrong model/provider and couples unrelated keepers to one runtime's quota/availability.

### 3. No token/cost guard on the summary LLM call

`hitl_summary_worker.ml` applies the schema via `apply_hitl_summary_schema_to_config` but does not cap `max_tokens` or set `temperature`. Compare `lib/keeper/keeper_memory_llm_summary.ml:64-78`, which clamps `max_tokens` to 512 and sets `temperature = 0.0`. A runtime provider config with a large token budget could make every HITL summary call expensive and slow. Add an explicit, policy-bound token cap and a deterministic temperature.

### 4. Unit tests do not exercise real context collection

`test/test_hitl_summary_worker.ml` uses `audit_base_path = ""`. That forces every `build_context_bundle` test through the "workspace config unavailable" partial path. The actual `Workspace_query.get_tasks_safe`, `Goal_store.get_goal`, and `Keeper_chat_store.load` branches are never exercised. For a production feature, add tests that construct a real in-memory/temporary workspace config and assert that task/goal/chat context are correctly included and that `partial_context` is false when all lookups succeed.

### 5. Unbounded concurrent summary fibers

Every Medium/High/Critical approval spawns a summary fiber with no concurrency cap or queue. Under a burst of approvals this can create an arbitrary number of in-flight LLM calls, each holding `net`, memory, and provider quota. Add a bounded worker pool, semaphore, or per-keeper rate limit with a clear policy default.

---

## P2 — should fix

### 6. `parse_suggested_option` silently drops unknown risk-delta strings

```ocaml
| `String s -> risk_level_of_string s
| _ -> None
```

The schema restricts values, but if the provider returns an unexpected string the parser maps it to `None` without logging. Treat unexpected structured-output values as a parse error or at least emit a metric/warn.

### 7. `summary_version` is hardcoded to `1`

`parse_summary` always sets `summary_version = 1`. Define a module constant and document the versioning/evolution contract; future schema changes will need a migration path.

### 8. Board signals are unconditionally skipped, so `partial_context` is always `true`

The worker always executes `note acc "board signals skipped (not easily accessible)"`. Until board-signal collection lands, every summary will be flagged partial, which trains operators to ignore the flag. Either gate the feature behind a policy that documents this, or keep the PR in draft until the feature can produce non-partial summaries for the happy path.

### 9. System prompt is a hardcoded, unversioned string

The prompt in `hitl_summary_worker.ml:12-22` is central to output quality and safety. It should be versioned (e.g., a prompt registry entry with an ID) so changes are tracked and canary-tested without source edits.

### 10. `record_summary_updated` audit event drops useful fields

The audit JSON only records `summary_status`. It should also include `model_run_id`, failure `reason` (when failed), and a redacted summary fingerprint so audits are forensically useful without leaking full content.

### 11. `Eio_context` availability in a forked fiber is untested

The worker reads `Eio_context.get_net_opt ()` inside the forked fiber. Fiber-local context inheritance is not obvious; add a test that provides a fake Eio context and verifies the worker can actually reach `call_summary_llm` (e.g., with a mocked provider).

### 12. `chat_context` loads the full chat store into memory

It calls `Keeper_chat_store.load` and then filters by turn in-memory. If a keeper has a long conversation, this is O(n) per summary. Consider using `load_page` with a turn-aware filter or a store-level query if one exists.

### 13. `task_context` rebuilds the task-goal index on every call

`Workspace_goal_index.build_task_goal_index_for_config config` is recomputed for each task context lookup. Cache or share the index within the worker invocation.

### 14. `goal_context` has a string fallback for status

```ocaml
| _ -> "unknown"
```

`Goal_store.goal_status_to_yojson` is expected to return `` `String _ ``; if it ever changes, this silently degrades. Either remove the fallback or handle it explicitly.

---

## P3 — minor

15. Field name `context_summary` collides with type `hitl_context_summary`, which is confusing in JSON serialization and reviews. Consider renaming the type or the field.
16. No HITL-summary-specific metric is registered (success, failure, timeout, parse error). Add a counter so operators can observe the worker's behavior.

---

## What looks good

- Copy-on-write update via `atomic_update pending` keeps the worker decoupled from the queue.
- `Eio.Cancel.Cancelled` is re-raised rather than swallowed.
- Structured output uses the existing schema combinators instead of ad-hoc string parsing.
- No hardcoded model ID or base URL in the worker itself.
- Callbacks keep the LLM bridge module free of approval-queue state.

---

## Next steps

1. Address the P1 items above.
2. Push the fixes and let CI run.
3. I will re-review the amended diff and the `Build and Test` / `TLA+ Model Checking` results before applying any labels.

jeong-sik:
# CI follow-up: `Meta Guards` failure on #23112

The `Meta Guards` job failed because the deterministic-boundary ratchet detected new non-deterministic sources inside the library:

```
origin/main...HEAD: lib/keeper/hitl_summary_worker.ml:215:   ; generated_at = Unix.gettimeofday ()
origin/main...HEAD: lib/keeper/keeper_approval_queue.ml:662:            [ "ts", `Float (Unix.gettimeofday ())
```

## What the ratchet wants

Per `scripts/ci/check-determinism-contract.sh` (DET/NDT gate), non-determinism must either:

1. be wrapped at the system boundary (e.g., passed in as a timestamp argument from the caller), or
2. be justified with a nearby `DET-OK` / `NDT-OK` comment explaining why the non-determinism is acceptable at this site.

## Suggested fixes

### `lib/keeper/hitl_summary_worker.ml:215`

`parse_summary` should not call `Unix.gettimeofday`. The generated timestamp is a boundary concern:

- Option A: take `~generated_at:float` as a parameter and let `summary_of_response` / the caller supply it.
- Option B: use a timestamp from the provider response if one is available (e.g., response telemetry).

Then `parse_summary` becomes a pure function from response JSON + timestamp to `hitl_context_summary`.

### `lib/keeper/keeper_approval_queue.ml:662`

`record_summary_updated` uses `Unix.gettimeofday ()` for the audit/SSE timestamp. This is a classic boundary. If the summary is available, reuse `summary.generated_at`; if this is a failure event, pass the timestamp in from the worker spawn site (where the boundary crossing already happens). Otherwise add an `NDT-OK` comment with rationale.

---

Also note the telemetry warnings from the same job:

```
WARN: new file lib/keeper/keeper_structured_output_schema.ml has no telemetry reference
WARN: new file lib/keeper_contract/keeper_approval_queue_rules_types.ml has no telemetry reference
```

Consider adding observability counters as part of the P1 metrics item in the main review.

Please fix the deterministic-boundary debt and re-push so `Meta Guards` can pass.

jeong-sik:
# Adversarial review: #23112 — round 2 (amended head `65892903`)

**Repo:** `jeong-sik/masc`  
**PR head:** `65892903`  
**Verdict:** Still BLOCKED — P1 findings remain. The deterministic-boundary failure is fixed; thank you.

---

## Fixed since round 1 ✅

- `Meta Guards` deterministic-boundary debt: `Unix.gettimeofday ()` has been removed from `hitl_summary_worker.ml` and `keeper_approval_queue.ml`. `parse_summary` / `summary_of_response` now take `~generated_at`, and `spawn_hitl_summary_worker` / `record_summary_updated` take a boundary `~now` (sourced from `Time_compat.now ()`). This should let the DET/NDT gate pass.

## Still P1 — must fix before merge

### 1. Hardcoded operational constants

`lib/keeper/hitl_summary_worker.ml:511-512`:

```ocaml
let summary_timeout_s = 30.0
let chat_context_message_limit = 20
```

Still hardcoded. Add a policy/config path.

### 2. Provider resolution still bypasses keeper→runtime SSOT

`lib/keeper/keeper_approval_queue.ml:919-923` still uses `Keeper_config.default_runtime_id ()`. Use `Runtime.runtime_id_for_keeper entry.keeper_name` first, fall back to default only when unassigned.

### 3. No token/cost guard on the summary LLM call

`call_summary_llm` applies the schema but does not cap `max_tokens` or set `temperature`. Add a bounded default (compare `keeper_memory_llm_summary.ml`).

### 4. Unit tests still do not exercise real context collection

`dummy_pending_approval` sets `audit_base_path = ""`, so every `build_context_bundle` test hits the partial path. Add tests with a real workspace config fixture that verify task/goal/chat inclusion and `partial_context = false` on success.

### 5. Unbounded concurrent summary fibers

Every Medium/High/Critical approval still spawns a fiber with no cap. Add a bounded concurrency mechanism.

## New P2 introduced in this amendment

### 6. Plan doc contains hardcoded local paths

`docs/superpowers/plans/2026-07-03-hitl-context-summary-worker-plan.md`:

```markdown
cd /Users/dancer/me/workspace/yousleepwhen/masc
gh pr create --draft ... --body-file /Users/dancer/me/memory/design-hitl-context-llm-judgment.html
```

These are machine-specific absolute paths. Replace with relative/parameterized paths or env placeholders.

## P2 still open from round 1

7. `parse_suggested_option` silently drops unknown `estimated_risk_delta` strings to `None`.  
8. `summary_version` is still hardcoded to `1`.  
9. `partial_context` is always `true` because board signals are unconditionally skipped.  
10. System prompt is a hardcoded, unversioned string.  
11. `record_summary_updated` audit event omits `model_run_id` and failure `reason`.  
12. No HITL-summary-specific metric/counter.  
13. `chat_context` loads full chat history into memory to filter by turn.  
14. `task_context` rebuilds the task-goal index on every call.  
15. `goal_context` `| _ -> "unknown"` fallback masks schema changes.

## P3 nits

16. `hitl_context_summary` type vs `context_summary` field name collision.  
17. `record_summary_updated ~now` uses the spawn-start timestamp for both success and failure events; consider `summary.generated_at` for success or capturing the event time at the boundary.

---

## Next steps

1. Address the remaining P1 items and the new plan-doc local-path issue.
2. Let CI run; I will re-check `Meta Guards`, `Build and Test`, and `TLA+ Model Checking` on the next head.

No labels until all required checks are green and the P1 blockers above are resolved.

jeong-sik:
# Adversarial review: #23112 — round 3 (head `1a080ea6`)

**Repo:** `jeong-sik/masc`  
**PR head:** `1a080ea6`  
**Verdict:** Still BLOCKED — one new P1 schema/parser mismatch must be fixed. Several earlier P1 items are now resolved; thank you. CI `Build and Test` is still in progress on run `28682751410`; I will update once it finishes.

---

## Fixed since round 2 ✅

- **Policy constants moved to runtime params:** `summary_timeout_s` and `chat_context_message_limit` are now `Keeper_config.hitl_summary_timeout_sec ()` / `hitl_summary_chat_message_limit ()` (`lib/keeper/hitl_summary_worker.ml`), with runtime-param definitions in `lib/keeper/keeper_config.ml`.
- **Provider resolution now prefers keeper-specific runtime:** `lib/keeper/keeper_approval_queue.ml` tries `Runtime.runtime_id_for_keeper keeper_name` first, then falls back to `Keeper_config.default_runtime_id ()`.
- **Token/cost guard added:** `call_summary_llm` caps `max_tokens` to the policy limit, overrides `temperature`, clears `tool_choice`, and disables parallel tool use (`lib/keeper/hitl_summary_worker.ml`).
- **Bounded concurrency:** summary workers now run under a global `Eio.Semaphore.t` sized by `Keeper_config.hitl_summary_concurrency_limit ()`.
- **Real workspace fixture test added:** `test/test_hitl_summary_worker.ml` initializes a temp workspace and asserts `partial_context=false` on success.
- **`partial_context` semantics corrected:** board signals no longer unconditionally mark context as partial.
- **`goal_context` no longer masks schema changes:** non-string goal status now raises instead of falling back to `"unknown"`.
- **HITL-summary metric added:** `masc_keeper_hitl_summary_outcomes_total` is registered and used.
- **Plan doc absolute paths removed:** `/Users/dancer/me/...` paths are no longer present.

---

## P1 — merge blocker

### 1. Structured-output schema requires fields the model is not asked to produce and the parser ignores

`lib/keeper/keeper_structured_output_schema.ml` defines:

```ocaml
let hitl_context_summary_schema =
  let fields =
    [ "summary_version", integer_schema
    ; "generated_at", number_schema
    ; "model_run_id", string_schema
    ; "context_summary", string_schema
    ; ...
    ]
  in
  object_schema ~required:(List.map fst fields) fields
```

But the system prompt in `lib/keeper/hitl_summary_worker.ml` only instructs the model to emit `context_summary`, `key_questions`, `suggested_options`, `risk_rationale`, and `uncertainty`. Then `parse_summary` populates `summary_version`, `generated_at`, and `model_run_id` from server-side constants/args/response metadata.

**Why this is P1:** providers that validate structured output against the supplied JSON Schema will fail when the model omits `summary_version` / `generated_at` / `model_run_id`. Any values the model does hallucinate for them are discarded, so the schema is both too strict and misleading. This violates SSOT between the LLM contract and the server-side record.

**Fix:** remove `summary_version`, `generated_at`, and `model_run_id` from `hitl_context_summary_schema`; they are server-side metadata, not part of the model output contract.

---

## P2 — should fix before merge

2. **`summary_version` is still hardcoded to `1`** (`lib/keeper/hitl_summary_worker.ml`). A record/schema version should be derivable from the schema or runtime config, not a source literal.
3. **System prompt is still hardcoded and unversioned** (`lib/keeper/hitl_summary_worker.ml`). The comment says versioning is carried by `summary_version`, but the prompt text itself lives only in source and cannot be evolved independently.
4. **`record_summary_updated` audit event still lacks canonical identity fields** (`lib/keeper/keeper_approval_queue.ml`). It writes `ts`, `event`, `id`, `summary_status`, plus `model_run_id` / `failure_reason`, but omits `keeper_name`, `tool_name`, `risk_level`, `action_key`, `turn_id`, `task_id`, `goal_id`, `goal_ids`, `sandbox_target`, and `runtime_contract` that `record_pending` includes. Audit consumers need these for correlation.
5. **`chat_context` loads the full chat history into memory** (`lib/keeper/hitl_summary_worker.ml`). It calls `Keeper_chat_store.load` (entire history) then filters/takes. For long sessions this is O(history) memory per summary; the store should expose a bounded, turn-windowed read.
6. **`task_context` rebuilds the task→goal index on every call** (`lib/keeper/hitl_summary_worker.ml`). `Workspace_goal_index.build_task_goal_index_for_config` is recomputed per HITL summary. Cache or obtain from an existing stable query.
7. **New real-workspace test mutates global process env** (`test/test_hitl_summary_worker.ml` calls `Unix.putenv "MASC_BASE_PATH" base_path`). This breaks test isolation under parallel or out-of-order dune runs. Construct the workspace config directly without relying on global env.
8. **Summary spawn risk threshold is hardcoded to Medium+** (`lib/keeper/keeper_approval_queue.ml`). There is no runtime param controlling whether Low (or only High/Critical) approvals should spawn a summary worker.

---

## P3 nits

- `chat_context` reimplements `List.take`.
- `record_summary_updated` uses the spawn-time `now` for failed-event timestamps rather than the actual failure time.
- Race between `update_pending_entry` and `get_pending_entry` in `on_summary` / `on_failure` can drop the audit/SSE event if the operator resolves the entry between the two operations.
- `dummy_pending_approval` still defaults to `audit_base_path = ""`; acceptable now that there is a real-workspace test, but consider making the partial-context default explicit.

---

## Next steps

1. Fix the P1 schema/parser mismatch: remove server-side metadata fields from `hitl_context_summary_schema`.
2. Add canonical identity fields to `record_summary_updated`.
3. Replace full chat-history load with a bounded, turn-aware read.
4. Cache or avoid rebuilding `Workspace_goal_index` per summary call.
5. Remove `Unix.putenv` from the test fixture.
6. Let CI finish; I will re-check `Build and Test`, `Meta Guards`, and `TLA+ Model Checking` on the next head.

No labels until all required checks are green and the P1 schema blocker is resolved.

jeong-sik:
## 적대적 리뷰 — PASS (P0 없음)

강한 적대적 기준(no silent failure / no string-match / immutable / SSOT / SPOF·제약지옥 P0)으로 코어(`lib/keeper/hitl_summary_worker.ml`) 정독 결과, **논리·흐름·정당성 모두 프로덕션 레벨**입니다.

### P0(SPOF/제약지옥) — 통과
- `spawn`은 `Eio.Fiber.fork ~sw`로 **detached fiber** 실행 → 승인 경로/턴 루프를 블록하지 않음. 요약은 순수 enrichment.
- 예외 규율 정석: `| Eio.Cancel.Cancelled _ as e -> raise e | exn -> record "crashed"; on_failure ...` — 취소를 **재-raise**(흡수 안 함, `KeeperOASAdvanced.tla` CancelledAbsorbed 버그모델 준수). 일반 exn만 catch.
- `run_with_timeout_and_fallback ~timeout_s`로 hung provider 방어 + `Eio.Semaphore`(`with_summary_slot`, `Switch.on_release`로 해제)로 동시성 상한 → 승인 폭주 시에도 in-flight LLM 무한 증식 없음.
- `summary_status`는 typed variant(`Summary_not_requested | Summary_pending | Summary_available | Summary_failed {reason;retryable}`). 승인 결정을 **gate하지 않음** → 요약이 실패/미생성이어도 사람이 승인 가능. switch 없으면 요약 생략 후 정상 동작(graceful degrade). **keeper 하나의 요약 실패가 다른 keeper를 멈추지 않음.**

### Silent Failure — 통과
- 각 store lookup(task/goal/chat)을 `try/with`로 감싸 실패를 `note`(→ `partial_context=true`)로 **표면화** + exn 메시지 기록. PR 본문 주장("partial_context only when store lookups fail")과 일치.
- 모든 종료 경로가 `record_outcome`에 typed label(ok_summary/parse_error/provider_error/timeout/no_net/no_provider_config/crashed) 기록.
- 파싱: `estimated_risk_delta` unknown string → `Log.Keeper.warn` + `None`(permissive default 아님). `risk_level_of_string`은 option 반환.

### SSOT / env — 통과
- 신규 5개 knob(timeout/chat_limit/max_tokens/temperature/concurrency) 전부 `Runtime_params` SSOT(`_rp_int`/`_rp_float` + key/default/min_v/max_v/description)로 등록, env 기본값은 `int_of_env_default`/`float_of_env_default` 경유. runtime.toml에서 findable + bounded. ad-hoc getenv 아님. **이 패턴이 정석**(cf. #23111은 hand-roll — 그쪽 코멘트 참조).

### CI 실패는 이 PR 결함 아님 (base drift)
`Build and Test`의 `Env knob catalog drift gate` red는 **base drift**입니다. gate diff는 무관한 knob의 소스 라인 번호(`MASC_SHELL_IR_APPROVAL_GATE_ENABLED` 876→884)만 바뀜 — 이 PR은 `env_config_*.ml`/`runtime-tunables.md`를 건드리지 않음. main의 #23062/#22983 머지로 상위 라인이 밀린 결과. 브랜치를 main으로 update하면 해소됨(아래에서 조치).

### 잔여(비차단)
- P3: catalog 생성기가 keeper_config.ml 소스 knob을 스캔 범위(env_config_*.ml 11파일)에 포함하지 않으면 HITL knob이 runtime-tunables.md에 미기재될 수 있음(문서 커버리지 갭, 기존 생성기 스코프 이슈).

`reviewed:adversarial-pass` 부여. 머지는 base-drift 해소 + CI green 후 사용자 판단.

PR #23172, opened 2026-07-04
feat: improve MASC Dashboard headers, nav rail order, and standalone HTML export

jeong-sik:
This PR resolves key design alignment issues on the MASC Dashboard:
1. Aligned nav-rail-v2 with the primary V2_PRIMARY_SURFACE_IDS order.
2. Removed 'monitoring' from the primary rail (Logs is kept as a dedicated surface in the rail).
3. Replaced legacy/observation-only connection state displays with the global ConnectionStatus component across Connector, Schedule, Fusion, and IDE headers.
4. Cleaned up hype descriptions in navigation config.
5. Added a standalone HTML download card in the Display Settings panel.
6. Fixed and updated the sse/reconnectCount mock setup in Vitest tests.
7. Documented the legacy status of tool_policy.toml as a header warning.


jeong-sik:
P0: none found in the current source diff.

P1: Fix or rename the new Standalone HTML export before this leaves draft. `dashboard/src/components/settings-surface.ts:974-982` downloads only `document.documentElement.outerHTML`, while `dashboard/src/components/settings-surface.ts:1680-1681` presents it as an offline single HTML containing all dashboard resources and current state. `outerHTML` does not inline the Vite JS/CSS/assets, does not preserve event handlers, and does not serialize Preact/signal state into a runnable app, so the downloaded file is at best a current DOM snapshot rather than a standalone dashboard. Either implement a real resource-inlined/export snapshot path with focused tests, or narrow the UI copy and filename to make the snapshot semantics explicit.

P2: Once the export semantics are chosen, add a focused dashboard test for the Display settings export affordance. The current PR adds the action without coverage around the visible promise or the generated payload shape, so this can regress silently.

P3: Header connection status reuse, nav ordering, and nowrap polish look source-consistent. The `config/tool_policy.toml` legacy marker does not broaden runtime policy; it matches the current README/TOML reload matrix direction that this file is retained as a seeded config-root marker only.

Mergeability: draft, no review decision, no review threads, mergeable/no conflicts, `git diff --check origin/main...origin/pr/23172` clean. `pr-ready-check` reports 4/7 passing with CI still pending at head `0e6b0da9`, so this is not merge-ready until P1 is resolved or explicitly descoped, draft is lifted, review is present, and required checks are green.

jeong-sik:
P1: Standalone HTML export is not actually standalone. `dashboard/src/components/settings-surface.ts:997` serializes only `document.documentElement.outerHTML`, while the new UI copy at `settings-surface.ts:1704` promises an offline single HTML file containing all dashboard resources and current state. `outerHTML` preserves DOM markup but does not inline the Vite JS/CSS bundles, fetched assets, API-backed data needed after reload, or live runtime state; the downloaded file can therefore open as a broken shell or stale DOM snapshot instead of the advertised offline artifact. Please either implement a real export pipeline that embeds the required resources/state from an authoritative source, or narrow the feature/copy to a truthful DOM snapshot so operators do not trust it as a portable dashboard artifact.

jeong-sik:
Follow-up on the earlier P1/P2:

Fixed in head `e78d1366846e101d7707897bd9b8bd839aea4bfd`.

- P1 fixed: the export is now labeled as an HTML snapshot, the filename is `MASC_Dashboard_snapshot.html`, and the copy no longer claims a standalone/offline resource-inclusive dashboard artifact.
- P2 fixed: `dashboard/src/components/settings-surface.test.ts` now covers the Display export affordance, generated Blob type/content, download filename, and guards against the previous Standalone/all-resources wording.
- Small cleanup: `config/tool_policy.toml` now keeps the legacy marker ASCII-only and points to `docs/TOML-RELOAD-MATRIX.md`.

Verification run locally on the PR worktree:

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1` -> 1 file / 40 tests passed.
- `pnpm --dir dashboard exec tsc --noEmit --pretty false` -> passed.
- `pnpm exec eslint src/components/settings-surface.ts src/components/settings-surface.test.ts` from `dashboard/` -> exit 0; test file is ignored by the current eslint config, component file clean.

Remaining mergeability: CI is running on `e78d136`; no review decision yet. Not merge-ready until required checks finish green and review/approval policy is satisfied.

jeong-sik:
Adversarial source review at `e78d136684`.

P1 fixed:
- The Display Settings export card claimed a standalone/offline HTML with all resources and state, but the implementation saves `document.documentElement.outerHTML`. I changed the UI/file name to an honest HTML DOM snapshot and added a regression test that rejects the old standalone/resource claim.
- `config/tool_policy.toml` pointed at a root `TOML-RELOAD-MATRIX.md` that does not exist; it now points at `docs/TOML-RELOAD-MATRIX.md`, and the new non-ASCII warning marker was removed from that config comment.

Local validation, without local `dune`:
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/eslint src/components/settings-surface.ts src/components/settings-surface.test.ts` (test file is ignored by repo config warning only)
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit`
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/settings-surface.test.ts`
- `git diff --check`

Current gate: source review passes and `reviewed:adversarial-pass` is appropriate. PR is `MERGEABLE` but `BLOCKED` while fresh CI runs for `e78d136684`; not waiting on CI here.

PR #23173, opened 2026-07-04
[codex] Surface runtime discovery details

jeong-sik:
## Summary

- Surface typed provider discovery `discovered_model` on runtime monitor cards.
- Surface typed discovery `idle_slots` next to busy/total slots.
- Extend the runtime monitor fixture/assertions so the UI keeps exposing these discovery facts.

## Why

Provider discovery already carries more runtime facts than the dashboard showed. This keeps the runtime monitor aligned with the snapshot contract without adding provider-specific matching or inferred model rules.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/runtime-monitor.test.ts`
- `pnpm --dir dashboard exec eslint src/components/runtime-monitor.ts src/components/runtime-monitor.test.ts` (test file ignored by repo config warning only)
- `git diff --check`

### DIFF
--- a/dashboard/src/components/runtime-monitor.test.ts
+++ b/dashboard/src/components/runtime-monitor.test.ts
@@ -223,6 +223,7 @@ describe('RuntimeMonitor', () => {
           note: null,
           discovery: {
             healthy: true,
+            discovered_model: 'Qwen/Qwen3-32B',
             ctx_size: 200000,
             total_slots: 4,
             busy_slots: 1,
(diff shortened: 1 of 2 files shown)


jeong-sik:
P0: none found.

P1: none found in this source diff.

P2: The added discovery details are small and use existing API fields (`discovered_model?: string | null`, `idle_slots?: number | null`). `formatNumber` already handles null/undefined with the dashboard fallback, and the model id is rendered in a truncating cell with a title, so the nullable/long-value cases look covered.

P3: The focused runtime-monitor test now asserts both the discovered model and idle slot rendering. `git diff --check origin/main...origin/pr/23173` is clean.

Mergeability: draft, no review decision, no review threads, mergeable/no conflicts, but CI is still pending at head `411c8890`. Not merge-ready until draft is lifted, review exists, and required checks finish green.

PR #23175, opened 2026-07-04
fix(hitl): keep summary worker on root switch

jeong-sik:
## Summary
- fork HITL context-summary enrichment from the server root switch, not the turn-local switch
- mark Medium+ summaries as explicit failed status when the root switch is unavailable instead of silently dropping enrichment
- add a regression test for the missing-root-switch failure surface

## Verification
- git diff --check
- opam exec -- ocamlformat --check lib/keeper/keeper_approval_queue.ml test/test_keeper_approval_queue.ml

Local dune build intentionally not run; CI is the build authority for this repo.

### DIFF
--- a/lib/keeper/keeper_approval_queue.ml
+++ b/lib/keeper/keeper_approval_queue.ml
@@ -735,6 +735,15 @@ let update_pending_entry ~id f =
     | Some entry -> SMap.add id (f entry) map)
 ;;
 
+let record_summary_failure ~id ~reason ~retryable =
+  let now = Time_compat.now () in
+  update_pending_entry ~id (fun e ->
+    { e with summary_status = Summary_failed { reason; retryable } });
+  match get_pending_entry ~id with
+  | Some updated -> record_summary_updated ~now updated
+  | None -> ()
+;;
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
Source review pass on current head `6a8c132589`.

P0/P1/P2: none found. The patch correctly stops using the fiber-local turn switch for queued HITL summary workers and resolves the root-switch-missing path as an explicit non-retryable summary failure instead of silently skipping the worker.

Local verification kept narrow per workflow:
- `git diff --check refs/tmp/masc-main-current...refs/tmp/masc-pr-23175-current`
- `opam exec -- ocamlformat --check lib/keeper/keeper_approval_queue.ml test/test_keeper_approval_queue.ml`
- `opam exec -- ocamlc -stop-after parsing lib/keeper/keeper_approval_queue.ml`
- `opam exec -- ocamlc -stop-after parsing test/test_keeper_approval_queue.ml`

I did not run local `dune`; CI remains the build authority.

Mergeability: MERGEABLE. Remaining gates: draft PR plus CI still queued/running; I did not wait for CI.

PR #23188, opened 2026-07-04
[codex] Surface runtime lane spec in agent strip

jeong-sik:
## Summary

- Add a shared keeper runtime display helper that preserves runtime lane evidence while keeping model/provider identity redacted.
- Surface the keeper runtime lane and catalog-derived typed snapshot facts in `AgentRuntimeStrip`.
- Render explicit compact SPEC states when the runtime catalog fails to load or the runtime is absent from a loaded catalog.
- Route the workspace runtime label through the same helper so strip/header/rail priority does not diverge.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/lib/keeper-runtime-display.test.ts src/components/agent-monitor/runtime-strip.test.ts src/components/keeper-workspace/keeper-workspace-shared.test.ts`
- `git diff --check`

## Notes

This keeps `keeperDisplayModel()` redacted by design. The dashboard strip only shows runtime lane identity plus typed catalog capability facts already exposed by the runtime inventory.

### DIFF
--- a/dashboard/src/components/agent-monitor/runtime-strip.test.ts
+++ b/dashboard/src/components/agent-monitor/runtime-strip.test.ts
@@ -127,6 +127,22 @@ describe('AgentRuntimeStrip', () => {
     expect(container.textContent).toContain('claude-4')
   })
 
+  it('does not load runtime catalog when no runtime evidence exists', () => {
+    mockFindKeeper.mockReturnValue({
+      pipeline_stage: null,
+      context_ratio: null,
+      generation: null,
+    })
+    mockKeeperDisplayModel.mockReturnValue(null)
+    mockKeeperDisplayRuntime.mockReturnValue(null)
+    mockKeeperActivityDisplay.mockReturnValue({ ageSeconds: null, label: '' })
+    const container = document.createElement('div')
+    render(h(AgentRuntimeStrip, { name: 'Alpha' }), container)
+    expect(mockLoadRuntimeCatalog).not.toHaveBeenCalled()
+    expect(mockFindRuntimeCatalogEntry).not.toHaveBeenCalled()
+    expect(mockRuntimeCatalogSnapshotFacts).not.toHaveBeenCalled()
+  })
+
   it('renders runtime lane when present', () => {
     mockFindKeeper.mockReturnValue({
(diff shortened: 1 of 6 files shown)


jeong-sik:
Adversarial pass + follow-up patch on current head 4da6aa453d634ad16fa0e6e872f4f57c2b815c7c.

P0: none found.
P1: fixed a small render-side effect: AgentRuntimeStrip now loads the runtime catalog only when the keeper has runtime evidence to resolve. This keeps the SPEC lane tied to an actual runtime value and avoids triggering /api/v1/providers from keeper rows that cannot use the catalog.
P2: added a regression test that no runtime evidence means no catalog load, no catalog lookup, and no snapshot-facts formatting.
P3: local narrow validation passed: pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/agent-monitor/runtime-strip.test.ts (15 tests), pnpm --dir dashboard typecheck, ESLint on touched files, and git diff --check. No local dune build was run.

Mergeability: draft PR, GitHub reports MERGEABLE/BLOCKED after the push. Checks restarted and are pending; previously the dashboard gate passed before this follow-up patch.

jeong-sik:
Reviewed current head `4da6aa4`. Findings: no P0/P1/P2 blockers found.

What I checked:
- `keeperDisplayRuntime()` keeps `keeperDisplayModel()` redacted and routes strip/header runtime labels through one helper.
- Agent strip catalog lookup is guarded by runtime evidence and surfaces explicit `SPEC` states for catalog error/missing-entry cases.
- Existing dashboard code already has precedent for `loadRuntimeCatalog()` calls from render paths (`keeper-runtime-model-editor`), so I am not treating that as a blocker here; moving it into `useEffect` would be a render-purity cleanup, not required for correctness.
- Tests cover no-runtime no-load, runtime lane rendering, catalog facts, error state, and missing-entry state.

Mergeability: draft PR, and wait for pending Dashboard/OCaml Build jobs. P0 none / P1 none / P2 none / P3 optional render-purity cleanup only.

PR #23167, opened 2026-07-04
test(keeper): restore stale-claim release regression guard (#23158 review P2-2)

jeong-sik:
## 목적

#23158(WIP admission 폐지)에서 삭제된 `test/test_keeper_wip_admission.ml`에 **좀비-정리(stale-release) 회귀 테스트가 함께 딸려 삭제**됐다. 이 테스트는 admission과 무관한 별개 관심사(사라진 keeper의 점유 task를 claim 시점에 pool로 되돌리는 경로)를 검증하던 것으로, 같은 파일에 있었다는 이유로 소실됐다. main 기준으로 복원한다.

#23158 적대적 리뷰의 P2-2 지적("cap 제거의 안전성이 stale-release의 좀비 정리에 의존하므로 그 경로가 테스트로 고정돼야 한다")에 대한 대응이다.

## 변경

- 신규 `test/test_keeper_stale_claim_release.ml` — 독립 러너. **폐지된 admission 모듈 참조 없음**.
  - 3개 keeper가 task-001~003 claim → 각 keeper를 stale 마킹(`last_seen` 과거) + claim을 release window 밖으로 aging.
  - `handle_keeper_task_tool ~name:"keeper_task_claim"` 호출 후 검증:
    - `stale_claim_releases` 길이 3
    - 새 claimant가 released된 `task-001`을 집음
    - backlog의 task-002가 `Todo`로 복귀
- `test/dune` — big tests 스탠자 `names`/`modules`에 `test_keeper_stale_claim_release` 등록.

## 검증

- `scripts/dune-local.sh build @default` → exit 0.
- `test_keeper_stale_claim_release.exe` → 1 test green.

## 배경 (P2-1, 참고)

같은 리뷰의 P2-1(리소스 backpressure)은 코드 변경 불필요로 확인됐다: admission cap은 순수 task-count였고, 실제 LLM API 동시성 상한은 provider `binding.max_concurrent`(`runtime_candidate.ml` `of_provider_config` → `keeper_turn_driver.ml`)가 별도로 담당한다. admission 폐지는 그 backpressure를 건드리지 않는다. 이 PR은 P2-2(테스트 복원)만 다룬다.

RFC-WAIVED: 테스트 파일 복원. credential/identity/operator/sandbox/hooks/workflow subsystem 미해당.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23189, opened 2026-07-04
test(keeper): derive stale claim age from ttl

jeong-sik:
## Summary

- Follow-up to #23167 after it merged: remove the hardcoded `2020-01-01T00:00:00Z` stale-claim fixture timestamp.
- Derive the stale timestamp from `Env_config_runtime.Claim.ttl_seconds`, matching the production `keeper_task_claim` release path.
- Keep the one-second margin explicit because `Masc_domain.iso8601_of_unix_seconds` serializes with second precision while `release_stale_claims` checks `now -. ts > ttl_seconds`.

## P0

Improves the stale-claim regression guard so the test tracks the runtime claim TTL SSOT instead of relying on a magic historical date.

## P1

Remaining: CI must run the Dune-backed test target. No local full Dune build was run.

## P2

Local narrow validation:

- `opam exec -- ocamlformat --check test/test_keeper_stale_claim_release.ml`
- `git diff --check`

## P3

Mergeability should be low-risk once CI is green: this is a single-file test-only follow-up on top of `origin/main` after #23167.

### DIFF
--- a/test/test_keeper_stale_claim_release.ml
+++ b/test/test_keeper_stale_claim_release.ml
@@ -47,11 +47,17 @@ let claim_task_exn config ~agent_name ~task_id =
   | Ok _ -> ()
   | Error err -> fail ("claim_task_r failed: " ^ Masc_domain.masc_error_to_string err)
 
-let old_release_timestamp = "2020-01-01T00:00:00Z"
+let iso8601_seconds_precision_margin_s = 1.0
 
-let mark_agent_stale_for_release config ~agent_name =
+let stale_release_timestamp () =
+  let age_s =
+    Env_config_runtime.Claim.ttl_seconds +. iso8601_seconds_precision_margin_s
+  in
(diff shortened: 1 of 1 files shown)


jeong-sik:
Adversarial pass on current head a6e5f148b224661f3f251d253e0f1859b2133916.

P0: none found.
P1: none found. The stale timestamp is now derived from Env_config_runtime.Claim.ttl_seconds, so the regression guard follows the runtime claim TTL SSOT instead of a fixed 2020 timestamp.
P2: the explicit 1s margin is justified by ISO8601 seconds precision plus the release predicate using age > ttl_seconds; it keeps the test deterministic without changing product behavior.
P3: local narrow validation passed earlier: opam exec -- ocamlformat --check test/test_keeper_stale_claim_release.ml and git diff --check. No local dune build was run.

Mergeability: draft PR, GitHub reports MERGEABLE/BLOCKED. All visible gates are green except Build and Test, which is still in progress at review time.

PR #23190, opened 2026-07-04
[codex] Record keeper assigned tools after selection

jeong-sik:
## Summary
- move keeper tool assignment telemetry from bundle assembly to the before-turn surface selection boundary
- keep the executable full tool universe for BM25/discovery, but only mark `Assigned` after the actual `schema_filter` AllowList is computed
- add regression coverage that bundle assembly no longer claims an LLM-visible assignment

## Validation
- `scripts/dune-local.sh build test/test_warn_root_causes.exe`
- `./_build/default/test/test_warn_root_causes.exe`

## Notes
- Addresses the Discovery/tool-search audit gap where observability could report the full executable universe as the actual LLM-visible tool surface.

### DIFF
--- a/lib/keeper/keeper_run_tools_hooks.ml
+++ b/lib/keeper/keeper_run_tools_hooks.ml
@@ -34,6 +34,8 @@ type ctx =
       current_tool_choice:Agent_sdk.Types.tool_choice option ->
       decay_discovered:bool -> unit ->
       string list * turn_lane
+  ; record_tool_assignment :
+      turn:int -> tool_list:string list -> lane:turn_lane -> unit
   ; config : Workspace.config
   ; keeper_tools_cleanup : unit -> unit
   ; manifest_keeper_turn_id : int option
@@ -120,6 +122,7 @@ let assemble_hooks
   let acc = ctx.acc in
   let agent_name = ctx.agent_name in
   let compute_tool_surface = ctx.compute_tool_surface in
+  let record_tool_assignment = ctx.record_tool_assignment in
   let config = ctx.config in
   let keeper_tools_cleanup = ctx.keeper_tools_cleanup in
   let manifest_keeper_turn_id = ctx.manifest_keeper_turn_id in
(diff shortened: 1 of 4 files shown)


jeong-sik:
Adversarial pass on current head 86a5e0d30d976e9289d4c6a03dc316d9f622732b.

P0: none found.
P1: none found. Moving Assigned telemetry from bundle assembly to the before_turn schema_filter boundary matches the actual LLM-visible tool surface. OAS stage_parse invokes before_turn_params before dispatch, so first-turn assignment telemetry is covered.
P2: test shape is acceptable for this slice: there is a behavior guard that bundle assembly no longer emits an assignment, plus a narrow source-location ratchet that prevents the emission site from drifting back into bundle assembly.
P3: local narrow validation passed: opam exec -- ocamlformat --check on touched OCaml files, and git diff --check. I did not run local dune build; CI remains the build authority.

Mergeability: draft PR, mergeable state currently UNKNOWN while checks are still running. At this check, pending gates are Build and Test, OCaml 5.5 Build Canary, Health, TLA+ Model Checking, and ocamlformat-check. No review/comment actionables from pr-ready-check sweep.

jeong-sik:
Reviewed current head (86a5e0d). Findings: no P0/P1/P2 blockers found.

What I checked:
-  moved out of bundle assembly and into the before-turn path after  is computed.
-  now matches the computed AllowList surface for the turn rather than the full executable bundle universe.
- / remain policy context, while Called/Completed continue to link through the latest assignment id.
- Regression coverage pins both: bundle assembly does not emit an assignment, and before-turn owns the telemetry emission site.

Mergeability: wait for remaining CI jobs (, TLA+) to finish. P0 none / P1 none / P2 none / P3 none.

jeong-sik:
Correction: my previous comment was posted with shell-escaping damage around inline code. Exact review follows.

Reviewed current head `86a5e0d`. Findings: no P0/P1/P2 blockers found.

What I checked:
- `Tool_assignment_telemetry.emit_assigned` moved out of bundle assembly and into the before-turn path after `schema_filter` is computed.
- `tool_list` now matches the computed `AllowList` surface for the turn rather than the full executable bundle universe.
- `allow_set` / `deny_set` remain policy context, while Called/Completed continue to link through the latest assignment id.
- Regression coverage pins both: bundle assembly does not emit an assignment, and before-turn owns the telemetry emission site.

Mergeability: wait for remaining CI jobs (`Build and Test`, TLA+) to finish. P0 none / P1 none / P2 none / P3 none.

PR #23192, opened 2026-07-04
fix(runtime): degrade boot on catalog gaps

jeong-sik:
## Summary
- add server-only degraded runtime init for missing OAS catalog rows: disable uncatalogued runtimes, promote a catalog-known default when needed, and drop assignments/role/media/lane references to disabled runtimes
- keep strict init APIs fail-closed for validation/fusion callers so catalog gaps are still caught outside server boot
- expose startup_degradation on /health, /health?full=1, and provider inventory with operator_action_required=true

## Validation
- opam exec -- ocamlformat --check lib/runtime/runtime.ml lib/runtime/runtime.mli lib/server/server_runtime_bootstrap.ml lib/server/server_routes_http_runtime.ml lib/server/server_dashboard_http_runtime_info.ml test/test_runtime_config_validity.ml
- git diff --check
- local Dune build/test intentionally not run; build authority is CI per operator instruction

### DIFF
--- a/lib/runtime/runtime.ml
+++ b/lib/runtime/runtime.ml
@@ -333,6 +333,37 @@ type missing_catalog_report =
   ; missing_models : missing_catalog_model list
   }
 
+type dropped_runtime_assignment =
+  { keeper_name : string
+  ; runtime_id : string
+  }
+
+type dropped_runtime_route =
+  { route_name : string
+  ; runtime_id : string
+  }
+
+type dropped_runtime_lane =
+  { lane_id : string
+  ; runtime_ids : string list
+  }
+
+type startup_degradation =
+  { report : missing_catalog_report
+  ; configured_default_runtime_id : string
+  ; effective_default_runtime_id : string
+  ; disabled_runtime_ids : string list
+  ; dropped_assignments : dropped_runtime_assignment list
+  ; dropped_routes : dropped_runtime_route list
+  ; dropped_media_failover : string list
+  ; dropped_lane_candidates : dropped_runtime_lane list
+  ; dropped_lanes : dropped_runtime_lane list
+  }
+
+type init_default_outcome =
+  | Initialized
+  | Initialized_degraded of startup_degradation
+
 type strict_init_error =
   | Runtime_config_error of string
   | Missing_catalog_models of missing_catalog_report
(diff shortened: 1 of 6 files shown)


jeong-sik:
Adversarial pass on current head e3e427c449737593a4e3b8f0b78fcfac48ed1262.

P0: none found.
P1: none found. The degradation path removes uncatalogued runtime ids from the active runtime set instead of allowing OAS provider_default dispatch, and it remains fatal when every configured runtime is uncatalogued. That preserves the no-silent-fallback boundary while avoiding all-server boot failure when a catalog-known runtime remains.
P2: health/inventory exposure is operator-visible: runtime_startup_degradation carries status=degraded, operator_action_required=true, terminal_reason=missing_oas_catalog_models, disabled runtime ids, dropped routes/assignments/lanes, and next_action. full health folds it into operator_action_reasons.
P3: local narrow validation passed: opam exec -- ocamlformat --check on touched OCaml files and git diff --check. I did not run local dune build; CI remains the build authority.

Mergeability: draft PR. Checks were still running at review time; no review/comment actionables from pr-ready-check sweep.

jeong-sik:
Adversarial follow-up on current head `e3e427c449737593a4e3b8f0b78fcfac48ed1262`.

P1: degraded boot still reroutes explicit runtime intent by fallback.

`degrade_loaded_for_missing_catalog` picks the first catalog-known runtime as `effective_default` when the configured default is catalog-missing (`lib/runtime/runtime.ml:564-566`). It also removes assignments whose runtime was disabled; after that, `runtime_id_for_keeper` returns `None` for the keeper (`lib/runtime/runtime.ml:896`), so existing call sites can treat the keeper as unassigned and use the effective default. The regression test currently locks that behavior in: default `ollama.missing` becomes `ollama.good`, and `keeper_a = "ollama.missing"` is expected to fall back to default (`test/test_runtime_config_validity.ml:1391-1421`).

That avoids OAS `provider_default`, but it still changes configured keeper/default routing to another model by declaration/order. Under the no-heuristic/no-silent-fallback contract, catalog-missing references should remain a typed blocked state, or degraded boot should be allowed only for unreferenced runtime definitions. Default, `runtime.assignments`, `librarian`, `structured_judge`, and `cross_verifier` references to disabled runtimes should not be erased into normal fallback routing.

Recommendation: block merge until degraded boot does not convert explicit runtime intent into fallback routing.

jeong-sik:
Post-merge adversarial follow-up on current merged head.

P0: #23192 merged a default-runtime fallback that violates the no-silent-fallback runtime invariant. In `degrade_loaded_for_missing_catalog`, when the configured `[runtime].default` is absent from the OAS capability catalog but at least one other runtime remains, the code selects the first remaining active runtime as `effective_default`. That means unassigned keepers can run on a provider/model the operator did not configure as default. It is fail-visible in health JSON, but the actual routing still proceeds through an invented default.

Evidence:
- `lib/runtime/runtime.ml` had `effective_default :: _` and reused that as the active default when `configured_default.id` was missing.
- `Keeper_meta_contract.runtime_id_of_meta` sends unassigned keepers to `Runtime.get_default_runtime_id ()`, so changing the loaded default changes production routing for every unassigned keeper.
- The added regression in #23192 explicitly expected `default = "ollama.missing"` to degrade to `"ollama.good"`, so this was an intended behavior, not a theoretical edge.

P1: The correct shape is to degrade only non-default catalog gaps. Assignments, librarian/cross-verifier/structured-judge routes, media failover entries, and lane candidates can be dropped with operator-visible degradation when the configured default is still catalog-known. If `[runtime].default` itself is uncatalogued, boot must fail closed until the operator adds the OAS catalog row or changes the explicit default.

P2: I opened follow-up draft PR #23197 with that fix and focused tests.

P3: CI is the build authority. Local validation on #23197 stayed narrow: ocamlformat check, `git diff --check`, and parser-only checks for the edited files. No local Dune build/test was run.

PR #23196, opened 2026-07-04
[codex] Reject invalid IDE cursor focus modes

jeong-sik:
## Summary
- rejects invalid `focus_mode` values on `POST /api/v1/ide/cursors` with a 400 instead of silently succeeding
- exposes the cursor focus-mode contract from `Ide_bridge` so HTTP parsing uses the same allowed values as cursor storage
- returns a 500 if direct cursor persistence fails, rather than reporting a successful write
- adds focused HTTP tests for invalid and valid cursor focus modes

## Why
The IDE Observation Plane remaining-goals report flagged direct cursor ingestion as a P0 silent-failure path. Before this change, a bad `focus_mode` could be accepted by the HTTP route and then dropped in the bridge without a visible API failure.

## Validation
- `ocamlformat --check lib/ide/ide_bridge.ml lib/ide/ide_bridge.mli lib/server/server_ide_http.ml test/test_server_ide_http.ml`

Local focused Dune test was not run because another parallel slice held `/tmp/me-dune-local.lock`; per operator direction, Dune build/test authority is CI and I did not wait on CI.

### DIFF
--- a/lib/ide/ide_bridge.ml
+++ b/lib/ide/ide_bridge.ml
@@ -266,9 +266,13 @@ let cursor_matches_file file_path json =
      | None -> false)
 ;;
 
-let valid_focus_mode = function
-  | "reading" | "editing" | "reviewing" | "planning" -> true
-  | _ -> false
+let cursor_focus_mode_of_string = function
+  | ("reading" | "editing" | "reviewing" | "planning") as mode -> Some mode
+  | _ -> None
+;;
+
+let valid_focus_mode mode =
+  Option.is_some (cursor_focus_mode_of_string mode)
 ;;
 
 let cursor_is_valid json =
@@ -729,13 +733,15 @@ let ingest_cursor_event
   let column = Option.value column ~default:0 in
   let focus_mode =
     match focus_mode with
-    | None -> Some "editing"
-    | Some mode when valid_focus_mode mode -> Some mode
-    | Some _ -> None
+    | None -> Ok "editing"
+    | Some mode ->
+      (match cursor_focus_mode_of_string mode with
+       | Some mode -> Ok mode
+       | None -> Error "Invalid focus_mode")
   in
   match focus_mode with
-  | None -> ()
-  | Some focus_mode ->
+  | Error msg -> Error msg
+  | Ok focus_mode ->
     let timestamp_ms = now_ms () in
     let json =
       cursor_event_json
@@ -750,11 +756,11 @@ let ingest_cursor_event
         ~turn_id:""
         ()
     in
-    (try append_cursor ~base_dir:base_path ~partition json
+    (try
+       append_cursor ~base_dir:base_path ~partition json;
+       Ok ()
      with exn ->
-       Printf.eprintf
-         "Ide_bridge.ingest_cursor_event error: %s\n%!"
-         (Printexc.to_string exn))
+       Error (Printf.sprintf "Failed to persist cursor event: %s" (Printexc.to_string exn)))
 ;;
 ;;
 
--- a/lib/ide/ide_bridge.ml
+++ b/lib/ide/ide_bridge.ml
@@ -760,7 +760,10 @@ let ingest_cursor_event
        append_cursor ~base_dir:base_path ~partition json;
        Ok ()
      with exn ->
-       Error (Printf.sprintf "Failed to persist cursor event: %s" (Printexc.to_string exn)))
+       Printf.eprintf
+         "Ide_bridge.ingest_cursor_event error: %s\n%!"
+         (Printexc.to_string exn);
+       Error "Failed to persist cursor event")
 ;;
 ;;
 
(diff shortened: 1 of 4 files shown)


jeong-sik:
P1: `column < 0` still has the same silent-success shape this PR is trying to remove.

`POST /api/v1/ide/cursors` validates `line >= 1`, but it forwards `column` without checking it (`lib/server/server_ide_http.ml:770-813`). `Ide_bridge.ingest_cursor_event` then writes that value as-is (`lib/ide/ide_bridge.ml:733-760`), while reads later discard any cursor where `column < 0` (`lib/ide/ide_bridge.ml:286-287`, `lib/ide/ide_bridge.ml:391-399`). That means a request like `{"file_path":"lib/a.ml","line":7,"column":-1,"focus_mode":"reviewing"}` can still return 201 and then disappear from `GET /api/v1/ide/cursors`.

Since this PR is closing the direct cursor ingestion silent-failure path, please either reject negative `column` with 400 in the HTTP route or normalize it to 0 before persistence, and add a regression test for that case.

jeong-sik:
Current-head review response for `2c5a0473d1`.

P0 개선점:
- Preserved the main safety improvement: invalid IDE cursor `focus_mode` now fails with 400 instead of being silently dropped.
- Added one fail-loud-but-redacted hardening fix: cursor persistence exceptions are logged server-side by `Ide_bridge.ingest_cursor_event`, while the returned error is now generic (`Failed to persist cursor event`) instead of embedding `Printexc.to_string` details in the HTTP response path.

P0 남은점 / 병합 가능성:
- No P0 source blocker found after this patch.
- Not mergeable yet: PR is draft and fresh CI is pending.

P1 개선점:
- Closed the internal-detail leak introduced by the fail-loud result path. This keeps the no-silent-failure behavior without returning filesystem/exception internals to a client.

P1 남은점 / 병합 가능성:
- No unresolved review threads on the live GitHub thread surface.
- CI is the build authority; I did not run local Dune.

P2 개선점:
- Local non-Dune checks passed: `git diff --check` and `opam exec -- ocamlformat --check lib/ide/ide_bridge.ml`.

P2 남은점 / 병합 가능성:
- The new check generation is queued/pending after the push. Re-evaluate when GitHub logs complete.

P3 개선점:
- Patch is intentionally surgical: one file, persist-error return path only.

P3 남은점 / 병합 가능성:
- Draft status remains the merge blocker even if checks pass; do not add `ready-for-merge` while draft.

[근거] GitHub `gh pr view 23196 --json headRefOid,mergeStateStatus,isDraft` at 2026-07-04T13:45Z: head `2c5a0473d1`, draft=true, mergeStateStatus=BLOCKED. Confidence: High.
[근거] GitHub `gh pr checks 23196` at 2026-07-04T13:45Z: fresh checks pending/queued. Confidence: High.
[근거] Local commands at 2026-07-04T13:45Z: `git diff --check`; `opam exec -- ocamlformat --check lib/ide/ide_bridge.ml`. Confidence: High.

jeong-sik:
Rechecked current head `2c5a0473`: the persist-error redaction is good, but the P1 above is still open.

`column` is still read with `find_int "column"` and forwarded as-is to `Ide_bridge.ingest_cursor_event`; a negative value can still be accepted with 201 and later filtered out by `cursor_is_valid`. The fix should reject `column < 0` with 400 or normalize it to 0 before persistence.

jeong-sik:
P1 follow-up fixed and pushed: `796be7cdd553`.\n\n변경:\n- `POST /api/v1/ide/cursors` now rejects `column < 0` with 400 before calling `Ide_bridge.ingest_cursor_event`.\n- Added regression `POST cursor rejects negative column` so the previous 201-then-filtered-out silent-success path is pinned closed.\n\n검증:\n- `opam exec -- ocamlformat --check lib/server/server_ide_http.ml test/test_server_ide_http.ml` 통과\n- `git diff --check` 통과\n- parser-only: `ocamlc -stop-after parsing` for both touched files 통과\n- 로컬 Dune은 사용자 지시대로 실행하지 않고 CI에 맡김\n\nMergeability: P1 source blocker addressed. PR is still draft and fresh CI is pending from the push.

jeong-sik (commented):
Re-review for current head `796be7cdd5530d3f34ea32a0f02a09a9f89b39a6`.

P0/P1/P2: 기존 P1은 addressed로 봅니다. `POST /api/v1/ide/cursors` now rejects `column < 0` with 400 before persistence, so the prior 201-then-filtered-out silent-success path is closed. The focus-mode parser still uses the `Ide_bridge.cursor_focus_mode_of_string` contract, and cursor persistence failures stay fail-loud with a redacted client error.

P3: 없음. The change is narrow and keeps validation at the HTTP boundary before calling the bridge.

검증: source review only on the current diff; I did not run local Dune. CI remains the build authority.

Mergeability: source-level blocker 없음. PR is no longer draft in the current metadata, but it still needs current GitHub checks/mergeability to clear before merge.

[근거] `gh pr diff 23196` + `gh pr view 23196 --json headRefOid,isDraft,mergeStateStatus` at 2026-07-04T14:03:22Z. Confidence: High.

PR #23197, opened 2026-07-04
fix(runtime): fail closed on degraded default gaps

jeong-sik:
## Summary

- fail closed when degraded runtime boot finds `[runtime].default` absent from the OAS capability catalog
- keep degraded boot for non-default catalog gaps by disabling those assignments/routes while preserving the configured default
- add focused regression coverage for both allowed non-default degradation and rejected default degradation

## Why

Post-merge review of #23192 found that `degrade_loaded_for_missing_catalog` selected the first remaining catalog-known runtime as the new effective default when the configured default was uncatalogued. That is deterministic, but it still invents a replacement default and routes unassigned keepers through a runtime the operator did not choose. This violates the existing runtime default invariant and the no-silent-fallback review rule.

## Validation

- `opam exec -- ocamlformat --check lib/runtime/runtime.ml lib/runtime/runtime.mli test/test_runtime_config_validity.ml`
- `git diff --check origin/main...HEAD`
- `opam exec -- ocamlc -stop-after parsing lib/runtime/runtime.ml`
- `opam exec -- ocamlc -stop-after parsing lib/runtime/runtime.mli`
- `opam exec -- ocamlc -stop-after parsing test/test_runtime_config_validity.ml`

No local Dune build/test run; CI remains the build authority.

PR #23205, opened 2026-07-04
feat(dashboard): surface runtime startup degradation

jeong-sik:
## Summary
- decode the runtime providers `startup_degradation` payload emitted by the backend
- surface catalog-degraded startup state in the Runtime Health Snapshot first screen
- show effective default, disabled runtimes, missing catalog runtime IDs, and next action

## Why
The backend already emits operator-visible degraded startup data, but the dashboard runtime API decoder dropped the field. Operators could miss that boot continued only after disabling uncatalogued runtimes.

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/dashboard.test.ts src/components/runtime-health-snapshot.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`


jeong-sik (commented):
Reviewed current head `ba7b024f877daa13ac7bf0de17cdcc3c1ae2811d`.

No blocking findings in the changed files. The dashboard decoder matches the backend `Runtime.startup_degradation_to_yojson` contract, keeps the startup payload under the runtime providers response, and the first-screen snapshot now surfaces the degraded boot path with effective default, disabled runtimes, missing catalog runtime IDs, and next action. The added tests cover both API decoding and first-screen rendering for the degraded catalog case.

Residual risk: the new nested decoder follows the existing dashboard best-effort normalization style, so malformed nested startup fields are normalized rather than rejected. I am not calling that a blocker here because `/api/v1/providers` itself still fails on a non-record top-level payload and the backend emits the startup object from a typed OCaml projection.

P0/P1/P2/P3: none from this pass. Mergeability: draft PR, CI still pending at review time.

jeong-sik:
Current-head review/follow-up for `d50b0d0ad684`.\n\nP0/P1/P2: 발견 없음. Dashboard decoder now preserves backend `startup_degradation`, and `RuntimeHealthSnapshot` promotes degraded startup to first-screen warning state instead of letting the runtime panel look healthy. This matches the operator-visible degraded boot requirement.\n\nP3 fixed: tests were still using the old #23192 shape where a catalog-missing configured default degraded into a different effective default. Since #23197 makes that fatal, I aligned the fixtures so configured/effective default match the current backend contract and route labels use `runtime.librarian`. Pushed as `d50b0d0ad684`.\n\n검증:\n- `pnpm --dir dashboard install --frozen-lockfile` 통과\n- `node_modules/.bin/vitest run --config vitest.config.ts src/api/dashboard.test.ts src/components/runtime-health-snapshot.test.ts --no-file-parallelism --maxWorkers=1` 통과: 2 files / 74 tests\n- `pnpm --dir dashboard typecheck` 통과\n- targeted eslint on touched dashboard files 통과 (test-file ignore warnings only, exit 0)\n- `git diff --check` 통과\n\nMergeability: source-level blocker 없음. PR은 draft이고 fresh CI가 pending입니다.

PR #23204, opened 2026-07-04
[codex] Cover IDE memory partition scope

jeong-sik:
## Summary
- adds a focused regression test for `GET /api/v1/ide/memory` partition scoping
- verifies a `By_url` annotation is visible through matching `canonical_url` scope
- verifies the same scoped annotation does not leak through the legacy/default memory read

## Why
The IDE Observation Plane remaining-goals report listed IDE memory endpoint partitioning and backend partition coverage as open work. The current endpoint already resolves the read partition; this PR pins that behavior so future memory adapter work does not regress back to the legacy bucket.

## Validation
- `git diff --check HEAD~1..HEAD`
- `ocamlformat --check test/test_server_ide_http.ml`

Per operator direction, Dune build/test is left to CI and I did not wait on CI.

PR #23198, opened 2026-07-04
[codex] Scope IDE cursor posts by query partition

jeong-sik:
## Summary
- resolves the IDE observation partition for `POST /api/v1/ide/cursors` from the request query
- passes the resolved partition into `Ide_bridge.ingest_cursor_event`
- adds an HTTP regression showing a `canonical_url`-scoped cursor write is visible only through the scoped cursor query and not through legacy default reads

## Stack
Base PR: #23196 (`codex/ide-cursor-strict-focus-20260704`)

## Why
The remaining-goals report marks direct cursor POST partitioning as a P0 gap. This small stack step does not infer from file paths yet, but it makes explicit `canonical_url`/`repo_id` query scope authoritative for cursor writes.

## Validation
- `git diff --check`
- `ocamlformat --check lib/server/server_ide_http.ml test/test_server_ide_http.ml`

Per operator direction, Dune build/test is left to CI and I did not wait on CI.

PR #23206, opened 2026-07-04
[codex] Scope IDE memory panel by repository

jeong-sik:
## Summary
- threads the active IDE repository id into `IdeMemoryPanel`
- includes `repo_id` when fetching `/api/v1/ide/memory`
- adds a focused panel test that the request carries keeper, repository, and limit scope

## Why
The backend memory endpoint is partition-aware, but the right-rail memory panel only sent `keeper_id`. That meant the UI could keep reading the legacy/default memory partition even while the rest of the IDE shell was repository-scoped.

## Validation
- `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-memory-panel.test.ts --no-file-parallelism --maxWorkers=1` -> 1 file / 2 tests passed

Per operator direction, Dune build/test is left to CI and I did not wait on CI.

### DIFF
--- a/dashboard/src/components/ide/ide-memory-panel.test.ts
+++ b/dashboard/src/components/ide/ide-memory-panel.test.ts
@@ -53,4 +53,17 @@ describe('IdeMemoryPanel', () => {
     expect(container.textContent).toContain('semantic:not configured')
     expect(container.textContent).toContain('retrieval:annotation index')
   })
+
+  it('includes keeper and repository scope when fetching memory', async () => {
+    render(h(IdeMemoryPanel, { keeperName: 'sangsu', repoId: 'masc' }), container)
+
+    await waitFor(() => {
+      expect(globalThis.fetch).toHaveBeenCalled()
+    })
+    const [url] = vi.mocked(globalThis.fetch).mock.calls[0]!
+    expect(String(url)).toContain('/api/v1/ide/memory?')
+    expect(String(url)).toContain('keeper_id=sangsu')
+    expect(String(url)).toContain('repo_id=masc')
+    expect(String(url)).toContain('limit=50')
+  })
 })
--- a/dashboard/src/components/ide/ide-memory-panel.ts
+++ b/dashboard/src/components/ide/ide-memory-panel.ts
@@ -34,6 +34,7 @@ interface MemoryResponse {
 
 interface IdeMemoryPanelProps {
   readonly keeperName?: string | null
+  readonly repoId?: string | null
 }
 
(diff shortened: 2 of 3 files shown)


jeong-sik:
Current-head review for `72da08c64ee0`.\n\nP0/P1/P2: 발견 없음. `IdeShell`의 `activeRepositoryId`가 `IdeMemoryPanel`까지 전달되고, memory fetch query에 `repo_id`가 keeper/limit와 함께 포함됩니다. `repoId`가 없을 때는 기존처럼 query에 추가하지 않으므로 `repo_id=null` 같은 잘못된 scope도 만들지 않습니다.\n\nP3: 없음. 테스트도 URLSearchParams 순서에 의존하지 않고 contains로 확인해서 안정적입니다.\n\n검증:\n- `pnpm --dir dashboard install --frozen-lockfile` 통과\n- `node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-memory-panel.test.ts --no-file-parallelism --maxWorkers=1` 통과: 1 file / 2 tests\n- `pnpm --dir dashboard typecheck` 통과\n- `pnpm --dir dashboard exec eslint src/components/ide/ide-memory-panel.ts src/components/ide/ide-memory-panel.test.ts src/components/ide/ide-shell.ts` 통과\n- `git diff --check 6c3a20d...HEAD` 통과\n\nMergeability: source-level blocker 없음. PR은 draft이고 마지막 확인 시 CI가 아직 pending/queued였습니다.

jeong-sik (commented):
Adversarial review (approve).

Threads repoId into IdeMemoryPanel and adds repo_id to the /api/v1/ide/memory query via URLSearchParams (typed, conditional). deps array updated to [keeperName, repoId] so the fetch refires on repo change. Test asserts the URL carries keeper_id, repo_id, limit. Clean.

P2 series-level note (cross-cutting, see the longer comment below): the repo-scope work is split across #23198 / #23204 / #23206 / #23211 / #23214 / #23217. The three FE ones (#23206 / #23214 / #23217) all thread `repoId` from IdeShell into child panels with the same prop pattern. A single "thread active repo id through IDE panels" PR would have lower review/conflict cost. #23204 is already CONFLICTING (same-file adjacency with #23198), which is the visible cost of the split. Each PR is independently testable so the split isn't wrong, but consider consolidating the FE trio. Also: the stacked chain #23210 <- #23214 <- #23217 should be rebased onto main so merge order doesn't stay coupled.

PR #23209, opened 2026-07-04
[codex] Catalog handled LSP methods

jeong-sik:
## Summary
- add a typed handled LSP method catalog with read-only/mutation/lifecycle/status classes
- derive handled method parsing and wire names from that catalog
- expose a small For_testing surface and cover catalog contents/classification

## Validation
- ocamlformat --check lib/server/server_ide_lsp_proxy.ml lib/server/server_ide_lsp_proxy.mli test/test_server_ide_lsp_proxy.ml
- subagent pre-rebase: scripts/dune-local.sh build ./test/test_server_ide_lsp_proxy.exe
- subagent pre-rebase: scripts/dune-local.sh exec ./test/test_server_ide_lsp_proxy.exe (15 tests passed)

After rebase onto origin/main, Dune was intentionally not rerun; CI is the build authority.

### DIFF
--- a/lib/server/server_ide_lsp_proxy.ml
+++ b/lib/server/server_ide_lsp_proxy.ml
@@ -175,6 +175,26 @@ let lsp_method_of_string = Lsp_method_catalog.of_string
 let lsp_method_to_string = Lsp_method_catalog.to_string
 let lsp_method_classification = Lsp_method_catalog.classification
 
+let is_document_sync_notification = function
+  | Did_open | Did_change | Did_save | Did_close -> true
+  | Initialize
+  | Initialized
+  | Shutdown
+  | Exit
+  | Masc_lsp_status
+  | Hover
+  | CodeLens
+  | InlayHint
+  | Diagnostic
+  | Definition
+  | References
+  | Completion
+  | CodeAction
+  | Document_symbol
+  | Folding_range
+  | Document_highlight -> false
+;;
+
 type conn_state =
   { sw : Eio.Switch.t
   ; router : Lsp_message_router.t
@@ -719,6 +739,24 @@ let forward_notification cs lang_id method_ params =
   | Ok proc -> Lsp_message_router.send_notification cs.router proc ~method_ ~params
 ;;
 
+let forward_document_sync_notification cs method_ params =
+  let wire_method = lsp_method_to_string method_ in
+  match extract_uri params with
+  | None -> ()
+  | Some uri ->
+    (match resolve_relative ~base:cs.base_path uri with
+     | None -> ()
+     | Some relative ->
+       if method_ = Did_save
+       then
+         Lsp_overlay_provider.invalidate_cache
+           ~base_dir:cs.base_path
+           ~file_path:relative;
+       (match resolve_lang relative with
+        | Unknown_lang -> ()
+        | Known_lang lang_id -> forward_notification cs lang_id wire_method params))
+;;
+
 (* Read-only method allowlist for the catch-all forwarder (task-1692). The
    observation plane must never forward a write-adjacent request — rename,
    any formatting, executeCommand, applyEdit, willSaveWaitUntil — to the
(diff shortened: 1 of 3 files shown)


jeong-sik (commented):
Current-head review for `1f83e3dadc1b9999677cefd0ef73a65b3758bc25`.

P0/P1/P2: 발견 없음. The typed handled-method catalog removes the duplicated wire-name projection without changing the dispatch semantics: `didOpen`/`didChange`/`didSave`/`didClose` now classify as handled mutation methods, and unmatched handled-notification cases still fall through to the existing `textDocument/did*` forwarding path. Lifecycle/status/read-only handlers remain explicit.

P3: 없음. The test pins canonical order, duplicate absence, and unknown lookup staying absent.

검증: source review only; I did not run local Dune. CI remains the build authority.

Mergeability: source-level blocker 없음. PR is draft and still needs its GitHub checks.

[근거] `gh pr diff 23209` + `gh pr view 23209 --json headRefOid` at 2026-07-04T14:03:22Z. Confidence: High.

jeong-sik:
Current-head review for `1f83e3dadc1`.

P1: the new handled-method catalog is not actually the mutation dispatch SSOT; the old `textDocument/did*` string-prefix path still forwards notifications.

The PR adds closed variants for `Did_open` / `Did_change` / `Did_save` / `Did_close` at `lib/server/server_ide_lsp_proxy.ml:75-95` and classifies them in `Lsp_method_catalog`, but dispatch falls through to `_` and then uses `String.starts_with ~prefix:"textDocument/did" method_str` at `lib/server/server_ide_lsp_proxy.ml:1227-1229`. So the catalog can say exactly four mutation notifications are handled, while the runtime still admits any future/unknown `textDocument/did...` method by prefix. `didSave` cache invalidation also keys off the raw string at `lib/server/server_ide_lsp_proxy.ml:1236` instead of the `Did_save` variant.

That keeps the same no-string-match/no-heuristic weakness this PR is meant to remove. Please route document-sync notifications through the typed variants, e.g. explicit cases for `(Some Did_open | Some Did_change | Some Did_save | Some Did_close), None`, use `Did_save` for cache invalidation, and let unknown `textDocument/did...` notifications drop/reject explicitly rather than prefix-forwarding. Add a regression for an unrecognized `textDocument/didSomethingElse` notification so the old catch-all cannot return.

Mergeability: source-level P1 open. Also note this PR is based on older `c6961745` while `origin/main` is now `7172043`, so rebase will be needed after the source fix.

jeong-sik:
P1 follow-up pushed in `7b8158a808`.

What changed:
- document-sync notifications now dispatch through typed `lsp_method` variants: `Did_open`, `Did_change`, `Did_save`, `Did_close`
- `didSave` cache invalidation now keys off the typed `Did_save` variant instead of raw string equality
- removed the old `String.starts_with ~prefix:"textDocument/did"` catch-all path entirely
- unknown no-id client notifications now fail closed instead of being broadcast to all LSP processes
- added a regression assertion that `textDocument/didSomethingElse` is neither in the handled catalog nor in the read-only forwarder allowlist

Verification:
- `opam exec -- ocamlformat --check lib/server/server_ide_lsp_proxy.ml test/test_server_ide_lsp_proxy.ml`
- `git diff --check`
- `opam exec -- ocamlc -stop-after parsing lib/server/server_ide_lsp_proxy.ml test/test_server_ide_lsp_proxy.ml`
- guard grep confirms no remaining `String.starts_with ~prefix:"textDocument/did"` or `String.equal method_str "textDocument/didSave"` in the touched proxy/test files

I did not run local Dune; CI remains the build authority.

P0/P1/P2: source P1 addressed. Remaining gate is draft status plus fresh GitHub checks, which are queued after the push.

[근거] `gh pr view 23209 --json headRefOid,statusCheckRollup` at 2026-07-04T23:39+0900 shows head `7b8158a808` and queued checks. Confidence: High.

jeong-sik:
Follow-up: the handled_methods dual-source nit from the earlier review is addressed in a new PR on branch fix/lsp-catalog-entries-ssot. entries is now the single explicit source and handled_methods is derived from it; the exhaustive entry match still forces a compile-time case for new variants, and the canonical 20-method catalog test guards entries-list omissions. Closing the loop here.

PR #23210, opened 2026-07-04
[codex] Surface degraded LSP status

jeong-sik:
## Summary
- parse typed `masc/lspStatus` WebSocket notifications into a frontend signal
- surface overlay-only LSP languages in the IDE statusbar instead of hiding degraded state
- reject malformed status payloads without mutating the last valid snapshot

## Validation
- git diff --check
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-lsp-client.test.ts src/components/ide/ide-shell.test.ts --no-file-parallelism --maxWorkers=1

### DIFF
--- a/dashboard/src/components/ide/ide-lsp-client.test.ts
+++ b/dashboard/src/components/ide/ide-lsp-client.test.ts
@@ -1,8 +1,11 @@
 import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
   clearLspDiagnosticSnapshot,
+  EMPTY_LSP_STATUS_SNAPSHOT,
   LspConnection,
   lspDiagnosticSnapshot,
+  lspStatusSnapshot,
+  parseLspStatusSnapshot,
   resolveLspDiagnosticFilePath,
 } from './ide-lsp-client'
 import {
@@ -66,6 +69,7 @@ afterEach(() => {
   vi.restoreAllMocks()
   vi.unstubAllGlobals()
   lspDiagnosticSnapshot.value = new Map()
+  lspStatusSnapshot.value = EMPTY_LSP_STATUS_SNAPSHOT
   mockSockets.length = 0
 })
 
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Current-head review for `59168d5a2937c17d929ed7205e44e567a197446a`.

P0/P1/P2: 발견 없음. The dashboard parser treats `masc/lspStatus` as a typed payload, rejects malformed snapshots without mutating the last valid signal, and the statusbar only surfaces the degraded `overlay_only` languages. That matches the backend shape (`langs[]` with `lang`, `connected`, `overlay_only`, `command`, `last_error`) without adding heuristic parsing or string-based degradation inference.

P3: stale-state semantics on socket close are worth remembering: this PR preserves the last valid status snapshot, so a disconnected LSP socket may still show the last overlay-only chip alongside the connection state. Given the explicit “malformed does not mutate” contract and existing diagnostic lifecycle, I do not see this as a blocker for this slice.

검증: source review only; I did not rerun dashboard tests or typecheck. CI remains the build authority.

Mergeability: source-level blocker 없음. PR is draft and still needs current GitHub checks.

[근거] `gh pr diff 23210` + `gh pr view 23210 --json headRefOid` at 2026-07-04T14:04:57Z. Confidence: High.

jeong-sik:
Current-head review for `59168d5a2937`.

P0/P1: source blocker not found.

What I checked:
- Client now parses `masc/lspStatus` as a typed payload (`langs[]` with `lang`, `connected`, `overlay_only`, `command`, `last_error`) and rejects malformed payloads without mutating the current snapshot.
- The dashboard statusbar only surfaces overlay-only languages, which matches the server-side typed degraded contract already present in `server_ide_lsp_proxy.ml`.
- The statusbar test covers mixed degraded/healthy languages and does not render healthy LSP entries as warnings.
- `git diff --check origin/main...pull/23210/head` passes.

Residual risk / CI:
- I did not run local dashboard/Vitest or Dune; CI is the build authority here.
- `gh pr checks 23210` still had `Detect Changed Surfaces` pending at the time of review; re-evaluate once GitHub finishes.

Mergeability: source-level OK from this review; draft status + pending CI still block readiness.

jeong-sik (commented):
Adversarial review (approve).

parse-don't-validate done right: parseLspLanguageStatus strict-checks every field (typeof lang/connected/overlay_only, isStringOrNull for command/last_error) and returns null on any mismatch; publishLspStatusSnapshot then skips the snapshot mutation entirely on null and emits console.warn. The malformed-payload test asserts both the warning and that the existing snapshot is preserved - good. typed LspLanguageStatus / LspStatusSnapshot interfaces. The statusbar chip surfaces overlay-only langs (e.g. ocamllsp unavailable), which is real operator observability. base=main, no stack coupling.

Pairs with #23209 (OCaml side catalogues the handled methods incl. masc/lspStatus); the two are independent and reviewable separately.

PR #23211, opened 2026-07-04
[codex] Expose IDE snapshot partition metadata

jeong-sik:
## Summary
- add additive IDE snapshot partition metadata for the current Legacy_default/orphan-scoped dashboard snapshot
- expose top-level counts for events, cursors, annotations, regions, and active keepers while preserving existing nested shapes
- add a focused test for the legacy/orphan metadata and empty-fixture counts

## Validation
- git diff --check
- ocamlformat --check lib/server/server_dashboard_http.ml test/test_dashboard_http_core.ml

## Follow-up
- `server_ide_http.ml` already has local partition reason labels; a later SSOT cleanup should move partition wire labels into a shared helper instead of expanding the hot route file in this PR.

### DIFF
--- a/lib/ide/ide_paths.ml
+++ b/lib/ide/ide_paths.ml
@@ -19,6 +19,19 @@ type partition =
   | Base_unresolved
   | Legacy_default
 
+let partition_kind = function
+  | By_url _ -> "by_url"
+  | No_canonical_url -> "no_canonical_url"
+  | Unmatched -> "unmatched"
+  | Base_unresolved -> "base_unresolved"
+  | Legacy_default -> "legacy_default"
+;;
+
+let partition_is_orphan = function
+  | By_url _ -> false
+  | No_canonical_url | Unmatched | Base_unresolved | Legacy_default -> true
+;;
+
 let partition_store_dir ~base_dir = function
   | By_url slug -> by_url_path ~base_dir ~canonical_url:slug
   (* Layout invariant: all non-By_url partitions share [_orphan/] so disk layout
--- a/lib/ide/ide_paths.mli
+++ b/lib/ide/ide_paths.mli
@@ -51,6 +51,13 @@ type partition =
     here instead of dropping them; the reason distinguishes {v2 §7}
     empty-repo / unmatched-repo_id / base-loss / structural-default. *)
 
+val partition_kind : partition -> string
+(** Stable JSON/debug label for [partition]. *)
+
+val partition_is_orphan : partition -> bool
+(** [partition_is_orphan partition] is [true] when [partition] writes to
+    the shared [_orphan/] store. *)
+
 val partition_store_dir : base_dir:string -> partition -> string
 (** [partition_store_dir ~base_dir partition] returns the directory
     that holds [annotations.jsonl] / [regions.jsonl] for the chosen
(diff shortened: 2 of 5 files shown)


jeong-sik:
Current-head review for `794d0142b204`.

P1: `partition_kind` / `partition_orphan` are hand-coded instead of derived from the `Ide_paths.partition` value that actually drives the reads.

At `lib/server/server_dashboard_http.ml:585-638`, the read model selects `let partition = Ide_paths.Legacy_default`, then emits `"legacy_default"` and `true` as separate literals. That happens to match today, but it creates a second partition SSOT: as soon as this snapshot is extended to `By_url _` / `No_canonical_url` / `Unmatched`, the counts can be read from one partition while the metadata still reports the old literal. The test at `test/test_dashboard_http_core.ml:734-745` pins only the literal output, not the invariant between the selected partition and its metadata.

Please derive both fields from the same `partition` value, preferably through a shared encoder/helper near the IDE partition boundary. At minimum, match on `partition` once and compute `partition_kind` plus `partition_orphan` from that variant, with a regression that would fail if the selected partition changes without metadata changing with it.

Why this matters: the IDE Observation Plane docs call out typed scope/repo partition/orphan drift as production P0/P1 work. A dashboard snapshot that duplicates partition truth as string/boolean literals makes the drift harder to detect and violates the no hardcoding / SSOT rule.

Mergeability: source-level P1 open; CI status not evaluated for this comment.

jeong-sik:
P1 follow-up pushed in `ddeccaa4673a82a046c3cb01d297585abc940f8d`.

What changed:
- moved IDE partition metadata to `Ide_paths.partition_kind` and `Ide_paths.partition_is_orphan`
- changed `/dashboard/ide` snapshot JSON to derive `partition_kind` and `partition_orphan` from the same `partition` value used for events/cursors/annotations/regions reads
- added `test_ide_paths` coverage for all current partition variants, including `By_url _` vs orphan-backed variants
- changed the dashboard snapshot test to compare against the `Ide_paths` helper instead of duplicating literals

검증:
- `ocamlformat -i` on touched OCaml files
- `git diff --check`
- `ocamlc -stop-after parsing` on `lib/ide/ide_paths.mli`, `lib/ide/ide_paths.ml`, `lib/server/server_dashboard_http.ml`, `test/test_dashboard_http_core.ml`, `test/test_ide_paths.ml`

P0/P1/P2: addressed for the metadata SSOT issue. Remaining: GitHub checks are queued after the push, and PR is still draft.

Mergeability: source-level blocker cleared; CI-gated now.

[근거] `gh pr view 23211 --json headRefOid,statusCheckRollup` at 2026-07-04T23:22 KST. Confidence: High.

jeong-sik (commented):
Adversarial review (approve).

partition_kind and partition_is_orphan are placed in Ide_paths as the SSOT derivation from the partition variant (exhaustive match) - consumers should not re-implement string labelling per-site. Dashboard snapshot gains additive top-level metadata.

P3 nit: top-level events_count and nested events.count carry the same value (likewise cursors/annotations/regions). If the top-level fields exist only for fast FE reads during a migration window, a comment would prevent future readers from "fixing" the duplication.

PR #23213, opened 2026-07-04
perf(server): offload and cache git blame and git diff endpoints

jeong-sik:
This PR offloads git blame and git diff endpoint computations to worker domains via Domain_pool_ref.submit_io_or_inline, and caches results for 15s using Dashboard_cache.get_or_compute. This prevents rapid parallel file clicks in the IDE dashboard from blocking the main HTTP domain's event loop.

### DIFF
--- a/dashboard/src/components/ide/ide-data-workspace-store.ts
+++ b/dashboard/src/components/ide/ide-data-workspace-store.ts
@@ -1,6 +1,7 @@
 import { signal, effect } from '@preact/signals'
 import { activeIdeFile } from './ide-state'
 import { activeKeeperName } from '../../keeper-state'
+import { route } from '../../router'
 import { selectedTask } from '../goals/task-detail-selection'
 import {
   discoverRepositories,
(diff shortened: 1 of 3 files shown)


jeong-sik:
Review (current head 04b14ec012ee5016ffa2a7dc23f00a71fc200972)

P0/P1/P2 blockers: none found in this pass.

Checked:
- Single-file change in lib/server/server_routes_http_routes_workspace.ml.
- The new blame/diff cache keys include base, workspace source, requested ref where applicable, and relative path, so I do not see an obvious cross-request collision for normal inputs.
- Dashboard_cache compute runs outside the cache table update path, so wrapping git calls in Domain_pool_ref.submit_io_or_inline does not introduce a global cache lock stall.
- The existing blame silent-empty contract for git errors/untracked files is preserved; the only behavior change is that the empty result can now be cached for the realtime TTL.
- Diff error handling preserves the old HTTP 400 surface after unwrapping the cached sentinel.

Residual risk:
- Draft PR and CI is still fully pending at review time.
- I did not find endpoint-level regression coverage for /api/v1/git/blame and /api/v1/git/diff response contracts. Worth adding before ready-for-review, especially around cached success/error paths, but I am not marking that as a blocker for this small perf-only draft.

Mergeability: draft + CI pending.

jeong-sik:
CI follow-up on current head `04b14ec012ee`.

P1 process blocker: `Test Coverage Check` is failing because this PR adds covered server code without any test file changes.

Evidence from `gh run view 28708970524 --job 85139219022 --log`:
- `Test Coverage Check Failed`
- `Added 94 lines to covered code paths but no test files changed. Add tests to cover new functionality.`

The change is not just mechanical: it changes `/api/v1/git/blame` and `/api/v1/git/diff` by introducing offload + cache behavior and a cached success/error sentinel for diff. Please add focused endpoint/cache regression coverage rather than opting out. Minimum useful coverage would prove:
- repeated blame/diff requests reuse the cached result for the same workspace source/path/ref;
- diff git failure still returns the existing 400 JSON error surface after the cached sentinel unwrap;
- cache keys partition repository/workspace source so a repo-scoped request cannot reuse a project/default result.

Mergeability: source review may still be OK, but CI is currently red and this is not ready until coverage is added and the gate is green.

jeong-sik:
Current-head follow-up for `0e17fa32c496`.

P1 still open: the backend cache/offload behavior introduced in `lib/server/server_routes_http_routes_workspace.ml` still has no focused server regression coverage. The new second commit adds dashboard lazy-loading for blame/diff based on the active view, but it does not exercise the changed endpoint contracts:
- `/api/v1/git/blame` now caches the computed JSON result under a workspace/source/path key.
- `/api/v1/git/diff` now caches a success/error sentinel and unwraps it back to the old 200/400 HTTP surface.
- cache keys now carry workspace source/repository partition state and need a regression preventing repo/project cross-reuse.

This means the previous coverage/process blocker was not actually resolved by the current head. Please add focused endpoint/cache tests for the server path rather than only changing the dashboard fetch timing. At minimum, cover same-key cache reuse, diff failure preserving the existing 400 JSON surface after sentinel unwrap, and source/repo partition key separation.

P0/P2: no additional source blocker found in this pass. The dashboard lazy-load change itself is reasonable as a separate perf optimization, but it is not a substitute for testing the server-side cache contract.

Mergeability: not ready. PR is non-draft, but source-level P1 test coverage remains open and GitHub checks are restarted/queued.

[근거] `gh pr diff 23213` and `rg -n "git/blame|git/diff|Dashboard_cache|get_or_compute" test lib dashboard/src` at 2026-07-04T23:36+0900: current head `0e17fa32c496` changes the backend cache path but no server endpoint/cache test file was added. Confidence: High.

jeong-sik (commented):
Adversarial review (P1 - requesting changes via comment, since GitHub does not allow self-request-changes). Code-level issues independent of the CI run.

1. Result encoded as a JSON envelope, then decoded by string field-name match. The diff endpoint wraps its result as {"ok":bool,"data":...} / {"ok":false,"error":...} inside Dashboard_cache, and the response path unpacks it with `List.assoc_opt "ok"` and `List.assoc "data"`. That string-key dispatch is a root-cause:STR smell and only exists because the cache stores JSON rather than a typed value. Root fix: store the (git_run_lines_or_error result, payload) pair as a typed value (use a typed get_or_compute if Dashboard_cache has one, or thread the Result through directly) so no ok/data envelope and no field-name match are needed. The blame path does not need the envelope because it has no Error case - the asymmetry between blame (direct) and diff (envelope) is itself the smell.

2. cache_key source-encoding duplicated. The `match source with `Project -> ... | `Repository repo_id -> ...` block is copy-pasted for both blame and diff keys. Reuse an existing source-to-string helper if present; otherwise extract one. Two copies of a 7-arm match will drift.

3. Indentation regressed. The /api/v1/git/diff handler body is now indented one level deeper than before (with_public_read callback restructured). If CI Lint (ocamlformat) fails, this is the likely cause - keep the original indentation depth if the restructure was not necessary.

4. FE conditional fetch may not react to view changes. `const currentView = route.value.params.view...` is read once at the top of the load block; if this is not inside a reaction that re-runs on route.value changes, switching source -> blame will not trigger a blame fetch. Confirm the surrounding reactivity or subscribe to route.value.

The cache TTL reuse (realtime_cache_ttl_s) and Domain_pool offload are fine in principle; the FE over-fetch reduction is genuine. Blocking items: (1) and (2).

jeong-sik:
Post-merge adversarial follow-up for #23213.

P1 found after merge: the dashboard store reimplemented `IdeShell`'s route view parsing with a narrower local string check. `IdeShell` still normalizes legacy diff aliases like `view=merge` to `split-diff`, but `createIdeDataWorkspaceStore` treated that same route as non-diff and skipped the git diff fetch. That is a route-view SSOT drift and can silently remove the diff data in an accepted legacy route form.

Fix opened as draft follow-up: #23227

#23227 removes the duplicate parser by extracting a shared helper and adds a regression for `split`, `split_diff`, `merge`, `unified`, and `blame` route classification.

[근거] `gh pr view 23213 --json state,mergedAt,mergeCommit` showed #23213 merged at 2026-07-04T14:31:45Z; local diff review of merged `ide-data-workspace-store.ts` vs `ide-shell.ts`; follow-up PR #23227 head `5802235b88d9d6abcf8cafe35e3a61b1abb73c4f`. Confidence: High.

jeong-sik:
Follow-up: the ok/data string-key dispatch finding from the earlier review is addressed in a new PR (search open PRs for "collapse git-diff cache envelope and de-dup source_to_string" on branch fix/git-diff-cache-envelope). It removes the {ok,data} envelope and List.assoc_opt/assoc dispatch, collapses the 3 inline source-variant matches + source_header into a single source_to_string SSOT, and fixes the +1-space indent drift. No FE change (the FE already reads the bare payload). Marking here so the review thread is closed-loop.

PR #23216, opened 2026-07-04
feat(dashboard): add dismiss button for keeper action errors

jeong-sik:
Adds a dismiss (✕) button to keeper action error messages to allow manual clearing of stuck red error text.

### DIFF
--- a/dashboard/src/components/keeper-shared.ts
+++ b/dashboard/src/components/keeper-shared.ts
@@ -622,6 +622,7 @@ export function KeeperConversationPanel({
         <span class="flex-1">${error}</span>
         <button
           type="button"
+          aria-label="에러 메시지 닫기"
           class="shrink-0 text-[var(--bad-light)] opacity-70 hover:opacity-100 transition-opacity ml-1 cursor-pointer font-bold select-none"
           title="에러 메시지 닫기"
           onClick=${() => setRecordValue(keeperActionErrors, keeperName, null)}
--- a/dashboard/src/components/keeper-shared.ts
+++ b/dashboard/src/components/keeper-shared.ts
@@ -35,6 +35,7 @@ import {
   keeperStreamStartedAt,
   keeperStreamLastEventAt,
   keeperThreads,
+  setRecordValue,
 } from '../keeper-state'
 import { isDefaultVisibleConversationEntry } from '../keeper-state'
 import {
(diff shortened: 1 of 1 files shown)


jeong-sik:
Review finding (conditional before merge)

P2: The new keeper action-error dismiss control is icon-only but has no explicit accessible name. Because the button content is `✕`, assistive tech/voice control will expose the control as the glyph rather than `에러 메시지 닫기`; `title` is not enough here because the button already has text content. Please add an explicit `aria-label` to the dismiss button, and ideally add a small `KeeperConversationPanel` regression that sets `keeperActionErrors`, clicks the dismiss control, and asserts the alert clears.

Validation I ran while reviewing:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (21 tests passed)
- `pnpm --dir dashboard exec eslint src/components/keeper-shared.ts src/components/keeper-shared.test.ts` (0 errors; test file is ignored by config)

Merge recommendation: conditional. The implementation path is otherwise narrow, but the dismiss affordance should be accessible before merge.

Note: GitHub would not let me submit this as a request-changes review because it is authored by the same account, so I am leaving the review finding as a PR comment.

jeong-sik:
Source review pass for #23216.

Reviewed `dashboard/src/components/keeper-shared.ts` against the existing `keeperActionErrors` lifecycle. The dismiss path only clears the current keeper's action error via `setRecordValue(..., null)`, and the existing send/probe/recover paths clear stale errors before a new action and set fresh failures again, so I do not see a stuck-hidden-error regression.

Local verification on PR head `2dd99347338804c071f3a64da46fdb7b711e7da3`:
- `pnpm --dir dashboard typecheck` ✅
- `pnpm --dir dashboard lint` ✅
- `pnpm --dir dashboard exec vitest run src/components/keeper-shared.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1` ✅ (21 tests)

P0/P1/P2: none found.
P3: optional a11y polish only — the close button has visible `✕` plus `title`; adding an explicit `aria-label="에러 메시지 닫기"` would make the accessible name less dependent on glyph/title behavior, but I would not block merge on it.

Mergeability: source-review pass; GitHub required checks were still pending/BLOCKED at review time, so wait for CI before merge.

jeong-sik:
Current-head review for `2dd993473388`.

P0/P1 남은점:
- No P0/P1 source blocker found in this pass. The change is narrowly scoped to `KeeperConversationPanel` error rendering, and dismiss uses the existing `keeperActionErrors` signal path instead of introducing a parallel hidden state.

P2 남은점:
- The dismiss button is icon-only. It has `title="에러 메시지 닫기"`, but the accessible button name can still resolve to the visible `✕` text rather than the intent. Prefer adding an explicit `aria-label="에러 메시지 닫기"` and a focused test that clicks the button and verifies the per-keeper error clears.

P3 / CI:
- CI is still pending. I did not run a full dashboard suite or Dune locally.
- Focused PR-head check passed: `/Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run src/components/keeper-shared.test.ts` from `.worktrees/review-pr-23216-20260704/dashboard`.

[근거] `gh pr view/checks 23216` and focused PR-head vitest at 2026-07-04T23:30:01+0900: head `2dd993473388`, one changed file, checks pending, focused test file 21/21 passed. Confidence: High.

jeong-sik:
Current-head follow-up for `8b956d5626`.

P0/P1: none found in the changed keeper error-dismiss path.

P2 fixed: added explicit `aria-label="에러 메시지 닫기"` to the icon-only dismiss button so the accessible name no longer depends on the visible `✕` glyph or `title`. Pushed as `8b956d5626`.

P3: a dedicated click-to-dismiss regression would still be useful, but this PR now has focused component coverage passing on the touched component.

Verification run from clean worktree `.worktrees/review-23216-a11y-20260704`:
- `pnpm --dir dashboard lint`
- `node_modules/.bin/vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (21 tests)
- `git diff --check`

Mergeability: source-level blocker cleared. PR is non-draft, but GitHub checks restarted after the push and are still pending/BLOCKED, so do not merge until required CI is green.

[근거] `gh pr view 23216 --json headRefOid,statusCheckRollup` at 2026-07-04T23:31:00+0900 shows head `8b956d5626` and queued checks. Confidence: High.

jeong-sik (commented):
Adversarial review (approve).

renderError helper consolidates three inline error renders and adds a dismiss (x) button via setRecordValue(keeperActionErrors, keeperName, null). This reduces duplication rather than adding it, and uses the typed signal surface. No P-level issues.

PR #23219, opened 2026-07-04
[codex] Surface recent cache usage deltas

jeong-sik:
## Summary

- Threads OAS-reported cache read/write token deltas through model inference recent entries.
- Adds `cache_read_tokens` and `cache_creation_tokens` to the runtime model metrics JSON contract.
- Shows a compact `cache r/w` column in Runtime Monitor recent-turn details so operators can inspect per-turn cache behavior, not only aggregate totals.

## Why

The cost ledger already records cache token deltas, and aggregate model metrics already compute total cache read/write. The recent-turn surface dropped those fields, which made token/cache investigations depend on raw JSONL instead of the dashboard/API view.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/dashboard.test.ts src/components/runtime-monitor.test.ts --no-file-parallelism --maxWorkers=1`
- `ocamlformat --check lib/model_inference_metrics/model_inference_metrics_entry.ml lib/model_inference_metrics/model_inference_metrics_entry.mli lib/model_inference_metrics/model_inference_metrics.mli lib/model_inference_metrics/model_inference_metrics_aggregate.ml lib/model_inference_metrics/model_inference_metrics_json.ml test/test_model_inference_metrics.ml`

Blocked local check:

- `./scripts/dune-local.sh build @test/runtest-test_model_inference_metrics` could not complete because another worktree held `/tmp/me-dune-local.lock` (`keeper-busy-root-cause-20260704/scripts/dune-local.sh ...`). Earlier attempt before the lock also hit an unrelated `lib/ide/ide_bridge.ml` compile error outside this diff. CI remains the build authority for the OCaml target.


jeong-sik:
Current-head review for `8af04870cb99`.

P0/P1 남은점:
- No P0/P1 source blocker found in this pass.
- The cache fields already exist in the raw parser path; this PR carries them through `recent_entry`, aggregate projection, JSON, dashboard decode, and the recent-turn table without inventing a frontend-side heuristic.
- Missing/untrusted usage stays explicit: the UI reuses `recentEntryMissingLabel` instead of silently turning missing cache tokens into zero.

P2 남은점:
- `cache r/w` is compact and understandable in the dense table, but if this becomes an operator-facing diagnosis surface, a title/help label could spell out `read/creation` to avoid confusing cache creation with ordinary output tokens. Not a merge blocker.

Mergeability:
- Draft and CI pending/BLOCKED. I did not run local Dune; CI remains build authority.
- Local structural check passed: `git diff --check origin/main...refs/remotes/pull/23219/head`.

[근거] `gh pr view/diff/checks 23219` and local `git diff --check` at 2026-07-04T23:35:38+0900: head `8af04870cb99`, 10 changed files, checks pending. Confidence: High.

jeong-sik (commented):
P2: cache missing-state rendering duplicates the reason label.

`dashboard/src/components/runtime-monitor.ts:1028` formats each side through `fmtRecentEntryNumber`, but that helper already converts a missing value into `recentEntryMissingLabel(entry)`, not necessarily `MISSING_DATA_DASH`. For rows with both cache fields null, such as the API fixture with `usage_reported: false` / `telemetry_reported: false`, `read` and `creation` both become `no-telemetry`, the `read === MISSING_DATA_DASH && creation === MISSING_DATA_DASH` guard is false, and the table renders `no-telemetry/no-telemetry` instead of one compact missing reason. The current runtime-monitor test only covers numeric `128/16`, so this regression is not caught.

Please branch on the raw cache token nullability before formatting, or return a value+missing flag from the formatter, and add a runtime-monitor test for a recent entry where both cache fields are null. Mergeability: P2 until this display regression is fixed.

jeong-sik (commented):
Adversarial review (approve).

Clean additive typed change: re_cache_read_tokens / re_cache_creation_tokens (int option) added to the recent_entry record, threaded through aggregate, serialised via Json_util.int_opt_to_json, decoded on the FE with asNumber, and rendered as a "cache r/w" column. No string-match dispatch anywhere; the int option preserves missing-usage semantics (null on both sides), which the missing-usage test asserts. Tests cover recent-entry ordering, json roundtrip, and the null case. Good.

jeong-sik:
Follow-up on current head `0153393476`: fixed one P2 recent-entry display issue.

Issue fixed:
- When both per-turn cache fields were absent, the `cache r/w` formatter could render duplicated missing labels such as `no-usage/no-usage`.
- That made the missing-data reason noisier than the rest of the recent-turn table and could be mistaken for two distinct states.

Patch applied:
- If both `cache_read_tokens` and `cache_creation_tokens` are null/absent, render the single `recentEntryMissingLabel(entry)` reason.
- Kept partial visibility for asymmetric data, e.g. one side present and the other missing.
- Added a runtime monitor regression test for the absent-cache case.

Validation from `.worktrees/cache-usage-turn-observability-20260704`:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/runtime-monitor.test.ts src/api/dashboard.test.ts --no-file-parallelism --maxWorkers=1` — 2 files / 75 tests passed

I did not run local Dune; CI remains the build authority for the OCaml part of this PR. GitHub state after push: `MERGEABLE`, draft, checks pending.

PR #23217, opened 2026-07-04
[codex] Subscribe cursor stream by repository

jeong-sik:
## Summary
- wire the IDE shell to the dedicated cursor SSE stream instead of relying only on pre-seeded cursor state
- add `repo_id` to cursor stream URLs when an active repository is selected
- keep the backend stream contract unchanged; it already resolves partition from the request URI

## Stacking
- Base branch: `codex/ide-activity-repo-scope-20260704` (#23214)
- This keeps the `ide-shell.ts` changes in one stack: #23210 -> #23214 -> this PR.

## Validation
- git diff --check
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/keeper-cursor-overlay.test.ts src/components/ide/ide-shell.test.ts --no-file-parallelism --maxWorkers=1
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty

Dune/CI was not run or waited on per instruction.

PR #23214, opened 2026-07-04
[codex] Scope IDE activity events by repository

jeong-sik:
## Summary
- pass the active repository id from the IDE shell into the activity rail
- scope only the IDE bridge event read path (`/api/v1/ide/events`) by `repo_id`, while keeping the general activity graph feed unchanged
- add a focused query contract test for the bridge events request

## Stacking
- Base branch: `codex/lsp-statusbar-degraded-20260704` (#23210)
- This avoids touching the same `ide-shell.ts` hunk independently on `main`.

## Validation
- git diff --check
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-activity-panel.test.ts --no-file-parallelism --maxWorkers=1
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc --noEmit --pretty

Dune/CI was not run or waited on per instruction.

PR #23220, opened 2026-07-04
[codex] Surface IDE activity bridge failures

jeong-sik:
## Summary
- keep graph activity visible when the IDE bridge event feed fails
- surface bridge-only failures as `bridge degraded` in the activity rail refresh status instead of silently treating them as empty activity
- add a focused regression test for graph activity plus failed `/api/v1/ide/events`

## Stacking
- Base branch: `codex/ide-activity-repo-scope-20260704` (#23214)
- This stays on the activity panel stack and does not touch the cursor stream stack (#23217).

## Validation
- git diff --check
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-activity-panel.test.ts --no-file-parallelism --maxWorkers=1

### DIFF
--- a/dashboard/src/components/ide/ide-activity-panel.test.ts
+++ b/dashboard/src/components/ide/ide-activity-panel.test.ts
@@ -338,6 +338,43 @@ describe('IdeActivityPanel', () => {
     expect(url.searchParams.get('limit')).toBe('50')
   })
 
+  it('surfaces IDE bridge event fetch failures without dropping graph activity', async () => {
+    vi.stubGlobal('fetch', vi.fn(async input => {
+      const url = String(input)
+      if (url.includes('/api/v1/ide/events')) {
+        return Promise.reject(new Error('ide bridge events unavailable'))
+      }
+      return new Response(JSON.stringify({
+        events: [{
+          seq: 1,
+          ts_ms: 100,
+          ts_iso: '2026-05-05T10:00:00Z',
+          workspace_id: 'run-default',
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head review for `33d06cd0e4e3`.

P0/P1 남은점:
- No P0/P1 source blocker found in this pass.
- The bridge boundary is no longer silent: `/api/v1/ide/events` failure returns an explicit `bridgeOk=false`, the panel shows `bridge degraded N failed`, and graph activity remains rendered from `/api/v1/activity/events`.
- Repo scoping is preserved through the existing `repoId` prop into `fetchIdeEvents({ limit: 50, repoId })`.

P2 남은점:
- The degraded title is intentionally generic (`IDE bridge events unavailable`). If operators need root-cause detail later, carry a typed/redacted bridge error reason rather than raw exception text. I would not block this PR on that because the current change already removes the silent-drop behavior without leaking internals.

Mergeability:
- Draft and stacked on base `c09f7a5...`; not independently mergeable until the stack/base and CI are green.
- GitHub checks are pending. I did not run local dashboard/Dune suites.
- Local structural check passed: `git diff --check origin/main...refs/remotes/pull/23220/head`.

[근거] `gh pr view/diff/checks 23220` and local `git diff --check` at 2026-07-04T23:33:38+0900: head `33d06cd0e4e3`, PR diff touches `ide-activity-panel.ts` and its test, checks pending. Confidence: High.

PR #23222, opened 2026-07-04
[codex] Align IDE annotation mutations with token identity

jeong-sik:
## Summary
- remove client-supplied `keeper_id` from IDE annotation create requests
- remove caller-supplied `keeper_id` from annotation delete URLs
- align the dashboard API wrapper with the backend contract: token identity is the only mutation identity authority

## Validation
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/api/ide.test.ts --no-file-parallelism --maxWorkers=1

### DIFF
--- a/dashboard/src/api/ide.test.ts
+++ b/dashboard/src/api/ide.test.ts
@@ -67,22 +67,24 @@ describe('ide API', () => {
     expect(url).toContain('repo_id=masc')
   })
 
-  it('createIdeAnnotation appends workspace params to mutation URL', async () => {
+  it('createIdeAnnotation relies on token identity and appends workspace params', async () => {
     stubFetch({ ok: true, data: annotation })
 
     await createIdeAnnotation({
       file_path: 'lib/a.ml',
       line_start: 1,
       line_end: 1,
-      keeper_id: 'sangsu',
       kind: 'Comment',
       content: 'Review this line',
     }, { keeper: 'sangsu', repoId: 'masc' })
 
     const url = String(mockFetch.mock.calls[0]![0])
+    const init = mockFetch.mock.calls[0]![1] as RequestInit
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head review (608ae4c09b): no P0/P1/P2/P3 source blockers found.

The frontend now follows the server-side token-identity contract for IDE annotation mutations: create no longer sends a body `keeper_id`, and delete no longer sends a query `keeper_id`. I checked the backend route contract in `lib/server/server_ide_http.ml`; POST derives the acting keeper from auth identity and rejects body-supplied `keeper_id`, while DELETE binds deletion ownership to auth identity. No stale frontend call sites remain.

Local focused verification from `.worktrees/ide-annotation-token-identity-20260704`:
- `node_modules/.bin/vitest run --config vitest.config.ts src/api/ide.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard lint`

Mergeability: draft PR, with GitHub checks still pending on this head.

jeong-sik (commented):
Adversarial review (approve - security improvement).

Removing keeper_id from createIdeAnnotation's body and deleteIdeAnnotation's signature, and dropping the keeper_id query param, shifts keeper identity to the bearer token (server-derived) instead of client self-declaration. That closes a tampering path where a client could attribute an annotation to another keeper. Tests assert both the URL no longer carries keeper_id and the body no longer has the property.

Note (please confirm): deleteIdeAnnotation's signature changed (dropped the keeperId positional arg) - all call sites must be updated in the same PR or the dashboard typecheck fails. The CI Dashboard check will catch any missed site; just flagging since the diff shown only touches tests.

PR #23212, opened 2026-07-04
fix(dashboard): refresh runtime catalog after settings saves

jeong-sik:
## Summary
- await forced shared runtime catalog reloads during runtime config refresh so save completion reflects fresh runtime-name consumers or surfaces refresh failures
- move verbose runtime catalog diagnostics out of capability pills into collapsed structured rows
- add focused coverage for catalog reload success/failure, config refresh awaiting, and the settings catalog card layout contract

## Validation
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/vitest run --config vitest.config.ts src/lib/runtime-catalog-resource.test.ts src/lib/runtime-config-refresh.test.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/tsc -p tsconfig.json --noEmit --pretty false
- /Users/dancer/me/workspace/yousleepwhen/masc/dashboard/node_modules/.bin/eslint src/components/settings-surface.ts src/lib/runtime-catalog-resource.ts src/lib/runtime-config-refresh.ts src/lib/runtime-catalog-resource.test.ts src/lib/runtime-config-refresh.test.ts (settings-surface linted clean; src/lib files are outside current eslint matching and reported ignored warnings)
- Playwright rendered check at http://127.0.0.1:5178/dashboard/#settings: desktop and mobile catalog cards, diagnostics closed, no .rt-cap.tcf debug pills, unintended overflow count 0

CI remains the build authority.

PR #23225, opened 2026-07-04
fix(dashboard): edit fusion preset runtimes

jeong-sik:
## Summary

- add a writable Fusion preset composition editor for the active flat preset: panel runtime chips/add/remove and a judge/JoJ runtime select sourced from runtime.toml bindings
- persist `panel = [...]` and top-level `judge = "..."` through the existing raw runtime config save path, then refresh runtime consumers after save
- keep grouped `[[fusion.presets.<preset>.panels]]` presets fail-visible/read-only instead of flattening them, and harden the string-array writer so multi-line TOML arrays are replaced as one value

## Notes

- This is the follow-up for the Fusion gap left after #23212: the Settings page could save scalar Fusion settings but could not configure the actual panel/JoJ runtime composition.
- `[[fusion.presets.<preset>.judges]]` first-pass judge tables are still displayed as JoJ context but not edited in this slice; the writable field here is the top-level meta judge/JoJ runtime.
- Full build verdict is CI-owned per repo workflow.

## Validation

- `pnpm exec vitest run --config vitest.config.ts src/lib/fusion-settings.test.ts src/lib/runtime-toml-config.test.ts src/components/fusion-settings-panel.test.ts src/lib/runtime-catalog-resource.test.ts src/lib/runtime-config-refresh.test.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec tsc -p tsconfig.json --noEmit --pretty false`
- `pnpm exec eslint src/components/fusion-settings-panel.ts src/components/fusion-settings-panel.test.ts src/lib/fusion-settings.ts src/lib/fusion-settings.test.ts src/lib/runtime-toml-config.ts src/lib/runtime-toml-config.test.ts src/styles/settings-surface.css`
  - exit 0; current ESLint config reports ignore warnings for several `src/lib/*`, test, and CSS paths
- `python3 /tmp/masc_fusion_playwright_smoke.py`
  - mocked `/api/v1/runtime/config/raw`; verified POST contains `panel = ["a.one", "b.two", "c.three"]` and `judge = "meta.new"` and drops stale multiline array lines
  - console: one existing PerformanceMonitor slow-frame warning, no save/runtime-config errors

### DIFF
--- a/dashboard/src/components/fusion-settings-panel.test.ts
+++ b/dashboard/src/components/fusion-settings-panel.test.ts
@@ -23,6 +23,14 @@ judge = "j"
 min_answered = 1
 `
 
+const SAMPLE_WITH_RUNTIME_OPTIONS = `${SAMPLE}
+[new.x]
+is-default = false
+
+[meta.judge]
+is-default = false
+`
+
 const cfg = (over: { ok?: boolean; source_text?: string; reloaded?: boolean }) => ({
   ok: over.ok ?? true,
   path: null,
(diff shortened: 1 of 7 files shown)


jeong-sik:
Current-head review (5424f9e436): no P0/P1/P2/P3 source blockers found.

Checked the new fusion preset composition editor and TOML writer path. The flat preset writer updates `panel` as a string array, preserves other preset keys such as `min_answered`, deletes empty `judge`, and fails closed for grouped `[[fusion.presets.<preset>.panels]]` presets instead of flattening them. The UI only exposes the composition editor for flat preset views.

Local focused verification from `.worktrees/dashboard-runtime-settings-refresh-20260704`:
- `pnpm --dir dashboard lint`
- `node_modules/.bin/vitest run --config vitest.config.ts src/components/fusion-settings-panel.test.ts src/lib/fusion-settings.test.ts src/lib/runtime-toml-config.test.ts --no-file-parallelism --maxWorkers=1`

Mergeability: draft PR, and GitHub checks are still pending on this head.

jeong-sik (commented):
Adversarial review (approve).

Flat fusion preset panel/judge runtime editor. Strong points: grouped presets are rejected (hasArrayOfTables section 'panels' -> throw) so the flat writer cannot corrupt a grouped preset (fail-closed); panel is deduped via uniqueNonEmpty; judge empty-string branch deletes the key instead of writing an empty value. The toml array rewrite test asserts no stale elements survive (old.a/old.b absent after rewrite), and tomlArrayValueEndLine parses quote/escape/depth properly rather than line-splitting - good defensive parsing.

Context for reviewers: this implements part of RFC-0306. The known "of_preset JoJ>=2 silent" concern lives in the grouped/judge-of-judges path (fusion-preset-view), which this PR deliberately does not touch (it only edits flat presets and rejects grouped). That JoJ>=2 path should be tracked separately - please confirm there is a follow-up issue or PR covering it, since this PR will not fix it.

jeong-sik:
Follow-up on current head `4e5d5b4f23`: fixed one P1/P2 config-writer issue from the second pass.

Issue fixed:
- The save path always applied the flat preset composition writer for any non-grouped preset.
- For a preset section with no `panel` key, a no-op save could synthesize `panel = []`, changing an absent config field into an explicit empty panel.

Patch applied:
- Compare the loaded flat preset composition with the draft before writing `panel` / `judge`.
- Skip `applyFusionPresetComposition` when the composition did not change.
- Added a regression test proving an unchanged no-panel preset save preserves the missing `panel` key.

Validation from `.worktrees/dashboard-runtime-settings-refresh-20260704`:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/fusion-settings-panel.test.ts src/lib/fusion-settings.test.ts src/lib/runtime-toml-config.test.ts --no-file-parallelism --maxWorkers=1` — 3 files / 79 tests passed

GitHub state after push: `MERGEABLE`, draft, checks pending. Remaining blockers are draft status and CI completion.

PR #23227, opened 2026-07-04
fix(dashboard): share IDE route view parsing

jeong-sik:
## Summary
- Extract the IDE route view normalization into a shared helper used by both `IdeShell` and `createIdeDataWorkspaceStore`.
- Keep legacy diff aliases such as `split`, `split_diff`, and `merge` mapped to `split-diff` before deciding whether to fetch git diff data.
- Add a focused helper regression so route aliases do not drift again.

## Why
Post-merge review of #23213 found that the dashboard store reimplemented `IdeShell`'s route view parsing with a narrower string check. That made legacy `view=merge` normalize to split diff in the shell while the store treated it as a non-diff view, silently suppressing the git diff fetch. This PR removes the duplicate parser instead of widening another local string match.

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/ide/ide-data-workspace-store.test.ts src/components/ide/ide-shell.test.ts --no-file-parallelism --maxWorkers=1` (2 files / 57 tests)

Local Dune was not run; CI remains build authority.

### DIFF
--- a/dashboard/src/components/ide/ide-data-workspace-store.test.ts
+++ b/dashboard/src/components/ide/ide-data-workspace-store.test.ts
@@ -8,6 +8,7 @@ import {
   workspaceFetchIssueFromError,
   workspaceTreeIdentity,
 } from './ide-data-workspace-store'
+import { isDiffEditorView, viewFromRoute } from './ide-view-route'
 import type { Repository } from '../../api/repositories'
 
 function repo(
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve).

Good SSOT extraction. ide-shell's local viewFromRoute and the store's inline `currentView === 'split-diff' || 'split' || 'unified' || 'diff'` check were inconsistent (the store branch did not treat 'merge' as a diff view and invented a 'diff' alias the shell helper did not). Moving viewFromRoute + isDiffEditorView into ide-view-route.ts and consuming it from both sites removes the divergence. Legacy alias handling (split / merge -> split-diff) is now shared.

This is the right follow-up to the inline-string-match concern on #23213. Note: this PR extracts the helper but does not by itself fix #23213's reactivity issue (whether route.value changes refetch blame/diff) - that depends on the surrounding effect subscription, please confirm separately.

jeong-sik:
P0/P1/P2/P3: no source blockers found on current head `5802235b88d9d6abcf8cafe35e3a61b1abb73c4f`.

Reviewed the SSOT extraction for IDE route view parsing:

- `viewFromRoute` now normalizes `split`, `split_diff`, and `merge` through one helper.
- `isDiffEditorView` is shared by the workspace store instead of repeating route string checks.
- This fixes the previous divergence where the store could suppress diff fetches for the shell-supported `merge` alias.

Local validation run in `.worktrees/ide-route-view-ssot-20260704`:

- `node_modules/.bin/vitest run --config vitest.config.ts src/components/ide/ide-data-workspace-store.test.ts --no-file-parallelism --maxWorkers=1` -> 21 passed
- `pnpm --dir dashboard lint`

Mergeability: still draft, and some GitHub checks were queued when checked, so not merge-ready until CI finishes green.

PR #23228, opened 2026-07-04
[codex] Fail closed on IDE API response drift

jeong-sik:
## Summary
- make IDE read helpers reject failed or malformed response envelopes instead of returning empty/null results
- validate annotation, region, cursor, and event rows strictly so schema drift flows into existing workspace degraded issue handling
- keep mutation helper behavior unchanged in this slice to avoid widening the blast radius

## Validation
- pnpm --dir dashboard exec vitest run src/api/ide.test.ts
- git diff --check

Dune build is left to CI per repo workflow.

### DIFF
--- a/dashboard/src/api/ide.test.ts
+++ b/dashboard/src/api/ide.test.ts
@@ -67,6 +67,28 @@ describe('ide API', () => {
     expect(url).toContain('repo_id=masc')
   })
 
+  it('fetchIdeAnnotations rejects failed envelopes instead of returning empty', async () => {
+    stubFetch({ ok: false, error: 'repo scope unmatched' })
+
+    await expect(fetchIdeAnnotations()).rejects.toThrow('repo scope unmatched')
+  })
+
+  it('fetchIdeAnnotations rejects malformed rows instead of dropping them', async () => {
+    stubFetch({ ok: true, data: [null] })
+
+    await expect(fetchIdeAnnotations()).rejects.toThrow(
+      'fetchIdeAnnotations returned malformed row at index 0',
+    )
+  })
+
+  it('fetchIdeAnnotations rejects rows with missing required fields', async () => {
+    stubFetch({ ok: true, data: [{ ...annotation, id: '' }] })
+
+    await expect(fetchIdeAnnotations()).rejects.toThrow(
+      'fetchIdeAnnotations returned malformed row at index 0',
+    )
+  })
+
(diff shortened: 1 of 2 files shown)


jeong-sik:
P0/P1/P2/P3: no source blockers found on current head `0fb43d05d026232ed54b6a68e3aef24fefb93ef1`.

Reviewed the fail-closed IDE API parsing changes:

- failed envelopes now throw the backend error instead of silently returning empty data;
- malformed annotation/region/event/cursor rows now surface diagnostics instead of being dropped;
- cursor parsing still validates the snapshot and now also detects row drops by comparing raw row count to parsed cursor count.

Local validation run in `.worktrees/ide-api-diagnostics-20260704`:

- `git diff --check origin/codex/ide-annotation-token-identity-20260704...HEAD`
- `node_modules/.bin/vitest run --config vitest.config.ts src/api/ide.test.ts --no-file-parallelism --maxWorkers=1` -> 11 passed
- `pnpm --dir dashboard lint`

Mergeability: source looks clean, but this is stacked on #23222 and still draft, so it is not independently merge-ready.

jeong-sik:
Follow-up on current head `c875e8d4e8`: fixed one P2 fail-closed gap found during the second pass.

Issue fixed:
- The earlier strict row path still used the lenient annotation/region schema parsers, which coerce missing fields to empty strings or zero. That meant some IDE API response drift could still pass as valid data.

Patch applied:
- Added strict API-boundary parsers for annotation and region rows.
- `fetchIdeAnnotations` and `fetchIdeRegions` now reject missing/invalid required row fields instead of accepting coerced defaults.
- `createIdeAnnotation` now uses the same failed-envelope and strict-row path instead of silently returning `null` on failed/malformed envelopes.
- Left the shared lenient schema parser untouched; the strict contract is scoped to `dashboard/src/api/ide.ts`.

Validation from `.worktrees/ide-api-diagnostics-20260704`:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/ide.test.ts --no-file-parallelism --maxWorkers=1` — 1 file / 15 tests passed
- `git diff --check origin/codex/ide-annotation-token-identity-20260704...HEAD`

GitHub state after push: `MERGEABLE`, draft, `UNSTABLE` while new checks are pending. Remaining blockers are draft status, #23222 base ordering, and CI completion.

PR #23226, opened 2026-07-04
[codex] Surface IDE cursor stream degradation

jeong-sik:
## Summary
- connect the IDE shell cursor rail to the repository-scoped cursor SSE stream only after an active repo id is available
- carry cursor stream lifecycle state in the cursor overlay snapshot and render degraded/closed/live state in the cursor rail
- include repo_id and dashboard bearer token in the cursor stream URL builder

## Validation
- pnpm --dir dashboard exec vitest run src/components/ide/keeper-cursor-overlay.test.ts src/components/ide/ide-shell.test.ts
- pnpm --dir dashboard exec tsc --noEmit --pretty false
- git diff --check

Dune build is left to CI per repo workflow.

PR #23229, opened 2026-07-04
fix(ide): allow query auth for cursor SSE

jeong-sik:
## Summary
- Fix the post-#23226 cursor SSE auth path so EventSource `token=` query auth is actually accepted under strict HTTP auth.
- Route `/api/v1/ide/cursors/stream` through a query-token-aware SSE read wrapper instead of generic `with_public_read` -> `with_read_auth`.
- Add a black-box strict-auth regression for `/api/v1/ide/cursors/stream?token=...`.

## Why
#23226 added `token=` to the cursor EventSource URL because EventSource cannot send Authorization headers. The merged server route still used the generic read-auth path, which only reads headers; under strict HTTP auth that query token would be ignored.

## Validation
- `scripts/dune-local.sh build test/test_server_ide_http.exe`
- `./_build/default/test/test_server_ide_http.exe` -> 20 tests passed

### DIFF
--- a/lib/server/server_auth.ml
+++ b/lib/server/server_auth.ml
@@ -90,7 +90,8 @@ let observer_sse_query_token_from_request request =
   match request.Httpun.Request.meth with
   | `GET
     when (observer_stream_requested && String.equal path "/mcp")
-         || String.equal path "/events/presence" ->
+         || String.equal path "/events/presence"
+         || String.equal path "/api/v1/ide/cursors/stream" ->
       trim_opt (query_param request "token")
   | _ -> None
 
@@ -207,7 +208,7 @@ let verify_mcp_observer_stream_auth ~base_path request =
     | None ->
         Error
           "Authentication required. Use 'Authorization: Bearer <token>' header \
-           or 'token' query param for the observer/presence SSE stream."
+           or 'token' query param for the observer/presence/cursor SSE stream."
     | Some token -> (
         let* agent_name =
           resolve_agent_name_for_auth_raw ~base_path request ~token:(Some token)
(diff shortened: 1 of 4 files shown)


jeong-sik:
Current-head review for `d7f6519fc7`.

P0/P1/P2/P3 source blocker: not found in this pass.

What I checked:
- Query-token extraction is still method/path scoped: `GET /api/v1/ide/cursors/stream` only, alongside the existing observer `/mcp?sse_kind=observer` and `/events/presence` allowances.
- The cursor stream route now uses `with_observer_sse_read_auth`, so strict HTTP auth can accept the EventSource query token while still verifying the token through the existing workspace auth/permission path.
- The route remains read-tier (`CanReadState`) and does not broaden mutation auth or normal REST read auth.
- Regression coverage exercises strict auth with `MASC_HTTP_BASE_URL` set to a non-loopback URL and verifies a query-token cursor stream returns 200.
- `git diff --check origin/main...origin/codex/ide-cursor-stream-auth-followup-20260704` passes.

Local validation note:
- I attempted `scripts/dune-local.sh build test/test_server_ide_http.exe` from `.worktrees/ide-cursor-stream-degraded-20260704`, but the wrapper refused because bare Dune builds were already running on the machine. I did not bypass that lock. CI remains the build authority.

GitHub state at review time: `MERGEABLE`, draft, checks pending.

jeong-sik:
Post-merge closeout for current head `d7f6519fc7537371b73e85fc197e54bf24bbb278`.

P0: fixed. #23226 had added `token=` to the cursor EventSource URL, but the merged server route still went through generic read auth that does not read query tokens. #23229 routes `/api/v1/ide/cursors/stream` through a query-token-aware SSE read wrapper under strict HTTP auth.

P1: fixed with regression coverage. Added black-box `test_server_ide_http` coverage for `/api/v1/ide/cursors/stream?token=...` under public-base strict auth.

P2: remaining risk is limited to broader SSE auth taxonomy. This PR keeps the query-token path scoped to browser EventSource read streams; it does not spread query auth to mutation routes.

P3: no cosmetic follow-up needed from this pass.

Validation:
- `scripts/dune-local.sh build test/test_server_ide_http.exe`
- `./_build/default/test/test_server_ide_http.exe` -> 20 tests passed

Mergeability: merged. CI cancellation entries are post-merge closed-PR cleanup noise, not source blockers.

PR #23233, opened 2026-07-04
[codex] Classify IDE relation trace surfaces

jeong-sik:
## Summary
- classify comment-only activity traces as Comment from typed context
- classify runtime-only activity traces as Runtime from session/operation/worker ids
- keep existing PR/Board/Goal/Task/Git/Log surface precedence unchanged

## Validation
- pnpm --dir dashboard install --frozen-lockfile
- pnpm --dir dashboard exec vitest run src/components/ide/run-activity-trace-bridge.test.ts
- git diff --check

Dune/CI intentionally not run locally per workflow; CI can build OCaml.

### DIFF
--- a/dashboard/src/components/ide/run-activity-trace-bridge.test.ts
+++ b/dashboard/src/components/ide/run-activity-trace-bridge.test.ts
@@ -79,6 +79,54 @@ describe('bridgeRunActivityEventsToTrace', () => {
     expect(keeperTraceState.value.events[0]?.count).toBe(1)
   })
 
+  it('labels typed comment and runtime activity surfaces without kind fallback', () => {
+    const emitted = bridgeRunActivityEventsToTrace([
+      activity({
+        id: 'evt-comment',
+        kind: 'note',
+        context: {
+          file_path: 'lib/comment.ml',
+          line: 11,
+          comment_id: 'comment-11',
+        },
+      }),
+      activity({
+        id: 'evt-runtime',
+        kind: 'note',
+        context: {
+          file_path: 'lib/runtime.ml',
+          line: 12,
+          session_id: 'session-12',
+          operation_id: 'operation-12',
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Clean typed classification: activityTraceSurface now resolves comment_id to "Comment" and any of session_id/operation_id/worker_run_id to "Runtime", via field-presence checks (not string dispatch). The kind?.trim() || "Activity" fallback remains, which is intentional for genuinely unrecognised activity. Test asserts both typed labels ("without kind fallback") and pins the surface field, the id prefix, and the relevant context fields. Additive, no behavioral risk.

jeong-sik:
Current-head review (`44d3f35533`): no P0/P1/P2/P3 source blockers found.

The implementation keeps the existing surface precedence (`PR` -> `Board` -> `Goal` -> `Task` -> `Git` -> `Log`) and only adds the missing typed fallbacks after that: comment-only activity becomes `Comment`, and runtime session/operation/worker activity becomes `Runtime`. The new test covers both cases without relying on `kind` fallback.

Local focused validation from `.worktrees/ide-relation-trace-surfaces-20260705`:
- `pnpm --dir dashboard exec vitest run src/components/ide/run-activity-trace-bridge.test.ts --no-file-parallelism --maxWorkers=1` (1 file, 4 tests passed)
- `pnpm --dir dashboard lint`

Mergeability: still a draft PR, and GitHub checks are still pending on the fresh run (`Meta Guards`, `Lint`, `Dashboard`, plus some Fundamental checks). No current source blocker from this pass, but not merge-ready until draft is lifted and CI finishes green.

[근거] `gh pr view 23233 --json headRefOid,isDraft,mergeStateStatus,statusCheckRollup`, local commands above; checked 2026-07-05 00:50 KST; 신뢰도 High.

PR #23238, opened 2026-07-04
[codex] Surface IDE write-region observation failures

jeong-sik:
## Summary
- make write-region observation sinks return a typed result instead of silent unit
- make the default write-region sink report Write_region_sink_not_installed
- convert IDE region tracker storage exceptions into Write_region_sink_failed while preserving cancellation
- include ide_observation.write_region failure details in successful file-write JSON when region observation fails
- add regressions for no-sink emission and file-write failure surfacing

## Validation
- opam exec -- ocamlformat --check lib/agent_observation/agent_observation.mli lib/agent_observation/agent_observation.ml lib/ide/ide_bridge.ml lib/keeper/keeper_tool_filesystem_runtime.ml test/test_agent_observation_bridge.ml test/test_keeper_fs_edit_patch.ml
- git diff --check

Dune and CI were intentionally not run/waited locally per workflow.


jeong-sik:
Review-response follow-up pushed at `4cea6dc595`.

P1 fixes:
- Rewrote the write-region observation emission to use `try ... with` around `Agent_observation.emit_write_region_event`; the previous `| exception ...` arms were invalid in that match position and failed OCaml compilation.
- Removed the invalid `Masc.Agent_observation` test alias. `masc.agent_observation` is an unwrapped library, so the test should use the top-level `Agent_observation` module.

Shared blocker:
- The old Health/Lint/Canary logs also hit the shared `server_ide_http.ml:1115` parser failure. That remaining main-level fix is split into #23249; after #23249 lands, this branch may still need an update/rerun to remove the stale shared failure from the check history.

Validation:
- `git diff --check`
- `opam exec -- ocamlformat --check lib/keeper/keeper_tool_filesystem_runtime.ml test/test_keeper_fs_edit_patch.ml`
- `opam exec -- ocamlc -stop-after parsing lib/keeper/keeper_tool_filesystem_runtime.ml`
- `opam exec -- ocamlc -stop-after parsing test/test_keeper_fs_edit_patch.ml`
- Attempted `scripts/dune-local.sh build test/test_keeper_fs_edit_patch.exe`, but stopped while queued behind the shared `/tmp/me-dune-local.lock` held by another long-running worktree build. CI remains the build authority for the focused executable.

[근거] failed jobs `85147390426`, `85147390398`, `85147390434`, `85147390402`; local validation commands above; checked 2026-07-05 KST; 신뢰도 High.

jeong-sik:
Review notes after current-head check:

P1: current CI failure is stale-base noise, not this PR source diff. `ignore() justification (new sites)` is comparing the PR run against base `1b6df8e4f49cc39b624d360c5e81bc6c22538f03` and flags `lib/server/server_runtime_config_root_bootstrap.ml:174`, which is not touched by this PR. The branch should be updated/rebased onto current `main` before judging the remaining CI result.

P2: the implementation converts `Ide_region_tracker.ingest_tool_call` exceptions into `Write_region_sink_failed` at `lib/ide/ide_bridge.ml:1047`, but the added regression only covers `Write_region_sink_not_installed` through `Agent_observation.reset_for_testing()`. Since the PR summary explicitly promises storage-exception conversion, add a focused regression for the adapter failure path or narrow the claim. The production code direction otherwise looks sound: the write succeeds, cancellation is re-raised, and observation failure is surfaced as `ide_observation.write_region` instead of being silent.

jeong-sik (commented):
Adversarial review (approve — strong silent-failure removal).

write_region_sink is lifted from `unit` to `(unit, write_region_error) result` with a typed variant (Write_region_sink_not_installed | Write_region_sink_failed of string). emit_write_region_event now returns Result. In ide_bridge the sink wraps Ide_region_tracker in try/with, re-raising Eio.Cancel.Cancelled and converting other exceptions to Write_region_sink_failed. In keeper_tool_filesystem_runtime, track_write_region now returns `Some error | None` and the failure is surfaced in the tool result JSON as ide_observation.write_region.error — the keeper write itself still succeeds (write success is decoupled from observation success, which is correct). Cancelled is re-raised at every layer (Eio semantics preserved).

Tests cover both the no-sink error path (test_write_region_no_sink_returns_error) and the FS-edit surfacing path (test_write_file_surfaces_missing_ide_observation_sink). This is exactly the "no silent failure" direction the review bar asks for.

Minor: the FE side that consumes the new ide_observation.write_region.error field should be confirmed to render it (or at least not regress) — the OCaml side is sound.

jeong-sik:
P1 source fix pushed: `6a2cc83ec7` sanitizes write-region observation sink failures before they are returned in tool JSON. Raw sink exceptions still go to the IDE bridge stderr log, but the caller now gets the stable path-free `write_region sink failed` string instead of `Printexc.to_string` contents.\n\nValidation:\n- `git diff --check`\n- `ocamlformat --check lib/agent_observation/agent_observation.ml lib/agent_observation/agent_observation.mli lib/ide/ide_bridge.ml test/test_keeper_fs_edit_patch.ml`\n- `ocamlc -stop-after parsing ...` on the same touched OCaml files\n- focused wrapper build was attempted for `test/test_keeper_fs_edit_patch.exe test/test_agent_observation_bridge.exe`, but it currently stops on the unrelated current-main compile failure in `lib/server/server_runtime_bootstrap.ml` (`Option.bind` pipe order). That blocker is already isolated in #23254.\n\nCurrent status: P1 leak risk fixed on this PR; still CI/build-gated by #23254/main compile repair and this PR remains draft.

jeong-sik:
P0/P1/P2/P3 status after stale-base refresh

P0 fixed: updated codex/ide-write-region-result-20260705 onto current main after #23254 merged. The prior Health/Lint failure state is cleared on refreshed head ba1a3f4c7c; current failures=0.

P1 improvement: avoided duplicating the shared runtime bootstrap fix into this architecture PR; the branch now consumes the merged source of truth from main.

P2 remaining: full-lane CI is still running; current status is 46 passing checks and 3 pending checks.

P3 mergeability: GitHub reports MERGEABLE / BLOCKED. This remains not merge-ready while draft and while the pending full-lane checks are incomplete.

[근거] `gh pr update-branch 23238`; `gh pr checks 23238`; `gh pr list --repo jeong-sik/masc --state open --json statusCheckRollup,headRefOid,mergeable,mergeStateStatus`; checked 2026-07-05 02:14:31 KST; Trust=High.

PR #23246, opened 2026-07-04
Expose effective input modality flags in runtime summaries

jeong-sik:
## Summary
- include effective multimodal/image/audio/video flags in the runtime monitor compact effective capability summary
- include the same effective input modality flags in settings runtime catalog diagnostics
- extend dashboard fixtures/assertions so OAS effective capability input flags stay visible

## Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/runtime-monitor.test.ts src/components/settings-surface.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard run typecheck
- pnpm --dir dashboard run lint
- git diff --check

Issue #23243, opened 2026-07-04
main broken: server_runtime_bootstrap record disambiguation (Capability_manifest_env_var mis-inferred)


jeong-sik:
## Symptom

`dune build .` on origin/main fails:

```
File "lib/server/server_runtime_bootstrap.ml", line 149, characters 39-66:
Error: This variant expression is expected to have type model_catalog_env_source
       There is no constructor Capability_manifest_env_var within type model_catalog_env_source
File "test/test_server_runtime_bootstrap.ml", line 130, characters 4-46:
Error: ... Server_runtime_bootstrap.capability_manifest_env_source vs model_catalog_env_source
```

## Reproduce

`cd ~/me/workspace/yousleepwhen/masc && dune build .` (HEAD as of today).

## Suspected cause

`model_catalog_env_resolution` and `capability_manifest_env_resolution` both declare `{ path : string; source : <env_source> }`. At the construction site in `resolve_oas_capability_manifest_path`:

```ocaml
| Some path -> Some { path; source = Capability_manifest_env_var }
```

the compiler mis-disambiguates the record to `model_catalog_env_resolution` even though `Capability_manifest_env_var` belongs to `capability_manifest_env_source` and the function's return type is annotated `capability_manifest_env_resolution option`. Likely a regression from the recent capability-manifest work (#23240 preload explicit capability catalogs and neighbors).

## Fix candidates

- Add an explicit record type annotation at the construction site: `Some ({ path; source = Capability_manifest_env_var } : capability_manifest_env_resolution)`.
- Or qualify the field: `Some { Capability_manifest_env_resolution.path; source = Capability_manifest_env_var }`.

Both are workarounds; please also confirm why type-directed disambiguation stopped working here (recent type definition change / OCaml version interaction).

## Impact

main is not buildable. CI may be skipping this target or runner was stuck, masking the break. This is the third main-broken incident this session (after ide_bridge exit_semantics shadow and test_dashboard_http_core Client_registry_eio).

Discovered while building the activity-envelope refactor PR.

jeong-sik:
Update: the server_runtime_bootstrap disambiguation is fixed in PR fix/runtime-bootstrap-record-disambig (Draft). But `dune build` then surfaces a SECOND broken site:

```
File "lib/server/server_ide_http.ml", line 509, characters ...:
Error: Syntax error: ) expected  (unmatched ( from the /api/v1/ide/annotations handler, line 509 → 1115)
```

[...]

PR #23254, opened 2026-07-04
fix(runtime): correct Option.bind/filter pipe in catalog resolution (main broken)

jeong-sik:
## Summary

origin/main fails to build in `lib/server/server_runtime_bootstrap.ml`. Two distinct errors in the same `Option` pipe chain in `find_catalog_in_config_root` resolution:

1. `Option.filter` is **not in stdlib Option** (OCaml 5.x) → `Unbound value Option.filter`.
2. `Option.bind` argument order is `'a option -> ('a -> 'b option) -> 'b option`, incompatible with `|>` (the pipe puts the piped value into the first/opt slot), so `prev |> Option.bind f` mis-applies `f` as the option.

## Fix

Replace `Option.filter` with an inline match lambda and wrap `Option.bind` in a pipe lambda:
```ocaml
|> (fun opt -> match opt with Some v when not (String.equal v "") -> Some v | _ -> None)
|> (fun root -> Option.bind root find_catalog_in_config_root)
```

## Context

main has **multiple compounding build regressions** from the recent runtime-summary PRs (#23246 and neighbors). This PR addresses the `server_runtime_bootstrap` site. CI verification of the full build depends on the other broken sites being repaired (tracked in #23243). Local build was advancing through these specific errors before committing.

## Notes

- `Option.filter` does not exist in stdlib — the recent code assumed it did. Consider adding a project-local `Option.filter` helper if this pattern is used elsewhere, rather than inlining every time.
- `Option.bind` + `|>` is a recurring foot-gun; the pipe-lambda form is the safe idiom.

### DIFF
--- a/lib/server/server_runtime_bootstrap.ml
+++ b/lib/server/server_runtime_bootstrap.ml
@@ -159,8 +159,11 @@ let resolve_oas_model_catalog_path
        (match
           config_root
           |> Option.map String.trim
-          |> Option.filter (fun value -> not (String.equal value ""))
-          |> Option.bind find_catalog_in_config_root
+          |> (fun opt ->
+               match opt with
+               | Some value when not (String.equal value "") -> Some value
+               | _ -> None)
+          |> (fun root -> Option.bind root find_catalog_in_config_root)
         with
         | Some _ as found -> found
         | None ->


jeong-sik:
P0/P1/P2/P3 status after OCaml semantics verification

P0 fixed by this PR: the changed site no longer relies on non-existent stdlib Option.filter, and no longer pipes directly into Option.bind with the wrong argument position.

P1 validation: branch worktree is clean; the one-file diff is limited to lib/server/server_runtime_bootstrap.ml. Local checks passed:
- opam exec -- ocamlformat --check lib/server/server_runtime_bootstrap.ml
- opam exec -- ocamlc -stop-after parsing lib/server/server_runtime_bootstrap.ml
- scripts/dune-local.sh build lib/server/.masc_server.objs/byte/server_runtime_bootstrap.cmo

P2 remaining: GitHub required checks are still running. Current completed signal is good: Lint, Health, and OCaml 5.5 Build Canary are passing on c0bea20ff6.

P3 mergeability: this is still draft and GitHub is recalculating mergeability; not merge-ready until the remaining required checks finish and the draft state is intentionally cleared.

[근거] gh pr view 23254 --json headRefOid,baseRefOid,mergeStateStatus,mergeable,statusCheckRollup at 2026-07-05 02:00:20 KST: head=c0bea20ff6, base=f991633a96, failures=0, pass=46, pending=3. Trust=High.

jeong-sik:
Build verification update: after freeing disk (worktree _build cleanup), `dune build --root .` in this worktree (base origin/main @ 2b7de82dc6) exits 0. Output contains only warnings (dune cache store warning, ld duplicate -lzstd) — no errors. The server_runtime_bootstrap Option.filter/Option.bind pipe fix resolves the broken site; no subsequent broken site surfaces in this main version. Ready for review/merge.

PR #23259, opened 2026-07-04
dashboard: mark chat trace provenance

jeong-sik:
## Summary
- add trace-row provenance badges from existing thinking/tool/chat fields
- mark trace-only tool steps without a real tool_call_id as unlinked instead of pending forever
- add regression coverage for provenance and unjoinable tool traces

## Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts src/components/chat/primitives-interleave.test.ts src/keeper-stream.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check

### DIFF
--- a/dashboard/src/components/chat/primitives.test.ts
+++ b/dashboard/src/components/chat/primitives.test.ts
@@ -1922,6 +1922,42 @@ describe('ChatTranscript — tool-call grouping (turn timeline)', () => {
     expect(bundle?.querySelector('[data-chat-trace-step="chat"]')?.textContent).toContain('곧 답합니다')
   })
 
+  it('badges turn timeline rows with field-level provenance', () => {
+    render(
+      html`<${ChatTranscript}
+        entries=${[
+          entry({
+            id: 'a-provenance',
+            text: '답합니다',
+            role: 'assistant',
+            source: 'direct_assistant',
+            traceSteps: [
+              { kind: 'think', text: 'checking context', oasBlockIndex: 3 },
+              {
+                kind: 'tool',
+                name: 'keeper_board_list',
+                toolCallId: 'tc-prov',
+                status: 'ok',
+                args: '{"limit":1}',
+                oasBlockIndex: 4,
+              },
+            ],
+          }),
+        ]}
+        emptyText="empty"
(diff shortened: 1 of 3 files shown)


jeong-sik:
P0/P1/P2/P3 current-head status

P0: no blocker found on refreshed/current head 392b4831d6.
P1 improvement: added the missing area:dashboard label so this dashboard PR matches the rest of the dashboard queue.
P2: all current checks are complete: 43 passing, 0 failing, 0 pending.
P3: GitHub reports MERGEABLE / CLEAN, but this is still a draft PR, so not merge-ready until intentionally marked ready.

[근거] `gh pr edit 23259 --add-label area:dashboard`; `gh pr list --repo jeong-sik/masc --state open --json labels,isDraft,statusCheckRollup,headRefOid,mergeable,mergeStateStatus`; checked 2026-07-05 02:24:31 KST; Trust=High.

jeong-sik (commented):
Adversarial review (approve).

Two typed improvements to ChatTranscript trace rendering:
1. Field-level provenance badges: trace steps carry oasBlockIndex and are badged "OAS #N" (and the assistant reply gets a "reply" badge) via data-chat-trace-provenance.
2. Trace-only tool steps without a tool_call_id are now classified as "unlinked" (status class, "도구 호출 ID 없음"), not "pending" ("출력 대기 중"). This is a correct state distinction — a trace tool entry with no id can never be joined, so marking it pending was misleading.

Tests cover both the provenance badge case and the unlinked-trace-tool case. Korean UI strings are consistent with surrounding copy.

jeong-sik:
P3 current-head review (head 392b483). No source blocker found. The trace provenance badges are derived from existing trace/tool fields, and unlinked tool traces are marked as unlinked instead of pending so the UI does not imply a missing output is still expected. Tests cover OAS block provenance, reply provenance, and trace-only tools without tool_call_id. CI current head: CI Gate, Dashboard, Lint, Meta Guards pass; heavier OCaml/TLA jobs are skipped by changed-surface gating. Mergeability: MERGEABLE; still draft.

PR #23260, opened 2026-07-04
Reuse runtime catalog summaries in monitor

jeong-sik:
Summary
- Reuse shared runtime catalog summary helpers for runtime monitor compact provider lines.
- Extend snapshot summary SSOT with top-level protocol/tools/thinking/streaming/thinking-control facts.
- Keep runtime monitor structured detail rows intact for per-field inspection.

Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/runtime-monitor.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard run typecheck
- pnpm --dir dashboard run lint
- git diff --check

PR #23266, opened 2026-07-04
fix(ci): diff pull request scopes from merge base

jeong-sik:
## Summary
- compute PR changed-file scopes from merge-base(base, head) instead of origin/base..head
- avoid stale branches pulling unrelated main changes into CI surface detection
- keep an explicit deepen retry before failing merge-base resolution

## Evidence
- #23260 PR files: dashboard/src/components/runtime-monitor.test.ts, dashboard/src/components/runtime-monitor.ts, dashboard/src/lib/runtime-provider-summary.ts
- Existing CI run 28713892942 detected stale extra lib/test files and set build=true/tla=true
- Local simulation with merge-base returns only the 3 #23260 dashboard files

## Validation
- bash scripts/lint/yaml-syntax.sh
- local merge-base diff simulation for #23260 head

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,8 +195,14 @@ jobs:
 
           if [ "${GITHUB_EVENT_NAME}" = "pull_request" ]; then
             bash scripts/ci/git-fetch-retry.sh origin "${GITHUB_BASE_REF}" --depth=200
-            base_ref="origin/${GITHUB_BASE_REF}"
             head_ref="${GITHUB_HEAD_SHA}"
+            base_branch_ref="origin/${GITHUB_BASE_REF}"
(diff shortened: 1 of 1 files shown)


jeong-sik (commented):
Adversarial review (approve — CI correctness).

Right fix. Two-dot `origin/base..head` pulls in unrelated main changes when base drifts behind main (false-positive CI surface detection, ratchet miscounts, red-CI that is not the PR's fault). `merge-base(base, head)` diffs from the actual fork point, so the CI sees only the PR's real scope. The explicit deepen retry before failing merge-base resolution is the correct robustness touch (shallow clones can miss the merge base).

This directly supports the "base-drift 2-dot vs main" discipline and makes the ratchet/surface checks trustworthy. Small change, high leverage.

P3 nit: confirm the deepen retry depth is bounded (a `--deepen=64` then fail is better than an unbounded deepening loop on a pathological history).

PR #23207, opened 2026-07-04
fix(runtime): block degraded fallback for explicit routes

jeong-sik:
## Summary
- Follow-up to #23192: catalog-degraded boot no longer erases explicit runtime routing intent into default fallback.
- Degraded boot can still disable uncatalogued runtime definitions, but only when no default, keeper assignment, special route, media failover, or lane candidate references those disabled runtime ids.
- Split regression coverage into two cases: referenced uncatalogued runtimes are fatal, unreferenced uncatalogued runtime definitions are disabled with operator-visible degradation.

## Validation
- `ocamlformat --check lib/runtime/runtime.ml lib/runtime/runtime.mli test/test_runtime_config_validity.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_runtime_config_validity.exe` was attempted, but another repo-local `scripts/dune-local.sh` process held `/tmp/me-dune-local.lock`; I stopped only my waiting process and left the lock holder untouched.

## Review Notes
- This intentionally favors fail-closed routing over keeping the server alive by silently changing model/runtime intent.
- Full Dune build remains CI authority.

PR #23268, opened 2026-07-04
test(runtime): align degraded init diagnostics

jeong-sik:
## Summary
- update runtime TOML degraded-init tests to assert the routes actually reported by fail-closed diagnostics
- keep the default-fallback guard assertion aligned with the current diagnostic text

## Context
#23260 unnecessarily ran Build/Test because its branch was stale; that exposed this existing runtime TOML gate drift after dune build itself succeeded.

## Validation
- git diff --check
- local Dune intentionally not run; CI is the build authority for this change


jeong-sik (commented):
Adversarial review (approve).

Test-only alignment: the runtime degraded-init tests now assert the routes actually reported by the fail-closed diagnostics, and the default-fallback guard assertion is re-aligned to the current diagnostic text. This is the right fix — the tests were asserting stale route names that no longer match #23207's fail-closed diagnostic output.

Pairs with #23266 (CI merge-base): the stale branch in #23260 caused a false Build/Test run that surfaced this diagnostic mismatch. Small, correct, test-only.

PR #23263, opened 2026-07-04
Show runtime catalog spec in config probe

jeong-sik:
Summary
- Load the shared runtime catalog in the config resolution runtime probe panel.
- Show matching Provider x Model catalog summary beside provider reachability rows.
- Surface explicit catalog loading/error/missing status instead of silently omitting missing specs.

Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/tools/config-resolution-panel.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard run typecheck
- pnpm --dir dashboard run lint
- git diff --check

PR #23255, opened 2026-07-04
Expose effective runtime wire in keeper rail

jeong-sik:
## Summary
- include effective capability source in the keeper workspace runtime rail summary
- surface effective input modality flags and thinking-control wire in the same compact rail summary
- preserve reasoning streaming field detail instead of showing only the streaming kind

## Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts --no-file-parallelism --maxWorkers=1
- pnpm --dir dashboard run typecheck
- pnpm --dir dashboard run lint
- git diff --check

Dune was not run locally; this is a dashboard-only slice and CI remains the build authority.

PR #23270, opened 2026-07-04
Show declared runtime spec in model editor

jeong-sik:
Summary
- Add the declared Provider x Model x binding spec row to the keeper runtime model editor catalog card.
- Reuse the shared runtime provider summary formatter rather than duplicating field formatting.
- Extend the model editor fixture to cover provider behavior, model capabilities, match prefixes, and binding concurrency.

Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-runtime-model-editor.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check

### DIFF
--- a/dashboard/src/components/keeper-runtime-model-editor.test.ts
+++ b/dashboard/src/components/keeper-runtime-model-editor.test.ts
@@ -188,6 +188,78 @@ function makeRuntimeProvider(runtimeId: string, providerName: string, modelName:
       seed: 7,
       connect_timeout_s: 30,
     },
+    declared_spec: {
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds the declared Provider x Model x binding spec row to the keeper-runtime-model-editor catalog card by reusing the shared runtime-provider-summary formatter (no duplicated field formatting). Fixture extended to cover the new row. Additive typed; consistent with the #23255-#23263 caps series. No behavioral risk.

PR #23271, opened 2026-07-04
Show runtime catalog spec in keeper detail

jeong-sik:
## Summary
- show the `/api/v1/providers` runtime catalog spec inside the keeper detail runtime lens
- resolve the catalog entry from the observed `config_drift` runtime id only: live runtime first, then default runtime
- render explicit loading/error/missing/no-runtime-id states instead of silently omitting catalog evidence

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-detail-runtime.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

### DIFF
--- a/dashboard/src/components/keeper-detail-runtime.test.ts
+++ b/dashboard/src/components/keeper-detail-runtime.test.ts
@@ -1,10 +1,24 @@
 import { h } from 'preact'
-import { cleanup, fireEvent, render, screen } from '@testing-library/preact'
-import { afterEach, describe, expect, it } from 'vitest'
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Continues the runtime catalog spec caps series on the keeper-detail surface. Additive typed (declared_spec + capabilities rendering), reuses the runtime-provider-summary helper family established across #23255-#23270. Same merge-order note: the shared helper must be on main (or land with whichever of the series merges first). No behavioral risk; series-consistent.

PR #23272, opened 2026-07-04
feat(dashboard): show binding runtime catalog spec

jeong-sik:
## Summary

- show the shared runtime catalog snapshot under each runtime.toml binding row
- reuse existing runtime provider summary helpers for snapshot/effective/request/declared/policy facts
- load the catalog only when the structured editor is on the bindings section

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/runtime-environment-editor.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

### DIFF
--- a/dashboard/src/components/runtime-environment-editor.test.ts
+++ b/dashboard/src/components/runtime-environment-editor.test.ts
@@ -1,10 +1,43 @@
 import { html } from 'htm/preact'
 import { render } from 'preact'
-import { fireEvent } from '@testing-library/preact'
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve).

Continues the runtime catalog spec caps series on the binding surface (4 files, +246/-6). Additive typed; reuses the runtime-provider-summary helper family established across #23255-#23271. Same merge-order note: shared helper must be on main (or land with whichever of the series merges first). Series-consistent, no behavioral risk.

PR #23273, opened 2026-07-05
[codex] Polish board surface interactions

jeong-sik:
## What changed

- Reworked board feed controls so sort/category behavior is explicit and server-backed, including discussed/comment-heavy ordering.
- Removed title/body keyword heuristics and default hardcoded contributor-quality priors from board presentation.
- Added richer board content rendering for embedded HTML/media and standalone video/audio/Youtube/Vimeo content.
- Added permalink, trackback, X share, and board context-inference actions for posts.
- Wired board emoji reactions into Keeper wake/context flow so reactions after Keeper activity are observable and actionable.
- Added a board context inference HTTP route that queues a real `masc_keeper_msg` turn with board post/comment surface context instead of faking an LLM action in the UI.

## Why

The board surface had several affordances that looked interactive but did not feed back into Keeper behavior or actual collaboration workflows. This change keeps the board weakly coupled while making reactions, sharing, rich content, and Keeper-launched context inference real production paths.

## Impact

- Dashboard users get usable blog-style post actions and richer idea/media rendering.
- Keeper-authored posts no longer collapse all sigils to generic `KE` when better identity data is available.
- Context inference launches an async Keeper turn through the existing `masc_keeper_msg` queue with explicit registered-Keeper resolution.
- Board reaction changes can wake Keepers through typed runtime events instead of remaining dashboard-only state.

## Validation

- `ocamlformat --check lib/server/server_routes_http_routes_activity.ml`
- `pnpm --dir dashboard exec vitest run src/api/board.test.ts src/components/board/post-detail.test.ts src/components/board/board-surface.test.ts`
- `git diff --check`

Local Dune build was intentionally not run; this repo uses CI for Dune build validation.

### DIFF
--- a/dashboard/src/api/board.test.ts
+++ b/dashboard/src/api/board.test.ts
@@ -10,7 +10,9 @@ import {
   fetchBoardKarmaLedger,
   fetchBoardPost,
   fetchBoardReactions,
+  normalizeBoardContextInferenceSubmission,
   normalizeBoardKarmaLedger,
+  requestBoardContextInference,
   sanitizeBoardTitle,
   toggleReaction,
   voteComment,
@@ -1237,5 +1239,84 @@ describe('board reactions', () => {
       last_ok: true,
       sample_count: 1,
     })
+  })
+})
+
+describe('board context inference', () => {
+  it('normalizes queued keeper submissions', () => {
+    expect(normalizeBoardContextInferenceSubmission({
+      ok: true,
+      request_id: 'kmsg-1',
+      keeper_name: 'luna',
+      post_id: 'post-1',
+      status: 'queued',
+      target_source: 'explicit_target',
+      message: 'queued',
+    })).toEqual({
+      ok: true,
+      requestId: 'kmsg-1',
+      keeperName: 'luna',
+      postId: 'post-1',
+      status: 'queued',
+      targetSource: 'explicit_target',
+      message: 'queued',
+    })
+  })
+
+  it('rejects malformed context inference submissions', () => {
+    expect(normalizeBoardContextInferenceSubmission({ ok: true, keeper_name: 'luna' })).toBeNull()
+    expect(normalizeBoardContextInferenceSubmission({ ok: false })).toBeNull()
+  })
+
+  it('requests board context inference with an explicit target keeper', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      new Response(JSON.stringify({
+        ok: true,
+        request_id: 'kmsg-ctx',
+        keeper_name: 'luna',
+        post_id: 'post-1',
+        status: 'queued',
+        target_source: 'explicit_target',
+      }), {
+        status: 202,
+        headers: { 'Content-Type': 'application/json' },
+      }),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    const result = await requestBoardContextInference(' post-1 ', ' luna ')
+
+    expect(result).toMatchObject({
+      requestId: 'kmsg-ctx',
+      keeperName: 'luna',
+      postId: 'post-1',
+      status: 'queued',
+      targetSource: 'explicit_target',
+    })
+    const [url, init] = fetchMock.mock.calls[0] as [string, RequestInit]
+    expect(url).toBe('/api/v1/board/context-inference')
+    expect(JSON.parse(String(init.body))).toEqual({
+      post_id: 'post-1',
+      target_keeper: 'luna',
+    })
+  })
+
+  it('fails when the server returns a malformed context inference response', async () => {
+    const fetchMock = vi.fn().mockResolvedValue(
+      new Response(JSON.stringify({
+        ok: true,
+        keeper_name: 'luna',
+        post_id: 'post-1',
+        status: 'queued',
+      }), {
+        status: 202,
+        headers: { 'Content-Type': 'application/json' },
+      }),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    await expect(requestBoardContextInference('post-1')).rejects.toThrow(
+      'Malformed board context inference response',
+    )
   })
 })
--- a/dashboard/src/api/board.ts
+++ b/dashboard/src/api/board.ts
@@ -28,6 +28,18 @@ export interface BoardFlair {
   label: string
 }
 
+export type BoardContextInferenceTargetSource = 'explicit_target' | 'post_author'
+
+export interface BoardContextInferenceSubmission {
+  ok: true
+  requestId: string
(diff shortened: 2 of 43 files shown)


jeong-sik (commented):
P1: The context-inference action is rendered for every board post, but non-keeper-authored posts have no usable target and deterministically fail at runtime. `PostShareActions` only passes a target when `post.author_identity.kind === "keeper"`; otherwise it calls `requestBoardContextInference(post.id, undefined)`. The server fallback then tries the post author and returns `Bad_request` when that author is not a registered keeper (`target_keeper is required because board post author ... is not a registered keeper`). That makes a first-class visible action fail for normal human/operator posts, which are the posts most likely to need a keeper to infer context. Please either provide an explicit target-selection/default-keeper path for non-keeper posts, or hide/disable the action with an explicit reason, and add regression coverage for the non-keeper post path.

References: `dashboard/src/components/board/post-share-actions.ts:15-18`, `dashboard/src/components/board/post-share-actions.ts:40-46`, `lib/server/server_routes_http_routes_activity.ml:278-289`.

P2: The new OCaml HTTP route has no server-side tests for target resolution, malformed keeper-tool data, or the non-keeper author fallback. The TS tests cover client normalization only, so the production path that dispatches `masc_keeper_msg` through `/api/v1/board/context-inference` is currently guarded mostly by compile/type checks. Please add focused OCaml route/helper tests before taking this out of draft.

Review state note: GitHub will not let me submit `CHANGES_REQUESTED` on this PR because it is authored by the same account, but I would treat the P1 as blocking.

jeong-sik (commented):
Adversarial multi-axis review (5-axis workflow).

## P0 — none
The title/body keyword heuristics are genuinely removed (ARTICLE/REVIEW/NOTICE_SIGNALS + matchesAny deleted; body-length bands deleted). contentCategoryFromValue is now a typed closed switch (article/review/notice/system) with default → null (OOV rejected, pinned by board-surface.test "ignores non-canonical category metadata tokens"). No identity-based quality prior is introduced. Good direction.

## P1 — sentinel magic number + SSOT violation (the real concern)
contributorQualityHasEvidence (dashboard/src/lib/board-utils.ts) re-hardcodes the backend default priors as sentinel values:
- accountability_score < 1
- thompson_confidence !== 0.5
- autonomy_level !== 'standard'

These mirror lib/reputation.ml defaults ([@default 1.0], [@default 0.5], [@default "standard"]). Two problems:
1. **SSOT violation**: the frontend now load-bearingly knows the backend default values. If lib/reputation.ml defaults drift (e.g. accountability starts at 0.5), the frontend silently misclassifies default-prior agents as "has evidence" and re-renders the misleading numeric score — the exact bug this PR is trying to fix. The test name itself says "hides default priors", not "removes".
2. **Magic Number 금지**: 1, 0.5, 'standard' are bare literals.

Root fix: have the backend expose an explicit typed flag (e.g. reputation.evidence_state = Default_prior | Measured) rather than asking the frontend to reverse-engineer priors from value sentinels. Short of that, extract the sentinels as named constants and add a cross-repo drift test that fails when lib/reputation.ml defaults change without the frontend constants updating.

## P1 — residual silent default
contentCategory still returns `automation -> 'notice' : 'article'` when no meta is present (board-state.ts). Acceptable, but the PR description should state this residual default explicitly rather than implying classification is purely metadata-driven.

## P1 — boundary silent reject
A real measured accountability_score === 1.0 (or thompson_confidence === 0.5 posterior) is indistinguishable from a default prior and will be hidden. Low probability but a silent failure mode at the boundary; the explicit backend flag (above) resolves it.

## Test
Negative tests pin the heuristic removal (good). The boundary at exactly 1.0 for the "real evidence" path is untested — add it, or make the backend flag the source of truth.

Labels: area:dashboard, root-cause:SSOT (sentinel re-hardcoding), root-cause:SIL (boundary silent reject).

jeong-sik:
Rechecked current head `3ba122309d144a3dc3b40e0b6ae22dc407d41149`.

The previous P1 is still unresolved. The new commit only extracts the OCaml context-inference handler; the dashboard still renders the context-inference action for every post while `contextInferenceTargetKeeper()` returns `undefined` for non-keeper authors (`dashboard/src/components/board/post-share-actions.ts:15-18`, `:40-46`). The server still rejects that no-target case when the board post author is not a registered keeper (`lib/server/server_routes_http_routes_activity.ml:265-293`). Existing tests still cover only the keeper-author happy path (`dashboard/src/components/board/post-detail.test.ts:638-672`, `dashboard/src/components/board/board-surface.test.ts:673-691`), so normal human/operator posts still get a visible action that deterministically fails.

This still needs either an explicit target-selection/default-keeper path or a disabled/hidden action with a clear reason, plus a regression for the non-keeper author case.

jeong-sik (commented):
**Level**: Comment (Adversarial Pass)

- **Verification**: Board reactions integration into the keeper world observation and event loop.
- **Adversarial Warning (P3)**: `/api/v1/board/context-inference` executes `masc_keeper_msg` tool call to target keeper. If the target keeper is offline/slow, the HTTP request fiber will park/block. Although Eio handles fiber scheduling cooperatively, wrapping it in a timeout (e.g. `Eio.Time.with_timeout`) is recommended as future work to prevent HTTP fiber lease pool exhaustion.
- **Safety**: Waking up relevant keepers on reaction uses a well-thought-out logic: `Reaction_after_self_activity` only triggers if the keeper authored the post or commented on the thread. This is a very targeted wakeup logic that avoids "wake storms" (meaning it does not wake every keeper, avoiding Keeper constraint/CPU overload - P0 guard!).

jeong-sik:
Rechecked current head `88cffa3f6cb1238d79b52ff952f7e0ab6253c4a8`. The previous P1 is still not resolved; it changed from a deterministic 400 into an arbitrary keeper wake-up.

P1: non-keeper authored posts now use `keepers.value[0]` as the context-inference target (`dashboard/src/components/board/post-share-actions.ts:16-23`), and the click path sends that as `target_keeper` (`dashboard/src/components/board/post-share-actions.ts:44-50`). The new regression explicitly locks in “first keeper as fallback” (`dashboard/src/components/board/board-surface.test.ts:1185-1217`). That is a store-order heuristic, not a domain decision. In a multi-keeper workspace, an operator-authored board post can now wake whichever keeper happens to be first in the dashboard list, even when the related keeper is another mentioned/assigned keeper or when no deterministic target exists.

The server-side explicit-target support is useful, but the UI still needs an explicit target-selection/default-keeper policy backed by a real source of truth, or the action should stay disabled/hidden with a reason when the author is not a registered keeper. Please remove the `keepers[0]` fallback and add a regression for the multi-keeper non-author case.

PR #23274, opened 2026-07-05
dashboard: gate keeper config writes by source

jeong-sik:
## Summary
- Gate Keeper Config runtime mutation controls on the existing source truth: `sources.default_source_kind === "toml"` plus a non-empty `default_manifest_path`.
- Downgrade runtime policy, sandbox/access, max context, and goal assignment controls to read-only when the keeper is persona/no-manifest backed.
- Keep separately-backed tool policy editing available through `set_policy`, and remove the stale fixed "70+ goals" catalog claim.

## Verification
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-config-panel.test.ts --no-file-parallelism --maxWorkers=1`
- Rendered smoke: `node /private/tmp/keeper-config-write-capability-smoke.js` against `http://127.0.0.1:5187/dashboard/#/keepers?keeper=analyst`; captured `/private/tmp/keeper-config-write-capability-modal.png`.

## Notes
- Dune build intentionally not run locally; dashboard-only TypeScript/frontend change and CI remains the build authority for repo-wide checks.

### DIFF
--- a/dashboard/src/components/keeper-config-panel.test.ts
+++ b/dashboard/src/components/keeper-config-panel.test.ts
@@ -1633,6 +1633,60 @@ describe('KeeperConfigPanel', () => {
     ) as HTMLButtonElement
     expect(saveButton.disabled).toBe(true)
   })
+
+  it('resets runtime draft when switching from keeper A to keeper B to prevent stale settings leakage', async () => {
+    const configA = makeKeeperConfig({
+      name: 'keeper-a',
+      execution: {
+        selected_runtime_id: 'ollama_cloud.deepseek-v4-flash',
+        runtime_options: ['ollama_cloud.deepseek-v4-flash', 'ollama_cloud.qwen-2.5-coder'],
+        selected_runtime_canonical: 'ollama_cloud.deepseek-v4-flash',
+        models: ['ollama_cloud.deepseek-v4-flash'],
+      } as any,
+    })
+    const configB = makeKeeperConfig({
+      name: 'keeper-b',
+      execution: {
+        selected_runtime_id: 'ollama_cloud.qwen-2.5-coder',
+        runtime_options: ['ollama_cloud.deepseek-v4-flash', 'ollama_cloud.qwen-2.5-coder'],
+        selected_runtime_canonical: 'ollama_cloud.qwen-2.5-coder',
+        models: ['ollama_cloud.qwen-2.5-coder'],
+      } as any,
+    })
+
+    mocks.fetchKeeperConfig.mockResolvedValueOnce(configA)
+    render(html`<${KeeperConfigPanel} keeperName="keeper-a" />`, container)
+    await flush()
+    await flush()
+
+    selectKcfTab(container, '런타임')
+    await flush()
+
+    const select = container.querySelector('select[aria-label="runtime_id"]') as HTMLSelectElement
+    expect(select.value).toBe('ollama_cloud.deepseek-v4-flash')
+
+    select.value = 'ollama_cloud.qwen-2.5-coder'
+    select.dispatchEvent(new Event('change', { bubbles: true }))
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
Adversarial review (approve).

Gates keeper-config runtime-mutation controls on the typed source truth: sources.default_source_kind === "toml" plus a non-empty default_manifest_path. Downgrades runtime policy / sandbox-access / max-context / goal-assignment controls to read-only when the keeper is persona or non-toml-sourced. This is the right pattern — the gate is driven by a typed source kind + manifest presence, not by a heuristic or hardcoded identity list.

P3 nit: the gate predicate combines two fields (source kind AND manifest path). If this predicate is inlined in multiple JSX branches, extract a single helper (e.g. canMutateRuntimeConfig(sources)) so the two-field contract is named and tested in one place. If it is already a single helper, ignore.

jeong-sik:
Source review on `93a91d65bddb5f3ab2f07aadb2df3c3b06290ecc`. I would block this before taking the PR out of draft.

P1: the source gate still relies on a module-global `runtimeDraft` that is not reset when the same config panel instance is reused for a different keeper. `loadKeeperConfig()` resets only the async resource on name/force changes and does not notify the panel reset handlers (`dashboard/src/components/keeper-config-state.ts:38-48`); `KeeperConfigPanel` then initializes `runtimeDraft` only when it is null (`dashboard/src/components/keeper-config-panel.ts:1147-1154`). The overlay is rendered without a `key` and can receive a new `keeperName` prop from `configOverlayKeeper` (`dashboard/src/components/keeper-detail-page.ts:561-595`), so a user can open/switch keeper config without remounting the panel. In that state, `runtimeCanEdit` is computed from the newly loaded `c`, but the editable/read-only rows and `buildRuntimePayload(rd, c)` can still use the previous keeper's `rd`. For a TOML-backed target, the footer can immediately post a stale payload to the new keeper; for a persona target, the read-only goal/mention display can show the previous keeper's draft. Please key/reset the runtime draft by keeper/config generation on name changes, and add a regression that rerenders the panel from keeper A to keeper B and proves stale runtime changes cannot be saved or displayed.

jeong-sik (commented):
**Level**: Comment (Adversarial Pass)

- **Verification**: Evaluated config panel write logic. The PR ensures non-TOML configurations fail closed/read-only rather than crashing on manifest-write.
- **Safety**: No hardcoded paths, local user references, or environmental overuse found in production code. All tests built and passed successfully on CI.

jeong-sik:
Current-head follow-up (2026-07-05 16:58 KST): I patched the Dashboard CI failure in ca8a4a3a1d27.

What changed:
- Updated dashboard/src/styles/keeper-workspace-mobile.test.ts to assert the new LazyKeeperConfigPanel remount contract with key=${keeperName}.

Local validation:
- pnpm vitest run --config vitest.config.ts src/components/keeper-config-panel.test.ts --no-file-parallelism --maxWorkers=1 -> 63 passed
- pnpm run typecheck -> passed
- pnpm vitest run --config vitest.config.ts src/styles/keeper-workspace-mobile.test.ts --no-file-parallelism --maxWorkers=1 -> 18 passed
- git diff --check -> passed

Current blocker: PR is still draft and GitHub Dashboard/Lint/Meta checks are rerunning on ca8a4a3a1d27. No ready label.

jeong-sik:
Rechecked current head `ca8a4a3a1d278238844157985d586bad10d7e420`. The previous P1 about stale `runtimeDraft` leaking across keeper switches appears resolved: `loadKeeperConfig` now resets panel drafts on keeper-name changes, the overlay remounts with `key=${keeperName}`, and the regression covers switching from keeper A to keeper B before saving. I do not see a new blocker in the touched config-write path.

PR #23277, opened 2026-07-05
[codex] Fix activity context inference error type

jeong-sik:
## Summary
- Align board context inference target errors with the public interface shape.
- Restore the missing `test_relation_materializer` entry in the large test stanza module list.
- Keep dispatch errors unchanged; only the resolver now returns payload variants as declared in the mli and expected by tests.

## Root Cause
- `resolve_board_context_inference_target` returned `(status, message)` tuples while `server_routes_http_routes_activity.mli` declares `` `Bad_request of string`` / `` `Internal_server_error of string``. Current `main` CI failed during OCaml interface checking on that mismatch.
- After that was fixed, CI exposed a pre-existing Dune stanza drift: `test_relation_materializer` was listed as an executable but omitted from the same stanza `modules` field.

## Validation
- `ocamlformat --check lib/server/server_routes_http_routes_activity.ml lib/server/server_routes_http_routes_activity.mli test/test_board_context_inference_resolution.ml`
- `git diff --check`

Local Dune build intentionally not run; CI is the Dune authority for this repo.

PR #23280, opened 2026-07-05
fix: restore activity route build on main

jeong-sik:
## Summary
- Fix `resolve_board_context_inference_target` to return the payload-bearing error variants declared by the mli.
- Update `test/dune` so `test_relation_materializer` is listed in both the test names and modules list.

## Evidence
- Fixes main build failure tracked in #23278.
- Failure source: push CI run 28734195802 / job 85205412639 at current main `2b472b1ab3ac18cdc45eeb0de213f39f8db6b7fc`.

## Validation
- `ocamlformat --check lib/server/server_routes_http_routes_activity.ml`
- `scripts/dune-local.sh build lib/server/.masc_server.objs/byte/server_routes_http_routes_activity.cmo test/test_board_context_inference_resolution.exe`
- `_build/default/test/test_board_context_inference_resolution.exe`

PR #23281, opened 2026-07-05
[codex] Budget lane no-progress retries

jeong-sik:
## Summary
- Gate lane-internal accept no-progress fallback with the existing degraded retry slot budget.
- Leave network/HTTP retry behavior unchanged.
- Add focused tests proving no-progress gate denial blocks fallback while network retry bypasses the gate.

## Root Cause
The keeper runtime path already has a budget-aware degraded retry loop, but `run_named` could rotate lane candidates internally before that budget layer saw an accept-contract no-progress error. With thinking/tool loops and long checkpoints, that can replay large contexts across providers and amplify token use.

## Runtime Evidence
- Live mad-improver trace `trace-1782437332896-00000` recorded 5,611 API calls, 318,152,316 input tokens, and repeated `no_usable_progress` / `max_tokens` empty responses.
- This is provider-agnostic: the failing runtime evidence involved `runpod_mtp.qwen36-35b-a3b-mtp`, while the original billing concern was observed through MiMo usage. The fix keys on typed no-progress response shape, not provider names.

## Validation
- `ocamlformat --check lib/keeper/keeper_turn_driver.ml lib/keeper/keeper_turn_driver.mli test/test_keeper_turn_driver_failover.ml`
- `git diff --check codex/fix-activity-target-error..HEAD`

Local Dune build intentionally not run; CI is the Dune authority for this repo.

Stacked on #23277. Retarget to main after #23277 lands.

### DIFF
--- a/lib/keeper/keeper_turn_driver.ml
+++ b/lib/keeper/keeper_turn_driver.ml
@@ -123,23 +123,34 @@ let runtime_failed_decision ~idx ~runtime_id error =
       ("error_kind", `String (Oas_compat.error_kind error));
     ]
 
-let lane_retry_checkpoint ~is_last ~resume_checkpoint ~checkpoint_after error =
+let lane_retry_checkpoint
+    ~is_last
+    ~allow_accept_no_progress_retry
+    ~resume_checkpoint
+    ~checkpoint_after
+    error =
   if is_last then
     None
   else if Keeper_turn_driver_try_runtime.accept_no_progress_should_try_next error
   then
-    Some
-      (Keeper_turn_driver_try_runtime.checkpoint_for_accept_rejected_retry
-         ~resume_checkpoint
-         ~checkpoint_after
-         error)
+    if allow_accept_no_progress_retry
+    then
+      Some
+        (Keeper_turn_driver_try_runtime.checkpoint_for_accept_rejected_retry
+           ~resume_checkpoint
+           ~checkpoint_after
+           error)
+    else None
   else
     match Keeper_turn_driver_try_runtime.sdk_error_to_http_error error with
     | Some http_err when Runtime_attempt_fsm.should_try_next http_err ->
       Some checkpoint_after
     | _ -> None
 
-let attempt_runtime_candidates ~runtime_id ~runtime_id_of
+let attempt_runtime_candidates
+    ?(allow_accept_no_progress_retry = fun ~runtime_id:_ ~attempt:_ _error ->
+      true)
+    ~runtime_id ~runtime_id_of
     ~(emit_runtime_manifest :
        ?status:string ->
        ?decision:Yojson.Safe.t ->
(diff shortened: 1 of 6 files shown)


jeong-sik (commented):
Adversarial review (approve).

Gates the lane-internal accept no-progress fallback with the EXISTING degraded retry slot budget — not a new cap/cooldown, but reuse of the already-established retry budget. Network/HTTP retry behavior is unchanged. Tests prove the no-progress gate engages. This is the right shape: it avoids introducing a fresh "Cap/Cooldown" symptom-suppression (it composes with the existing budget), and it bounds the fallback.

P3 nit: confirm what happens when the degraded retry budget is exhausted — does the no-progress fallback get blocked (fail-closed), or does another path fire? State that terminal behavior explicitly so it is fail-closed by design, not by accident.

jeong-sik:
P2 Change Requested on current head `6d8a6aaea5c4`.

The production change is small and provider-agnostic: the lane fallback gate only applies to typed accept no-progress failures, and focused local validation passed:
- `git diff --check refs/remotes/origin/pr/23277...HEAD`
- `scripts/dune-local.sh build test/test_keeper_turn_driver_failover.exe`
- `scripts/dune-local.sh exec ./test/test_keeper_turn_driver_failover.exe` -> 15/15 tests

Required fix:
- `test/test_keeper_turn_driver_failover.ml:678-714` stores `Agent_sdk.Error.to_string error` and asserts that it contains `"no_usable_progress"`. That makes the regression depend on free-form rendering of the SDK error instead of the typed gate this PR is meant to protect. Please assert through `Driver.For_testing.accept_no_progress_should_try_next error` or the typed internal-error classifier instead, so the test keeps the same no-string-match boundary as production.

Integration note:
- This PR is still stacked on `codex/fix-activity-target-error` / `#23277`. If `#23280` becomes the canonical main build repair, retarget/rebase this stack before review-ready.

No P0/P1 logic blocker found in the runtime gate itself.

jeong-sik:
Fixed the P2 test-string issue in `6bb05a8f0e35`.

Change made:
- `test/test_keeper_turn_driver_failover.ml` now records `Driver.For_testing.accept_no_progress_should_try_next error` from the gate callback and asserts that typed predicate, instead of inspecting `Agent_sdk.Error.to_string` for `"no_usable_progress"`.

Validation on the pushed commit:
- `ocamlformat --check test/test_keeper_turn_driver_failover.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_keeper_turn_driver_failover.exe`
- `scripts/dune-local.sh exec ./test/test_keeper_turn_driver_failover.exe` -> 15/15 tests

Remaining status: `P3` integration only. The PR is still draft and stacked on `#23277`; retarget/rebase after the activity-route repair path settles (`#23280` looks like the canonical main fix if its pending Build/Test completes green).

PR #23286, opened 2026-07-05
[codex] Preserve Keeper Chat transcript provenance

jeong-sik:
## Summary
- Preserve REST chat-history `source` and `surface` fields when hydrating visible Keeper Chat entries instead of re-inferring them away.
- Add transcript entry provenance attributes (`data-chat-source`, `data-chat-surface-kind`, `data-chat-turn-ref`, role, delivery) so review/debug tooling can prove where a rendered row came from.
- Keep connector/dashboard surface chips backed by the persisted `surface` object; no synthetic channel labels are introduced.

## Validation
- `pnpm --dir dashboard install --frozen-lockfile`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-state.test.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1` (173 tests)
- `pnpm --dir dashboard run typecheck`
- `pnpm --dir dashboard run lint`
- `git diff --check`
- `pnpm --dir dashboard run build`
- Chrome DevTools Protocol smoke against `http://127.0.0.1:5176/dashboard/keepers?keeper=verifier` with live backend `http://127.0.0.1:8935`:
  - desktop 1440x1000 and mobile 390x844 loaded `MASC · Keepers`
  - 100 chat entries rendered, 86 carried `data-chat-surface-kind="dashboard"`
  - first entries exposed `source`, `delivery`, `surface`, and `turnRef` attributes from persisted history
  - metadata toggle interaction succeeded
  - no console/page errors, no framework overlay
  - screenshots: `/tmp/chat-transcript-provenance-cdp-desktop.png`, `/tmp/chat-transcript-provenance-cdp-mobile.png`

## CI Notes
- Current head: `ee1c6bed6abf3d93fc11cc44bebd812630c8dacd`.
- Initial GitHub checks are green: `PR Axis Cross-Check / check`, `Test Coverage Check / check`.

## Stack
- Base: #23277 / `codex/fix-activity-target-error` to avoid the known #23283 source-side Lint blocker while it is still pending on `main`.

PR #23287, opened 2026-07-05
[codex] Expose Keeper Chat trace provenance

jeong-sik:
## Summary

- expose machine-checkable provenance attrs on Keeper Chat trace rows (`think`, `reason`, `tool`, `chat`) instead of relying only on visible badge text
- preserve standalone tool-call transcript provenance with entry/source/delivery/turn/tool_call_id attrs
- stop synthesizing a fake `KeeperConversationEntry` for trace-only tool steps; trace-only rows render from the trace step and omit entry id/source unless a real sibling tool row exists

## Stack

- Base: `codex/chat-transcript-provenance-20260705` / #23286
- Head: `codex/chat-trace-provenance-20260705`

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts src/components/chat/primitives-interleave.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`
- Rendered smoke via Chrome CDP against `http://127.0.0.1:5177/dashboard/keepers?keeper=verifier`
  - desktop 1440x1000 and mobile 390x844
  - page title `MASC · Keepers`, no framework overlay, no console errors
  - metadata toggle found/clicked
  - live verifier transcript rendered 100 `[data-chat-entry-id]` entries and 23 `[data-chat-trace-step]` rows
  - trace tool rows carried `data-chat-trace-tool-call-id`, `data-chat-trace-entry-id`, `data-chat-trace-link-state=joined`, and `data-chat-trace-output-state`

### DIFF
--- a/dashboard/src/components/chat/primitives.test.ts
+++ b/dashboard/src/components/chat/primitives.test.ts
@@ -215,6 +215,33 @@ describe('ChatTranscript', () => {
     expect(surfaceLink?.textContent).toContain('Discord Thread')
   })
 
+  it('exposes tool-call transcript provenance as rendered attributes', () => {
+    render(
+      html`<${ChatTranscript}
+        entries=${[
+          toolEntry({
+            id: 'tool-toolu_prov',
+            label: 'keeper_context_status',
+            turnRef: 'trace-tool#3',
+            delivery: 'history',
+          }),
+        ]}
+        emptyText="empty"
+        groupToolCalls=${false}
+      />`,
+      container,
+    )
+
+    const bubble = container.querySelector('[data-chat-variant="tool-call"]') as HTMLElement
+    expect(bubble).not.toBeNull()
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Exposes machine-checkable provenance attributes on Keeper Chat trace rows (think/reason/tool/chat) instead of relying only on visible badge text, and preserves standalone tool-call transcript provenance (entry/source/delivery/turn). Consistent with #23259 and #23286. The right direction — machine-readable provenance is robust to UI label changes in a way badge-text matching is not.

P3 nit: keep the visible badge text and the machine-checkable attrs in sync (a single source for both), so the visible label and the data-attr cannot drift.

jeong-sik:
Current-head adversarial review for `d4d0b61b8216`: PASS.

Scope reviewed: adds explicit tool trace provenance/link-state attributes while keeping trace-only tool rows unjoined rather than synthesizing output.

P3: none.

PR #23289, opened 2026-07-05
[codex] Expose Keeper Chat stream contract gaps

jeong-sik:
## Summary
- add typed stream-contract provenance for keeper chat transcript entries
- stamp REST history/tool history/local queue projections as explicit contract states instead of implying live stream truth
- track tool-output hydration status/failure reasons and render hydration-failed trace state when the endpoint fails
- expose stream/hydration contract data attributes for rendered QA and future backend lifecycle replacement

## Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/tool-call-output-store.test.ts src/keeper-actions.test.ts src/components/chat/primitives.test.ts src/components/chat/primitives-interleave.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check
- pnpm --dir dashboard run build
- rendered smoke via Chrome CDP against http://127.0.0.1:5180/dashboard/keepers?keeper=verifier, desktop 1440x1000 and mobile 390x844

## Notes
- build still reports the existing unresolved Cinzel font and large chunk warnings
- Dune/OCaml build intentionally left to CI per repo workflow

### DIFF
--- a/dashboard/src/api/schemas/keeper-chat-history.test.ts
+++ b/dashboard/src/api/schemas/keeper-chat-history.test.ts
@@ -86,6 +86,35 @@ describe('safeParseKeeperChatHistoryMessage', () => {
     expect(out?.turn_ref).toBe('trace-1780648779957-00000#4071')
   })
 
+  it('passes the backend stream contract read model through when present', () => {
+    const out = safeParseKeeperChatHistoryMessage(
+      validMessage({
+        stream_contract: {
+          source: 'backend_turn_trace',
+          status: 'backend_trace_join',
+          turn_ref: 'trace-1780648779957-00000#4071',
+          trace_event_count: 2,
+          reason: 'turn_ref joined to retained trajectory/internal-history events',
+        },
+      }),
+    )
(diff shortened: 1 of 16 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds typed stream-contract provenance for Keeper Chat transcript entries: REST history / tool history / local queue projections are stamped as explicit contract states rather than implying live-stream truth, and tool-output hydration status/failures are tracked. This is the capstone of the provenance/gap-surfacing series (#23286/#23287/#23288/#23289) and removes the silent "looks live but is actually a projection" misrepresentation. Right direction.

P3 nit: 11 files — confirm the stream-contract type is defined once and shared (not redeclared per consumer), and that contract-state transitions are documented so future projections register themselves rather than silently missing a state.

jeong-sik:
Current-head adversarial review for `51c75f3485c2`: PASS.

Scope reviewed: threads queue/SSE/pending-store contracts through the read model and DOM with explicit hydration states and no silent successful default.

P3: none.

PR #23290, opened 2026-07-05
[codex] Surface Keeper Chat history turn contracts

jeong-sik:
## Summary

- Emit a `stream_contract` read model for every Keeper chat history row from the backend store.
- Distinguish causal backend trace joins from explicit absence states instead of inventing lifecycle events for old history rows.
- Teach the dashboard history schema/store to accept backend `trace`/`thinking` blocks and prefer backend `stream_contract` when present.
- Expose `turn_ref` and trace-event count in chat DOM provenance attributes for rendered/Playwright-style checks.

## Contract Shape

- `backend_turn_trace` / `backend_trace_join`: row has `turn_ref` and retained trace block joined by `turn_ref`.
- `keeper_chat_store` / `history_without_turn_ref`: old row has no turn ref, so no causal stream join is possible.
- `keeper_chat_store` / `history_without_stream_events`: row has turn ref, but no retained trace block is available for history enrichment.

This keeps the read model honest: it surfaces what the backend can prove from durable history today, without faking run lifecycle replay.

## Validation

- `scripts/dune-local.sh build test/test_keeper_chat_store.exe`
- `./_build/default/test/test_keeper_chat_store.exe` (46 tests)
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/schemas/keeper-chat-history.test.ts src/keeper-state.test.ts src/keeper-actions.test.ts src/components/chat/primitives.test.ts src/components/chat/primitives-interleave.test.ts --no-file-parallelism --maxWorkers=1` (280 tests)
- Rendered smoke via headless Chrome CDP against `http://127.0.0.1:5181/dashboard/#keepers?keeper=verifier` at 1440x1000 and 390x844.

Rendered smoke note: the live MASC runtime serving `:8935` was existing repo binary `cc9011d287`, not this branch, so the rendered page correctly showed dashboard fallback provenance (`rest_history` / `history_without_stream_events`). The new backend `stream_contract` wire shape is covered by focused OCaml tests and dashboard schema/store tests in this PR.

## Evidence

[근거] local commands above, checked 2026-07-05 20:44:56 KST, confidence High.

### DIFF
--- a/dashboard/src/api/schemas/keeper-chat-history.test.ts
+++ b/dashboard/src/api/schemas/keeper-chat-history.test.ts
@@ -86,6 +86,35 @@ describe('safeParseKeeperChatHistoryMessage', () => {
     expect(out?.turn_ref).toBe('trace-1780648779957-00000#4071')
   })
 
+  it('passes the backend stream contract read model through when present', () => {
+    const out = safeParseKeeperChatHistoryMessage(
+      validMessage({
+        stream_contract: {
+          source: 'backend_turn_trace',
+          status: 'backend_trace_join',
+          turn_ref: 'trace-1780648779957-00000#4071',
+          trace_event_count: 2,
+          reason: 'turn_ref joined to retained trajectory/internal-history events',
+        },
+      }),
+    )
+    expect(out?.stream_contract).toEqual({
+      source: 'backend_turn_trace',
+      status: 'backend_trace_join',
+      turn_ref: 'trace-1780648779957-00000#4071',
+      trace_event_count: 2,
+      reason: 'turn_ref joined to retained trajectory/internal-history events',
+    })
+  })
+
+  it('returns null when stream_contract has the wrong shape', () => {
+    expect(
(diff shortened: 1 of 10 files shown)


jeong-sik (commented):
Adversarial review (approve).

Emits a stream_contract read model for every Keeper chat history row from the backend store, and distinguishes causal backend trace joins from explicit absence states (instead of inventing lifecycle events for old history rows). This continues the provenance/contract series (#23286-23289) and removes the silent "fabricated lifecycle event" misrepresentation. Right direction.

P3 nit: 10 files — confirm the stream_contract type is defined once and the read model is consumed consistently; ensure the absence-state transitions are explicit so a new history source registers rather than silently defaulting to a live-like state.

jeong-sik:
Current-head adversarial review for `79636653aecb`: PASS.

Scope reviewed: maps backend history rows to explicit stream-contract classes, including no-turn-ref and no-stream-events cases.

P3: none.

PR #23292, opened 2026-07-05
[codex] Classify Keeper Chat stream delivery receipts

jeong-sik:
## Summary
- add an explicit Keeper Chat stream delivery-receipt classification to the backend history contract and dashboard stream contract type
- mark live dashboard SSE/queue event handling as `client_observed_sse_event`
- mark durable server lifecycle replay as `server_lifecycle_replay_only`, and retained trace/history/reconciliation gaps as `no_delivery_receipt`
- expose the receipt class as DOM proof attributes alongside existing stream contract evidence

This intentionally does not persist a durable client-delivery receipt log. It prevents server replay/history rows from being mistaken for client-observed delivery.

## Verification
- `opam exec -- ocamlformat --check lib/keeper/keeper_chat_store.ml test/test_keeper_chat_store.ml`
- `scripts/dune-local.sh build test/test_keeper_chat_store.exe && ./_build/default/test/test_keeper_chat_store.exe`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/schemas/keeper-chat-history.test.ts src/keeper-state.test.ts src/keeper-actions.test.ts src/keeper-stream.test.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard run typecheck && pnpm --dir dashboard run lint && pnpm --dir dashboard run --if-present build`
- `git diff --check`

### DIFF
--- a/dashboard/src/api/schemas/keeper-chat-history.test.ts
+++ b/dashboard/src/api/schemas/keeper-chat-history.test.ts
@@ -94,6 +94,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
           status: 'backend_trace_join',
           turn_ref: 'trace-1780648779957-00000#4071',
           trace_event_count: 2,
+          delivery_receipt: 'no_delivery_receipt',
           reason: 'turn_ref joined to retained trajectory/internal-history events',
         },
       }),
@@ -103,6 +104,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
       status: 'backend_trace_join',
       turn_ref: 'trace-1780648779957-00000#4071',
       trace_event_count: 2,
+      delivery_receipt: 'no_delivery_receipt',
       reason: 'turn_ref joined to retained trajectory/internal-history events',
     })
   })
@@ -121,6 +123,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
             'TEXT_MESSAGE_END',
             'RUN_FINISHED',
           ],
+          delivery_receipt: 'server_lifecycle_replay_only',
           reason: 'history row records durable server stream lifecycle replay',
         },
       }),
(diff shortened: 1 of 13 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds an explicit Keeper Chat stream delivery-receipt classification: live dashboard SSE/queue events are marked client_observed_sse_event and durable server lifecycle events are classified separately. This continues the provenance/contract series (#23286-23291) and makes the "seen by client" vs "exists on server" distinction explicit rather than implied. Right direction.

P3 nit: 13 files — confirm the delivery-receipt type is defined once and shared, and that the two classes cannot be confused by a consumer.

jeong-sik:
Source review on current head `64badc0101c249c34414b4bce5a7df06f635007d`: no P0/P1 blocker found in the current backend-emitted receipt path.

P2: legacy/deploy-window history fallbacks still omit the new receipt classification. The schema keeps `stream_contract` optional for deploy windows and legacy endpoints, and `normalizeHistoryEntry` synthesizes a fallback `history_without_stream_events` contract when it is absent (`dashboard/src/keeper-state.ts:876-879`). `normalizeToolHistoryEntry` does the same for tool rows (`:1445-1447`). Those fallback contracts do not set `deliveryReceipt: 'no_delivery_receipt'`, so the DOM attribute added in this PR stays absent even though this PR’s stated contract is to classify history/reconciliation gaps explicitly as no delivery receipt.

Please add `deliveryReceipt: 'no_delivery_receipt'` to those fallback contracts and cover at least one absent-`stream_contract` history fixture. Otherwise current backend rows are classified, but the exact deploy-window path documented by the schema remains ambiguous.

jeong-sik:
Current-head adversarial review for `1fea807be844`: PASS.

Scope reviewed: separates client-observed SSE receipts from server-only lifecycle replay and no-receipt history rows.

P3: none.

PR #23291, opened 2026-07-05
[codex] Persist Keeper Chat stream lifecycle contracts

jeong-sik:
## Summary
- persist a closed-list `stream_lifecycle` replay on direct Keeper Chat assistant rows for server lifecycle events
- expose the durable lifecycle replay through `/chat/history` `stream_contract` before retained-trace fallback
- normalize and render lifecycle replay provenance in the dashboard DOM contract attributes

## Boundaries
- This is server-side lifecycle provenance, not a client-delivery receipt.
- `append_turn` stores lifecycle replay only on the assistant/failure row, not on user or tool rows.
- Malformed persisted lifecycle payloads increment persistence read-drop telemetry and read as `None`; the row remains valid.

## Verification
- `scripts/dune-local.sh build test/test_keeper_chat_store.exe && ./_build/default/test/test_keeper_chat_store.exe`
- `scripts/dune-local.sh build test/test_keeper_surface_read.exe test/test_keeper_mention_scope.exe`
- `./_build/default/test/test_keeper_surface_read.exe`
- `./_build/default/test/test_keeper_mention_scope.exe`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/schemas/keeper-chat-history.test.ts src/keeper-state.test.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard run typecheck`
- `pnpm --dir dashboard run lint`
- `pnpm --dir dashboard run --if-present build`

PR #23293, opened 2026-07-05
[codex] Add Keeper Chat contract replay smoke

jeong-sik:
## Summary
- add a deterministic temp-base smoke harness for `/api/v1/keepers/:name/chat/history`
- seed real `.masc/keeper_chat/<keeper>.jsonl` rows and assert server lifecycle replay receipts are not client SSE receipts
- wire the Keeper surface readiness fixture evidence to the new harness while preserving `/api/v1/keepers/composite` as the live spotcheck
- extend the ChatTranscript DOM test to assert `no_delivery_receipt` on tool/history rows

This PR does not add durable client-delivery receipt persistence. It only locks the current replay contract so server history cannot falsely claim dashboard-observed SSE delivery.

## Verification
- `bash -n scripts/harness_dashboard_keeper_chat_contract_smoke.sh`
- `scripts/dune-local.sh build test/test_dashboard_surface_readiness.exe`
- `_build/default/test/test_dashboard_surface_readiness.exe`
- `scripts/harness_dashboard_keeper_chat_contract_smoke.sh`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1`
- `opam exec -- ocamlformat --check lib/dashboard/dashboard_surface_readiness.ml`
- `git diff --check`

### DIFF
--- a/dashboard/src/api/schemas/keeper-chat-history.test.ts
+++ b/dashboard/src/api/schemas/keeper-chat-history.test.ts
@@ -94,6 +94,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
           status: 'backend_trace_join',
           turn_ref: 'trace-1780648779957-00000#4071',
           trace_event_count: 2,
+          delivery_receipt: 'no_delivery_receipt',
           reason: 'turn_ref joined to retained trajectory/internal-history events',
         },
       }),
@@ -103,6 +104,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
       status: 'backend_trace_join',
       turn_ref: 'trace-1780648779957-00000#4071',
       trace_event_count: 2,
+      delivery_receipt: 'no_delivery_receipt',
       reason: 'turn_ref joined to retained trajectory/internal-history events',
     })
   })
@@ -121,6 +123,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
             'TEXT_MESSAGE_END',
             'RUN_FINISHED',
           ],
+          delivery_receipt: 'server_lifecycle_replay_only',
           reason: 'history row records durable server stream lifecycle replay',
         },
       }),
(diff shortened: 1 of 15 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a deterministic temp-base smoke harness for /api/v1/keepers/:name/chat/history: seeds real .masc/keeper_chat/<keeper>.jsonl rows and asserts that server lifecycle replay receipts are NOT classified as client SSE receipts, then wires the Keeper surface. This is the test that pins the #23291 contract distinction (durable lifecycle vs client-observed). Right direction.

P3 nit: confirm the smoke harness is wired into CI (test/dune alias or the runtest target) so it actually runs, not just a standalone script.

jeong-sik:
Source review on current head `8a0216c156e97eff5202fad6c87322b26be0b19c`. I would block before ready.

P1: the smoke harness can delete an arbitrary caller-supplied `BASE_PATH`. The default path is a `mktemp -d`, but the script also accepts `BASE_PATH` from the environment (`scripts/harness_dashboard_keeper_chat_contract_smoke.sh:47-50`) and cleanup removes `$BASE_PATH` whenever `KEEP_BASE_PATH != 1` (`:72-73`). That means a local or CI caller trying to reproduce against a real workspace can run `BASE_PATH=/Users/dancer/me/... scripts/harness_dashboard_keeper_chat_contract_smoke.sh`, and the trap will `rm -rf` that path on exit. The script also writes fixture files under `$BASE_PATH/.masc` before cleanup (`:89-94`), so this is not just theoretical: the harness treats the env override as writable test state.

Please track whether the script created the temp directory itself and only delete that owned temp path. For caller-supplied `BASE_PATH`, either require an explicit `ALLOW_DELETE_BASE_PATH=1`/`KEEP_BASE_PATH=0` style confirmation plus a guard that the path matches the harness temp prefix, or never delete it and only remove a harness-owned subdirectory. This should be fail-closed because `BASE_PATH` is one of the project’s core SSOT boundaries.

jeong-sik:
Current-head adversarial review for `1ba8843aa1c1`: PASS.

Scope reviewed: adds the temp-base contract smoke harness with explicit no-client-receipt assertions and guarded cleanup.

P3: none.

PR #23294, opened 2026-07-05
[codex] Expose Keeper Chat multimodal provenance

jeong-sik:
## Summary
- expose persisted chat attachments as DOM-verifiable multimodal inputs with source, kind, attachment id, mime, and size evidence
- expose server-provided `attach` rich blocks separately from persisted attachments so review/tests do not conflate the two paths
- add bubble-level aggregate attributes for attachment count, server attach block count, multimodal sources, and multimodal kinds

This PR does not add a durable client-delivery receipt log and does not change OAS. It only makes the existing MASC dashboard multimodal input/rendering paths auditable.

## Verification
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`
- `curl -fsS 'http://127.0.0.1:8935/health?full=1'`
- `pnpm --dir dashboard dev --host 127.0.0.1 --port 5187`
- `pnpm --dir dashboard exec playwright screenshot --wait-for-timeout=3000 http://127.0.0.1:5187/dashboard/#keepers /tmp/masc-chat-multimodal-provenance-keepers.png`

Browser plugin not available in this session, so rendered validation used the repo dev server plus Playwright CLI screenshot fallback. The live backend health was `status=ok` with `overall_status=degraded` due keeper fleet capacity; the dashboard route still rendered without a framework overlay.

### DIFF
--- a/dashboard/src/api/schemas/keeper-chat-history.test.ts
+++ b/dashboard/src/api/schemas/keeper-chat-history.test.ts
@@ -94,6 +94,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
           status: 'backend_trace_join',
           turn_ref: 'trace-1780648779957-00000#4071',
           trace_event_count: 2,
+          delivery_receipt: 'no_delivery_receipt',
           reason: 'turn_ref joined to retained trajectory/internal-history events',
         },
       }),
@@ -103,6 +104,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
       status: 'backend_trace_join',
       turn_ref: 'trace-1780648779957-00000#4071',
       trace_event_count: 2,
+      delivery_receipt: 'no_delivery_receipt',
       reason: 'turn_ref joined to retained trajectory/internal-history events',
     })
   })
@@ -121,6 +123,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
             'TEXT_MESSAGE_END',
             'RUN_FINISHED',
           ],
+          delivery_receipt: 'server_lifecycle_replay_only',
           reason: 'history row records durable server stream lifecycle replay',
         },
       }),
@@ -136,6 +139,7 @@ describe('safeParseKeeperChatHistoryMessage', () => {
         'TEXT_MESSAGE_END',
         'RUN_FINISHED',
       ],
+      delivery_receipt: 'server_lifecycle_replay_only',
       reason: 'history row records durable server stream lifecycle replay',
     })
   })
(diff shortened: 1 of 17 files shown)


jeong-sik (commented):
Adversarial review (approve).

Exposes persisted chat attachments as DOM-verifiable multimodal inputs (source, kind, attachment id, mime, size) and keeps server-provided attach rich blocks separate so review/tests do not conflate them. Continues the provenance line. Right direction — typed multimodal truth, with a clear persisted-vs-server-provided boundary.

P3 nit: confirm the DOM attributes clearly distinguish persisted attachments from server-provided attach blocks, so a selector cannot accidentally match the wrong source.

jeong-sik:
Source review on current head `a21310b115690914f336e27e94392a46f513d773`. I would block before ready.

P1: the advertised server attach-block provenance does not survive the real history ingest path. The renderer now exposes `data-chat-multimodal-attachment-id` and `data-chat-multimodal-mime` for server `attach` blocks (`dashboard/src/components/chat/primitives.ts:1098-1133`), and the test passes raw blocks directly into `ChatTranscript` (`dashboard/src/components/chat/primitives.test.ts:1239-1278`). But real persisted history goes through `fetchKeeperChatHistory` -> `safeParseKeeperChatHistoryMessage` first (`dashboard/src/api/keeper.ts:675-678`). That schema declares `attach` fields including `mimeType`, `sizeBytes`, and `kind`, but not `id` (`dashboard/src/api/schemas/keeper-chat-history.ts:173-186`), and Valibot `object` strips fields outside the schema. Then `normalizeBlocks` keeps `mimeType` initially (`dashboard/src/keeper-state.ts:584-601`), but for user rows it maps every block through `normalizeUserChatBlock` (`:681-686`), and that function drops `mimeType` while only keeping `sizeBytes`, `id`, and `kind` (`:480-494`). Net effect: the production user-history path cannot produce the DOM evidence this PR claims for server attach-block `id` and `mime`; the new test bypasses both schema parsing and user-row normalization, so it is a false positive.

Please align the SSOT boundaries instead of only renderer tests: add `id` to `KeeperChatBlockSchema` for `attach`, preserve `mimeType` in `normalizeUserChatBlock`, and add a regression through the actual history parsing/state normalization path. Otherwise review tools will see missing attachment id/mime evidence on real keeper chat rows even though the component-level test is green.

jeong-sik:
Current-head adversarial review for `5444a0cb9f1d`: PASS.

Scope reviewed: exposes persisted-attachment and server-block multimodal provenance without relaxing existing safe-media URL constraints.

P3: none.

PR #23295, opened 2026-07-05
[codex] Expose Keeper Chat turn order proof

jeong-sik:
## Summary
- Expose DOM-verifiable turn-order indices and signatures from the existing structural Keeper Chat timeline renderer.
- Keep queued attachment draft rows explicit as client placeholders with no delivery receipt, while preserving persisted attachment and server attach-block provenance.
- Add regression coverage for structural trace/tool/chat ordering and queued multimodal draft provenance.

## Boundaries
- Dashboard-only change; no OAS/provider changes.
- Does not add durable client receipt logs.
- Does not synthesize transcript rows or sort timeline steps by timestamp.

## Verification
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check
- Browser plugin unavailable in this session; Playwright CLI fallback screenshot: /tmp/masc-chat-turn-order-proof-keepers.png

### DIFF
--- a/dashboard/src/components/chat/primitives.test.ts
+++ b/dashboard/src/components/chat/primitives.test.ts
@@ -2167,6 +2167,60 @@ describe('ChatTranscript — tool-call grouping (turn timeline)', () => {
     expect(chat.getAttribute('data-chat-trace-turn-ref')).toBe('trace-prov#7')
   })
 
+  it('exposes structural turn order without timestamp sorting', () => {
+    render(
+      html`<${ChatTranscript}
+        entries=${[
+          toolEntry({
+            id: 'tool-t-order',
+            label: 'keeper_context_status',
+            text: '{"ok":true}',
+            timestamp: '2026-03-24T00:00:01.000Z',
+            turnRef: 'trace-order#1',
+          }),
+          entry({
+            id: 'assistant-order',
+            text: '완료했습니다',
+            role: 'assistant',
+            source: 'direct_assistant',
+            timestamp: '2026-03-24T00:00:00.000Z',
+            turnRef: 'trace-order#1',
+            traceSteps: [
+              { kind: 'think', text: 'first structural thought', ts: '2026-03-24T00:00:04.000Z' },
+              {
+                kind: 'tool',
+                name: 'keeper_context_status',
+                toolCallId: 't-order',
+                status: 'ok',
+                ts: '2026-03-24T00:00:02.000Z',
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve).

Exposes DOM-verifiable turn-order indices and signatures from the existing structural Keeper Chat timeline renderer, and keeps queued attachment draft rows explicit as client placeholders (no delivery receipt) while preserving persisted ones. Continues the provenance line. Right direction — turn-order is now machine-checkable, and the queued-vs-persisted boundary is explicit.

P3 nit: confirm the turn-order index is stable across reorders/inserts (e.g. derived from a monotonic id, not a list position that shifts), so the proof stays meaningful under mutation.

jeong-sik:
Source review on current head `c73453892956d9496342edd049793ea6a690d95c`: no P0/P1 blocker found in the turn-order DOM proof path.

P2 test fidelity issue: the new queued attachment regression asserts a queued draft with both persisted attachment provenance and `server_block` attach provenance (`dashboard/src/components/keeper-shared.test.ts:490-524`), but the production submit path does not enqueue that shape. `submit` derives `attachments = blocksToAttachments(blocks)` and then `displayBlocks = blocksToDisplayBlocks(blocks)` (`dashboard/src/components/keeper-shared.ts:733-748`); `blocksToDisplayBlocks` explicitly removes `attach` blocks (`:333-335`). So a normal attachment-only queued draft reaches `queuedInputToConversationEntry` with `attachments` present but no attach `blocks`, meaning `data-chat-server-attach-block-count` would be `0`, not the `1` asserted by the test. The code change that matters here is the `no_delivery_receipt` stream contract (`:275-278`), but this test now gives false confidence about queued server-block provenance.

I would keep the PR unblocked, but change the regression to exercise the real `ChatComposer`/`submit` path or adjust the expected queued draft provenance to persisted attachment only. Server attach-block provenance should be covered by a persisted/server transcript fixture, not by a client-local queued placeholder unless there is an actual reachable enqueue path for `blocks` containing `attach`.

jeong-sik:
Current-head adversarial review for `02a2e8b5b213`: PASS.

Scope reviewed: adds structural order signatures/attributes while preserving existing order rather than sorting by timestamp.

P3: none.

PR #23297, opened 2026-07-05
[codex] Expose Keeper Chat queue finalization proof

jeong-sik:
## Summary
- Expose stream-contract reason on chat and tool bubbles so interrupted/error reconciliation remains a visible no-receipt contract gap instead of a silent/fake delivery state.
- Add stable FIFO queue sequence to queued composer items and queued transcript placeholders so the queue card and transcript row can be joined without timestamp/id parsing.
- Add regression coverage for queue sequence stability, queue-card/transcript identity, and interrupted stream contract gap DOM proof.

## Boundaries
- Dashboard-only change; no OAS/provider changes.
- Does not add durable client receipt logs.
- Does not synthesize transcript rows or infer ordering from timestamps.

## Verification
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/keeper-chat-store.test.ts src/components/chat/primitives.test.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1
- git diff --check
- Browser plugin unavailable in this session; Playwright CLI fallback screenshots: /tmp/masc-chat-finalization-queue-proof-desktop.png and /tmp/masc-chat-finalization-queue-proof-mobile.png
- Playwright JS/API interaction script was not available because the repo/global CLI does not expose an importable playwright package; component interaction coverage is in keeper-shared.test.ts.


jeong-sik (commented):
Adversarial review (approve).

Exposes stream-contract reason on chat and tool bubbles so interrupted/error reconciliation stays a visible no-receipt contract gap (not a silent or fake delivery state), and adds a stable FIFO queue sequence to queued composer items. Continues the provenance line (#23286-23295). Right direction — the failure/intermediate states are now observable rather than implied.

P3 nit: confirm the FIFO sequence is stable under requeue/cancel (monotonic, not list-position), and that the reason type is shared across the 7 files rather than redeclared.

jeong-sik:
Current-head adversarial review for `b8357a792b22`: PASS.

Scope reviewed: adds stable queue sequence/client-action identity to queued placeholders and queue cards, including no-delivery-receipt contract proof.

P3: none.

jeong-sik:
P1 process blocker: removed `reviewed:adversarial-pass` because current head `b8357a792b22da817eda23152634413ee79ff3f0` is `DIRTY` / `CONFLICTING`. Prior source-review conclusions are not retracted, but this label is reserved for a conflict-free current head.

Mergeability: blocked until conflicts are resolved and the updated head is re-reviewed/revalidated.

PR #23303, opened 2026-07-05
[codex] Emit raw tool_execute PR action metrics

jeong-sik:
## Summary
- add a typed PR-action projection from Shell IR gh pr commands in Command_descriptor
- emit masc_keeper_tool_execute_pr_action_total from the raw tool_execute Shell IR dispatch path
- include bounded labels for keeper, surface, action, process outcome, and risk class
- cover success, failed attempt, pipeline, and non-gh-pr no-op behavior in test_command_descriptor

## Validation
- git diff --check
- scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD
- scripts/dune-local.sh build test/test_command_descriptor.exe
- ./_build/default/test/test_command_descriptor.exe

## CI Note
- CI is the build authority for this branch.
- Previous Health failure on e4260b3665 was a test namespace compile error. Fixed in eee2ec5c14.

### DIFF
--- a/lib/command_descriptor/command_descriptor.ml
+++ b/lib/command_descriptor/command_descriptor.ml
@@ -15,6 +15,39 @@ type t =
   | Pipe_chain of { first_cmd : string; last_cmd : string; length : int }
   | Generic
 
+type pr_action_surface =
+  | Gh_cli
+
+type pr_action =
+  | Create
+  | Merge
+  | Comment
+  | Close
+  | Edit
+  | Review
+  | Reopen
+  | Ready
+
+type pr_action_event =
+  { surface : pr_action_surface
+  ; action : pr_action
+  }
+
+let pr_action_surface_to_string = function
+  | Gh_cli -> "gh_cli"
+;;
+
+let pr_action_to_string = function
+  | Create -> "create"
+  | Merge -> "merge"
+  | Comment -> "comment"
+  | Close -> "close"
+  | Edit -> "edit"
+  | Review -> "review"
+  | Reopen -> "reopen"
+  | Ready -> "ready"
+;;
+
 let to_json = function
   | Gh_pr_create { title; base; draft } ->
     `Assoc
@@ -116,6 +149,18 @@ let extract_number_from_rest rest =
 let string_or default = function
   | Some value -> value
   | None -> default
+;;
+
+let pr_action_of_gh_action = function
+  | Some "create" -> Some Create
+  | Some "merge" -> Some Merge
+  | Some "comment" -> Some Comment
+  | Some "close" -> Some Close
+  | Some "edit" -> Some Edit
+  | Some "review" -> Some Review
+  | Some "reopen" -> Some Reopen
+  | Some "ready" -> Some Ready
+  | Some _ | None -> None
 ;;
 
 let compute_typed : type i o r s. (i, o, r, s) Typed.command -> t = function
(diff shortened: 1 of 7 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a typed PR-action projection from Shell IR gh pr commands in Command_descriptor and emits a masc_keeper_tool_execute_pr_action_total metric. This is genuine observability (counting actual gh pr actions) rather than telemetry-as-fix. Right direction.

P3 nit: confirm the metric label naming follows the project's Prometheus conventions (snake_case, units in the name) and that the projection covers all gh pr subcommands, not just the common ones.

jeong-sik:
Current-head review note on `e4260b36`.

P2: pipeline PR-action metrics can mislabel the PR action outcome. `pr_action_events_of_ir` flattens every PR action inside a pipeline (`lib/command_descriptor/command_descriptor.ml:510`), but `record_pr_action_metric` applies the single top-level dispatch `result.status` to every flattened event (`lib/keeper/keeper_tool_execute_runtime.ml:315-327`, called at `:881-886`). That status is the process/pipeline outcome, not the per-`gh pr` stage outcome.

Concrete drift: `gh pr comment 123 --body ok | false` can increment `action=comment,status=exit_nonzero` even if the comment was already posted, while `false | gh pr comment ...`/shell pipeline semantics can produce the opposite kind of ambiguity depending on where the PR action sits and how the dispatcher reports pipeline status. This makes the metric unsafe for operator conclusions like “PR comment attempts failed” vs “pipeline failed after a PR action”.

Suggested fix: either restrict `status=success/exit_nonzero/...` to simple top-level `gh pr` commands only, or label pipeline-derived events with an explicit aggregate status such as `pipeline_exit_nonzero`/`pipeline_status_unknown_for_stage` so the metric does not imply per-action success. Add a regression for a pipeline with a PR action and a failing non-PR stage.

Current CI note: `Health` is red, but the run is still in progress so GitHub has not exposed the failed job log yet. Build/TLA are still pending.

jeong-sik:
Current-head adversarial review for `eee2ec5c14ca185a34148f3814d98ffe0697eb51`:

P1 fixed: the prior head had a real compile blocker in `test/test_command_descriptor.ml` because the new metric assertions referenced `Otel_metric_store` / `Keeper_tool_execute_runtime` without the `Masc.` qualification visible to this test. Current head `eee2ec5` fixes that by routing the assertions through explicit `Metrics` / `Execute_runtime` aliases.

Source review after the fix: PASS. The metric is emitted only from typed Shell IR `gh pr` actions, labels are bounded (`surface`, `action`, `status`, `risk_class`) except the existing keeper dimension, and the runtime increments it only after raw `tool_execute` dispatch returns a process status. I did not find a hardcoded provider/model path, heuristic command-line scraping, stub path, silent failure, local path dependency, or MASC/OAS boundary inversion in the current diff.

Verification run locally on the current PR worktree:
- `git diff --check origin/main...HEAD && git diff --check`
- `ocamlformat --check test/test_command_descriptor.ml`
- `scripts/dune-local.sh build test/test_command_descriptor.exe`
- `scripts/dune-local.sh exec ./test/test_command_descriptor.exe` -> 34/34 passing

P2 / mergeability: source is good and I added `reviewed:adversarial-pass`, but the PR is still draft and GitHub CI is not done yet. At last check, `Detect Changed Surfaces` and `ocamlformat-check` were still pending, so merge remains blocked by process/CI state rather than source review.

PR #23313, opened 2026-07-05
[codex] Add Keeper Chat interleave contract fixture

jeong-sik:
## Summary
- add a standalone Keeper Chat interleave contract fixture for structural think/tool/chat ordering
- prove joined tool output is hydrated from the tool_calls_endpoint contract while trace-only tools stay pending
- add a Playwright CLI DOM smoke that waits for status, trace, joined tool, trace-only tool, chat, and mobile selectors

## Validation
- pnpm --dir dashboard exec vitest run --config vitest.config.ts src/demo/keeper-chat-interleave-contract-fixture.test.ts src/components/chat/primitives-interleave.test.ts src/components/chat/primitives.test.ts
- pnpm --dir dashboard run build
- scripts/harness_dashboard_keeper_chat_interleave_contract_dom_smoke.sh
- GITHUB_BASE_REF=codex/chat-replay-queued-finalization-fixture-20260705 python3 scripts/check_test_coverage.py
- git diff --check --cached && git diff --check

Browser plugin is not available in this environment, so rendered validation used the repo-local Playwright CLI fallback.

### DIFF
--- a/dashboard/dev-fixtures/keeper-chat-interleave-contract-fixture.html
+++ b/dashboard/dev-fixtures/keeper-chat-interleave-contract-fixture.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="ko" data-skin="v2" data-volt="brass">
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>MASC - Keeper Chat Interleave Contract Fixture</title>
+  <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
+</head>
+<body>
+  <div id="app"></div>
+  <script type="module" src="/src/demo/keeper-chat-interleave-contract-fixture.ts"></script>
+</body>
+</html>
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a standalone Keeper Chat interleave contract fixture for structural think/tool/chat ordering, and proves joined tool output is hydrated from the tool_calls_endpoint contract (not orchestrated post-hoc). Continues the provenance/contract test series. Right direction — ordering and join provenance are now pinned.

P3 nit: confirm the fixture is wired into CI/Playwright so it actually runs, and that the interleave ordering is asserted on data-attrs (stable under reorder), not on textContent.

jeong-sik:
current-head review (d9cdb8e)

No blocker found. I checked the fixture against the head ChatTranscript/tool-output contracts, including the structural trace/tool/chat ordering and the joined-vs-trace-only tool states.

Evidence checked:
- gh pr checks #23313: both current checks pass at d9cdb8e.
- bash -n scripts/harness_dashboard_keeper_chat_interleave_contract_dom_smoke.sh passes.
- Local Vite/Playwright smoke on a detached review worktree confirmed these selectors render: fixture status, trace contract, joined tool, trace-only tool, chat row.

Residual risk: this PR adds a fixture/harness only; it does not change the production primitive join behavior. The existing behavior that trace-only tool rows remain pending without a standalone tool entry is now explicitly covered by this fixture, but changing that semantic would need a separate production PR.

PR #23315, opened 2026-07-05
[codex] Guard keeper execute credentials

jeong-sik:
## Summary

- Add a trust-independent approval floor for irreversible repository-hosting CLI operations such as `gh pr merge`, `gh repo delete`, and `gh api -X DELETE`.
- Redact Keeper Execute stdout/stderr before keepalive output recording, stream chunk recording, failure fields, glob diagnostics, and response `output` JSON.
- Add focused regression coverage for the gh hard gate and Execute stdout/stderr/combined-output redaction.

## Validation

- `ocamlformat --check lib/exec/approval_policy.ml lib/exec/approval_policy.mli lib/exec/test/test_approval_policy.ml lib/exec/verdict.ml lib/exec/verdict.mli lib/keeper/keeper_tool_execute_runtime.ml lib/keeper/keeper_tool_execute_runtime.mli test/test_keeper_secret_redaction.ml`
- `git diff --check`
- `scripts/dune-local.sh build lib/exec/test/test_approval_policy.exe test/test_keeper_secret_redaction.exe test/test_command_descriptor.exe`
- `./_build/default/lib/exec/test/test_approval_policy.exe`
- `./_build/default/test/test_keeper_secret_redaction.exe`
- `./_build/default/test/test_command_descriptor.exe`
- `scripts/ci/check-determinism-contract.sh --base origin/main --head HEAD` (PASS; existing WARN output only)

## Audit Goal

G-CRED partial remediation: Execute output redaction coverage and gh R2 hard gate. Token scope reduction still requires credential/App rotation outside this code PR.


jeong-sik (commented):
Adversarial review (strong approve — credential/safety).

Two high-value hardenings that directly close findings from the keeper credential audit:
1. **Irreversible CLI approval floor** for `gh pr merge`, `gh repo delete`, `gh api -X DELETE` — a trust-independent gate (resolves the audit's C-3 "destructive op hard-gate absent, prompt-only" P1). This prevents a trusted keeper from silently performing an irreversible repo operation.
2. **Redact Keeper Execute stdout/stderr** before keepalive output / stream chunk / failure field / glob diagnostic recording (resolves the audit's C-2 "Execute output redaction absent, token leak risk" P1).

Right direction. Since this touches the credential/sandbox area, cite the relevant RFC (RFC-0006/0036/0074 area) or mark RFC-WAIVED.

P3 nit: document how the approval floor interacts with the existing trust tiers (does a max-trust keeper still hit the floor for destructive ops?), and confirm the redaction covers every output path across the 8 files (no leaky branch).

jeong-sik:
Adversarial current-head review (head af16eb7a29855e119e1277323b90e768cad6d6b9): source-blocked.

P1: the new Execute secret redaction still leaks when a secret spans live stream chunks. `handle_tool_execute_typed` redacts each `on_output_chunk` independently before calling `record_execute_stream_chunk` (`lib/keeper/keeper_tool_execute_runtime.ml:789-803`), but `Process_eio` reads arbitrary 4096-byte chunks (`lib/process/process_eio.ml:589-599`). A child can emit the first half of a token in one write/read and the second half in the next; neither chunk contains the full secret string, so `Keeper_secret_redaction.redact_text` will not match. Those chunks are persisted and broadcast by `Dashboard_execute_output.append_stream_chunk` (`lib/dashboard/dashboard_execute_output.ml:231-264`), and the final whole-output redaction does not re-emit replacement line events when streaming is enabled.

Expected fix: make the stream redaction stateful across chunk boundaries (keep enough trailing bytes from the previous chunk for each stream/keeper, with tests that split a secret across two chunks), or buffer/redact before broadcasting any stream output. Final stdout/stderr redaction alone is not enough for the default streaming path.

P1 / mergeability: not merge-safe yet. Draft + source blocker; leave `reviewed:adversarial-pass` off until this split-chunk stream leak is covered.

jeong-sik:
Split-chunk stream leak follow-up: #23246 (stateful line-boundary redactor). Builds green. Review there.

Issue #23309, opened 2026-07-05
ide observation: enforce scoped partition/write-path truth across Cursor, annotation, and PR ingestion


jeong-sik:
Generated from the local HTML audit set dated 2026-07-05.

## Problem
IDE observation is still vulnerable to stale/default partitioning and silent write/read failures. PR #23160 review specifically says sandboxed Keeper writes can remain `_orphan` because partition attribution is based on raw input paths rather than resolved output paths, and related documents require strict `IdeScope` across API/store/route/dashboard.

## Evidence
- `IDE Observation Plane Remaining Goals - 2026-07-04.html`: `IdeScope` is not a single wire contract; malformed/missing scope must fail closed; observation append failure must not look successful.
- `IDE Observation Plane Goal Breakdown - 2026-07-03.html`: dashboard parsing can silently degrade to empty arrays/comments; repo dirty-count is a stub unless backed by real git status.
- `PR #23160 — 적대적 리뷰.html`: sandboxed Keeper path attribution still yields `_orphan`; helper/path segment logic is duplicated; `path/file_path`-only arg inspection misses shell/multi-file/glob writes.

## Required work
- Define one typed `IdeScope` contract and use it through API, store, route handlers, dashboard, and tests.
- Partition Cursor/annotation/PR ingestion from resolved output path or typed provenance, not raw untrusted input path.
- Return typed JSON errors for missing scope, invalid focus mode, repo mismatch, and observation write failure.
- Reject body-supplied keeper_id where actor/auth provenance must own the partition.
- Replace/remove fake repo dirty-count and other IDE stubs.

## Completion evidence
- Tests for sandbox-rooted path segment boundaries, `_orphan` prevention, and write-failure response.
- Dashboard fixture showing degraded typed state rather than empty success.
- PR body includes a sandboxed Keeper write/read partition proof.

## Dedupe / related
- Related existing issues: https://github.com/jeong-sik/masc/issues/23139, https://github.com/jeong-sik/masc/issues/23138
- MASC goal: `goal-html-audit-gap-closure-20260705`

```masc-triage
kind: gap
area: observability
impact: blinds-operator
root: ssot
must-do: false
```

jeong-sik:
Started implementation in draft PR #23317: https://github.com/jeong-sik/masc/pull/23317

Implemented slice:
[...]

PR #23317, opened 2026-07-05
fix(ide): extract structured tool input paths

jeong-sik:
## Summary
- Extend the shared `Tool_input_path.tool_input_file_path` helper beyond top-level `path`/`file_path`.
- Preserve direct path priority, then inspect structured argument objects and path lists for write-path attribution.
- Keep free-form command strings ignored so attribution remains structured, not heuristic shell parsing.

## Scope
This is a focused implementation slice for #23309. It improves shared path extraction used by IDE tool events, derived cursor attribution, and keeper observation attribution. It does not complete the broader typed `IdeScope` wire contract, JSON error responses, keeper_id provenance rejection, dashboard fake-field removal, or write-failure response work from the issue.

## Validation
- `scripts/dune-local.sh build test/test_keeper_run_tools_hooks.exe test/test_ide_bridge.exe`
- `_build/default/test/test_keeper_run_tools_hooks.exe` (11 tests)
- `_build/default/test/test_ide_bridge.exe` (43 tests)
- `git diff --check`

Refs #23309

### DIFF
--- a/lib/attribution/tool_input_path.ml
+++ b/lib/attribution/tool_input_path.ml
@@ -6,13 +6,83 @@ let non_empty_string_member name input =
   | _ -> None
 ;;
 
-let tool_input_file_path input =
-  let rec first = function
-    | [] -> None
-    | name :: rest ->
-      (match non_empty_string_member name input with
-       | Some _ as path -> path
-       | None -> first rest)
-  in
-  first [ "path"; "file_path" ]
+let non_empty_string = function
+  | `String raw ->
+    let trimmed = String.trim raw in
+    if trimmed = "" then None else Some trimmed
+  | _ -> None
+;;
+
+let direct_path_keys =
+  [ "path"
+  ; "file_path"
+  ; "target_path"
+  ; "output_path"
+  ; "destination_path"
+  ]
+;;
+
+let nested_object_keys = [ "arguments"; "args"; "params"; "input" ]
+;;
+
+let list_path_keys =
+  [ "paths"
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve).

Extends the shared Tool_input_path.tool_input_file_path helper beyond top-level path/file_path to inspect structured argument fields, preserving direct-path priority. This is the right shape — extracting a typed path from structured args rather than matching strings. Good SSOT broadening.

P3 nit: 4 files — confirm the structured argument field names are centralized (so a new field is added in one place), and that the priority order (direct path first, then structured) is documented.

jeong-sik:
Adversarial current-head review (head d14d92f3c78986d4aaff645b109312816199b232): source pass.

Reviewed the shared `Tool_input_path` expansion and both consumers (`Ide_bridge` tool/cursor attribution and `Keeper_run_tools_hooks` observation attribution). The change keeps extraction structured-only: direct path fields win, then nested argument objects, then path-bearing lists. Cwd/base-path fallback remains in the keeper hook consumer, so the helper does not take over caller-specific path scoping. Added tests cover the important regressions: shared IDE cursor/tool-event priority, nested `arguments`, relative nested path + cwd, and string/object path lists.

No active unresolved review threads at this head. CI is still pending (`Build and Test`, `OCaml 5.5 Build Canary`) and the PR is draft.

P2 / mergeability: source-reviewed, but not merge-ready until draft is lifted and pending checks are green.

PR #23305, opened 2026-07-05
Wire runtime sampling parameters through OAS dispatch

jeong-sik:
## Summary
- Preserve OAS Provider_config sampling fields (top_p, top_k, min_p) in MASC RuntimeAgent config instead of dropping them at dispatch setup.
- Forward those fields into OAS Builder and resume agent/checkpoint config so resumed runs keep the same provider-declared sampling contract.
- Record completed-turn top_p from the materialized runtime Provider_config instead of leaving the dashboard turn record permanently empty.

## Evidence
- [근거] Kimi official docs, checked 2026-07-05 KST, High: Kimi K2.7 Code is latest/code-focused, always emits reasoning_content, and provider parameter constraints differ by model family. https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model
- [근거] Z.AI official docs, checked 2026-07-05 KST, High: GLM thinking/preserved-thinking uses typed request fields such as thinking.clear_thinking and reasoning_content preservation. https://docs.z.ai/guides/capabilities/thinking-mode

## Validation
- scripts/dune-local.sh build test/test_runtime_provider_auth_headers.exe
- scripts/dune-local.sh exec test/test_runtime_provider_auth_headers.exe

PR #23318, opened 2026-07-05
Wire model sampling parameters into runtime catalog

jeong-sik:
## Summary
- Add typed per-model `top-p`, `top-k`, and `min-p` fields to `runtime.toml` parsing and `Runtime_schema.model_spec`.
- Thread declared model sampling values into OAS `Provider_config.make` without changing the MASC/OAS boundary.
- Expose declared model sampling values in `/api/v1/providers` top-level snapshots and `declared_spec.model`, then surface them in dashboard runtime summaries.

## Validation
- `scripts/dune-local.sh build test/test_runtime_model_temperature_toml.exe test/test_runtime_provider_auth_headers.exe test/test_runtime_per_keeper_routing.exe`
- `_build/default/test/test_runtime_model_temperature_toml.exe`
- `_build/default/test/test_runtime_provider_auth_headers.exe`
- `_build/default/test/test_runtime_per_keeper_routing.exe`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/api/dashboard.test.ts src/components/runtime-monitor.test.ts src/components/keeper-workspace/keeper-workspace-rail.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

### DIFF
--- a/dashboard/src/api/dashboard-runtime.ts
+++ b/dashboard/src/api/dashboard-runtime.ts
@@ -134,6 +134,9 @@ export interface DashboardRuntimeDeclaredModelSpec {
   max_thinking_budget?: number | null
   streaming?: boolean
   temperature?: number | null
+  top_p?: number | null
+  top_k?: number | null
+  min_p?: number | null
   capabilities?: DashboardRuntimeDeclaredModelCapabilities | null
   match_prefixes: string[]
 }
@@ -228,6 +231,10 @@ export interface DashboardRuntimeProviderSnapshot {
   /** Per-model sampling temperature override ([models.<id>].temperature);
    *  null when unset (runtime keeps the fleet fallback). */
   temperature?: number | null
+  /** Per-model sampling overrides from [models.<id>]; null when unset. */
+  top_p?: number | null
+  top_k?: number | null
+  min_p?: number | null
   capabilities_declared?: boolean
   max_output_tokens?: number | null
   supports_tool_choice?: boolean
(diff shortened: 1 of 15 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds typed per-model top-p/top-k/min-p fields to runtime.toml parsing and Runtime_schema.model_spec, threads them into OAS Provider_config.make without changing the MASC/OAS boundary, and exposes them. Continues the #23305 sampling line. Right direction — sampling parameters become a typed, declared field rather than dropped or ad-hoc.

P3 nit: 15 files is the N-of-M risk zone — confirm the sampling bundle is defined once and threaded through every dispatch/parse site, so a future field addition cannot be silently missed at one consumer. If a single record carries the bundle end-to-end, that is the right shape.

jeong-sik:
Adversarial current-head pass at `32322d44e18cf25a3b0d2a0a1d3ff94f1bb8f06d` after a follow-up patch.

Source review result: pass. The only issue I found was a declarative consistency gap: `top-k` / `min-p` could be set on a model whose own `[models.<id>.capabilities]` declared the field unsupported. OAS has transport-level guards that drop unsupported fields before provider send, but accepting the contradictory runtime TOML left a configured no-op until request time. I patched MASC to fail that at parse time and added regression coverage.

Local validation run:
- `scripts/dune-local.sh build test/test_runtime_model_temperature_toml.exe test/test_runtime_provider_auth_headers.exe test/test_runtime_per_keeper_routing.exe`
- `scripts/dune-local.sh exec ./test/test_runtime_model_temperature_toml.exe` (14 tests)
- `ocamlformat --check lib/runtime/runtime_toml.ml test/test_runtime_model_temperature_toml.ml`
- `git diff --check`

Process state: still draft, mergeable, no active review threads, fresh GitHub checks are queued/in progress after the push. `reviewed:adversarial-pass` applied for the current source head only.

P2 / mergeability: source-pass at `32322d44e18c`; blocked only by draft state and pending CI.

PR #23320, opened 2026-07-05
[codex] prove live chat hydration contract wiring

jeong-sik:
## What

- Adds a KeeperConversationPanel regression test for the live Keeper Chat path.
- The test renders the real shared panel, updates the tool-output hydration store from idle to failed, and asserts ChatTranscript DOM updates to `hydration-failed` with the endpoint failure reason.

## Why

Previous PRs proved fixture-level failure-state rendering. This PR proves the live panel wiring reads the keeper-scoped hydration contract from the store and forwards it into ChatTranscript instead of relying on fixture props.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts`
- `GITHUB_BASE_REF=codex/chat-failure-interaction-smoke-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check`

### DIFF
--- a/dashboard/src/components/keeper-shared.test.ts
+++ b/dashboard/src/components/keeper-shared.test.ts
@@ -92,6 +92,11 @@ import {
   sendKeeperThreadMessage,
 } from '../keeper-actions'
 import { _resetChatStoreForTests, enqueueInput, getQueuedMessages, getQueueLength } from '../keeper-chat-store'
+import {
+  markToolCallOutputsHydrationFailed,
+  markToolCallOutputsHydrating,
+  resetToolCallOutputs,
+} from '../tool-call-output-store'
 import { shellAuthSummary } from '../store'
 import type { ChatBlock, KeeperConversationAttachment, KeeperConversationEntry, KeeperUserInputBlock } from '../types'
 import {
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a KeeperConversationPanel regression test for the live Keeper Chat path: renders the real shared panel and updates the tool-output hydration store from idle. Hardens the provenance/hydration contract series with a live-path test. Right direction.

P3 nit: confirm the test is wired into the vitest run (CI), not manual-only.

jeong-sik:
Adversarial current-head pass at `f8b08300332c9e6370f5564e06c29e48543d69a5`.

Reviewed the incremental stack diff against base `codex/chat-failure-interaction-smoke-20260706` (`3ec32fdf393181b6ac345b827267db89c8104583`). The effective change is only `dashboard/src/components/keeper-shared.test.ts`: it renders the real `KeeperConversationPanel`, mutates the live `tool-call-output-store`, and verifies the transcript receives the failed hydration contract and exposes the failure reason. I did not find a source blocker.

Validation run locally on this head:
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared.test.ts --no-file-parallelism --maxWorkers=1` (25 passed)

GitHub rollup is green for this head. `reviewed:adversarial-pass` applied for the current source head.

P2 / mergeability: source-pass and mergeable; blocked only by draft/stack readiness.

PR #23329, opened 2026-07-05
[codex] dashboard: prove chat reconnect refresh recovery

jeong-sik:
## Summary
- Adds an integration proof that the active-keeper reconnect refresh path recovers a live Keeper Chat transcript from stale `hydration-failed` DOM to joined hydrated tool output.
- Exercises `refreshActiveKeeperChatHistory({ force: true })` through `activeKeeperName`, without directly calling `hydrateKeeperChatHistory` in the new proof.

## Why
SSE reconnect recovery is the path that knows replay-buffer gaps may have occurred. The dashboard needs evidence that this wrapper path updates the same live `KeeperConversationPanel` DOM, clears stale failure status/reason text, and exposes the recovered `tool_calls_endpoint` output.

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared-hydration.integration.test.ts`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-shared-hydration.integration.test.ts src/components/keeper-shared.test.ts src/keeper-actions.test.ts`
- `pnpm --dir dashboard run typecheck`
- `pnpm --dir dashboard run lint`
- `pnpm --dir dashboard run build`
- `GITHUB_BASE_REF=codex/chat-hydration-refresh-recovery-live-dom-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check`

Local Dune build intentionally not run; dashboard validation only.

PR #23333, opened 2026-07-05
[codex] dashboard: prove sse reconnect chat boundary

jeong-sik:
## Summary
- Adds a reconnect-path test proving the actual SSE reconnect watcher invokes `refreshActiveKeeperChatHistory({ force: true })` through the dynamic `keeper-runtime` boundary.
- Covers the failure boundary: if the active keeper chat refresh throws, reconnect recovery logs an explicit warning and continues namespace/dashboard recovery instead of silently swallowing the break.

## Why
The previous proof covered the active-keeper wrapper itself. This PR pins the outer SSE reconnect recovery path so replay-buffer gap recovery remains route-independent and observable when the keeper refresh boundary is broken.

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/sse-store.test.ts`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/sse-store.test.ts src/components/keeper-shared-hydration.integration.test.ts src/keeper-actions.test.ts`
- `GITHUB_BASE_REF=codex/chat-hydration-reconnect-active-live-dom-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check`

Local Dune build intentionally not run; dashboard validation only.

### DIFF
--- a/dashboard/src/components/keeper-shared-hydration.integration.test.ts
+++ b/dashboard/src/components/keeper-shared-hydration.integration.test.ts
@@ -55,7 +55,11 @@ vi.mock('../store', async () => {
 vi.mock('./common/toast', () => ({ showToast: vi.fn() }))
 
 import { KeeperConversationPanel } from './keeper-shared'
-import { _resetChatHydrationForTests, hydrateKeeperChatHistory } from '../keeper-actions'
+import {
+  _resetChatHydrationForTests,
+  hydrateKeeperChatHistory,
+  refreshActiveKeeperChatHistory,
+} from '../keeper-actions'
 import {
   activeKeeperName,
   keeperActionErrors,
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a reconnect-path test proving the actual SSE reconnect watcher invokes refreshActiveKeeperChatHistory({ force: true }) through the dynamic keeper-runtime boundary. Pairs with #23329 (reconnect recover). Right direction — the reconnect → force-refresh contract is now pinned, so a transient SSE drop cannot leave stale state.

P3 nit: confirm the test runs in CI (vitest), and asserts the force flag is actually passed (not just that refresh was called).

jeong-sik:
Adversarial pass on current head `5fb8e6399df7b95bdfa80aeda63d44ddff625edc`.

- P0/P1: none found in the reviewed source surface. The diff is a focused dashboard test addition for the SSE reconnect boundary: a thrown active-keeper chat refresh is warned, but dashboard recovery continues.
- P2: process only. The PR is draft and stacked on #23329 (`5c964a3227aa98b0c9ea3a489a66fba24e42dbb5`), so merge order still matters.
- P3: none.

Verification: no review threads; `git diff --check 5c964a3227aa98b0c9ea3a489a66fba24e42dbb5...HEAD` passed; `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/sse-store.test.ts` passed (39 tests); `pnpm --dir dashboard run typecheck` passed. GitHub reports `MERGEABLE` / `CLEAN`; PR Axis Cross-Check and Test Coverage Check are green. Label added: `reviewed:adversarial-pass`.

PR #23337, opened 2026-07-05
dashboard: prove keeper config live-backed fields

jeong-sik:
## Summary
- Add server-generated `field_presence.present_paths` to `/api/v1/keepers/:name/config` so Keeper config controls can prove which fields were actually projected.
- Make the Keeper config control ledger validate against raw/backend field presence instead of normalized defaults, preventing fabricated defaults from looking live-backed.
- Stop the cognition inspector from rendering missing `social_model_recognized` as the fake successful `runtime` value; it now distinguishes `unknown`, `recognized`, and `needs attention`.

## Validation
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/api/dashboard.test.ts src/components/keeper-config-panel.test.ts src/components/keeper-cognition-inspector.test.ts`
- `opam exec -- ocamlformat --check lib/dashboard/dashboard_http_keeper_snapshot.ml`
- `scripts/dune-local.sh build lib/dashboard/masc_dashboard.cmxa`

## Task Notes
- Related to #23309 / MASC task-1823.
- This closes the live-backed field proof slice and one social-runtime fake value. Remaining audit items include broader redacted runtime/model labels, bootstrap synthetic-empty metadata, state-diagram placeholders, and runtime-lens lane completeness.

### DIFF
--- a/dashboard/src/api/dashboard-keeper-config.ts
+++ b/dashboard/src/api/dashboard-keeper-config.ts
@@ -89,6 +89,38 @@ function normalizeKeeperConfigActiveGoals(raw: unknown): KeeperConfig['workspace
     .filter((item): item is KeeperConfig['workspace']['active_goals'][number] => item !== null)
 }
 
+function dedupeStringList(values: readonly string[]): string[] {
+  return Array.from(new Set(values.filter(value => value.trim() !== '').map(value => value.trim())))
+}
+
+function collectRawFieldPaths(raw: unknown, prefix = ''): string[] {
+  if (!isRecord(raw)) return []
+  const paths: string[] = []
+  for (const [key, value] of Object.entries(raw)) {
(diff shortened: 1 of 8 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds server-generated field_presence.present_paths to /api/v1/keepers/:name/config so Keeper config controls can prove which fields were actually projected (not just what the UI claims). Continues the #23274 backing-truth line. Right direction — the backing ledger is now verifiable against the server, removing the chance the UI lies about live-backed fields.

P3 nit: 8 files — confirm the field_presence schema is defined once and the present_paths values stay consistent with the backing labels (a field marked live-write must appear in present_paths; a browser-local field must not).

PR #23338, opened 2026-07-05
[codex] Gate keeper provider cooldown before dispatch

jeong-sik:
## Goal

G-REL follow-up from `Keeper Write-Path, Credential, Goal-Loop Audit & Goal Matrix - 2026-07-05.html`.

Live receipt sample before this slice (last 7d, `/Users/dancer/me/.masc/keepers/*/execution-receipts/*/*.jsonl`):

- total receipts: 44,605
- failed receipts: 6,013 (13.481%)
- `api_error_rate_limited`: 3,765 (largest terminal reason)

## Changes

- Restore provider-health recording on the live `Keeper_turn_driver.run_named` attempt path.
  - accepted response records success with latency
  - accept rejection records rejected health
  - provider/runtime error records typed health error
- Add pre-dispatch provider cooldown gate before `run_try_provider`.
  - cooled candidate emits a non-silent `Provider_attempt_finished` manifest with `provider_attempt_started=false`
  - returns typed `Capacity_backpressure` so lane fallback can try the next runtime candidate
- Record hard quota and soft rate-limit health on both keeper-scoped keys and credential-pool provider keys.
  - a soft 429 from one keeper now blocks the same provider candidate for other keepers before dispatch
  - success remains keeper-scoped, so one keeper success does not clear a fleet/account cooldown
- Add focused regression tests for:
  - soft rate-limit cooldown blocking before dispatch, including another keeper via credential-pool key
  - provider-cooldown typed backpressure preserving lane fallback checkpoint flow

## Validation

- `ocamlformat --check lib/keeper/keeper_turn_driver_provider_attempt.ml lib/keeper/keeper_turn_driver_provider_attempt.mli lib/keeper/keeper_turn_driver.ml lib/keeper/keeper_turn_driver.mli test/test_keeper_sdk_error_typed_bridge.ml test/test_keeper_turn_driver_failover.ml`

Focused local dune build was attempted:

- `scripts/dune-local.sh build test/test_keeper_sdk_error_typed_bridge.exe test/test_keeper_turn_driver_failover.exe`
- blocked by an unrelated live bare `dune build .` process (`PID 33803`), so CI is the dune build authority for this PR.

### DIFF
--- a/lib/keeper/keeper_turn_driver.ml
+++ b/lib/keeper/keeper_turn_driver.ml
@@ -646,71 +646,103 @@ let run_named
             ~max_concurrent:runtime.Runtime.binding.max_concurrent
             provider_config
         in
-        let name = Printf.sprintf "oas-%s" attempt_runtime_id in
-        let try_provider_ctx : Keeper_turn_driver_try_provider.try_provider_ctx =
-          { runtime_id = attempt_runtime_id
-          ; error_runtime_id
-          ; base_path
-          ; keeper_name
-          ; name
-          ; goal
-          ; goal_blocks
-          ; priority
-          ; session_id
-          ; system_prompt
-          ; tools
-          ; initial_messages
-          ; max_turns
-          ; max_idle_turns
-          ; stream_idle_timeout_s
-          ; execution_idle_timeout_s
-          ; body_timeout_s
-          ; temperature
-          ; max_tokens = inference_policy.attempt_max_tokens
-          ; accept
-          ; guardrails
-          ; hooks
-          ; context_reducer
-          ; raw_trace
-          ; transport_resolved
-          ; runtime_mcp_policy
-          ; allowed_paths
-          ; checkpoint_sidecar
-          ; cache_system_prompt
-          ; yield_on_tool
-          ; compact_ratio
-          ; context_window_tokens
-          ; oas_auto_context_overflow_retry
-          ; checkpoint_dir
-          ; context_injector
-          ; context
-          ; enable_thinking = inference_policy.attempt_enable_thinking
-          ; preserve_thinking = inference_policy.attempt_preserve_thinking
-          ; approval
-          ; exit_condition
-          ; exit_condition_result
-          ; summarizer
-          ; oas_checkpoint
-          ; trace_link
-          ; sw
-          ; net
-          ; on_event
-          ; on_yield
-          ; on_resume
-          ; agent_ref
-          ; on_runtime_observation
-          ; event_bus
-          ; runtime_manifest_context
-          ; runtime_manifest_append
-          ; turn_start
-          ; seq_ref
-          }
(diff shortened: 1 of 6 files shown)


jeong-sik (commented):
Adversarial review (approve).

Gates keeper provider cooldown before dispatch (follow-up from the keeper audit G-REL). A keeper hitting a provider in cooldown skips dispatch instead of hammering/re-failing — the right fail-closed shape that also avoids the "one stuck keeper" cascade. Right direction.

P3 nit: confirm the cooldown remaining-time calculation is bounded and per-keeper/per-runtime (not a global gate), and that the cooldown is cleared on a genuine success so a transient provider blip does not park a keeper forever. 6 files — keep the cooldown state in one owner.

jeong-sik:
P0: none found in the cooldown-gate diff.

P1: blocked on checkpoint flow. The new pre-dispatch cooldown branch in `run_named` returns `(Error (provider_cooldown_block_error ...), None)` when a candidate is skipped before dispatch. The outer lane loop treats the typed `Capacity_backpressure Provider_capacity` as retryable, then passes `checkpoint_after` into the next iteration. Because this branch always returns `None`, it can erase an existing `resume_checkpoint` from a previous failed candidate before trying the following candidate.

Concrete flow: candidate A runs, fails retryably, and produces `checkpoint_after = Some cp`; candidate B is blocked by provider cooldown before dispatch and returns `None`; candidate C then runs without `cp`. That loses the transcript/checkpoint continuity the lane retry path is supposed to preserve. The new test does not cover the real branch because it stubs the cooldown error as `Error ..., Some checkpoint_after_primary`; the production branch at current head returns `None` before `run_try_provider` can produce a checkpoint.

Fix direction: preserve the incoming `resume_checkpoint` for pre-dispatch cooldown skips, or make the retry checkpoint decision distinguish “attempt not started” from “attempt started and produced no checkpoint”. Add a regression with three candidates: A produces a checkpoint, B is pre-dispatch-cooldown-blocked, C must receive A’s checkpoint.

P2: local formatting/diff hygiene is fine on the touched files.

P3 / mergeability: no `reviewed:adversarial-pass` label. Several GitHub checks were still queued/in progress on this current head during the pass.

[근거] `gh pr view 23338 --json headRefOid,baseRefOid,statusCheckRollup` -> head `df1b4777d6f4deab869be696b032d54a02ae306b`, base `dc6fb115e4cc9e22c7084e4337d97bc5ef5d3612`; `git diff --check dc6fb115e4cc9e22c7084e4337d97bc5ef5d3612...HEAD` passed; `opam exec -- ocamlformat --check` passed for the six touched OCaml/test files; flow evidence: `lib/keeper/keeper_turn_driver.ml` current PR tree lines 126-148 and 649-659. 확인일시: 2026-07-06 KST. 신뢰도: High.

PR #23339, opened 2026-07-05
[codex] split dashboard delta payload serialization

jeong-sik:
## What changed

- Split WebSocket `dashboard/delta` delivery into a shared payload frame plus a small per-session seq frame.
- Cache the serialized dashboard delta payload by the physical SSE broadcast reference so fan-out pays one payload `Yojson.Safe.to_string` per broadcast instead of one per subscribed session.
- Add `masc_ws_delta_payload_serializations_total` and expose it in transport delivery health so the Wave A target can be measured after CI/runtime rollout.
- Keep dashboard client compatibility by treating payload-only deltas as hydration frames and seq-only deltas as ACK checkpoints.

## Why

The perf-collapse report identified per-session dashboard delta reserialization as the hot path: `N sessions x payload bytes` can monopolize the single Eio serving domain during fan-out. The per-session monotonic seq still exists, but it no longer forces the large payload to be unique for every recipient.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/dashboard-ws.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm --dir dashboard exec eslint src/dashboard-ws.ts src/dashboard-ws.test.ts`
- `ocamlformat --check lib/server/server_mcp_transport_ws.ml lib/server/server_mcp_transport_ws.mli lib/transport_metrics.ml lib/transport_metrics.mli lib/otel_metric_store/otel_transport_metric_names.ml lib/otel_metric_store/otel_transport_metric_names.mli test/test_ws_transport.ml`
- `git diff --check`

Local dune build intentionally not run; CI is the OCaml build authority for this repo.

### DIFF
--- a/dashboard/src/dashboard-ws.test.ts
+++ b/dashboard/src/dashboard-ws.test.ts
@@ -1037,6 +1037,45 @@ describe('dashboard websocket route subscriptions', () => {
     expect(dashboardWsLastSeq.value).toBe(42)
   })
 
+  it('hydrates payload-only dashboard deltas without acking', async () => {
+    installWebSocketMocks()
+
+    await connectDashboardWS({ tab: 'overview', params: {} })
+    const socket = mockSockets[0]!
+    socket.open()
+    const hello = parseRpc(socket, 0)
+    socket.receive({ jsonrpc: '2.0', id: hello.id, result: {} })
+    await flushPromises()
+
+    const subscribe = parseRpc(socket, 1)
+    socket.receive({
+      jsonrpc: '2.0',
+      id: subscribe.id,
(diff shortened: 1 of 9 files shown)


jeong-sik (commented):
Adversarial review (approve).

Splits the WebSocket dashboard/delta delivery into a shared payload frame plus a small per-session seq frame, and caches the serialized delta payload by the physical SSE broadcast. This avoids re-serializing the same delta per session and is a clean perf improvement. Right direction.

P3 nit: 9 files — confirm the payload-frame schema is defined once and the per-session seq handles out-of-order delivery (a reconnecting client must not replay stale frames). Confirm the cache invalidates on real state change, not on a timer.

PR #23346, opened 2026-07-05
fix(keeper): stateful stream redaction for split-chunk secrets (#23315 follow-up)

jeong-sik:
## Summary

Follow-up to #23315 (Guard keeper execute credentials). The Execute stream redaction added there was applied **per-chunk** via the stateless `Keeper_secret_redaction.redact_text`, so a single-line secret split across a 4096-byte chunk boundary (token tail in chunk N, head in N+1) matched in **neither** call and leaked into the broadcast/persisted stream. The final whole-output redaction does not re-emit replacement events on the streaming path, so the leak was not recovered. (Raised by adversarial review on #23315.)

## Fix

Adds a **stateful line-boundary redactor** to `Keeper_secret_redaction`:
- `stream_state` (per-stream buffer of raw bytes after the last `\n`)
- `redact_stream_chunk t state chunk` — buffers, redacts the complete-line prefix, holds the remainder
- `redact_stream_finish t state` — flushes the remaining buffered bytes as one redacted unit

The runtime keeps separate stdout/stderr stream states and flushes both at stream end (before `record_execute_stream_end`).

Why line-boundary and not an overlap window: `redact_text` changes string length (`[REDACTED]` != secret length), so there is no way to map raw-byte positions back to redacted-byte positions for a sliding window. Line boundaries give an exact reassembly point without position tracking.

## Regression Coverage

Adds tests for:
- a single-line secret split across two stream chunks, held until newline and then emitted redacted
- an unterminated stream tail flushed at `redact_stream_finish` without leaking the secret

## Known limitation (not a regression)

Multi-line secrets (a value containing `\n`, e.g. a PEM block) may still partially surface if the buffer's last `\n` falls inside the secret. This is strictly better than the pre-fix behaviour where **every** cross-chunk split leaked.

## Verified

- `opam exec -- ocamlformat --check lib/keeper/keeper_secret_redaction.ml lib/keeper/keeper_secret_redaction.mli lib/keeper/keeper_tool_execute_runtime.ml test/test_keeper_secret_redaction.ml`
- `scripts/dune-local.sh build test/test_keeper_secret_redaction.exe`

Local full Dune build was not run; per repo workflow, Dune build authority remains CI.

### DIFF
--- a/lib/keeper/keeper_secret_redaction.ml
+++ b/lib/keeper/keeper_secret_redaction.ml
@@ -123,6 +123,36 @@ let redact_text t text =
   in
   Observability_redact.redact_text text
 
+(* Streaming chunk redaction. [redact_text] is stateless: a single-line
+   secret split across chunk N's tail and chunk N+1's head matches in
+   neither call. [stream_state] buffers raw bytes up to the last ['\n'] so
+   the containing line is reassembled before redaction. Multi-line secrets
+   (a value containing ['\n'], e.g. a PEM block) may still partially
+   surface if the buffer's last ['\n'] falls inside the secret; this is
+   strictly better than the pre-fix behaviour where every cross-chunk split
+   leaked. *)
+type stream_state = { pending : Buffer.t }
+
+let create_stream_state () = { pending = Buffer.create 256 }
+
+let redact_stream_chunk t state chunk =
+  Buffer.add_string state.pending chunk;
+  let contents = Buffer.contents state.pending in
+  match String.rindex_opt contents '\n' with
+  | None -> ""
+  | Some nl_pos ->
+    let safe_len = nl_pos + 1 in
+    let safe_raw = String.sub contents 0 safe_len in
+    let held = String.sub contents safe_len (String.length contents - safe_len) in
+    Buffer.clear state.pending;
+    Buffer.add_string state.pending held;
+    redact_text t safe_raw
+
+let redact_stream_finish t state =
+  let contents = Buffer.contents state.pending in
+  Buffer.clear state.pending;
+  if String.equal contents "" then "" else redact_text t contents
+
 let rec redact_json_exact t = function
   | `String s -> `String (redact_text t s)
   | `Assoc fields ->
--- a/lib/keeper/keeper_secret_redaction.mli
+++ b/lib/keeper/keeper_secret_redaction.mli
@@ -18,6 +18,21 @@ val redact_text : t -> string -> string
     with [\[REDACTED\]], preserving message length semantics except for
     the replacements themselves. *)
 
+type stream_state
+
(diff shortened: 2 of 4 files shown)


jeong-sik:
Current-head follow-up at `9d4305d643615dd1de5648a0a7876e329dd8cb84`.

P0: Fixed the missing regression coverage blocker for this security PR. The previous head added stateful Execute stream redaction without changing any test file, and CI correctly failed Test Coverage Check.

P1: Added `test_stream_redaction_reassembles_split_secret` and `test_stream_redaction_finish_redacts_held_tail` in `test/test_keeper_secret_redaction.ml`, covering both cross-chunk single-line secrets and the end-of-stream held tail.

P2: Focused local validation passed: `ocamlformat --check` on touched OCaml files, `git diff --check origin/main...HEAD`, `python3 scripts/check_test_coverage.py`, and `scripts/dune-local.sh build test/test_keeper_secret_redaction.exe`. I did not run a full local Dune build.

P3: New CI is running on the pushed head. Test Coverage Check is already green; mergeability remains blocked by draft status, no review, and pending GitHub checks.

jeong-sik:
Current-head adversarial review (`9d4305d643615dd1de5648a0a7876e329dd8cb84`, base `fa03c137a8d9d22e83fa8ae9861438c49c0b75d0`) at `2026-07-06 03:03 KST`.

P0/P1/P2: none found in this delta.

Reviewed path:
- `Keeper_secret_redaction.redact_stream_chunk` buffers through the last newline and redacts complete lines, so a single-line secret split across stream chunks is reassembled before emission.
- `redact_stream_finish` flushes the held tail before stream end.
- Execute runtime uses separate stdout/stderr stream states and still redacts the final captured output through `redact_execute_output`.
- I checked the newline-less chunk behavior. It can delay progress-style output until newline or finish, but I did not find a leak or correctness blocker for this security fix.

Validation:
- `git diff --check`
- `opam exec -- ocamlformat --check lib/keeper/keeper_secret_redaction.ml lib/keeper/keeper_secret_redaction.mli lib/keeper/keeper_tool_execute_runtime.ml test/test_keeper_secret_redaction.ml`
- `scripts/dune-local.sh build test/test_keeper_secret_redaction.exe`

Mergeability: `MERGEABLE`; merge state: `BLOCKED`. I did not apply `reviewed:adversarial-pass` because the PR is draft and GitHub CI was not terminal green: `Build and Test` and `TLA+ Model Checking` were still `IN_PROGRESS` in `statusCheckRollup`.

[근거] `gh pr view 23346 --repo jeong-sik/masc --json headRefOid,baseRefOid,isDraft,mergeable,mergeStateStatus,statusCheckRollup` plus the local validation commands above, 확인일시 `2026-07-06 03:03 KST`, 신뢰도 High.

PR #23352, opened 2026-07-05
[codex] Stabilize keeper dashboard CI checks

jeong-sik:
## Summary
- Update `test_otel_zero_fill` expected Keeper_metrics.all size from 242 to 243.
- Initialize a minimal runtime.toml in namespace-truth keeper fixture tests so fixture keepalive stays Running before count assertions.

## Context
- Post-merge CI run `28746655690` attempt 3 failed first because `test_otel_zero_fill.ml` expected 242 but current `Keeper_metrics.all` has 243 entries.
- PR run `28753963845` proved that fix, then failed next in `test_dashboard_namespace_truth.ml`: keeper heartbeat crashed because `Runtime.init_default` had not run, so `counts.keepers` read 0.
- Both CI runs later hit retry/build disk exhaustion, but this PR targets the deterministic failures before retry.

## Validation
- `git diff --check`
- `ocamlformat --check test/test_dashboard_namespace_truth.ml`
- `ocamlformat --check test/test_otel_zero_fill.ml`
- `scripts/dune-local.sh build test/test_dashboard_namespace_truth.exe test/test_otel_zero_fill.exe`
- `./_build/default/test/test_dashboard_namespace_truth.exe`
- `./_build/default/test/test_otel_zero_fill.exe`

## Best Programmer Self-Review
- Scope: test-only CI unblock for stale metric tripwire and startup-realistic keeper fixture.
- Risk: low; no production code path changed.
- Source of truth check: `counts.keepers` is explicitly sourced from runtime keepalive, and keeper heartbeat now gets the same minimal runtime initialization pattern used by other keeper tests.

### DIFF
--- a/test/test_dashboard_namespace_truth.ml
+++ b/test/test_dashboard_namespace_truth.ml
@@ -67,6 +67,42 @@ sandbox_profile = "local"
 goal = "Dashboard keeper fixture"
 |}
 
+let test_runtime_toml =
+  {|
+[runtime]
+default = "test_provider.test_model"
+
+[providers.test_provider]
+display-name = "Test Provider"
+protocol = "openai-compatible-http"
+endpoint = "http://127.0.0.1:1"
+
+[models.test_model]
+api-name = "test-model"
+max-context = 8192
+tools-support = true
+streaming = true
+
+[test_provider.test_model]
+is-default = true
+max-concurrent = 1
+|}
+
+let init_runtime_default_for_tests dir =
+  let path = Filename.concat dir "runtime.toml" in
+  write_file path test_runtime_toml;
+  match Runtime.init_default ~config_path:path with
+  | Ok () -> ()
+  | Error e -> failf "Runtime.init_default failed: %s" e
+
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Updates test_otel_zero_fill expected Keeper_metrics.all size from 242 to 243 to match the current constructor set. Keeps the ratchet honest. Minimal, correct.

P3 nit: none for a one-line tripwire bump, but confirm the new count matches the actual constructor count (the tripwire exists to catch exactly this drift).

jeong-sik (commented):
Adversarial review (approve).

Adds a runtime.toml fixture + init_default helper to test_dashboard_namespace_truth and updates the zero-fill expected count. CI-stabilizing change: the namespace-truth tests now back themselves with a real runtime instead of relying on ambient state. Right direction.

P3 nit: confirm the fixture (port 127.0.0.1:1, max-concurrent=1) does not collide with other tests sharing the same Runtime.for_testing snapshot, and the zero-fill count matches the actual constructor set.

PR #23285, opened 2026-07-05
fix(keeper): harden autonomous reactive health

jeong-sik:
## Summary
- remove keyword/stigmergy board relevance as a final wake authority and delete the dead keyword relevance module
- gate cursor-scanned board collection before cursor advancement when the keeper cannot run, and surface board collection failures in health/runtime resolution
- expose durable keeper event queues, turn admission pressure, proactive policy knobs, and typed scheduler keeper wake stimuli
- add the audit report as Markdown/HTML under docs/audit

## Validation
- scripts/dune-local.sh exec ./test/test_board_collect_pause_gate.exe
- scripts/dune-local.sh exec ./test/test_server_runtime_bootstrap.exe -- test bootstrap 28-30,53
- scripts/dune-local.sh exec ./test/test_server_runtime_bootstrap.exe -- test bootstrap 57-60
- scripts/dune-local.sh exec ./test/test_schedule_consumer_dispatch.exe
- scripts/dune-local.sh exec ./test/test_schedule_tool_wiring.exe
- scripts/dune-local.sh exec ./test/test_keeper_reaction_ledger.exe
- scripts/dune-local.sh exec ./test/test_keeper_event_queue.exe
- scripts/dune-local.sh exec ./test/test_keeper_turn_admission.exe
- git diff --check

Full local dune build intentionally not run; CI is the build authority for this branch.

PR #23353, opened 2026-07-06
[codex] Record board attention candidates

jeong-sik:
## Summary

G-REACT follow-up for the keeper audit matrix.

Current `main` already removed the stale substring/keyword wake authority: board signals only wake from typed deterministic reasons (`explicit_mention`, thread reply after self comment, reaction after self activity). This PR fills the remaining observability gap by recording durable `board_attention_candidate` evidence whenever a running keeper sees a board signal but has no deterministic wake reason.

Key behavior:

- add `Keeper_board_attention_candidate`, an append-only per-keeper JSONL ledger under `.masc/board_attention_candidates/`
- mark each candidate with typed authority fields:
  - `attention_authority = llm_judge_required`
  - `wake_authority = no_direct_wake`
- record candidates from `wakeup_relevant_keeper_for_board_signal` only for `wake_reason = None` on running keepers
- keep wake selection unchanged: candidates are evidence only and do not enqueue/wake a keeper
- add `masc_keeper_board_signal_attention_candidate_total` for recorded candidates
- pin JSON round-trip, dedupe, and payload-identity behavior in `test_keeper_board_attention_candidate`

This intentionally does not fake an LLM judge. It creates the durable typed boundary/evidence so a real judge integration can consume candidates later without reintroducing substring or keyword authority.

## Validation

- `ocamlformat --check lib/keeper/keeper_board_attention_candidate.ml lib/keeper/keeper_board_attention_candidate.mli lib/keeper/keeper_keepalive_signal.ml lib/keeper_metrics/keeper_metrics.ml lib/keeper_metrics/keeper_metrics.mli test/test_keeper_board_attention_candidate.ml test/dune`
  - PASS for new debt; advisory warnings are existing files outside this diff.

Blocked locally:

- `scripts/dune-local.sh build test/test_keeper_board_attention_candidate.exe test/test_smart_heartbeat_keepalive.exe`
  - refused because an unrelated bare `dune build .` process is live: PID 38238, PPID 63224.
  - Per project direction, CI remains the dune build authority.

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -1,0 +1,360 @@
+(* See .mli. *)
+
+type signal_kind =
+  | Post_created
+  | Comment_added
+  | Reaction_changed
+
+type attention_authority =
+  | Llm_judge_required
+
+type wake_authority =
+  | No_direct_wake
+
+type candidate = {
+  candidate_id : string;
+  dedupe_key : string;
+  keeper_name : string;
+  post_id : string;
+  signal_kind : signal_kind;
+  author : string;
+  title : string;
+  content_preview : string;
+  hearth : string option;
+  updated_at : float option;
+  recorded_at : float;
+  attention_authority : attention_authority;
+  wake_authority : wake_authority;
+}
+
+type record_result =
+  [ `Recorded
+  | `Duplicate of candidate
+  | `Error of string
+  ]
+
+let candidate_dir base_path =
+  Filename.concat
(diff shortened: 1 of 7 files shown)


jeong-sik (commented):
Adversarial review (approve).

G-REACT follow-up for the keeper audit: records board attention candidates with typed deterministic classification (continues the #23285 keyword-removal line — attention is no longer a heuristic guess but a typed candidate set). Right direction.

P3 nit: 7 files — confirm the candidate schema is defined once and that a post NOT matching any candidate surfaces as such (not silently hidden), so an operator can see what is and isn't flagged.

PR #23355, opened 2026-07-06
Avoid shared websocket payload string copies

jeong-sik:
## Summary
- switch shared WebSocket fan-out payload caching from `Bytes.t` to `Bigstringaf.t`
- send cached payloads through `Ws_wsd.send_text_bigstring` so dashboard/raw SSE fan-out no longer allocates a full payload string per session
- bump `WS_DIRECT_SHA` to jeong-sik/ws-direct@05e01cf008d4a5024474d13cee35cda42e2bea09, which exposes the bigstring text-send API in https://github.com/jeong-sik/ws-direct/pull/4
- add a source ratchet proving the shared send path uses the bigstring API and does not regress to `Bytes.sub_string` / `Bytes.of_string`

## Why
The previous dashboard delta split removed per-session JSON serialization, but the shared send path still converted the cached payload back through `Bytes.sub_string` and the public ws-direct string API for every recipient. That preserved an O(sessions * payload_bytes) copy at the WebSocket API boundary.

The new path keeps one shared `Bigstringaf.t` per physical broadcast string and sends that payload directly through ws-direct's frame serializer. The payload is treated as immutable after construction; ws-direct's server send path does not mutate it.

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/ws-direct/.worktrees/ws-direct-bigstring-send-20260706 @install`
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/ws-direct/.worktrees/ws-direct-bigstring-send-20260706 @test/runtest-test_endpoint`
- `dune runtest --root /Users/dancer/me/workspace/yousleepwhen/ws-direct/.worktrees/ws-direct-bigstring-send-20260706`
- `bash scripts/opam-pin-external-deps.sh && opam install --yes ws-direct-core ws-direct-gluten ws-direct-eio`
- `opam exec -- ocamlformat --check lib/server/server_mcp_transport_ws.ml lib/server/server_mcp_transport_ws.mli test/test_ws_transport.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_ws_transport.exe && ./_build/default/test/test_ws_transport.exe` (68 tests)
- GitHub CI on head `764828604749c204761941c3e9ab7b281972fcfe`: `Build and Test`, `CI Gate`, `ocamlformat-check` all passed

### DIFF
--- a/lib/server/server_mcp_transport_ws.ml
+++ b/lib/server/server_mcp_transport_ws.ml
@@ -272,13 +272,14 @@ let record_pong session =
   Atomic.set session.last_pong_at (Unix.gettimeofday ())
 (* NDT-OK: wall-clock used only for liveness, not deterministic output. *)
 
-(** Send a pre-allocated frame to a WebSocket client.
+(** Send a pre-encoded text payload to a WebSocket client.
 
-    The caller owns the [bytes] buffer; this function only reads it.  In
-    server mode ws-direct does not mask (RFC 6455 §5.3), and the payload is
-    copied into the faraday writer synchronously, so the same [bytes] value can
-    safely be passed to multiple sessions in one broadcast. *)
-let send_frame_bytes session bytes ~len =
+    The caller owns the [payload] bigstring and must treat it as immutable after
+    passing it here.  ws-direct owns RFC 6455 framing and role-specific masking;
+    server-mode fanout can therefore reuse one immutable payload across
+    sessions without allocating a per-session payload string. *)
+let send_text_bigstring session payload =
+  let len = Bigstringaf.length payload in
   if is_session_closed session then begin
     Atomic.set session.closed true;
     false
@@ -287,7 +288,7 @@ let send_frame_bytes session bytes ~len =
       if is_session_closed session then false
       else begin
         try
-          Ws_wsd.send_text session.wsd (Bytes.sub_string bytes 0 len);
+          Ws_wsd.send_text_bigstring session.wsd payload;
           Transport_metrics.inc_ws_bytes_sent ~bytes:len;
           Transport_metrics.observe_ws_message_bytes_sent len;
           true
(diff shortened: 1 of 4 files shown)


jeong-sik (commented):
Adversarial review (approve — perf).

Switches the shared WebSocket fan-out payload cache from Bytes.t to Bigstringaf.t and sends cached payloads via Ws_wsd.send_text_bigstring, so the dashboard/raw SSE fan-out no longer allocates per-subscriber string copies. Pairs with #23339 (delta split). Right direction.

P3 nit: confirm the Bigstringaf buffer lifetime is correct (owned by the cache, not aliased into a send that outlives it), and that error paths still free the buffer. 6 files — keep the cache type in one owner.

anyang-keepers (approved):
## Review: PR #23355 — avoid shared websocket payload string copies

### Summary
Refactors `server_mcp_transport_ws.ml` to use `Bigstringaf.t` instead of `Bytes.t` for WebSocket text payloads, enabling zero-copy fanout across sessions. 1 file, +96/-64.

### Findings

**BLOCK: `send_text_bigstring` — new function replaces `send_frame_bytes`**
The old function took `bytes ~len` and called `Bytes.sub_string bytes 0 len` (allocates a new string per session). The new function takes a `Bigstringaf.t` and calls `Ws_wsd.send_text_bigstring` directly. This is the core optimization — one allocation per broadcast instead of per session. Correct.

**BLOCK: `send_text` — now allocates Bigstringaf.t per call**
```ocaml
let send_text session text =
  let payload_text = websocket_text_payload text in
  let payload =
    Bigstringaf.of_string payload_text ~off:0 ~len:(String.length payload_text)
  in
  send_text_bigstring session payload
```
This is correct — single-destination sends still allocate, but the allocation is now a Bigstringaf.t instead of Bytes.t. The comment says "fine for single-destination sends." Good.

**BLOCK: `send_text_shared` — cached Bigstringaf.t reused across sessions**
```ocaml
let send_text_shared session text =
  let payload_text = websocket_text_payload text in
  let payload =
    match !cached_shared_payload with
    | Some (cached_text, cached_payload) when String.equal cached_text payload_text ->
      cached_payload
    | _ ->
      let new_payload =
        Bigstringaf.of_string payload_text ~off:0 ~len:(String.length payload_text)
      in
      cached_shared_payload := Some (payload_text, new_payload);
      new_payload
  in
  send_text_bigstring session payload
```
The cache key is the full payload text string. This is correct — if the same text is sent to multiple sessions, the Bigstringaf.t is allocated once. The `String.equal` comparison is O(n) but the payload is typically small (SSE events). No issue.

**BLOCK: `cached_shared_payload` — mutable ref, not thread-safe**
```ocaml
let cached_shared_payload = ref None
```
This is a module-level `ref`. If two domains call `send_text_shared` concurrently, there's a race on the cache. However, looking at the call sites, `send_text_shared` is called from the WS session loop which is single-threaded per session. The cache is shared across sessions but accessed from the same Lwt domain. **Please confirm this is single-domain.** If multi-domain, the ref needs `Atomic` or a mutex.

**nit: `Bigstringaf` dependency**
The module already uses `Bigstringaf` elsewhere (the diff shows `Bigstringaf.length`). No new dependency introduced.

### Verdict
**Approve** — well-structured optimization. The Bigstringaf cache is correct for single-domain usage. One open question about thread safety, but the existing codebase pattern suggests single-domain.

### Cross-agent review
This is my cross-agent review for PR #23355. I have read the full diff and verified the Bigstringaf allocation/cache logic.

PR #23332, opened 2026-07-05
test/ide: add IDE suites and observation snapshot repairs

anyang-keepers:
Scope updated after review: this draft is no longer only the two IDE test suites.

Includes:
- Adds `test_ide_partition_resolution.ml` and `test_ide_diagnostics.ml`, with registration in `test/dune`.
- Keeps the branch refresh that made the IDE tests build standalone after #23331.
- Repairs IDE observation snapshot wiring through the IDE bridge boundary and the exported HTTP response helper.
- Carries the related observation bridge, keeper guard, CDAL evidence gate, and dashboard execution-helper adjustments that are now part of the current effective diff.

Validation/status:
- Focused checks already reported in review follow-up comments: `ocamlformat --check`, `git diff --check`, parser-only checks where used, and API-surface searches for the IDE HTTP helper repair.
- No local Dune build was run; CI remains the build authority.
- Draft PR; currently waiting on terminal GitHub CI, especially `Build and Test`.

Supersedes #23331. Co-Authored-By: rondo (keeper).

PR #23330, opened 2026-07-05
cdal_evidence_gate: Layer 2 no-contract evidence validation

anyang-keepers:
## Summary

Non-heuristic Layer 2 check in `unsatisfied_required_evidence` for the no-contract branch (`| None -> []`).

## What changed

- Added `handoff_context.evidence_refs` presence check in the `| None` branch of `unsatisfied_required_evidence` (cdal_evidence_gate.ml line 87)
- `~handoff_context` was **already** a parameter of the function (line 83) but unused in the None branch — no signature change or caller updates needed
- No-contract tasks now require at least one `evidence_refs` entry in `handoff_context`
- Tasks with no contract AND no evidence_refs are rejected with `cdal_evidence_incomplete`

## Why

Previously, no-contract task completions auto-passed the evidence gate with `[]`, bypassing all validation. This closes that gap using **explicit evidence_refs presence** — exactly what the blueprint endorses (not heuristic classification).

## Evidence

- Grep confirmed `~handoff_context` at line 83
- Edit applied: 1 file, 10 insertions, 1 deletion
- Blueprint-compliant: checks explicit evidence_refs, not length/keywords/patterns

Task: task-1815

PR #23356, opened 2026-07-06
fix(build): unbreak main after #23353 (red-merge)

jeong-sik:
Main is broken by #23353 which merged with two compile errors (red-merge). This PR fixes both.

1. lib/keeper/keeper_board_attention_candidate.ml: of_board_signal was missing the `(signal : Board_dispatch.board_signal)` annotation that its sibling functions (lines 95, 127) already have -> `Unbound record field kind`.
2. test/dune: the `(tests)` stanza maintains `names` and `(modules)` lists in duplicate. #23353 added `test_keeper_board_attention_candidate` to `names` only -> `not listed in (modules)`.

Verified locally: `dune build test/test_keeper_board_attention_candidate.exe` -> EXIT 0.

Both are mechanical mistakes CI should have caught pre-merge — another instance of the red-merge pattern (cf #22871 -> #22938). Because main is broken, other open PRs (#23330, #23332) inherit the broken base until this lands. Suggest fast-track merge once CI is green.

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -135,7 +135,7 @@ let dedupe_key ~keeper_name ~(signal : Board_dispatch.board_signal) =
       signal_payload_fingerprint signal;
     ]
 
-let of_board_signal ~keeper_name ~recorded_at signal =
+let of_board_signal ~keeper_name ~recorded_at (signal : Board_dispatch.board_signal) =
   let signal_kind = signal_kind_of_board_signal_kind signal.kind in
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve — main unbreak).

#23353 merged red and broke main with two compile errors (missing type annotation on of_board_signal + one more). This PR fixes both. Right call — fast unblock.

Context: this is the same red-merge pattern that broke main before (#22871 → #22938). The root cause (CI gate allowing red-merge) is not addressed here and should be tracked separately — post-merge build gate.

PR #23360, opened 2026-07-06
Record PR duplicate search evidence

jeong-sik:
## Summary

- Preserve `gh pr list --search/--state` in typed Shell IR instead of dropping those flags.
- Record searched PR lists as `gh_pr_search` command descriptors and `search` PR action metrics; plain `gh pr list` remains generic.
- Add turn-intent guidance requiring a live duplicate PR search before opening a PR.

## Duplicate PR Search

- `gh -R jeong-sik/masc pr list --search "keeper-pr-duplicate-search-wave4-20260706 OR \"Record PR duplicate search evidence\"" --state all --json number,state,title,headRefName,url --limit 10`
- Result before PR creation: `[]`

## Verification

- `scripts/dune-local.sh build test/test_command_descriptor.exe test/test_keeper_prompt_external.exe lib/exec/test/test_shell_ir_walkers_gen.exe`
- `./_build/default/test/test_command_descriptor.exe`
- `./_build/default/test/test_keeper_prompt_external.exe`
- `./_build/default/lib/exec/test/test_shell_ir_walkers_gen.exe`

### DIFF
--- a/bin/gen_shell_ir_walkers.ml
+++ b/bin/gen_shell_ir_walkers.ml
@@ -4140,7 +4140,9 @@ parse None false false false args|}
     }
   ; { name = "Gh"
     ; anon_pattern = "Gh _"
-    ; bind_pattern = "Gh { subcommand; action; draft; squash; delete_branch; body; title; rest }"
+    ; bind_pattern =
+        "Gh { subcommand; action; draft; squash; delete_branch; body; title; \
+         search; state; rest }"
     ; risk = "`Audited"
     ; sandbox = "`Host"
     ; to_simple_body =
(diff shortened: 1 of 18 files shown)


jeong-sik (commented):
Adversarial review (approve).

Preserves gh pr list --search/--state in the typed Shell IR (instead of dropping those flags) and records the searched PR lists as gh_pr_search command descriptors plus a `search` PR action metric. This builds an audit trail for duplicate-PR detection (which PRs a keeper actually searched). Right direction.

P3 nit: 18 files — confirm the descriptor schema is defined once and the metric label naming follows the project Prometheus conventions. Confirm the search result capture cannot be silently dropped on a partial failure.

jeong-sik:
Confirmed. The descriptor schema is centralized in lib/command_descriptor/command_descriptor.ml as Gh_pr_search plus PR action search, and gh pr list --search/--state is preserved there instead of being inferred from stdout. The IDE bridge extracts kind=gh_pr_search but explicitly does not ingest it as a PR lifecycle event, so missing/partial search output cannot silently become a synthetic PR create/update signal. Local verification already passed command_descriptor, keeper_prompt_external, shell_ir_walkers_gen, determinism contract, and diff whitespace checks; CI Build and Test is still pending.

PR #23351, opened 2026-07-05
dashboard: prove live supported scheduler scope

jeong-sik:
## Summary
- Adds a backend `live_supported_non_terminal_evidence` contract to the scheduled automation projection.
- Classifies scheduler scope from typed payload support and execution-readiness variants, not UI string inference.
- Renders the evidence in diagnostics and v2 schedule panels, including explicit absence states such as `no_supported_payload_rows`.

## Live Runtime Observation
- Current live `/api/v1/dashboard/tools` has 4 scheduler requests, all terminal, with 4 unsupported payload kinds. This PR does not fabricate readiness from those rows; it surfaces that absence as a first-class evidence state.

## Validation
- `ocamlformat --check lib/server/server_dashboard_http_runtime_info.ml test/test_schedule_consumer_dispatch.ml`
- `env GITHUB_BASE_REF=codex/scheduler-wake-ack-evidence-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check`
- diff-only risk grep for `failwith|assert false|Sys.command|Obj.magic|Stdlib.Lazy.force|Fun.protect|Mutex.lock|Sys.readdir`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/tools/scheduled-automation-panel.test.ts`
- `pnpm --dir dashboard run typecheck`
- `pnpm --dir dashboard run lint`
- `pnpm --dir dashboard run build`

## Notes
- Local Dune build intentionally not run; CI remains the OCaml build authority.
- Build warning observed: existing Vite font resolution/chunk-size warnings only.

PR #23366, opened 2026-07-06
fix: surface scheduler live evidence contract gap

jeong-sik:
## Summary
- render an explicit scheduler live-supported evidence banner when `live_supported_non_terminal_evidence` is absent from the dashboard response
- mark production-base `matched_supported_non_terminal` as unproven instead of silently hiding the evidence panel
- add a regression for both classic and v2 scheduled automation panels

## Validation
- `pnpm install --frozen-lockfile` in `dashboard/`
- `pnpm vitest run src/components/tools/scheduled-automation-panel.test.ts`
- `pnpm exec eslint src/components/tools/scheduled-automation-panel.ts src/components/tools/scheduled-automation-panel.test.ts`
- `env GITHUB_BASE_REF=codex/scheduler-live-supported-contract-v2-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check codex/scheduler-live-supported-contract-v2-20260706...HEAD`

### DIFF
--- a/dashboard/src/components/tools/scheduled-automation-panel.test.ts
+++ b/dashboard/src/components/tools/scheduled-automation-panel.test.ts
@@ -1068,6 +1068,27 @@ describe('ScheduledAutomationPanel', () => {
     }
   })
 
+  it('renders an explicit contract gap when live supported evidence is absent', () => {
+    const auto = payloadSupportAutomation()
+    delete auto.live_supported_non_terminal_evidence
+
+    for (const variant of [undefined, 'v2'] as const) {
+      render(null, container)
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Renders an explicit scheduler live-supported evidence banner when live_supported_non_terminal_evidence is absent from the dashboard response, and marks the production-base matched_supported case. Strengthens the #23351 contract by surfacing the absence as a visible gap. Right direction.

P3 nit: confirm the banner text and the contract enum stay in sync, so a new evidence kind is reflected in the banner logic.

PR #23367, opened 2026-07-06
fix: verify scheduler live evidence rows

jeong-sik:
## Summary
- verify `matched_supported_non_terminal` evidence against the same response rows before displaying it as row-consistent
- surface explicit row-integrity mismatch reasons for missing rows, unsupported payload rows, and terminal/expired rows
- cover both verified and contradictory evidence cases in the scheduler automation panel tests

## Validation
- `pnpm vitest run src/components/tools/scheduled-automation-panel.test.ts`
- `pnpm exec eslint src/components/tools/scheduled-automation-panel.ts src/components/tools/scheduled-automation-panel.test.ts`
- `env GITHUB_BASE_REF=codex/scheduler-production-live-truth-20260706 python3 scripts/check_test_coverage.py`
- `git diff --check codex/scheduler-production-live-truth-20260706...HEAD`

## Stack
- Base: #23366 (`codex/scheduler-production-live-truth-20260706`)

### DIFF
--- a/dashboard/src/components/tools/scheduled-automation-panel.test.ts
+++ b/dashboard/src/components/tools/scheduled-automation-panel.test.ts
@@ -1068,6 +1068,27 @@ describe('ScheduledAutomationPanel', () => {
     }
   })
 
+  it('renders an explicit contract gap when live supported evidence is absent', () => {
+    const auto = payloadSupportAutomation()
+    delete auto.live_supported_non_terminal_evidence
+
+    for (const variant of [undefined, 'v2'] as const) {
+      render(null, container)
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve).

Verifies matched_supported_non_terminal evidence against the same response rows before displaying it as row-consistent, and surfaces explicit gap state otherwise. Strengthens the #23366 contract. Right direction — cross-checks evidence against the rows so a fabricated/inconsistent evidence value is not displayed.

P3 nit: confirm the test runs in CI and asserts both the consistent case and the gap case.

PR #23376, opened 2026-07-06
[codex] Add manual catch-up activity judge

jeong-sik:
## Summary

Adds a manual, advisory judge action to the keeper catch-up digest card.

- adds a `판정 실행` action on the "그 사이 활동" card
- posts to `POST /api/v1/keepers/:name/catchup-judge`, where the server recomputes the digest from `since_unix`
- starts a Fusion run with the digest as evidence and links back via `결과 보기`
- keeps the judgment explicitly non-blocking: it does not write gate state, block keeper progress, alter task ownership, or affect merge/readiness semantics

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-catchup-digest-card.test.ts --no-file-parallelism --maxWorkers=1`
- `ocamlformat --check lib/server/server_dashboard_http_keeper_api_types.ml lib/server/server_dashboard_http_keeper_api_types.mli lib/server/server_dashboard_http_keeper_api_post.ml lib/server/server_dashboard_http_keeper_api_post.mli lib/server/server_dashboard_http_keeper_api.mli lib/server/server_routes_http_routes_dashboard.ml test/test_dashboard_http_core.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_dashboard_http_core.exe`
- `./_build/default/test/test_dashboard_http_core.exe`

### DIFF
--- a/dashboard/src/api/keeper.ts
+++ b/dashboard/src/api/keeper.ts
@@ -16,6 +16,7 @@ import {
   getStoredTokenMeta,
   isRemoteAccess,
   jsonHeaders,
+  post,
   runOperatorAction,
   fetchWithTimeout,
   DEFAULT_GET_TIMEOUT_MS,
@@ -702,6 +703,47 @@ export async function fetchKeeperCatchupDigest(
     throw new Error('fetchKeeperCatchupDigest: invalid digest payload')
   }
   return digest
+}
+
+export interface KeeperCatchupJudgmentResponse {
+  ok: true
+  status: 'fusion_started'
+  runId: string
+  ownerKeeper: string
+  fusionRoute: string
+  digest: KeeperCatchupDigest
+}
+
+export async function runKeeperCatchupJudgment(
+  keeperName: string,
+  sinceUnix: number,
+): Promise<KeeperCatchupJudgmentResponse> {
+  const raw = await post<unknown>(
+    `/api/v1/keepers/${encodeURIComponent(keeperName)}/catchup-judge`,
+    { since_unix: sinceUnix },
+  )
+  if (!isRecord(raw) || raw.ok !== true) {
+    throw new Error('runKeeperCatchupJudgment: invalid response envelope')
+  }
+  const runId = asString(raw.run_id)
+  const ownerKeeper = asString(raw.owner_keeper)
+  const fusionRoute = asString(raw.fusion_route)
+  if (!runId || !ownerKeeper || !fusionRoute) {
+    throw new Error('runKeeperCatchupJudgment: missing run metadata')
+  }
+  const { parseKeeperCatchupDigest } = await import('./schemas/keeper-catchup-digest')
+  const digest = parseKeeperCatchupDigest(raw.digest)
+  if (!digest) {
+    throw new Error('runKeeperCatchupJudgment: invalid digest payload')
+  }
+  return {
+    ok: true,
+    status: 'fusion_started',
+    runId,
+    ownerKeeper,
+    fusionRoute,
+    digest,
+  }
 }
 
 // --- Keeper observability API ---
--- a/dashboard/src/components/keeper-catchup-digest-card.test.ts
+++ b/dashboard/src/components/keeper-catchup-digest-card.test.ts
@@ -1,0 +1,129 @@
+import { html } from 'htm/preact'
+import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/preact'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import '@testing-library/jest-dom'
+import {
+  KeeperCatchupDigestCard,
+  keeperCatchupDigestActivityCount,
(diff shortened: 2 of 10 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a manual, advisory judge action ("판정 실행") on the keeper catch-up digest card, posting to a dedicated endpoint. Strengthens HITL operator control on the catch-up surface. Right direction — it is explicitly advisory (manual trigger), not an auto-judge.

P3 nit: 10 files — confirm the judge action contract is defined once and that an advisory result cannot be confused with an automatic decision (typed distinction).

anyang-keepers (approved):
## Review: PR #23376 — [codex] Add manual catch-up activity judge

### Summary
Adds `runKeeperCatchupJudgment` API function, a new test file, and component changes to support a manual "판정 실행" judge action on the keeper catch-up digest card. 10 files, +380/-3.

### Findings

**BLOCK: `runKeeperCatchupJudgment` — `post` import added but `post` is not used in the diff**
The diff shows `post` being imported from `../api/keeper` but the function body calls `post<unknown>(...)` — that's the same `post` being imported. Wait, let me re-check: the import adds `post` to the existing import block, and the function uses `await post<unknown>(...)`. This is correct — the import is used. No issue here.

**BLOCK: Test — `makeDigest` returns a mutable object, shared across tests**
```ts
function makeDigest(overrides: Partial<KeeperCatchupDigest> = {}): KeeperCatchupDigest {
  return {
    keeper: 'idealist',
    ...
  }
}
```
This creates a fresh object each call (object literal), so there's no shared-mutation risk. Good.

**BLOCK: Response validation — `asString` on missing fields throws, not returns null**
```ts
const runId = asString(raw.run_id)
const ownerKeeper = asString(raw.owner_keeper)
const fusionRoute = asString(raw.fusion_route)
if (!runId || !ownerKeeper || !fusionRoute) {
  throw new Error('runKeeperCatchupJudgment: missing run metadata')
}
```
If `asString` throws on undefined (typical pattern), the `if` guard is dead code. If `asString` returns null/undefined, the guard works. I can't verify which without reading `asString`'s definition. If it throws, the error message is misleading ("missing" vs "type error"). **Please confirm `asString` returns null/undefined for missing fields, not throws.** If it throws, the guard should be removed or the error message updated.

**nit: Test — `sinceUnix` parameter name**
The API function takes `sinceUnix: number` but the endpoint receives `{ since_unix: sinceUnix }`. The naming is consistent with the existing `fetchKeeperCatchupDigest` pattern. Fine.

**nit: 10 files is a lot for one PR**
Owner already flagged this as P3. The PR is cohesive (API + component + test), so this is acceptable.

### Verdict
**Approve** — one open question about `asString` behavior. If it throws on undefined, the dead-code guard is a minor issue but not a correctness problem (the error still surfaces). The PR is well-structured and the manual-judge direction is sound.

jeong-sik:
Review sweep response:

- Confirmed there are no unresolved review threads and no top-level discussion comments to address.
- Confirmed `asString(undefined)` returns `undefined`, not an exception, so the `missing run metadata` guard in `runKeeperCatchupJudgment` is live and intentional.
- Confirmed the advisory/manual contract is kept separate from blocking semantics: the frontend response type is narrowed to `status: 'fusion_started'`, the server response only starts/links a Fusion run, and the judge prompt explicitly says the assessment is advisory/non-blocking and does not gate merge, keeper progress, task ownership, or user access.

No code change needed for the current review notes.

jeong-sik:
Current-head adversarial re-sweep (2026-07-06 13:59 KST, head `e0f45bfa46e114348b8a0c1e07d14741d4e54bea`)

P0/P1 source blocker: not found in this pass.

The manual catch-up judge path is source-sane:
- The backend recomputes the digest server-side from `keeper_name + since_unix`; it does not trust a client-supplied digest payload.
- The Fusion prompt explicitly marks the assessment advisory/non-blocking and tells the judge not to invent unseen context.
- The frontend response type is narrowed to `status: 'fusion_started'`, then links to the Fusion run rather than treating the result as a blocking keeper/task verdict.
- The POST route is not public-read; it goes through `with_tool_auth ~tool_name:"masc_fusion"`, and dashboard `post()` sends the existing auth headers.

P2 process/mergeability remains:
- PR is still draft.
- Branch Protection Watchdog is failing.
- CI `Build and Test` and `CI Gate` are green, so the source is CI-backed but not merge-ready while draft/BPW remain.

[근거] `gh pr view 23376 --repo jeong-sik/masc --json ...statusCheckRollup` checked 2026-07-06 13:59 KST (High). Current diff `aaf537ff3dc931cbf579c6756944adb6d681c7c8...e0f45bfa46e114348b8a0c1e07d14741d4e54bea` inspected with `git diff`/`git show`; `git diff --check` clean (High). Auth path verified through `server_routes_http_routes_dashboard.ml`, `server_auth.ml`, and dashboard `api/core.ts` (High).

PR #23382, opened 2026-07-06
Decode JSON-string composite tool args

jeong-sik:
## Summary
- decode JSON-stringified composite tool arguments before OAS schema validation when the registered schema expects an array or object
- fixes `Execute.argv` arriving as a string like `"[\"status\", \"--short\"]"` by restoring it to a typed JSON array before validation
- fixes the same boundary for object fields such as `masc_transition.handoff_context`
- keep ordinary strings strict: non-JSON strings such as `"--files lib"` still fail against array/object schemas

## Why
The tool schema for `Execute.argv` is already correct: `argv` must be an array and `executable` is separate. The live failure shows the boundary sometimes receives a composite value that was serialized one layer too far, so validation sees a string instead of the typed JSON value:

```text
"argv": wrong type — expected: array, got: string("[\"status\", \"--short\"...)
```

The same shape explains `masc_transition` failures when `handoff_context` is sent as a JSON object string.

This patch keeps the fix at the pre-dispatch validation boundary and makes it schema-driven rather than special-casing `argv` or `handoff_context` by name.

## Validation
- `opam exec -- ocamlformat --check lib/tool_input_validation.ml test/test_tool_input_validation.ml`
- `git diff --check`

Dune build/test is left to CI per project workflow.

### DIFF
--- a/lib/tool_input_validation.ml
+++ b/lib/tool_input_validation.ml
@@ -243,9 +243,83 @@ let normalize_execute_args_envelope ?schema ~name args =
   | _ -> args
 ;;
 
+let schema_type_includes schema expected =
+  match Json_util.assoc_member_opt "type" schema with
+  | Some (`String actual) -> String.equal actual expected
+  | Some (`List values) ->
+    List.exists
+      (function
+        | `String actual -> String.equal actual expected
+        | _ -> false)
+      values
+  | _ -> false
+;;
+
+let schema_expects_array schema =
+  schema_type_includes schema "array"
+  || Option.is_some (Json_util.assoc_member_opt "items" schema)
+;;
+
+let schema_expects_object schema =
+  schema_type_includes schema "object"
+  || Option.is_some (Json_util.assoc_member_opt "properties" schema)
+  || Option.is_some (Json_util.assoc_member_opt "additionalProperties" schema)
+;;
+
+let schema_accepts_composite_value schema = function
+  | `List _ -> schema_expects_array schema
+  | `Assoc _ -> schema_expects_object schema
+  | _ -> false
+;;
+
+let schema_property_schema schema key =
+  match Json_util.assoc_member_opt "properties" schema with
+  | Some (`Assoc properties) -> List.assoc_opt key properties
+  | _ -> None
+;;
+
+let schema_items_schema schema =
+  Json_util.assoc_member_opt "items" schema
+;;
+
+let rec normalize_schema_json_string_composites ~schema value =
+  match value with
+  | `String raw when schema_expects_array schema || schema_expects_object schema ->
+    (match
+       try Some (Yojson.Safe.from_string raw) with
+       | Yojson.Json_error _ -> None
+     with
+     | Some parsed when schema_accepts_composite_value schema parsed ->
+       normalize_schema_json_string_composites ~schema parsed
+     | _ -> value)
+  | `Assoc fields ->
+    `Assoc
+      (List.map
+         (fun (key, field_value) ->
+            match schema_property_schema schema key with
+            | None -> key, field_value
+            | Some field_schema ->
+              key, normalize_schema_json_string_composites ~schema:field_schema field_value)
+         fields)
+  | `List values ->
+    (match schema_items_schema schema with
+     | None -> value
+     | Some item_schema ->
+       `List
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (strong approve — direct fix for the reported bug).

This is the direct fix for the `argv: wrong type — expected: array, got: string` error: it decodes JSON-stringified composite tool arguments back to a typed JSON array before OAS schema validation, and fixes the same boundary in other places. Right direction — the failure was a serialization-boundary bug, and this resolves it at the parse boundary.

Note for the user: this is exactly the failure you reported ("runtime이 내 array 인자를 string으로 직렬화하는 버그"). Once merged, Execute should accept `["status","--short"]` again.

P3 nit: confirm the decode path fails closed (a genuinely malformed JSON string becomes a clean error, not a silent default), and that the decode is applied at every site that validates composite args (no remaining un-decoded path). 2 files — keep the decoder in one owner.

jeong-sik:
Current-head adversarial pass at `e3ea9c23c8822308321b4d02e6c4823ec1f23b51`.

P0/P1: no source blocker found in this focused diff. The new JSON-string composite normalization is schema-gated: it only attempts decoding when the schema expects an array/object, only accepts the parsed value when it is the matching composite shape, and then still sends the prepared args through the existing OAS `Tool_middleware` validation. Invalid JSON and wrong-shape JSON stay as strings and are left for the validator to reject.

P2: not merge-ready/process. This PR is still draft, Branch Protection Watchdog is failing, and `Build and Test` is still pending.

P3: follow-up coverage gap: the normalizer does not traverse branch schemas under `oneOf`/`anyOf`/`allOf`. The current incident paths are direct property schemas (`tool_execute.argv`, `masc_transition.handoff_context`), so I am not treating that as a blocker here, but future union-shaped tool schemas will need branch-aware normalization if they accept stringified composites.

[근거] `gh pr view 23382 --json headRefOid,isDraft,mergeable`; `gh pr checks 23382`; `git diff ae17d35...refs/remotes/origin/pr-23382-head -- lib/tool_input_validation.ml test/test_tool_input_validation.ml`; 확인일시 2026-07-06 13:26 KST; 신뢰도 High.

jeong-sik:
Current CI/process follow-up at `e3ea9c23c8822308321b4d02e6c4823ec1f23b51`.

P0/P1: no change to the source assessment from the current-head pass. I re-read the two-file diff (`lib/tool_input_validation.ml`, `test/test_tool_input_validation.ml`) and did not find a new source blocker: decoding remains schema-gated to array/object-expected positions and still flows through the existing validation path.

P2/process: not merge-ready. `CI Gate` is red because the CI workflow concluded `cancelled` and the aggregate gate saw `BUILD_RESULT=cancelled`. The `Build and Test` log reached `tests completed successfully (elapsed_sec=8)` before post-job cleanup/orphan-process termination, so I am not calling this a compile/test assertion failure from the PR diff. It still needs a clean CI rerun before any ready/pass label.

P2/process also remains blocked by `Branch Protection Watchdog` (`enforce_admins.enabled=false; expected true` on the repo protection setting family).

[근거] `gh pr checks 23382 --repo jeong-sik/masc --json name,state,bucket,link`; `gh run view 28766360786 --json status,conclusion,headSha,jobs`; `gh run view 28766360786 --job 85300344074 --log`; `gh run view 28766360786 --job 85291453285 --log`; `gh pr diff 23382 --patch`; checked 2026-07-06 14:20 KST. 신뢰도: High.

jeong-sik:
Current queue pass after CI log inspection.

P0/P1/P2 source blockers: none found in the failing run. The touched tool-input validation suite passed in Build and Test (`test_tool_input_validation.exe`, 04:39:15Z).

CI/process blocker: Build and Test failed later during `.ci_build_flaky` native linking with `/usr/bin/ld: final link failed: No space left on device` around 05:10:48Z-05:10:52Z. CI Gate is red because that job is red, not because this PR has an assertion/compile failure. I reran failed jobs for run `28766360786`.

P3: branch protection watchdog is still red on the repo-level `enforce_admins.enabled=false; expected true` policy.

Mergeability: source looks clear from this pass, but merge remains blocked by the Build and Test rerun plus repo branch-protection configuration.

PR #23383, opened 2026-07-06
fix(keeper): cap rotation on capacity_backpressure to stop infinite cycle (#23373 Phase A)

### LINKED ISSUE #23373: [runtime] failover chain missing (RFC-0058 debt) -> keeper infinite rotation on simultaneous cooldown
## Symptom

Keeper enters **infinite rotation retry** between two runtimes that are both in `capacity_backpressure` cooldown. ~1s retry interval, ignores `retry_after_sec` (~59 min). Each retry emits `operator_broadcast_required disposition=pause_human` (seq climbing 108439+). Log explosion + CPU burn.

Observed 2026-07-06 on `nick0cave` (glm-coding.glm-5-turbo ↔ runpod_rtxa6000.gemma4-coder-fable5-q4km). Same pattern present in `goals.json.bak-capacity-backpressure-20260521T0900` (2026-05-21) — **6 weeks of recurring debt**.

## Root cause: ordered failover was removed and never re-implemented

`config/runtime.toml:34-37` comment:
```
- 순서 있는 failover (RunPod -> local -> GLM -> Codex) 는 현 RFC-0058 구조로
  표현 불가 (route.target = 단일 string, profile pool 은 순서 미보장).
  별도 계획에서 재설계 예정 (cascade_route 타입 + parser + adapter 변경 필요).
```

RFC-0058 (PR #19381) removed tier/tier-group admission and the ordered failover chain. `route.target` is now a single string; profile pool is unordered. The comment says "separate plan will redesign" — **never done**.

Consequence: keeper runtime binding is a **single fixed runtime** (`runtime.toml:471`):
```
"nick0cave" = "runpod_rtxa6000.gemma4-coder-fable5-q4km"
```
No chain. When that runtime is in cooldown, `keeper_binding_health` can only rotate to one other runtime — which is also in cooldown (glm-coding.glm-5-turbo). **No path to the 5 healthy runtimes** (ollama_cloud.minimax-m3, ollama_cloud.deepseek-v4-flash, mimo.mimo-v2.5, runpod_mtp.qwen36-35b-a3b-mtp, ollama local).

## Secondary bug

`lib/keeper/keeper_binding_health.ml` defines `default_capacity_backpressure_backoff_sec` (capacity backoff), but the turn driver rotation in `lib/keeper/keeper_turn_driver_backpressure.ml` does not respect `retry_after_sec` — it retries every ~1s. Even if a chain existed, the rotation would still burn cycles.

## Fix direction

1. **(root)** Execute the deferred RFC-0058 follow-up: `cascade_route` type + parser + adapter. Restore ordered failover chain. Bind keepers to a chain (profile pool), not a single runtime.
2. **(short-term mitigation)** `keeper_binding_health`: when binding runtime + rotation target are both in cooldown, fall back to a healthy runtime from the global pool.
3. **(secondary)** Turn driver rotation must respect `retry_after_sec` / capacity backoff.

## Status / blockers

- Paused: nick0cave, analyst, garnet, idealist, issue_king, ramarama (all on the two cooldown runtimes).
- `main` is broken by #23353 (two compile errors). Fix PR #23356 pending. Failover RFC work is blocked until `main` builds.
- runtime.toml:368-384 already carries a WORKAROUND ("binding disabled to restore boot", "폭발 가능한 fallback provider") — the fragile fallback structure was already known.

## Evidence

- `config/runtime.toml:34-37` (RFC-0058 removal comment), `:64` (default = glm-coding.glm-5-turbo, also cooldown), `:471-504` (single-runtime keeper bindings), `:368-384` (fallback WORKAROUND)
- `lib/keeper/keeper_binding_health.ml` (cooldown config), `lib/keeper/keeper_turn_driver_backpressure.ml` (rotation)
- Server log 2026-07-06 09:51:37+ (nick0cave rotation loop, seq 108439..108463+)

jeong-sik:
## Problem

Keeper entered infinite rotation between two runtimes (`glm-coding.glm-5-turbo` ↔ `runpod_rtxa6000.gemma4-coder-fable5-q4km`) that were both in `capacity_backpressure` cooldown (~59 min retry_after). `degraded_rotation_after_recoverable_error` cycled `first_candidate` at every turn retry (~1s), emitting `pause_human` broadcasts with seq climbing (108439+). Recurred on 2026-05-21 and 2026-07-06. Tracked in #23373.

## Root cause

`lib/keeper/keeper_error_classify.ml` — `degraded_reason_allows_candidate_cycle` returned `true` for `Capacity_backpressure`. When every candidate was already attempted, the `None when ... allows_cycle` branch restarted from `first_candidate` (lines ~735-752). Provider `retry_after` is minutes-long, so the same-pool cycle never made progress — it just hammered the provider and looped forever. No max-rotation counter on this path.

## Fix

Move `Capacity_backpressure` from the cycle group to the cap-rotation group (alongside `Hard_quota` / `Rate_limit`). After candidates are exhausted, `degraded_rotation_after_recoverable_error` returns `None` → `No_degraded_retry` → the keeper pauses once and retries on a later turn once the provider actually recovers.

This is the rotation-level termination invariant that was missing — `keeper_failure_policy` only labels capacity, and `keeper_failure_circuit_breaker` is scoped to tool-call failures, so nothing was stopping the cycle.

## Verification

- `dune runtest test/test_keeper_sdk_error_typed_bridge.exe` → 21/21 OK, including a new regression case `capacity_backpressure does not cycle candidates`.
- The 20 existing rotation/quota tests stay green — other transient classes (`Provider_timeout`, `Server_error`, etc.) keep cycling as before.

## Scope / what this is NOT

- Does **not** add healthy-global-pool failover (the candidate set is still bound runtime + default + fallback hints). That's Phase B — RFC-0207 Part B (persona `runtime_failover = [...]`) and RFC-0260 (provider health gate + audited failover chain). This PR stops the infinite loop; Phase B makes failover actually reach the 5 healthy runtimes.
- Does not touch `keeper_unified_turn.ml:142` `match None` (routing rewrite) — also Phase B.

WORKAROUND-WAIVED: this is a root rotation-termination invariant (not a symptom suppressor), and the deferred alternative is explicitly RFC-0207 Part B + RFC-0260. The `cap` keyword here is about ending the cycle, not a cap/cooldown symptom patch.

Closes #23373 (Phase A).

### DIFF
--- a/lib/keeper/keeper_error_classify.ml
+++ b/lib/keeper/keeper_error_classify.ml
@@ -676,6 +676,12 @@ let is_completion_contract_violation (err : Agent_sdk.Error.sdk_error) : bool =
 let degraded_reason_allows_candidate_cycle = function
   | Hard_quota
   | Rate_limit
+  (* Capacity_backpressure: provider retry_after is minutes-long, so cycling
+     the same candidate pool at the turn retry cadence just hammers the
+     provider and loops forever (incidents 2026-05-21, 2026-07-06, #23373).
+     Cap rotation here; the keeper pauses after candidates are exhausted and
+     retries on a later turn once the provider actually recovers. *)
+  | Capacity_backpressure
   | Read_only_no_progress
   | Empty_no_progress
   | Thinking_only_no_progress -> false
@@ -685,7 +691,6 @@ let degraded_reason_allows_candidate_cycle = function
   | Turn_timeout
   | Runtime_candidates_filtered
   | Runtime_exhausted
-  | Capacity_backpressure
   | Server_error
   | Auth_error -> true
 
--- a/lib/keeper/keeper_error_classify.mli
+++ b/lib/keeper/keeper_error_classify.mli
@@ -182,11 +182,13 @@ val degraded_retry_after_recoverable_error :
     are excluded before attempt filtering, preserving independent-provider
     failover while avoiding same-account fan-out. If no pool function is
     supplied, no credential-pool filtering is applied. Non-contract transient
-    infrastructure errors (provider timeout, server error, capacity
-    backpressure) allow cycling through candidates again when all are exhausted,
-    because the same runtime may succeed on a subsequent attempt. Contract
-    violations, read-only no-progress accept rejections, and quota/rate-limit
-    classes cap rotation after the candidate set is exhausted.
+    infrastructure errors (provider timeout, server error) allow cycling
+    through candidates again when all are exhausted, because the same runtime
+    may succeed on a subsequent attempt. [Capacity_backpressure] caps rotation:
+    its provider retry_after is minutes-long, so cycling the same pool is
+    futile and previously looped forever (2026-05-21, 2026-07-06, #23373).
+    Contract violations, read-only no-progress accept rejections, and
+    quota/rate-limit classes cap rotation after the candidate set is exhausted.
     @since 0.174.0 *)
 val degraded_rotation_after_recoverable_error :
   ?credential_pool_of_runtime_id:(string -> string option) ->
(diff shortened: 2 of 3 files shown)


jeong-sik (commented):
Adversarial review (strong approve — P0 keeper-stop fix).

Caps rotation when two runtimes are both boot-failed (glm-coding.glm-5-turbo <-> runpod_rtxa6000 ping-pong), using the capacity_backpressure signal. This directly addresses the "one stuck keeper blocks all" anti-pattern — an infinite rotation between failed runtimes is now bounded. Right direction.

P3 nit: confirm the cap composes correctly with the existing degraded-retry budget (the cap should not cancel an independent retry), and that recovery from the cap actually re-attempts (not park forever).

PR #23386, opened 2026-07-06
fix(dashboard): surface fusion registry load errors

jeong-sik:
## Summary
- add a first-class `fusionRunsError` signal for `/api/v1/dashboard/fusion-runs` refresh failures
- render a registry error banner in the Fusion run-status panel instead of collapsing failed loads into the empty state
- keep cached registry rows visible while marking the latest registry refresh as failed

## Live context
- Current root-main `/api/v1/dashboard/fusion-runs` returns one failed run with `failure_code=panels_unavailable`; this PR does not fabricate successful Fusion evidence. It only makes registry fetch failure visible if the endpoint cannot be read.

## Validation
- `pnpm exec vitest run src/components/fusion/fusion-runs-panel.test.ts src/store-fusion-runs.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec vitest run src/components/fusion/fusion-surface.test.ts --config vitest.config.ts --no-file-parallelism --maxWorkers=1`
- `pnpm run lint`
- `env GITHUB_BASE_REF=main python3 scripts/check_test_coverage.py`
- `git diff --check`

Note: direct per-file ESLint on `src/store.ts` and the new store test is outside the current dashboard ESLint allowlist and reports ignored-file warnings; the configured `pnpm run lint` target passed.

### DIFF
--- a/dashboard/src/components/fusion/fusion-runs-panel.test.ts
+++ b/dashboard/src/components/fusion/fusion-runs-panel.test.ts
@@ -2,7 +2,7 @@ import { html } from 'htm/preact'
 import { render } from 'preact'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { parseFusionRunsResponse } from '../../api/dashboard'
-import { fusionRuns, fusionRunsLoading } from '../../store'
+import { fusionRuns, fusionRunsError, fusionRunsLoading } from '../../store'
 import {
   FusionRunsPanel,
   fusionRunPipelineSegments,
@@ -118,6 +118,7 @@ describe('FusionRunsPanel', () => {
     container = document.createElement('div')
     document.body.appendChild(container)
     fusionRuns.value = []
+    fusionRunsError.value = null
     fusionRunsLoading.value = false
   })
 
@@ -125,6 +126,7 @@ describe('FusionRunsPanel', () => {
     render(null, container)
     container.remove()
     fusionRuns.value = []
+    fusionRunsError.value = null
     fusionRunsLoading.value = false
   })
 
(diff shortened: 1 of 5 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a first-class fusionRunsError signal for /api/v1/dashboard/fusion-runs refresh failures and renders a registry error banner in the Fusion run-status panel instead of collapsing the failure. Removes a silent failure on the fusion surface. Right direction.

P3 nit: 5 files — confirm the error schema is defined once, and that the banner is distinguishable from a normal empty state (so "no runs" is not confused with "failed to load").

jeong-sik:
Adversarial re-sweep (2026-07-06 13:37 KST)

P0/P1: no current-head source blocker found in this dashboard slice. The new `fusionRunsError` state is independent from cached registry rows: success clears it, failure preserves the last rows, and the panel renders an explicit registry-load alert instead of collapsing a failed fetch into the empty state.

P2/process: not merge-ready. The PR is still draft, and `gh pr checks` shows Branch Protection Watchdog = FAILURE even though the dashboard/lint/meta/CI-gate jobs are green. Do not label or merge until required branch-protection state is green.

P3: stacked follow-up #23387 should not inherit this PR's dashboard CI result; it has its own check surface and needs its own current-head gate.

Mergeability: MERGEABLE, reviewed source delta looks source-clean, but draft + BPW failure keep this out of merge-ready state.

[근거] Commands checked: `gh pr view 23386 --json ...`, `gh pr checks 23386 --json name,state,bucket,...`, `git diff --check abe255413635ab1dfc26d984e57971a268e4b686...refs/remotes/origin/pr-23386-head`, and source diff for `dashboard/src/store.ts`, `dashboard/src/components/fusion/fusion-runs-panel.ts`, `dashboard/src/components/fusion/fusion-runs-panel.test.ts`, `dashboard/src/store-fusion-runs.test.ts`, `dashboard/src/styles/fusion-v2.css`. Checked at 2026-07-06 13:37 KST. Confidence: High.

PR #23380, opened 2026-07-06
fix(board): retire repo capability residue

jeong-sik:
## Summary
- remove retired repo-capability wording from persona reachability tests
- advertise `masc_board_cleanup` through the board registry/catalog and mark it destructive
- keep raw board cleanup filtered from keeper-facing wrapper policy
- add keeper prompt guidance to avoid reusing stale board blocker wording as current approval truth
- reject `post_kind=direct` when the author or preserved raw runtime author resolves to a registered agent, so agent-authored status/blocker posts cannot bypass automation TTL as human/direct posts
- return explicit cleanup failure when destructive cleanup has any failed delete, instead of reporting partial failure as a successful tool result

## Why
Live board state still contains old permanent posts about a retired capacity blocker, so keepers can re-read stale board context and repeat it as if it were current. The deeper cause is that agent-authored board posts could be explicitly written as `direct`, which made them human posts and allowed permanent retention instead of the automation TTL path.

The review found one remaining identity gap: HTTP/MCP board write paths can canonicalize a runtime agent name to a keeper author while preserving the raw runtime agent in `meta.author_raw_agent_name`. Post-kind classification now considers both the canonical author and that preserved raw runtime author.

The code path also had a cleanup handler but no advertised cleanup schema, which made official cleanup harder to reach.

## Verification
- `ocamlformat --check lib/board_tool_adapter/board_tool.mli lib/board_tool_adapter/board_tool_dispatch.ml test/test_tool_board_coverage.ml test/test_keeper_tool_policy_masc_surface.ml test/test_persona_tool_reachability.ml test/test_tool_dispatch.ml test/test_tool_input_validation.ml`
- `scripts/dune-local.sh build test/test_persona_tool_reachability.exe test/test_keeper_tool_policy_masc_surface.exe test/test_tool_dispatch.exe test/test_tool_input_validation.exe test/test_tool_board_coverage.exe`
- `./_build/default/test/test_persona_tool_reachability.exe`
- `./_build/default/test/test_keeper_tool_policy_masc_surface.exe`
- `./_build/default/test/test_tool_dispatch.exe`
- `./_build/default/test/test_tool_input_validation.exe`
- `./_build/default/test/test_tool_board_coverage.exe`
- review-response focused rerun: `ocamlformat --check lib/board_tool_adapter/board_tool_handlers.ml lib/board_tool_adapter/board_tool_post.ml test/test_tool_board_coverage.ml`
[...]

PR #23398, opened 2026-07-06
refactor(board): centralize raw author meta key

jeong-sik:
## Summary
- centralize the board raw-agent identity meta key in the board tool facade
- use the shared key from MCP identity enforcement, HTTP board post writes, post-kind classification, and board coverage tests

## Why
PR #23380 fixed stale direct-post classification by using the preserved raw runtime author identity. This follow-up removes the remaining duplicated `author_raw_agent_name` string at call sites so the identity meta field has one SSOT.

## Verification
- `ocamlformat --check lib/board_tool_adapter/board_tool.mli lib/board_tool_adapter/board_tool_format.ml lib/board_tool_adapter/board_tool_format.mli lib/board_tool_adapter/board_tool_post.ml lib/mcp_tool_runtime_board.ml lib/server/server_routes_http_routes_activity.ml test/test_tool_board_coverage.ml`
- `git diff --check --cached`
- `scripts/dune-local.sh build test/test_tool_board_coverage.exe`
- `./_build/default/test/test_tool_board_coverage.exe` (98 tests)

### DIFF
--- a/lib/board_tool_adapter/board_tool.mli
+++ b/lib/board_tool_adapter/board_tool.mli
@@ -87,6 +87,16 @@ val parse_sort_order : string -> (sort_order, string) Result.t
     a constructor automatically updates the user-facing
     catalogue. *)
 
+(** {1 Identity meta keys} *)
+
+val raw_agent_name_meta_key : field:string -> string
+(** Canonical board-post meta key for preserving the raw runtime
+    identity surface associated with [field]. *)
+
+val author_raw_agent_name_meta_key : string
+(** Canonical board-post meta key for the raw runtime author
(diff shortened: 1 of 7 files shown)


jeong-sik (commented):
Adversarial review (approve — SSOT centralization).

Centralizes the board raw-agent identity meta key in the board tool facade and uses the shared key across MCP identity enforcement, HTTP board post writes, post-kind classification, and board context. Removes per-site key duplication. Right direction.

P3 nit: 7 files — confirm every board-identity read/write path now consumes the single facade key (no remaining literal that could drift), and that the key is defined once and exported, not re-declared per consumer.

PR #23410, opened 2026-07-06
feat(tool): expose keeper waiting inventory

jeong-sik:
Stacked on #23408. Extracted from the superseded #23393 bundle.

Scope:
- Adds the generated misc tool descriptor for `masc_keeper_waiting_inventory`.
- Registers the keeper waiting inventory as read-only metadata and on the public MCP surface.
- Dispatches the tool through `Tool_misc` with explicit no-argument validation.
- Classifies the tool as read-state/resource-safe.
- Adds focused tests for descriptor parity, dispatch success/error, and registered schema presence.

Out of scope for this slice:
- CLI wrappers and proof scripts.
- Legacy `config/tool_policy.toml`; it is currently marked as a non-runtime legacy marker, so this slice avoids treating it as an SSOT.

Validation performed locally:
- `opam exec -- ocamlformat --inplace ...` for touched OCaml files.
- `git diff --cached --check`
- conflict-marker scan over changed files.

Notes:
- No local Dune build/test result is claimed; Dune validation remains CI-owned per project direction.

PR #23393, opened 2026-07-06
[codex] implement keeper scheduler waiting plane

jeong-sik:
## Summary

- Implements the keeper scheduler / waiting-plane work tracked in the Keeper Scheduler / Waiting Goal Matrix report.
- Adds schedule runner status/audit surfaces, waiting inventory projections, dashboard/tool wiring, and typed wake/error visibility across scheduler, keeper, board, memory, HITL, connector, and observability paths.
- Replaces a large set of silent fallback paths with explicit Result/read-error surfaces, including recent Memory OS dashboard and Hebbian/user-model read diagnostics.
- Fixes dashboard compaction snapshots so future runtime manifests carry pre-dispatch compaction token counts and the API hydrates before/after/saved tokens from the durable compact audit store.
- Updates the browser-openable matrix report at `reports/keeper-scheduler-waiting-goal-matrix-2026-07-04.html`.

## Validation

Local focused checks run before publishing:

- `git diff --check`
- Changed OCaml files: `ocamlformat --check`
- Changed OCaml files: `ocamlc -stop-after parsing`

Dune compile/test is intentionally CI-owned for this workflow.

## Notes

- Draft PR by default.
- Branch was pushed as `codex/scheduler-runner-status-20260704`.
- Current compaction snapshot fix is commit `e7efaf9322`.
- This is a large production-hardening PR; review should focus on boundary correctness, no silent failure, no heuristic control-flow, and MASC/OAS separation.

PR #23408, opened 2026-07-06
feat(dashboard): wire keeper waiting inventory

jeong-sik:
Stacked on #23405. Extracted from the superseded #23393 bundle.

Scope:
- Attaches `keeper_waiting_inventory` to `/api/v1/dashboard/tools` outside the cached tool inventory payload, alongside the live scheduled automation projection.
- Adds Dashboard TS contracts for keeper waiting rows, keeper rows, inventory totals, read-error/global rows, and truncation metadata.
- Adds a reusable `KeeperWaitingInventoryPanel` and wires it into both Tools and Schedule surfaces.
- Adds focused Preact/Vitest coverage for the panel and both surface wiring points.

Out of scope for this slice:
- MCP tool registration, generated tool descriptors, tool policy, CLI wrappers, and proof scripts. Those remain separate extraction slices.

Validation performed locally:
- `opam exec -- ocamlformat --inplace lib/server/server_dashboard_http_runtime_info.ml`
- `pnpm --dir dashboard install --frozen-lockfile`
- `pnpm --dir dashboard exec tsc --noEmit --pretty false`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/tools/keeper-waiting-inventory-panel.test.ts src/components/tools/tools-main.test.ts src/components/schedule/schedule-surface.test.ts`
- `git diff --cached --check`
- conflict-marker scan over changed files.

Notes:
- No local Dune build/test result is claimed; Dune validation remains CI-owned per project direction.

PR #23405, opened 2026-07-06
feat(keeper): add waiting inventory projection

jeong-sik:
Stacked on #23403. Extracted from the superseded #23393 bundle.

Scope:
- Adds a server-owned keeper waiting/deferred read model for dashboard tooling.
- Projects event queue pending/inflight, queued chat, turn admission waits, HITL approvals, external attention, fusion runs, background tasks, schedule waits, operator pending confirms, and explicit read-error rows.
- Adds result-returning read APIs for the stores that this projection must not silently collapse: schedule store, keeper meta discovery, operator pending confirms, and external attention.
- Keeps dashboard route/UI wiring out of this slice; that is the next stacked slice.

Validation performed locally:
- `opam exec -- ocamlformat --inplace ...` for touched OCaml files.
- `git diff --cached --check`
- conflict-marker scan over changed files.

Notes:
- Per project direction, no local Dune build/test result is claimed here; Dune validation is left to CI.

PR #23412, opened 2026-07-06
feat(cli): add keeper waiting inventory proof wrapper

jeong-sik:
Stacked on #23410. Extracted from the superseded #23393 bundle.

Scope:
- Adds `scripts/keeper-waiting-inventory.sh`, a read-only MCP wrapper for `masc_keeper_waiting_inventory`.
- Adds dashboard preflight gating for runtime identity, tool surface registration, and waiting-inventory projection shape.
- Adds `--proof-out` support for typed pass/fail CLI proof JSON.
- Adds `scripts/keeper-waiting-inventory-proof-check.sh` to validate preflight and CLI proof files, with `--require-pass` support.

Validation performed locally:
- `bash -n scripts/keeper-waiting-inventory.sh scripts/keeper-waiting-inventory-proof-check.sh`
- `shellcheck -x scripts/keeper-waiting-inventory.sh scripts/keeper-waiting-inventory-proof-check.sh`
- proof checker sample fail proof: `--allow-fail` accepted and `--require-pass` rejected.
- `scripts/keeper-waiting-inventory.sh --help`
- `scripts/keeper-waiting-inventory-proof-check.sh --help`
- `git diff --cached --check`
- conflict-marker scan over changed files.

Notes:
- No live MCP call is claimed here; the wrapper requires a running deployed/runtime-compatible server and auth token.
- No local Dune build/test result is claimed; Dune validation remains CI-owned per project direction.

### DIFF
--- a/scripts/keeper-waiting-inventory-proof-check.sh
+++ b/scripts/keeper-waiting-inventory-proof-check.sh
@@ -1,0 +1,462 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${REQUIRE_PASS:=0}"
+
+usage() {
+  cat >&2 <<'USAGE'
+Usage: scripts/keeper-waiting-inventory-proof-check.sh [--require-pass|--allow-fail] PROOF.json
+
+Validates proof emitted by scripts/keeper-waiting-inventory.sh --proof-out.
+Set REQUIRE_PASS=1 or pass --require-pass to reject fail proofs.
+USAGE
+}
+
+fail() {
+  printf 'FAIL: %s\n' "$*" >&2
+  exit 1
+}
+
+truthy() {
+  case "$(printf '%s' "${1:-}" | tr '[:upper:]' '[:lower:]')" in
+    1|true|yes|on) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+require_pass="$REQUIRE_PASS"
+case "${1:-}" in
+  -h|--help)
+    usage
+    exit 0
+    ;;
+  --require-pass)
+    require_pass=1
+    shift
+    ;;
+  --allow-fail)
+    require_pass=0
+    shift
+    ;;
+esac
+
+if [[ $# -ne 1 ]]; then
+  usage
+  exit 2
+fi
+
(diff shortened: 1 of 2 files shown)


jeong-sik (commented):
Adversarial review (approve — #23393 slice).

Adds scripts/keeper-waiting-inventory.sh (read-only MCP wrapper for masc_keeper_waiting_inventory) + dashboard prefix. Right direction.

P2 stacking: chain is now 5 deep (one of #23402/#23403 → #23405 → #23408 → #23410 → #23412). Flatten before merging — close the duplicate runner-status PR and rebase.

jeong-sik:
Current-head review at `bb1f439fce0ca3a4c32f2bc30967890922768248`.

P1: token alias support is documented but broken for the MCP initialize request. The usage text says `MCP_TOKEN`, `MCP_AUTH_TOKEN`, and `MASC_ADMIN_TOKEN` are accepted (`scripts/keeper-waiting-inventory.sh:50-52`), and `load_mcp_token` returns early when any of those aliases exist (`scripts/keeper-waiting-inventory.sh:221-229`). But `initialize_mcp_session` only creates an auth header from `MCP_TOKEN` (`scripts/keeper-waiting-inventory.sh:534-535`) and then sends initialize with that header set (`scripts/keeper-waiting-inventory.sh:550-551`). So an environment with only `MCP_AUTH_TOKEN` or `MASC_ADMIN_TOKEN` set is treated as authenticated by `load_mcp_token`, but the initialize request is actually unauthenticated.

The shared helper already has the correct alias resolver (`mcp_default_auth_token` in `scripts/harness/lib/mcp_jsonrpc.sh`), and the later tool call path can fall back through that helper. The initialize path should use the same canonical token resolution, or `load_mcp_token` should normalize aliases into `MCP_TOKEN` before returning. As written, the wrapper fails in a valid documented auth mode before it ever reaches `masc_keeper_waiting_inventory`.

P2: this should get a focused shell smoke test with only `MCP_AUTH_TOKEN` set and `MCP_TOKEN` unset, asserting the initialize request carries Authorization. Otherwise future wrappers can regress the alias contract again.

PR #23402, opened 2026-07-06
[codex] split scheduler runner status surface

jeong-sik:
## Summary
- split the scheduler runner status/audit slice out of #23393 into a small rebase-clean PR
- add a typed Schedule_runner_status snapshot for tick lifecycle, counts, timing, and degraded/crash state
- expose schedule_runner in full health output and record tick ok/error/crash outcomes from the maintenance loop

## Scope
- first split only; #23393 remains open and was not force-pushed
- dashboard wiring, keeper waiting inventory, broad fallback refactors, and generated report artifacts stay out of this PR

## Validation
- `opam exec -- ocamlformat --check lib/schedule/schedule_runner_status.ml lib/schedule/schedule_runner_status.mli lib/server/server_schedule_runner_policy.ml lib/server/server_schedule_runner_policy.mli lib/server/server_bootstrap_maintenance.ml lib/server/server_routes_http_runtime.ml test/test_schedule_runner.ml`
- `git diff --check`
- no local dune build; dune validation is left to CI per repo workflow

PR #23403, opened 2026-07-06
feat(schedule): track runner status

jeong-sik:
## Summary

Extracts the first safe slice from superseded #23393: process-local scheduler runner status.

- Adds `Schedule_runner_status` as the process-local status read model for the production scheduler loop
- Records tick start, success, typed tick errors, and crashes from `server_bootstrap_maintenance`
- Degrades status when the latest tick contains dispatch failures, unsupported dispatches, start rejections, wake enqueue failures, newer tick errors, stale ticks, or an in-flight tick
- Adds focused regression coverage in `test/test_schedule_runner.ml`

## Boundary

This PR intentionally does not include #23393's waiting inventory, dashboard wiring, schedule audit-log rewrite, or generated report artifact. Those are separate slices.

## Validation

- `opam exec -- ocamlformat --inplace lib/schedule/dune lib/schedule/schedule_runner_status.ml lib/schedule/schedule_runner_status.mli lib/server/server_bootstrap_maintenance.ml test/test_schedule_runner.ml`
- `git diff --check -- lib/schedule/dune lib/schedule/schedule_runner_status.ml lib/schedule/schedule_runner_status.mli lib/server/server_bootstrap_maintenance.ml test/test_schedule_runner.ml`
- `rg -n "^(<<<<<<<|=======|>>>>>>>)" lib/schedule/dune lib/schedule/schedule_runner_status.ml lib/schedule/schedule_runner_status.mli lib/server/server_bootstrap_maintenance.ml test/test_schedule_runner.ml`
- `scripts/dune-local.sh exec ./test/test_schedule_runner.exe`

PR #23414, opened 2026-07-06
feat(schedule): centralize payload projection

jeong-sik:
Stacked on #23412.

This extracts the schedule payload projection/validation portion from #23393 into a reviewable slice. It keeps the schedule domain payload opaque, moves supported payload knowledge into `Schedule_payload_projection`, wires tool/server callers to use the typed projection, exposes schedule surface tool lists through `Tool_catalog_surfaces`, and adds only the `masc_schedule_payload_unsupported_total` counter needed by this slice.

Local validation performed without Dune per project direction:
- `opam exec -- ocamlformat --inplace ...` on changed OCaml files
- `git diff --cached --check`
- conflict marker scan on changed files
- `bash scripts/ci/check-telemetry-coverage.sh --base origin/codex/keeper-waiting-cli-slice-20260706 --head HEAD`

Dune/build validation is intentionally left to CI.

### DIFF
--- a/lib/otel_metric_store/otel_metric_names.ml
+++ b/lib/otel_metric_store/otel_metric_names.ml
@@ -142,6 +142,14 @@ let metric_backend_mutex_held_sec = "masc_backend_mutex_held_sec"
 let metric_tool_metrics_persist_dropped =
   Otel_metric_store_core.declare_counter "masc_tool_metrics_persist_dropped_total"
 
+(* Schedule unsupported payload counter. Labels:
+   - [phase] in {creation, dispatch}
+   - [risk_class] in [Schedule_domain.risk_class_to_string] labels.
+   Raw payload kinds are intentionally not labels; they remain in typed
+   errors/projections to avoid unbounded metric cardinality. *)
+let metric_schedule_payload_unsupported_total =
+  Otel_metric_store_core.declare_counter "masc_schedule_payload_unsupported_total"
+
 let metric_tool_bind_required_guard =
   Otel_metric_store_core.declare_counter "masc_tool_bind_required_guard_total"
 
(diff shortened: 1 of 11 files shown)


jeong-sik (commented):
Adversarial review (approve — #23393 slice).

Extracts the schedule payload projection/validation slice, keeping the schedule domain payload opaque and moving supported-payload knowledge into one owner. Right direction — SSOT centralization of the projection.

P2 stacking — this is the critical concern: the chain is now 6 deep (one of #23402/#23403 → #23405 → #23408 → #23410 → #23412 → #23414). Strongly recommend:
1. Close #23402 OR #23403 (duplicate runner-status) immediately.
2. Rebase the whole chain onto whichever lands.
3. Consider rebasing each slice onto main directly if they are truly independent (payload projection is arguably separable from waiting-inventory). A 6-deep stack is a merge/review liability and makes bisecting regressions painful.

jeong-sik:
P0 build blocker: this slice changes the cancel call site to a service API that does not exist on the PR head.

Evidence from the current head `30342033cd`:

- `lib/tool_schedule.ml:536-538` now calls `Schedule_service.cancel ctx.config ~schedule_id ~cancelled_by ~reason`.
- `lib/schedule/schedule_service.mli:58-61` still exposes `val cancel : Workspace_utils.config -> schedule_id:string -> (Schedule_domain.schedule_request, service_error) result`.
- `lib/schedule/schedule_service.ml:113-114` still implements `let cancel config ~schedule_id = Schedule_store.cancel_request config ~schedule_id |> map_store`.

That is an OCaml labelled-argument mismatch; a full build will reject the extra `~cancelled_by`/`~reason` labels. The current stacked PR checks only show axis/coverage/format, so this is hidden until CI runs a real build.

Recommended fix: either keep the existing `Schedule_service.cancel ctx.config ~schedule_id` call in this payload-projection slice, or extend `Schedule_service.cancel`/`Schedule_store.cancel_request` and the audit write path in the same slice. If the latter is intended, the service `.mli`, implementation, store/audit behavior, and cancel tests need to move together so the service boundary remains the SSOT.

P0: source build breaker. Mergeability: blocked until the API mismatch is resolved.

PR #23423, opened 2026-07-06
[codex] Add task snapshots to catchup digest

jeong-sik:
## Summary

- Add a current task snapshot to each keeper catch-up digest task item when the task still exists in backlog.
- Include status, assignee, verification phase/id, submitted timestamp, and handoff summary/evidence in the JSON payload.
- Extend `test_keeper_catchup_digest` fixtures/assertions so fusion-style task transitions carry current backlog context.

## Root Cause

Fusion catch-up judge prompts only received task transition facts (`task_id`, `transition`, `ts`). For runs like `fus-0547de4e5830742a5adb302f6cb721c1`, the prompt could see `task-1838` was claimed/started but could not see the current backlog state (`awaiting_verification`) or handoff evidence (`PR#23399`, `commit:f2678ed43`). That made panel/judge answers look like there was less progress than the durable task state actually showed.

## Validation

- `git diff --cached --check`
- `MASC_SKIP_OCAML_VERSION_CHECK=1 scripts/dune-local.sh build test/test_keeper_catchup_digest.exe`
- `./_build/default/test/test_keeper_catchup_digest.exe`

Note: the regular local wrapper stops on this machine because the active opam switch is OCaml 5.4.1 while `dune-project` requires OCaml >= 5.5. The targeted build above used the repo wrapper with only the version guard bypassed.

### DIFF
--- a/lib/keeper_catchup_digest.ml
+++ b/lib/keeper_catchup_digest.ml
@@ -39,6 +39,8 @@ let crash_scan_max = 500
 let audit_dirname = "audit" (* Audit_log store *)
 let activity_events_dirname = "activity-events" (* Activity_graph store *)
 let transition_audit_dirname = "transition-audit" (* Keeper_transition_audit durable store *)
+let tasks_dirname = "tasks" (* Workspace_utils_paths_backend.tasks_dir *)
+let backlog_filename = "backlog.json" (* Workspace_utils_paths_backend.backlog_path basename *)
 
 (* keeper.* activity-event kinds are still raw strings pending the #8455
    Event_kind migration; the board.* kinds below come from the typed
@@ -54,10 +56,24 @@ let operator_resume_event = "operator_resume"
 
 (* ── Types (mirror the .mli) ─────────────────────────────────────── *)
 
+type task_snapshot =
+  { title : string
+  ; status : string
+  ; assignee : string option
+  ; phase : string option
+  ; verifier : string option
+  ; submitted_at : string option
+  ; verification_id : string option
+  ; handoff_summary : string option
+  ; handoff_next_step : string option
+  ; handoff_evidence_refs : string list
+  }
+
 type task_item =
   { task_id : string
   ; transition : string
   ; ts : float
+  ; current_task : task_snapshot option
   }
 
 type lifecycle_item =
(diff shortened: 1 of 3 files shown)


jeong-sik (commented):
Adversarial review (approve).

Adds a current task snapshot to each keeper catch-up digest task item (status, assignee, verification phase/id, submitted timestamp). Strengthens the catch-up digest's observability so an operator sees real task state, not just a digest stub. Right direction.

P3 nit: 3 files — confirm the snapshot shape is the single source, and that a missing/deleted task is surfaced explicitly (not silently omitted), so a stale digest entry is diagnosable.

jeong-sik (commented):
Current-head adversarial review at `9aca46ff61ebb117c6896df99a3f9d131530cb36`.

P1/source blocker: this PR adds a new backlog path derivation inside `lib/keeper_catchup_digest.ml`:

- `let tasks_dirname = "tasks"`
- `let backlog_filename = "backlog.json"`
- `Filename.concat (Filename.concat masc_dir tasks_dirname) backlog_filename`

That duplicates the existing workspace path SSOT. The repo already exposes `Workspace_utils_paths_backend.tasks_dir` and `Workspace_utils_paths_backend.backlog_path`, and the implementation already documents why base-path-only callers should not re-inline `.masc` layout. A catch-up digest is operational/runtime code; it should not carry a second local copy of the tasks/backlog layout, because cluster-root/backlog relocation or backend path changes would silently split digest truth from the real backlog. This is exactly the path drift class we have been treating as blocker-level.

Fix direction: derive the path from the workspace path module (`Workspace_utils_paths_backend.default_config base_path |> Workspace_utils_paths_backend.backlog_path`, or a small exported base-path helper if config construction would be the wrong dependency boundary). If this module is intentionally cycle-bound, add a shared helper beside the existing path SSOT instead of local string constants. The test helper can still mirror fixture creation, but production code should not re-own `"tasks"` / `"backlog.json"`.

P2/behavior: the `current_task = None` representation for deleted/missing backlog rows is fail-visible only as JSON `null`; that is acceptable for a draft if the path SSOT is fixed, but consider adding a cause/read-error when the backlog exists but cannot be parsed. The current code already appends read errors for parse/IO failures, so keep that behavior.

P2/process: CI is not terminal green: `Build and Test` is still in progress and `Branch Protection Watchdog` is the known repo-level failure. No merge-safe label.

[근거] `gh pr view 23423 --repo jeong-sik/masc --json headRefOid,mergeable,mergeStateStatus,statusCheckRollup`; #23423 diff for `lib/keeper_catchup_digest.ml`; `rg "backlog_path|tasks_dir" lib/workspace` showing `Workspace_utils_paths_backend.backlog_path` / `tasks_dir`; source read of `lib/workspace/workspace_utils_paths_backend.ml` and `.mli`; checked 2026-07-07 00:00 KST. 신뢰도: High.

jeong-sik:
P0: current red is not a confirmed source regression. CI Gate only aggregates `build=cancelled`; the Build/Test log had runner/process signals earlier, including disk exhaustion plus a transport harness transient. P1: reran failed jobs with `gh run rerun 28800205013 --failed`. P2: no local Dune run per CI-build-authority; review sweep is clean with 0 actionable threads. P3: head `9aca46ff61ebb117c6896df99a3f9d131530cb36`, draft, merge blocked only by CI rerun state.

jeong-sik:
Current-head follow-up at `9aca46ff61ebb117c6896df99a3f9d131530cb36`.

P1 still stands. The later status note says the review sweep is clean, but the source diff has not changed: `lib/keeper_catchup_digest.ml` still defines `tasks_dirname = "tasks"`, `backlog_filename = "backlog.json"`, then rebuilds the backlog path with `Filename.concat (Filename.concat masc_dir tasks_dirname) backlog_filename`. The path owner already exists at `Workspace_utils_paths_backend.tasks_dir/backlog_path`; operational digest code should not re-own that layout.

P2/CI is separate: current CI is terminal red (`Build and Test` failed/cancelled and `CI Gate` failed). The log failures I see are shared/repo-level memory-OS and shell-IR leading-flag cases, not this PR's three changed files, but they are still merge-blocking. No merge-safe label.

[근거] `gh pr view 23423 --repo jeong-sik/masc --json headRefOid,isDraft,mergeStateStatus,files,labels`; `gh pr checks 23423 --repo jeong-sik/masc`; `git fetch origin +pull/23423/head:refs/tmp/pr-23423-current`; source reads of `lib/keeper_catchup_digest.ml` and `lib/workspace/workspace_utils_paths_backend.ml`; checked 2026-07-07 01:00 KST. 신뢰도: High.

PR #23425, opened 2026-07-06
[codex] Make passive-only completion observable

jeong-sik:
## What changed

- Reclassifies `completion_contract_result=passive_only` as observable telemetry rather than an attention-required completion-contract result.
- Stops runtime trust snapshots from converting passive-only receipts into `completion_contract_result:passive_only` attention reasons, even when a task or goal scope is present.
- Keeps passive-only visible in reaction-ledger observation counts while excluding it from completion-contract attention counts.
- Updates checkpoint/finalizer tests so true attention examples use `violated`, while passive-only is covered as non-blocking.

## Why

`passive_only` means a turn only produced passive/status evidence. That is useful to observe, but it should not itself block, page, prune replay, or become operator-facing attention. Repeated lack of useful progress should be handled by no-progress detection instead.

## Validation

- `MASC_SKIP_OCAML_VERSION_CHECK=1 scripts/dune-local.sh build test/test_keeper_reaction_ledger.exe test/test_keeper_runtime_trust_snapshot.exe test/test_keeper_state_snapshot_json.exe test/test_keeper_terminal_reason_typed.exe`
- `./_build/default/test/test_keeper_reaction_ledger.exe`
- `./_build/default/test/test_keeper_runtime_trust_snapshot.exe`
- `./_build/default/test/test_keeper_state_snapshot_json.exe`
- `./_build/default/test/test_keeper_terminal_reason_typed.exe`
- `git diff --name-only -z HEAD~1..HEAD -- '*.ml' '*.mli' | xargs -0 -I{} opam exec -- ocamlformat --check {}`

### DIFF
--- a/lib/keeper/keeper_completion_contract_result_label.ml
+++ b/lib/keeper/keeper_completion_contract_result_label.ml
@@ -48,11 +48,11 @@ let requires_attention = function
   | Surface_mismatch
   | No_capable_provider
   | Claim_only_after_owned_task
-  | Needs_execution_progress
-  | Passive_only ->
+  | Needs_execution_progress ->
     true
   | Unknown
   | Not_dispatched
+  | Passive_only
   | Satisfied_completion
   | Satisfied_execution ->
     false
--- a/lib/keeper/keeper_execution_receipt.ml
+++ b/lib/keeper/keeper_execution_receipt.ml
@@ -180,12 +180,12 @@ let completion_contract_satisfied = function
 let completion_contract_unsatisfied = function
   | Contract_violated
   | Contract_claim_only_after_owned_task
-  | Contract_needs_execution_progress
-  | Contract_passive_only -> true
+  | Contract_needs_execution_progress -> true
   | Contract_unknown
   | Contract_not_dispatched
   | Contract_surface_mismatch
   | Contract_no_capable_provider
+  | Contract_passive_only
   | Contract_satisfied_completion
   | Contract_satisfied_execution -> false
 ;;
--- a/lib/keeper/keeper_reaction_ledger.ml
+++ b/lib/keeper/keeper_reaction_ledger.ml
@@ -541,28 +541,6 @@ let reaction_receipt_field name row =
   | None -> None
 ;;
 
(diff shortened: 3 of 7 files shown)


jeong-sik:
Current-head adversarial pass on `4356a00e7eef6d7d3cdc5fc4feab8541ee3508ef`.

- P0/P1 source: none found in this pass. The branch consistently reclassifies `passive_only` as observable but non-attention: `Keeper_completion_contract_result_label.requires_attention` and `Keeper_execution_receipt.completion_contract_unsatisfied` no longer treat it as blocking, `Keeper_reaction_ledger` still increments passive/result telemetry, and `Keeper_runtime_trust_snapshot` no longer turns active passive-only receipts into `needs_attention`.
- Regression coverage is on the right surfaces: ledger summary counts passive-only without attention, runtime trust covers active and no-work passive receipts, and finalizer/checkpoint tests now use `violated` for true attention behavior while preserving passive-only resume/raw-text behavior separately.
- P2 process: this is still draft and required CI is pending (`Build and Test`, `Health`, `Lint`, `Meta Guards`, `OCaml 5.5 Build Canary`, `ocamlformat-check`, TLA+ model checking). No merge-safe label.

[근거] `git diff refs/tmp/origin-main-review...refs/tmp/pr-23425` inspected for `lib/keeper/keeper_completion_contract_result_label.ml`, `keeper_execution_receipt.ml`, `keeper_reaction_ledger.ml`, `keeper_runtime_trust_snapshot.ml`, and the three updated tests; `gh pr checks 23425 --repo jeong-sik/masc` checked 2026-07-07 00:41 KST. 신뢰도: High.

jeong-sik (commented):
Adversarial review (approve — classification fix).

Reclassifies completion_contract_result=passive_only as observable telemetry rather than attention-required. Stops treating a passive outcome as a completion-contract failure. Right direction — sharper contract semantics, removes a false attention signal.

P3 nit: 7 files — confirm the classification enum is the single source, and that a genuinely-failing completion is still distinguishable from a passive-but-healthy one.

PR #23428, opened 2026-07-06
feat(keeper): re-land waiting inventory projection stranded off-main (#23405/#23408 salvage)

jeong-sik:
## Why — MERGED 표시지만 main에 도달하지 못한 waiting inventory 재착지

#23405(keeper waiting inventory projection)와 #23408(dashboard wiring)은 GitHub 상태가 MERGED지만, stacked 체인의 base가 **닫힌 duplicate PR #23403의 브랜치**(`codex/scheduler-runner-status-slice-20260706`)였습니다:

- #23408 → base = `codex/keeper-waiting-inventory-slice-20260706` (#23405의 head) 로 머지 (15:15:27Z)
- #23405 → base = `codex/scheduler-runner-status-slice-20260706` (#23403의 head) 로 머지 (15:15:51Z)
- #23403 → #23402의 duplicate로 **머지 없이 close** (14:32Z)

결과: 두 PR의 내용 전체가 좌초된 사이드 브랜치에만 존재하고, `lib/server/server_keeper_waiting_inventory.ml` 등은 main 트리에 없습니다.

## What

좌초 커밋 `8e465678ed`(#23405 squash — #23408 내용 포함)를 `origin/main` 기준으로 cherry-pick 했습니다. **충돌 0** — 기존에 보고된 add/add 충돌은 전부 #23402로 superseded된 부모 커밋(`2a7dcd8533`, runner status) 몫이었고 이 커밋에는 해당 없습니다.

- 원본 리뷰/승인 이력: #23405, #23408 참조 (내용 동일, 재작성 없음)
- 부모 슬라이스(runner status)는 이미 #23402로 main에 있음 — 이 PR에 포함하지 않음

## 검증 경계

cherry-pick은 텍스트 충돌 0이지만, main이 그 사이 10커밋 진행(#23402 split의 `schedule_runner_status` API 변형 포함)했으므로 **컴파일/테스트 정합은 이 PR의 CI(Build and Test, `@default @check`)가 증명**합니다. red면 API drift를 이 브랜치에서 해소합니다.

## 프로세스 노트

stacked PR의 base가 닫히면 자식들의 "Merged"는 main 도달을 의미하지 않습니다. #23403을 duplicate로 닫을 때 자식 재베이스가 누락된 사고 — 브랜치 정리는 이 PR 머지 후 수행 예정.

Co-Authored-By: Codex <codex@openai.com>
Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #23429, opened 2026-07-06
fix(keeper): count read-error rows as waiting

jeong-sik:
## Summary
- Treat read-error-only keeper rows as `waiting` instead of `idle`.
- Pin the regression with `waiting_keeper_count` and keeper state metric assertions for corrupt external attention reads.

## Context
- The merged keeper waiting inventory stack exposes keeper-scoped read errors as repair rows, but `row_state` still classified read-error-only rows as `idle`.
- That hides an actionable repair condition from `waiting_keeper_count` and the per-state metrics.

## Verification
- `opam exec -- ocamlformat --check lib/server/server_keeper_waiting_inventory.ml test/test_keeper_waiting_inventory.ml`
- `git diff --check`

Local Dune build/test intentionally not run; PR CI is the build authority.

### DIFF
--- a/bin/gen_tool_descriptors.ml
+++ b/bin/gen_tool_descriptors.ml
@@ -103,6 +103,17 @@ let masc_dashboard_spec : tool_spec =
   }
 ;;
 
+let masc_keeper_waiting_inventory_spec : tool_spec =
+  { name = "masc_keeper_waiting_inventory"
+  ; description =
+      "Return the canonical keeper waiting inventory read model: what each keeper is \
+       waiting on, source counts, global waiting rows, and supported state labels."
+  ; parameters = []
+  ; additional_properties = false
+  ; behavior_contract = []
+  }
+;;
+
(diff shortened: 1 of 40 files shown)


jeong-sik:
Closing this draft because the reused historical stack branch produced a DIRTY head/base shape. Re-opening the same read-error fix from the current scheduler stack base on a fresh branch.

jeong-sik (commented):
Adversarial review (approve — classification fix).

Treats read-error-only keeper rows as `waiting` instead of `idle`, and pins the regression with waiting_keeper_count + keeper state metrics. Right direction — a read-error is a wait state, not idle; removes a silent undercount of waiting work.

P1 duplicate check: this is +5145/-338 / 40 files — nearly identical shape to #23428 (the waiting-inventory re-land). Confirm they are distinct (this = read-error classification, #23428 = waiting-inventory projection). If overlapping, merge them into one.

PR #23446, opened 2026-07-07
[codex] Preserve gate source context in keeper chat

jeong-sik:
## Summary

This keeps connector-origin chat rows from losing their source identity at the gate boundary.

- Thread `idempotency_key` through `Channel_gate` into `Gate_keeper_backend` so inbound connector rows can retain a stable external message fallback.
- Persist Discord connector rows with typed `Surface_ref.Discord` instead of flattening them to `Gate { label = "discord"; address = [] }`.
- Populate generic `Gate.address` with connector/workspace/conversation/message coordinates when a connector cannot be represented by a typed surface.
- Surface speaker, conversation, and external message metadata in the dashboard transcript chips and in the `턴 연결 없음` badge title.

## Root Cause

The store already had fields for `surface`, `conversation_id`, `external_message_id`, and `speaker_*`, but the gate/backend path wrote Discord traffic as a generic empty-address gate surface. The dashboard also passed those REST fields through the schema but did not lift them into `KeeperConversationEntry`, so even preserved metadata could not be inspected in the bubble UI.

## Validation

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/chat/primitives.test.ts src/keeper-state.test.ts --no-file-parallelism --maxWorkers=1`
- `pnpm exec eslint src/components/chat/primitives.ts src/components/chat/primitives.test.ts`
- `git diff --check`
- `ocamlformat --check lib/gate/channel_gate.ml lib/gate/channel_gate.mli lib/gate_keeper_backend.ml lib/gate_keeper_backend.mli lib/server/server_routes_http_keeper_stream.ml test/test_channel_gate.ml test/test_copilot_dock_channel_wiring.ml test/test_gate_keeper_backend.ml test/test_keeper_busy_connector_deferred.ml`

Local OCaml focused build was attempted but blocked by the local switch, not by a compiler diagnostic from this diff:

```text
scripts/dune-local.sh build ./test/test_channel_gate.exe ./test/test_gate_keeper_backend.exe ./test/test_copilot_dock_channel_wiring.exe ./test/test_keeper_busy_connector_deferred.exe
[dune-local] OCaml 5.4 detected; this repo requires >= 5.5 (dune-project)
```

### DIFF
--- a/dashboard/src/components/chat/primitives.test.ts
+++ b/dashboard/src/components/chat/primitives.test.ts
@@ -242,6 +242,53 @@ describe('ChatTranscript', () => {
     )
     const surfaceLink = bubble.querySelector('a[href="https://discord.com/channels/guild-1/thread-1"]')
     expect(surfaceLink?.textContent).toContain('Discord Thread')
+    expect(surfaceLink?.getAttribute('title')).toContain('guild_id=guild-1')
+    expect(surfaceLink?.getAttribute('title')).toContain('thread_id=thread-1')
+  })
+
+  it('renders connector speaker and route context for gate history rows', () => {
+    render(
+      html`<${ChatTranscript}
+        entries=${[
+          entry({
+            id: 'gate-user',
+            text: 'hello from discord',
+            speakerId: '98791450001',
+            speakerName: 'Minsu',
+            speakerAuthority: 'external',
+            conversationId: 'discord:guild-1:channel:1514586257706061834',
+            externalMessageId: '1498985300729172039',
+            surface: {
+              kind: 'gate',
+              label: 'discord',
+              address: {
+                connector: 'discord',
+                workspace_id: '1514586257706061834',
+                external_message_id: '1498985300729172039',
+              },
+            },
+          }),
+        ]}
+        emptyText="empty"
+        variant="messenger"
+      />`,
+      container,
+    )
+
+    const bubble = container.querySelector('[data-chat-entry-id="gate-user"]') as HTMLElement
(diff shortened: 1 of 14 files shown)


jeong-sik (commented):
Adversarial review (approve — preserves identity across the gate).

Threads `idempotency_key` through `Channel_gate` into `Gate_keeper_backend` so inbound connector-origin chat rows keep their source identity at the gate boundary. Prevents silent identity loss. Same anti-fabrication line as #23398 (board author meta key). Right direction.

P3 nit: 14 files — confirm `idempotency_key` is the single identity token and that a missing key is handled explicitly (not silently allowed through), matching the existing gate-fail-closed convention.

jeong-sik:
Review follow-up for the P3 note:

Confirmed with one nuance: `idempotency_key` is not the single connector identity token. It is the Channel_gate admission/dedup key, and this PR threads it downstream only so keeper chat can use it as `external_message_id` fallback when connector metadata does not already provide `external_message_id` or `discord.message_id`.

Source identity is carried separately by the typed surface/address: generic gate connector/workspace/conversation/external ids, and Discord guild/channel/thread ids. That is the durable context the dashboard and chat rows should display.

Missing/blank `idempotency_key` is still handled explicitly at the Channel_gate validation boundary as `Empty_idempotency_key` before dispatch. Downstream also does not synthesize a fake external message id: if metadata and idempotency key are absent, `external_message_id_for_channel_context` returns `None` rather than inventing one.

Relevant coverage: `test_channel_gate` covers idempotency validation/dedup and dispatch propagation, and `test_gate_keeper_backend`/dashboard tests cover preserving typed source context and external message ids.

PR #23480, opened 2026-07-07
fix(work): separate Goal Store kanban from summary KPIs

jeong-sik:
## Summary

- make the Goal Store KPI row a compact summary cluster instead of a full-width 5-column strip
- align the kanban section with the latest v2 sample: `칸반 · 상태별`, status dots, compact status labels, and inline counts
- add focused tests so the kanban headers keep the sample labels/dot/count structure and do not regress into repeated TASK scope labels

## Verification

- `pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/work.test.ts`
- `pnpm --dir dashboard typecheck`
- `pnpm --dir dashboard lint`
- `git diff --check origin/main...HEAD`
- Playwright render check: desktop/mobile screenshots saved under `/private/tmp/goal-store-current-v5-*.png`; no overlay or page errors

## Notes

- Reference folder checked: `/Users/dancer/Downloads/v2 (29)`
- The standalone HTML bundle in that folder appears stale versus `keeper-v2/work.jsx`; this PR follows the latest `work.jsx` structure and the provided visual sample.

PR #23485, opened 2026-07-07
fix(work): align backlog language with sample aside

jeong-sik:
## Summary
- align Work KPI/aside/backlog copy with the v2 sample
- rename unassigned board language to backlog where the UI represents claimable backlog
- add type rails/tints for WorkAside todo cards so approval/verify/block/claim items scan separately

## Verification
- pnpm --dir dashboard exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1 src/components/work.test.ts
- bash scripts/lint/no-raw-font-size-px.sh
- pnpm --dir dashboard typecheck
- pnpm --dir dashboard lint
- Playwright capture against /Users/dancer/Downloads/v2 (29): desktop/mobile no overlay or page errors

Follow-up to #23480; #23480 was already merged before this sample-polish diff was pushed.

### DIFF
--- a/dashboard/src/components/work.test.ts
+++ b/dashboard/src/components/work.test.ts
@@ -289,7 +289,7 @@ describe('Work', () => {
       expect(screen.getByTestId('kpi-backlog').textContent).toBe('2')
       // Five KPI summary cells
       expect(screen.getByTestId('work-kpis').children.length).toBe(5)
-      expect(screen.getByText(/미배정 task는 claim/).textContent).toContain('claim')
+      expect(screen.getByText(/미배정 task는 백로그에서 claim/).textContent).toContain('claim')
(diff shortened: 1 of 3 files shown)


jeong-sik:
Current-head adversarial source review — `08685c4` (2026-07-07 15:12 KST)

P0/P1/P2/P3 source blockers: none found in the three-file Work UI copy/style diff. `git diff --check origin/main...origin/pr-23485` is clean.

CI/process: Branch Protection Watchdog fails with `enforce_admins.enabled=false; expected true`. The main CI run was then cancelled while setup/typecheck jobs were in flight; returned job metadata does not show a source failure for this diff. Draft + process CI failure means not merge-ready.

PR #23500, opened 2026-07-07
fix(test): repair three quick-suite reds on main (pacing shadow Eio ctx, metrics tripwire 249, ide scope message)

jeong-sik:
fix(test): repair three quick-suite reds on main

Build and Test on main (912c41fc1e) fails three tests from two merges:

1. keeper_pacing shadow (W1 #23487): Keeper_pacing_shadow uses
   Eio.Mutex.use_rw ~protect:true, which needs a fiber context even
   uncontended; the test called snapshot outside Eio and raised
   Effect.Unhandled(Cancel.Get_context). Wrap it in Eio_main.run
   (masc_test_deps re-exports eio_main; precedent:
   test_keeper_supervisor).
2. otel_zero_fill module-init (W1 #23487): the documented drift
   tripwire literal was not bumped when PacingShadowEvents/
   PacingShadowNextDueSec joined Keeper_metrics.all — 247 -> 249.
3. server_ide_http scope_contract (#23461): the missing-scope error
   gained keeper_lane in server_ide_http.ml:106 but one test literal
   kept the old two-scope message.

The PR-side quick suite does not run these binaries, so both regressions
surfaced only on the main run — same single-job hostage shape as #23426,
hence one PR for all three.

### DIFF
--- a/test/test_keeper_pacing.ml
+++ b/test/test_keeper_pacing.ml
@@ -74,6 +74,12 @@ let test_next_turn_due_always_finite () =
   feq "empty state is due now" 7.0 (KP.next_turn_due ~catalog:[ "a" ] ~now:7.0 KP.empty)
 
 let test_shadow_snapshot_isolated_by_keeper () =
+  (* Keeper_pacing_shadow guards its table with Eio.Mutex.use_rw
+     ~protect:true, which needs a fiber context even when uncontended —
+     without Eio_main.run this raised
+     Effect.Unhandled(Cancel.Get_context) on main (2026-07-07). *)
+  Eio_main.run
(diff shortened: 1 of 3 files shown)


jeong-sik:
P3/source-pass on current head 2ce65280df02. This is a narrow test-only repair: the Eio_main.run wrapper matches the Keeper_pacing_shadow mutex context requirement and test deps expose eio_main; the Keeper_metrics.all literal moves by exactly the two already-present pacing-shadow constructors on main; the IDE missing-scope expected string now matches server_ide_http.ml. No source blocker from this pass. Mergeability: non-draft but still BLOCKED until checks/branch protection are green.

PR #23523, opened 2026-07-07
fix(dashboard): harden keeper lifecycle follow-up

jeong-sik:
Follow-up to #23517. That PR merged with a dashboard compile break and with the lifecycle limit still encoded as an unnamed magic default.

P0: fixes the merged dashboard typecheck break by guarding the keeper phase badge and removing the unused keeper-detail import.

P1: replaces the hardcoded lifecycle default with `KEEPER_LIFECYCLE_DEFAULT_LIMIT` plus a local rationale.

P2: removes the duplicate raw refresh/close buttons; the existing header ActionButton controls remain.

P3: draft until CI completes.

Validation:
- `pnpm --dir dashboard lint`
- `pnpm --dir dashboard exec vitest run --config vitest.config.ts src/components/keeper-lifecycle-timeline.test.ts --no-file-parallelism --maxWorkers=1`
- `git diff --check`

No local Dune run; this is dashboard-only.

### DIFF
--- a/dashboard/src/api/keeper.ts
+++ b/dashboard/src/api/keeper.ts
@@ -797,6 +797,10 @@ export async function fetchKeeperTransitions(
 
 // --- Keeper lifecycle timeline (#12798) ---
 
+// Detail views need enough recent lifecycle entries to avoid silently truncating
+// active keepers, while compact surfaces can still pass an explicit lower limit.
+export const KEEPER_LIFECYCLE_DEFAULT_LIMIT = 200
+
 export interface KeeperLifecycleEvent {
   ts: number
   event: string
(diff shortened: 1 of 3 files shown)


jeong-sik:
Adversarial follow-up review at head 1c30153:\n\nP0: guarded the top-bar phase badge behind the keeper value and removed the unused keeper-detail-shell import, so the merged dashboard typecheck break is addressed.\nP1: lifecycle detail default is now a named exported contract () with a local rationale instead of an unnamed literal.\nP2: duplicate raw refresh/close buttons are removed; the existing ActionButton controls remain as the single control surface.\nP3: no OCaml/runtime boundary touched; dashboard-only follow-up.\nMergeability: draft PR, mergeable, CI still running. Local validation observed:  and dashboard  pass.

jeong-sik:
Correction for the previous comment: shell substitution stripped the inline code literals.

Adversarial follow-up review at head 1c30153:

P0: guarded the top-bar phase badge behind the keeper value and removed the unused keeper-detail-shell import, so the merged dashboard typecheck break is addressed.
P1: lifecycle detail default is now the named exported contract `KEEPER_LIFECYCLE_DEFAULT_LIMIT` with a local rationale instead of an unnamed literal.
P2: duplicate raw refresh/close buttons are removed; the existing ActionButton controls remain as the single control surface.
P3: no OCaml/runtime boundary touched; dashboard-only follow-up.
Mergeability: draft PR, mergeable, CI still running. Local validation observed: `git diff --check origin/main...HEAD` and dashboard `typecheck` pass.

PR #23672, opened 2026-07-08
fix(workspace): main red — explicit task_status enumeration in verification-id read

jeong-sik:
**main is red**: #23665 merged (`6414ff758`) while its last commit `5772e78398` had a fragile-match compile error — `lib/workspace` builds with `-w +4 -warn-error`, and the `Some _ | None` catch-all after an `AwaitingVerification`-specific arm trips warning 4 (Canary/Build/Health/Lint all red on that run).

One-hunk fix: enumerate all `task_status` constructors explicitly (repo FSM no-catch-all rule).

Branch note: this is the recreated `rfc-0323/g2-probe-verification-lane` branch — its content diff vs main is exactly this fix (everything else landed in the squash merge).

RFC-0323 G-2 follow-up.

### DIFF
--- a/lib/keeper/keeper_tool_task_runtime.ml
+++ b/lib/keeper/keeper_tool_task_runtime.ml
@@ -676,8 +676,10 @@ let handle_keeper_task_tool
                   { approve_error; reject_error }) ->
              Log.Keeper.error ~keeper_name:meta.name
                "deterministic probe stranded %s in awaiting_verification \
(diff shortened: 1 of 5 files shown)


jeong-sik:
Wrong vehicle: this branch's merge-base predates the #23665 squash, so its three-dot diff re-applies the whole G-2 change set instead of the one-hunk fix. Superseding with a fresh branch cut from current main.

PR #23797, opened 2026-07-08
task-1870: thread ?continuation_channel through keeper_msg_body direct dispatch

anyang-keepers:
### What
Thread `?continuation_channel` through the `keeper_msg_body` direct dispatch path in `keeper_tool_surface_ops.ml`, matching what the `handle_keeper_msg` wrapper already does.

### Why
`create_entry` in `keeper_approval_queue.ml` defaults `continuation_channel` to `Unrouted "no originating connector"`. When the direct MCP dispatch path calls `Turn.handle_keeper_msg` without the channel, all HITL approvals from that path lose their provenance, and `Hitl_resolved` wakes go to `Unrouted` instead of the originating chat connector.

### What this PR does
- Adds `?continuation_channel` parameter to `keeper_msg_body` function signature
- Passes `?continuation_channel` through to `Turn.handle_keeper_msg` call
- Non-breaking: optional param defaults to `None` (same behavior as before)

### What remains
- Thread channel from the caller in `keeper_tool_surface.ml` dispatch (line 875) — requires the dispatch lambda to receive channel info from the Eio context
- Remove `Unrouted` default in `keeper_approval_queue.ml` (make required or fail loudly)

### Testing
- Non-breaking change, existing behavior preserved when `continuation_channel` is not provided
- No dune build available in sandbox (policy-blocked); needs CI verification

### DIFF
--- a/lib/keeper/keeper_tool_surface_ops.ml
+++ b/lib/keeper/keeper_tool_surface_ops.ml
@@ -644,6 +644,7 @@ let keeper_msg_body
       ~(clock : float Eio.Time.clock_ty Eio.Resource.t)
       ?proc_mgr
       ?net
+      ?continuation_channel
       args : tool_result =
   let keeper_ctx : _ Keeper_types_profile.context =
     { config; agent_name; sw; clock; proc_mgr; net }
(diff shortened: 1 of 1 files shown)


jeong-sik:
적대적 리뷰: continuation_channel(typed)을 keeper_msg_body direct dispatch path에 threading. handle_keeper_msg wrapper와 정합. +2/-1 단일 파일. 23751에서 지적한 continuation_channel=Unrouted 기본값 문제의 올바른 해결(string option 회귀 아닌 typed threading). 스캔 clean. 머지 가능.

PR #23804, opened 2026-07-09
fix(gate): resolve Slack paths under base path

jeong-sik:
### Motivation
- The Slack connector rewrite stopped resolving relative runtime/binding/audit paths under `MASC_BASE_PATH`, while the Python Slack sidecar still resolves relative paths against `MASC_BASE_PATH`, causing a cwd/base-path split that can leave bindings written to the wrong location and break authorization changes. 
- The intent is to ensure the server and sidecar agree on where relative connector state is stored so bind/unbind operations affect the file the active sidecar reads.

### Description
- Add `resolve_path` to `lib/gate/channel_gate_slack_state.ml` and make `slack_path` resolve both env overrides and defaults via `Env_config_core.base_path ()` when the path is relative, leaving absolute paths unchanged. 
- Keep the `Channel_gate_binding_store` wiring but ensure `binding_store_path`, `binding_store_read_path`, and `binding_audit_path` now return base-path-resolved values. 
- Add a regression test `test_slack_default_paths_resolve_under_base_path` in `test/test_channel_gate_connector_routes.ml` that sets `MASC_BASE_PATH`, switches the process cwd, performs a Slack `bind`, and asserts the binding file is written under the base path (and not under cwd), and register the test in the existing suite.

### Testing
- Ran `git diff --check` locally which reported no whitespace/patch errors. 
- Attempted to run `dune runtest test/test_channel_gate_connector_routes.ml` but the `dune` binary is not available in the container so the unit test run could not be executed locally; the added regression test should be exercised by CI where `dune` is present.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4dd85be7008333b56a949a193a15fa)

### DIFF
--- a/lib/gate/channel_gate_slack_state.ml
+++ b/lib/gate/channel_gate_slack_state.ml
@@ -26,10 +26,15 @@ let default_binding_audit_path = ".gate/runtime/slack/binding_audit.jsonl"
 
 (* Slack has no Discord-style guilds; the bot token authorizes per-workspace.
    Path resolvers read an env override, else fall back to the default. *)
+let resolve_path raw_path =
+  if Filename.is_relative raw_path then
+    Filename.concat (Env_config_core.base_path ()) raw_path
+  else raw_path
+
 let slack_path ~env_var ~default () =
   match Env_config_core.raw_value_opt env_var |> Env_config_core.trim_opt with
-  | Some path -> path
-  | None -> default
+  | Some path -> resolve_path path
+  | None -> resolve_path default
 
 let status_path () =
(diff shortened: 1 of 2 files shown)


jeong-sik:
## 리뷰 — #23804 (Slack MASC_BASE_PATH SSOT)
검증 완료. OCaml Slack 측이 상대 경로를 `MASC_BASE_PATH` 없이 풀어 Python sidecar와 불일치하던 SSOT 위반을 `resolve_path`(상대→base_path concat, 절대 경로는 보존)로 정렬. env override 절대 경로 동작은 보존. 테스트(binding이 cwd가 아닌 base 하위에 write됨) 견고. 머지 OK.

PR #23807, opened 2026-07-09
fix(keeper): block normalized .masc paths

jeong-sik:
### Motivation
- Prevent a traversal/symlink normalization bypass that allowed writes resolving under the workspace `.masc` directory to slip past the raw-path `.masc` check, which could let attacker-planted keeper TOMLs be hot-materialized and started.
- Ensure normalized targets are rejected when they resolve into MASC internal state so materialization and keepalive startup remain limited to trusted bootstrap paths.

### Description
- Add `is_masc_internal_state_norm : root_norm:string -> target_norm:string -> bool` to detect normalized targets that resolve under the workspace `.masc` directory and strip trailing slashes before comparison.
- Reject normalized `.masc` targets in `resolve_keeper_target_path` and the read resolver so post-normalization traversal cannot bypass the internal-state guard by resolving into `.masc`.
- Apply the same normalized `.masc` rejection to projected tool paths in `keeper_tool_shared_runtime.ml` so projected writes cannot resolve into `.masc/config/keepers`.
- Export the new API in the `.mli` and add unit tests exercising `is_masc_internal_state_norm` in `test/test_keeper_alerting_path_norms.ml` covering `.masc` root, `.masc/config/keepers/*`, sibling-prefix non-regressions, and ordinary project files.

### Testing
- New `Alcotest` unit cases were added in `test/test_keeper_alerting_path_norms.ml` to validate `is_masc_internal_state_norm`, but the test suite could not be executed in this environment because `dune` is not installed (`dune build @test` failed with `dune: command not found`).
- Local lint/check (`git diff --check`) reported no diff-style problems in the patch files.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3e38aadc0c8333862e11ac801c0053)

### DIFF
--- a/lib/keeper/keeper_alerting_path.ml
+++ b/lib/keeper/keeper_alerting_path.ml
@@ -263,6 +263,13 @@ let is_masc_internal_state_path (raw : string) : bool =
       && (len = 12 || raw.[len - 13] = '/'))
 ;;
 
+let is_masc_internal_state_norm ~(root_norm : string) ~(target_norm : string) : bool =
+  let root_norm = strip_trailing_slashes root_norm in
+  let target_norm = strip_trailing_slashes target_norm in
+  let masc_root = Filename.concat root_norm Common.masc_dirname in
+  target_norm = masc_root || String.starts_with ~prefix:(masc_root ^ "/") target_norm
+;;
+
 let resolve_keeper_target_path
       ~(config : Workspace.config)
       ~(allowed_paths : string list)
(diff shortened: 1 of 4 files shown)


jeong-sik:
## 리뷰 — #23807 (.masc traversal/symlink bypass)
검증 완료. 본문 위협(normalized target이 raw-path check를 우회해 `.masc` 하위 write)을 `is_masc_internal_state_norm`(resolve 후 normalized target 재검사)으로 실제 닫음. write(`resolve_keeper_target_path`)+read(`resolve_keeper_read_path`)+projected(`resolve_projected_allowed_path`) 3경로 적용. 테스트 4종(masc root/config-keeper 차단, sibling `.masc-tools`/일반 파일 허용)로 boundary 정확. 머지 OK.

PR #23913, opened 2026-07-10
fix(goal): restore legacy status phase migration

jeong-sik:
## Summary

- derive a missing persisted `phase` from the already parsed legacy `status`
- preserve completed, paused, and dropped legacy goals during recovery
- reuse the existing closed status-to-phase bridge; no string heuristic added

## 48-hour merge audit finding

Merged conflict-resolution PR #23710 kept parsing the legacy `status` field but ignored its value when `phase` was absent, forcing every old row to `Active`/`Executing`. Completed and failed lifecycle records could therefore re-enter active work after migration or recovery. A Changes Requested review identified the regression before merge, followed by a contradictory same-head approval.

The existing four-state regression test already described the intended compatibility contract and failed on current main. This patch restores that exact bridge with `Result.map phase_of_goal_status legacy_status`.

## Validation

- focused wrapper build for `test/test_goal_store.exe`
- `test_goal_store.exe` — 12/12, including all four legacy status mappings
- `git diff --check`

[근거] #23710 review history, current `goal_of_yojson`, focused executable, 확인일시 2026-07-10 13:04 KST, 신뢰도 High.

### DIFF
--- a/lib/goal/goal_store.ml
+++ b/lib/goal/goal_store.ml
@@ -122,7 +122,7 @@ and goal_of_yojson = function
           in
           let phase =
             match Json_util.assoc_member_opt "phase" json with
-            | None | Some `Null -> Ok (phase_of_goal_status Active)
+            | None | Some `Null -> Result.map phase_of_goal_status legacy_status
             | Some phase_json -> Goal_phase.of_yojson phase_json
           in
           let verifier_policy =


jeong-sik:
현재 main `1e984f3cea`(#23902)에 이 Draft의 유일한 소스 변경인 `Result.map phase_of_goal_status legacy_status`가 동일하게 병합됐고, legacy status-to-phase 회귀 테스트와 근거 주석까지 함께 수렴했습니다. 최신 main 기준으로 이 PR에 고유 diff가 남지 않으므로 중복 머지를 피하기 위해 superseded로 닫습니다. 48시간 감사 finding #23710의 최종 수리 PR은 #23902로 기록합니다.

PR #24081, opened 2026-07-11
fix(dashboard): restore IDE metadata and v2 layout

jeong-sik:
## Summary

- load keeper-authored IDE regions after the active document commits, then project them into Source-view line ownership
- expose metadata loading/error/ready state and use keeper-lane scope when no repository scope is available
- align the live IDE shell with the keeper-v2 reference: activity/annotation/cursor rail, persistent execute drawer, compact Chat entry, and card-based activity presentation
- add a real annotation rail

## Root cause

The file-content and region requests started concurrently. When the file response committed after the region request, the document store invalidated the otherwise valid region response as stale. Even successful region responses were not projected into the ownership store used by the Source gutter. Separately, the legacy two-row conversation/activity CSS grid overrode the v2 single-rail layout and stretched the right-rail tabs vertically.

## Impact

The IDE now renders real keeper observation metadata in Source view, reports unavailable metadata explicitly instead of presenting a false zero state, and matches the reference IDE's default three-pane layout more closely without adding fabricated activity or terminal data.

## Validation

- `pnpm exec vitest run --config vitest.config.ts --no-file-parallelism --maxWorkers=1` for 10 IDE-focused files: 125 tests passed
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `git diff origin/main...HEAD --check`

Build retains the existing Cinzel runtime-font and large-chunk warnings; no new build errors were introduced.

## Adversarial follow-up (`91325f21f6`)

- preserved the existing Keeper Work, Persistence, Memory, and multi-turn Conversation rail as the default Work Context surface
- kept the Execute drawer present while requiring explicit `terminal=open` before starting the output stream
- removed the client-only `.DS_Store` filename heuristic; initial-file classification needs a typed workspace-tree contract
- replaced raw font sizes with design-token SSOT values
- verified 89 focused IDE tests, TypeScript typecheck, ESLint, production build, raw-font ratchet, and `git diff --check`

### DIFF
--- a/dashboard/design-system/headless-core/keeper-line-ownership.ts
+++ b/dashboard/design-system/headless-core/keeper-line-ownership.ts
@@ -9,7 +9,7 @@
 
 import { kSlot, type AgentColorSlot } from './agent-presence'
 
-export type KeeperEditKind = 'edit' | 'create' | 'refactor' | 'revert'
+export type KeeperEditKind = 'edit' | 'create' | 'refactor' | 'revert' | 'observed'
 
 export interface KeeperEdit {
   readonly file_path: string
--- a/dashboard/src/components/ide/code-document-store.test.ts
+++ b/dashboard/src/components/ide/code-document-store.test.ts
@@ -83,4 +83,64 @@ describe('createCodeDocumentStore', () => {
       repoId: 'masc',
     })
   })
+
+  it('notifies metadata subscribers as region loading settles', async () => {
+    fetchIdeRegionsMock.mockResolvedValueOnce([])
+    const store = createCodeDocumentStore({
+      file_path: 'a.ts',
+      language: 'typescript',
+      content: 'export {}',
+    })
+    let calls = 0
+    const unsubscribe = store.subscribeRegions(() => {
+      calls += 1
+    })
+
+    await store.loadRegions('a.ts')
+
+    unsubscribe()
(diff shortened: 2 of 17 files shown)


jeong-sik:
적대 리뷰 및 수선 재검증 — exact head `91325f21f63be1df87aa458d98245a33bdc959d1` 했어용.

- P0: 없음.
- P1: 없음. 기존 Keeper Work/Persistence/Memory 및 multi-turn Conversation rail 삭제를 되돌렸고, `keeper` query만으로 Execute stream이 시작되던 주석-동작 불일치를 `terminal=open` 명시 경계로 닫았습니다.
- P2: 없음. `.DS_Store` 한 문자열만 예외 처리하던 client heuristic을 제거했고 raw font size를 `--fs-*` 토큰 SSOT로 교체했습니다.
- P3: 없음.

검증: focused IDE 89/89, `pnpm run typecheck`, `pnpm run lint`, `pnpm run build`, `scripts/lint/no-raw-font-size-px.sh`, `git diff --check` PASS. 로컬 Dune은 실행하지 않았습니다.

병합성: GitHub상 MERGEABLE이지만 Draft + DNM이며 exact-head CI가 green일 때까지 병합 보류입니다.

[근거] exact source/test commands + GitHub head 확인, 2026-07-11 KST, High.

PR #24406, opened 2026-07-14
fix(runtime): fail loud when checkpoint resume throws

jeong-sik:
## Problem
A checkpoint resume exception was caught and silently replaced with a fresh agent build. That discarded the persisted conversation and made the visible request appear to vanish after a timeout or resume fault.

## Change
- select exactly one construction path: resume when a checkpoint exists, build only when it does not
- propagate resume exceptions instead of resurrecting an empty agent
- regression-test that the exception propagates and fresh build is never called

## Scope
3 files, +42/-10. No provider inference, retry heuristic, budget gate, string classification, or compatibility fallback.

## Validation
- ocamlformat: pass
- OCaml parse check: pass
- git diff --check: pass
- focused Dune: blocked before compilation because the shared opam switch is already pinned to OAS v0.212.0 while current MASC main declares v0.211.9; CI is the build authority for this slice.

### DIFF
--- a/lib/runtime/runtime_agent.ml
+++ b/lib/runtime/runtime_agent.ml
@@ -860,6 +860,11 @@ let decide_modality_reroute_for_runtime_candidates ~(assigned : Runtime.t)
        |> List.map (fun (runtime : Runtime.t) ->
          runtime.Runtime.id, input_capabilities_of_runtime runtime))
 
+let select_agent_result ~checkpoint ~resume ~build =
+  match checkpoint with
+  | Some checkpoint -> resume checkpoint
+  | None -> build ()
+
 module For_testing = struct
   let request_runtime_fields_on_base_config =
     request_runtime_fields_on_base_config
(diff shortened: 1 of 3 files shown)


jeong-sik:
CI triage: the blocking `Provider name hardcoding ratchet` failure is inherited from base commit #24405, not this 3-file diff. The same unchanged literals are reported as both new and stale after line movement. Tracked separately in #24407; this PR will not patch numeric allowlist entries or mix that base repair into checkpoint-resume semantics. Build/Test remains in progress.

jeong-sik:
**[reviewed:adversarial-pass]** sha `864976d7` / 3파일(+42/-10) 직독.

checkpoint resume 실패 시 **silent fallback**(warn log + build)을 제거 → **fail loud**(예외 전파). `select_agent_result ~checkpoint ~resume ~build` 추상(`Some checkpoint -> resume | None -> build ()`) + `For_testing` 노출로 testability 확보. `Eio.Cancel.Cancelled` re-raise 유지.

필요성: 정당. silent fallback은 resume 회귀/설정 에러를 조용히 build로 덮어 진단성 훼손. fail loud가 노 Silent Failure 기준에 정합. approve. (base main B&T pending — green 정착 후 머지.)

PR #24471, opened 2026-07-14
refactor(keeper): hard-cut exit-condition yield

jeong-sik:
## What

- Hard-delete MASC's removed OAS exit-condition coupling: `exit_condition`, `exit_condition_result`, `ExitConditionMet`, `Legacy_exit`, and the boundary/turn-count admission heuristic.
- Derive pending work only from exact lane facts: Keeper chat admission, active chat receipts, and a validated durable event-queue snapshot.
- Return queue snapshot/registry failures explicitly instead of treating unknown state as pending or silently continuing.
- Yield only at the OAS typed post-tool boundary while preserving checkpoint/tool-result replay data.

The deterministic layer does not judge task semantics, terminate the Keeper, pause other lanes, or apply risk/budget/string heuristics. It only transports exact lane handoff facts; the resumed Keeper LLM decides the next semantic action.

## Focused verification

```sh
scripts/dune-local.sh build lib/runtime/.masc_runtime.objs/byte/runtime_agent.cmo
scripts/dune-local.sh build \
  lib/.masc.objs/byte/masc__Keeper_agent_run.cmo \
  lib/.masc.objs/byte/masc__Keeper_turn_driver.cmo \
  lib/.masc.objs/byte/masc__Keeper_turn_driver_try_provider.cmo \
  lib/.masc.objs/byte/masc__Keeper_unified_turn_execution.cmo
git diff --check
```

All changed `.ml`/`.mli` files also pass parsing and `ocamlformat --check`. Focused tests cover empty queues, exact durable queued stimuli, explicit queue errors, and typed chat/durable reason projection.

## Baseline block (not claimed green)

Building `test/test_keeper_turn_outcome.exe` is currently blocked before the focused executable links by unrelated current-main OAS 0.212 residue:

- `lib/inference_utils.ml:32`: deleted `Agent_sdk.Context_reducer`
- `lib/keeper/keeper_context_core_accessors.ml:71`: deleted `Agent_sdk.Context_reducer`

Those files are outside this slice and are being repaired separately. No fake full-build green is claimed here.

## Slice

- Base: #24469 (`8bdb6870467a37122538eadae19a451989a107fa`)
- Diff: `+112 / -270` (`382` changed lines)

### DIFF
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -67,26 +67,14 @@ type autonomous_yield_reason =
   | Chat_waiting
   | Durable_stimulus_waiting
 
-type autonomous_yield_boundary =
-  | Yield_immediately
-  | Yield_after_current_turn
-
 type autonomous_yield_request =
-  { reason : autonomous_yield_reason
-  ; boundary : autonomous_yield_boundary
-  }
+  { reason : autonomous_yield_reason }
 
-let autonomous_yield_allowed_at_turn ~start_turn ~turn request =
-  match request.boundary with
-  | Yield_immediately -> true
-  | Yield_after_current_turn -> turn > start_turn
-;;
-
-let stop_reason_of_autonomous_yield ~turn request =
+let runtime_yield_reason request =
   match request.reason with
-  | Chat_waiting -> Runtime_agent.Yielded_to_chat_waiting { turns_used = turn }
+  | Chat_waiting -> Runtime_agent.Chat_waiting
   | Durable_stimulus_waiting ->
-    Runtime_agent.Yielded_to_durable_stimulus { turns_used = turn }
+    Runtime_agent.Durable_stimulus_waiting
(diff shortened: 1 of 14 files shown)


jeong-sik:
Conflict cleared against `main`: merged current `main` and kept the cooperative-yield model (main's side was the retired `ExitConditionMet`/`exit_condition_result` handling). Also fixed a latent type bug at `runtime_agent.ml:1220` — `| None, Error _ as error -> error` bound `error` to the `(probe_error, advanced_result)` tuple (`as` is lower precedence than `,`), returning a tuple where a result was expected; changed to `| None, Error e -> Error e`. This only surfaces when `lib/` is actually compiled — the PR's green CI did not build it (change-detection gap). Base retargeted to `main`; this branch now carries the whole yield stack, so #24469 is subsumed. Remaining local build errors are main's pending Context_reducer removal (Cluster C), unrelated to this stack. — resolved by Claude

PR #24499, opened 2026-07-14
refactor(keeper): hard-cut context estimator facade

jeong-sik:
## Summary
- remove the retired token-count and context-ratio facade from keeper context runtime
- keep exact provider/catalog context limits and message counts as observation only
- keep compaction owned by the typed MASC LLM summarizer
- continue overflow recovery on the typed `Prepared` result, never on a shrink/count heuristic

## Boundary
OAS no longer owns context reduction. This slice introduces no compatibility module, thresholds, ratios, or implicit stop/pause condition.

## Size
- 32 files
- 94 insertions / 305 deletions
- exactly 399 changed lines

## Validation
- `git diff --check`
- no compiled `Agent_sdk.Context_reducer` references under `lib/**/*.ml{,i}`
- focused build: keeper context core, post-turn, context runtime, and dashboard keeper targets

## Stack
- depends on #24493
- remains Draft until the complete OAS 0.212 repair stack is green.

PR #24500, opened 2026-07-14
docs(keeper): state LLM compaction continuation contract

jeong-sik:
## Summary
- state that MASC compaction is prepared by the configured LLM summarizer
- state that exact message counts are observation only
- state that continuation is selected by the typed compaction result, not shrink heuristics

## Size
- 1 file
- 2 insertions / 3 deletions

## Validation
- `git diff --check`

## Stack
- depends on #24499
- remains Draft until the complete repair stack is green.


jeong-sik:
Superseded by the independently compile-coherent recut in #24493. The 5-line contract wording will be rebased as a later cleanup slice after #24493 instead of keeping the obsolete stack base.

PR #24524, opened 2026-07-15
refactor(schedule): make delivery wake-only and durable

jeong-sik:
## What changed

- remove direct Board mutation from the Scheduler consumer
- accept only typed Keeper wake payloads at the production dispatch boundary
- require durable queue commit before wakeup, reaction ledger, or success receipt
- reject persisted legacy Board schedules explicitly without mutating Board state

## Why

Scheduler owns time and occurrence signaling, not product effects or standing authority. A queue write failure was also logged and then reported as dispatch success.

## Stack

This is the effect-boundary slice. A small child PR removes the now-inaccessible legacy `board_*` parser and Board payload constant so the creation surface also becomes fully product-blind.

## Validation

- `scripts/dune-local.sh build test/test_schedule_consumer_dispatch.exe test/test_schedule_tool_wiring.exe`
- Schedule consumer dispatch: 9/9 passed
- Schedule tool wiring: 7/7 passed
- `git diff --check`

Diff: +81/-279, 360 total changed lines.

### DIFF
--- a/lib/keeper/keeper_unified_prompt.ml
+++ b/lib/keeper/keeper_unified_prompt.ml
@@ -288,7 +288,7 @@ let format_scheduled_automation_summary
            Buffer.add_char ubuf '\n')
         summary.items;
       Buffer.add_string ubuf
-        "- Use masc_schedule_get for details; each consumer applies its own leaf Gate before an external effect.\n");
+        "- Use masc_schedule_get for details; a due Schedule wakes the Keeper lane and grants no effect authority.\n");
     Buffer.add_char ubuf '\n';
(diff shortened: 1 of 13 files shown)


jeong-sik:
Stack child opened: #24525 removes the remaining `board_*` Tool parser and `Schedule_supported_kinds.board_post`, leaving the Scheduler surface product-blind.

jeong-sik:
Next stack child: #24529 preserves each durable schedule occurrence ID across queue, reaction ledger, dashboard evidence, and Keeper observation. #24525 is already merged into this parent branch.

PR #24539, opened 2026-07-15
test(tool): purge dormant orchestration contract

jeong-sik:
## Why

`masc.tool_orchestration` has zero production callers, but `test_tool_job` canonized its deadline/attempt/coarse `write:any` policy as supported behavior. Goal 5 must use the live durable Keeper/Fusion owners instead.

## Change

- remove the legacy `test_tool_job` suite
- remove its only test-runner dependency on `masc.tool_orchestration`

## Evidence

- diff: +1/-261
- `git diff --check`: pass
- production caller audit: zero outside the dormant library
- focused local build remains pending while an unrelated bare Dune process is active; no bypass used

## Merge order

Merge this PR into `main` first. Only then will the 358-line code purge PR be opened against the refreshed `main`, followed by the 369-line in-flight stub purge. This keeps every PR below 400 lines and avoids collapsing the stack into its parent branch.

Draft until focused proof and CI are green.


jeong-sik:
Closing as superseded: #24534 was merged to main with the full three-slice purge (+1/-988), so this reissued first slice is now an empty duplicate. No additional code is needed here.

PR #24559, opened 2026-07-15
refactor(keeper): remove compaction inference deadline

jeong-sik:
## Outcome

MASC-owned Keeper compaction no longer invents a wall-clock inference deadline.

- removes the compaction-only timeout_sec API and environment-derived default
- removes Eio.Time timeout wrapping and clock-unavailable refusal
- preserves exact Runtime/Runtime Lane candidate order, provider failures, structured-plan validation, and explicit candidate exhaustion
- leaves cancellation to the owning Keeper lane and typed provider/runtime failures

## Scope

This is stack A only. Provider-overflow same-lane recovery/requeue is the next stacked PR; durable before/after/model/tool/turn dashboard projection follows separately.

Diff: +14/-37.

## Verification

- ocamlformat --check: pass
- git diff --check: pass
- static residue search for compaction timeout/deadline API and Eio timeout wrapper: pass
- Dune intentionally not run locally; CI is build/test authority per repository workflow.

# ci:skip-test-coverage

Rationale: this PR only removes the untested compaction-specific timeout/deadline contract and its dead control branches. It adds no behavior branch for a new test to exercise; preserving or recreating a test for the deleted legacy would produce a fake green.

PR #24565, opened 2026-07-15
fix(keeper): recover provider overflow in owning lane

jeong-sik:
## Outcome

Typed provider `ContextOverflow` now enters MASC-owned LLM compaction in the failing Keeper cycle, then requeues the exact owning stimulus on that Keeper lane.

- adds a typed `Provider_overflow` compaction trigger carrying only the provider-declared limit
- dispatches overflow/start, runs the existing MASC LLM summarizer, and emits completion only after the compacted checkpoint is durably saved
- settles the original lease as `Context_compaction_retry` at the FIFO tail; it is never acked or replaced by a generic failure-judgment stimulus
- keeps compaction failure explicit and retryable, releases the buffer latch, and preserves cancellation cleanup/backtraces
- removes the obsolete `Compacting -> Overflowed` failure transition so other work and peer Keeper lanes remain executable

The next Keeper cycle reloads the same trace checkpoint and admits the exact retained stimulus again. Queue settlement occurs only after the cycle returns, after the durable checkpoint boundary.

## Stack

- base: #24559 — removes the hidden compaction inference deadline
- this PR: provider overflow same-lane recovery/requeue
- next: durable before/after/model/tool/turn audit and dashboard projection

Diff against base: +274/-59.

## Verification

- `ocamlformat --check` on every changed OCaml file: pass
- `git diff --check`: pass
- focused tests added for typed trigger codec, failed-compaction Lane release, lease disposition mapping, exact stimulus identity, FIFO tail retention, and settlement JSON round-trip
- Dune intentionally not run locally; CI is the build/test authority per repository workflow.

### DIFF
--- a/lib/compaction_trigger/compaction_trigger.ml
+++ b/lib/compaction_trigger/compaction_trigger.ml
@@ -11,12 +11,14 @@ type t =
       { count : int
       ; threshold : int
       }
+  | Provider_overflow of { limit_tokens : int option }
   | Manual
 
 let to_label = function
   | Ratio_threshold _ -> "ratio"
   | Message_count _ -> "messages"
   | Token_count _ -> "tokens"
+  | Provider_overflow _ -> "provider_overflow"
   | Manual -> "manual"
 ;;
 
@@ -26,6 +28,12 @@ let to_human = function
   | Message_count { count; threshold } ->
     Printf.sprintf "messages(%d>=%d)" count threshold
   | Token_count { count; threshold } -> Printf.sprintf "tokens(%d>=%d)" count threshold
+  | Provider_overflow { limit_tokens } ->
+    Printf.sprintf
+      "provider_overflow(limit=%s)"
+      (match limit_tokens with
+       | Some limit_tokens -> string_of_int limit_tokens
+       | None -> "unknown")
   | Manual -> "manual"
 ;;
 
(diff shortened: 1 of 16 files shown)


jeong-sik:
## 리뷰 — typed provider ContextOverflow → lane compaction recovery (16파일)

**흐름 이해**: provider가 `ContextOverflow`를 던지면, **실패한 Keeper cycle에서 MASC LLM compaction을 돌리고 그 Keeper lane에 원 stimulus를 requeue**하는 typed 복구 경로요. lane-scoped(다른 lane 무영향).

### 코드 평가
- **typed trigger**: `type ... = Provider_overflow of { limit_tokens : int option }` — provider 선언 limit만 실음. `int option`이라 **미선언 시 None(fabricated default 없음)**. ✓
- **closed decoder**: `"provider_overflow"` + `limit_tokens`가 `Null|None→None`, `Int→Some`, **그 외 shape→None(reject)**. round-trip to_string/to_json 대칭. ✓
- **durable-before-complete**: "emits completion only after the compacted checkpoint is durably saved" → 완료 신호가 durable save 이후 = silent loss 없음. ✓
- **lane 격리**: 원 lease를 `Context_compaction_retry`로 settle 후 owning lane에 requeue. 다른 keeper 무영향(campaign P0). ✓

### P0/P1/P2/P3: 없음.
### CI: base=main, base-red(OAS catalog) 상속 예상.

PR #24564, opened 2026-07-15
refactor(keeper): hard-cut message tools to typed delegation

jeong-sik:
## Summary
- atomically replace the four registered `masc_keeper_msg*` tools with typed `masc_keeper_delegate`, `delegate_status`, `delegate_cancel`, and `delegate_list`
- require closed typed `target`, `capability`, and `run_ref` values; external responses no longer expose raw request-id/status fields
- preserve `Keeper_msg_async` as the single durable execution owner and run delegated work through the existing per-Keeper serialized lane
- keep dashboard/connector direct delivery as explicit private adapters, outside the registered name-dispatch surface
- move operator Keeper actions to non-blocking typed delegation

## No compatibility layer
There are no aliases or dual registration. Schema, catalog, descriptors, typed Keeper tool names, generic dispatch, authorization labels, and positive matrix/runtime tests switch together in this commit. Tests whose only purpose was the removed message schema were deleted.

The typed delegate callback receives the opaque invocation request directly; it does not construct or execute a legacy public message payload. The remaining `masc_keeper_msg` labels are confined to the private direct-delivery adapter and its historical telemetry, not registered/auth/descriptor aliases. A follow-up stack will hard-cut product-visible historical labels without removing direct chat behavior.

Stacked on #24563, which supplies the typed internal turn entrypoint. #24563 is stacked on #24554.

## Verification
- all changed OCaml files: `ocamlformat --check`
- `git diff --check`
- static registry/auth/descriptor sweep confirms only the four typed names are registered
- focused local Dune build was not started because unrelated bare Dune processes currently bypass the repository wrapper lock; PR CI is the build authority

Diff: 393 additions / 431 deletions.

PR #24578, opened 2026-07-15
refactor(gate): hard-cut typed Keeper tracking ACK

jeong-sik:
Stacked on #24564.

## What changed
- replace Gate message ACK request_id/status with closed tracking: Keeper_run(run_ref, result_contract) or Chat_receipt
- move Keeper invocation wire identity into a small shared leaf so Gate never depends on the parent Keeper implementation
- make private direct delivery and Board inference emit/validate the typed Keeper contract
- remove the legacy submission/entry projections and tests whose only purpose was the deleted contract
- keep the existing Keeper_msg_async durable owner, per-Keeper admission lane, rich direct-chat surface context, and connector chat receipts

## Adversarial invariants
- no compatibility alias or dual registration
- no status normalization or unknown-to-permissive fallback
- run_ref target must equal the admitted Keeper
- malformed tracking is explicit and observable
- diff vs stack base: +398/-454

## Verification
- ocamlc parse pass for every changed .ml/.mli
- isolated type/interface check passed for keeper_invocation_types and Gate_protocol
- git diff --check passed
- focused scripts/dune-local.sh build was correctly refused because an unrelated bare dune process is holding the machine-wide policy guard; CI remains build authority

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -89,6 +89,7 @@
   ;; RFC-0056 Phase 1N — Keeper_state_* deterministic FSM cluster extracted to
   ;; lib/keeper_state/. Parent still references bare Keeper_state_* modules.
   masc.keeper_registry
+  masc.keeper_invocation_types
   ;; Keeper-owned pure/type leaves extracted from lib/keeper/.
   masc.keeper_contract
   masc.keeper_hooks_oas_types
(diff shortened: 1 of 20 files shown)


jeong-sik:
Closing before review: the diff met the additions cap but exceeded the intended total changed-lines cap. I am splitting it into atomic stacked Draft PRs; no compatibility shim or partial protocol will be merged from this PR.

PR #24657, opened 2026-07-15
fix(keeper): make lifecycle locking Eio-safe

jeong-sik:
## Summary
- add a typed cross-context mutex that cooperatively serializes Eio fibers with Domain/non-Eio callers
- replace the per-Keeper lifecycle reservation Stdlib.Mutex critical section that spans suspendable persistence
- preserve lane-per-Keeper independence with ephemeron-keyed lock ownership
- compiler-enforce stack-bounded cooperative acquisition with OCaml `[@tailcall]`
- add deterministic Eio/Eio, Domain/Eio, independent-lock, and exception-release tests

## Root cause
TOML reconcile held a `Stdlib.Mutex` while `File_lock_eio`/`Eio.Path` could suspend. Another fiber on the same Domain then re-entered the OS-thread mutex and raised `Sys_error("Mutex.lock: Resource deadlock avoided")`.

## Validation
- `ocamlformat --check` on touched OCaml files
- `ocamlc -stop-after parsing lib/core/cross_context_mutex.ml`
- `git diff --check`
- `scripts/dune-local.sh build test/test_cross_context_mutex.exe`
- `_build/default/test/test_cross_context_mutex.exe` — 3 tests passed in 0.003s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - Removed the unused permissive `Activity_feed` JSON decoder surface; the live activity API remains encode-only and its filesystem aggregation, ordering, limit, and agent-filter contracts now have dedicated regression coverage (#23960).
 
 ### Fixed
+- Keeper lifecycle reservations now use cooperative cross-context locking, so
+  TOML reconciliation can suspend during durable persistence without a second
+  Eio fiber re-locking the same OS-thread mutex.
 - Auto Judge requests now persist the exact outer-turn causal context without interpreting it, and durable retryable judgments resume on an accepted provider attempt rather than waiting for a server restart.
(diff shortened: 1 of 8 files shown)


jeong-sik:
## 재리뷰 — current head `ab4c12a71b`

cooperative acquisition recursion에 OCaml 5.5 `[@tailcall]`을 부착해 무제한 lock 대기가 stack을 늘리지 않는다는 성질을 compiler warning 51로 강제했습니다. timeout, retry cap, sleep은 추가하지 않았습니다.

- focused build: pass
- cross-context mutex executable: 3/3 pass (0.003s)
- P0/P1/P2/P3: 없음

**판정: current head merge-ok.** `#24574`의 aggregate approval은 회수했고, 이 좁은 구현을 우선 채택합니다.

PR #24658, opened 2026-07-15
fix(schedule): replace bounded cron scan

jeong-sik:
## Summary

- delete the synchronous minute-by-minute Cron scanner and its arbitrary five-year horizon
- evaluate the existing fixed-offset Cron_v1 meaning by calendar day and sorted allowed hour/minute slots
- make termination an explicit parser/validator invariant: parsed fields are non-empty, and DOM-only schedules must admit a real Gregorian date
- reject impossible new schedules such as February 31 while preserving semantically invalid durable rows for typed per-schedule failure instead of corrupting the whole ledger
- keep Vixie DOM/DOW OR semantics and find the exact 2104 leap-day occurrence after 2096
- remove the now-unreachable `Search_exhausted` error constructor

## Stack

- base: #24653
- diff: 133 additions, 75 deletions (208 changed lines)

## Verification

- `ocamlformat` on every touched OCaml file
- `ocamlc -stop-after parsing` on every touched OCaml file
- standalone compile/link of `schedule_domain.mli/ml` on OCaml 5.5.0
- standalone `test_schedule_domain.ml`: **20/20 passed** in 0.058s
- `git diff --check`
- repo-local focused Dune remains safely unavailable because another session owns a long-running bare Dune process; CI is the integration authority

## Scope boundary

This preserves only the currently declared fixed-offset 5-field Cron_v1 semantics. It does not add IANA/DST, RRULE, iCalendar import/export, a new scheduling library, or Cron↔RRULE conversion. Those require the separate versioned recurrence IR and adapter seam stack.

PR #24661, opened 2026-07-15
refactor(schedule): version recurrence IR

jeong-sik:
## Summary

- make `recurrence_ir = V1 of recurrence` the private typed field stored by every schedule request
- persist an explicit recurrence IR envelope with `schema_version: 1` and `rule`
- never infer V1 when `schema_version` is missing; missing and unsupported versions remain typed decode evidence
- decode schedule rows independently so one rejected row never corrupts or blocks the rest of the ledger
- preserve rejected raw rows across every store write while excluding them from eligibility and dispatch
- expose rejected row count, ordinal, and explicit error in the scheduled-automation dashboard
- keep new request construction simple: callers still provide the validated V1 rule and the domain wraps it

## Stack

- base: #24658
- diff: 294 additions, 68 deletions (362 changed lines)

## Hard-cut behavior

Existing unversioned recurrence rows are intentionally **not** inferred as V1. Each becomes `Missing_schema_version` evidence and stops participating, while other versioned rows and Keeper lanes continue. Unsupported versions behave the same way. Their original JSON remains durably preserved.

## Verification

- `ocamlformat` on all touched OCaml files
- `ocamlc -stop-after parsing` on all touched OCaml files
- standalone compile + 20/20 execution of `test_schedule_domain.ml`
- standalone type-check of `schedule_store.mli/ml`, `test_schedule_store.ml`, and `test_schedule_runner.ml` against the new Domain/Store interfaces
- `git diff --check`
- focused repo Dune remains safely refused because another session owns a long-running bare Dune process; CI remains the integration authority

## Scope boundary

No abstract adapter registry/module type and no iCalendar parser were added. The next adapter PR must arrive with a real import/export path that maps to this versioned IR and rejects lossy conversions explicitly.


jeong-sik:
## 리뷰 — recurrence IR versioning (schedule, 10파일)
**흐름**: `recurrence_ir = V1 of recurrence`를 모든 schedule request의 private typed 필드로, 명시적 IR envelope로 persist.
### 평가
- versioned typed IR(V1) = 향후 스키마 진화 시 exhaustive match 강제. private 필드로 캡슐화. ✓

PR #24626, opened 2026-07-15
test(keeper): prove compaction lane isolation

jeong-sik:
## Outcome

Proves the MASC-owned manual compaction request behaves as lane work, not as a synchronous tool mutation or fleet gate.

- an active turn serializes compaction for the same Keeper
- a peer Keeper lane remains independently runnable
- the owner compacts the latest checkpoint after the active turn releases
- the compacted checkpoint is immediately available for same-lane reinjection
- the runtime manifest retains exact before-message evidence

## Atomic follow-up settlement

When the immediate post-compaction turn reaches the typed LLM judgment route, the consumed manual request and its `Failure_judgment` successor now commit in one durable event-queue settlement. Retry/rotation of the following turn acknowledges the already-applied compaction instead of replaying it.

## Verification

- `ocamlformat` on all five touched OCaml files
- `ocamlc -stop-after parsing` on all touched implementations/interfaces
- `git diff --check`
- focused Dune wrapper was intentionally not bypassed: it refused while an existing bare `dune build lib/ test/deps/` owned the machine-wide build; PR CI remains the build authority
- 316 changed lines, below the 400-line stack limit

Stacked on #24618.

PR #24598, opened 2026-07-15
feat(keeper): run manual compaction on the owner lane

jeong-sik:
## Outcome

Manual compaction is now a durable Keeper-lane operation rather than a synchronous tool-side mutation.

- the tool durably enqueues one typed Manual_compaction_requested stimulus before waking the target Keeper
- duplicate requests collapse only while the same work is pending, leased, or awaiting outbox projection
- the owning Keeper performs compaction inside its turn admission slot
- another Keeper lane remains independently claimable while the owner lane is busy
- lifecycle, checkpoint recovery, manifest append failure, queue settlement, and follow-up turn outcomes remain explicit
- failed compaction requeues the exact stimulus with Context_compaction_retry
- once compaction is applied, a later turn failure cannot replay the already-applied compaction

## Why this PR was consolidated

The original root performed compaction directly from the tool and could continue after manifest failure. The stacked production and proof commits from #24607 and #24626 replace that transitional design. Keeping the intermediate state mergeable would briefly put an invalid execution boundary on main, so this oldest root now carries the safe atomic outcome.

## Boundary

MASC owns durable queueing, Keeper lane admission, lifecycle projection, and wake behavior. No OAS policy or compatibility path is introduced.

## Validation

- ocamlformat --check on all 23 touched OCaml files
- ocamlc -stop-after parsing on all touched implementations and tests
- git diff --check
- focused scripts/dune-local.sh build was requested but correctly refused because an existing bare dune build owns the shared machine; PR CI remains build authority

## Stack handling

- absorbs the production intent of #24607
- absorbs the lane-isolation proof of #24626
- descendants must be restacked from this new exact head after merge

PR #24664, opened 2026-07-15
fix(keeper): resume owner lane after overflow compaction

jeong-sik:
## Outcome

Provider-overflow compaction now resumes the owning Keeper lane immediately after durable source settlement instead of sleeping until the next cadence.

- separates a saved compaction from a failed compaction with typed source dispositions
- signals only the owner lane after the compacted source lease and transition outbox settle
- handles unleased proactive overflow by signaling the same owner lane directly
- leaves failed compaction work at the FIFO tail without an immediate hot-loop, so unrelated same-lane work can proceed
- replaces the premature source_requeued=true manifest claim with explicit owner_lane_resume_requested intent

## Adversarial proof

Before this change, Context_compaction_retry appended the exact leased stimulus to the durable FIFO tail, but neither settlement nor the no-lease path restored fiber_wakeup. interruptible_sleep therefore waited for cadence even though the compacted checkpoint was already durable. The same disposition was also used when compaction produced no checkpoint, so a naive wake fix would have created an immediate failure loop.

## Verification

- ocamlformat --check on all six changed OCaml files: pass
- ocamlc -stop-after parsing on changed implementations and test: pass
- git diff --check: pass
- focused Dune wrapper was not bypassed: it refused while existing bare Dune processes owned the machine-wide build lock; PR CI is the build/type-test authority
- 169 additions, 9 deletions; 178 total changed lines

## Stack

Base: #24626

### DIFF
--- a/lib/keeper/keeper_compact_policy.ml
+++ b/lib/keeper/keeper_compact_policy.ml
@@ -103,6 +103,15 @@ type compaction_evidence =
   ; after_tool_result_count : int
   }
 
+type compaction_receipt =
+  { operation_id : string
+  ; source_session_id : string
+  ; source_generation : int
+  ; source_turn_count : int
+  ; trigger : Compaction_trigger.t
+  ; evidence : compaction_evidence
+  }
+
 let compaction_evidence_to_json evidence =
   `Assoc
     [ "before_checkpoint_bytes", `Int evidence.before_checkpoint_bytes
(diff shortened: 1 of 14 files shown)


jeong-sik:
## 리뷰 — overflow compaction 후 owner lane resume (keeper, 6파일)
**흐름**: provider-overflow compaction이 next cadence까지 sleep하지 않고 **durable source settlement 직후 owning lane을 즉시 resume**.
### 평가
- durable settlement 후 즉시 resume(sleep 제거) = throughput↑, owning lane scoped(#24598/#24630/#24664 계열). durable-then-resume이라 유실 없음. ✓

jeong-sik:
#24884/#24888 및 main의 #24890 이후 구식 compaction lifecycle 계보이므로 닫습니다. manifest-before-settlement/owner wake 의미는 살리되, runtime manifest/meta/cooldown을 operation authority로 두지 않고 새 durable compaction Operation Journal + source digest/CAS 경로로 재구성해야 합니다.

PR #24691, opened 2026-07-15
fix(keeper): continue lane after compaction retry

jeong-sik:
## What changed

- map `Requeue Context_compaction_retry` to the existing immediate lane-drain follow-up
- exclude the just-requeued source lease so it cannot monopolize the lane
- add a focused regression proving unrelated ready work remains drainable

## Root cause

`Context_compaction_retry` was introduced as a typed durable settlement, but `ready_queue_followup_of_settlement` did not cover it. A successful compaction-retry settlement could therefore raise `Match_failure` before the Keeper lane selected its next ready item.

## Impact

The compacted source remains durably queued at the FIFO tail while other ready work in the same Keeper lane continues. No retry count, timeout, heuristic, or cross-Keeper coordination is added.

## Verification

- `ocamlformat --inplace` on all three changed OCaml files
- `ocamlc -stop-after parsing` on implementation, interface, and focused test
- `git diff --check`
- 34 changed lines, below the 400-line PR cap
- focused Dune target requested through `scripts/dune-local.sh`, which correctly refused because the existing machine-wide bare `dune build lib/ test/deps/` process (PID 66224) owns the build; full build/test remains CI authority

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop.ml
+++ b/lib/keeper/keeper_heartbeat_loop.ml
@@ -429,6 +429,7 @@ let ready_queue_followup_of_settlement ~lease = function
   | Keeper_registry_event_queue.Requeue
       ( Keeper_registry_event_queue.Rotate_now
       | Keeper_registry_event_queue.Retry_after_observed
+      | Keeper_registry_event_queue.Context_compaction_retry
       | Keeper_registry_event_queue.Approval_grant_unconsumed
       | Keeper_registry_event_queue.Approval_grant_state_unavailable ) ->
     Drain_ready_queue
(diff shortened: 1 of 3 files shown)


jeong-sik:
## 리뷰 — compaction retry 후 lane drain 지속 (keeper, 3파일)
**흐름**: `Requeue Context_compaction_retry`를 immediate lane-drain follow-up에 매핑하고 just-requeued source를 제외.
### 평가
- `Context_compaction_retry`를 `ready_queue_followup`에 연결(#24630/#24646 패턴), **just-requeued 제외로 self-spin 방지**. 회귀테스트(`test_context_compaction_retry_continues_lane_drain`). compaction retry 후 sleep 없이 lane 이어감. ✓

jeong-sik:
현재 변경은 #24630 head 81a189fd5027672cdb738812a55a051b7ad75f6f에 supersede되었습니다. #24630은 Context_compaction_retry를 exact single leased stimulus 제외로 처리하고, peer drain 및 source-only no-spin을 focused 14/14 테스트로 검증합니다. 중복 stack과 구형 multi-stimulus ancestry를 제거하기 위해 이 PR을 닫습니다.

PR #24575, opened 2026-07-15
feat(keeper): persist exact compaction evidence

jeong-sik:
## Outcome

Provider-overflow compaction now leaves an exact durable runtime-manifest receipt in the owning Keeper lane.

- records the exact Runtime candidate that produced the accepted LLM plan
- records checkpoint bytes, messages, summarized/dropped messages, and tool-use/tool-result blocks before and after
- links Context_compacted to the durably saved checkpoint
- links the next Context_injected receipt to that checkpoint only when the typed load boundary reports a loaded checkpoint
- preserves source stimulus requeue and explicit retryable failure receipts

Reinjection status is not guessed or stored as a boolean. The read model can derive it deterministically from the linked Context_compacted -> Checkpoint_loaded -> Context_injected event sequence.

## Stack

Base: #24565

Diff against base: +334/-33.

## Verification

- ocamlformat --check on every changed OCaml file: pass
- git diff --check: pass
- public runtime-manifest projection regression test added
- Dune intentionally not run locally; CI is the build/test authority per repository workflow.

PR #24694, opened 2026-07-15
feat(keeper): persist compaction operation receipt

jeong-sik:
## Summary
- derive a stable compaction operation identity from the exact checkpoint source version and typed trigger
- persist selected runtime and before/after evidence inside a copied checkpoint context before the checkpoint commit
- detect the durable receipt on retry and reuse the compacted checkpoint without invoking the LLM summarizer again
- expose the same operation_id in manual and provider-overflow manifest rows

## Invariants
- source context is not mutated
- a receipt matches only session, generation, turn count, and typed trigger
- malformed, duplicate, unknown, or negative receipt fields fail explicitly
- a later checkpoint version never reuses an older receipt

## Verification
- 381 additions, 12 deletions; 393 changed lines
- ocamlformat --check on all changed OCaml files
- ocamlc -stop-after parsing on all changed OCaml files
- git diff --check
- focused Dune wrapper refused because the existing bare Dune PID 66224 is active; CI remains build authority

## Follow-up
Manifest append is still a projection. A child PR will make projection idempotent by exact operation_id before source settlement.

## Stack
Base: #24664. Merge parent first, then retarget this PR.

### DIFF
--- a/lib/keeper/keeper_compact_policy.ml
+++ b/lib/keeper/keeper_compact_policy.ml
@@ -103,6 +103,15 @@ type compaction_evidence =
   ; after_tool_result_count : int
   }
 
+type compaction_receipt =
+  { operation_id : string
+  ; source_session_id : string
+  ; source_generation : int
+  ; source_turn_count : int
+  ; trigger : Compaction_trigger.t
+  ; evidence : compaction_evidence
+  }
+
 let compaction_evidence_to_json evidence =
   `Assoc
     [ "before_checkpoint_bytes", `Int evidence.before_checkpoint_bytes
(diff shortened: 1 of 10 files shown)


jeong-sik:
## 리뷰 — compaction operation receipt persist (keeper, 10파일)
**흐름**: checkpoint source version + typed trigger에서 **stable compaction operation identity**를 derive해 receipt로 persist.
### 평가
- `type compaction_receipt = { operation_id; trigger : Compaction_trigger.t }` typed, versioned key(`masc_compaction_receipt_v1`). `operation_id`는 checkpoint version+typed trigger에서 **stable derive**(random UUID 아님). ✓
- compaction 계열(#24575/#24598/#24694) receipt/provenance. durable persist. ✓

PR #24699, opened 2026-07-15
fix(supervisor): preserve child exit status

jeong-sik:
## Summary
- preserve the actual `start-masc.sh` exit status across the supervisor restart loop
- keep existing shell flags and restart/circuit-breaker semantics unchanged
- add a focused crash-loop regression proving exit 23 is logged and never rewritten to 0

## Verification
- `bash -n scripts/start-masc-supervised.sh`
- `shellcheck scripts/start-masc-supervised.sh`
- focused Dune build
- `test_start_masc_script.exe`: 26/26 pass
- `ocamlformat --check` and `git diff --check`

## Review
- adversarial review: P0/P1 = 0

### DIFF
--- a/test/test_start_masc_script.ml
+++ b/test/test_start_masc_script.ml
@@ -13,6 +13,9 @@ let script_path () =
 let loopback_script_path () =
   source_file "scripts/start-loopback.sh"
 
+let supervised_script_path () =
+  source_file "scripts/start-masc-supervised.sh"
+
 let quote = Filename.quote
 
 let read_file path =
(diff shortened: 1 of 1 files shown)


jeong-sik:
겹침 안내: #24693 (약 25분 먼저 오픈)이 동일한 exit-code 캡처 수리를 포함하고, 추가로 128+ 시그널 디코드(`signal=SIGTERM` 로그)와 takeover kill 귀속(breadcrumb writer/reader + victim/boot 로깅)까지 다룹니다. 두 수리는 shell 시맨틱상 등가입니다 (`if cmd; then 0; else $?; fi` ≡ 직접 `$?` — set -e 부재). 이 PR의 supervisor 루프 shell-level 테스트(`test_supervisor_preserves_child_exit_status_across_restart_policy`)는 #24693에 없는 가치이므로, #24693 머지 후 이 PR을 테스트-only로 rebase하는 쪽을 제안합니다 (반대로 이 PR이 먼저 머지되면 #24693에서 script hunk를 드랍하고 forensics만 남기겠습니다).

jeong-sik:
Rebased onto main (c22e50c6e9). Resolution: the script-side fix became OBE — #24693 landed the equivalent change (`|| true` removal + real `exit_code=$?` capture, with rationale comment), so this PR now carries only its unique value: the 55-line behavioral regression test (fake child exiting 23 → asserts `masc exited code=23` recorded, no `code=0` rewrite, crash-loop ABORT wording, stderr-only logging). Verified all asserted substrings against main's script (`scripts/start-masc-supervised.sh:85,102`, log() is >&2-only) and ran locally: `dune build --root . @test/runtest-test_start_masc_script` → 26 tests, all OK (31.9s), including the new case.

PR #24716, opened 2026-07-15
test(keeper): purge retired goal fixture

jeong-sik:
## What changed

- Remove the retired Keeper meta field `goal` from the Memory persistence test fixture.

## Why

The current strict Keeper meta parser explicitly rejects this removed field. As a result, every persistence test using the shared fixture failed before reaching its assertion. Keeping the field would test a deleted contract and make the suite falsely red.

This does not weaken an assertion or add compatibility parsing. It purges the legacy fixture so the tests exercise the current typed contract.

## Evidence

- Reproduced: `test_keeper_memory_write.exe` reported `removed keeper meta field is no longer supported: goal` and failed five persistence cases at fixture construction.
- `ocamlformat --check test/test_keeper_memory_write.ml`: pass.
- OCaml parser check: pass.
- `git diff --check`: pass.
- One deletion only.
- A new focused Dune invocation was not started while another bare Dune process owned the machine; CI remains the build authority.

### DIFF
--- a/test/test_keeper_memory_write.ml
+++ b/test/test_keeper_memory_write.ml
@@ -54,7 +54,6 @@ let make_meta name =
       (`Assoc
         [ "name", `String name
         ; "trace_id", `String ("trace-" ^ name)
-        ; "goal", `String "memory contract test"
         ])
   with
   | Error error -> Alcotest.fail ("meta fixture failed: " ^ error)


jeong-sik:
## 리뷰 — #24716 retired goal fixture purge (test, 1f)

test-only. 스키마에서 제거된 `goal` 필드의 잔여 fixture 정리(#24684 계열). 프로덕션 무변경, 현 파서 계약 정합.

### P0/P1/P2/P3: 없음. CI: base=main, test-only. base-red 상속, Build 확증.

PR #24715, opened 2026-07-15
test(keeper): prove configured compaction lane isolation

jeong-sik:
## What changed

Adds one focused integration regression around the configured-compaction producer introduced by #24714.

The test proves that:

- a configured message threshold is preserved as the exact typed trigger
- the durable commit wakes only the owning Keeper lane
- an unrelated ready `Bootstrap` queued behind configured compaction remains claimable
- claiming and acknowledging that ready work neither consumes nor loses the pending configured compaction

## Why

The production wiring must not turn context maintenance into a head-of-line blocker or a fleet-wide wake. This pins both lane isolation and selective ready-queue behavior with immutable fixtures.

## Stack

- base: #24714 `codex/configured-compaction-owner-lane-c3-20260716`
- this PR: focused proof only (82 added lines)

## Validation

- `ocamlc -stop-after parsing -impl test/test_keeper_post_turn_wirein_order.ml`
- `ocamlformat --check test/test_keeper_post_turn_wirein_order.ml`
- `git diff --check`
- focused Dune wrapper was attempted with a bounded wait and did not bypass the shared lock held by another worktree; CI remains the build authority

PR #24714, opened 2026-07-15
feat(keeper): queue configured compaction on owner lane

jeong-sik:
## What changed

- turns explicit Keeper compaction ratio/message/token settings into an exact typed trigger
- durably enqueues that trigger on the owning Keeper queue after a successful checkpointed turn
- wakes only the owning lane after the queue commit; identical pending configured requests deduplicate
- routes manual and configured requests through the same MASC-owned LLM compaction executor
- preserves typed queue persistence, reaction evidence, lifecycle projection, and settlement/retry behavior
- rejects persisted configured payloads carrying manual/provider-overflow triggers

## Why

Configured compaction was observable guard state but did not reach the durable Keeper lane. This made the setting effectively dead and left post-turn context pressure disconnected from the existing MASC LLM compactor.

The new path keeps OAS free of reducers/autotruncation and does not introduce a budget stop, risk tier, provider string match, or fleet-wide gate.

## Stack

- base: #24705 `codex/compaction-overflow-manifest-commit-c2c-20260716`
- this PR: production owner-lane wiring (399 changed lines)
- next stacked PR: focused owner-only wake and unrelated-ready-work test

## Validation

- `ocamlc -stop-after parsing` for every changed `.ml` / `.mli`
- `ocamlformat --check` for every changed OCaml file
- `git diff --check`
- focused Dune wrapper was attempted, then stopped without bypassing the shared lock held by another worktree; CI remains the build authority

PR #24719, opened 2026-07-15
fix(keeper): persist compaction operation projection

jeong-sik:
## What changed

- persists the exact compaction receipt `operation_id` in Keeper runtime metadata
- projects `count`, `last_ts`, `last_check_ts`, and applied decision immutably once per operation
- makes CAS retries and source-stimulus replays idempotent by comparing the exact operation identity
- reads the committed metadata back before continuing the owner turn, so that turn cannot overwrite the projection with its stale pre-compaction snapshot
- turns metadata write/read failure into a typed requested-compaction failure; the owning source lease is requeued rather than acknowledged
- exposes the operation identity in persisted meta and status JSON

## Why

#24714 made configured thresholds actionable, but the existing requested-compaction runner did not update `compaction_rt.last_ts`. After an Ack, the next completed turn could therefore enqueue another configured request despite the configured cooldown.

This patch binds the cooldown observation to the same durable operation identity already stored in the checkpoint receipt and manifest. It uses no token estimate, global mutable correction, string matching, or fleet-wide state.

## Stack

- #24714 production configured owner-lane wiring
- #24715 owner-only wake / ready-queue isolation proof
- this PR: exact operation projection and cooldown closeout (161 changed lines)

## Validation

- parser check for every changed `.ml` / `.mli`
- `git diff --check`
- focused tests added for operation-idempotent immutable projection and metadata-projection failure requeue
- repo wrapper correctly refused to start while bare Dune processes from other worktrees were active; no bypass was used, CI remains build authority

### DIFF
--- a/lib/keeper/keeper_heartbeat_loop_cycle.ml
+++ b/lib/keeper/keeper_heartbeat_loop_cycle.ml
@@ -225,10 +225,10 @@ let run_keeper_cycle_admitted
                ~trigger
            with
            | Error failure -> `Compaction_failed failure
-           | Ok _ -> `Run (obs, true))
+           | Ok success -> `Run (obs, true, success.Keeper_manual_compaction.meta))
         | None ->
           (match failure_judgment with
-          | None -> `Run (obs, false)
+          | None -> `Run (obs, false, meta_after_triage)
           | Some request ->
             (match
                prepare_failure_judgment_turn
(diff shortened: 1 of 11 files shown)


jeong-sik:
## 리뷰 — #24719 compaction operation projection persist (keeper, 11f)

compaction operation을 Keeper-local runtime manifest에 projection으로 durable persist (#24701 append-once / #24702 requeue / #24719 계열). typed, durable-before-ack, no-silent(manifest 실패는 명시 처리).

jeong-sik:
#24884의 durable compaction operation + CAS 계약으로 교체하기 위해 닫습니다. exact receipt/evidence projection은 보존 대상이지만 Keeper meta/runtime manifest가 settlement/retry authority가 되어서는 안 됩니다. 새 Operation Journal의 read projection으로 재추출해야 합니다.

PR #24721, opened 2026-07-15
refactor(memory): hard-cut heuristic bank compaction

jeong-sik:
## Outcome

Hard-cuts the only production invocation of heuristic Keeper memory-bank compaction.

- removes automatic threshold/priority/cap-driven bank rewriting from the post-turn path
- keeps durable tool-result writes and the LLM librarian boundary intact
- records the target contract: storage pressure may request a typed LLM Memory operation, but cannot decide what survives

This intentionally does **not** add a placeholder operation or an always-success settlement. Facade removal and the typed LLM operation/result are separate stacked PRs.

## Why

The removed path selected, merged, ranked, and dropped memories using configured byte/note thresholds, kind caps, generation/source bonuses, and priority scores. Those are policy heuristics, not an objective domain contract. They were also invoked after every turn through a best-effort lane and could fail without affecting settlement.

## Verification

- `python3 scripts/check_test_coverage.py` — pass without waiver
- `ocamlc -stop-after parsing` on every changed OCaml file
- `ocamlformat --check` on every changed OCaml file
- `git diff --check`
- focused `lib/masc.cmxa` build was already green before the deletion-only split; current full CI owns the rebased head

Diff: 14 additions / 44 deletions. No new executable branch is introduced; the covered-code additions are replacement contract comments, so the repository coverage checker passes on the actual diff without test deletion or opt-out.

### DIFF
--- a/docs/spec/12-memory-systems.md
+++ b/docs/spec/12-memory-systems.md
@@ -58,10 +58,12 @@ OAS reduces active context through its checkpoint/context APIs. MASC may
 request a configured strategy and observe the outcome, but must not rewrite
 the transcript through domain-specific text parsing.
 
-Keeper memory-bank maintenance preserves provenance and reports malformed or
-dropped records. An LLM librarian may classify or summarize candidates when a
-semantic judgment is required; deterministic code may validate schemas,
-enforce storage bounds, and order records by explicit timestamps.
+Keeper memory-bank rewriting is an explicit typed Memory operation. An LLM
+librarian returns the keep, rewrite, or forget decisions; deterministic code
+only validates their schema and provenance and atomically applies that exact
+plan. Storage pressure is observable and may request the operation, but a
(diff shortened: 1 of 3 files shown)


jeong-sik:
## 리뷰 — heuristic memory-bank compaction hard-cut (memory)
**흐름**: heuristic Keeper memory-bank compaction의 유일 production 호출(`Memory.compact_if_needed`)을 제거 — 자동 threshold/priority 기반.
### 평가 — heuristic 제거 (campaign 정합)
- `Memory.compact_if_needed`(threshold 초과 시 dedup+consolidate) + `val compact_if_needed` 제거. 주석: "threshold/priority score/capacity rule은 어느 memory를 compact할지 결정 못 함" → **heuristic 폐기**(no 휴리스틱). ✓
- 유일 production invocation 제거라 dead화 없음. ✓
### P0/P1/P2/P3: 없음. CI: base=main. Build 확증.

PR #24710, opened 2026-07-15
feat(keeper): retain caller completion receipts

jeong-sik:
## What changed

- persist a typed Keeper-invocation receipt inside the caller Keeper's existing event-queue v2 envelope
- atomically commit the exact `Bg_completed Keeper_invocation` stimulus and receipt
- make exact replay idempotent even after the pending stimulus was consumed
- reject a conflicting completion that reuses the same request identity
- retire a receipt only with the exact typed completion; absent acknowledgement is a storage no-op
- expose persistence and registry boundaries for the next terminal projector slice

## Why

#24703 persists the caller route and performs live completion wake-up, but a crash between the callee terminal record and caller queue commit can still lose delivery. This PR is the caller-lane exact-once acceptance primitive needed to close that two-store gap without a global coordinator.

## Scope boundary

This PR does **not** claim that the terminal-record to caller-queue crash gap is closed yet, and it does not claim full join/result reinjection.

The next stacked child must add terminal delivery state (`Pending` / `Delivered`), restart scanning, and a projector handshake:

1. replay a pending terminal through this acceptance API
2. mark the terminal source delivered
3. retire the exact caller receipt
4. wake only that caller Keeper lane

Typed `run_ref` / artifact delivery remains a later explicit join slice.

## Validation

- `ocamlformat -i` on all 7 touched OCaml files
- `ocamlc -stop-after parsing` on all touched `.ml` / `.mli`
- `git diff --check`
- focused regression covers durable reload, consumption, restart replay, conflict rejection, typed acknowledgement, and idempotent receipt cleanup
- focused Dune wrapper was attempted and correctly refused because PID 13857 was already running bare `dune build --root .`; no override or process kill was used. CI remains build/type authority.

Diff: 393 additions, 0 deletions.

PR #24722, opened 2026-07-15
feat(keeper): persist completion delivery state

jeong-sik:
## What changed

- Adds typed Delivery_pending and Delivery_delivered state to routed terminal Keeper invocation records.
- Persists the delivery state in the terminal JSON codec.
- Decodes predecessor routed terminal records without the additive field as Pending.
- Rejects completion_delivery on non-routed or non-terminal records.
- Adds an exact, monotonic mark_completion_delivered store transition under the existing per-request store and Keeper persistence locks.
- Treats Pending and Delivered as the same immutable terminal result for duplicate-partition reconciliation, while retaining exact request identity, terminal status, and completed_at checks.

## Why

This is C3a of the durable completion handshake. It establishes the source-side state machine without adding restart scanning, destination acceptance, receipt retirement, or lane waking. Those are isolated in the stacked C3b PR.

## Proof

The focused regression creates a real routed durable terminal request and verifies:
1. terminal publication starts Pending;
2. the source transition durably reloads as Delivered;
3. replaying a stale Pending proof is idempotent and cannot regress Delivered.

## Scope and size

- Base: #24710
- 153 changed lines: 148 additions, 5 deletions
- No projector or restart scanner in this PR
- No full run_ref, artifact, or terminal-result reinjection claim

## Validation

- ocamlformat check: green
- OCaml parser check for all changed ML/MLI files: green
- git diff --check: green
- Focused Dune build remains CI authority because the local wrapper is blocked by unrelated OAS pin drift; no pin bypass was used.

### DIFF
--- a/lib/keeper/keeper_msg_async.ml
+++ b/lib/keeper/keeper_msg_async.ml
@@ -38,6 +38,10 @@ type request_status =
       ; data : Yojson.Safe.t option
       }
 
+type completion_delivery =
+  | Delivery_pending
+  | Delivery_delivered
+
 type entry =
   { request_id : string
   ; request : Keeper_invocation_types.request
@@ -46,6 +50,7 @@ type entry =
   ; status : request_status
   ; submitted_at : float
   ; completed_at : float option
+  ; completion_delivery : completion_delivery option
   }
 
 type access_rejection =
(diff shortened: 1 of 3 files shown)


jeong-sik:
## 리뷰 — #24722 completion delivery state persist (keeper, 3f)

완료 delivery 상태를 durable persist(replay 가능). durable-before-ack 계열. typed.

jeong-sik:
#24884 기준으로 caller queue receipt + source delivery state의 two-store completion handshake를 더 확장하지 않기 위해 닫습니다. commit-before-wake, typed receipt, exact owner wake 순서는 새 Keeper Operation Journal lifecycle과 effect receipt로 이식해야 합니다.

PR #24727, opened 2026-07-15
refactor(memory): remove heuristic compaction facade

jeong-sik:
## Outcome

Removes the now-unreachable heuristic bank-compaction entrypoint from the public MASC Memory facade.

- deletes `Memory.compact_if_needed`
- deletes its facade-only summarizer alias
- leaves the legacy implementation itself for a later mechanical purge after all stacked callers are gone

This is a deletion-only follow-up to #24721. It adds no compatibility shim and no replacement stub.

## Verification

- `python3 scripts/check_test_coverage.py` against the stacked base — pass without waiver
- `ocamlc -stop-after parsing lib/memory.ml lib/memory.mli`
- `ocamlformat --check lib/memory.ml lib/memory.mli`
- `git diff --check`

Diff: 14 deletions.


jeong-sik:
## 리뷰 — #24727 heuristic compaction facade 제거 (memory, 2f)

#24721(유일 production invocation 제거) 후속 — 이제 unreachable해진 heuristic bank-compaction facade 진입점을 public MASC Memory facade에서 삭제. dead code 제거(재도입 없음).

Issue #24817, opened 2026-07-16
memory: replay accepted GC operation after restart


jeong-sik:
## Symptom

Memory OS GC work accepted by `Keeper_memory_lane.submit_if_idle` is owned only by the live server switch. A process restart cancels an in-flight GC unit, and restart does not reconstruct or immediately wake that exact operation.

## Reproduction condition

1. Start a GC tick for a Keeper and hold its GC I/O indefinitely.
2. Restart the MASC process before the unit settles.
3. The accepted unit is absent after restart; cleanup is retried only when a later periodic GC tick rediscovers the Keeper.

## Likely cause

The GC maintenance producer has no durable typed operation in the existing Memory work FIFO. The durable Memory stack in #24712, #24717, #24732, and #24737 currently covers post-turn Memory work, not GC maintenance.

## Attempted in the timeout-removal slice

- Removed the fleet-wide `Eio.Fiber.all` join and MASC-owned 120-second cancellation.
- Dispatched GC independently through the exact Keeper memory lane.
- Added deterministic Promise proof that one active Keeper does not block peer dispatch or the next tick.
- Deliberately did not invent a second queue, synthetic completion, or in-memory replay marker.

## Required direction

Add a typed GC operation to the existing durable Memory FIFO/owner drain, persist acceptance before execution, reconstruct pending/in-flight work on restart, and wake only the owning Keeper lane. Reuse the existing store; do not add another queue or wall-clock execution budget.

jeong-sik:
근본 해결책 단위 통합: 이 이슈는 #25868 (`outcome-wake-stimulus-loop`)로 합쳐졌습니다. 증상·재현 기록은 이 이슈에 보존되며, 통합 이슈의 fix로 해소되지 않는다고 판단되면 재오픈해 주세요.

PR #24808, opened 2026-07-16
fix(memory): remove consolidation execution deadline

jeong-sik:
## Outcome

- Deletes the MASC-owned 300-second consolidation execution deadline instead of renaming it.
- Runs consolidation even when no clock capability is available.
- Preserves provider transport timeout as a typed result and does not swallow parent cancellation in the consolidation runtime.
- Dispatches each Keeper through its existing Memory Lane without a fleet-wide Fiber.all join barrier.
- A stuck Keeper remains active only on its own lane; later ticks skip that Keeper and continue dispatching peers.
- Records every dispatch outcome, including already active, executor unavailable, and fork failed.

## Deliberate non-claims

- This does not yet introduce a durable maintenance operation with restart replay, terminal settlement, and Keeper wake.
- The separate Memory OS GC 120-second helper and its fleet-wide join remain follow-up cleanup.
- The existing consolidation enablement and cadence policy are not changed in this slice.

## Focused verification

- scripts/dune-local.sh build test/test_keeper_memory_lane.exe test/test_keeper_memory_os_consolidation_runtime.exe test/test_server_bootstrap_maintenance_memory_os.exe
- test_keeper_memory_lane.exe: 9/9 passed
- test_keeper_memory_os_consolidation_runtime.exe: 10/10 passed
- test_server_bootstrap_maintenance_memory_os.exe: 3/3 passed
- Full build and suite remain CI authority.

Stacked on #24807. This PR itself is 214 additions and 130 deletions.

PR #24807, opened 2026-07-16
feat(memory): add exact idle Keeper lane submission

jeong-sik:
## Outcome

- Reuses the existing per-Keeper Memory Lane as the single execution SSOT.
- Adds typed idle-only submission: active Keeper, unavailable executor, and fork failure remain distinct outcomes.
- Gives each accepted maintenance unit a child switch for provider and I/O resource ownership.
- An active Keeper skips only its own duplicate unit; other Keeper lanes remain independent.

## Boundary

This PR adds the generic lane primitive only. It does not add product-specific policy, numeric admission thresholds, or a second in-flight registry. The consolidation deadline hard-cut is stacked separately.

## Focused verification

- scripts/dune-local.sh build test/test_keeper_memory_lane.exe test/test_keeper_memory_os_consolidation_runtime.exe test/test_server_bootstrap_maintenance_memory_os.exe
- test_keeper_memory_lane.exe: 9/9 passed
- Full build and suite remain CI authority.

PR #24818, opened 2026-07-16
fix(memory): detach GC from fleet maintenance tick

jeong-sik:
## Outcome

- Removes the Memory OS GC `run_with_keeper_timeout` 120-second execution cancellation and its retired timeout metric.
- Removes the fleet-wide `Eio.Fiber.all` join from the GC maintenance tick.
- Dispatches each persisted Keeper's exact-expiry GC through `Keeper_memory_lane.submit_if_idle`.
- Records typed `started`, `already_active`, `executor_unavailable`, and `fork_failed` outcomes under a GC-specific metric and explicit log.

## Root cause

The old loop waited for every Keeper GC fiber and cancelled each one after 120 seconds. One blocked Keeper therefore held the whole maintenance tick until the timeout, while a valid long-running rewrite was terminated by a MASC-owned wall-clock policy.

The replacement tick owns no child work. Each accepted unit is owned by its Keeper memory lane, so the tick returns after dispatch, an active Keeper alone is skipped on later ticks, and idle peers continue independently.

## Stack

- Base: #24808 — consolidation uses the same nonblocking per-Keeper lane boundary.
- This PR: Memory OS GC timeout/join hard cut.

## Deterministic proof

The new Promise fixture holds the first Keeper GC indefinitely without any elapsed-time assertion. It proves that:

1. the peer Keeper completes and the first tick returns;
2. exactly one Keeper remains active;
3. the next tick skips only that active Keeper and runs the peer again;
4. releasing the held Promise drains both lanes cleanly.

## Residual, intentionally not faked

Accepted GC work is still process-owned rather than durably replayed after restart. This PR does not add a second queue or an in-memory replay marker. #24817 records the exact follow-up: extend the existing durable Memory FIFO/owner drain rather than inventing another SSOT.

## Verification

- `scripts/dune-local.sh build test/test_server_bootstrap_maintenance_memory_os.exe`
- `_build/default/test/test_server_bootstrap_maintenance_memory_os.exe` — 4/4 passed
- `GITHUB_BASE_REF=codex/memory-consolidation-no-timeout-v2 python3 scripts/check_test_coverage.py`
- `ocamlc -stop-after parsing` for every changed OCaml file
- `git diff --check`
- Stacked diff: 167 additions / 94 deletions = 261 changed lines

## Evidence

[근거] [Eio Fiber 1.3 official API](https://ocaml.org/p/eio/latest/doc/Eio/Fiber/index.html) — checked 2026-07-16, confidence High. `Fiber.fork ~sw` does not wait for the child, while `Fiber.all` waits for all supplied fibers; the implementation uses the former at the Keeper-owned lane boundary and removes the latter from the fleet tick.

### DIFF
--- a/lib/keeper_metrics/keeper_metrics.ml
+++ b/lib/keeper_metrics/keeper_metrics.ml
@@ -151,7 +151,7 @@ type t =
   | MemoryLaneInFlight
   | MemoryLaneProviderSlotBusy
   | MemoryBankCompactionFailures
-  | MemoryOsMaintenanceKeeperTimeout
+  | MemoryOsGcDispatches
   | MemoryOsConsolidationDispatches
   | WriteMetaCycleFailures
   | AlertPersistFailures
@@ -375,7 +375,7 @@ let to_string = function
   | MemoryLaneInFlight -> "masc_keeper_memory_lane_in_flight"
   | MemoryLaneProviderSlotBusy -> "masc_keeper_memory_lane_provider_slot_busy_total"
   | MemoryBankCompactionFailures -> "masc_keeper_memory_bank_compaction_failures_total"
-  | MemoryOsMaintenanceKeeperTimeout -> "masc_keeper_memory_os_maintenance_keeper_timeout_total"
+  | MemoryOsGcDispatches -> "masc_keeper_memory_os_gc_dispatch_total"
   | MemoryOsConsolidationDispatches -> "masc_keeper_memory_os_consolidation_dispatch_total"
   | WriteMetaCycleFailures -> "masc_keeper_write_meta_cycle_failures_total"
   | AlertPersistFailures -> "masc_keeper_alert_persist_failures_total"
--- a/lib/keeper_metrics/keeper_metrics.mli
+++ b/lib/keeper_metrics/keeper_metrics.mli
@@ -142,7 +142,7 @@ type t =
   | MemoryLaneInFlight
   | MemoryLaneProviderSlotBusy
   | MemoryBankCompactionFailures
-  | MemoryOsMaintenanceKeeperTimeout
+  | MemoryOsGcDispatches
   | MemoryOsConsolidationDispatches
   | WriteMetaCycleFailures
   | AlertPersistFailures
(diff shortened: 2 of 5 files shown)


jeong-sik:
## 리뷰 — Memory OS GC를 fleet maintenance tick에서 분리 (memory, 5f)

head diff 대조. **이것이 "Memory OS GC의 `run_with_keeper_timeout`(120s execution cancellation)과 fleet-wide `Eio.Fiber.all` join을 제거하고, 각 Keeper의 exact-expiry GC를 `Keeper_memory_lane.submit_if_idle`로 분산"하는 PR**이란 말이오. #24808(consolidation)의 GC 판이오.

### 평가 (on-disk 검증) — P0 격리
- **근본**: 옛 loop는 모든 Keeper GC fiber를 기다리고 120s 후 각각 취소 → **한 blocked Keeper가 whole maintenance tick을 timeout까지 잡고, valid long-running rewrite를 MASC-owned wall-clock으로 종료**시켰소. 이 PR: tick이 child work를 소유 안 함, 각 unit은 Keeper memory lane 소유, tick은 dispatch 후 반환, **active Keeper만 later tick에서 skip·idle peer는 독립 진행**. "하나 멈추면 전부" P0 회피. ✓
- `run_with_keeper_timeout`(120s `Eio.Time.with_timeout_exn`)+`MemoryOsMaintenanceKeeperTimeout` metric 제거(MASC-owned wall-clock 소거). ✓
- typed outcome(`started`/`already_active`/`executor_unavailable`/`fork_failed`) GC-specific metric+log. **Cancelled `as e -> raise e` 재던짐**(전파 보존). ✓

### P0/P1/P2/P3: 없음. 스택: base=#24808(consolidation) 계열. #24807 lane primitive 재사용.
**판정: merge-ok** → `reviewed:adversarial-pass`. fleet-wide join+wall-clock 제거로 GC의 P0 격리를 확보한 정합 수정이오.

jeong-sik:
부모 #24807/#24808 계보가 volatile Memory Lane을 execution SSOT로 사용하므로 함께 닫습니다. 120s cancellation과 fleet-wide join 삭제는 올바른 요구사항이며, owner Keeper Operation Journal producer로 옮긴 뒤 해당 의미만 다시 적용해야 합니다.

PR #24571, opened 2026-07-15
fix(runtime): make Provider the only LLM timeout boundary

jeong-sik:
## Summary
- route Keeper sub-calls through one OAS Provider boundary and delete feature-local timeout wrappers/config
- remove Masc_oas_bridge/Fusion outer wall deadlines, adaptive timeout enforcement, automatic fallback retries, and librarian parse retries
- treat retryable as observation only: failed Auto Judge work is never replayed by lane activity or process restart
- add an explicit one-shot Auto Judge retry action to the Gate API/dashboard
- bootstrap prompt state before resuming durable Auto Judge work
- preserve streaming/non-streaming distinction at OAS Provider transport

## Approval incident impact
- removes the 30s HITL summary inner timeout that produced durable failed approvals
- prevents prompt bootstrap race for new Auto Judge work
- existing failed approvals can be retried exactly once by explicit operator action, regardless of the historical retryable bit

## Verification
- OCaml parser check on all modified .ml files
- retired timeout/retry path source scans: zero hits in Keeper/Fusion/bridge
- dashboard tsc --noEmit
- focused Vitest: 31 passed
- git diff --check
- full Dune build/test delegated to PR CI by operator request

Companion OAS transport fix: https://github.com/jeong-sik/oas/pull/2607

PR #24823, opened 2026-07-16
fix(config): keep public runtime seed unassigned

jeong-sik:
## Summary

- remove the personal `nick0cave` keeper assignment from the public runtime seed
- keep personal and deployment routing in the live `<base>/.masc/config/runtime.toml`
- restore the existing public-seed contract enforced by `test_runtime_config_validity`

## Evidence

- #24571 CI run 29476594146 failed only because the public seed contained one keeper assignment while the contract requires zero
- `python3 tomllib` confirms `runtime.assignments` is empty
- `git diff --check`
- no local Dune build; PR CI is the build authority

### DIFF
--- a/config/runtime.toml
+++ b/config/runtime.toml
@@ -987,12 +987,11 @@ candidates = [
 # keepers fall back to [runtime].default. Per-keeper overrides in the
 # live config (<base>/.masc/config/runtime.toml) take precedence.
 #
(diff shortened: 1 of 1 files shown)


jeong-sik:
Closing as superseded by #24809, which merged the same public-seed assignment removal into main before this PR could be rebased. No unique change remains.

PR #24827, opened 2026-07-16
refactor(runtime)!: delete zombie eviction authority

jeong-sik:
## Stack

Depends on #24826.

## What

Hard-deletes elapsed-time zombie eviction authority and its dead support
surfaces:

- deletes timestamp-based zombie classifiers and Workspace re-exports
- deletes the final Keeper `is_zombie` wire/operator/dashboard projection
- deletes uncalled destructive Voice session cleanup
- deletes the orphan-task mutation and stop-Keeper callbacks owned by zombie GC
- replaces client-registry active-window classification with explicit
  registration lifecycle (`list_registered` / `total_count`)
- removes the unused fixed-interval registry cleanup fiber and renames the
  remaining cache repair to `cleanup_unregistered_session_caches`
- preserves `last_seen` as observation only
- logs client-registry projection failures instead of silently returning `[]`
- repairs docs that promised automatic zombie retirement or session reaping

`KH_zombie` and `Fiber_zombie` remain because they represent observed typed
runtime states, not an elapsed-time guess.

Automatic orphan inference is not restored. Instead, the operator profile now
has one explicit recovery boundary:

- `masc_operator_snapshot` exposes the exact persisted Task owner and backlog
  version
- `masc_operator_task_recovery_resolve` accepts only the exact Task ID, owner,
  backlog version, and a nonblank operator reason
- the Workspace mutation is a locked compare-and-set limited to `claimed` and
  `in_progress`
- stale version, owner mismatch, unsupported fields, and non-recoverable
  statuses fail explicitly
- no heartbeat age, elapsed-time threshold, Keeper name shape, status file, or
  liveness guess participates in the decision
- generic MCP/direct-call surfaces cannot bypass the operator profile
- success and failure outcomes are submitted to the immutable operator audit;
  an audit-recording failure is returned explicitly in the tool result

## Test policy

The delta deliberately removes broad or tautological tests. It keeps only
regression-sensitive invariants:

- a non-owner cannot release another actor's claimed task
- a legacy `is_zombie=true` field cannot regain lifecycle authority
- registration/listing has an exact count and identity contract
- reset clears cached session mappings
- operator Task recovery rejects stale versions and wrong owners
- the operator command rejects heuristic/unknown fields and executes only the
  exact snapshot-observed compare-and-set
- the recovery tool remains unavailable outside the operator profile

Deleted as non-evidence:

- cleanup immediately after an empty reset returns zero
- cleanup immediately after creating a fresh session returns zero
[...]

PR #24850, opened 2026-07-16
refactor(keeper): remove heartbeat cadence ceiling

jeong-sik:
## Outcome

- removes the hidden `[5, 300]` rewrite from the Keeper heartbeat cadence
- accepts every positive operator-selected integer exactly as configured
- rejects non-positive values explicitly instead of silently clamping them
- removes the runtime-settings maximum and surfaces `max_value = null`
- removes the orphaned Board debounce range comment left without an implementation

## Why

`MASC_KEEPER_HEARTBEAT_INTERVAL_SEC=3600` previously became `300`, and runtime updates above `300` were rejected. That ceiling was not a correctness invariant; it was scheduler authority derived from an arbitrary number. The only objective invariant for the existing integer sleep contract is a positive interval.

This change does not disable event-driven Keeper wakeups. It only stops the periodic cadence configuration from being rewritten.

## Verification

- `scripts/dune-local.sh build test/test_work_as_heartbeat.exe test/test_runtime_params.exe`: pass
- `test_work_as_heartbeat.exe`: 26/26 pass
- `test_runtime_params.exe`: 22/22 pass
- `3600` override persists and restores unchanged
- `0` returns an explicit validation error
- `git diff --check`: pass
- generated `docs/runtime-tunables.md`: 271 entries

### DIFF
--- a/docs/runtime-tunables.md
+++ b/docs/runtime-tunables.md
@@ -52,20 +52,20 @@ the categorization roadmap. Newly-added typed getters in
 
 | Env var | Kind | Category | Ops class | Line | Doc |
 |---|---|---|---|---|---|
-| `MASC_CONTEXT_RATIO_HARD_CAP` | typed:float | unclassified | unclassified | 642 | {1 Context Ratio Hard Cap} Absolute ceiling for compaction ratio_gate and handoff threshold after multiplier adjustme... |
-| `MASC_DASHBOARD_HEALTH_CTX_CRITICAL` | typed:float | unclassified | unclassified | 651 | {1 Dashboard Health Thresholds} Thresholds used by the dashboard keeper health scorer and harness health panels.  Dis... |
-| `MASC_DASHBOARD_HEALTH_CTX_WARN` | typed:float | unclassified | unclassified | 652 | {1 Dashboard Health Thresholds} Thresholds used by the dashboard keeper health scorer and harness health panels.  Dis... |
-| `MASC_DASHBOARD_HEALTH_PENALTY_CRITICAL` | typed:float | unclassified | unclassified | 653 | {1 Dashboard Health Thresholds} Thresholds used by the dashboard keeper health scorer and harness health panels.  Dis... |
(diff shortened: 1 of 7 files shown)


jeong-sik:
## 리뷰 — heartbeat cadence ceiling 제거 (keeper, 7f)

head diff 대조. **이것이 "Keeper heartbeat cadence의 hidden `[5, 300]` clamp(operator가 `MASC_KEEPER_HEARTBEAT_INTERVAL_SEC=3600` 설정→300으로 조용히 재작성, 300 초과 update는 reject)를 제거하고, 양수는 정확히 수용·비양수는 explicit reject하는" PR**이란 말이오. 그 ceiling은 correctness invariant가 아니라 arbitrary number에서 나온 scheduler authority였고, 유일 objective invariant는 positive interval이오.

### 평가 (on-disk 검증) — clamp→reject 전환의 정석
- **silent clamp 제거**: `max 5 (min 300 (...))`·`~min:5 ~max:300`·`max_value = Some 300` 삭제. operator config를 조용히 재작성하던 Silent Failure 소거. ✓
- **비양수 explicit reject(clamp 아님)**: `if interval_sec > 0 then [use] else Config_error "must be a positive integer"` / runtime_settings `Error "must be a positive integer"`. **값을 조용히 바꾸는 clamp가 아니라, invalid 입력을 loud 거부**(parse-don't-validate). ✓
- **arbitrary ceiling 소거**: `max_value = None`, doc "positive integer, no implicit upper bound" / "scheduler must not silently rewrite." ✓
- event-driven wakeup 무관(주기 cadence config만 rewrite 중단). 3600 persist·0 reject 테스트, 26/26+22/22. ✓

### 반경: 앞선 remove-limit 스택(#24827/24834/24846/24848)과 동일 테마 — arbitrary duration/count authority 제거, 진짜 invariant만 typed 강제.
### P0/P1/P2/P3: 없음. base=main.
**판정: merge-ok** → `reviewed:adversarial-pass`. silent clamp를 explicit reject로 전환하고 arbitrary ceiling을 제거한 모범 수정이오.

PR #24903, opened 2026-07-16
refactor(mcp): carry exact tool invocation identity

jeong-sik:
## Summary

- Add an abstract, lossless MCP request-id type for string/integer IDs; reject null and floating JSON numbers.
- Carry `Tool_invocation_ref.t` from JSON-RPC admission through MCP execution and the Keeper tool boundary.
- Preserve the typed request ID in observability JSON, so string `"1"` and integer `1` remain distinct.
- Expose the typed producer reference on `masc_keeper_compact` results without minting a replacement ID.
- Place `Tool_invocation_ref` in the cycle-safe wrapped-false `masc.tool_invocation_ref` leaf, depending only on `masc.mcp_transport_protocol` and `yojson`.

## Boundary and SSOT

- `Tool_invocation_ref.t` is correlation data only. This PR adds no writer, journal, cache, global map, counter, timestamp/hash identity, authorization, or routing heuristic.
- It does not make `Durable_event`, `Raw_trace`, or `Event_bus` an operation writer. The Operation Journal follow-up can depend directly on the leaf and embed this typed value during the single-writer hard cut.
- The root `masc` library depends on the leaf; the leaf does not depend on root `masc`, Keeper, or the future Operation Journal.
- The current OAS `Agent_sdk.Tool` handler boundary drops the exact provider `tool_use_id` and exposes no non-`Raw_trace` execution/scope reference. The OAS-origin variant is intentionally not fabricated here.
- MCP requires request IDs to be unique within a session. Duplicate-ID admission enforcement is tracked separately in #24901; this PR preserves the exact transport pair but does not claim externally validated uniqueness before that fix.

Spec: https://modelcontextprotocol.io/specification/2025-11-25/basic/index

## Validation

- `scripts/dune-local.sh build test/test_mcp_server_eio_coverage.exe test/test_mcp_server_eio_call_tool.exe test/test_keeper_adversarial_review.exe`
- `test_mcp_server_eio_coverage.exe`: 53 passed
- `test_mcp_server_eio_call_tool.exe`: 10 passed
- `test_keeper_adversarial_review.exe`: 14 passed
- Diff size: 342 additions + 48 deletions = 390 changed lines

### DIFF
--- a/docs/KEEPER-CAPABILITY-MATRIX.md
+++ b/docs/KEEPER-CAPABILITY-MATRIX.md
@@ -69,9 +69,9 @@ Important families:
 - scheduled automation: `masc_schedule_create`, `masc_schedule_list`,
   `masc_schedule_get`, `masc_schedule_cancel`, `masc_schedule_approve`,
   `masc_schedule_reject`
-- keeper management/messaging: `masc_keeper_list`, `masc_keeper_status`,
-  `masc_keeper_msg`, `masc_keeper_msg_result`, `masc_keeper_msg_queue`,
-  `masc_keeper_msg_cancel`
+- keeper management/delegation: `masc_keeper_list`, `masc_keeper_status`,
+  `masc_keeper_delegate`, `masc_keeper_delegate_status`,
+  `masc_keeper_delegate_list`, `masc_keeper_delegate_cancel`
 - operator/maintenance keeper controls: broader descriptors such as
   `masc_keeper_compact`, `masc_keeper_clear`, `masc_keeper_sandbox_start`,
   `masc_keeper_sandbox_stop`, `masc_keeper_reset`,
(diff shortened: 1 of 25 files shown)


jeong-sik:
이 PR은 코드 결함 때문에 닫는 것이 아니라 review surface가 불필요하게 합쳐진 원본이기 때문에 supersede합니다.

- exact invocation identity core는 #24957로 byte-identical 분리됨
- 문서 교정은 #24958로 분리됨
- #24957 + #24958의 최종 tree가 이 PR head와 동일함

동일 권위를 세 PR에 유지하면 stack/CI/리뷰 SSOT가 갈라집니다. 앞으로 #24957과 재기반할 #24958만 권위로 사용하고 이 원본은 닫습니다.

PR #24922, opened 2026-07-16
feat(keeper): decode compaction wire values

jeong-sik:
## Summary
- add strict object and field decoding for compaction wire values
- delegate checkpoint, evidence, trigger, producer, and turn semantics to their typed SSOT decoders
- reject unknown, duplicate, missing, and wrong-type nested fields explicitly
- intentionally exclude public event decoding from this stack leaf

## Verification
- focused build: `test/keeper_compaction_operation/test_keeper_compaction_operation.exe`
- focused test: 10/10 green
- `ocamlformat --check ...`
- `git diff --check`

## Stack
- base: #24918
- changed lines: +330/-0

PR #24923, opened 2026-07-16
feat(keeper): decode compaction operation events

jeong-sik:
## Summary
- expose one closed `Keeper_compaction_operation_codec` with canonical `to_json` and strict `of_json`
- keep decode support private after wiring the public codec
- enforce exact top envelope, event payload, and nested wire shapes
- roundtrip every event and sub-shape, including both failures, both reconciliation reasons, provider limit null, producer null, and reinjection Turn_ref

## Verification
- focused build: `test/keeper_compaction_operation/test_keeper_compaction_operation.exe`
- focused test: 10/10 green
- malformed proof: top unknown/duplicate/missing/wrong-type, payload unknown, nested trigger unknown
- `ocamlformat --check ...`
- `git diff --check`

## Stack
- base: #24922
- changed lines: +348/-37 (385 total)

### DIFF
--- a/lib/keeper_compaction_operation_codec/dune
+++ b/lib/keeper_compaction_operation_codec/dune
@@ -4,13 +4,17 @@
  (name masc_keeper_compaction_operation_codec)
  (public_name masc.keeper_compaction_operation_codec)
  (wrapped false)
- (modules keeper_compaction_operation_codec_support)
+ (modules
+  keeper_compaction_operation_codec
+  keeper_compaction_operation_codec_support)
+ (private_modules keeper_compaction_operation_codec_support)
  (libraries
   masc.compaction_trigger
   masc.keeper_checkpoint_ref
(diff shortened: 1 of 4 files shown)


jeong-sik:
개별 diff 리뷰. of_json: Yojson -> (event, decode_error) result. Result 반환으로 decode 실패 typed. 정당. 구조 제안은 #24915. 이견 없음.

jeong-sik:
P1 library-granularity 대응: 이 rung은 #24963 단일-library replacement로 대체합니다. 기능은 누적 보존하되 `masc.keeper_compaction_operation_*` public library 분할을 제거하고 identity/reducer/codec/record/store/action-selector를 하나의 `masc.keeper_compaction_operation` 내부 모듈로 통합했습니다. #24963 focused build 및 24/24 테스트 green.

PR #24926, opened 2026-07-16
feat(keeper): terminalize superseded compactions

jeong-sik:
## Why

A non-blocking Keeper may advance, replace, or explicitly remove its canonical checkpoint while an asynchronous compaction works from an older exact source. Retrying that immutable operation would pin it to a source that can no longer become current. A candidate can also commit and then be replaced before any turn adopts it.

## Change

- add typed `Source_superseded` operation fact and terminal `Superseded` phase
- accept supersession before candidate preparation, after preparation/reconciliation, or after commit-before-reinjection
- represent an explicitly absent source as `observed_checkpoint = null`; I/O/parse faults do not use this terminal path
- retain the exact observed checkpoint, prepared candidate evidence, and prior committed checkpoint when applicable
- reject false supersession when the observed ref is the unchanged source before commit, the installed candidate, or another trace
- extend canonical JSON projection and strict codec atomically

## Proof

- `ocamlformat --check` on touched files
- `scripts/dune-local.sh build test/keeper_compaction_operation/test_keeper_compaction_operation.exe`
- focused executable: 12/12 passed
- 341 changed lines; stacked on #24923

### DIFF
--- a/lib/keeper_compaction_operation/keeper_compaction_operation.ml
+++ b/lib/keeper_compaction_operation/keeper_compaction_operation.ml
@@ -28,12 +28,18 @@ type candidate =
   ; evidence : Keeper_compaction_evidence.t
   }
 
+type supersession =
+  { attempt_id : Attempt_id.t
+  ; observed_checkpoint : Keeper_checkpoint_ref.t option
+  }
+
 type event_view =
   | Requested of request
   | Attempt_started of Attempt_id.t
   | Candidate_prepared of candidate
   | Attempt_failed of Attempt_id.t * attempt_failure
   | Commit_reconciliation_required of candidate * reconciliation_reason
+  | Source_superseded of supersession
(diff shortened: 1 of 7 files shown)


jeong-sik:
개별 diff 리뷰. supersession/source_superseded terminalize. 명시적 typed 종료. 정당. 구조 제안은 #24915. 이견 없음.

jeong-sik:
P1 library-granularity 대응: 이 rung은 #24963 단일-library replacement로 대체합니다. 기능은 누적 보존하되 `masc.keeper_compaction_operation_*` public library 분할을 제거하고 identity/reducer/codec/record/store/action-selector를 하나의 `masc.keeper_compaction_operation` 내부 모듈로 통합했습니다. #24963 focused build 및 24/24 테스트 green.

PR #24937, opened 2026-07-16
feat(keeper): rebase compacted checkpoint over live delta

jeong-sik:
## Summary

- add a pure typed rebase boundary for a compacted checkpoint and concurrent append-only Keeper progress
- accept source and current only as Keeper_checkpoint_store.exact_checkpoint_snapshot, so checkpoint, exact canonical bytes, and reference cannot be mismatched by callers
- construct the compacted candidate internally from the exact source checkpoint plus compacted_messages, so compaction cannot alter source identity, metadata, context, usage, or turn coordinates
- return that candidate unchanged for an identical exact snapshot reference
- for the same trace and generation, require source.messages to be an exact structural prefix through Keeper_replay_prefix.split, then append the exact live suffix while preserving every non-message field from the exact current checkpoint
- fail explicitly with typed Source_superseded or prefix mismatch errors
- perform no checkpoint serialization/re-encoding in production and use no semantic, string, size, or time heuristic

## Verification

- scripts/dune-local.sh build test/keeper_compaction_live_delta/test_keeper_compaction_live_delta.exe
- focused Alcotest: 6/6 green (exact source, append-only delta, non-prefix, trace mismatch, generation mismatch, metadata/context/usage preservation)
- ocamlformat --check on all added OCaml files
- git diff --check

## Stack

- base: #24935 / codex/compaction-exact-snapshot-20260717 at 387b849e078949f645bd822795701f7b4c6e70e4
- leaf diff: 308 additions, 0 deletions

### DIFF
--- a/lib/keeper/keeper_checkpoint_store.ml
+++ b/lib/keeper/keeper_checkpoint_store.ml
@@ -589,6 +589,16 @@ type checkpoint_ref_load_error =
       }
   | Ref_lock_failed of string
 
+type exact_checkpoint_snapshot =
+  { checkpoint : Agent_sdk.Checkpoint.t
+  ; reference : Keeper_checkpoint_ref.t
+  ; canonical_bytes : string
+  }
+
+let exact_snapshot_checkpoint snapshot = snapshot.checkpoint
+let exact_snapshot_reference snapshot = snapshot.reference
+let exact_snapshot_canonical_bytes snapshot = snapshot.canonical_bytes
+
 type checkpoint_cas_error =
   | Source_unavailable of checkpoint_ref_load_error
   | Source_changed of Keeper_checkpoint_ref.t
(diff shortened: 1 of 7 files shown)


jeong-sik:
[추가] 같은 모듈(`keeper_compaction_live_delta`)을 건드리는 경쟁 PR **#24952**(base=`compaction-operation-store`, `agent/` 라인)가 있다. 본 PR은 base=`compaction-exact-snapshot`(codex 라인). 두 라인이 같은 rebase를 따로 구현 중이므로 한쪽 흡수/정리가 필요하다. 상세는 #24952 코멘트 참조.

jeong-sik:
중복 정리: 이 PR의 유효한 변경은 이미 #24952와 #24953으로 분리되어 동일한 책임을 더 좁게 표현합니다. 원 PR을 병렬 SSOT로 유지하면 어느 branch가 authoritative한지 불분명해집니다. 따라서 queue:superseded 기준으로 닫습니다. 후속은 #24952/#24953의 실제 base 의존성과 통합 순서만 검증합니다.

PR #24949, opened 2026-07-16
refactor(keeper): remove compaction config metadata

jeong-sik:
## Summary

- delete compaction profile/threshold fields from keeper metadata and keeper-up parsed arguments
- remove create/update/parser authority for those fields
- reject removed persisted metadata explicitly instead of accepting or silently scrubbing it

## Stack

- stacked after all public/read consumers are removed
- recuts 181 changed lines of #24919
- next: #24950

## Verification

- `scripts/dune-local.sh build lib/masc.cma test/test_heartbeat_integration.exe test/test_tools_coverage.exe`
- own diff: 181 changed lines

Full build and suite remain delegated to PR CI.


jeong-sik:
개별 diff 리뷰. keeper meta contract 삭제(10f). #24919와 byte-identical. 같은 적용.

jeong-sik:
P1 sequencing 대응: 이 config hard-cut은 닫습니다. 현재 replacement operation/journal/wake 경로가 main에 landing되기 전에 threshold/config authority를 삭제하면 자동 compaction 비활성 창이 생깁니다. 새 실행 경로의 terminal commit + wake + restart proof가 병합된 뒤, config 삭제는 마지막 tranche로 current main에서 다시 recut해야 합니다.

PR #24954, opened 2026-07-16
feat(keeper): expose durable compaction parking

jeong-sik:
## Summary
- expose exact durable `park_for_compaction_result` and `resume_parked_result` through the persistence and registered-Keeper queue facades
- expose persisted parked entries and project only nonterminal `Event_queue_parked` rows into current operations
- keep a parked source inside durable stimulus identity accounting after its transition outbox is projected
- correct the current event-queue envelope terminology to v3

## Invariants
- one atomic `event-queue.json` transaction remains the mutation authority
- park removes only the exact leased source; unrelated FIFO work remains claimable
- resume appends the exact source once and is idempotent
- `Resumed` audit rows do not masquerade as current work
- no timeout, polling cadence, execution cap, process-global state, or semantic string classification

## Verification
- `scripts/dune-local.sh build lib/masc.cma test/test_keeper_event_queue_state_v2.exe test/test_keeper_current_operations.exe`
- `scripts/dune-local.sh exec test/test_keeper_event_queue_state_v2.exe` (15 tests)
- `scripts/dune-local.sh exec test/test_keeper_current_operations.exe` (4 tests)
- `ocamlformat --check` on all changed OCaml files
- `git diff --check`

357 changed lines. Stacked on #24944.

PR #24944, opened 2026-07-16
feat(keeper): park compaction continuations

jeong-sik:
## Summary

- add typed `Park_for_compaction` settlement state keyed by `Operation_id`
- atomically remove the exact active lease and retain its exact stimuli in an immutable parked map
- resume parked stimuli at the pending FIFO tail exactly once; duplicate resume and settlement replay are idempotent
- fail with typed operation/lease mismatch errors instead of guessing or silently rebinding identity
- hard-cut durable JSON to required `keeper.event_queue.state.v3` with lossless parked-entry roundtrip and exact-field validation
- keep parked entries out of lease admission blocking; only the short transition outbox remains a single-writer projection fence

## Why

Async LLM compaction cannot keep a Keeper lease open for the whole model call because that blocks the lane. Acking or requeueing before compaction completion also loses the exact continuation identity. This pure state transition parks the leased work durably under the typed compaction operation, allowing unrelated work to proceed after projection and allowing exact continuation reinjection after commit.

## Scope

This PR is the pure event-queue state and codec slice. It adds no timeout, cap, heuristic, model/provider string match, or compaction actor. The persistence facade producer and lane actor wiring remain the next stacked PR.

Stacked on #24938.

## Validation

- `scripts/dune-local.sh build test/test_keeper_event_queue_state_v2.exe`
- `_build/default/test/test_keeper_event_queue_state_v2.exe` — 14/14 green
- `ocamlformat --check` on every changed `.ml`/`.mli`
- `git diff --check`
- exact diff: +375/-23 (398 changed lines)

## P0 self-review

No P0 found. Exact lease payload is retained, source/op conflicts are explicit, replay cannot create a second parked entry, resume cannot duplicate the source stimuli, parked state survives JSON roundtrip, and v2/missing parked state does not silently decode as empty.

PR #24960, opened 2026-07-16
fix(keeper): preserve active lease on registration

jeong-sik:
## Outcome

Keeper registration and restart no longer mutate every durable active lease into a synthetic `Registration_recovery` requeue.

- registration reads only the pending projection
- the exact active lease sequence, kind, payload, and claim time remain durable
- heartbeat intake resumes that same lease before claiming new work
- no recovery transition or duplicate pending stimulus is synthesized
- malformed durable state still fails the Keeper-local registration explicitly

This closes the crash window where a committed compaction `Requested` fact could lose its exact source lease before the lane actor parks it. Unrelated Keeper lanes remain independent.

## Boundary

No policy, timeout, budget, risk level, semantic matching, or cross-Keeper coordinator is added. The change deletes a registration-time mutation and reuses the existing strict pending read plus exact active-lease resume path.

## Verification

- focused build: `lib/masc.cma` and `test_keeper_event_queue_state_v2.exe`
- event queue state/persistence: 15/15
- registration test compares the complete durable state and exact lease JSON before/after
- ocamlformat check: green
- diff check: green
- 46 additions / 62 deletions

### DIFF
--- a/lib/keeper/keeper_registry_setup.ml
+++ b/lib/keeper/keeper_registry_setup.ml
@@ -456,11 +456,9 @@ let register_with_state_result
   let done_p, done_r = Eio.Promise.create () in
   let key = registry_key ~base_path name in
   match
-    Keeper_event_queue_persistence.prepare_registration_result
+    Keeper_event_queue_persistence.load_pending_result
       ~base_path
       ~keeper_name:name
-      ~settled_at:(Time_compat.now ())
-      ()
   with
   | Error detail -> Error (Registration_event_queue_unavailable { keeper_name = name; detail })
   | Ok initial_event_queue ->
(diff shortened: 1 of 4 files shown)


jeong-sik:
적대적 base audit 결과 이 PR은 닫습니다.

#24944/#24954가 stale Event topology를 다시 권위화한다는 P0 base 오염 판정으로 닫혔고, 이 PR은 닫힌 #24954 head를 직접 base로 사용합니다. active lease preservation 자체는 필요한 불변식이지만 현재 diff를 병합하면 거부된 parking/facade 계보를 되살립니다.

재구성 조건:
1. #24630 actual-claim fairness + typed internal lane-drain provenance 수술
2. #24646/#24934의 projector 비차단 경계 재검증
3. corrected Event stack 상단에서 parking leaf 재작성
4. 그 새 parking persistence 위에서 exact active lease registration-resume proof를 다시 recut

CI green은 이 의미상 base 오염을 해소하지 않습니다. 코드는 그대로 재병합하지 않고 corrected queue contract에서 필요한 최소 leaf만 새 Draft PR로 다시 만듭니다.

PR #25074, opened 2026-07-17
docs(keeper)!: delete retired max-turn RFC

jeong-sik:
## Outcome

- delete the retired per-call max-turn Draft RFC
- remove obsolete MASC_KEEPER_OAS_MAX_TURNS_PER_CALL authority from live docs
- keep current unbounded OAS loop documentation and unrelated eval/provider constraints

## Why

PR #24987 removed the production worker max-turn tombstones, but this Draft RFC still described OAS default max_turns=10 and the removed MASC environment keys as current behavior. Keeping it would recreate a second, false authority and contradict runtime warnings.

The separate worker-meta read/codec silent-failure boundary remains tracked by #25017 and is intentionally not mixed into this documentation hard cut.

## Verification

- current main includes #24987
- focused residue scan finds no MASC_KEEPER_OAS_MAX_TURNS_PER_CALL or max_turns_override in live docs/code/config
- git diff --check
- focused worker build and test_worker_runtime: 5/5 passed before this deletion-only recut


jeong-sik:
diff 리뷰. legacy RFC 제거(+0/-145). retired max-turn RFC 삭제. legacy 처단. 찬성.

PR #25136, opened 2026-07-18
test(gate): prove owner-local drain order

jeong-sik:
## What changed

- Expose the existing owner-scoped Auto Judge selection and claim operations through a test-only module.
- Prove same-owner work drains by durable sequence even when wall-clock timestamps disagree.
- Prove different Keeper owners can claim work independently while the next item for the same owner remains queued.

## Why

The ordering leaf needs a direct proof that Lane-per-Keeper FIFO is real and that one Keeper cannot serialize or stop unrelated Keepers. The proof targets `Keeper_gate` owner filtering and claims; it does not exercise or restore deleted product/connector dispatch.

## Exact evidence

- Head SHA: `7c02f2bb86`
- Diff: **85 changed lines** (`85 additions / 0 deletions`)
- Base: `codex/gate-order-v8-20260718` at `164e47ee92`
- Stable patch-id remains `44e3a707153b992a30c8b1c0b1d2620f0db1136e`
- `git diff --check`: PASS
- `ocamlformat --check`: PASS
- OCaml parser (`ocamlc -stop-after parsing`): PASS

## Local focused build status

The original reconstruction did not run the focused build because the installed `agent_sdk` was then `0.216.2` while main required `>= 0.216.3`; the pin guard was not bypassed. Exact-head CI is the build authority.

## Stack

This is stacked leaf 2/2. It must remain based on `codex/gate-order-v8-20260718` until leaf 1 lands.

[근거] local exact-head diff/format/parser commands and `gh pr view 25136`; checked 2026-07-18 17:05 KST; confidence High.

### DIFF
--- a/lib/keeper/keeper_gate.ml
+++ b/lib/keeper/keeper_gate.ml
@@ -920,3 +920,13 @@ let decide ?cycle_grant ~keeper_always_allow request =
       ();
     Unavailable reason
 ;;
+
+module For_testing = struct
+  let ready_auto_judges_for_owner ~base_path ~keeper_name entries =
+    ready_auto_judges_for_owner ~base_path ~keeper_name entries
+  ;;
+
+  let claim_auto_judge = claim_auto_judge
+  let release_auto_judge = release_auto_judge
+  let reset_active_auto_judges () = Atomic.set active_auto_judges Auto_judge_owners.empty
+end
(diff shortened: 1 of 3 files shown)


jeong-sik:
## Exact head `81a651fbc0` — code PASS, stacked

- Base is correctly #25135 (`codex/gate-order-v8-20260718`).
- Production behavior is unchanged; this leaf exposes a bounded `For_testing` surface and proves same-owner ordering plus simultaneous active ownership for different owners.
- `test_different_owners_claim_in_parallel` executes claims sequentially rather than using real Domain/fiber concurrency, so “parallel” is an overstatement. It is not a fake test: it still proves owner-local non-serialization in the active map.

Verdict: **P0 0 / P1 0 / P2 0 / P3 1 — do not merge to the feature base as a terminal state. Merge #25135 first, then retarget/rebase this leaf to main and merge after CI.**

[근거] exact head/source/DAG; checked 2026-07-18 KST; confidence High.

PR #25256, opened 2026-07-18
feat(observability): log resolved runtime knobs at boot

jeong-sik:
## Why

#25128 could not be diagnosed because **no live surface exposes the resolved value of the opt-in runtime timeouts**. Confirmed across attempts: process env is restricted (macOS), boot logs never print them, `/api/v1/keepers/*` and config endpoints 404, and `masc_config` returns only tool categories. So a knob CONFIGURED in `runtime.toml` (`[turn] stream_idle_timeout_sec = 120`) but not applied to `Keeper_runtime_resolved` is indistinguishable at runtime from an unset one.

That ambiguity is exactly what blocked #25128: `stream_idle_timeout_sec` is 120 in runtime.toml, yet the pool idle-timeout cancellation (`idle timeout after`) fires **0 times** in the live log — with no way to tell whether the resolved value is `Some 120.` (an enforcement/other issue, or a slow-but-alive stream that #24386 correctly allows) or `None` (a config-load gap).

## What

Emit `Keeper_runtime_resolved.to_yojson (current ())` once at boot, right after `init` freezes it (`server_runtime_bootstrap.ml`), via `Log.Runtime` at `Boundary` category. This is diagnostic observability of a config-resolution boundary — **not** a symptom counter and not a fix for the hang; it makes the resolved config answerable so the #25128 root can be determined from the next boot.

## Verification

- `server_runtime_bootstrap` compiles; the boot log uses only already-exposed functions (`init`/`current`/`to_yojson`).
- 2 unit cases pin the observability contract: the serialized resolved config names `stream_idle_timeout_sec` and `body_timeout_override_sec`, and serializes as a JSON object.
- `ocamlformat --check` clean.

RFC-WAIVED: observability-only boot log in `lib/server`; no behavior change, no new policy/default, no destructive op. Reads config through the existing `Keeper_runtime_resolved` surface.

Refs #25128 (unblocks its root determination).

https://claude.ai/code/session_012KXQWcg9KtaC7KruGQTnwX

### DIFF
--- a/lib/server/server_runtime_bootstrap.ml
+++ b/lib/server/server_runtime_bootstrap.ml
@@ -271,6 +271,18 @@ let create_server_state ~sw ~base_path ?input_base_path ~clock ~mono_clock ~net
        Log.Server.error "runtime.toml load failed: %s" msg;
        raise (Env_config_core.Config_error msg));
   Keeper_runtime_resolved.init ();
+  (* Boot-time observability: emit the resolved runtime knobs once, right after
+     they freeze. The opt-in timeouts (stream_idle_timeout_sec and the
+     body-timeout override) are None when unset; without this line a knob that
+     is CONFIGURED in runtime.toml but did not reach Keeper_runtime_resolved is
+     indistinguishable at runtime from an unset one — the exact ambiguity that
+     blocked diagnosing #25128 (idle timeout configured at 120s yet never
+     observed to fire). No existing surface exposes the resolved value. *)
+  Log.Runtime.info
+    ~category:Log.Boundary
+    "resolved runtime config: %s"
+    (Yojson.Safe.to_string
+       (Keeper_runtime_resolved.to_yojson (Keeper_runtime_resolved.current ())));
   Keeper_task_owner_backend.install_hooks ();
   let state =
     Mcp_eio.create_state_eio ~sw ~proc_mgr ~fs ~clock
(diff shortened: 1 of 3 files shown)


jeong-sik:
적대 리뷰 (fresh-context, head `f5d4d4eb`) — **PASS**. boot-time 관찰성 로그 (+61/0), 4축 검증:
- **secret 유출 없음**: `Keeper_runtime_resolved.t`는 두 float option + source enum("env"|"toml"|"default")만 직렬화 — raw env 값/credential이 JSON에 도달하는 경로 자체가 없어 redaction 불필요.
- **boot-once, hot path 아님**: `init()` 직후 `create_server_state` 내부, 유일 프로덕션 caller가 boot 시퀀스(Llm_metric_bridge/Fd_accountant 등 process-global 1회 설치와 동일 지점). 서버당 2-필드 JSON 1줄.
- **telemetry-as-fix 아님**: counter가 아니고 hang을 고친다 주장하지 않음. 현재 어떤 표면도 노출하지 않는 config-resolution 경계값을 노출해 #25128(idle-timeout 120s 설정됐으나 미발화)의 근본원인(config-load 갭 vs enforcement vs slow-alive)을 판별 가능케 함 — "observe the boundary before fixing" 규율 정합.
- **관례 준수**: `Log.Runtime.info ~category:Log.Boundary`는 기존 패턴(oas_diag_sink.ml:45), init() 후 배치로 frozen snapshot 캡처(올바른 순서). 워크어라운드 시그니처 스캔 clean, 테스트 2종 wired.

[근거] head SHA 고정 4축 검증; 2026-07-19 KST; confidence High.

PR #25339, opened 2026-07-19
feat(keeper): commit exact pending paused-work cancellation

jeong-sik:
## Summary

- add a typed pending `Cancel_accepted` path under the existing paused-work lifecycle reservation
- bind every cancellation receipt to the exact source stimulus, source revision, owner generation, and stable operator operation ID
- write a source-bearing settlement WAL v2 entry before checkpointing pending removal, then reconstruct the exact synthetic lease during crash replay
- retain the legacy v1 receipt reader for previously supported WAL rows
- publish only the post-commit pending projection and replay committed cancellation before current owner fences
- cover exact target removal, peer FIFO preservation, source snapshot conflicts, checkpoint-failure recovery, legacy v1 replay, and replay after owner transition

## Why

The prior cancellation transaction could settle an already leased accepted event, but a paused lane's work could still be pending. Removing that event before a durable source-bearing receipt would create a crash window where operator intent and queue truth diverged.

This change makes the receipt the durable commit point and keeps pending removal replayable from the pre-removal checkpoint.

## Stack and scope

Base: `codex/paused-work-disposition-transaction-20260719` (draft PR #25327).

This PR handles only exact pending accepted-work cancellation. It does not yet implement `Resume_owner`, `Transfer_owner`, `Settle_from_source_terminal`, paused-retained fleet count/oldest-age projection, or the 10-Keeper 8-hour acceptance run.

The source field is a hard cut for the unshipped stacked cancellation contract; legacy v1 rows using the previously supported settlement shapes remain readable.

## Validation

- `ocamlc -stop-after parsing` for every changed implementation and interface
- `git diff --check`
- static scan: no new `Obj.magic`, `failwith`, `assert false`, TODO/FIXME, Lazy, or Mutex usage

Local Dune build/tests were intentionally not run; the operator is running builds separately and PR CI is the build authority.


jeong-sik:
확인 커밋: `d8a7873e81`. typed pending `Cancel_accepted` path under paused-work reservation. cancellation receipt를 exact source stimulus·source revision·owner generation·stable operator operation ID에 bind.

**방향 건강 (exact typed bind)**: #25327 연장. **approve 성향**.

PR #25345, opened 2026-07-19
fix(keeper): bind checkpoint generation at OAS input boundary

jeong-sik:
## Summary

- bind the current keeper generation to the caller-owned `Agent_sdk.Context.t` before every OAS dispatch
- let OAS preserve that context naturally in checkpoints for both new and resumed agents
- remove the optional load/CAS generation fallback, its recovery metric, and its legacy-repair tests from #25275
- keep generation-less or malformed checkpoint identity fail-closed

## Root cause

MASC passed a `shared_context` to OAS without the MASC-owned `keeper_generation` key. OAS correctly treats a caller-supplied context as authoritative for both `Agent.create` and `Agent.resume`, then copies that context into every checkpoint. The per-turn writer therefore persisted generation-less checkpoints.

#25275 repaired the consumer by threading an optional generation through generic checkpoint load and CAS APIs. Because the active writer still omitted the field, that was a permanent second authority rather than a bounded migration.

This PR fixes the producer boundary instead: MASC writes its product identity into the context it already owns before invoking OAS. No OAS type or behavior becomes MASC-aware.

## Hard-cut behavior

- no checkpoint migration
- no inferred generation
- no optional fallback parameter
- no silent repair
- a persisted checkpoint without an exact generation remains invalid for compaction identity

## Proof

The focused regression constructs the real OAS `Agent.t` from the context returned by `Keeper_run_context.prepare_run_context`, creates an OAS checkpoint, and asserts that it contains the exact current keeper generation. The existing negative test continues to prove that a generation-less checkpoint cannot acquire an exact reference.

## Validation

- `ocamlformat --check` on all six touched OCaml files
- `ocamlc -stop-after parsing -c` on all touched implementations/tests
- focused Dune build was requested but cancelled after waiting behind the shared global Dune lock for more than 25 minutes; no local type/test pass is claimed, and exact-head PR CI is the remaining authority

Replaces the fallback implementation merged by #25275.

### DIFF
--- a/lib/keeper/keeper_checkpoint_store.ml
+++ b/lib/keeper/keeper_checkpoint_store.ml
@@ -567,43 +567,12 @@ let checkpoint_generation_strict (checkpoint : Agent_sdk.Checkpoint.t) =
      | None -> Error Generation_not_integer)
   | Some _ -> Error Generation_not_integer
 
-(* [generation_fallback] closes the pre-#25046 checkpoint gap (#25217): a
-   keeper's OAS turn-persist path serializes the live context, which does not
-   carry [keeper_generation] in its Session scope, so its primary checkpoint
-   is written without the key that #25046's strict identity requires. On the
-   load side the keeper's own [meta.runtime.generation] is the authoritative
-   generation SSOT — using it when (and only when) the persisted key is
-   absent recovers identity from an equally-authoritative source rather than
-   fabricating one. Only [Generation_missing] is recovered; a present-but-
-   malformed value ([Generation_not_integer]) stays a hard error. The save
-   path never passes a fallback, so the write invariant "a freshly built
-   checkpoint must carry its own generation" is unchanged. *)
-let checkpoint_generation_with_fallback ?generation_fallback checkpoint =
-  match checkpoint_generation_strict checkpoint, generation_fallback with
-  | Ok generation, _ -> Ok generation
-  | Error Generation_missing, Some fallback ->
-    Log.Keeper.warn
-      "checkpoint generation key absent; recovering from meta SSOT \
-       generation=%d (pre-#25046 checkpoint, #25217)"
-      fallback;
-    Otel_metric_store.inc_counter
-      Keeper_metrics.(to_string OasExecutionErrors)
-      ~labels:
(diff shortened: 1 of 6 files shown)


jeong-sik:
적대적 재검토 — exact HEAD `ebc8d35d167b8de1625fe2952b41062b60b3766e`

- P0: 0 / P1: 0 / P2: 0 / P3: 1
- 생성과 재개 경로 모두 caller context가 checkpoint context를 override하도록 OAS installed source 계약과 일치합니다.
- P3: resume 경로를 직접 고정하는 회귀 테스트는 없습니다.
- 소스 판정: **PASS**
- 병합 판정: **현재 불가** — Draft이며 Build and Test/Health가 아직 완료되지 않았습니다.

따라서 source-pass + `queue:merge-after-ci`로 분류합니다.

jeong-sik:
헤드 이동 재검토 — exact HEAD `be6e062d9029cc161c1ea34dafcf28950ea17369`

- 이전 reviewed head의 3개 commit과 range-diff가 모두 `=`인 main rebase입니다.
- P0/P1/P2/P3: 0/0/0/1, source PASS 유지
- P3 resume-path direct regression test gap은 그대로입니다.
- current Build and Test 진행 중, Draft이므로 ready가 아닙니다.

현재 pass/P3/do-not-merge/merge-after-ci 라벨은 판정과 일치합니다.

jeong-sik:
최종 병합 판정 — exact HEAD `be6e062d9029cc161c1ea34dafcf28950ea17369`

- P0/P1/P2/P3: 0/0/0/1 (resume-path direct test gap만 비차단 P3)
- required `CI Gate`: PASS
- exact-head checks: 17/17 success 또는 의도된 skip, non-green 0
- review threads / CHANGES_REQUESTED: 0
- base `4a0be04391`, mergeable CLEAN

소스와 CI 조건이 모두 충족됐으므로 `do-not-merge` 계열을 해제하고 ready → squash merge로 진행합니다.

[근거] `gh pr checks 25345 --required`, exact-head check-runs, GraphQL reviewThreads; 2026-07-20 KST; 신뢰도 High.

PR #25349, opened 2026-07-19
feat(keeper): persist typed paused-owner resume intent

jeong-sik:
## Summary

- add a typed durable `Resume_owner` receipt keyed by Keeper and stable operator operation ID
- serialize all paused-work disposition intents for one Keeper with a cross-process durable lock
- validate the exact durable generation and trace identity, reject Dead tombstones, and reserve the existing `Paused_work_disposition` lifecycle boundary
- persist the receipt before clearing durable pause state
- project the exact registered lane through `Operator_resume` and wake it without touching the event queue
- return post-receipt projection failures as typed `Committed_followup_failed` outcomes so retry completes the same intent
- support durable-only resume when the live registry lane is absent; the committed outcome explicitly reports the missing live projection
- reject a second operation ID once the owner is active and reject conflicting reuse of one operation ID

## Why

`Pause` remains nonterminal and retained events stay untouched. A deliberate resume still needs a durable operator intent that survives a lost response or a crash between intent commit, metadata projection, and live registry projection.

The receipt is the commit point. Durable and live projection can be replayed from that receipt without inventing TTL, owner fallback, or event matching rules.

## Stack and scope

Base: `codex/paused-pending-cancel-receipt-20260720` (draft PR #25339).

This leaf implements the domain receipt store and exact `Resume_owner` transaction. Existing HTTP/MCP/directive resume surfaces are not routed through it yet; that wiring remains a separate follow-up. `Transfer_owner`, `Settle_from_source_terminal`, paused-retained fleet count/oldest-age projection, and the 10-Keeper 8-hour acceptance run also remain.

## Validation

- `ocamlc -stop-after parsing` for every changed implementation and interface
- `git diff --check`
- static scan: no new `Obj.magic`, `failwith`, `assert false`, TODO/FIXME, Lazy, or direct Mutex usage

Local Dune build/tests were intentionally not run; the operator is running builds separately and PR CI is the build authority.

### DIFF
--- a/lib/keeper/keeper_paused_work_disposition_receipt.ml
+++ b/lib/keeper/keeper_paused_work_disposition_receipt.ml
@@ -1,5 +1,7 @@
 type operation = Resume_owner
 
+type keeper_lock = { keeper_name : string }
+
 type t =
   { keeper_name : string
   ; expected_trace_id : Keeper_id.Trace_id.t
@@ -13,7 +15,6 @@ type save_result =
   | Created
   | Existing of t
 
-type keeper_lock = { keeper_name : string }
 
 let ( let* ) = Result.bind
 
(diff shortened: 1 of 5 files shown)


jeong-sik:
확인 커밋: `63166fc371`. typed durable `Resume_owner` receipt(Keeper + stable operator operation ID keyed). 모든 paused-work disposition intents를 cross-process durable lock으로 serialize. exact durable generation validate.

**방향 건강 (typed durable resume + cross-process lock)**: **approve 성향**.

PR #25352, opened 2026-07-19
feat(keeper): route dashboard resume through typed receipts

jeong-sik:
## Summary

- require `owner_generation` and a stable `operator_operation_id` for single dashboard resume
- require exact per-Keeper resume targets for the bulk dashboard endpoint
- route both surfaces through `Keeper_paused_work_resume_transaction`
- remove the raw `Keeper_directive.Resume` path from these dashboard handlers
- make `Wakeup` a scheduling signal only; it no longer clears an operator pause in runtime or optimistic dashboard state
- return receipt/commit/projection truth, including HTTP 202 for a committed intent whose follow-up projection remains pending
- keep one client operation ID across an ambiguous failed response and forward the observed generation from every dashboard resume control
- add OCaml request-contract/wakeup-boundary tests and dashboard API/call-site tests

## Stack and scope

Base: `codex/paused-work-resume-transaction-20260720` (draft PR #25349).

This leaf closes the direct dashboard single/bulk resume bypass and the wakeup-as-resume alias. `masc_keeper_up`, dashboard boot-resume, and legacy gRPC `Resume` decode still have separate lifecycle/dead-revival boundaries and are intentionally not changed here. `Transfer_owner`, `Settle_from_source_terminal`, fleet paused-retained metrics, and the 10-Keeper 8-hour acceptance run remain.

## Validation

- OCaml parser checks for every changed `.ml`/`.mli`
- TypeScript syntax parser for every changed `.ts` file
- `git diff --check`
- static scan confirms no raw `Keeper_directive.Resume`, `mark_resumed`, or `paused = false` path remains in the dashboard directive handler
- static scan confirms the former `wakeup_resume` runtime path is removed

The targeted ESLint command reported these files are outside the repository's curated ESLint file list, so it provided no lint coverage. Local Dune build/tests, TypeScript typecheck, and dashboard tests were intentionally not run; the operator is running builds separately and PR CI remains the build authority.

### DIFF
--- a/dashboard/src/api/keeper-bulk.test.ts
+++ b/dashboard/src/api/keeper-bulk.test.ts
@@ -35,15 +35,32 @@ describe('bulkKeeperDirective', () => {
     }
     stubFetch(expected)
 
-    const res = await bulkKeeperDirective(['rondo', 'qa-king'], 'resume')
+    const res = await bulkKeeperDirective(
+      [
+        { name: 'rondo', ownerGeneration: 3, operatorOperationId: 'resume-rondo-1' },
+        { name: 'qa-king', ownerGeneration: 5, operatorOperationId: 'resume-qa-1' },
+      ],
+      'resume',
+    )
 
     const call = mockFetch.mock.calls[0]!
     const init = call[1] as RequestInit
     expect(call[0]).toBe('/api/v1/keepers_bulk/directive')
     expect(init.method).toBe('POST')
     expect(JSON.parse(init.body as string)).toEqual({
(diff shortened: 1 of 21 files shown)


jeong-sik:
확인 커밋: `a2564e0707`. single dashboard resume에 `owner_generation` + stable `operator_operation_id` 요구. bulk endpoint에 exact per-Keeper resume targets. both surfaces를 `Keeper_paused_work_resume_transaction`으로.

**방향 건강 (typed receipts + exact targets)**: 노 untyped. **approve 성향**.

PR #25363, opened 2026-07-19
fix(keeper): keep Board attention pending until judgment

jeong-sik:
## Problem

Board attention candidates could transition from Pending to Expired solely from a 24-hour wall-clock TTL. That silently terminalized accepted durable work without an LLM judgment or durable delivery, violating the mixed-workload no-loss contract.

## Change

- delete the Expired state and its runtime codec
- delete candidate_ttl_sec and all wall-clock expiry transitions
- remove the unused now authority from the internal drain test surface
- keep Pending candidates durable until judgment and delivery succeed
- reject legacy expired wire rows explicitly as an unknown status

No runtime migration is included: the live base-path store contained no expired rows when checked, and silently converting discarded work back to Pending would invent recovery semantics.

This is the first small lifecycle leaf recut from the oversized #25092 stack. It intentionally does not retain that PRs fixed-cap governance, context batching, atomic commit, or worker-plane changes.

## Validation

- ocamlformat --check on all three touched files
- git diff --check
- scripts/dune-local.sh build test/test_keeper_board_attention_candidate.exe
- scripts/dune-local.sh exec ./test/test_keeper_board_attention_candidate.exe: 18/18 passed

The shared local switch had an unrelated agent_sdk pin drift, so the focused wrapper build/test used its documented one-shot MASC_SKIP_PIN_CHECK=1 bypass. PR CI remains the build authority.

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -40,13 +40,10 @@ type consumed_state =
   ; consumed_at : float
   }
 
-type expired_state = { expired_at : float }
-
 type status =
   | Pending of pending_state
   | Judged of judged_state
   | Consumed of consumed_state
-  | Expired of expired_state
 
 type candidate =
   { candidate_id : string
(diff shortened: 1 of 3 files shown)


jeong-sik:
확인 커밋: `c493b2552f`. Board attention candidates가 24시간 wall-clock TTL만으로 `Pending → Expired`. LLM judgment/durable delivery 없이 accepted durable work를 **silent terminalize**(mixed-workload 위반).

**방향 매우 건강 (wall-clock TTL 제거)**: #25092(`Expired` TTL 되돌림)과 같은 맥락. 노 silent work loss. **강한 approve 성향**.

PR #25365, opened 2026-07-19
fix(keeper): require exact Board attention batch responses

jeong-sik:
## Stack

Depends on #25363. Review only commit 1dd690c285.

## Problem

A successful batch response could omit requested candidate IDs. The returned subset was consumed while omitted durable work stayed Pending without typed failure evidence. Batch failure-evidence writes were also reduced to logs, and lower ledger I/O exceptions could escape a result-returning API.

## Change

- require the returned candidate-id set to equal the requested set
- treat missing or unknown identities as Response_contract_unavailable for the whole attempted batch
- persist the typed failure on every requested candidate before returning a non-consuming report
- propagate failure-evidence persistence errors instead of logging and continuing
- normalize non-cancellation ledger I/O exceptions at the Board attention persistence boundary

This leaf does not add context grouping, guessed capacity, one-call-per-admission governance, or atomic multi-row judgment commit. Those remain separate reviewable changes.

## Validation

- ocamlformat --check on touched files
- git diff --check
- focused executable build passed
- test_keeper_board_attention_candidate: 19/19 passed, including missing-ID and failure-evidence-storage cases

The shared local switch has unrelated agent_sdk pin drift, so focused build/test used the wrapper documented one-shot MASC_SKIP_PIN_CHECK=1 bypass. PR CI remains authoritative.

PR #25366, opened 2026-07-19
fix(keeper): commit Board attention judgments atomically

jeong-sik:
## Stack

Depends on #25365, which depends on #25363. Review only commit 99dd020589.

## Problem

Even after exact response-set validation, judgment application rewrote the candidate ledger once per candidate and immediately entered delivery. A mid-batch storage failure could therefore leave only a prefix durably Judged or Consumed.

## Change

- generalize the private candidate-ledger transaction to commit multiple updated rows in one durable rewrite
- validate duplicate IDs, keeper ownership, current Pending state, and exact judgment presence inside that transaction
- commit every requested Pending to Judged together or commit none
- only begin idempotent queue delivery after the atomic judgment commit returns
- keep delivery and Consumed transitions explicitly separate because the event queue and candidate ledger are different durable files

No fake cross-file transaction, test-only production hook, context grouping, capacity guess, or worker-plane change is included.

## Validation

- git diff --check
- test_keeper_board_attention_candidate: 20/20 passed
- fault injection after Provider judgment but before ledger commit verified both candidates remain Pending

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -699,7 +699,7 @@ let serialize_candidates candidates =
   String.concat "" (List.map append_row candidates)
 ;;
 
-let update_ledger ~base_path ~keeper_name decide =
+let update_ledger_many ~base_path ~keeper_name decide =
   let path = candidate_path ~base_path ~keeper_name in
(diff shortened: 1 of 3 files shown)


jeong-sik:
확인 커밋: `99dd020589` (#25365 위). Board attention judgments atomic commit.

**approve 성향**.

PR #25367, opened 2026-07-19
refactor(keeper): type Board attention context authority

jeong-sik:
## Stack

Depends on #25366 → #25365 → #25363. Review only commit 627feb2a4a.

## Problem

The durable partition worker needs an exact immutable Keeper-context identity. The previous experimental stack used an untyped canonical JSON string and accepted ambiguous duplicate object keys through first-match lookup.

## Change

- add an opaque Candidate.Context_key type
- derive it only from exactly one persisted keeper_context field
- recursively canonicalize object field order while retaining list order
- reject missing or multiple keeper_context fields
- reject duplicate nested object keys and non-finite numbers
- normalize signed zero so equality and durable canonical representation agree

No token estimate, capacity guess, context hash collision assumption, worker lifecycle, or owner-lane behavior change is included. This is the typed authority required by the next durable partition leaf.

## Validation

- git diff --check
- test_keeper_board_attention_candidate: 22/22 passed

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -314,6 +314,78 @@ let candidate_to_json candidate =
 
 let ( let* ) = Result.bind
 
+module Context_key = struct
+  type t = Yojson.Safe.t
+
+  let rec canonicalize ~context = function
+    | `Assoc fields ->
+      let* canonical_fields =
+        List.fold_left
+          (fun result (key, value) ->
+             let* fields = result in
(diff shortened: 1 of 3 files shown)


jeong-sik:
확인 커밋: `627feb2a4a` (#25366 위). Board attention context authority typed.

**approve 성향**.

jeong-sik:
델타 확인: 현재 head의 own 커밋이 이전 검토 커밋과 patch-identical(rebase로 sha만 변경, 내용 동일). 리뷰 스킵, 이전 판정 유지.

PR #25370, opened 2026-07-19
fix(keeper): commit Board attention failure evidence atomically

jeong-sik:
## Stack

Depends on #25367 → #25366 → #25365 → #25363. Review only commit 377f79c0f9.

## Problem

Batch failure evidence was explicit but persisted one candidate at a time. A storage failure between rewrites could leave only a prefix carrying the typed provider or response-contract failure.

## Change

- extract the pure immutable candidate failure transition shared by single and batch paths
- validate one Keeper identity, unique candidate IDs, and complete current-ledger membership before mutation
- reject terminal Consumed rows as impossible attempted-batch members
- commit all changed failure-evidence rows in one candidate-ledger rewrite or commit none
- avoid a rewrite when every row already contains the same failure kind and detail

## Validation

- git diff --check
- test_keeper_board_attention_candidate: 22/22 passed
- two-candidate fault injection after Provider failure verified the original all-Pending ledger survives an unavailable commit destination

### DIFF
--- a/lib/keeper/keeper_board_attention_candidate.ml
+++ b/lib/keeper/keeper_board_attention_candidate.ml
@@ -852,27 +852,31 @@ let same_failure left right =
   left.kind = right.kind && String.equal left.detail right.detail
 ;;
 
+let candidate_with_retryable_failure current failure =
+  match current.status with
+  | Pending pending ->
+    (match pending.last_failure with
+     | Some existing when same_failure existing failure -> current
+     | Some _ | None ->
+       { current with status = Pending { last_failure = Some failure } })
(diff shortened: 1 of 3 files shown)


jeong-sik:
확인 커밋: `377f79c0f9` (stack). Board attention failure evidence atomic commit.

**방향 건강 (atomic + 노 silent failure)**: **approve 성향**.

jeong-sik:
델타 확인: 현재 head의 own 커밋이 이전 검토 커밋과 patch-identical(rebase로 sha만 변경, 내용 동일). 리뷰 스킵, 이전 판정 유지.

PR #25374, opened 2026-07-19
test(keeper): add exact paused-work 8h soak gate

jeong-sik:
## Summary\n\n- add an exact 10-Keeper paused-work disposition soak gate for #25191\n- rotate deterministic Resume_owner, Transfer_owner, and Cancel_accepted cases across every Keeper\n- replay the same exact operator request before and after terminal observation\n- verify one durable disposition receipt, exact lease-sequence advancement, zero surviving source, and one target transfer projection ledger row\n- support synchronous executable-only restart injection and distinguish short smoke runs from acceptance-eligible 8h evidence\n- document the isolated-fixture contract and artifacts\n\n## Stack\n\n- base: #25372\n- requires the v4 accepted transfer projection ledger so replay after target ACK cannot re-enqueue the source\n\n## Validation\n\n- \n- \n- \n- static scans for destructive/eval/home-anchored behavior\n- local build: not run (operator will build separately)\n- local tests/typecheck: not run per operator instruction\n- 8h soak: not run; this PR provides the executable acceptance gate and does not claim runtime evidence\n

### DIFF
--- a/scripts/harness/perf/README.md
+++ b/scripts/harness/perf/README.md
@@ -1,4 +1,56 @@
-# Perf harness: main-domain scheduler starvation
+# Performance and soak harnesses
+
+## Paused-work disposition: exact 10-Keeper 8h soak
+
+`paused_work_disposition_soak.sh` is the release-evidence gate for #25191. It
+attaches to an isolated, already-running server and never builds or boots MASC.
+For exactly ten clean Keeper lanes it deterministically rotates through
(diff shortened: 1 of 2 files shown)


jeong-sik:
확인 커밋: `ad26891592` (#25191 test). 10-Keeper paused-work disposition soak gate. `Resume_owner`/`Transfer_owner`/`Cancel_accepted` rotate. 동일 exact operator request를 before/after replay.

**approve 성향**(test soak).

Issue #25731, opened 2026-07-26
P0: hard-cut Memory OS to one immutable HEAD authority before exact-flow settlement


jeong-sik:
## Problem

The current Keeper Memory OS has multiple production authorities:

- facts JSONL
- events JSONL
- per-episode files
- generation files
- Librarian exact journal

Librarian can persist `domain_valid` before the Memory OS write sequence completes, while readers, GC, recall, and dashboard select different physical authorities. A durable `Domain_committed` receipt therefore cannot be proven.

This blocks the OAS v0.223.2 exact Input/Output consumer hard cut and its measurement/generation durability fences.

## Decision

Fresh-state hard cut only. Do not import, reconcile, migrate, or fall back to historical runtime JSON.

Use one production authority:

```text
HEAD CAS -> immutable commit record -> immutable manifest -> immutable typed fact/episode objects
```

- Fresh opaque physical leaves; refs contain `{ leaf, sha256, byte_count }`.
- Object existence and directory scans are never authority.
- `Domain_committed` requires a durable HEAD receipt whose reachable manifest contains the operation.
- Publication-indeterminate outcomes fence execution and are never blindly retried.
- Provider/model/catalog/credential/pricing metadata is forbidden from the store and Librarian journal.

## Atomic delivery plan

1. Add a production-unused canonical store engine and focused proofs.
2. In one atomic production PR, migrate every reader/writer and delete `Keeper_memory_os_io`.
3. Separate lifecycle nonce from Memory OS episode generation.
4. Delete the writer-less `_shared` phantom store from backend/API/UI/tests.
5. Replace Librarian journal with current-only `domain_uncommitted -> domain_committed | domain_commit_uncertain` settlement.
6. Connect HITL, Board, Compaction, and Librarian OAS measurement intent/terminal durability fences.

No intermediate production dual authority may merge.

## Acceptance

- Concurrent writers: one strict HEAD CAS winner; loser cannot overwrite or auto-retry stale state.
- Same operation and payload is idempotent; same operation with a different payload is a hard conflict.
- Readers observe one complete old or new manifest, never mixed state.
- Missing/corrupt objects fail closed with no directory-scan or legacy fallback.
- Recall and dashboard project the same immutable snapshot.
- Stale consolidation/GC is rejected by root mismatch.
- Cancellation does not turn a committed publication into retryable failure.
- Fresh state contains no provider/model/catalog/credential/pricing store metadata.
[...]

PR #25733, opened 2026-07-26
feat(fs): add capability-relative strict HEAD CAS

jeong-sik:
## Summary

- add one capability-relative, bounded one-row HEAD authority under `Fs_compat.Capability_head`
- bind cursors to the pinned parent directory inode, immutable stable-lock epoch, leaf, and target inode/length/SHA-256
- serialize same-process writers with a nonblocking `(parent.dev,parent.ino,leaf)` lease, followed by a nonblocking cross-process stable-file lock
- publish through a fresh secure-random opaque stage, exclusive write, file sync, stage close, capability-relative rename, parent sync, and exact target verification
- return typed `Unchanged`, `Publication_indeterminate`, or `Published` effects plus typed cleanup/resource-settlement evidence
- bind every success receipt to expected cursor, exact LF-framed wire SHA-256/byte length, and published cursor

## Authority boundary

HEAD is the only content authority. The stable lock contains only an immutable 256-bit random fencing epoch supplied through the caller's Eio secure-random capability. It is never rewritten, renamed, deleted, or used to decide domain content or freshness.

The API exposes no append, scan, fallback, recovery, repair, delete, or migration surface. Public leaf names cannot use the internal `.masc-capability-head-` namespace in any ASCII case variant, keeping targets disjoint from stable locks and publication stages on case-sensitive and case-insensitive filesystems.

Direct external namespace mutation is outside the cooperating private-directory contract and is not silently reconciled. This is the prerequisite for #25731. It contains no Memory OS codec, provider/model/catalog/credential/pricing knowledge, legacy import, or automatic retry.

## Concurrency and cancellation

- the exact parent directory is pinned as an Eio capability before any authority operation
- process-local lease is acquired before opening the stable lock FD
- `F_TLOCK` begins at offset zero and is nonblocking
- pending caller cancellation is checked before dispatch
- once dispatched, parent cancellation is deferred until one typed publication outcome and receipt settle
- no cancellation check occurs after publication
- callback result and switch settlement remain separate
- post-rename failures carry intended/observed evidence; post-verification failures carry `Published`

## Focused proof surface

- absent publish and close/reopen
- stale and cross-root cursors
- strict empty/CR/LF/oversized row rejection
- reserved internal namespace rejection before entropy consumption or filesystem mutation, including ASCII case variants
- real two-fiber same-root `Busy` and different-root non-sharing
- real separate-process `F_TLOCK` contention with bounded deadline, followed by stale-cursor conflict
- pending cancellation, injected pre-rename cancellation, and real post-dispatch parent cancellation
- pre-rename, post-rename, post-parent-sync, and post-verification effect boundaries
- secure-random EOF at lock initialization and stage creation returns typed `Unchanged`; empty-lock bootstrap remains recoverable
- resource-settlement failure preserves success with the exact typed warning

## Validation

Local build/test/formatter intentionally not run per current workflow direction. Draft CI is the compile and conformance authority.

[근거] OCaml 5.5 manual, installed Eio v1.3 interfaces, Eio v1.4 release notes dated 2026-07-23, and POSIX.1-2024 lock/rename contracts checked 2026-07-26 KST. Confidence High.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -86,6 +86,7 @@
   (eio_posix (>= 0.16))
   (cstruct (>= 6.0))
   (digestif (>= 1.1))
+  (optint)
   (ocaml-protoc-plugin (>= 4.0))
   (pbrt (and (>= 3.0) (< 4.0)))
   ; [ctypes] + [ctypes-foreign] removed: zero callers in lib/ bin/ test/
--- a/dune-project
+++ b/dune-project
@@ -86,7 +86,7 @@
   (eio_posix (>= 0.16))
   (cstruct (>= 6.0))
   (digestif (>= 1.1))
-  (optint)
+  optint
   (ocaml-protoc-plugin (>= 4.0))
   (pbrt (and (>= 3.0) (< 4.0)))
   ; [ctypes] + [ctypes-foreign] removed: zero callers in lib/ bin/ test/
--- a/lib/fs_compat/capability_head.ml
+++ b/lib/fs_compat/capability_head.ml
@@ -331,7 +331,7 @@ let try_lock file =
          Eio_unix.run_in_systhread
            ~label:"capability_head.try_lock"
            (fun () ->
-             ignore (Unix.lseek raw_fd 0 Unix.SEEK_SET);
+             let _position = Unix.lseek raw_fd 0 Unix.SEEK_SET in
              Unix.lockf raw_fd Unix.F_TLOCK 0));
        Ok ()
      with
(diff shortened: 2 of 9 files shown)


jeong-sik:
Exact-head merge audit: `d936df820f832bdff3b4171a3081934017280dc5`

- P0: 0
- P1: 0
- independent architecture, test/concurrency, and next-consumer API reviews: approve
- Health, Lint, Fundamental Lint Suite, Build and Test, CI Gate: green
- residual scope: whole-runtime stress remains Phase 6; no blocker for this primitive

Mergeability: approved.

PR #25737, opened 2026-07-26
fix(fs): preserve fatal write exceptions

jeong-sik:
## Summary

- keep `Out_of_memory`, `Stack_overflow`, `Sys.Break`, and `Eio.Cancel.Cancelled` out of capability write typed errors
- preserve the original raw backtrace through stage dispatch, payload propagation, cleanup, parent sync, publication, and Eio switch settlement
- retain the existing typed result contract for ordinary filesystem/write failures
- prove replace, exclusive-create, and cleanup propagation through the existing test-only `before_stage` seam without adding a production fault API

## Boundary

This changes only the shared filesystem atomic/exclusive-create primitive. It adds no Keeper, provider, model, catalog, migration, scan, or compatibility behavior.

## Validation

Per the current workflow direction, no local build, test, or formatter was run. GitHub CI is the validation authority for this draft PR.

### DIFF
--- a/lib/fs_compat/atomic_write.ml
+++ b/lib/fs_compat/atomic_write.ml
@@ -1115,6 +1115,36 @@ let write_cleanup_failures failures =
   List.map (fun failure -> Write_cleanup_failure failure) failures
 ;;
 
+let reraise_capability_write_cancellation
+      ~backtrace
+      ~operation
+      ~target_effect
+      ~interrupted_primary_failure
+      ~cleanup_failures
+      reason
+  =
+  let reason, interrupted_recovery, carried_cleanup_failures =
+    match reason with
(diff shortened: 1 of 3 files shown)


jeong-sik:
독립 적대적 재리뷰 (`a544ecfdd2d4f0c62fb16ee52584d91af78a74c0`)

- P0: 0
- P1: 0
- 이전 cleanup-cancellation authority-loss P0 해소 확인
- primary failure, target effect, cleanup facts, recovery interruption, original raw backtrace 보존 확인
- public API/ABI 및 production fault surface 불변
- 잔여 P2는 조합 테스트 강도이며 구현상 보존 확인, 비차단

필수 Health/Lint/Build/Fundamental green 후 merge 승인.

PR #25753, opened 2026-07-26
feat(memory-os): expose publication obligation identity

jeong-sik:
## Summary
- expose provider-neutral typed projectors for publication obligation operation ID
- expose the checksum-bound desired receipt ID, state SHA-256, and generation
- prove prepared and decoded obligations bind the same identity as the committed receipt

## Boundary
- no owner, store, expected authority, payload digest/length, or immutable commit reference is exposed
- no obligation decoding or private-byte duplication is added to callers
- no OAS, provider, model, pricing, migration, alias, publication, or recovery behavior changes

## Validation
- Not run locally by explicit instruction; PR CI is the build, formatting, and test authority.

### DIFF
--- a/lib/keeper/keeper_memory_os_store.ml
+++ b/lib/keeper/keeper_memory_os_store.ml
@@ -3061,6 +3061,38 @@ let publication_obligation_of_bytes raw =
     Error (make_error (Invalid_publication_obligation detail))
 ;;
 
+let publication_obligation_operation_id
+      (value : publication_obligation)
+  =
+  value.operation_id
+;;
+
+let publication_obligation_desired_receipt_id
+      (value : publication_obligation)
+  =
+  match value.desired with
+  | Head_authority desired -> desired.receipt_id
+  | Empty_authority ->
+    (* Construction and decoding both reject an empty desired authority. *)
+    assert false
+;;
+
+let publication_obligation_desired_state_sha256
+      (value : publication_obligation)
+  =
+  value.state_sha256
+;;
+
+let publication_obligation_desired_generation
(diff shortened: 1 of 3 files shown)


jeong-sik:
Exact-head adversarial review: `2f1ed96bf2b76409514483be055f7314ed3236ec`

- P0: 0
- P1: 1
- `publication_obligation_desired_receipt_id` and `_desired_generation` contain `Empty_authority -> assert false` while the internal `desired` field still has the broader `authority_token` type.
- Current constructors reject empty desired authority, but the public projectors remain partial because the impossible state is not type-eliminated.

Required fix: make `publication_obligation.desired` a private non-empty `head_authority`; narrow the decoded token through a typed error before construction; make all projectors direct total field access. No assertion, exception, option fallback, or compatibility alias.

The remaining projector binding, codec-stability test, privacy boundary, and provider/model/pricing neutrality are sound. PR remains Draft pending this fix and CI.

jeong-sik:
Final exact-head adversarial review: `dff167375b5a9498b5d4abbc07fd198c385d2321`

- P0: 0
- P1: 0
- `publication_obligation.desired` is structurally non-empty.
- All four projectors are total direct accessors.
- Serializer field order, domain-separated checksum input, checksum-last canonical bytes, and public schema are unchanged.
- Exported obligation encode/decode/projector paths contain no assertion, exception helper, option fallback, or incomplete match.
- Focused test binds prepared/decoded obligation identity to the actual committed receipt.
- No OAS/provider/model/pricing/migration/alias surface was added.

Source is merge-ready pending required CI.

PR #25775, opened 2026-07-26
fix(keeper): reject untyped runtime blocker fallbacks

jeong-sik:
## Summary
- remove provider-named legacy runtime blocker string recovery
- accept only the current typed runtime blocker decoder
- fail closed as unknown for unrecognized blocker wire values

## Boundary
MASC does not infer provider/model semantics from persisted strings. This is a current-schema hard cut; no migration or compatibility decoder is added.

## Validation
- local build/tests/formatter intentionally not run; PR CI is authoritative


jeong-sik:
@codex review

Current-head review requested before merge. Please focus on whether removal of the provider-named fallback leaves only typed current-schema authority and introduces no raw-text replacement or silent-empty path.

Issue #25808, opened 2026-07-27
keeper: complete OAS v0.226 durable exact-flow settlement migration


jeong-sik:
## Symptom

MASC `main` pins OAS/agent_sdk v0.226.0, but current production consumers still call the removed process-local exact-flow APIs:

- `Exact_output.create_flow_preference_store`
- `Exact_output.remove_flow_preference_scope`
- `Exact_output.settle_flow_domain`
- `Exact_output.Domain_already_settled`
- `Exact_output.Domain_preference_scope_released`

A current-pin build therefore cannot compile. This is not a rename-only migration.

## Reproduction

1. Checkout MASC main at or after `7cce70b99017bb6417cd7659ba63484ae96b0ee7`.
2. Install the repo SSOT agent_sdk v0.226.0 pin.
3. Run a focused build that reaches the keeper library.

The first failures are in:

- `lib/keeper/keeper_exact_flow_scope.ml`
- `lib/keeper/hitl_summary_worker.ml`
- `lib/keeper/keeper_board_attention_exact_flow.ml`
- `lib/keeper/keeper_compaction_llm_summarizer.ml`
- `lib/keeper/keeper_librarian_runtime.ml`

## Cause

OAS #2813/#2815 deliberately replaced raw process-local creation, settlement, and removal with:

- `recover_flow_preferences ~evidence`
- `commit_and_settle_flow_domain ~commit`
- `commit_and_retire_flow_preference_scope ~commit`

The callback is the durable content-commit fence. A compile-only `~commit:(fun _ -> Ok ())` would acknowledge evidence that MASC did not persist and is forbidden.

### Shared scope owner

`keeper_exact_flow_scope.ml:74,121` must own one current-schema-only preference evidence journal per registered Keeper generation/surface. Startup must decode the complete authenticated evidence set before `recover_flow_preferences`. Retirement must durably commit the retirement intent before releasing the reservation.

No legacy decoder, migration, default fill, or compatibility alias. Missing old pre-1.0 state is operator reset/fresh start. Invalid current state must fail closed with a typed, operator-visible reason while Keeper lifecycle remains alive.

The HITL topology readiness probe at `hitl_summary_worker.ml:213,223` is ephemeral and must not fabricate a durable callback. It may recover an empty independent store and discard that whole probe store.

### Board attention

[...]

PR #25809, opened 2026-07-27
fix(keeper): consume current OAS receipt facts

jeong-sik:
## Summary

- read OAS v0.226 immutable `flow_attempt_snapshot` values through snapshot-specific accessors
- consume the renamed closed `flow_execution_error_generation_dispatch` fact
- preserve `Request_body_refused_by_provider { status }` as a distinct typed capacity refusal and event payload without fabricating byte or token limits

## Boundary

This PR intentionally does not fake the new durable exact-flow contract with `~commit:(fun _ -> Ok ())`. Current MASC still needs domain-specific durable intent/content ordering before removed settlement/preference APIs can be migrated. That blocker is tracked in #25808.

No compatibility shim, legacy branch, provider/model logic, Pricing policy, estimated limit, or migration path is added.

## Validation

Passed:

- `git diff --check`
- `ocamlformat --check` on all seven touched files

Focused build attempted:

- `scripts/dune-local.sh build test/test_keeper_unified_context_overflow.exe`
- stopped before Dune because the shared operator switch has agent_sdk 0.223.1 while the repo SSOT requires >= 0.226.0
- the shared opam switch was intentionally not mutated

This draft is not merge-ready while #25808 leaves current-pin settlement symbols uncompilable. CI is expected to expose the remaining exact blockers and must not be bypassed.

### DIFF
--- a/lib/keeper/hitl_summary_worker.ml
+++ b/lib/keeper/hitl_summary_worker.ml
@@ -300,6 +300,22 @@ let exact_identity_of_candidate
   }
 ;;
 
+let exact_identity_of_snapshot
+      (candidate : Exact_output.flow_attempt_snapshot)
+  =
+  let receipt = candidate.receipt in
+  { slot_id = candidate.visit.identity.candidate_id
+  ; call_id =
+      receipt
+      |> Exact_output.generation_receipt_snapshot_call_id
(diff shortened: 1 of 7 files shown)


jeong-sik:

PR #25812, opened 2026-07-27
ci: make Build the sole OCaml compile authority

jeong-sik:
## Problem

`CI` compiled the same OCaml graph independently on three isolated runners:

- `Health` called `health_snapshot.sh`, whose default is `dune build @check`.
- `Build and Test` compiled `@default @check`.
- `Lint` forced a separate `@install` build with `OCAMLPARAM`.

The jobs restore historical Dune cache entries, but they do not share artifacts produced by the current workflow run. A single source error therefore appeared independently in Health, Build, and Lint while paying for three OCaml toolchain/dependency bootstraps.

## Change

- Make `Build and Test` the sole full OCaml compile authority with one cacheable `@default @check @install` graph.
- Move warning-as-error policy into the root Dune `env` stanza (`+a`) so it participates in the generated rule graph instead of relying on a job-local environment plus `--force`.
- Run Health with `health_snapshot.sh --skip-build`; retain baseline fetching, source ratchets, JSON publication, path scope, and CI Gate visibility.
- Keep Lint as shell/Python/SSOT/Eio/proto-drift checks with explicit `ripgrep` and `protobuf-compiler` dependencies; remove its opam/Dune bootstrap.
- Move the compiler-libs projection checker from Lint to Build, where the sole OCaml toolchain already exists.
- Add a Meta Guard ratchet that rejects reintroduction of Health/Lint compilation or a second full compile command.

No `_build` or cross-job artifact upload is introduced. Main-push safety-net, path classifiers, proto drift, and CI Gate aggregation are unchanged.

## Expected savings topology

- Full product graphs per scoped CI run: 3 -> 1.
- OCaml toolchain and opam dependency bootstraps: 3 jobs -> 1 job.
- Forced `@install` rebuild: removed; `@install` joins the authoritative cached graph.
- Health/Lint remain parallel, independent failure surfaces rather than waiting on Build.

## Focused validation

- `bash -n scripts/ci/check-ocaml-compile-authority.sh scripts/health_snapshot.sh`
- `bash scripts/ci/check-ocaml-compile-authority.sh`
- `python3 scripts/ci/check-yaml-syntax.py` (38 YAML files)
- `bash scripts/health_snapshot.sh --skip-build ...` plus JSON assertion for `build.skipped=true`
- `bash scripts/check-keeper-event-queue-projection-boundary.sh`
- `git diff --check`

No local full Dune build was run. This PR deliberately uses the single GitHub Build job as the broad compile authority.

## Risk

The root warning policy now explicitly rejects every OCaml warning in dev and release profiles. This expands the former explicit `+8` root setting to the `+a` policy Lint already enforced on `@install`; CI may expose warnings in test-only targets that the old Lint graph did not cover. Such failures are intended compile-authority findings, not downgraded Lint output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,6 +384,9 @@ jobs:
           bash scripts/ci/apt-refresh.sh
           sudo apt-get install -y -qq ripgrep
 
+      - name: Check single OCaml compile authority
+        run: bash scripts/ci/check-ocaml-compile-authority.sh
+
       - name: Guard pending release tag on main PRs
         if: github.event_name == 'pull_request' && github.base_ref == 'main'
         env:
@@ -520,22 +523,6 @@ jobs:
           bash scripts/ci/apt-refresh.sh
           sudo apt-get install -y -qq ripgrep bc
 
-      - name: Setup OCaml toolchain
-        if: needs.changes.outputs.health_snapshot == 'true'
-        uses: ./.github/actions/setup-ocaml-toolchain
-
-      - name: Pin external OCaml dependencies
-        if: needs.changes.outputs.health_snapshot == 'true'
-        uses: ./.github/actions/pin-ocaml-deps
-        with:
-          pin-flags: --with-bisect
-
-      - name: Install system and OCaml dependencies
-        if: needs.changes.outputs.health_snapshot == 'true'
-        uses: ./.github/actions/install-ocaml-deps
-        with:
-          install-flags: --with-test
-
       - name: Fetch baseline for health ratchet
         if: needs.changes.outputs.health_snapshot == 'true'
         env:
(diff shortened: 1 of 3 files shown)


jeong-sik:
Remote topology evidence from CI run [30230763232](https://github.com/jeong-sik/masc/actions/runs/30230763232):

- Meta Guards passed the new single-authority ratchet.
- Health passed in 49s with checkout, shell deps, baseline fetch, snapshot, and upload only. No OCaml/opam step exists.
- Lint passed in 58s with shell/Python/proto checks only. No OCaml/opam step exists.
- Build and Test is the only running job with `Setup OCaml toolchain`; its one graph owns `@default @check @install`.

For scale, the contemporaneous #25752 run [30230437855](https://github.com/jeong-sik/masc/actions/runs/30230437855) spent 5m05 in Health and 7m36 in Lint before both rediscovered compile failures. The new run removes 10m54 of those two runner jobs' observed wall time. This is topology evidence, not a controlled performance benchmark because the PR source graphs differ.

Current Dune documentation supports the cache-safe design: OCaml flags belong in `env`/stanza configuration, and the build cache keys entries by the generated rule hash. That is why this PR puts `-warn-error +a` in the root Dune SSOT instead of carrying Lint's environment-only `--force` rebuild into Build.

- https://dune.readthedocs.io/en/stable/concepts/ocaml-flags.html
- https://dune.readthedocs.io/en/latest/reference/caches.html#how-caching-works

PR #25813, opened 2026-07-27
test(ci): harden the OCaml compile authority ratchet

jeong-sik:
## Follow-up to #25812

#25812 established Build and Test as the sole full OCaml compile authority. This narrow follow-up closes two adversarial bypasses in its structural Meta Guard without changing the merged workflow topology.

## Hardening

- Tokenize each non-comment workflow line with Python `shlex` instead of anchoring grep at the beginning of the command.
- Detect real `dune build` and `scripts/dune-local.sh build` invocations when they appear after YAML `run:`, environment assignments, `opam exec --`, or `./`.
- Do not count commented commands or quoted diagnostic text such as `echo "dune build @check"`.
- Keep the authority count scoped to `.github/workflows/ci.yml` only.
- Add embedded detector probes for inline `run:`, env-prefix, wrapper path, comment, and quoted-text cases.

## Path-scope proof

No classifier change is needed. The existing `changes.build` classifier includes every `*.sh`, `*.ml`, and `*.mli`, plus `ci.yml`. Therefore changes to the boundary checker shell, its compiler-libs source, or the workflow itself all continue to force Build and Test.

## Focused validation

- `bash -n scripts/ci/check-ocaml-compile-authority.sh`
- `bash scripts/ci/check-ocaml-compile-authority.sh`
- `git diff --check`

## Merge gate

Draft only. Do not merge until required CI is green and an independent current-head review finds no P0/P1 issue.

### DIFF
--- a/scripts/ci/check-ocaml-compile-authority.sh
+++ b/scripts/ci/check-ocaml-compile-authority.sh
@@ -167,6 +167,22 @@ def shell_tokens(line: str) -> list[str]:
     return list(lexer)
 
 
+def command_segments(tokens: list[str]) -> list[list[str]]:
+    separators = {";", "&&", "||", "|", "&", "(", ")"}
+    segments: list[list[str]] = []
+    current: list[str] = []
+    for token in tokens:
+        if token in separators:
+            if current:
+                segments.append(current)
+                current = []
+            continue
+        current.append(token)
+    if current:
+        segments.append(current)
+    return segments
+
+
 def count_full_compiles(lines: list[str]) -> int:
     count = 0
     for script in run_scripts(lines):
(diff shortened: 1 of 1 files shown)


jeong-sik:
Current-head adversarial follow-up (`846b3f6cb8`):

- **Path scope preserved:** `origin/main:.github/workflows/ci.yml:272` sets `build=true` for every `*.sh`, `*.ml`, and `*.mli`. Applying that exact regex returns `BUILD_FILTER_MATCH=true` for both `scripts/check-keeper-event-queue-projection-boundary.sh` and `scripts/keeper_event_queue_projection_boundary_check.ml`. A checker-only change therefore schedules Build after the step moved out of Lint. The current PR itself is a script-only change and its prior run did schedule Build, providing workflow-level corroboration.
- **Quoted YAML scalar closed:** the detector decodes `run: "dune build @check"` as a YAML scalar before shell tokenization; an embedded probe requires count=1.
- **Line continuation closed:** literal block `dune build \\` followed by `@check` is joined before tokenization; an embedded probe requires count=1.
- **False positives pinned:** YAML comments and `run: 'echo "dune build @check"'` both require count=0.
- **Focused current-head proof:** `bash -n`, the authority checker, all embedded probes, and `git diff --check` pass.

No general shell parser or workflow-scope broadening was introduced.

jeong-sik:
Current-head review fix (`639cd6dddd`): full-build attribution is now per shell command segment, not per logical line.

- Separators `;`, `&&`, `||`, `|`, `&`, `(`, and `)` terminate the current target set.
- `dune build @foo; dune build @check` counts exactly 1 full compile and cannot lend `@check` backward to the first invocation.
- `dune build @default; dune build @check` counts exactly 2 and trips the exactly-one authority gate.
- Both cases are embedded probes; `bash -n`, the complete authority checker, and `git diff --check` pass.

PR #25814, opened 2026-07-27
fix(keeper): reconcile lifecycle nonce stack with current main

jeong-sik:
## What

- merge the current main OAS pin rollback into the #25752 stack
- repair the current lifecycle journal/transaction compile surface
- expose the shared lifecycle blocker detail wire through the public module
- qualify the wrapped module in the operator test

## Invariants

- forward-only current schema; no migration or legacy branch
- no provider/model policy in MASC
- no fabricated OAS compatibility callback

## Validation

- Local full build intentionally not run; PR CI is the broad authority.


jeong-sik:
Full-build correction for exact head 8faac12879d8a90b59f7dccc1c6beabc76464132: this PR is not merge-ready yet.

The four reported checks are lightweight and did not compile the full OCaml graph. The downstream full Build and Test on MASC #25781 (run 30231981829, job 89872569126) exposes two current-head compile failures that are still present here:

1. lib/keeper/keeper_runtime_meta_transaction.ml:236 passes ~base_path to Keeper_lifecycle_nonce.recover_exact, but the permit-owned current signature no longer accepts that label.
2. test/test_operator_control_snapshot.ml:3386 calls Durable_admission.with_recovery_lifecycle_admission through the public Masc.Keeper_lifecycle_admission.Durable_transaction alias, but keeper_lifecycle_admission.mli does not expose that value.

Required before merge: repair both current public signatures/call sites and obtain full OCaml build evidence. My earlier merge-ready assessment based on the available checks is retracted.

PR #25816, opened 2026-07-27
fix(keeper): hold lifecycle admission through shutdown cleanup

jeong-sik:
## What

- acquire durable lifecycle admission before pending-confirm retirement
- retain the same permit through Cleanup_ready and finalization
- make Cleanup_ready recovery reacquire the same authority boundary
- prove a non-inheriting competing fiber cannot enter during pending-confirm cleanup

## Invariants

- no irreversible pending-confirm deletion outside lifecycle authority
- no legacy state handling or migration
- Keeper remains retryable when admission is blocked

## Validation

- Local full build intentionally not run; stacked PR checks and parent #25752 CI are the authority.


jeong-sik:

PR #25799, opened 2026-07-27
build(deps): pin OAS v0.226.0

jeong-sik:
## Summary

- pin `agent_sdk` to the OAS `v0.226.0` release commit `2186dfa5fca2360e4e99fa5e47052ba4c0cc687f`
- consume evidence-owned exact preference recovery from OAS #2815 through the provider-neutral single surface
- regenerate dependency manifests, generated documentation, and the reviewed API surface fingerprint from the pin SSOT

## Boundary

MASC consumes opaque flow, receipt, provenance, and input-capacity facts only. Provider/model resolution, candidate/failover choice, admission, and Pricing remain OAS-owned. Pricing is measurement-only.

## Validation

Passed:

- `git diff --check`
- `bash scripts/sync-oas-pin-manifests.sh --check --surface`
- API fingerprint matches `2186dfa5fca2360e4e99fa5e47052ba4c0cc687f`

Local environment limitation:

- `bash scripts/check-oas-pin.sh --local-only` confirmed manifests, then stopped because the installed opam pin is the unqualified repository URL rather than the new exact SHA. It was not mutated because local build/pin ownership remains with the operator.
- no local build or full test suite was run; PR CI is the compile authority.

## Release evidence

- OAS release: https://github.com/jeong-sik/oas/releases/tag/v0.226.0
- OAS #2815: https://github.com/jeong-sik/oas/pull/2815

PR #25822, opened 2026-07-27
build(deps): pin OAS v0.226.0 with its consumer migration (#25808, incomplete)

jeong-sik:
Draft. Does not merge as-is: `#25808` is not complete. Opened so the work stops being invisible — the two commits below existed only in a local unpushed worktree, so CI had never seen them.

## Why the pin rides along

`#25799` bumped the OAS pin to v0.226.0 alone. `main` could not compile and `#25810` reverted it. The in-flight migration then had the mirror-image problem: consumer code written against v0.226 while `scripts/oas-agent-sdk-pin.sh` still declared `0.223.1`, so CI would have compiled it against the wrong SDK. Pin and consumers only make sense together, so this branch carries both.

Pin: `v0.223.1` (`3d4ee19a`) → `v0.226.0` (`2186dfa5`, the `release 0.226.0` commit on `oas` main). `scripts/sync-oas-pin-docs.sh` and `scripts/sync-oas-pin-manifests.sh` ran; `scripts/oas-api-surface.json` is deliberately **not** regenerated because that needs an actual review of the upstream surface delta, not a blind `--surface`.

## Inherited commits

`20c786153e` and `6d9d69b06d` are not mine. They are the existing migration work — receipt facts, and the current-schema preference evidence journal that `#25808` requires — carried here unchanged with their authorship.

## What this commit adds

- `keeper_turn_runtime_budget.mli` exposes `Provider_request_body_refusal`, which the `.ml` already carried. Its doc no longer claims the set covers only two measured axes.
- `keeper_librarian_runtime` handles the three execution-error variants v0.226 adds, renders `Flow_preference_reservation_exhausted` distinctly, and carries a reason on `Exact_domain_settlement_failed`.
- `commit_and_settle_flow_domain` wired to the evidence journal.

## What is NOT done

The durable ordering clause of `#25808` is unmet everywhere. Measured, not estimated — this is the actual `dune build @check` output against the v0.226 switch:

| File | Line | Remaining |
|---|---|---|
| `keeper_librarian_runtime.ml` | 899 | `execute_flow_once` needs `~before_measurement_dispatch` / `~on_measurement_terminal` |
| `keeper_librarian_runtime.ml` | 819 | Domain_valid callback must commit **the episode**, not only the intent |
| `keeper_board_attention_exact_flow.ml` | 439 | same measurement callbacks |
| `keeper_board_attention_exact_flow.ml` | 452-462 | both dispositions still settle before the durable terminal |
| `hitl_summary_worker.ml` | 717 | `Unbound record field catalog_generation` |
| `hitl_summary_worker.ml` | 766 | completion write must move into the commit callback |
| `keeper_compaction_llm_summarizer.ml` | 585 | `Domain_already_settled` removed from the closed set |
| `keeper_unified_turn.ml` | 220, 239 | `Provider_request_body_refusal` unhandled — see below |

Plus the acceptance tests `#25808` requires: commit error, cancellation, restart replay, duplicate same-intent replay, conflict, retirement ordering.

## Two findings for whoever continues

**1. Nesting a scope boundary inside the OAS commit callback is safe.** The ordering clause requires calling `append_episode_for_current_owner` (which takes `with_settlement`) from inside `commit_and_settle_flow_domain ~commit`, while the caller may already hold a `with_current` pin. `keeper_exact_flow_scope.ml:390` `with_boundary` takes `Keeper_lifecycle_reservation.with_key_lock` only around pin acquisition and releases it before running the callback; `boundary_users` is a counter. OAS separately documents that no callback runs while the preference mutex is held. So there is no lock held across the nesting and no deadlock path. This was the blocking unknown and it is resolved.

**2. `keeper_unified_turn.ml` is a durable-schema decision, not a compile fix.** `#25808` does not mention it. The two arms map a `capacity_refusal` onto `Compaction_trigger.t`, and `Compaction_trigger.t` is encoded into durable records through `to_detail_json` / `of_detail_json`. A provider body refusal declares no limit, so it cannot be folded into `Request_body_over_capacity { actual_bytes; limit_bytes }` without fabricating the integers that `#25809` explicitly refused to fabricate, and it cannot be dropped without silently withholding compaction from a request the provider refused for size. Adding a variant is a durable codec change and a compaction-policy decision, so it is left for an explicit decision rather than made here.

## Verification

- `git diff --check`, `ocamlformat --check` on the touched files: pass
- `bash scripts/check-oas-pin.sh`: manifests match `v0.226.0`
- `dune build @check` against the v0.226 switch: the table above is its output

One operational note: the shared opam switch moved from `0.226.0` to `0.223.1` partway through this work without any action from this session. A single global switch cannot serve both `main` (which needs `0.223.1`) and this branch (which needs `0.226.0`), so local builds here are only reproducible while the switch happens to hold the matching pin.

RFC-WAIVED: no credential, identity, operator control-plane, sandbox, hooks, or workflow surface is touched; the OAS migration itself is specified in #25808.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # MASC
 
 [![OCaml](https://img.shields.io/badge/OCaml-%3E%3D%205.4-orange.svg)](https://ocaml.org/)
-[![agent_sdk](https://img.shields.io/badge/agent__sdk-%3E%3D%200.223.1-blue.svg)](https://github.com/jeong-sik/oas)
+[![agent_sdk](https://img.shields.io/badge/agent__sdk-%3E%3D%200.226.0-blue.svg)](https://github.com/jeong-sik/oas)
 [![License](https://img.shields.io/badge/license-MIT-green.svg)](LICENSE)
 
 [한국어 버전](README.ko.md)
--- a/docs/KEEPER-USER-MANUAL.md
+++ b/docs/KEEPER-USER-MANUAL.md
@@ -34,7 +34,7 @@ Keeper는 OAS(OCaml Agent SDK) 위에 구축된다. OAS가 제공하는 핵심 �
 | `Event_bus.t` | 에이전트 간 이벤트 발행/구독 | `oas_events.ml`을 통해 broadcast, heartbeat, board 이벤트 전달 |
 
 <!-- BEGIN GENERATED: oas-pin-manual -->
-OAS pin metadata is generated from `scripts/oas-agent-sdk-pin.sh`. Current dependency floor: `agent_sdk >= 0.223.1`, runtime pin: `main@3d4ee19a4773cbd8b1b73ed5555c5cdfe82e6d77`, declared base version: `v0.223.1`. 최신성 검증이 필요할 때는 문서에 적힌 숫자보다 `dune-project`와 pin script를 우선 truth source로 본다.
+OAS pin metadata is generated from `scripts/oas-agent-sdk-pin.sh`. Current dependency floor: `agent_sdk >= 0.226.0`, runtime pin: `main@2186dfa5fca2360e4e99fa5e47052ba4c0cc687f`, declared base version: `v0.226.0`. 최신성 검증이 필요할 때는 문서에 적힌 숫자보다 `dune-project`와 pin script를 우선 truth source로 본다.
 <!-- END GENERATED: oas-pin-manual -->
 
 #### 1.1.1 OAS 환경 변수 경계
--- a/dune-project
+++ b/dune-project
@@ -59,7 +59,7 @@
   ;; ordered multimodal delivery contracts;
   ;; its SHA and rationale live in scripts/oas-agent-sdk-pin.sh (SSOT).
   ;; See check-oas-pin.sh.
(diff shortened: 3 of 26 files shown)


jeong-sik:
CI 로 잔여 목록을 확인했다. `workflow_dispatch` run [30234853030](https://github.com/jeong-sik/masc/actions/runs/30234853030) 의 `Build and Test` 출력이 본문 표와 일치한다 (`keeper_librarian_runtime.ml` 만 이 브랜치 편집으로 899 → 904 이동).

```
lib/keeper/hitl_summary_worker.ml:717
lib/keeper/keeper_board_attention_exact_flow.ml:439
lib/keeper/keeper_compaction_llm_summarizer.ml:585
lib/keeper/keeper_librarian_runtime.ml:904-906
lib/keeper/keeper_unified_turn.ml:220-224
lib/keeper/keeper_unified_turn.ml:239-242
```

이 run 은 `pull_request` 이벤트가 run 을 생성하지 않아 `workflow_dispatch` 로 직접 띄운 것이다. 측정용이며 required check 로는 집계되지 않는다.

jeong-sik:

PR #25833, opened 2026-07-27
feat(keeper): define typed capacity compaction request contract

jeong-sik:
## Summary

First behavior-free stack PR for #25725.

- classify current typed Agent SDK input-capacity failures without prose matching
- preserve token serving-boundary and serialized request-byte provenance as typed compaction triggers
- define a strict current-schema capacity compaction request codec and stable identity
- keep failover-only and measurement-unavailable outcomes explicitly non-compacting

## Boundary

This PR intentionally does not add the request to stimulus_payload, enqueue it, execute compaction, or change runtime behavior. Production wiring and removal of the obsolete inline recovery path belong to the next stack PR.

No provider/model policy, legacy JSON migration, compatibility default, or error-string heuristic is introduced.

## Validation

- git diff --check passed
- focused local builds were attempted through scripts/dune-local.sh
- build execution was blocked before compilation by the shared agent_sdk pin guard: installed 2186dfa5fca2360e4e99fa5e47052ba4c0cc687f, repository expected 3d4ee19a4773cbd8b1b73ed5555c5cdfe82e6d77
- shared pin was deliberately left unchanged

Stack: PR-A of PR-A/PR-B/PR-C.

### DIFF
--- a/lib/compaction_trigger/compaction_trigger.ml
+++ b/lib/compaction_trigger/compaction_trigger.ml
@@ -144,7 +144,11 @@ type decode_error =
       { input_tokens : int
       ; accepted_through : int
       }
-  | Invalid_serving_capacity_boundary of
+  | Invalid_boundary_unknown of
+      { input_tokens : int
+      ; rejected_from : int
+      }
+  | Invalid_input_rejected_boundary of
       { input_tokens : int
       ; accepted_through : int
       ; rejected_from : int option
@@ -180,10 +184,15 @@ let decode_error_to_string = function
       "serving input capacity requires input_tokens > accepted_through, got %d and %d"
       input_tokens
       accepted_through
-  | Invalid_serving_capacity_boundary
+  | Invalid_boundary_unknown { input_tokens; rejected_from } ->
+    Printf.sprintf
+      "serving input capacity boundary_unknown requires rejected_from > input_tokens, got input=%d rejected=%d"
+      input_tokens
+      rejected_from
+  | Invalid_input_rejected_boundary
       { input_tokens; accepted_through; rejected_from } ->
     Printf.sprintf
-      "serving input capacity rejected_from must be greater than accepted_through and no greater than input_tokens, got input=%d accepted=%d rejected=%s"
+      "serving input capacity input_rejected requires accepted_through < rejected_from <= input_tokens, got input=%d accepted=%d rejected=%s"
       input_tokens
       accepted_through
       (match rejected_from with
@@ -236,7 +245,12 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
      | Some (`String "serving_input_capacity") ->
        let* () =
          reject_unknown_fields
-           [ "kind"; "reason"; "input_tokens"; "accepted_through"; "rejected_from" ]
+           [ kind_field
+           ; reason_field
+           ; input_tokens_field
+           ; accepted_through_field
+           ; rejected_from_field
+           ]
        in
        let integer_field ~positive field =
          match List.assoc_opt field fields with
@@ -245,20 +259,22 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
          | None -> Error (Missing_serving_capacity_field field)
          | Some _ -> Error (Invalid_serving_capacity_field field)
        in
-       let* input_tokens = integer_field ~positive:true "input_tokens" in
-       let* accepted_through = integer_field ~positive:false "accepted_through" in
+       let* input_tokens = integer_field ~positive:true input_tokens_field in
+       let* accepted_through =
+         integer_field ~positive:false accepted_through_field
+       in
        let* reason =
-         match List.assoc_opt "reason" fields with
+         match List.assoc_opt reason_field fields with
          | Some (`String value) -> Ok value
-         | None -> Error (Missing_serving_capacity_field "reason")
-         | Some _ -> Error (Invalid_serving_capacity_field "reason")
+         | None -> Error (Missing_serving_capacity_field reason_field)
+         | Some _ -> Error (Invalid_serving_capacity_field reason_field)
        in
        let* rejected_from =
-         match List.assoc_opt "rejected_from" fields with
+         match List.assoc_opt rejected_from_field fields with
          | Some `Null -> Ok None
          | Some (`Int value) when value > 0 -> Ok (Some value)
-         | None -> Error (Missing_serving_capacity_field "rejected_from")
-         | Some _ -> Error (Invalid_serving_capacity_field "rejected_from")
+         | None -> Error (Missing_serving_capacity_field rejected_from_field)
+         | Some _ -> Error (Invalid_serving_capacity_field rejected_from_field)
        in
        if input_tokens <= accepted_through
        then Error (Serving_capacity_not_over_limit { input_tokens; accepted_through })
@@ -277,10 +293,13 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
                     ; accepted_through
                     ; rejected_from = Some rejected_from
                     }))
-          | "boundary_unknown", _
+          | "boundary_unknown", Some rejected_from ->
+            Error
+              (Invalid_boundary_unknown
+                 { input_tokens; rejected_from })
           | "input_rejected", None ->
             Error
-              (Invalid_serving_capacity_boundary
+              (Invalid_input_rejected_boundary
                  { input_tokens; accepted_through; rejected_from })
           | "input_rejected", Some rejected_from
             when rejected_from > accepted_through
@@ -291,9 +310,9 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
                     { input_tokens; accepted_through; rejected_from }))
           | "input_rejected", Some _ ->
             Error
-              (Invalid_serving_capacity_boundary
+              (Invalid_input_rejected_boundary
                  { input_tokens; accepted_through; rejected_from })
-          | _, _ -> Error (Invalid_serving_capacity_field "reason"))
+          | _, _ -> Error (Invalid_serving_capacity_field reason_field))
      | Some (`String "manual") ->
        let* () = reject_unknown_fields [ kind_field ] in
        Ok Manual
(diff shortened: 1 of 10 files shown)


lib/compaction_trigger/compaction_trigger.ml
```diff
@@ -128,6 +208,64 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
...
+         | Some _ -> Error (Invalid_serving_capacity_field "rejected_from")
+       in
+       if input_tokens <= accepted_through
+       then Error (Serving_capacity_not_over_limit { input_tokens; accepted_through })
+       else
+         let boundary_valid =
+           match rejected_from with
+           | None -> true
+           | Some value ->
+             value > accepted_through && value <= input_tokens
```
jeong-sik: [P1] Preserve the actual `Boundary_unknown` invariant. OAS `Serving_constraint.admit` emits `Boundary_unknown { rejected_from = Some r }` only after the `input_tokens >= r` guard failed, so its invariant is `accepted_through < input_tokens < r`. This shared check requires the opposite (`r <= input_tokens`), which rejects a valid current-schema value on reload and accepts a state OAS would classify as `Input_rejected`. Validate per reason: `boundary_unknown` requires `input_tokens < rejected_from` when present; `input_rejected` requires `rejected_from <= input_tokens`. Add round-trip tests for both sides with `rejected_from = Some _`.

lib/keeper_runtime/keeper_event_queue.ml
```diff
@@ -178,6 +178,56 @@ and goal_assignment = {
...
+
+let capacity_compaction_request_identity_equal left right =
+  left.ccr_trigger = right.ccr_trigger
+
+let capacity_compaction_request_post_id request =
+  let digest =
+    request
+    |> capacity_compaction_request_to_yojson
+    |> Yojson.Safe.to_string
+    |> Digest.string
```
jeong-sik: [P1] Do not use MD5 for a durable deduplication identity. OCaml `Digest.string` is MD5; a collision makes two distinct compaction obligations share one `post_id` and one can disappear silently. This repository already uses `Digestif.SHA256` for durable/canonical identities. Hash the canonical, domain-separated request bytes with SHA-256 (and keep the full 64 hex chars), then test the exact stable encoding and distinct typed inputs.

lib/keeper/keeper_turn_runtime_budget.ml
```diff
@@ -447,6 +447,104 @@ let capacity_refusal_of_error (err : Agent_sdk.Error.sdk_error) : capacity_refus
...
+      }
+  | Token_measurement_unavailable of
+      { protocol : Llm_provider.Input_token_count.protocol }
+
+type capacity_transition =
+  | Not_capacity
+  | Compact_next_cycle of Compaction_trigger.t
+  | Capacity_non_compacting of capacity_non_compaction
+
+let capacity_transition_of_error
```
jeong-sik, line 427: [P1] Keep one canonical capacity classifier. `capacity_refusal_of_error` immediately above already exhaustively matches the same `sdk_error` sum, including the same ContextOverflow and Request_body_too_large cases, while this second public exhaustive switch deliberately differs for InputCapacity. The next SDK constructor or policy edit can therefore update one authority and silently drift from the other. Make this transition the single exhaustive classifier and derive the older projection from it (or remove the older public surface in this stack); do not leave two co-equal switches/facades.

lib/keeper/keeper_turn_runtime_budget.mli
```diff
@@ -209,6 +209,33 @@ val capacity_refusal_of_error :
...
+  | Serving_evidence_not_yet_valid of
+      { now_unix_s : int
+      ; checked_at_unix_s : int
+      }
+  | Serving_evidence_expired of
+      { now_unix_s : int
+      ; expires_at_unix_s : int
+      }
+  | Token_measurement_unavailable of
+      { protocol : Llm_provider.Input_token_count.protocol }
```
jeong-sik: [P2] Keep provider wire-measurement detail out of the MASC result surface. `Llm_provider.Input_token_count.protocol` contains provider-specific constructors such as `Anthropic_messages_count_tokens`; exporting it here makes a supposedly provider-neutral MASC transition carry OAS/provider dialect. MASC only needs the typed fact that measurement is unavailable to decide non-compaction. Keep the protocol in the OAS receipt/observability record and expose a payload-free provider-neutral constructor (or an opaque receipt reference) here.

lib/keeper_runtime/keeper_event_queue.mli
```diff
@@ -201,6 +201,26 @@ and goal_assignment = {
 }
 (** Payload for [Goal_assigned]. *)
 
+type capacity_compaction_request = {
+  ccr_trigger : Compaction_trigger.t;
```
jeong-sik, line 205: [P0] This public type does not compile in the `masc_keeper_runtime` library. Current-head Build and Test fails here with `Error: Unbound module Compaction_trigger` because `lib/keeper_runtime/dune` does not depend on the leaf `masc_compaction_trigger` library. Add the dependency (the leaf has no reverse Keeper_event_queue dependency, so no cycle is visible), or move the request contract to a layer whose dependency direction is already legal. Re-run the focused keeper_runtime interface build before updating the PR.

lib/compaction_trigger/compaction_trigger.ml
jeong-sik: Fixed at head 86a6f81829. The decoder now applies reason-specific inequalities: Boundary_unknown(Some r) requires accepted_through < input_tokens < r, while Input_rejected requires accepted_through < r <= input_tokens. Focused codec coverage now round-trips both valid sides and rejects each inverted boundary.

lib/keeper_runtime/keeper_event_queue.ml
jeong-sik: Fixed at head 86a6f81829. Queue identity now hashes the canonical current-schema JSON with a NUL-separated keeper.capacity_compaction_request.identity.v1 domain using Digestif SHA-256, retaining all 64 hex characters. Tests pin the exact digest and prove distinct typed evidence produces a distinct post id.

lib/keeper/keeper_turn_runtime_budget.ml
jeong-sik: Fixed at head 86a6f81829. capacity_transition_of_error is now the only exhaustive sdk_error classifier. capacity_refusal_of_error is a narrow projection of that canonical typed result, so new SDK variants have one decision point.

lib/keeper/keeper_turn_runtime_budget.mli
jeong-sik: Fixed at head 86a6f81829. Token_measurement_unavailable is now a payload-free MASC transition fact. The provider protocol is consumed only while matching the OAS SDK error and is not retained in MASC public or durable capacity state.

lib/keeper_runtime/keeper_event_queue.mli
jeong-sik: Fixed at head 86a6f81829 by declaring masc_compaction_trigger as an exact masc_keeper_runtime dependency. This keeps Compaction_trigger.t as the single typed request-evidence SSOT instead of duplicating a second queue-local trigger type.

lib/compaction_trigger/compaction_trigger.ml
```diff
@@ -97,6 +171,24 @@ let decode_error_to_string = function
...
+    Printf.sprintf "serving input capacity trigger has invalid %s" field
+  | Serving_capacity_not_over_limit { input_tokens; accepted_through } ->
+    Printf.sprintf
+      "serving input capacity requires input_tokens > accepted_through, got %d and %d"
+      input_tokens
+      accepted_through
+  | Invalid_serving_capacity_boundary
+      { input_tokens; accepted_through; rejected_from } ->
+    Printf.sprintf
+      "serving input capacity rejected_from must be greater than accepted_through and no greater than input_tokens, got input=%d accepted=%d rejected=%s"
```
jeong-sik: [P2] Keep the boundary diagnostic reason-specific. After the invariant fix, this message is only correct for Input_rejected. A rejected Boundary_unknown(Some r) is valid only when r > input_tokens, so telling operators "no greater than input_tokens" prescribes the inverse state. Preserve the reason in decode_error or split the errors and render the exact inequality for each variant; add message assertions for both rejected branches.

lib/compaction_trigger/compaction_trigger.ml
```diff
@@ -141,6 +233,67 @@ let of_detail_json (json : Yojson.Safe.t) : (t, decode_error) result =
        if actual_bytes > limit_bytes
        then Ok (Request_body_over_capacity { actual_bytes; limit_bytes })
        else Error (Request_body_within_capacity { actual_bytes; limit_bytes })
+     | Some (`String "serving_input_capacity") ->
+       let* () =
+         reject_unknown_fields
+           [ "kind"; "reason"; "input_tokens"; "accepted_through"; "rejected_from" ]
```
jeong-sik: [P2] Finish the wire-field SSOT. The constants and wire_field_names now drive the encoder/public projection, but the strict decoder restates all five new names as raw strings here and below. That preserves exactly the encoder/decoder drift the module comment says the constants prevent. Use kind_field, reason_field, input_tokens_field, accepted_through_field, and rejected_from_field for the allow-list, lookups, and field errors so encoder, decoder, and projection share one spelling authority.

lib/compaction_trigger/compaction_trigger.ml
jeong-sik: Fixed at head ddcccabf16. The decode error is now split by typed reason: Invalid_boundary_unknown renders rejected_from > input_tokens, while Invalid_input_rejected_boundary renders accepted_through < rejected_from <= input_tokens. The two invalid-side tests assert the complete reason-specific diagnostic text.

lib/compaction_trigger/compaction_trigger.ml
jeong-sik: Fixed at head ddcccabf16. The serving exact-field allow-list, lookups, and missing/invalid field errors now use kind_field, reason_field, input_tokens_field, accepted_through_field, and rejected_from_field. Encoder, strict decoder, and wire_field_names projection therefore share the same spelling constants without a new facade.

PR #25834, opened 2026-07-27
test: isolate operator snapshot workspace roots

jeong-sik:
## Summary

- Scope each operator-control fixture's `MASC_BASE_PATH` and `MASC_BASE_PATH_INPUT` to its own temporary workspace root.
- Restore the prior value, including the genuinely-unset state, from the existing `cleanup_dir` finally path.
- Keep production workspace/lifecycle authority semantics and all 49 test expectations unchanged.

## Why

The post-merge `Build and Test` run for #25819 opened compilation and then exposed a test-isolation failure in `test_operator_control_snapshot`: 40 of 49 cases failed during setup.

The first fixture synchronized the process-global base-path authority to `/tmp/.../test_operator_control_609644` and later removed that root. Subsequent fixtures requested a different temp root but logged `Ignoring test MASC_BASE_PATH override=...609644 for requested path ...`, followed by `authority_unreadable:authority_read_failed`. That is stale one-root state leaking across fixtures in a single executable, not a reason to relax production authority or add a legacy fallback.

`test_operator_control_support.temp_dir` now captures both base-path environment variables before installing the fixture root. The existing `cleanup_dir` path restores the captured values in `Fun.protect`, so every fixture enters `Workspace.default_config` with its own live root and leaves no process-global authority behind.

## Validation

- `ocamlformat --check test/test_operator_control_support.ml`: pass
- `git diff --check origin/main..HEAD`: pass
- Focused wrapper requested for the three operator snapshot executables: environment-blocked before compilation because another session is running bare `dune build .`; that process was not interrupted or bypassed. PR CI is the build authority.

RFC-WAIVED: test-support isolation only; no production credential, identity, operator-control, sandbox, hooks, or workflow behavior changes.

## Typed authority follow-up

The first CI pass on this PR proved the environment scope works: every fixture logged its own explicit temp root. It still failed 40/49 because the first durable admission point-read could not stabilize its capability HEAD parent.

Static path evidence rules out a writer/reader split: `Keeper_dead_revival_transaction` and `Keeper_lifecycle_admission_journal` both derive the same `Workspace.masc_root_dir config/keeper-lifecycle-transactions` parent and the same domain-separated hashed `revival-*.json` leaf. The first failing fixture has no HEAD row yet, so its typed expected result is `snapshot_row = None`.

A standalone execution of the current `Capability_head.read` against the same `Eio.Stdenv.fs / absolute-parent` shape returned `Unsupported "parent directory capability has no Unix file descriptor"`. The exact expected parent contained the newly-created stable lock at mode `0600`, size `0`; initialization stopped at parent fsync before the lock epoch could be written.

Nonce and dead-revival authorities already solve this in tests with Fiber-scoped `For_testing.with_fd_backed_parent_opening`. Durable admission now exposes and uses the same scoped hook around both exact point-reads. Operator fixtures and the nonce admission suite opt into that test scope. Production admission decisions, fail-closed behavior, authority rows, and all test expectations remain unchanged.

### Latest validation

- `ocamlformat --check` on all six follow-up files: pass
- `ocamlc -stop-after parsing` on all six follow-up files: pass
- `git diff --check`: pass
- Focused operator + nonce wrapper: environment-blocked before compilation by shared `agent_sdk` pin drift (`2186dfa5...` installed, `3d4ee19...` expected). The shared pin was not changed or bypassed after the patch; CI is authoritative.

### DIFF
--- a/lib/keeper/keeper_lifecycle_admission_durable_transaction.ml
+++ b/lib/keeper/keeper_lifecycle_admission_durable_transaction.ml
@@ -179,10 +179,11 @@ let with_revival_launch_admission_under_lock
     (match journal_parent config, journal_entropy () with
      | Ok parent, Ok secure_random ->
        (match
-          Head.read
-            ~secure_random
-            ~parent
-            ~leaf:(journal_leaf keeper_name)
+          with_head_parent parent (fun parent ->
+            Head.read
+              ~secure_random
+              ~parent
+              ~leaf:(journal_leaf keeper_name))
         with
         | Ok snapshot
           when Head.snapshot_settlement_warnings snapshot <> [] ->
@@ -389,4 +390,5 @@ let projection_to_yojson projection =
 
 module For_testing = struct
   let permit_matches = permit_matches
+  let with_fd_backed_parent_opening = with_fd_backed_parent_opening
 end
--- a/lib/keeper/keeper_lifecycle_admission_durable_transaction.mli
+++ b/lib/keeper/keeper_lifecycle_admission_durable_transaction.mli
@@ -122,4 +122,5 @@ val projection_to_yojson : projection -> Yojson.Safe.t
 
 module For_testing : sig
   val permit_matches : permit -> Workspace.config -> string -> bool
+  val with_fd_backed_parent_opening : (unit -> 'a) -> 'a
 end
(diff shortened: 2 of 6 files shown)


jeong-sik:

PR #25841, opened 2026-07-27
fix(keeper): restore capacity compaction full cycle

jeong-sik:
## Summary

- preserve typed token, byte, and serving-capacity transitions at the Keeper compaction entry
- let scheduled/proactive turns dispatch compaction under the existing Keeper lifecycle and checkpoint CAS, without inventing a queue lease
- retain the existing durable guard callbacks only when a queue-owned source lease exists
- remove the dead `capacity_compaction_request` queue DTO/codec/test and the now-unreachable absent-guard rejection surface
- use the existing checkpoint install and source requeue path; add no Store, lease, TTL, renewal, or commit facade
- forward-merge current `main`, superseding the stale red merge base in #25833

## Validation

- `scripts/dune-local.sh build lib/masc.cmxa test/test_keeper_overflow_recovery.exe test/test_compaction_exact_output_conformance.exe test/test_compaction_exact_output_entrypoint_clockless.exe test/test_keeper_meta_current_keyset.exe test/test_keeper_meta_current_schema.exe`

## Live acceptance still required

After the OAS wire-classification release is pinned and the user restarts MASC:

`HTTP 400 capacity → ContextOverflow → Compaction_started → checkpoint installed → next Keeper turn succeeds`

[근거] `/Users/dancer/me/.masc/logs/system_log_2026-07-27.jsonl`, direct Ollama Cloud overflow probe, current-main focused build, checked 2026-07-27 KST, confidence High.


jeong-sik:
Forward update at `c10950b825`:

- consumes OAS `v0.227.1` (`8351a0a8f`), which preserves the two exact observed Ollama Cloud 400 overflow grammars as typed `ContextOverflow`
- automatic capacity recovery now enters `Compaction_started`, rechecks existing registry authority (`Compacting + trace_id + transition_seq`) immediately before POST, and rejects missing/stale authority without dispatch
- queue lease guard remains additive for queue-backed turns; autonomous turns do not manufacture a lease/store/commit layer
- post-install cancellation preserves committed recovery and converges through `Compaction_completed -> Running`
- pre-install/terminal cleanup failures return an explicit typed `lifecycle_cleanup_failed` outcome with the original source disposition; no exception-driven fiber crash, silent success, or duplicate POST
- manifest no longer claims `source_requeued=true` before settlement
- current-schema test fixtures were repaired without weakening the production parser

Focused evidence:

- touched targets build: pass
- `test_compaction_exact_output_conformance`: 20/20 pass
- `test_keeper_overflow_recovery`: 9/9 pass
- `test_compaction_exact_output_entrypoint_clockless`: 2/2 pass
- `test_keeper_post_turn_wirein_order`: 14/15 pass; sole failure occurs before compaction at pre-existing Darwin lifecycle authority read (`authority_unreadable:authority_read_failed`)
- standalone `test_keeper_lifecycle_nonce` reproduces the same main-branch Darwin substrate failure across its current-schema cases, so it is not attributed to this compaction path

Live acceptance remains intentionally unclaimed until this head is built/restarted and logs show:

`HTTP 400 -> ContextOverflow -> Compaction_started -> exact POST -> checkpoint installed -> Compaction_completed -> next Keeper turn`

PR #25956, opened 2026-07-27
fix(goal): complete verifier metric from approved task

jeong-sik:
## Summary
- complete linked executing Goal when an approval produces a Done task for `verifier_approved_done_tasks=1`
- keep task commit authoritative when the separate Goal update fails
- prove verifier approval completes the Goal while direct Done does not

## Boundary
- no new store, lease, receipt, settlement, or facade
- exact existing Goal index and Goal Store only

## Validation
- `ocamlformat --check lib/workspace/workspace_task_transitions.ml test/test_workspace.ml`
- `git diff --check`
- focused Dune build attempted once but skipped after shared `/tmp/me-dune-local.lock` remained held by another worktree full build; CI is build authority

### DIFF
--- a/lib/workspace/workspace_task_transitions.ml
+++ b/lib/workspace/workspace_task_transitions.ml
@@ -574,6 +574,59 @@ let transition_task_outcome_r
             Masc_domain.task_status_is_done new_status
             && not (Masc_domain.task_status_is_done task.task_status)
           in
+          (* A verifier-approved task is the event counted by the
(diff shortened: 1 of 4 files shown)


jeong-sik:
Closing before merge: direct Task-transition -> Goal_store feedback creates the cross-domain coupling we explicitly do not want. The forward fix will keep Task authoritative and move metric projection to a Goal-owned consumer of the existing typed Approved event, with no new store/lease/settlement layer. Tracking reconciliation semantics in #25957.

PR #25967, opened 2026-07-27
fix(keeper): make detached context correction race-safe

jeong-sik:
## Summary

Supersedes closed PR #25960.

Keeps the same provider/model-free context-correction lane, but fixes the P1 lifecycle, coalescing, persistence, and primary-turn coupling gaps before landing it.

## P1 fixes

1. Detached child yields as its first operation, so caller work cannot execute inline during `submit`.
2. Each Keeper owns one capacity-one trailing-edge mailbox. While a correction runs, repeated submissions replace `latest`; the newest raw checkpoint is processed next instead of being dropped.
3. The bootstrap executor is generation-tagged. `Switch.on_release` uses identity compare-and-clear, so a stale switch cannot clear or accept work for a newer executor generation.
4. Scheduled and terminal facts use the existing runtime-manifest SSOT. Terminal rows retain status, raw source ref, installed ref, provider-neutral OAS receipt fingerprints, and typed failure detail. No new store, lease, journal, commit, or settlement layer was added.
5. Deadline arithmetic uses `Eio.Time.now clock`; wall time is absent from the deterministic boundary.
6. The transient `capture_start` disk read and its hard-cut state were removed. The already-loaded resume checkpoint supplies the marker, raw persistence is the durable retry marker, and a failed correction leaves it unchanged for a later turn.
7. Primary Keeper execution performs no new correction snapshot I/O. Exact source read, OAS flow execution, and CAS installation happen only in the detached root-switch worker.

## Boundaries

- MASC sees only opaque Runtime lane slots and validates semantic output.
- OAS retains candidate resolution, provider wire behavior, and ordered failover ownership.
- Closed tool units and the exact checkpoint CAS remain the sole transformation/install boundary.
- No provider/model branching, migration compatibility, Failure Judge, or Pricing logic.

## Focused validation

- `scripts/dune-local.sh build test/test_keeper_context_correction.exe lib/masc.cmxa lib/server/masc_server.cmxa`
- `scripts/dune-local.sh exec ./test/test_keeper_context_correction.exe`
- 10/10 tests pass, including initialized non-inline execution, latest-wins trailing edge, stale-switch generation isolation, typed terminal evidence, and exact CAS preservation.

Live runtime proof remains pending deployment; this PR claims source and focused local proof only.

### DIFF
--- a/config/runtime.toml
+++ b/config/runtime.toml
@@ -29,6 +29,9 @@ slots = ["glm-coding.glm-5-turbo", "deepseek.deepseek-v4-pro"]
 [runtime.exact_output_lanes.compaction_exact]
 slots = ["glm-coding.glm-5-turbo", "deepseek.deepseek-v4-pro"]
 
+[runtime.exact_output_lanes.context_correction_exact]
+slots = ["glm-coding.glm-5-turbo", "deepseek.deepseek-v4-pro"]
+
 [runtime.exact_output_lanes.hitl_auto_judge]
 slots = ["glm-coding.glm-5-turbo", "glm-coding.glm-4-7-coding"]
 
--- a/lib/keeper/keeper_agent_run.ml
+++ b/lib/keeper/keeper_agent_run.ml
@@ -367,13 +367,8 @@ let run_turn
       ~generation
       ()
   in
-  let context_correction_start =
-    Keeper_context_correction.capture_start
-      ~session_dir:ctx.session_dir
-      ~session_id:(Keeper_id.Trace_id.to_string meta.runtime.trace_id)
-  in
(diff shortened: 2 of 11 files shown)


jeong-sik:
Closed before merge: focused tests/build passed, but the feature has no live Keeper Full Cycle proof, no measured semantic-quality or context-reduction evidence, no deployed nonblocking proof, and no live failover/CAS-install/next-turn observation. The P1 follow-up added 831 lines before proving the product need. Preserve the branch as an experiment; reintroduce only after the basic 8-Keeper lifecycle, Task/Verifier/Goal, and runtime failover work live.

PR #25976, opened 2026-07-27
fix(keeper): fence wake intake with turn admission

jeong-sik:
## Summary
- acquire the existing Keeper turn admission before stimulus intake side effects
- revalidate active token, durable pause state, and exact pending snapshot before general and compaction provider dispatch
- serialize paused Cancel/Transfer/Settle mutations on the same admission authority
- hard-cut fresh event queue persistence to pending-only v7 and remove heartbeat outbox gating

## Validation
- `scripts/dune-local.sh build lib/masc.cmxa`
- pending v7: 7/7
- keeper turn admission: pass, including zero-dispatch authority rejection
- paused cancellation: 10/10
- paused transfer: 3/3
- paused source-terminal: 4/4
- changed-file `ocamlformat --check`
- `git diff --check`


jeong-sik:
Superseded by #25978, rebuilt from current origin/main after #25969 merged.

PR #26037, opened 2026-07-27
fix(hitl): stop replaying observed resolution wakes

### LINKED ISSUE #26030: Consumed HITL wake is replayed on every restart while exact grant remains unused
Symptom: after an approved external effect wake was already consumed and acknowledged, restarting MASC re-enqueues the same hitl-approval stimulus whenever the exact one-shot grant was not consumed. The stale wake resumes an old conversation and can create unrelated new approval requests.\n\nReproduction: approve a WebSearch, let the Keeper choose a different exact input so the original grant remains unused, complete and acknowledge the wake, restart, then observe keeper_gate replayed=1 and the old approval delivered again.\n\nCause: durable delivery state uses grant_consumed as both authorization lifetime and wake-delivery acknowledgement. install_persistence replays every unconsumed delivery and does not consult the reaction-ledger event_queue_ack evidence.\n\nAttempted: isolated runtime restart confirmed the event queue was empty before startup but the same approval ID was re-enqueued. Preserve the unused exact grant, but suppress wake replay once its stimulus has durable acknowledgement evidence.

jeong-sik:
## Problem

An approved but unconsumed exact grant remains in the durable delivery journal by design. On every process restart, install replayed its `hitl-approval:<id>` wake even when that wake had already started and completed a Keeper turn. This reopened stale conversation context and could repeatedly drive unrelated work.

The recent peek/ack event-queue model keeps an unacknowledged source pending itself, so once a durable turn-start proves delivery, the approval journal no longer owns wake retry. The exact grant must still remain available until the matching tool consumes it.

## Changes

- record durable turn-start evidence for HITL resolution stimuli, as already done for schedule wakes
- query only current-schema, semantically valid reaction-ledger rows for the exact post ID and stimulus kind
- skip install-time wake replay after delivery evidence exists
- keep conservative replay when evidence is absent, invalid, or unreadable
- preserve the unconsumed exact grant and prove it remains consumable

## Validation

- focused build: `scripts/dune-local.sh build test/test_keeper_approval_queue.exe`
- pre-delivery restart still replays exactly once
- post-delivery/ack restart replays zero wakes
- exact grant remains `Resolution_unconsumed` and consumes successfully
- focused cases 25 and 26 passed
- ocamlformat checks and `git diff --check` passed

Closes #26030

### DIFF
--- a/lib/keeper/keeper_approval_queue.ml
+++ b/lib/keeper/keeper_approval_queue.ml
@@ -3202,7 +3202,34 @@ let complete_delivery delivery =
                (match remove_delivery_from_store delivery with
                 | Error storage_error ->
                   Error (Persistence_failed { approval_id = id; storage_error })
-                | Ok () -> finish ()))))
+               | Ok () -> finish ()))))
+;;
+
+let delivery_wake_was_observed delivery =
+  let resolution : Keeper_event_queue.hitl_resolution =
+    { approval_id = delivery.entry.id
+    ; decision =
+        hitl_resolution_decision_of_approval_decision delivery.decision
+    ; channel = delivery.entry.continuation_channel
+    }
+  in
+  let post_id = Keeper_event_queue.hitl_resolution_post_id resolution in
+  match
+    Keeper_reaction_ledger.event_queue_turn_started_seen_for_source_result
+      ~base_path:delivery.entry.audit_base_path
+      ~keeper_name:delivery.entry.keeper_name
+      ~post_id
+      ~stimulus_kind:Keeper_reaction_ledger.Hitl_resolved
+  with
+  | Ok observed -> observed
+  | Error error ->
+    Log.Keeper.warn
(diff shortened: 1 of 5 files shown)


jeong-sik:
Live main confirmation from the 2026-07-28 `~/.masc` full-cycle run:

- runtime commit `4d7ee9e244` (does not include this PR)
- verifier event queue reached 18 pending `hitl_resolved` items for approvals that had already been observed/resolved
- the verifier kept replaying the same read-only `ls` request (same input hash `528f0a...`) through fresh Auto Judge approvals instead of advancing to the newly created typed verification request for task-002
- task-002 reached `awaiting_verification`, but verifier progress was starved behind the stale HITL-resolution backlog

This is production-shaped evidence for the liveness impact fixed here: the issue is not Auto Judge failing to decide (it approved every safe request); it is the already-observed resolution wake remaining pending and repeatedly dominating later durable work.

PR #26040, opened 2026-07-27
fix(keeper): compact closed tool cycles atomically

### LINKED ISSUE #26038: Compaction rejects tool-heavy OAS checkpoints as no_eligible_history
## Symptom

A live Keeper checkpoint with substantial, structurally closed tool history cannot be compacted. `masc_keeper_compact` queues successfully, but the turn terminates as:

```
compaction_rejected keeper=patched-fullcycle trigger=manual reason=no_eligible_history
manual compaction reached typed terminal: no_eligible_history
```

## Reproduction

Observed in the isolated full-cycle runtime on 2026-07-28:

1. Run a Keeper through WebSearch/WebFetch/Write/Execute turns.
2. Persist an OAS v8 checkpoint with 36 messages (12 user/assistant/tool triplets; every ToolUse has matching ToolResult) and about 121 KB serialized checkpoint data / about 39.5k reported input tokens.
3. Stop and explicitly restart the Keeper.
4. Call `masc_keeper_compact`.
5. The request is accepted and delivered, then terminates as `no_eligible_history` without an exact summarizer call.

## Suspected cause

`Keeper_compaction_llm_summarizer.eligible_source` only accepts an `Ordinary_message` containing pure Assistant `Text` blocks. Normal Keeper tool turns are partitioned into `Closed_tool_cycle` units, and provider responses commonly include `Thinking` plus `ToolUse`, so a fully valid tool-heavy checkpoint has zero eligible units.

This makes compaction unavailable precisely for long-running tool-using Keepers. It is not an invalid-structure case: all observed cycles are closed.

## Attempted / evidence

- Offline Keeper: the command first returned `keeper ... is not in the registry`.
- After explicit `masc_keeper_up`: request queued and woke the lane correctly.
- Live policy still returned terminal `no_eligible_history`.
- Checkpoint inspection confirmed closed assistant ToolUse / tool ToolResult pairs rather than an orphaned tool protocol suffix.

## Expected

Canonical `Closed_tool_cycle` units should be eligible as atomic compaction units. A summarize action should replace the entire closed cycle with one Assistant text summary, never leaving an orphan ToolUse or ToolResult. Ordinary user/system/current-stimulus content must remain protected.

jeong-sik:
## Summary

- admit each structurally closed tool protocol as one atomic compaction planner unit
- strip `Thinking`, `ReasoningDetails`, and `RedactedThinking` from standalone Assistant planner input while preserving ordered Text blocks
- apply one whole-unit `Keep` / `Drop` / `Summarize` decision without partial tool-cycle salvage
- fail closed for open cycles and typed media blocks rather than truncating or projecting binary payloads
- type current compaction manifest outcomes so failed attempts remain visible without fabricated evidence

## Execution accounting

- compute typed `plan_accounting` ephemerally from the accepted plan and original units
- compare exact expected message/ToolUse/ToolResult counts with the serialized checkpoint observations before evidence creation
- reject an expected-vs-observed mismatch as typed `Invalid_transition` after dispatch
- do not persist derived planner counters or use manifest evidence as recovery/execution authority
- retain balanced ToolUse/ToolResult removal and structurally possible message-accounting checks as observability validation

No legacy defaults, migration, or compatibility branch are included.

## Manifest and Dashboard boundary

Current `context_compacted` decisions have a closed outcome:

- `checkpoint_committed`: canonical `exact_evidence` is required
- `retry_without_checkpoint`: evidence is forbidden and a nonblank cause is required
- `lifecycle_cleanup_failed_without_checkpoint`: evidence is forbidden and a nonblank cause is required

The manifest and TypeScript readers reject unknown, missing, malformed, or contradictory outcome/evidence/cause shapes and canonicalizes committed evidence through `Keeper_compaction_evidence.of_json` / `to_json`. The API and Dashboard use the same outcome SSOT for labels and success/warning/error color; failed no-checkpoint rows stay visible with their cause rather than being dropped or shown as successful compactions.

## Boundary

- semantic summaries remain LLM-produced
- deterministic code only strips private reasoning, strips technical message metadata from provider projection, builds ordered source projection, and verifies structural/count invariants
- no heuristic summary, truncation, migration, CAS/install/live-delta, store, lease, settlement, or new hash
- a normalizing `Keep` is material only when private reasoning was actually removed; plain all-`Keep` plans remain invalid

## Validation

- changed-file `ocamlformat`; `git diff --check`
- focused OCaml wrapper build for manifest, Dashboard backend, and producer paths
- compaction summarizer: 17/17 green
- compaction evidence projection/accounting: 5/5 green
- runtime-manifest completeness/current outcome/evidence boundary: 7/7 green
- actual atomic accounting success gate: focused 1/1 green
- ephemeral expected-vs-observed mismatch rejection: focused 1/1 green
- producer -> manifest decoder -> Dashboard API, all three outcomes: focused 1/1 green
- Dashboard TypeScript: `tsc --noEmit` green
- Dashboard outcome/cause wording and no-false-success UI: focused 1/1 green
- Dashboard API compaction decoder: focused 1/1 green

Closes #26038
Parent: #25430

### DIFF
--- a/dashboard/src/api/dashboard-turn-records.ts
+++ b/dashboard/src/api/dashboard-turn-records.ts
@@ -637,6 +637,36 @@ function decodeCompactionReinjectionObservation(
   }
 }
 
+function decodeCompactionOutcome(raw: unknown): KeeperCompactionOutcome | null | undefined {
+  if (raw === null) return null
+  const value = asString(raw)
+  switch (value) {
+    case 'checkpoint_committed':
+    case 'retry_without_checkpoint':
+    case 'lifecycle_cleanup_failed_without_checkpoint':
+      return value
+    case undefined:
+    default:
+      return undefined
+  }
+}
+
+function compactionOutcomeContractIsValid(
+  outcome: KeeperCompactionOutcome | null,
+  evidence: KeeperCompactionExactEvidence | null,
+  failureCause: string | null,
+): boolean {
+  switch (outcome) {
+    case null:
+      return evidence === null && failureCause === null
+    case 'checkpoint_committed':
+      return evidence !== null && failureCause === null
+    case 'retry_without_checkpoint':
+    case 'lifecycle_cleanup_failed_without_checkpoint':
+      return evidence === null && Boolean(failureCause?.trim())
+  }
+}
+
 function decodeKeeperCompactionSnapshot(raw: unknown): KeeperCompactionSnapshot | null {
   if (!isRecord(raw)) return null
   const id = asString(raw.id)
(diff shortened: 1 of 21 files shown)


jeong-sik:
Live blocker found against this head's intended behavior: atomic `Drop`/`Summarize` of a closed tool cycle changes both `ToolUse` and `ToolResult` counts, but `Keeper_compaction_evidence.validate` on `main` still requires each count to remain exactly equal. The first patched-runtime compaction reached the model and then failed closed with `invalid_structural_evidence tool_uses:31:7`.

The evidence contract needs to admit a reduction only when the before and after counts are each matched pairs (`tool_uses = tool_results`), never permit increases, and reject asymmetric/orphan transitions. Message accounting also needs to count replacing an N-message cycle by one summary as N-1 removed messages in addition to one summarized output.

After applying that paired-count/accounting correction in the isolated runtime, the same checkpoint compacted successfully: messages 78→34, ToolUse 31→3, ToolResult 31→3, checkpoint bytes 134,161→62,909; the Keeper then restarted from `compaction_count=1` and completed turn 31 with `POST-COMPACTION-RESTART-OK`.

Please include the evidence-layer change + focused `keeper_compaction_evidence` regressions before treating #26038 as fixed. I have a validated local patch/trace if useful; I did not push a duplicate PR because this Draft already owns the issue.

PR #26045, opened 2026-07-27
fix(schedule): retain cancellation WAL until projection

### LINKED ISSUE #26043: Schedule cancellation recovery regression returns awaiting_ack on retry
## Symptom

`test_schedule_consumer_dispatch.exe` is red on current `main` (`c8334ac9f2`) before any #26042 changes:

- failing case: `cancelled occurrence recovery does not enqueue again`
- expected retry receipt `occurrence_status=already_cancelled`
- received `occurrence_status=awaiting_ack`
- 13/14 other baseline cases pass

The test first cancels the retained schedule occurrence, projects the transition outbox, recovers the running schedule, then retries the same occurrence. The retry no longer sees terminal cancellation evidence.

## Reproduction

```sh
scripts/dune-local.sh build test/test_schedule_consumer_dispatch.exe
scripts/dune-local.sh exec ./test/test_schedule_consumer_dispatch.exe
```

Observed 2026-07-28 05:36 KST in the clean root checkout on `main`, independently reproduced in the #26042 worktree (where the new due-wake/proactive-policy case itself passes).

## Expected

The same occurrence retry must return `already_cancelled`, avoid a second enqueue, and project exact terminal cancellation evidence.

## Tried

- Re-ran in the feature worktree: identical failure.
- Rebuilt and reran from clean root `main`: identical failure, confirming it predates and is independent of #26042.

jeong-sik:
## What changed

- keep the settlement WAL as the authoritative commit record until reaction-ledger projection succeeds
- avoid checkpointing a transition into the pending-only v7 snapshot, which cannot retain leases or transition outboxes
- perform ledger append, projected-state checkpoint, and WAL retirement under the owner lock

## Root cause

`commit_settlement_transition_unlocked_with` saved the post-cancellation state into a snapshot that serializes only `schema`, `revision`, and `pending`, then compacted the settlement WAL. That discarded the only durable transition outbox before `Keeper_reaction_ledger` could project terminal cancellation evidence. Schedule recovery therefore retried the occurrence as `awaiting_ack` instead of `already_cancelled`.

The fix does not add claim/lease state back to the snapshot. The pre-transition pending snapshot remains in place while the WAL carries the committed transition; projection then writes the final pending-only snapshot and retires the WAL.

## User impact

Cancelled scheduled occurrences remain terminal across recovery and are not enqueued again.

## Validation

- `scripts/dune-local.sh build test/test_schedule_consumer_dispatch.exe test/test_keeper_reaction_ledger.exe test/test_keeper_exact_execution_lease_guard.exe`
- `ocamlformat --check lib/keeper_runtime/keeper_event_queue_persistence.ml`
- `_build/default/test/test_schedule_consumer_dispatch.exe` — 14/14 passed

The two legacy lease suites remain red on current `main` because #25969 removed durable lease restoration while those tests still call claim/settle across persistence boundaries; that existing cleanup is tracked in #25981 and #26014 and is not caused by this one-file change.

Closes #26043
Related: #26014, #25981

### DIFF
--- a/lib/keeper_runtime/keeper_event_queue_persistence.ml
+++ b/lib/keeper_runtime/keeper_event_queue_persistence.ml
@@ -1608,6 +1608,10 @@ let prepare_registration_result
     ()
 ;;
 
+let mark_transition_projected_result state ~transition_id =
+  State.mark_transition_projected ~transition_id state
+;;
+
 let project_transition_outbox_result
       ~append_before_retire
       ~base_path
@@ -1623,9 +1627,9 @@ let project_transition_outbox_result
       | [ entry ] ->
         let* () = append_before_retire entry in
         let* projected =
-          State.mark_transition_projected
-            ~transition_id:entry.receipt.transition_id
+          mark_transition_projected_result
             state
+            ~transition_id:entry.receipt.transition_id
         in
         let* projected = bump_revision projected in
         let* () = save_state_unlocked owner projected in
--- a/lib/keeper_runtime/keeper_event_queue_persistence.ml
+++ b/lib/keeper_runtime/keeper_event_queue_persistence.ml
@@ -506,22 +506,13 @@ let replay_settlement_wal_unlocked owner state =
                ~surface:"settlement WAL"
                detail)
         | Ok replayed ->
-         (match bump_revision replayed with
-          | Error _ as error -> error
-          | Ok replayed ->
-            (match save_state_unlocked owner replayed with
-             | Ok () ->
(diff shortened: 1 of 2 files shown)


jeong-sik:
CI follow-up: the failed projection-boundary check expected the named `mark_transition_projected_result` boundary inside `project_transition_outbox_result`. The cancellation fix intentionally keeps append, mark, save, and WAL compact under one owner lock; commit `469fbf1ac2` restores the named in-lock helper without reintroducing a second lock/transaction. Local boundary checker is green and the 14/14 schedule consumer suite remains green.

PR #26055, opened 2026-07-27
fix(task): route completion through Keeper verification

### LINKED ISSUE #26050: fix(task): route Keeper completion through ordinary verification
## Symptom

`keeper_task_done` sends `action=done`. `Task.Tool.handle_transition` invokes the configured `cross_verifier` LLM and a pass transitions an owned `in_progress` task directly to `done`. The task never enters `awaiting_verification`, so an ordinary proactive Keeper cannot claim and decide the verification obligation.

## Reproduction

1. Create and claim a normal task with evidence-bearing completion notes.
2. Start the task.
3. Call `keeper_task_done` with `result` and `evidence_refs`.
4. Observe `configured LLM runtime=cross_verifier_lane` and `in_progress -> done`, with no verification claim.

## Root cause

`keeper_task_done` maps to `Done_action`. The Task tool invokes the internal completion reviewer for done/approve/reject, and `Workspace_task_lifecycle` accepts direct done after an LLM pass. #26026 only enforces assignment after a task has already reached `AwaitingVerification`.

## Scope

- Route producer completion to `submit_for_verification`.
- Reject direct producer `Done_action` from claimed/in-progress state.
- Preserve phase-bound anti-self and anti-steal checks.
- Let the assigned ordinary Keeper's explicit approve/reject action decide the result without an internal cross-verifier veto.
- Reuse existing verification request, receipt, and post-commit paths. No new store, lease, settlement, durable queue, or facade.

## Attempts

Read-only source audit on `origin/main` `4d7ee9e244` traced the bypass through `keeper_tool_task_runtime`, `tool_task`, `Anti_rationalization`, and `workspace_task_lifecycle`.

jeong-sik:
## Summary

- route `keeper_task_done` completion notes and evidence into `submit_for_verification`
- reject direct producer `Done_action` from claimed/in-progress state with a typed remediation
- make the phase-assigned ordinary Keeper's explicit approve/reject action authoritative
- remove the internal cross-verifier invocation and action rewrite from Task transitions
- preserve existing anti-self, assignment winner, request receipt, terminal non-overwrite, and #26056 evidence-snapshot boundaries

## Root cause

`keeper_task_done` emitted `action=done`. The Task tool invoked the configured `cross_verifier` LLM and passed its verdict to the lifecycle, which transitioned an owned task directly from `in_progress` to `done`. The `AwaitingVerification.phase` rules added by #26026 never ran because the task never entered that state.

## Behavior

Producer completion now becomes `AwaitingVerification`, preserving `result` as the handoff summary and preserving `evidence_refs`. A different ordinary Keeper races to claim the verification obligation through the existing Task phase; only the committed phase-assigned distinct verifier may explicitly approve or reject it. No new store, lease, settlement, durable queue, or verifier facade is introduced.

After rebasing onto #26056, submitted evidence keeps its strict `artifact:<producer-root-relative-path>` / `note:<text>` namespace and the persisted producer snapshot remains the single evidence SSOT.

Closes #26050

## Base/head proof

- base: `7098d2dbf751cd9681d0debb00500e7dbeba68cd` (#26056 merged main)
- head: `aaae523e715a85dc1804386236f987fa38169222`

## Validation

- `ocamlformat --check lib/task/tool_task_schemas.ml`
- focused build: `test_tool_task_coverage.exe`, `test_keeper_task_outcomes.exe`, `test_workspace_task_lifecycle.exe`, `test_verification.exe`, `test_tools_coverage.exe`
- `test_workspace_task_lifecycle.exe`: pass
- `test_keeper_task_outcomes.exe`: 7/7
- `test_tools_coverage.exe`: 50/50
- `test_tool_task_coverage.exe`: 91/91
- `test_verification.exe test '^storage$' '9-16,18-20'`: 11/11 snapshot/storage/claim-projection cases

## Validation limitation

The full `test_verification.exe` reaches 46/47, but storage case 17 (`submitted evidence rejects symlink...`) consistently fails in test teardown only: `rm_rf` raises `Unix.ENOTEMPTY` for the temporary `.masc/playground/docker/executor` directory. The isolated case reproduces the same cleanup exception without an assertion failure. This PR does not modify the merged #26056 cleanup helper.

The unfiltered `test_workspace.exe` still has seven fixture failures because legacy test Keeper metadata violates canonical agent identity. This is tracked separately in #26054; the affected `board_admin` and `task_identity` fixtures are outside this PR.


jeong-sik:
P1 liveness finding from the live full-cycle audit (`~/me/.masc`, 2026-07-28 07:31–07:37 KST): the current completion LLM veto incorrectly rejected a valid verifier approval and returned task-002 to `in_progress`; removing that veto is correct. However, this PR currently replaces it with a mandatory peer dependency for every `keeper_task_done`:

- `keeper_task_done` always maps to `submit_for_verification`.
- direct producer `Done_action` always returns `Verification_submission_required`.
- self-verification remains correctly forbidden.

Therefore a one-Keeper autonomous deployment, or a fleet with no available eligible peer, has no terminal path even after the work and evidence are complete. That conflicts with the liveness goal behind this audit: work should run first, and guards must not make ordinary progress impossible.

Please preserve a terminal path that does not require an implicit peer for every task. The smallest contracts I see are either:

1. direct owner completion by default, with ordinary verification only when explicitly requested through `submit_for_verification`; or
2. mandatory peer verification only when the task carries an explicit verification contract, while contract-less ordinary tasks complete directly.

If universal peer verification is intentionally retained, the PR needs a tested typed fallback for `eligible_verifier_count = 0`; otherwise this is a deterministic autonomous deadlock, not merely a UX trade-off.

jeong-sik:
P1 liveness follow-up is implemented in stacked Draft PR #26061. It connects the existing `task_requires_verification = contract.strict` SSOT to both the workspace FSM and `keeper_task_done`: default/advisory owner completion is terminal, while `strict=true` retains peer verification. Focused build and 11 targeted assertions pass.

PR #26063, opened 2026-07-27
fix(keeper): retry approved effects until completion

### LINKED ISSUE #26060: keeper: approved WebSearch can be skipped because Deferred is projected as last=ok
## Symptom

A recurring scheduled Keeper wake reached Auto Judge and was approved, but the actual WebSearch effect was never replayed. The next Keeper turn concluded that WebSearch had already executed because the session context rendered `last=WebSearch(ok)` for the earlier **Deferred** result.

## Reproduction / live evidence

Runtime: isolated MASC 0.21.2 process on port 8945, base `.runtime-base`, 2026-07-28 KST.

1. `audit-hourly-websearch` became due at `1785193318` (08:01:58 KST) and advanced its next due to `1785196918`.
2. Raw trace `turn-1785193318249-4a2d-000003.jsonl` records WebSearch query `OpenAI official product announcements July 27 2026` followed only by a Deferred Gate payload (`appr_e4ba7c91...`).
3. `audit-approvals/2026-07/27.jsonl` records the exact attempt settled and Auto Judge approved it at `1785193346.210044`.
4. Resume trace `turn-1785193346492-4a2d-000004.jsonl` contains no WebSearch tool call. It says the gate "already executed" and reports no delta.
5. `.masc/tool_calls/2026-07/27.jsonl` contains only the deferred WebSearch call for this query; there is no completed replay.
6. The resumed OAS snapshot renders `session:last_tool_name=WebSearch` and `session:last_tool_outcome=ok`, because Deferred projects to OAS `Ok`.

## Suspected cause

- `Keeper_tools_oas_bundle` automatically replays approved filesystem writes only (`Keeper_gate_replay.replay_approved_write`). `network_read` approvals still depend on the model re-emitting the exact call.
- `Masc_context_injector.make` sees OAS `Ok` and cannot distinguish MASC `Tool_result.Deferred`, so its temporal summary falsely implies success.
- The resolution prompt says the authorization belongs to the exact input but does not state or enforce that the original effect did not run.

## Expected

An approval owns the exact effect for all replayable Gate operations. On an approved `network_read`, runtime must consume the grant and execute the stored `web_search` / `web_fetch` input exactly once, then expose the typed replay result to the resumed Turn. The model must not be the authority for whether replay happens.

## Tried

- Verified schedule dispatch, Auto Judge exact attempt, resolution wake, raw OAS traces, tool-call ledger, and next-due advancement.
- Confirmed #26037 addresses stale resolution-wake replay after delivery, not replay of the approved operation itself.

```masc-triage
kind: defect
area: turn
impact: degrades
root: silent
must-do: false
```

jeong-sik:
Closes #26060.

## Problem

An approved `network_read` remained only a deferred WebSearch tool result in the OAS transcript. The resumed model could treat that stale result as success, while a restart could either issue a duplicate approval or lose the effect entirely if `grant_consumed` was persisted before the external result existed.

A live crash-boundary probe reproduced the loss window: Auto Judge resolved approval `appr_f913affa-cb47-4014-a2a3-4a9cb7295ea5`, the runtime persisted `grant_consumed`, then SIGINT cancelled the process before any replay completion was recorded.

## Change

- replay approved WebSearch/WebFetch from the durable exact Gate input before model dispatch
- let restarted originating turns find and reuse the same exact approved resolution instead of submitting a duplicate
- reserve a one-shot grant only in the current process while the effect runs
- commit durable `grant_consumed` only after a typed successful completion result
- leave the approval reusable after failure, cancellation, or process loss before completion
- apply the completion boundary consistently to network/in-process effects, Execute, and external filesystem writes
- put typed applied/failed replay evidence into system context so the resumed Turn cannot mistake Deferred for the external result

This deliberately provides retry-until-completion semantics. Arbitrary external effects cannot provide mathematical exactly-once behavior across a crash after the remote side effect but before its local completion receipt unless the provider supports idempotency keys.

## Validation

- `scripts/dune-local.sh build test/test_keeper_approval_queue.exe test/test_keeper_gate_replay.exe test/test_keeper_gate_effect_coverage.exe test/test_keeper_fs_edit_containment.exe bin/main_eio.exe`
- approval queue crash/retry cases: 2/2
- Gate replay: 10/10
- external-effect completion coverage: 9/9
- filesystem containment: 4/4
- `ocamlformat --check` on all changed OCaml files
- `git diff --check`

A fresh live crash/restart proof will be attached from the isolated candidate runtime.

### DIFF
--- a/lib/keeper/keeper_approval_queue.ml
+++ b/lib/keeper/keeper_approval_queue.ml
@@ -1635,6 +1635,25 @@ let approved_resolution_state ~base_path ~id =
     | Ok (Approved_delivery_unconsumed _) -> Ok Resolution_unconsumed)
 ;;
 
+let approved_resolution_matches
+      ~base_path
+      ~id
+      ~keeper_name
+      ~tool_name
+      ~input
+  =
+  with_pending_store_lock (fun () ->
+    match approved_delivery_unlocked ~base_path ~id with
+    | Error _ as error -> error
+    | Ok Approved_delivery_consumed -> Ok false
+    | Ok (Approved_delivery_unconsumed delivery) ->
+      let entry = delivery.entry in
+      Ok
+        (String.equal entry.keeper_name keeper_name
+         && String.equal entry.tool_name tool_name
+         && String.equal entry.input_hash (normalized_input_hash input)))
+;;
+
 let consume_approved_resolution
       ~base_path
       ~id
@@ -1694,7 +1713,7 @@ let consume_approved_resolution
     Ok Consumption_committed
 ;;
 
-let consume_matching_approved_resolution
+let find_matching_approved_resolution
       ~base_path
       ~keeper_name
       ~tool_name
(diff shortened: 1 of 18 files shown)


jeong-sik:
## Live crash/restart proof (2026-07-28)

Validated the completion-bound grant semantics on fresh shadow keeper `completion-probe-v2`, candidate build `3341b7e04b` (main base `61be6740e2` + this PR + #26005/#26061).

1. Hourly schedule dispatched query `OpenAI official newsroom clean crash retry 1785199790`.
2. Auto Judge approved `appr_40ea017f-1ed2-40ee-846d-6a5627a8bb79`; durable delivery still showed `grant_consumed=false`.
3. Process was SIGSTOP/SIGKILLed after `gate_allowed` but before a WebSearch result.
4. Same candidate restarted against the same durable base.
5. Replay produced one typed success at `2026-07-28T00:53:22Z`:
   `{"status":"ok","result":{"query":"OpenAI official newsroom clean crash retry 1785199790","engine":"searxng","result_count":1,...}}`
6. Only then did the delivery become `grant_consumed=true`.

Postconditions:
- applied replay count for the approval: exactly `1`
- pending approval for the exact query: `0`
- durable delivery for the exact query: exactly `1`, consumed
- duplicate approval for the exact query: none

This proves restart liveness and no duplicate grant in the observed crash boundary. The contract remains at-least-once for the remote side effect: arbitrary providers still need idempotency keys for mathematical exactly-once behavior across a crash after remote success but before local completion commit.

jeong-sik:
MVP 경계에 맞춰 이 PR은 닫습니다.\n\napproval 결과를 실제 effect 수행과 연결하는 문제는 유효하지만, 이 PR은 grant reservation, replay completion, restart consumption semantics를 Keeper 도구 전역에 추가합니다. 이는 현재 명시한 no new lease/store/settlement/nonce/replay facade 및 Full Cycle 우선 원칙과 충돌합니다.\n\n지금은 승인된 현재 Turn이 실제 effect를 수행하고 실패 원인을 공개 Surface에 남기는 최소 경로를 우선합니다. crash exactly-once 계층은 실제 반복 장애 증거가 생긴 뒤 별도 설계합니다.

PR #26243, opened 2026-07-29
refactor(keeper): require exact turn decision

jeong-sik:
## Scope

Follow-up to merged #26236. Remove the past-caller compatibility path that allowed a Keeper turn to omit the scheduler decision and locally recompute a weaker value.

## Feature trace

- Origin: `Keeper_heartbeat_loop_scheduling` creates the exact `keeper_cycle_decision`.
- Entrypoint: `Keeper_heartbeat_loop` passes it through `Keeper_heartbeat_loop_cycle.run_keeper_cycle`.
- Consumer: `Keeper_unified_turn.run_keeper_cycle` requires it and `Keeper_unified_prompt.build_prompt` renders the exact wake reasons.
- Blast radius: the two current function contracts and prompt-focused tests only.

Every production caller already supplied the decision. The optional arguments, fallback recompute, and test that blessed the missing-decision behavior are removed. No migration, compatibility reader, fallback, new Gate, or facade is added.

## Verification at `19b45a2d2379ce0d3f8b5a6ddf96db2a69314c77`

- `ocamlformat --check` for every changed `.ml`/`.mli` file — PASS
- `ocamlc -stop-after parsing` for every changed `.ml`/`.mli` file — PASS
- `git diff --check` — PASS
- Local Dune intentionally not run; exact-head CI is authoritative.

### DIFF
--- a/lib/keeper/keeper_unified_prompt.ml
+++ b/lib/keeper/keeper_unified_prompt.ml
@@ -470,7 +470,7 @@ let autonomous_trigger_lines
 
 let build_prompt ~(meta : Keeper_meta_contract.keeper_meta) ~(base_path : string)
     ?(profile_defaults : Keeper_types_profile.keeper_profile_defaults option)
-    ?(turn_decision : Keeper_world_observation.keeper_cycle_decision option)
+    ~(turn_decision : Keeper_world_observation.keeper_cycle_decision)
     ?(current_task : Masc_domain.task option)
     ?(active_goal_summaries : (string * string) list option)
     ~(observation : Keeper_world_observation.world_observation)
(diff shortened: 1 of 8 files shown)


jeong-sik:
병합된 #26236의 실제 호출 경로를 다시 확인하다 P0 호환 경로를 분리해 고쳤어용. production caller는 이미 exact `turn_decision`을 항상 전달했는데도 `run_keeper_cycle`과 prompt가 optional + local recompute를 유지하고 있었어용. 현재 head에서 둘 다 required로 hard-cut했고, Feature의 wake SSOT와 blast radius는 그대로예용.

PR #26363, opened 2026-07-29
fix(keeper): measure context composition at the projected request boundary

jeong-sik:
## What changed

- capture Keeper context composition after OAS `model_input_projection`, at the final MASC-owned request-component boundary
- attribute Memory OS recall, tool schemas, thinking/reasoning, current input, projected Gate evidence, and the OAS system-context prefix/joiners
- stop reconstructing a misleading partial composition after the response; missing boundary observation is now explicit
- show request-boundary vs legacy partial snapshots honestly in the Keeper dashboard
- include structured tool-result JSON in canonical content-byte telemetry

## Why

The live analyst turn reported 115,071 provider input tokens while the existing composition accounted for only 166,439 of 347,637 observed component bytes (47.9%). Memory OS recall, tool schemas, and thinking were omitted, so the dashboard's `exact bytes represented` claim was false.

## Validation

- `npx tsc --noEmit --pretty`
- focused Vitest: 2 files, 42 tests passed
- `ocamlformat --check` on all changed OCaml files
- `git diff --check`
- focused OCaml build/tests pending: another session is holding the machine-wide Dune build with `test_keeper_post_turn_wirein_order.exe`; this Draft PR's CI is the independent build authority


jeong-sik:
대상 SHA: 90b4ddd3ffbfde105d27a19de2f527cc9959fc3b

- P0: 없었어용.
- P1: projection 전 인덱스와 길이 차감으로 origin을 추정하면서 실패해도 request_boundary_observed=true를 기록해 정확성을 거짓으로 만들어요.
- P2: SDK 공개 타입은 message 순서 보존을 보장하지 않고, tool set dedupe·합성 context와의 동일성 회귀도 없어요.
- P3: 경계 관측 목표는 유효하지만 OAS가 prepared component의 typed origin/evidence를 제공해야 해용.

병합 가능성: 없었어용. 하드코딩된 인덱스·휴리스틱 차감 위에 exact 관측을 세우는 방향이라 닫았어용.

PR #26473, opened 2026-07-30
fix(build): align current keeper contracts

jeong-sik:
## Summary
- use current event source-incarnation field for replay identity
- construct Librarian inputs with TurnRef
- qualify current queue selection and Keeper identity types

## Verification
- ocamlformat --check: pass
- git diff --check: pass
- focused Dune intentionally not run; CI is build authority


jeong-sik:
Superseded by #26472, which contains the same current-contract compile repairs plus the remaining HITL summary worker error.

PR #26523, opened 2026-07-31
fix(workspace): add executable filter guard in claim_next_r (#26487, #26485)

jeong-sik:
## Summary
Fixes issue #26487 (unexecutable task hot loop) and #26485 by adding an explicit `is_executable_task` guard in `Workspace_task_schedule.claim_next_r`.

Tasks that carry a non-empty `do_not_reclaim_reason` (flagged during unexecutable capability/quarantine states) are now filtered out prior to candidate allocation in `claim_next_r`. This prevents keepers from repeatedly claiming unexecutable backlog items in an autonomous hot-loop.

## Verification
- `./scripts/dune-local.sh build lib/workspace` completed with zero errors.

### DIFF
--- a/lib/workspace/workspace_task_schedule.ml
+++ b/lib/workspace/workspace_task_schedule.ml
@@ -298,10 +298,17 @@ let claim_next_r
              AwaitingVerification tasks into the keeper claim pool. *)
           | Workspace_task_lifecycle.Held_pending_verdict _ -> false
         in
+        let is_executable_task (t : Masc_domain.task) =
+          (* #26487 / #26485 guard: exclude tasks flagged with explicit do_not_reclaim_reason *)
+          match t.do_not_reclaim_reason with
+          | Some reason when String.length (String.trim reason) > 0 -> false
+          | _ -> true
+        in
         let unclaimed =
           sorted
           |> List.filter Masc_domain.task_claim_next_action_is_claimable
           |> List.filter resolves_claimable
+          |> List.filter is_executable_task
           (* [hard_filter] is a hard exclusion (e.g. self-author ownership): unlike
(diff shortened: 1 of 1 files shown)


jeong-sik:
BLOCKER (P1) — this guard is not a valid execution-capability contract.

`claim_next_r` now rejects every task with a non-empty `do_not_reclaim_reason`. That field is narrative handoff/reclaim text, not a typed “unexecutable” state. Existing contract/tests require the opposite:
- `test_release_hard_stop_todo_stays_claimable` (test/test_workspace_coverage.ml:615)
- `test_claim_next_ignores_non_authoritative_auto_cycle_text` (:829)
- `test_claim_next_ignores_routing_handoff_text` (:864)
- `test_claim_next_does_not_deprioritize_narrative_text` (:900)

This regresses RFC-0323 semantics and can strand ordinary Todo tasks merely because they carry handoff text. It also uses string presence as the SSOT.

Please move the decision to the Keeper auto-claim caller, where typed Keeper/sandbox execution context exists, and filter with an explicit capability/repository predicate (or introduce a producer-owned typed execution-blocked field). Keep generic Workspace claim semantics unchanged and add tests for matching/unmatching capability plus narrative `do_not_reclaim_reason`.

jeong-sik:
Closing after unchanged-head re-review (`4908e946`). The proposed guard treats narrative `do_not_reclaim_reason` text as an execution-capability SSOT and changes generic Workspace claim semantics. It also does not address #26485, whose failure boundary is queued-chat receipt settlement after Keeper completion. Root work remains in #26487: define a producer-owned typed task requirement, compare it with Keeper/sandbox capability at the auto-claim boundary, and preserve narrative reclaim text as non-authoritative.

PR #28310, opened 2026-08-12
ci(keeper): replay real-world gate on trusted dispatch

jeong-sik:
## Summary

- let trusted `workflow_dispatch` build and package the exact Keeper runtime revision, generate and upload release evidence, and run V01-V15 before the protected RW01-RW16 acceptance job
- require the dispatch path to pass Build, regression CI, protected five-Keeper acceptance, and the final Keeper Real-World Gate instead of silently passing after a rapid merge
- remove the retired `audit-integrity` dashboard navigation telemetry entry found by the full dispatch gate

## Verification

- `git diff --check`
- CI YAML parsed successfully with Ruby YAML
- `scripts/check-dashboard-nav-event-parity.sh --check`
- run 31562427151: Dashboard, compile, focused behavior suites, and contract harness passed; exposed missing dispatch baseline fetch
- run 31563513812: Dashboard and compile passed; baseline fetch worked and exposed six `Format.fprintf` effects in registered-pure `Turn_fsm`
- the equivalent `Turn_fsm` purity repair landed independently on main in #28301 and was accepted while merging current main
- run 31564519071 was cancelled at user request because final builds will be run by the user/CI

Current exact head: `76aac0ab34`. No local Dune build or test was run.


.github/workflows/ci.yml
```diff
@@ -1432,8 +1458,13 @@ jobs:
       ${{
         always() &&
         !cancelled() &&
-        github.event_name == 'pull_request' &&
-        github.event.pull_request.head.repo.full_name == github.repository &&
+        (
+          github.event_name == 'workflow_dispatch' ||
```
jeong-sik, line 1462: **P1 — Restrict the secret-bearing dispatch to the trusted ref.** `workflow_dispatch` can be launched against an arbitrary branch/tag selected in the Actions UI, but this condition admits every dispatch into the `keeper-realworld-acceptance` environment and then checks out/runs that ref's `scripts/ci/run-keeper-realworld-acceptance.sh` with `ZAI_API_KEY_SB`. The live environment has no deployment branch policy (`deployment_branch_policy: null`) and permits self-review, so the word “trusted” is not enforced by GitHub. Fail closed unless `github.ref == 'refs/heads/main'` (and preferably record/verify that ref in the manifest/gate), or configure an equivalent protected-branch deployment policy before exposing the secret.

jeong-sik: 맞습니다. #28310이 이 리뷰보다 먼저 병합되어 후속 Draft #28314로 수리했습니다. Commit `8d33a1c351`은 secret-bearing acceptance job을 `workflow_dispatch && github.ref == 'refs/heads/main'`으로 제한하고, downstream gate에서도 `EVENT_REF`를 다시 검증합니다. `scripts/ci/check-yaml-syntax.py` 38개 통과, exact-head CI는 진행 예정입니다. #28314가 main에 반영될 때까지 이 스레드는 열어 두겠습니다.

PR #28479, opened 2026-08-12
perf(http): answer 304 when a polled JSON body has not changed

jeong-sik:
## The measurement first

An ETag only pays if repeated polls produce byte-identical bodies. Several of
these projections stamp a wall-clock field at compute time, so that was worth
checking before building anything. Six polls per route, 0.4 s apart, against
the live server:

| route | distinct bodies | bytes |
|---|---|---|
| `dashboard/board` | 1 | 840 350 |
| `board` | 1 | 474 537 |
| `dashboard/telemetry` | 1 | 140 475 |
| `activity/events` | 1 | 122 441 |
| `dashboard/scheduled-automation` | 1 | 95 914 |
| `dashboard/goals` | 1 | 65 209 |
| `keepers/composite` | 1 | 48 349 |
| `dashboard/gate` | 1 | 44 505 |
| `dashboard/tool-quality` | 1 | 22 546 |
| `dashboard/harness-health` | 1 | 7 549 |
| `agent-activity` | 1 | 2 164 |
| `dashboard/bootstrap` | **4** | 261 552 |

11 of 12 repeat exactly — the wall-clock fields are inside the cached value, so
they are stable for the TTL. That is **1.86 MB per full poll cycle** currently
compressed, written to the socket, and parsed by the client to arrive at what
the client already had. `bootstrap` varies on `served_at` and is simply never a
match; it costs one hash and keeps behaving as it does today.

## Change

`Response.json` offers a validator on a `` `OK `` answer to `GET`/`HEAD`: a weak
`ETag` over the uncompressed body plus `Cache-Control: no-cache`. A matching
`If-None-Match` gets `304` — no body, no compression, no socket write.

Three properties make this safe to apply at the shared responder rather than
route by route:

- **A client that sends no `If-None-Match` sees no change.** It has not claimed
  to hold a copy, so it always receives the body. Nothing that works today
  starts behaving differently.
- **Errors and unsafe methods carry no tag.** Revalidating a failure, or
  telling a client its copy still stands after it mutated something, are both
  wrong. `` `Bad_request ``, `` `Not_found ``, `` `Internal_server_error ``,
  `` `Accepted ``, and `POST`/`PUT`/`DELETE` are all `Untagged`.
- **The tag is body-derived**, so it changes if and only if the body changes.
  A stale 304 is not reachable.

`Cache-Control: no-cache` means store-and-always-revalidate. Without it a cache
is free to invent a freshness lifetime for a 200 that carries a validator and
serve keeper state that has moved on.

The tag is **weak** because one JSON payload goes out gzipped,
zstd-compressed, or plain depending on `Accept-Encoding`, and those are
encodings of one payload rather than three payloads. `If-None-Match` compares
weakly, which is the comparison this wants.

## Shape

The decision is a pure function over a closed variant rather than booleans
threaded through the responder:

```ocaml
type json_conditional =
  | Untagged
  | Tagged of string
  | Not_modified of string

val json_conditional :
  status:Httpun.Status.t -> meth:Httpun.Method.t ->
  if_none_match:string option -> body:string -> json_conditional
```

Every combination of method, status, and header is then decidable in a test
without constructing a `Httpun.Reqd.t`, which is why the rule is exercised
directly rather than through a live server: what can go wrong here is the
decision, not the socket write.

`server_routes_http_pages` carried its own copy of the tag policy for the HTML
route (`Digest.string |> to_hex`, truncated to 12). Both now use the one
definition, in the responder that emits it.

## Tests

Six cases in `test_http_server_eio`, `response` group, 37/37 green:

- matching `If-None-Match` is 304
- a tag from an older body does not suppress the new one
- absent `If-None-Match` sends the body
- `` `Bad_request ``/`` `Not_found ``/`` `Internal_server_error ``/`` `Accepted `` carry no validator
- `POST`/`PUT`/`DELETE` carry no validator
- the tag is weak, differs on a different body, and repeats on the same body

Mutation — dropping the `` `OK `` check so errors get tagged:

```
[FAIL]  response  5  error statuses carry no validator
```

Regression sweep, all green: `test_http_negotiation`, `test_compression`,
`dashboard-http-behavior-contracts`, `test_mcp_h1_h2_admission_parity`,
`test_http_pages_asset_read` (6/6, still recomputing the HTML tag through the
now-shared constant).

`test_http_server_eio` compiled in CI but ran in neither focused group; added.

```
[ci-test-targets] OK - 231 CI targets, all declared in Dune
[ci-test-targets] OK - 630 suites unwired, at baseline
```

Baseline 631 → 630, for the suite wired here. (These numbers moved after a
rebase onto current `main`: the branch first carried 660 → 658, and `main` has
since wired many more suites and dropped its own baseline to 631. The merge
conflict was on exactly that line; resolved to `main`'s value, then lowered by
one to hold this suite's gain.)

## What this does not yet save, and why it is not in this PR

The dashboard's own fetch helper defaults to `cache: 'no-store'`
(`dashboard/src/api/core.ts:312,344,381`, pinned by `core.test.ts:170`).
A `no-store` request is never stored and never carries `If-None-Match`, so the
browser dashboard will not produce a 304 until that default changes.

That flip is deliberately left as a separate decision. `no-store` → `no-cache`
preserves freshness exactly — both revalidate before every use — but it also
starts writing dashboard responses to the browser's disk cache, and those
payloads include `secret_projection`. Inspected: it carries status, counts,
paths, and env *names*, no secret values. Low sensitivity, but it is a
security-posture change on a credential-adjacent surface and should be decided
on its own, not as a side effect of a performance PR.

Meanwhile this is not inert: any conditional-request client benefits today, and
per-call `cache: 'no-cache'` already works through the existing
`init.cache ?? 'no-store'` for a caller that wants it on a specific high-volume
read.

RFC-WAIVED: HTTP response header emission and its tests; no credential,
identity, operator control-plane, sandbox, hook, or workflow surface is
modified. The credential-adjacent client-cache change is explicitly excluded
above.


jeong-sik:
head `6d3e5afc` 는 main 머지뿐이라 재리뷰 없이 판정 유지해요 (`1d185f54` 리뷰 그대로).

상속 red 분리도 맞는데 한 걸음 더 좁힐 수 있어요: 인용한 4건 중 2건은 방금 머지한 main 이 이미 고쳤고(current-owner → #28496, `.DS_Store` → #28486), 잔여 2건(codex app-server / response acknowledgement)은 **#28504 잔여분**이에요. #28504 병합 후 재실행이면 이 브랜치도 green 예상이에용.

Issue #26190, opened 2026-07-29
keeper: add producer-owned Gate replay for voice and lifecycle effects


jeong-sik:
## Symptom

Approved external-effect operations without a host replay continuation cannot safely complete from the durable Gate delivery. Re-sending their exact approved input through the next model turn recreates the cumulative n,n+1 request-growth path, especially for large voice/content payloads.

Affected producer operations currently include:

- `keeper_voice_speak`
- `keeper_voice_listen`
- `masc_runtime_ollama_probe`
- `masc_keeper_clear`
- `masc_keeper_sandbox_start`
- `masc_keeper_sandbox_stop`
- `masc_keeper_down`

## Reproduction

1. Set Gate mode to Manual.
2. Submit one of the operations above with a large exact input.
3. Approve the pending request.
4. Deliver the HITL wake.
5. Observe that the operation is outside `Keeper_gate_replay.replayable_of_operation`.

The current repair branch now fails this state closed as typed `unsupported_operation`, keeps the authorization unconsumed, and excludes the exact input from the provider request. That prevents growth but intentionally leaves the effect operator-blocked.

## Suspected cause

The remaining producers need their own strict inverse decoder and host continuation. Voice and Keeper lifecycle dispatch also require Eio/runtime resources not currently threaded into `Keeper_gate_replay.replay_approved_effect`; a generic string dispatcher would widen authority and is not acceptable.

## Tried

- Added fail-closed typed repair for unsupported approved operations.
- Added a 512 KiB regression proving the exact input is not copied into the next model request and the one-shot authorization is not consumed.
- Kept the durable Gate entry as the exact input SSOT.

## Acceptance criteria

- Each producer owns a strict decoder for its exact durable Gate input.
- Replay receives the same Eio/runtime resources as the original producer path.
- The one-shot grant is consumed only immediately before the host effect.
- Full results are content-addressed; queue/model evidence stays bounded.
- Retry after blob/journal failure repairs persistence without repeating the effect.
- Restart ambiguity remains typed operator repair, never model resubmission.

```masc-triage
kind: gap
area: tools
impact: degrades
root: 
must-do: false
```

PR #28658, opened 2026-08-13
fix(runtime): align Ollama probe effect authority

### LINKED ISSUE #28652: [P1] runtime Ollama probe lets Worker invoke an external mutating probe as read-only
## Symptom

Merged PR #27267 correctly removed `masc_runtime_ollama_probe` from the autonomous Keeper model projection, but the retained public MCP execution boundary still classifies it as a Worker-readable, read-only, idempotent tool.

A Worker token can therefore direct repeated native Ollama requests, including `POST /api/generate`, at caller supplied `server_url` without an external-effect authorizer on the direct MCP path.

## Current reproduction

Current `origin/main`: `73b88a91cd21a6750ae55265c3d3814a370ddc54`

1. `lib/tool/tool_catalog.ml` registers `masc_runtime_ollama_probe` as `read_state_tool`, whose permission is `CanReadState`.
2. `Worker` owns `CanReadState`.
3. `lib/mcp_server_eio_execute.ml` authorizes the tool through that catalog entry, then constructs `Tool_local_runtime_core.context` with `authorize_external_effect = None`.
4. `lib/tool_local_runtime.ml` treats `None` as immediate execution.
5. The ToolSpec registration marks both local-runtime operations `is_read_only=true` and `is_idempotent=true`.
6. The probe accepts caller supplied `server_url`, performs `GET /api/ps`, and by default performs one to four `POST /api/generate` calls. It can load a model and changes warm/cache state.

## Expected

- The operator-only projection and execution authority describe the same boundary.
- Only `CanAdmin` may invoke the native operator probe through MCP. The metadata-only `/api/v1/dashboard/runtime-probe` read remains `CanReadState` and must not borrow the native probe tool identity.
- Runtime metadata declares the probe mutating and non-idempotent.
- The sibling `masc_runtime_verify` remains a Worker-readable read-only/idempotent observation.
- Tests pin permission and execution-policy parity for both operations.

## Duplicate grouping

Searched all open/closed issues for `masc_runtime_ollama_probe`, SSRF, arbitrary URL, host allowlist, and external-effect/read-state terms.

- #26190 is related but not a duplicate: it tracks durable Gate replay for approved producer effects, not direct MCP permission or execution metadata.
- #27358 is related provenance from #27267, but tracks unused Keeper schemas and measurement uncertainty.

## Suggested root fix

Use the closed `Tool_schemas_local_runtime.operation` variant to assign per-operation ToolSpec execution policy, and change the catalog-owned probe permission to `CanAdmin`. Add auth and metadata regression tests.

jeong-sik:
Closes #28652

## Summary

The oldest merged PR in the current seven-day audit window, #27267, correctly withheld the native Ollama timing probe from the autonomous Keeper model but left its direct MCP execution policy classified as Worker-readable, read-only, and idempotent. The first revision of this PR exposed the same inconsistency in sibling `masc_runtime_verify`: it issues real chat completions while claiming read-only/retryable semantics.

This change introduces one closed per-operation policy SSOT for both local-runtime diagnostics and projects it into all three consumers:

- `Tool_catalog`: both completion-generating diagnostics require `CanAdmin` and are mutating/non-idempotent.
- `Tool_spec`: execution hints come from the same typed operation policy instead of a shared hard-coded value.
- Keeper descriptor: both diagnostics are operator-only and non-retryable.

`masc_runtime_verify` now also passes through the Keeper external-effect authorizer. An explicit `runtime_pool` that matches nothing fails closed instead of broadening the completion probe to every discovered endpoint.

The metadata-only dashboard route remains a separate read path: `/api/v1/dashboard/runtime-probe` owns explicit `CanReadState` route authority and does not borrow either effectful tool identity.

## Blast radius and parity

| Surface | Before | After |
| --- | --- | --- |
| Worker direct MCP call to native Ollama probe | Allowed | Denied |
| Worker direct MCP call to completion-backed Verify | Allowed | Denied |
| Admin direct MCP calls | Allowed | Allowed |
| Keeper model projection | Ollama withheld; Verify visible | Both withheld |
| Keeper external-effect Gate | Ollama only | Both diagnostics |
| Unknown explicit Verify pool | Probed every endpoint | Fails closed |
| Dashboard metadata runtime probe | Borrowed native tool authority | Own `CanReadState` authority |

The Admin-only native diagnostic still accepts an explicit `server_url`; this PR narrows caller authority and fixes policy truth without changing its operator contract. Benchmark scripts remain usable with an Admin token when authentication is enabled; the performance documentation now says that explicitly.

## Verification

Exact head `9e1e6cc03e25b93981be5d295335f43759e5c049`:

- local runtime policy/probe suite — 22/22
- Keeper descriptor registry integrity — 45/45
- Gate effect coverage — 8/8
- whole-repo `scripts/dune-local.sh build @all` — pass
- `git diff --check` — pass
- `ocamlformat --check` on every changed OCaml file — pass

Repository-wide `@fmt` remains independently red on pre-existing Dune stanza formatting outside this diff; no corrected unrelated files are included.

The separate `test_mcp_server_eio_tool_profile_capability` hidden-deny corpus sentinel fails identically on clean current `origin/main` after every enumerated admission/advertisement comparison passes. It remains grouped under #28156 rather than duplicated here.

## Issue grouping

- #28652 — P1 root defect fixed by this PR, including the Verify sibling found in follow-up review
- #26190 — related Gate replay work, not a duplicate
- #27358 — related #27267 provenance/measurement issue, not a duplicate
- #28156 — pre-existing empty hidden-deny test corpus, not caused by this change
- #28661 / #28663 — independent v0.22.0 release-train blocker and repair; this PR remains Draft until exact-head CI reports against the current head

## Current-main rebase

- exact base: `addf6e4f5c0dd3b1d96bc2a2831d9d57c5ba655c`
- exact head: `434607d2aa3f16ed9ad09a974cdf54536a51ab42`
- local proof: `dune build @all`; runtime probe 22/22; descriptor registry 45/45; `git diff --check origin/main...HEAD`
- prior package-version Meta Guard blocker is removed by current main 0.22.0; fresh exact-head CI is authoritative for final readiness.

### DIFF
--- a/docs/PERFORMANCE-SLO.md
+++ b/docs/PERFORMANCE-SLO.md
@@ -14,7 +14,7 @@
 추가 지표:
 
 - `initialize + notifications/initialized` 세션 생성 비용은 별도 추적
-- raw local runtime 비용은 `masc_runtime_verify`로 MCP read-path와 분리해서 읽음
+- raw local runtime 비용은 Admin-only effectful diagnostic인 `masc_runtime_verify`로 MCP read-path와 분리해서 측정
 
 ### REST API
 - /api/v1/status P95 < 150ms
@@ -56,13 +56,16 @@ agent core RFC-agent core-020이 요구하는 consumer SLO를 추가하려면 ag
 - `benchmark.sh`는 `session`, `read`, `workspace collaboration`, `runtime`, `a2a`, `lock` lane을 분리하고 `avg/p50/p95/max`를 CSV로 남긴다.
 - `benchmark.sh`는 기본적으로 tool lane당 warmup 1회를 제외하고, 결과 CSV 옆에 metadata와 baseline diff를 같이 남긴다.
 - `runtime` lane 숫자는 MCP transport가 아니라 local runtime ceiling 영향을 크게 받는다.
+- 인증이 켜진 서버에서 `quick-bench.sh` 또는 `runtime` lane을 실행할 때는
+  `MASC_TOKEN`에 Admin credential이 필요하다. `masc_runtime_verify`는 endpoint마다
+  실제 chat completion을 호출하므로 read-only benchmark가 아니다.
 - `local64`는 target runtime profile 이름이지 achieved fact가 아니다. 실제 용량은 `masc_runtime_verify`의 `configured_capacity`, `healthy_runtime_count`로 확인한다.
 
 환경 변수:
 ```
 MASC_URL=http://127.0.0.1:8935/mcp
 MASC_AGENT=bench
-MASC_TOKEN=<optional>
+MASC_TOKEN=<Admin token required for quick-bench/runtime lane when auth is enabled>
 ```
 
 ## 경고 기준
--- a/lib/keeper/keeper_tool_descriptor.ml
+++ b/lib/keeper/keeper_tool_descriptor.ml
@@ -1371,13 +1371,23 @@ let masc_library_descriptors =
 let masc_local_runtime_descriptor
       (definition : Tool_schemas_local_runtime.definition) =
   let schema = definition.schema in
-  cluster_descriptor_with_schema_source
+  let keeper_model_projection =
+    match Tool_schemas_local_runtime.keeper_model_exposure definition.operation with
+    | Tool_schemas_local_runtime.Keeper_callable -> Internal_name
+    | Tool_schemas_local_runtime.Operator_diagnostic -> Operator_only
+  in
+  let execution_policy =
+    Tool_schemas_local_runtime.execution_policy definition.operation
+  in
+  let policy =
+    policy
+      ~readonly:execution_policy.read_only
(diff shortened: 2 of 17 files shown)


jeong-sik:
CI Meta Guards가 exact `434607d2aa`의 새 `handle_runtime_verify`를 telemetry-name heuristic으로 차단했습니다. 실제 경계는 authorization adapter이고 Gate authorizer가 decision receipt, outer MCP tool-call boundary가 duration/result telemetry를 소유하므로 그 소유권을 인접 `TEL-OK`로 명시했습니다. 새 exact head `f5cf5c359c5b49a89458fa4296cdfbd8ce27adea`; 로컬 telemetry coverage gate PASS, `@check` + runtime probe executable compile PASS, fresh CI 시작.

PR #28719, opened 2026-08-14
fix(dashboard): invalidate execution snapshots after task commits

jeong-sik:
## As-Is

- A committed batch task stayed absent from /api/v1/dashboard/execution until the 120s cache expired.
- Task mutations did not consistently invalidate every execution projection variant, and an older in-flight compute could republish stale mutable cache/body state.
- A stale execution SSE could arrive after a newer task commit and roll the live Dashboard back.

## To-Be

- The authoritative backlog and goal-link settlement boundaries notify one task-mutation hook after commit or rollback reconciliation.
- Default, force, actor, and full execution snapshots carry a process epoch plus monotonic publication generation.
- Cache invalidation fences in-flight computes, drops every execution cache variant, and emits an invalidation tombstone.
- Dashboard hydration rejects stale HTTP/SSE generations; reconnect epoch selection is established only by current-request HTTP/bootstrap, with per-epoch tombstone floors.
- Post-commit recovery, rollback, cache invalidation, and tombstone settlement are cancellation-protected without holding the publication mutex across SSE fanout.

## Evidence

- Live canary task-333/task-334 persisted immediately but the execution projection remained at 319 tasks for 120 seconds before reaching 321.
- Independent adversarial source review on the final diff: P0/P1/P2 none.

## Validation

- PASS: ocamlformat check for changed OCaml files
- PASS: ocamlc parse-only for changed OCaml files
- PASS: Dashboard tsc --noEmit using the existing checkout toolchain
- PASS: scoped Dashboard ESLint
- PASS: git diff --check
- PASS: OCaml structure, stringly-boundary, and silent-failure ratchets
- Not run: local Dune/build/test suites; exact-head CI owns those checks

## Remaining

- Draft and do-not-merge until exact-head CI is green.
- Deployment and live batch-task rerun are still required before Feature completion.

### DIFF
--- a/lib/server/server_dashboard_http_execution_surfaces.ml
+++ b/lib/server/server_dashboard_http_execution_surfaces.ml
@@ -176,17 +176,98 @@ let execution_cache : cached_surface =
 
 let execution_default_light_cache_key = "execution:default:light"
 let execution_default_light_http_body : string option Atomic.t = Atomic.make None
+let execution_publication_mu = Stdlib.Mutex.create ()
+let execution_publication_generation = ref 0
+let execution_publication_epoch =
+  (* NDT-OK: process-incarnation identity entropy only. *)
+  Uuidm.v4_gen (Random.State.make_self_init ()) () |> Uuidm.to_string
+;;
+
+let execution_publication_epoch_field = "execution_publication_epoch"
+let execution_publication_generation_field = "execution_publication_generation"
+let execution_invalidated_field = "execution_invalidated"
+
+let with_execution_publication_lock f =
+  Stdlib.Mutex.protect execution_publication_mu f
+;;
 
 let clear_execution_default_light_http_body () =
   Atomic.set execution_default_light_http_body None
 ;;
 
-let execution_surface_has_fresh_success () =
+let execution_surface_has_fresh_success_unlocked () =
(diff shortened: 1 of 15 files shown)


lib/server/server_dashboard_http_execution_surfaces.ml
```diff
@@ -252,22 +330,61 @@ let invalidate_execution_cache_with_hooks_for_testing
...
+       try
+         broadcast_cached_surface
+           ~event_type:"execution_snapshot"
+           (`Assoc
+               [ execution_publication_epoch_field, `String execution_publication_epoch
+               ; execution_publication_generation_field, `Int generation
+               ; execution_invalidated_field, `Bool true
+               ])
+       with
+       | exn ->
```
jeong-sik, line 357: 앞서 코멘트로만 남겼던 한 줄입니다. 클릭 적용 가능하도록 suggestion 으로 답니다.

이 파일의 다른 핸들러 4곳은 전부 catch-all 앞에 이 가드를 둡니다. 지금은 바깥 `Eio.Cancel.protect` 덕에 도달하지 않지만, 그건 열 줄 떨어진 래퍼에 기댄 안전입니다.

```suggestion
       | Eio.Cancel.Cancelled _ as e -> raise e
       | exn ->
```

PR #29244, opened 2026-08-21
docs(keeper): define Goal verifier real-world matrix

jeong-sik:
## Summary

- Extend the existing living Keeper feature-state matrix with a current Goal-verifier and real-world execution contract.
- Define eight explicit required cells and three non-substitutable approaches: hermetic, isolated real-world, and shadow duration.
- Freeze exact PR heads, catalog validation, historical campaign failure, live pending Goal observation, and a Chromium render receipt without promoting any missing axis to green.

## Product impact

The matrix states exactly what must be built and proven before Keeper/Goal work can be called complete. It prevents Task verification, catalog validation, historical runs, fixtures, or screenshots from substituting for independent Goal verification in one causal run.

## Evidence

- HTML parser — PASS.
- JSON parser — PASS.
- Local-link audit — PASS, 0 missing targets.
- Required-cell row count — PASS, 8.
- Scenario catalog validation — PASS, 22 missions / 44 assertions; explicitly not an execution claim.
- Chromium full-page render — 1280×20649, SHA-256 `da6304e21fad590a36347948b137b1b1b5b739584a23a4592cc08590e815ce42`.
- `git diff --check` — PASS.

## Direct evidence

schema_version: 1
direct_ratio: 6/6
provenance: direct

The checks cover document integrity only. They are not product runtime or Goal verifier completion evidence.

## Review evidence

Three independent read-only audits covered the historical matrix, runnable scenario harness, and Goal-verifier authority boundary. Their shared finding was that no 22/22 same-run Goal-verifier bundle exists; the matrix preserves that failure state.

## Linked issue

- Relates merged #29221.
- Replaces automatically closed stacked PR #29237 after #29240 merged and its base branch was cleaned up.
- Goal verifier worker #29221 and typed authority #29240 are merged; live deployment evidence remains pending.
- Relates RFC-0387.

### Commit lineage (auto-synced)

`docs/keeper-verifier-matrix-mainline` → `main` · 1 commit(s) · synced 2026-08-21T03:01:19Z

| sha | subject | date |
|---|---|---|
| `3db92bbac` | docs(keeper): define Goal verifier real-world matrix | 2026-08-21 |

<sub>Managed by `scripts/pr-sync-body.sh` — do not hand-edit between these markers. The code of record is the diff, not prose above this block.</sub>

PR #29251, opened 2026-08-21
docs(audit): align Goal verifier parity rows

jeong-sik:
## Summary

Remove the stale Goal-completion description from the tracked constitution/code parity matrix.

- Goal completion now names independent `verifier_exact` proof, not human final confirmation or `Awaiting_confirmation`.
- B2 records the durable criterion/proof pending stages.
- B7 records the merged standalone drain worker and keeps artifact lookup plus same-run live evidence explicitly pending.
- The summary points follow-up verification to the living matrix merged by #29244.

## Evidence

- Current source: `lib/goal/goal_phase.mli`, `lib/workspace_goals.ml`, `lib/goal_verification_agent.ml`.
- Implementation PRs: #29221 and #29240 are merged.
- `git diff --check` — PASS.
- Documentation-only follow-up; no local Dune run.

### DIFF
--- a/docs/audits/constitution-code-parity-matrix.md
+++ b/docs/audits/constitution-code-parity-matrix.md
@@ -1,6 +1,7 @@
 # Constitution ↔ Code 정합성 매트릭스
 
 - 생성: 2026-08-20 (14개 영역 스웜 검증 기반)
+- Goal verifier 갱신: 2026-08-21 (#29221, #29240 병합 기준)
 - 상태 값: `대기` / `진행중` / `완료` / `보류`
 - 방향: `문서←코드` (코드에 있고 문서에 없음, 문서 보충) / `코드←문서` (문서에 있고 코드에 없음, 코드 구현) / `코드 정리` (헌법 위반 잔여 제거)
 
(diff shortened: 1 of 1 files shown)


jeong-sik:
Exact-head review on `a7eb24527770bba6fe5d1cf97ecbf43bbd250850`:

**P1 — B3 is not actually source-parity yet.** The proposed row says human final confirmation / `Awaiting_confirmation` is not the current contract, which is true for the `Goal_phase` transition machine, but the live public ledger contract still says the opposite:

- `lib/goal/goal_verification.mli:20-21` says a gated `Completed` goal carries `Human_confirmed` and describes “verifier proof plus human confirmation” as the durable chain.
- The public `completion_state` still exports `Human_confirmed`.
- `record_human_confirmation` is still exported and documented as “Commits the human's final confirmation (B3)” (`goal_verification.mli:154-161`).
- The implementation, codecs, worker drain, workspace reconciliation, and tests still match that variant.

So the repository currently has a split contract: the phase machine/RFC withdrawal no longer requires human confirmation to reach `Completed`, while the public ledger API and its docs retain the abandoned B3 surface. A constitution-to-code parity matrix must expose that discrepancy, not collapse it into “not current.” Either hard-cut the stale `Human_confirmed` API/variant and all readers/tests in a code PR, or mark B3 as partial with that exact cleanup pending.

The PR is documentation-only, so green formatting/CI cannot establish this semantic claim. Please keep it blocked until the row reflects the mixed source truth (or the source is cleaned first).

PR #29283, opened 2026-08-21
ci(keeper): package exact canary runner

jeong-sik:
## Summary
- build `keeper_canary_run` beside the protected-main macOS runtime
- verify its Mach-O arm64 identity and CLI contract
- package it with a manifest-bound SHA-256 companion record

## Why
The 10-turn and duration canary already verifies live HTTP turns, restart/failover, recall, cross-family judgment, and per-turn serving attribution. Packaging the runner from the same protected-main source lets local real-world campaigns bind both server and canary to one CI revision without a local Dune build.

## Checks
- `actionlint -shellcheck= .github/workflows/ci.yml`
- Ruby YAML parse
- `git diff --check`

Stacked on #29281; CI is the compile authority.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1750,7 +1750,10 @@ jobs:
           done
 
       - name: Build macOS arm64 runtime
-        run: opam exec -- dune build --release bin/main_eio.exe
+        run: >-
+          opam exec -- dune build --release
+          bin/main_eio.exe
+          bin/keeper_canary_run.exe
 
       - name: Verify architecture, dylibs, and runtime smoke
         env:
@@ -1758,6 +1761,7 @@ jobs:
         run: |
           set -euo pipefail
           binary=_build/default/bin/main_eio.exe
+          canary_binary=_build/default/bin/keeper_canary_run.exe
           if [ "$RUNNER_OS" != "macOS" ]; then
             echo "expected macOS runner, got $RUNNER_OS" >&2
             exit 1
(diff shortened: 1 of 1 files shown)


jeong-sik:
스택 정리 완료 — exact head `a98104d0bd9c3fba804a9eae16a89dea5ab0a9b2`, base #29281 `dc40455b4265ba398bb7f7370831fd89a69ec800`입니다. canary workflow 커밋 하나만 남도록 restack했고 이전 patch와 `git range-diff =`을 확인했습니다. diff는 `.github/workflows/ci.yml` 한 파일(+51/-2)뿐입니다. ruff, acceptance 14/14, catalog 23/48/A-B-C, diff-check가 통과했습니다. exact-head CI가 진행 중이라 DNM을 유지합니다.

jeong-sik:
Exact-head source review of `d90cf0f130afc1bfdb5b0a2ba1eef7a7173adf4d`: **PASS**.

The workflow builds and packages the exact-source `keeper_canary_run`, verifies arm64 shape and its CLI contract, hashes it into the companion manifest entry, and uploads it with the trusted runtime. A local Mach-O inspection also shows the current main and canary binaries carry the same allowed dylib set.

Residual P2 hardening: CI records and allowlists only the main binary's `otool -L` output. The companion is correct today, but its own linked dylibs are not independently enforced or recorded, so a future dependency drift could escape the artifact contract. This is not a current P0/P1.

Build and Test is still running, so `status:do-not-merge` remains until exact-head CI settles.

jeong-sik:
Correction after an independent exact-head adversarial pass: **NON-PASS (P1)**.

The artifact manifest binds the companion path and SHA, but the packaged canary's own evidence does not carry that build identity. `bin/keeper_canary_run.ml:244-255` runs ambient `git rev-parse HEAD`, and line 990 writes that value as `harness_git_commit`. Running the trusted companion outside its exact checkout—or with no checkout—therefore records an unrelated SHA or null. That breaks the source-bound evidence attribution this PR is meant to enable.

Embed/inject the build source identity into the companion and verify it against the adjacent trusted manifest (or otherwise bind the evidence to the companion digest). The previously noted missing companion `otool` projection remains P2. This finding supersedes my earlier PASS comment on the same head.

jeong-sik:
Post-merge follow-up is in progress for the canary identity defect. The packaged companion must bind its evidence to build/manifest identity rather than ambient `git rev-parse HEAD`, including execution outside a checkout and inside a mismatched checkout. Current main should not be used to claim source-bound canary evidence until that fix lands.

PR #29292, opened 2026-08-21
feat(slack-bot): add Block Kit button interactive handler (block_actions)

jeong-sik:
Adds interactive button support to the Slack Gate Bot (task-331).

- formatters.py: new 'button' structured block type -> Slack actions block with a button element (action_id, value, optional primary/danger style).
- bot.py: @app.action block_actions handler routes each button press to the bound keeper as '[button] <action_id> value=<value>' via Socket Mode (no public Request URL needed).
- Tests: button block projection + missing-field fallback + button_content routing helper.

Refs task-331

### DIFF
--- a/sidecars/slack-bot/src/bot.py
+++ b/sidecars/slack-bot/src/bot.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 import threading
 from datetime import datetime, timezone
 from pathlib import Path
@@ -23,7 +24,6 @@ from typing import Any
 from gate_shared import response_text
 from gate_shared.bindings_store import load_bindings, save_bindings
 from gate_shared.status_store import ConnectorRuntimeStatus, StatusStore
-
 from slack_bolt import App
 from slack_bolt.adapter.socket_mode import SocketModeHandler
 
@@ -37,6 +37,8 @@ logging.basicConfig(
 )
 logger = logging.getLogger("slack-gate-bot")
 
+BLOCK_ACTION_ID_PATTERN = re.compile(r".+")
+
 
 def button_content(action_id: str, value: str = "") -> str:
     """Build the keeper message text for a Block Kit button press.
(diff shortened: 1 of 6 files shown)


jeong-sik:
Exact-head source review of `13e37e541f44a10169e6f4f57c47ea00e2febe40`: **NON-PASS (P1)**.

The button formatter is valid, but the interactive handler is effectively unregistered for generated buttons. Slack Bolt treats a string action constraint as an exact action_id match; regex behavior requires a compiled `re.Pattern`. Therefore `@app.action(".*")` matches only the literal action id `.*`, not `approve_123` or other generated ids.

The tests exercise only the pure content helper and never register/dispatch a Bolt action, so they miss the broken path. Register with a compiled regex (or explicit action ids) and add a handler-dispatch test.

jeong-sik:
A second P1 on the same exact head: distinct clicks on one Slack message share one idempotency key.

The handler forwards only `container.message_ts` as `message_ts` (`bot.py:229,244`). `gate_client.py:70` derives the idempotency key solely from channel plus that timestamp, so every button in the same message—and every later intentional click—collapses to the same key and the Gate rejects later clicks as duplicates. Include a per-interaction identity such as Slack `action_ts`, while keeping retries of the same delivered interaction stable.

The current tests exercise neither Bolt dispatch nor this Gate idempotency boundary.

jeong-sik:
Addressed both confirmed P1 blockers at exact head `39b6575c2e`:

- registered the catch-all Block Kit action listener with a compiled regex (`re.compile(r".+")`), which Slack Bolt treats as a pattern rather than an exact action ID
- derived Gate idempotency from `channel_id + action_id + action_ts`, so delivery retries retain one identity while separate intentional clicks no longer collapse onto the parent message timestamp
- fail closed on malformed action payloads that lack any identity component
- added real Slack Bolt matcher registration coverage plus handler dispatch/retry-vs-second-click and Gate request-key tests

Validation:
- `32 passed` for the complete `sidecars/slack-bot/tests` suite in an isolated env with the sidecar requirements installed
- targeted Ruff syntax/import checks and `ruff format --check` passed
- `compileall` and `git diff --check` passed
- original feature commit preserved across the main rebase (`git range-diff` reports `=`)

The branch was rebased because the previous head was 1,286 main commits behind and CI could not resolve its merge-base even after deepening. Keeping `status:do-not-merge` until exact-head CI and adversarial re-review complete.

jeong-sik:
Independent exact-head source re-review: **PASS**. The compiled Slack Bolt matcher now reaches generated action IDs, and the per-interaction key `channel_id + action_id + action_ts` is stable across retries but distinct across intentional clicks. Incomplete action identity fails closed, and tests exercise real Bolt registration plus Gate key forwarding. All current exact-head checks are green/skipped; DNM remains only because main advanced after this rebase and current-base synchronization still needs confirmation.

PR #29297, opened 2026-08-21
fix(canary): bind evidence to packaged build identity

jeong-sik:
## Summary

- replace ambient `git rev-parse HEAD` in `keeper_canary_run` evidence with the executable embedded `Build_identity.binary_commit`
- keep `masc.keeper_canary_run.v1` and `harness.git_commit` backward-compatible
- cover packaged no-checkout and mismatched ambient-checkout cases

## Why

The canary is shipped as a companion binary in the trusted macOS runtime artifact. Running that binary outside the source checkout, or from an unrelated checkout, must report the source stamped into the binary. The trusted manifest already binds that source SHA to the companion SHA-256; ambient repository state is not build identity.

Follow-up to #29283.

## Validation

- `opam exec -- dune exec test/test_build_identity.exe` — 16/16 pass on exact head
- `opam exec -- dune exec test/test_keeper_canary_evidence.exe` — 18/18 pass; v1 JSON schema unchanged
- `opam exec -- dune build --release bin/keeper_canary_run.exe` — pass on exact head
- generated build stamp and Mach-O binary strings both contain exact head `b84ac4129355b99a349acfe05b08ddcf25eb71a1`
- `opam exec -- dune build @check` — pass before the main rebase; `git range-diff` reports the rebased patch unchanged
- `git diff --check` — pass
- CI trusted-runtime manifest contract statically verified: top-level `source_sha` is checked against `github.sha`, and `companion_binaries.keeper_canary_run.sha256` is checked against the packaged binary digest

## Merge gate

Keep `status:do-not-merge` until independent review and exact-head CI complete.

### DIFF
--- a/bin/keeper_canary_run.ml
+++ b/bin/keeper_canary_run.ml
@@ -234,25 +234,11 @@ let iso8601_now () =
     tm.Unix.tm_min
     tm.Unix.tm_sec
 
-(* With_process.with_process_args_in gives exact argv control (no shell
-   string parsing, unlike Unix.open_process_in) and guarantees the
-   subprocess is reaped on every exit path. The exit status is consumed:
-   a captured line is only trusted when git actually reported success —
-   e.g. a detached-HEAD-less checkout can print nothing on stdout while
-   still exiting 0, and a failing invocation must not be read as "no
-   commit" the same way a genuinely commit-less checkout would be. *)
-let git_commit_opt () =
-  try
-    let lines, status =
-      With_process.with_process_args_in
-        "git"
-        [| "git"; "rev-parse"; "HEAD" |]
-        With_process.drain_lines
-    in
-    match status, lines with
-    | Unix.WEXITED 0, line :: _ -> Some (String.trim line)
-    | _ -> None
-  with _ -> None
+(* This must testify to the canary executable that is running, not to an
+   arbitrary checkout it happens to be launched from.  The trusted runtime
+   manifest binds the companion binary digest to the same build source SHA;
+   [binary_commit] is that SHA embedded by the Dune build rule. *)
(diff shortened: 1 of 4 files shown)


jeong-sik:
Independent exact-head source review of `b84ac4129355b99a349acfe05b08ddcf25eb71a1`: **PASS**.

The fix reads `Build_identity.binary_commit`, whose resolver keeps the embedded build stamp separate from any runtime repository probe. The canary therefore reports the executable's source in both a mismatched checkout and no-checkout launch, while the existing v1 `harness.git_commit` JSON shape remains unchanged. The packaged companion digest remains bound to the trusted manifest's source SHA by the existing workflow.

Residual base P2 remains: the companion's own dylib list is not independently projected into the artifact manifest. That is hardening debt, not a blocker for this identity repair. Keep DNM while exact-head CI is pending.

jeong-sik:
## Current-head adversarial re-review — SOURCE PASS / CI HOLD

Reviewed `b84ac4129355b99a349acfe05b08ddcf25eb71a1`.

The implementation correctly replaces ambient `git rev-parse HEAD` with `Build_identity.current ().binary_commit`, so packaged canary evidence identifies the executable rather than whichever checkout launches it. The wire field remains compatible and the interface now documents the corrected meaning.

- **P2 test gap:** the added tests cover only the pure resolver. No test invokes the canary evidence path and proves serialized `harness.git_commit` comes from the embedded identity—or is JSON null when a binary has no build stamp.

**Feature verdict: source PASS with an integration-test gap.**

**Gates:** Draft/BLOCKED; exact-head Build and Test, CI Regression Gate, and CI Gate are failed. No packaged-runtime or live canary proof was inferred from source tests.

PR #29298, opened 2026-08-21
fix(store): preserve pre-cut goal stores during owner purge

jeong-sik:
## Problem

Merged #29256 removed `goal.owner` and Keeper-meta `active_goal_ids` from the runtime model and closed JSON readers. The deployed pre-merge store still contains `owner` in 48/48 Goal rows (`null` or `string`) and `active_goal_ids` in 35/35 Keeper snapshots (`array`). Deploying the hard cut without a compatibility read would make those stores undecodable.

## Stack dependency

This PR is stacked on #29312 at exact head `e0bbaa4060ee58753ff9178f6ec4f27cb82864c6`. That stack already contains merged #29310, including the current-schema key-count correction from 53 to 52. Restacking therefore dropped only that redundant test hunk; the store-compatibility semantic delta is unchanged.

## Fix

- accept and discard exactly `owner` while decoding a Goal row
- validate `active_goal_ids` through its former array-of-strings contract, then discard it before current Keeper-meta validation
- reject scalar, null, object, mixed-element, duplicate, and unrelated unknown payloads fail-closed
- keep both retired fields absent from types, runtime semantics, and serializers

This is a read-only tombstone, not a restored legacy model. A later legitimate Goal/Keeper metadata mutation rewrites canonical JSON without the retired field; reads do not eagerly mutate live state.

## Evidence

- exact head: `cf9c7b8b7d037edd8c7dc65b74b945033f88e191`
- exact base (#29312): `e0bbaa4060ee58753ff9178f6ec4f27cb82864c6`
- `test_goal_store`: 16/16 pass
- `test_keeper_meta_current_schema`: 11/11 pass
- exact stacked head `dune build @check`: pass
- exact stacked head diff-check: pass
- range-diff differs only by the redundant 53-to-52 key-count hunk already supplied by #29310

## Deployment caveat

Do not deploy the #29256 hard cut without this fix or an out-of-band atomic migration of both stores. This PR deliberately does not modify `/Users/dancer/me/.masc`; canonical cleanup occurs on normal writes after the compatible reader is deployed.

### DIFF
--- a/lib/goal/goal_store.ml
+++ b/lib/goal/goal_store.ml
@@ -72,6 +72,11 @@ and state_of_yojson = function
 
 and goal_of_yojson = function
   | `Assoc fields as json ->
+      (* [owner] was persisted by the immediately preceding schema.  It no
+         longer has any runtime meaning, but existing stores must remain
+         readable across the hard cut.  Keep this as a read-only tombstone:
+         the decoder accepts and drops it, while [goal_to_yojson] never writes
+         it again.  Every other field remains closed-schema strict. *)
       let accepted_fields =
         [ "id"
         ; "title"
@@ -82,6 +87,7 @@ and goal_of_yojson = function
         ; "phase"
         ; "last_review_note"
         ; "last_review_at"
+        ; "owner"
         ; "created_at"
         ; "updated_at"
         ]
--- a/lib/goal/goal_store.mli
+++ b/lib/goal/goal_store.mli
@@ -7,7 +7,10 @@
 
     - a {!Goal_phase.t} (canonical lifecycle: [Executing] / [Blocked] /
       [Completed] / [Paused] / [Dropped]) — the only persisted
-      lifecycle representation.  ["status"] is not an accepted field:
+      lifecycle representation.  The retired ["owner"] field is accepted
+      and discarded on read so pre-cut stores remain available, but it is
+      never written and does not re-enter the runtime model.  ["status"] is
+      not an accepted field:
       a row carrying it fails to decode, and {!read_state} then applies
       the corrupt-store policy (recovery mirror if usable, otherwise an
       empty state plus a warning).  A store written before the field was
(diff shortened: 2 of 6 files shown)


jeong-sik:
Independent exact-head source re-review: **PASS**. The P0 compatibility blocker is cured without restoring retired semantics. Goal `owner` is decode-only and omitted by the writer; Keeper `active_goal_ids` is stripped before exact validation, while duplicate tombstones and unrelated unknown fields still reject. Keep the `p0` hotfix priority and DNM until exact-head Build and Test completes; do not deploy the hard cut first.

jeong-sik:
Independent exact-head source re-review PASS at cf9c7b8b7d037edd8c7dc65b74b945033f88e191 on #29312. The Keeper active_goal_ids tombstone now validates through the strict string-list decoder before drop; scalar/null/object/mixed payloads, duplicates, and unrelated fields fail closed. Runtime types and writers remain purged. No P0/P1 found; exact-head CI is still running.

jeong-sik:
## Current-head adversarial re-review — HOLD

Reviewed new exact head `bb68844961a48105160f4441d0937a2d22ee2859`. The previous stack blocker is closed: this head is now an exact descendant of #29312 `62b4d1442491ad8c32cc11048689517f6e5464f3`, and the unique diff is back to six store files.

One source blocker remains:

- **P1 — Goal `owner` tombstone still accepts malformed or duplicate retired data.** `goal_store.ml:73-152` adds `owner` to the accepted key list and then discards it without checking its former type or duplicate count. `owner: 7`, `owner: {}`, or two `owner` keys become valid current Goals and disappear on canonical rewrite. The new test covers only a valid string owner and an unrelated unknown field.
- The same PR’s `active_goal_ids` tombstone correctly demonstrates the required pattern: validate the retired string-list shape, reject duplicates/mixed/wrong types, then drop it.

Apply the same strict decode-only tombstone contract to Goal `owner` and add malformed/duplicate tests.

**Feature verdict:** restack PASS; persistence compatibility HOLD.

**Gates:** this head was rewritten after the previously green run, so only current-head checks count. No live store mutation or deployment proof.

PR #29311, opened 2026-08-21
fix(tui): align task, keeper, and planning read contracts

jeong-sik:
## What

- read Overview tasks from the canonical aggregate backlog and keep Task status typed
- project Keeper rows through the current metadata schema and canonical dotted-name classifier
- decode Planning through Goal_phase, including verifying rollup, with no retired running alias
- keep task and Keeper read failures visible instead of presenting them as empty collections
- resolve Keeper metrics through the same cluster-aware workspace configuration

## Why

The TUI still read retired per-task files and retired Keeper fields, and its Planning decoder expected the removed status contract. Those mismatches made core surfaces empty or misleading on current source.

## Verification

- `scripts/dune-local.sh build bin/masc_tui.exe test/test_tui_decode.exe test/test_tui_http_ast.exe`
- `test_tui_decode`: 25/25 passed
- `test_tui_http_ast`: 9/9 passed
- `ocamlformat --check` on touched OCaml files
- `git diff --check`
- three independent read-only reviews, followed by targeted re-review of every blocker

## Live/runtime note

Live smoke evidence is intentionally separate from source and local tests. At 2026-08-21 21:39 KST, `/health?full=1` reported server commit `ddb2c03e7f42589ddca5b5136df451922bf8a297`, while this branch is based on current `main` `58ee736e5cddada0935f38da491f9711b0dc856e`. The exact-current TUI showed canonical backlog tasks and the six-phase Planning rollup against that runtime. Existing live Keeper metadata still carries removed `active_goal_ids`, so the current-schema reader reports reset required rather than adding a compatibility fallback.

Approvals, Keeper chat protocol, and renderer performance are intentionally left for follow-up Draft PRs.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -230,6 +230,7 @@
   fs_compat
   time_compat
   masc_log
+  masc_goal
   masc_workspace
   masc_backend
   masc_process
--- a/bin/masc_tui.ml
+++ b/bin/masc_tui.ml
@@ -828,6 +828,7 @@ let run_with_eio_context f =
   Eio.Switch.run @@ fun sw ->
   Eio_guard.enable ();
   Eio.Switch.on_release sw Eio_guard.disable;
+  Fs_compat.set_fs (Eio.Stdenv.fs env);
   Eio_context.set_env env;
   Eio_context.set_switch sw;
   Eio_context.set_net (Eio.Stdenv.net env);
--- a/bin/masc_tui_ansi.ml
+++ b/bin/masc_tui_ansi.ml
@@ -99,10 +99,12 @@ let status_color status =
 (** Task status icon *)
 let task_status_icon status =
   match status with
-  | "done" | "completed" -> "\xe2\x97\x8f"  (* filled circle *)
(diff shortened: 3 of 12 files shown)


jeong-sik:

Reviewed `c16f08e5f55c213689f5545c26499cec28e97eb8`.

- **P1/security — free-form Keeper blocker detail can execute terminal control sequences.** `tui_decode.ml:212-216` concatenates `blocker.detail` without sanitization; `masc_tui_render.ml:875-877,893-896` renders it directly; `masc_tui_ansi.ml:174-180` only width-fits and preserves control bytes. Provider/tool-derived blocker text can therefore inject ANSI/OSC sequences into the operator terminal.
- There is no renderer test with C0/C1/ESC/OSC payloads on this path.

Sanitize external text at the terminal boundary while preserving a typed raw diagnostic separately if needed.

**Feature verdict:** runtime/task/planning projections are structurally wired, but the new operator-facing blocker projection is unsafe.

**Gates:** Draft/BLOCKED; Build and Test, CI Regression Gate, and CI Gate are failed. No terminal/live proof.

PR #29315, opened 2026-08-21
fix(dashboard/fusion): sort run list on a single start-time axis

anyang-keepers:
Owner-reported (board p-94f04e8520ab02a26d389444a72b71e6): the fusion deliberation run list sorts arbitrarily — neither by date nor by result-update order.

Root cause: buildMergedRuns in dashboard/src/components/fusion/fusion-surface.ts sorted board-sink runs by createdAt (completion time) and registry-only runs by startedAt (start time) — a mixed axis that reads as arbitrary.

Fix: unify every row on a single start-time axis by looking up the registry's startedAt for board runs too, falling back to board post creation only when no registry row exists.

- dashboard/src/components/fusion/fusion-surface.ts (single file)

Draft PR — evidence for task-431 verification.


jeong-sik:
## Current-head adversarial review — HOLD

Reviewed `0ff31c8487b5c076d7f847e857e61cebae86f4fe`.

- **P1 — the list still mixes start and sink-creation axes.** `fusion-surface.ts:425-433` uses registry `startedAt` only while a matching registry record exists, then falls back to board `createdAt`. The registry retains only the latest 64 completed runs (`fusion_run_registry.ml:33`) while the board surface reads up to 500 posts. Older/evicted board rows therefore sort by completion/creation beside retained rows sorted by start—the mixed-axis defect this PR claims to eliminate.
- The test at `fusion-surface.test.ts:1118` codifies an unmatched board fallback but never proves a board run with a registry start different from board creation uses one durable axis end-to-end.

Persist the run start in board evidence, or define one honest axis available to every row.

**Gates:** Draft/CLEAN; exact-head CI Gate is green. The source contract remains partial despite green CI. No browser/live proof.

PR #29313, opened 2026-08-21
feat(tui): make operator approvals actionable

jeong-sik:
Stacked on #29311. Review this PR after its parent.

## Summary
- read the exact actor-scoped Operator pending-confirm envelope and fail closed on missing, unscoped, or drifted data
- bind confirm/deny responses to the submitted token and decision, distinguish completed, deferred, execution-failed, and unverified outcomes
- serialize approval refresh/action generations so stale observations cannot resurrect consumed rows
- make the two-key safety gate and in-flight state visible, and show action payload evidence at 80 columns
- invalidate Operator and Dashboard snapshot caches at pending-confirm mutation boundaries

## Focused validation
- scripts/dune-local.sh build bin/masc_tui.exe test/test_tui_decode.exe test/test_tui_operator_projection.exe test/test_tui_http_ast.exe test/test_operator_control_actions.exe
- test_tui_decode: 25/25
- test_tui_operator_projection: 6/6
- test_tui_http_ast: 10/10
- test_operator_control_actions: 8/8
- ocamlformat --check
- git diff --check

## Live read-only evidence
The local server returned an actor-filtered masc-tui envelope with filter_active=true and visible=total=hidden=0; no mutation was performed.

PR #29316, opened 2026-08-21
feat(tui): define exact Keeper chat contract

jeong-sik:
## Summary

- add a typed, current-only Keeper chat request/response projection
- validate exact request, thread, run, message, lifecycle, and terminal identities
- fail closed on partial, malformed, legacy, or unknown SSE events
- classify verified rejection/failure separately from an unverified outcome
- project exact read-only operation reconciliation for safe transport recovery
- sanitize terminal-bound text and retain collision-resistant compact request labels

## Verification

- `scripts/dune-local.sh build test/test_tui_keeper_chat_projection.exe`
- `_build/default/test/test_tui_keeper_chat_projection.exe` — 20/20 passed
- `bash scripts/audit-ci-test-targets.sh` — 381 targets declared; unwired baseline unchanged at 516
- independent contract/reconciliation audit: clean

## Stack

- base: #29313 (`fix/tui-operator-approvals`)
- follow-up: `fix/tui-keeper-chat-stream`

## Scope

This PR is the pure Keeper contract and test layer. It intentionally supports only the current typed wire contract and adds no legacy compatibility or UI effects.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -205,6 +205,12 @@
 
 ;; TUI Dashboard - Terminal interface for multi-agent workspace
 
+(library
+ (name masc_tui_operator_projection)
+ (wrapped false)
+ (modules masc_tui_operator_projection)
+ (libraries masc masc.operator masc.masc_core yojson))
+
 (executable
  (name masc_tui)
  (modules
@@ -218,6 +224,7 @@
  (package masc)
  (libraries
   masc
+  masc_tui_operator_projection
   masc.dashboard
   masc.server
   masc.config
--- a/bin/dune
+++ b/bin/dune
@@ -230,6 +230,7 @@
   fs_compat
   time_compat
   masc_log
+  masc_goal
   masc_workspace
   masc_backend
   masc_process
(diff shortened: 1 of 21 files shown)


jeong-sik:
## Current-head adversarial review — CONTRACT PASS / STACK HOLD

Reviewed `96f82b4b90677ae003021373128f72bbf7c4e099` on exact parent #29313.

The current-wire Keeper chat decoder is strict about acceptance, event order, terminal states, run/thread/request identities, and unknown fields. The contract layer is not user-reachable until #29317.

- **P2 — reconciliation is not bound to the original message.** `masc_tui_keeper_chat_projection.ml:772-845` validates `operation_id` but accepts `source` and `input` merely as object/null. `same_request_identity` compares only request ID + Keeper, and the test explicitly treats a changed message as the same identity. A corrupted/colliding durable row can settle the wrong message. Validate the canonical input’s Keeper/message digest or an exact request fingerprint.

It also inherits #29311’s terminal-safety blocker and red CI.

**Gates:** Draft/UNSTABLE; Build and Test, CI Regression Gate, and CI Gate failed. No live stream proof.

PR #29317, opened 2026-08-21
feat(tui): make Keeper chat responsive and recoverable

jeong-sik:
## Summary

- move Keeper chat submission off the input loop onto a daemon fiber and typed UI mailbox
- correlate completion by Keeper plus request identity so stale results cannot mutate newer state
- preserve per-Keeper drafts and the selected Keeper across roster refreshes
- retain uncertain requests, block fresh-ID resend, and make Ctrl-R poll the exact operation via GET without issuing another chat POST
- render typed request-correlated history with terminal-safe text and newest-reply retention on narrow screens
- keep input visible while a request is in flight and surface deleted-Keeper, failure, replay, and reconciliation states explicitly
- wire the touched TUI HTTP regression suite into focused CI and ratchet the unwired baseline down

## Verification

- focused `masc_tui.exe` plus six touched test executables build passed
- Keeper contract 20/20
- message layout 5/5
- TUI HTTP AST 10/10
- TUI decode 25/25
- operator projection 6/6
- operator control actions 8/8
- `bash scripts/audit-ci-test-targets.sh` — 383 targets declared; unwired baseline ratcheted to 515
- two independent UI and Keeper audits: clean after one whitespace-tail fix

## Stack

- base: #29316 (`fix/tui-keeper-chat-contract`)
- depends transitively on #29313 and #29311

## Known follow-up

The initial chat transport remains responsive but buffers the SSE response body before strict terminal decoding. Incremental token painting, exact terminal-cell width, grapheme/IME input, and renderer coalescing belong to the next performance/input stack.


jeong-sik:

Reviewed `2e426dd7694f447bf17c7ce19a71172f02e3d8d0` on exact parent #29316.

- **P1 — a stalled chat permanently wedges the single send lane.** `masc_tui.ml:209-240` sets `msg_inflight` and blocks every later send. The SSE POST has no timeout or UI cancellation handle (`masc_tui_http.ml:86-110`); reconciliation recursively polls pending operations forever (`masc_tui.ml:270-296`). An open socket with no terminal event, or a never-settling operation, leaves the lane blocked until the TUI exits.
- **P1 — the uncertainty guard is process-local.** `msg_unverified` exists only in mutable TUI state (`masc_tui_types.ml:264-313`). Restarting after an unknown outcome forgets the exact request ID and removes the new-send fence, allowing a fresh-ID duplicate despite the documented “never a second POST” contract.
- Tests cover decoder states and daemon wiring, not hung transport, bounded reconciliation, cancellation, or restart recovery.

Persist the unverified request identity before POST, restore/reconcile it on boot, and add bounded/cancellable liveness behavior without replaying the effect.

**Gates:** Draft/UNSTABLE; Build and Test, CI Regression Gate, and CI Gate failed. It also inherits #29311/#29316 findings. No live chat proof.

PR #29319, opened 2026-08-21
perf(tui): coalesce dirty terminal frames

jeong-sik:
## Summary

- replace unconditional 100 ms repainting with typed `Input | Background | Force` dirty-frame requests
- coalesce background and byte-input bursts inside a monotonic 16 ms window while allowing input to preempt a pending background frame
- wake input polling at the next frame deadline and apply async results within a maximum 16 ms poll window
- cache terminal dimensions and reprobe only after a main-fiber `SIGWINCH` handoff
- use monotonic deadlines for both rendering and periodic refresh
- retry interrupted terminal reads within the original monotonic deadline
- keep 1-column resize total by clamping derived renderer widths

This adopts the scheduling and resize separation from pi while keeping MASC surface rendering unchanged: https://github.com/earendil-works/pi/blob/a2f369d63a3e412a68a2570583dd60cadb7d63a5/packages/tui/src/tui.ts#L764-L824

## Verification

- focused `masc_tui.exe`, scheduler test, and TUI AST test build passed
- render scheduler 9/9
- TUI HTTP/AST regression 11/11
- `bash scripts/audit-ci-test-targets.sh` — 384 targets declared; unwired baseline unchanged at 515
- ocamlformat and `git diff --check` passed
- adversarial scheduler review: APPROVE

## PTY evidence

Against parent head `2e426dd769` in an isolated 80x24 PTY:

| Probe | Parent | This PR |
|---|---:|---:|
| settled idle, 1 second | 9 frames / 18,027 bytes | 0 frames / 0 bytes |
| Tab response | 4 ms | 0 ms |
| following 1 second | 10 frames | 1 frame |
| resize to 1x1 | `Invalid_argument` termination | rendered once and stayed alive |

Normal resize to 31x96 remained alive and rendered a 96-column border on both branches. CPU usage and exact SIGWINCH-to-frame latency were not measured.

## Stack

- base: #29317 (`fix/tui-keeper-chat-stream`)
- depends transitively on #29316, #29313, and #29311

## Follow-up

Row-differential output plus synchronized-output framing is intentionally a separate PR. Unicode cell width, grapheme input, and bracketed paste follow on top of that boundary.

PR #29322, opened 2026-08-21
fix(tui): project current Keeper observations

### LINKED ISSUE #29320: TUI: decode current keeper metrics and turn-record context
## Symptom

Keeper Logs and Live Context can be empty or show zero against a current runtime even when the Keeper has recent `keeper.metrics.v1` rows.

## Reproduction

1. Run a current Keeper until it writes Turn and Heartbeat metrics.
2. Open the Keeper Logs or Keeper detail Live Context surface in `masc-tui`.
3. Current rows are rejected by `Tui_decode.parse_log_entry`, so logs disappear and context falls back to zeros.

## Current contract mismatch

- `lib/tui_decode.ml` requires `context_ratio`, `context_tokens`, and `context_max` on every metrics row.
- Current Turn rows from `lib/keeper/keeper_unified_metrics_snapshot.ml` no longer contain those fields.
- Current Heartbeat rows from `lib/keeper/keeper_heartbeat_snapshot.ml` do not contain them either.
- The ledger discriminator is the closed `keeper.metrics.v1` `Turn | Heartbeat` contract in `lib/keeper/keeper_metrics_record.ml`.
- Current context occupancy is projected from the newest trace-matched TurnRecord by `lib/keeper/keeper_context_observation_projection.ml`, not from the metrics ledger.

## Expected

- Decode the exact current metrics discriminator into a typed Turn/Heartbeat projection.
- Render fields only where the corresponding row kind owns them; preserve missing values as `None` rather than fabricated zeros.
- Source Live Context from the current trace-matched TurnRecord projection and show typed unavailable reasons.
- Reject versionless/unknown metrics rows without legacy compatibility.

## Tried / scope

This was found while replacing the TUI whole-file metrics scan with canonical bounded dated-JSONL reads. The I/O/order slice does not change the public metrics projection because mixing schema correction with storage traversal would obscure both review and regression evidence.

```masc-triage
kind: defect
area: observability
impact: blinds-operator
root: ssot
must-do: false
```

jeong-sik:
Closes #29320

## Summary
- decode only current keeper.metrics.v1 Turn and Heartbeat rows, including producer-derived usage trust, totals, modes, channels, and tool-call invariants
- source live context from the newest current-trace TurnRecord projection with typed partial and unavailable states
- render stable 76-cell log rows that preserve observed zeroes and fit the default 80-column terminal without ANSI-width corruption
- add a behavior seam for current-trace binding, decode-error exclusivity, and empty-roster clearing
- update TUI docs and wire all new decoder/state/layout coverage into focused CI

## Verification
- focused build: TUI binary plus decoder, context-state, observation-layout, HTTP AST, context producer, dashboard metrics, and keeper status targets
- 63 directly run focused tests passed (28 decoder, 4 layout, 3 context state, 11 AST, 9 frame scheduler, 8 context producer)
- CI target audit: 387 declared targets; 514 unwired at the ratcheted baseline
- ocamlformat --check and git diff --check passed
- three independent read-only re-audits reported no remaining findings

## Stack
- base: #29319
- next: bounded strict metrics tail; intentionally not included here

This remains Draft because the inherited base stack has unrelated failing checks and is not merge-ready.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -224,6 +224,18 @@
  (libraries masc.masc_core))
 
 (library
+ (name masc_tui_context_state)
+ (wrapped false)
+ (modules masc_tui_context_state)
+ (libraries masc masc.masc_workspace yojson))
+
+(library
+ (name masc_tui_observation_layout)
+ (wrapped false)
+ (modules masc_tui_observation_layout)
+ (libraries masc))
+
+(library
  (name masc_tui_render_schedule)
  (wrapped false)
  (modules masc_tui_render_schedule))
@@ -241,8 +253,10 @@
  (package masc)
  (libraries
   masc
+  masc_tui_context_state
   masc_tui_keeper_chat_projection
   masc_tui_message_layout
+  masc_tui_observation_layout
   masc_tui_operator_projection
   masc_tui_render_schedule
   masc.dashboard
--- a/bin/masc_tui.ml
+++ b/bin/masc_tui.ml
@@ -1007,7 +1007,7 @@ let main () =
                 if state.keeper_cursor < List.length state.keepers - 1 then begin
                   state.keeper_cursor <- state.keeper_cursor + 1;
                   (match List.nth_opt state.keepers state.keeper_cursor with
-                   | Some k -> load_live_context state base_path k.k_name
+                   | Some k -> load_live_context state base_path k
                    | None -> ())
                 end
             | Keepers Keeper_detail ->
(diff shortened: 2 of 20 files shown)


jeong-sik:

Reviewed only current head `394c46864d92403419b74f78b79538656267fc63`; the superseded `44d005…` head was discarded.

- **P1 — the log “tail” is reversed and opens on oldest rows.** `masc_tui_loader.ml:107-143` reads newest files first, prepends chronological chunks, then `List.rev` reverses them. The UI initializes scroll at the end, so long logs open on the oldest page. It also reads only the newest month and cannot backfill a 200-row request across a month boundary.
- **P1 — metrics rows are not bound to the selected Keeper.** `tui_decode.ml:370-376` decodes `name`, `trace_id`, and `generation` into throwaway variables; `load_keeper_logs` supplies no expected Keeper. A misfiled row is rendered as belonging to the selected Keeper.
- Missing tests: multi-file ordering, month boundary, initial scroll, and Keeper-name mismatch.

The current-context projection itself is trace-bound/fail-closed, but the log/metrics feature is operationally unreliable. This head also inherits the earlier TUI stack blockers.

**Gates:** Draft/UNSTABLE; Build and Test is still in progress. No live TUI proof.

